PR #15, opened 2013-04-11
systemd service file for rabbitmq-server

jens-na:
A service file for systemd. In the file rabbitmq-env.conf are the environment variables like MNESIA_BASE=/var/lib/rabbitmq/mnesia.


michaelklishin:
I'm not sure this has to be a part of this repo.

Please start a conversation about this proposed change on `rabbitmq-users`. Thank you!

PR #19, opened 2013-08-08
Added basic support for MINGW in the server script

q42jaap:
Mingw seems to start RabbitMQ just fine with this addition.


michaelklishin:
Looks like this PR is incomplete. Please consider submitting it again next week as we are preparing to make GitHub repositories primaries (instead of just mirrors).

PR #24, opened 2013-10-25
Allow use of long or short names via a variable in rabbitmq-env

dezmodue:
Very simple change to make it possible running with long names when needed


michaelklishin:
This was fixed on the branch `bug18626`.

PR #29, opened 2014-01-16
Expiring queues should dead-letter messages when a dead-letter-exchange is defined

arobson:
### Present Behavior
Even when a dead-letter-exchange is defined for a queue that has a TTL set (x-expires), when the queue expires, the messages are lost.

### Use Case
One approach to providing isolation guarantees in addition to horizontal scalability is to allow every service instance to have it's own queue. Unfortunately, if the connection is severed between the service instance and its queue, (or the instance crashes) messages are 'orphaned' in the queue with no elegant way to detect this issue and recover the messages. (per-message TTL is very problematic because its dependent upon average message process time and queue depth)

### Test Case
In a system that needs to provide some level of isolation guarantees (say by using a consistent hashing exchange using correlation id for hashing), each new service instance would define its own queue bound to the consistent hash exchange with a TTL and the consistent hash exchange as the dead-letter-exchange.

Messages were published to the consistent hash exchange and split amongst queues. One queue's consumer is terminated and after the TTL expires, the queue is removed and all messages are republished to the consistent hash exchange and effectively redistributed to the remaining queues.

### Request For Comments
Please let me know if this behavior would be undesirable, it fits our use case perfectly but if there's a different set of configurations we could use to get the same benefits, we'd like to know about those.

If there are problems with the code as its written, I'd be happy to make any changes necessary in order for this PR to be suitable for further consideration.

If my implementation is so problematic that it would be easier to just ignore and have a core committer add the feature properly, I totally understand. I tried it partly for the learning experience and partly out of a desire to potentially contribute to an excellent piece of software I've been using every day for a long time.

### Workflow
It looks like GitHub is not the normal avenue for submitting changes but I'm far less familiar with Hg and wasn't sure how to (or if I could) create a ticket in your system for a feature in order to match the branch to the ticket number. 

It says a lot about RabbitMQ's team and implementation that I could pull something like this off in a few hours of source diving. Sorry if I've made a big mess, as I mentioned, this would be a huge win for us going forward if we could get this behavior implemented.

### DIFF
--- a/src/rabbit_amqqueue_process.erl
+++ b/src/rabbit_amqqueue_process.erl
@@ -681,24 +681,36 @@ drop_expired_msgs(Now, State = #q{backing_queue_state = BQS,
 
 drop_expired_queue_msgs(State = #q{backing_queue_state = BQS,
                                   backing_queue       = BQ }) ->
-    case is_empty(State) of
-        true -> State;
-        false ->
-            ExpireAll = fun (_) -> true end,
-            {_, State1} =
-                with_dlx(
-                  State#q.dlx,
-                  fun (X) -> dead_letter_expired_msgs(ExpireAll, X, State) end,
-                  fun () -> {Next, BQS1} = BQ:dropwhile(ExpireAll, BQS),
-                            {Next, State#q{backing_queue_state = BQS1}} end),
-            State1
-    end.
+    ExpireAll = fun (_) -> true end,
+    {_, State1} =
+        with_dlx(
+          State#q.dlx,
+          fun (X) -> dead_letter_all_msgs(ExpireAll, X, State) end,
+          fun () -> {Next, BQS1} = BQ:dropwhile(ExpireAll, BQS),
+                    {Next, State#q{backing_queue_state = BQS1}} end),
+    State1.
 
 with_dlx(undefined, _With,  Without) -> Without();
 with_dlx(DLX,        With,  Without) -> case rabbit_exchange:lookup(DLX) of
                                             {ok, X}            -> With(X);
                                             {error, not_found} -> Without()
                                         end.
+
+dead_letter_all_msgs(ExpirePred, X, State = #q{backing_queue = BQ, q = #amqqueue{name = Name}}) ->
+    %% In the event that the dead letter exchange had previous bindings to the queue,
+    %% bindings are removed from the destination before dead-lettering.
+    %% A manual call to remove_bindings for the exchange is necessary for exchanges like
+    %% the consistent hash exchange because they also keep internal routes which
+    %% do not get removed as a result of removing the bindings from the queue.
+    Bindings = [ B || {binding, X#exchange.name, _, _, _}=B <- rabbit_binding:list_for_destination(Name)],
+    rabbit_misc:execute_mnesia_transaction(
+        fun() -> 
+            rabbit_binding:remove_for_destination(Name),
+            rabbit_exchange:callback(X, remove_bindings, transaction, [X, Bindings])
+        end),
+    dead_letter_msgs(fun (DLFun, Acc, BQS1) ->
+                             BQ:fetchwhile(ExpirePred, DLFun, Acc, BQS1)
+                     end, queue_expired, X, State).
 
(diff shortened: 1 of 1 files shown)


arobson:
@michaelklishin - just following up: I realize the edge case _could_ cause performance issues but the same edge case already exists for any queue that is backlogged with a dead-letter exchange and long enough message TTL. In other words: there are already other ways one can create performance problems if one doesn't take the proper precautions when designing their topology.

Would the Rabbit team consider including cautionary language around this feature as part of the documentation? I've built several systems around Rabbit as the messaging backbone - there has not been a single case where message loss was preferable to temporary performance degradation. The easy way to prevent the performance issue with this feature is to ensure your queues aren't being overrun. With a consistent-hash-exchange (or exchange-queue pairs for each service instance), it's rather simple to solve for this by simply spinning up an additional service instance.

Without this feature, we're left with developing an entire subsystem to monitor queues and consumers that waits for a queue's consumer to drop, unbinds the queue from everything, and directs an existing consumer to pick up the slack then monitors the queue so it can be removed when empty. This is actually quite problematic and has the follow undesirable consequences:
 * load imbalance across consumers
 * uneven latency spikes in processing messages from the orphaned queue (since it can't be evenly redistributed)
 * constant load on rabbit's HTTP API, especially if you need to catch the issue quickly to minimize time lost

Distributed mutual exclusion is a very difficult problem to solve. I believe that with the inclusion of this feature (or a selective-consumer feature), Rabbit would be uniquely positioned to remove or at least greatly reduce the role of a distributed locking mechanism. Without this capability, we're left with complex work-arounds and a new set of problems to address.

michaelklishin:
I am positive about including this feature. Let's wait for other team members' opinions.

dumbbell:
I agree this seems a useful feature. It may even be what users expect when using TTL on a queue.

simonmacmullen:
I don't think anyone is debating whether dlx-on-queue-delete is a useful feature in the abstract. It totally is!

My concerns with the design of this pull request are:

1) It doesn't use flow control; therefore it's possible that if the deleted queue is large then it can spam the DLX-ing queues with messages fast enough to use up all available memory and crash the server.

Yes theoretically this issue exists already with messages getting DLXed by other means. But it's quite hard to hit; this patch makes it very easy. There's a good chance deleting any large queue with a DLX set would do it. So this really needs addressing.

2) It doesn't guarantee messages will not be lost when queues with DLX set are deleted (e.g. if you shut down at the wrong time). That's more fixable by documentation though.

I also have concerns with the implementation:

3) The bit about removing bindings first is dubious; other clients could add bindings back after this point. And calling into exchange callbacks is especially dubious. It happens to work for the CHX but I am not at all convinced that it would do the right thing for (e.g.) federation. See @rade's previous discussion here:

http://lists.rabbitmq.com/pipermail/rabbitmq-discuss/2014-January/033354.html

Now, @rade was not too concerned with offering solutions there beyond "it's complicated". But I would suggest that doing this in the termination phase (after `rabbit_amqqueue:internal_delete/1` has already happened) is definitely a preferable way forward.

You still have the issue with "dark queues" he mentioned, but there is now infrastructure that could let you fix that too; we can show `crashed` queues which are there-but-dead, storing the fact that they're crashed in Mnesia: https://github.com/rabbitmq/rabbitmq-server/blob/9388c6613c1789516614077868b1b7417979ddff/src/rabbit_amqqueue_process.erl#L237

This could be extended to support `deleting` queues too.

arobson:
@simonmacmullen - I can *try* to address your concerns but I won't pretend that I understand Rabbit well enough to handle it. It's been over a year since I've even looked at this code, I was actually pleasantly surprised to see the PR hadn't been closed.

If someone on the core team wanted to take on adding the feature properly, addressing your concerns, we can certainly close this out. Otherwise, do you have recommendations on how I could start working towards addressing your concerns and get some team feedback? Would you prefer I just continue making commits to this branch and request review?

Thanks for being open to discussion and for the guidance provided so far.

michaelklishin:
I'll take the liberty to close this. Not because it's a feature we don't want but because the implementation doesn't cover some edge cases we are quite worried about.

PR #52, opened 2015-02-18
Add systemd notification support

jeckersb:
We've been using this patch in Fedora now for a while.  It uses Peter Lemenkov's erlang-sd_notify module (if available) to notify systemd when RabbitMQ has completed startup.

### DIFF
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -335,6 +335,11 @@ broker_start() ->
     Plugins = rabbit_plugins:setup(),
     ToBeLoaded = Plugins ++ ?APPS,
     start_apps(ToBeLoaded),
+    case code:load_file(sd_notify) of
+        {module, sd_notify} -> SDNotify = sd_notify,
+                               SDNotify:sd_notify(0, "READY=1");
+        {error, _} -> ok
+    end,
     ok = log_broker_started(rabbit_plugins:active()).
 
 start_it(StartFun) ->


michaelklishin:
@jeckersb thank you.

@simonmacmullen are you happy with having such conditional code? We have seen some complaints about how our package works with `systemd` recently.

mbroadst:
:+1:

simonmacmullen:
I don't like it as-is since it will presumably cause our xref tests to fail due to a call to a "nonexistent" module. It needs to fool xref like this: https://github.com/rabbitmq/rabbitmq-auth-backend-ldap/blob/83044408f7e0fa6da2969e7532376e37256d66df/src/rabbit_auth_backend_ldap.erl#L304

Other than that it's OK.

michaelklishin:
@jeckersb would it be possible to update this PR as @simonmacmullen suggested above?

jeckersb:
Updated, passes check-xref for me

michaelklishin:
Thank you @jeckersb!

michaelklishin:
`check-xref` passes for me without errors on OS X and Linux.

binarin:
@jeckersb Hi. Is there a need to do this hard way, with the architecture-specific `sd_notify`? Or this change could be replaced with simple `os:cmd("systemd-notify --ready")` ?

mbroadst:
@binarin I'm pretty sure that wasn't available back then, it was still early days for systemd :smile: I'd say go ahead and submit the change, see what the maintainers say!

michaelklishin:
@binarin @mbroadst I'd recommend using `os:comd/1` as a fallback to what we have. We'd be much more comfortable with including such change in a `3.6.x` release than a complete replacement.

WDYT?

mbroadst:
@michaelklishin sounds like the right way to me, unless @binarin has actually seen an issue with the current code

binarin:
Not an issue, but more of annoyance when backporting/building RedHat packages - and looks like RedHat package is primary user of this feature.

But removing code and dependencies is always good )

@lemenkov, what do you think about removing sd_notify dependency from RedHat package altogether, and just executing `systemd-notify` instead?

michaelklishin:
Debian and Ubuntu seem to be adopting systemd as well.

binarin:
Yes, but currently systemd unit in debian uses `Type=simple` so it's not affected by presence/absence of `sd_notify`.

lemenkov:
@binarin RHEL/CentOS and Fedora does use Type=notify (at least since December 2014 according to Git logs). Are you sure you're using the latest package?

lemenkov:
@binarin sorry didn't understand initially what's your problem. I believe this should be fixed in Debian / Ubunty.

michaelklishin:
@lemenkov not to put words into @binarin's mouth but I think the issue is that `sd_notify` support currently requires a separate RPM. Using `os:cmd/1` is a zero-dependency option.

binarin:
@lemenkov Exactly, it's about removing `sd_notify` dependency altogether. I think I'll check that everything is indeed working correctly with `os:cmd/1`, and propose following changes:
- Add `os:cmd/1` as a fallback for `stable` branch
- Remove `sd_notify` and only leave `os:cmd/1` there - for `master` branch

michaelklishin:
Sounds like a plan. We can ask a few `sd_notify` users to test things with a one-off build from master.

michaelklishin:
The original solution was superseded by #573 and #574 (for `3.7.0`).

PR #54, opened 2015-02-24
Track and expose the timestamp property of the first msg in a queue

alexethomas:
### Summary
The timestamp property is part of AMQP but until now has not been used within Rabbit, only passed through.
This patch makes the timestamp of the head message visible in the management stats, so (typically) identifying the oldest message in the queue.
Monitoring systems can then create alerts when service times are exceeding, or about to exceed, prescribed limits.

### Requirement
Service times are a common aspect of a service level agreement that obtains between organisations or parts of an organisation. As such, it is particularly relevant where Rabbit is being used for integration across zones of responsibility such as application support teams or external service providers.

### Alternatives
Tracking this directly via the message timestamp has a number of advantages over alternatives such as monitoring queue depths:
* *Unambiguous:* a large queue depth, a slow ingress rate or an increasing reject count _may_ indicate a significant problem or may just indicate a bursty sending pattern, a temporary slowdown in the consumer or other transient problem. Monitoring such statistics can result in false alarms while still failing to detect a few aged and unprocessed messages.
* *Flexible:* because the AMQP timestamp is provided explicitly it can reflect the time a series of activities started rather the single activity associated with a specific message.
* *Efficient:* this monitoring does not involve message processing and so is much more efficient than alternative such as peeking at the head message or (in some JMS systems) using a QueueBrowser to watch the front of the queue.

### History
* 6/2009 Feature implemented in a JMS system using QueueBrowser at large financial services organisation
* 12/2010 Suggested as a RabbitMQ feature in customer meeting with VMware/Pivotal
* 4/2011 RabbitMQ replaces JMS system, service time alerts are missed by ops staff
* 2/2014 Discussed possible implementation with Rabbit team member
* 7/2014 v1 implementation, reviewed by Rabbit team member - performance concerns as it extended the message_properties record; also impacted backing_queue_behaviour 
* 10/2014 v2 much less intrusive version leaving the timestamp as a backing queue prop
* 2/2015 RabbitMQ moves to GitHub and invites contribs! Temptation proves too great.

### References
* [Similar request for IBM MQ](http://stackoverflow.com/questions/5038834/need-to-get-oldest-mqmessage-from-mqqueue) 
* [Ditto](http://stackoverflow.com/questions/6711609/monitoring-queue-having-old-message) 
* [Ditto for Glassfish OMQ](https://java.net/projects/mq/lists/users/archive/2009-04/message/14)
* [Description of similar feature added to IBM MQ v6](http://www.mqseries.net/phpBB2/viewtopic.php?t=37589&sid=a4202469d8c31ea3c543a12ae36f333a)
* [Description of similar feature in NServiceBus](http://docs.particular.net/nservicebus/monitoring-nservicebus-endpoints)

### TODO
* Current testing script (relying on management plugin) needs replacing by native rabbit-server tests. However, this isn't particularly straightforward as there are currently no similar tests of queue stats (only channel stats) to extend/adapt so will attempt once feature accepted in principle.


michaelklishin:
Thank you.

I can see the need for this feature but have a couple of concerns.

 * Backing queue state was exposed via HTTP API for debugging. It is not meant to be relied on by the users.
 * Throughput impact.

Do you have any data on what kind of throughput effect this change has?

dumbbell:
Alex, thank you, your pull request is well detailed.

You add `head_msg_timestamp` to `#vqstate` but I don't see where you use it. Is it a leftover from your previous patch?

You could squash the two commits into one, so `rabbit_variable_queue.erl.rej` never appears.

michaelklishin:
@dumbbell the entire VQ state is emitted to the stats DB, so yes, it is used.

michaelklishin:
After giving it some thought, I'm inclined to not merge this because

 * I expect it to have a noticeable throughput impact, especially when total queue length in bytes exceeds the amount of RAM.
 * If we merge it, we pretty much cannot remove internal message store stats from the HTTP API.

I'll wait for some data to see if my guess is correct about 1. Not sure what would make me believe exposing internal message store stats for much longer is a good idea.

simonmacmullen:
@michaelklishin I don't know what you mean by "the entire VQ state is emitted to the stats DB" but that field is not used and can safely be removed. Also, while measuring throughput is definitely valuable, but this patch does not look too bad in terms of throughput. Note that it will not page messages in from disk if necessary, just ignore them.

michaelklishin:
Excellent, then lets wait for some benchmark data. The consensus seems to be that we should keep BQ statistics in the HTTP API.

simonmacmullen:
Now, my comments on the pull request itself...

Firstly @alexethomas, thank you for your work here! This looks like a useful feature, and in principle I have no objection to it. In practice there are a few issues with the implementation.

As mentioned, `backing_queue_status` is not really correct, that's a dumping ground of debug information. You can make this into a "real" queue info item by adding `head_msg_timestamp` to `rabbit_backing_queue:info_keys/0` then adding a new head to `rabbit_variable_queue:info/2`. This will then make it available through `rabbitmqctl list_queues name head_msg_timestamp` or similar (so you should then update the manpage, see `docs/rabbitmqctl.1.xml`).

The implementation of `head_msg_timestamp` has a couple of bugs. Firstly, it assumes that (the head of) RPA is newer than DPA, which is newer than Q4, which is newer than Q3. I think of those relations only Q4 to Q3 is guaranteed. You need quite a pathological case to get DPA newer than RPA, but it's trivial for RPA/DPA to be older than Q3/Q4, unacked messages do not inherently live at the head of the queue.

Also, the master branch has a `qi_pending_ack` field as distinct from RPA and DPA (for unacked messages subject to the queue index embedding optimisation), so you need to think about that.

So I think for every source of messages here you need to get the head timestamp and then compare timestamps, trying to find the "newest" head is not going to work.

There's also the issue that some of these sources of messages have the message paged out (i.e. `#msg_status.msg` is `undefined`) (DPA definitely, Q3 if the message is *not* small enough to be embedded in the queue index). Your patch doesn't crash with these, but it does return `undefined`, which suggests that checking DPA is completely redundant, and Q3 rather dubious.

You could page messages in if needed, but you'd certainly have to work harder to argue that the patch has an acceptable performance impact.

Personally I'd forget RPA / DPA / QPA and just define it as working for ready messages only. I might be tempted to also write a peeking version of `queue_out/1` which pages in from Q3 if necessary (since the queue is presumably going to page in from Q3 soon if Q4 is empty) so that empty Q4 does not lead to the info item becoming blank.

I was going to comment on the tests being in Python and needing management, but I see that's already a TODO :-)

alexethomas:
Very impressed with the prompt feedback, thank you all!

@dumbbell Yes, the head_msg_timestamp in #vqstate is a relic of the previous version. Thank you! Removed and previous 2 commits squashed.
@michaelklishin Very happy to benchmark this, would the following be acceptable:
* Single persistent queue
* Single sender (confirms)
* Single consumer
* Java
* Time-to-send-N 1kB messages and time-to-receive N (separate phases)
@simonmacmullen The tricky stuff! 
Per the best home for the head_msg_timestamp stat, I'm certainly happy to promote it to being an info key as you suggest. (I was originally wary of changing anything in rabbit_backing_queue but this was probably a misinterpretation of earlier advice). 
Thanks particularly for the explanation of the likely status of the internal queue heads. Although it's tempting to only include ready messages (as a JMS QueueBrowser impl. would do) I think we can score a win here by including those pending acks. This catches the case where a consumer is in a reject loop and the message spends most of its time pending ack, or pending nack rather, and so remains mostly invisible to monitoring. (On another MQ this situation sometimes resulted in the head_msg_timestamp oscillating between the true head msg value and the next in the queue and so "flapping" the associated alert).
So if you agree I'd like to try peeking all of Q3, Q4, RPA, DPA and QPA heads and finding the oldest timestamp among them for the stat.

alexethomas:
Implemented as per above except better tests still ToDo.
No attempt is made to page messages in as I'm assuming that this will happen "soon enough" for our purposes.
On this basis, no code being added to the send/consume paths, I suggest benchmarking focuses on the monitoring loop only.

michaelklishin:
@alexethomas I'd still do some benchmarks with PerfTest, say with 100 queues, even though stats emission is not on the hot path.

alexethomas:
@michaelklishin Is there a way to use multiple queues in PerfTest? I can see settings for multiple producers and consumers of a single queue.

michaelklishin:
@alexethomas `-y 100` should use 100 server-named queues.

alexethomas:
@michaelklishin Thanks again, just done a couple of runs with that setting and it seems OK on the main path.

**./runjava.sh com.rabbitmq.examples.PerfTest -z 10 -y 100**

I also tried adding a timestamp to the Producer to see if that would make a difference here, and it does - a slowdown of about 6% - but the slowdown affects master and patched versions equally.

### master (commit bcb88a2e53ac7e2c63e5a5cbf7346c193992cf1b)
#### run 1
sending rate avg: 230 msg/s
recving rate avg: 23041 msg/s
#### run 2
sending rate avg: 225 msg/s
recving rate avg: 22544 msg/s

### sla_tracking_v2 (commit 6317a384b0ac5fc4490f8668bad74866f7d8c92a)
#### run 1
sending rate avg: 233 msg/s
recving rate avg: 23396 msg/s
#### run 2
sending rate avg: 243 msg/s
recving rate avg: 24376 msg/s

michaelklishin:
@alexethomas thank you. That's re-assuring.

simonmacmullen:
So this is moving in a decent direction, cool!

If you are not going to page messages in for this (which is probably the right thing to do, or rather not do) then:

* There's no point checking `disk_pending_ack` at all, messages in that are guaranteed to be paged out.
* It might be worth only checking q3 if q4 is empty, since q3 can't contain newer messages than q4.
* The manpage should mention the fact that paged out messages are ignored for this, somehow.

Also:

* The final return from the info item should be `''` (i.e. the atom with a zero-length name) not `undefined` if there is no value. Yes, that's weird, but it's how our info items work.
* Indentation is currently pretty weird. "Whatever emacs does" would be a good start :-)
* Since this isn't paging messages in, I'm not really worried about performance.

alexethomas:
Changes made as per above. 

I've simply said in the doc that timestamps for messages paged out won't appear, and I'm assuming that they'll "soon" be paged in - is that valid?

(Re q4 and q3 messages, q3 can't contain *earlier* messages than q4, I think, hence skipping q3 unless q4 is empty).

Also now re-indented as per Emacs.

simonmacmullen:
I would rephrase "The message must be paged-in from disk to appear" since a determined user could misinterpret that as meaning "inspecting this info item could cause extra messages to be paged in". Maybe "Messages which are paged out will not contribute to this field" or similar?

I think this is nearly ready. If you are going to keep the tests they need to be moved to Java or Erlang. Having said that, we haven't had automated tests for info items in the past. Maybe we should though.

alexethomas:
Ah right, yes it is ambiguous, thanks. How about "Timestamps of messages only appear when they are in the paged-in state."?

The idea of Java tests is tempting if they would be acceptable - any objections? I suppose direct Erlang tests could avoid going through the mgmt plugin but will be harder to wrangle.

michaelklishin:
We would certainly prefer tests in Erlang because no other tests currently requires the HTTP API to be enabled.

simonmacmullen:
That phrasing is fine.

Actually either Java or Erlang tests could work - Erlang tests have the advantage that they can bypass invoking `rabbitmqctl` and just RPC to `rabbit_amqqueue:info_all()` or whatever, but we do have some Java tests which invoke `rabbitmqctl list_*` operations too - see https://github.com/rabbitmq/rabbitmq-java-client/blob/master/test/src/com/rabbitmq/tools/Host.java#L148

alexethomas:
OK I will have a look at doing Erlang tests first. 

In principle it doesn't look too hard to extend the existing test_statistics fn but I'm wary of simply throwing more checks into the existing logic so think I should clone it. I can follow the same portmanteau style of multi-step test, so doing the existing 5 little tests in 1.

I'll be travelling for the next few days so may not get around to this for a while but won't forget!

michaelklishin:
@alexethomas just checking (as it has been a month since the last update): is this still something you'd like to finish?

alexethomas:
Hi Michael, yes despite the interval I'm actually still on it, just got diverted to rabbitmq/rabbitmq-auth-backend-ldap#4. I'll be checking in the new tests today or at latest tomorrow.

alexethomas:
@michaelklishin have now replaced the Python tests with Erlang ones so ready for review. I also renamed the stat to head_message_timestamp in line with other stats that have 'message' in full. Happy to merge in current master if that helps.

michaelklishin:
@alexethomas thank you, we'll take a look soon.

michaelklishin:
@alexethomas this PR can no longer merge cleanly with master. Can you please rebase?

alexethomas:
Yes of course, will try that now.

alexethomas:
Resolved.
(The conflict was in rabbit_basic.erl where both my extract_timestamp fn and the new header fns had been added to the exported fns.)

michaelklishin:
When testing this with a queue that has messages in it and no consumers, HTTP API always returns ``
as an empty string:

```
"head_message_timestamp": ""
```

```
curl -u guest:guest http://127.0.0.1:15672/api/queues/%2F/pr-54 | python -m json.tool            
…
{
    "arguments": {},
    "auto_delete": false,
    "backing_queue_status": {
        "avg_ack_egress_rate": 0.0,
        "avg_ack_ingress_rate": 0.0,
        "avg_egress_rate": 1.0165876629440466e-10,
        "avg_ingress_rate": 0.9991605609953158,
        "delta": [
            "delta",
            "undefined",
            0,
            "undefined"
        ],
        "len": 94,
        "next_seq_id": 166,
        "q1": 0,
        "q2": 0,
        "q3": 0,
        "q4": 94,
        "target_ram_count": "infinity"
    },
    "consumer_details": [],
    "consumer_utilisation": "",
    "consumers": 0,
    "deliveries": [],
    "disk_reads": 0,
    "disk_writes": 0,
    "durable": false,
    "exclusive_consumer_tag": "",
    "head_message_timestamp": "",
    "idle_since": "2015-04-10 22:53:18",
    "incoming": [],
    "memory": 142904,
    "message_bytes": 0,
    "message_bytes_persistent": 0,
    "message_bytes_ram": 0,
    "message_bytes_ready": 0,
    "message_bytes_unacknowledged": 0,
    "message_stats": {
        "deliver_get": 163,
        "deliver_get_details": {
            "rate": 0.0
        },
        "deliver_no_ack": 163,
        "deliver_no_ack_details": {
            "rate": 0.0
        },
        "publish": 257,
        "publish_details": {
            "rate": 0.0
        }
    },
    "messages": 94,
    "messages_details": {
        "rate": 0.0
    },
    "messages_persistent": 0,
    "messages_ram": 94,
    "messages_ready": 94,
    "messages_ready_details": {
        "rate": 0.0
    },
    "messages_ready_ram": 94,
    "messages_unacknowledged": 0,
    "messages_unacknowledged_details": {
        "rate": 0.0
    },
    "messages_unacknowledged_ram": 0,
    "name": "pr-54",
    "node": "rabbit@ubuntu",
    "policy": "",
    "recoverable_slaves": "",
    "state": "running",
    "vhost": "/"
}
```

The same happens with `rabbitmqctl`:

```
./scripts/rabbitmqctl list_queues name messages_ready head_message_timestamp
Listing queues ...
pr-54	181
```

I'm publishing messages as transient, `pr-54` is non-durable.

am I missing something?

alexethomas:
Thanks Michael, I've reproduced this but only once! Clearly there's some subtlety about first surfacing the message properties - perhaps on a queue that's never had a consumer - that I need to investigate.

alexethomas:
Strange! I've still only reproduced this once and can't make it do it again.
I've re-added the old python test script in case this helps isolate it (currently that works too of course, at least for me). Any further pointers welcome.
I noticed that your message bodies are empty so thought it might be something to do with that but it appears not.
Of course, I can't rule out that I made a typo in the property name the first time (e.g. left a space after it) and that's caused it not to appear.

alexethomas:
@michaelklishin no luck here - tried various permutations in the python script but couldn't reproduce. I'll ask @ash-lshift to have a go.

michaelklishin:
@alexethomas thank you, I'll give it another go tomorrow.

ash-lshift:
just as another data point, i've run the python script here and all the tests pass.

michaelklishin:
@ash-lshift yes, I plan on giving it another try soon.

michaelklishin:
I still can reproduce the blank value issue with a very basic Ruby script:

``` ruby
conn = Bunny.new; conn.start
ch = conn.create_channel
 => #<Bunny::Channel:70159864418180 @id=1 @connection=#<Bunny::Session:0x7f9ebb9a24f8 guest@127.0.0.1:5672, vhost=/, hosts=[127.0.0.1]>>
q = ch.queue("issue-54", durable: true)
# publishes over default exchange
loop { sleep 0.5; q.publish("", persistent: true) }
```

```
λ ~/ curl -u guest:guest http://127.0.0.1:15672/api/queues/%2F/issue-54 | python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1140  100  1140    0     0   688k      0 --:--:-- --:--:-- --:--:-- 1113k
{
    "arguments": {},
    "auto_delete": false,
    "backing_queue_status": {
        "avg_ack_egress_rate": 0.0,
        "avg_ack_ingress_rate": 0.0,
        "avg_egress_rate": 0.0,
        "avg_ingress_rate": 1.510160092604313,
        "delta": [
            "delta",
            "undefined",
            0,
            "undefined"
        ],
        "len": 40,
        "next_seq_id": 40,
        "q1": 0,
        "q2": 0,
        "q3": 0,
        "q4": 40,
        "target_ram_count": "infinity"
    },
    "consumer_details": [],
    "consumer_utilisation": "",
    "consumers": 0,
    "deliveries": [],
    "disk_reads": 0,
    "disk_writes": 40,
    "durable": true,
    "exclusive_consumer_tag": "",
    "head_message_timestamp": "",
    "incoming": [],
    "memory": 176760,
    "message_bytes": 0,
    "message_bytes_persistent": 0,
    "message_bytes_ram": 0,
    "message_bytes_ready": 0,
    "message_bytes_unacknowledged": 0,
    "message_stats": {
        "disk_writes": 40,
        "disk_writes_details": {
            "rate": 2.0
        },
        "publish": 40,
        "publish_details": {
            "rate": 2.0
        }
    },
    "messages": 40,
    "messages_details": {
        "rate": 2.0
    },
    "messages_persistent": 40,
    "messages_ram": 40,
    "messages_ready": 40,
    "messages_ready_details": {
        "rate": 2.0
    },
    "messages_ready_ram": 40,
    "messages_unacknowledged": 0,
    "messages_unacknowledged_details": {
        "rate": 0.0
    },
    "messages_unacknowledged_ram": 0,
    "name": "issue-54",
    "node": "rabbit@mercurio",
    "policy": "",
    "recoverable_slaves": "",
    "state": "running",
    "vhost": "/"
}
```

So, simply publishing a few messages doesn't make the new field non-null.

michaelklishin:
OK, there's one obvious this I've missed: my messages didn't have a timestamp set on them. Things seem to work as expected now.

michaelklishin:
@alexethomas @ash-lshift merged. Thank you for contributing!

alexethomas:
Terrific, thank you!

PR #61, opened 2015-03-08
Add timeout option to some rabbitmqctl commands

michaelklishin:
Our default (infinity) helps detect stuck processes but there should be a timeout option for operations convenience, see this [rabbitmq-users thread](https://groups.google.com/d/msg/rabbitmq-users/Ja0iyfF0Szw/WPbZ1DOeM5IJ) for example.

We keep the default as is because historically, any picked value wasn't good enough because someone had a slow enough server.


michaelklishin:
We decided to expend this in scope and list items (e.g. queues) that were unavailable or seem stuck.

PR #77, opened 2015-03-18
Alter RabbitMQ database directory permissions for .deb and .rpm packages

### LINKED ISSUE #68: Make internal database files group-readable
They are currently world-readable.

michaelklishin:
* Revoke group writing permission
 * Revoke all "others" permissions

It was decided that doing something similar for generic UNIX packages (in `rabbitmq-server`) would be too fragile because we don't know what UNIX flavour the package may be used on and what are common practices (or available tools) there.

Fixes #68.


dumbbell:
You can do the chmod in a single run:
```sh
chmod -R g-w,o-rwx .../mnesia
```

PR #79, opened 2015-03-20
Limit x-death growth to one entry per queue

### LINKED ISSUE #78: Limit x-death header length growth
Original bug: 26538.

See https://groups.google.com/forum/#!topic/rabbitmq-users/dwdE-DySy3E.

After discussing a few options, we've decided to cap the growth by having one value per queue (the most recent one).

michaelklishin:
Otherwise the list can grow forever in some cases.

Fixes #78.


dumbbell:
This PR fixes #78.

Issue #74, opened 2015-03-17
Put channel log messages in their own log category


dumbbell:
Reproduce what is done for connection log messages.

This should be extended to other message sources.

dumbbell:
The problem was reported on the mailing-list:
https://groups.google.com/d/topic/rabbitmq-users/F-Fbc7sxTrY/discussion

dumbbell:
An update to the documentation on the website is ready:
rabbitmq-website@735c99268a1b619b14e45cd94b12a25dade05b1c

michaelklishin:
@dumbbell perhaps you mean https://github.com/rabbitmq/rabbitmq-website/commit/735c99268a1b619b14e45cd94b12a25dade05b1c.

dumbbell:
Yes, thanks!

michaelklishin:
@dumbbell is this ready for QA? I don't see a pull request for it.

dumbbell:
@michaelklishin, yes, it's ready.

dumbbell:
Please QA the following PRs:
* rabbitmq/rabbitmq-server#81
* rabbitmq/rabbitmq-website#2

michaelklishin:
I QA'ed this branch yesterday. Both PRs are now merged => closing.

PR #81, opened 2015-03-26
PR for rabbitmq/rabbitmq-server#74

dumbbell:
See rabbitmq/rabbitmq-server#74.


michaelklishin:
I QA'ed this yesterday. Works as advertised, the change is what I expected it to be.

PR #83, opened 2015-03-26
Reusable worker pool

alexethomas:
Change to allow multiple instances of the worker pool to support the LDAP enhancement rabbitmq/rabbitmq-auth-backend-ldap#4. 

This follows the suggestion in the other PR to reuse this rather than roll our own. The other implementation hints have also been taken.

### DIFF
--- a/src/worker_pool_sup.erl
+++ b/src/worker_pool_sup.erl
@@ -18,7 +18,7 @@
 
 -behaviour(supervisor).
 
--export([start_link/0, start_link/1]).
+-export([start_link/0, start_link/1, start_link/2]).
 
 -export([init/1]).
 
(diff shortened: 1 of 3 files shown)


michaelklishin:
Thank you. One thing that stands out immediately is that we should be able to keep the existing API compatible, e.g. keep `start_link/0` which uses a default name.

We'll take a closer look later.

src/worker_pool_sup.erl
```diff
@@ -28,26 +28,29 @@
...
--define(SERVER, ?MODULE).
-
-%%----------------------------------------------------------------------------
-
 start_link() ->
     start_link(erlang:system_info(schedulers)).
 
 start_link(WCount) ->
-    supervisor:start_link({local, ?SERVER}, ?MODULE, [WCount]).
+    start_link(WCount, worker_pool).
```
ash-lshift: we should probably have a header which defines the default worker pool name rather than repeating it here. see also this line in `worker_pool.erl`:
```erlang
-define(DEFAULT_SERVER, ?MODULE).
```

PR #99, opened 2015-04-06
prevent unbounded gc interval backoff

dallasmarlow:
It has been relatively common in our environment to see rabbitmq use up to 10-20x more memory (usually allocated to the queue_procs or mgmt_db heaps) than the canonical queue data should consume after the rabbitmq-server processes have been running for some time with a production workload using versions 3.3.5, 3.4.4, and 3.5.0. The memory growth rate has been slow enough that a manually gc can be run which bring the processes back to a operable state, however running gc more often would be preferable to long intervals of unbounded growth/gc delay.

@michaelklishin let me know if more context or data is needed, any feedback is greatly appreciated!

### DIFF
--- a/src/background_gc.erl
+++ b/src/background_gc.erl
@@ -26,6 +26,7 @@
 
 -define(MAX_RATIO, 0.01).
 -define(IDEAL_INTERVAL, 60000).
+-define(MAX_INTERVAL, 240000).
 
 -record(state, {last_interval}).
 
@@ -70,7 +71,7 @@ terminate(_Reason, State) -> State.
 interval_gc(State = #state{last_interval = LastInterval}) ->
     {ok, Interval} = rabbit_misc:interval_operation(
                        {?MODULE, gc, []},
-                       ?MAX_RATIO, ?IDEAL_INTERVAL, LastInterval),
+                       ?MAX_RATIO, ?MAX_INTERVAL, ?IDEAL_INTERVAL, LastInterval),
     erlang:send_after(Interval, self(), run),
     State#state{last_interval = Interval}.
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
@dallasmarlow mind providing some background/context on this?

dallasmarlow:
@michaelklishin you beat me to it, updating description now

michaelklishin:
No pressure :)

michaelklishin:
@dallasmarlow limiting the interval to 4 minutes sounds reasonable. Thank you!

dallasmarlow:
@michaelklishin thanks, do you think these should be configurable?

michaelklishin:
@dallasmarlow unless more people ask for GC intervals being configurable, I'm fine with what we have.
One improvement I realise we need to is collecting GC interval metrics. While not nearly as important as on the JVM, it's still something we should expose to operations.

We may cherry pick this to `3.5.x` if @dumbbell and @videlalvaro have no objections.

rade:
@dallasmarlow any idea why in your setup the gc would have backed off so much?

rade:
Also, I'd like to question the utility of this whole logic now. With the upper bound set at 3 minutes, and the default at 1 minute, there is preciously little adjusting going on. You might as well ditch all the adjusting and run at a fixed interval.

dallasmarlow:
@rade i'm not sure, but after reading the interval_operation code and seeing the effects of a manual gc during these events it looked like as gc times got slower, the next interval would always get pushed back making the amount of gc work needed next time even more (with a constant load). that was my impression after seeing the graphs and stats after several of these events, and memory usage has been table and never above 10G after patching.

dallasmarlow:
@michaelklishin sounds good, thanks again.

michaelklishin:
@rade yes, makes sense, I'll file an issue.

michaelklishin:
Answering such questions would be a lot easier if GC related metrics were available.

dumbbell:
Changing how/when the garbage collector is called can cause unexpected behaviour. I'm wary about merging this commit to a stable branch.

Issue #88, opened 2015-03-31
Stderr/out redirection in package scripts should respect RABBITMQ_LOG_BASE


michaelklishin:
See [this mailing list thread](https://groups.google.com/forum/#!topic/rabbitmq-users/Ak1v_UQ6_18).

michaelklishin:
Marked as Enhancement because such redirection is not part of the core, so I'm not sure if this can be considered a bug.

PR #103, opened 2015-04-08
PR for rabbitmq/rabbitmq-server#88

dumbbell:
This PR addresses #88. It is against the `master` branch because it changes the behaviour as well as loading `rabbitmq-env` earlier than before. I tested it successfully on Fedora 21 and Ubunth 12.04.5.

@michaelklishin, @videlalvaro, this is ready for QA.


michaelklishin:
Overriding `LOG_BASE` in `rabbitmq-env.conf` works as expected, including for stream redirection.

Issue #96, opened 2015-04-03
Crash dump with standalone OS X package


michaelklishin:
See [this thread](https://groups.google.com/forum/#!topic/rabbitmq-users/3oYTb9feQr8).

Even though RabbitMQ starts and stops successfully, something still produces a crash dump (I'm attaching one).

Steps to reproduce:

 * Run `./sbin/rabbitmq-server -detached` with the standalone OS X package
 * Observe `erl_crash.dump` in the current directory.
 * `./sbin/rabbitmqctl status` suggests the node is fine
 * Running various quick tests suggests the node is fine
 * `./sbin/rabbitmqctl stop` stops the node successfully

videlalvaro:
There's no need to start/stop RabbitMQ to reproduce this issue. Just running rabbitmqctl will generate the crash dump

videlalvaro:
As explained in our internal bug system here https://bugzilla.rabbitmq.com/show_bug.cgi?id=26441#c3 we need to pass the boot file explicitly when using the standalone release **in every place where we call `erl`**.

The code here https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_nodes.erl#L207 doesn't pass this option, therefore Erlang crashes since it can't find the boot script.

Adding: `"-boot", "/path/to/start_clean"` to the list of arguments passed to `open_port` solves the problem. I'll try to fix this once I'm back.

vitaly-krugl:
Good point @videlalvaro, I just noticed that simply running `rabbitmqctl status` produces a crash dump.

michaelklishin:
The `erl` call that fails is supposed to force start `epmd` (if needed) because of certain platform-specific issues. So this is completely harmless.

PR #108, opened 2015-04-14
adds start.boot to bin folder for mac-standalone release

videlalvaro:
This file is required by rabbit_nodes:ensure_epmd/0 in order to be
able to programatically start epmd

Fixes rabbitmq/rabbitmq-server#96


videlalvaro:
Note that we were having an `erl_crash.dump` because `rabbit_nodes:ensure_epmd/0` wasn't finding the `start.boot` script. 

According to the [docs](http://www.erlang.org/doc/man/erl.html) `-boot` defaults to `$ROOT/bin/start.boot` therefore the proposed fix.

michaelklishin:
@videlalvaro this no longer merges cleanly. Please rebase.

michaelklishin:
Cherry-picked to `stable`.

PR #133, opened 2015-04-29
Make gm flow control configurable

### LINKED ISSUE #114: Make gm flow control optional
See #98. It may take a while to reproduce so for `3.5.2` we should investigate making flow control in gm optional.

Relevant commits:

 * 75a472ee850bbc95e50d60f1fed5a13512f8d1e5
 * 061f7339924c8b610840f223a22c56f4646c5c6b
 * 8a605773ce360d3f29edbf639bebbd204d8c7b7a

suggest that it won't be a breaking change, which is fortunate.

michaelklishin:
Fixes #114.

Inter-node flow control sometimes has unexpected effects and it's not clear to some users whether it solves more problems than introduces or not. Fortunately we can make it optional without breaking internal communication protocol. I believe that's what we should until we can begin collecting more metrics on flow control.

Note: this is against the `stable` branch and I really would like to get it into `3.5.2`.

### DIFF
--- a/src/rabbit_channel.erl
+++ b/src/rabbit_channel.erl
@@ -237,10 +237,8 @@ init([Channel, ReaderPid, WriterPid, ConnPid, ConnName, Protocol, User, VHost,
     process_flag(trap_exit, true),
     ?store_proc_name({ConnName, Channel}),
     ok = pg_local:join(rabbit_channels, self()),
-    Flow = case rabbit_misc:get_env(rabbit, mirroring_control, flow) of
-             flow   -> flow;
+    Flow = case rabbit_misc:get_env(rabbit, mirroring_flow_control, true) of
              true   -> flow;
-             noflow -> noflow;
              false  -> noflow
            end,
     State = #ch{state                   = starting,
--- a/src/rabbit_channel.erl
+++ b/src/rabbit_channel.erl
@@ -40,7 +40,9 @@
              queue_consumers, delivering_queues,
              queue_collector_pid, stats_timer, confirm_enabled, publish_seqno,
              unconfirmed, confirmed, mandatory, capabilities, trace_state,
-             consumer_prefetch, reply_consumer}).
+             consumer_prefetch, reply_consumer,
+             %% flow | noflow, see rabbitmq-server#114
+             delivery_flow}).
 
 -define(MAX_PERMISSION_CACHE_SIZE, 12).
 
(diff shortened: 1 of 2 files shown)


videlalvaro:
Code-wise I'm fine with the implementation. Also whenever we modify basic.publish or related methods, we should be very wary about performance impacts (in this case I see we are resorting to the process dictionary as a cache, but since this is a global setting, we could even moving outside basic.publish and have this as another flag inside the channel state, passed during channel creation).

Solution-wise I'm not sure this is the right thing to do. This seems more like avoiding flow control entirely, than solving the issue behind the original bug. Perhaps `26527` did the wrong thing to solve issue? I don't know, but overriding it altogether doesn't seem right.

Also once we put a new feature out, people will use it and _depend on it_, so if this happens to be the wrong thing to do, it will be hard to take it back.

About putting this on 3.5.2, seems rushed, for the previous reasons and because .2 is a Patch release, so no new features should go in there.

michaelklishin:
We cannot move this flag out anywhere for 3.5.x. This is a way for some users to go back to the old 3.4.x behaviour. Cross-node flow control effects are difficult to predict, as you can see in #114. There wasn't enough loading testing as part of bug `26527 `: in fact, if you take a look at the test suite, flow is disabled there.

We don't advertise this setting anywhere. This affects a couple of high profile users. I'm strongly in favour of having this option for 3.5.x. The current gm won't be around forever either, so we can make this permanent.

michaelklishin:
After discussing it via chat it looks like we can move this flag to channel's record while preserving `#delivery`. This is a good idea.

michaelklishin:
@videlalvaro ready for another round.

src/rabbit_channel.erl
```diff
@@ -235,6 +237,12 @@ init([Channel, ReaderPid, WriterPid, ConnPid, ConnName, Protocol, User, VHost,
     process_flag(trap_exit, true),
     ?store_proc_name({ConnName, Channel}),
     ok = pg_local:join(rabbit_channels, self()),
+    Flow = case rabbit_misc:get_env(rabbit, mirroring_control, flow) of
```
videlalvaro: the key name here doesn't match the name in the config ^^

src/rabbit_channel.erl
```diff
@@ -235,6 +237,12 @@ init([Channel, ReaderPid, WriterPid, ConnPid, ConnName, Protocol, User, VHost,
...
+             flow   -> flow;
+             true   -> flow;
+             noflow -> noflow;
+             false  -> noflow
```
videlalvaro: I wouldn't have so many options, either `flow/noflow` or `true/false`

videlalvaro:
Codewise is fine. About merging it into stable for the next patch release, I would wait for @dumbbell's opinion. I particularly don't agree.

michaelklishin:
Several high profile customers want the `3.4.x` behaviour back. This is a change that's compatible with `3.5.x`. I strongly feel this should be in `3.5.2`, otherwise we'd have to ship snowflake builds.

PR #136, opened 2015-04-30
erlang:now/0 => os:timestamp/0 in credit_flow and rabbit_reader

### LINKED ISSUE #135: Switch credit_flow and related rabbit_reader call sites to use os:timestamp/0
`os:timestamp/0` is compatible, is not monotonic (can produce non-unique values), and does not acquire a global lock. This is perfectly fine for `credit_flow` and `rabbit_reader`'s needs.

michaelklishin:
Fixes #135.

### DIFF
--- a/src/credit_flow.erl
+++ b/src/credit_flow.erl
@@ -116,7 +116,7 @@ state() -> case blocked() of
                true  -> flow;
                false -> case get(credit_blocked_at) of
                             undefined -> running;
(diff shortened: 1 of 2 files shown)


michaelklishin:
Taking the liberty to merge this as we've discussed this and I'd like to avoid chaining branches that revolve around `credit_flow`.

PR #139, opened 2015-05-02
Introduce conditional tracing of credit_flow events

### LINKED ISSUE #137: Conditional tracing for credit_flow events
We need to have a way to collect events about processes being blocked and unblocked by `credit_flow`. However, it is important to not introduce runtime overhead while doing so. Since this feature is intended for contributors and not users (at least for now), it was agreed that a pre-processor condition would be sufficient.

michaelklishin:
This introduces conditional (compile time) tracing of `credit_flow` events over `rabbit_event`. This will be used to collect stats on credit flow behaviour with a tool similar to `rabbitmq_event_exchange` (which needs to be extended to convert credit flow events, because they involve compound data structures as keys).

Fixes #137.

### DIFF
--- a/src/credit_flow.erl
+++ b/src/credit_flow.erl
@@ -70,9 +70,15 @@
 
 -ifdef(CREDIT_FLOW_TRACING).
 -define(TRACE_BLOCKED(THIS, FROM), rabbit_event:notify(credit_flow_blocked,
-                                      [{process, THIS}, {from, FROM}])).
+                                     [{process, THIS},
+                                      {process_info, erlang:process_info(THIS)},
+                                      {from, FROM},
+                                      {from_info, erlang:process_info(FROM)},
+                                      {timestamp, os:timestamp()}])).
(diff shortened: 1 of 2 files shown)


src/credit_flow.erl
```diff
@@ -68,6 +68,22 @@
...
+                                      {from, FROM},
+                                      {from_info, erlang:process_info(FROM)},
+                                      {timestamp, os:timestamp()}])).
+-define(TRACE_UNBLOCKED(THIS, FROM), rabbit_event:notify(credit_flow_unblocked,
+                                      [{process, THIS},
+                                       {from, FROM},
+                                       {timestamp, os:timestamp()}])).
+-else.
+-define(TRACE_BLOCKED(THIS, FROM), ok).
+-define(TRACE_UNBLOCKED(THIS, FROM), ok).
```
videlalvaro: can you use SELF instead of THIS. THIS sounds too OOP'ish

michaelklishin:
@videlalvaro renamed.

Issue #143, opened 2015-05-05
Make credit flow values (initial value, more-credit-after) configurable


michaelklishin:
Some users would benefit from lower values, some *may* benefit from higher ones. We need an easier way to experiment with different settings ourselves.

michaelklishin:
Note that we may be able to do it for 3.5.x if we cache the values in process dictionary. For 3.6.0, we'll use state records and do more refactoring as needed.

michaelklishin:
I'm labeling this as `3.6.0` because I'm not sure if I should cherry-pick to `stable`. If that's the plan, let me know or do it, then change the milestone here.

PR #144, opened 2015-05-05
Allows credit flow Credits to be set via configuration

videlalvaro:
Fixes rabbitmq/rabbitmq-server#143


videlalvaro:
Something we didn't think about is what would happen if a user sets negative values here…

michaelklishin:
@videlalvaro we can introduce config validation later. This is not an option we will heavily advertise, only to selected users who've expressed interest.

videlalvaro:
Setting a negative value for InitialCredit will just disable flow control https://github.com/rabbitmq/rabbitmq-server/blob/master/src/credit_flow.erl#L90

PR #150, opened 2015-05-07
Make credit flow state transition interval 1s instead of 5s

### LINKED ISSUE #138: Consider reducing flow status threshold
Currently connections and channels would report itself as "blocked by credit flow" if they were blocked in the last 5 seconds. This means that in some environments, it may seem to an operator like they are in flow control permanently, even though they go in and out of the blocked state many times a second under high load.

We should consider adjusting the threshold to, say, 1 second.

michaelklishin:
Management UI and HTTP API currently report connections and channels
as in flow if they've been in flow for the last 5 seconds. That
can confuse the user with inter-node flow control, making them
believe the flow is permanent (it is not: the actual state
toggles many times a second).

This reduces the interval to 1s, which seems more reasonable and
accurate (in a way).

Fixes #138.

### DIFF
--- a/src/credit_flow.erl
+++ b/src/credit_flow.erl
@@ -68,6 +68,11 @@
             put(Key, Expr)
         end).
 
+%% If current process was blocked by credit flow in the last
+%% STATE_CHANGE_INTERVAL milliseconds, state/0 will report it as "in
+%% flow".
+-define(STATE_CHANGE_INTERVAL, 1000000).
+
 %%----------------------------------------------------------------------------
 
 %% There are two "flows" here; of messages and of credit, going in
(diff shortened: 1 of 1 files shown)


src/credit_flow.erl
```diff
@@ -68,6 +68,11 @@
             put(Key, Expr)
         end).
 
+%% if current process was blocked by credit flow
+%% in the last STATE_CHANGE_INTERVAL millisecond,
+%% state/0 will report it as "in flow".
+-define(STATE_CHANGE_INTERVAL, 100000).
```
videlalvaro: I think you are missing a zero.

PR #153, opened 2015-05-09
Support x-death event values from before #78

### LINKED ISSUE #152: Messages with x-deaths from pre-3.5.1 versions cause queue crash on dead-lettering
#78 needs to handle existing messages with `x-deaths`. See [this mailing list thread](https://groups.google.com/d/msg/rabbitmq-users/uQbeiWq5R3o/8QzpAhgbX6kJ).

michaelklishin:
We group x-death header values before processing them
to make sure there's only one per {queue, reason}.

Fixes #152.

### DIFF
--- a/src/rabbit_dead_letter.erl
+++ b/src/rabbit_dead_letter.erl
@@ -84,32 +84,31 @@ x_death_event_key(Info, Key, KeyType) ->
         {value, {Key, KeyType, Val}} -> Val
     end.
 
-maybe_append_to_event_group(Info, _Queue, _Reason, []) ->
-    [Info];
-maybe_append_to_event_group(Info, Queue, Reason, Acc) ->
+maybe_append_to_event_group(Table, _Queue, _Reason, []) ->
+    [Table];
+maybe_append_to_event_group(Table, Queue, Reason, Acc) ->
     case lists:any(queue_and_reason_matcher(Queue, Reason), Acc) of
         true  -> Acc;
-        false -> [Info | Acc]
+        false -> [Table | Acc]
     end.
 
+group_by_queue_and_reason([]) ->
+    [];
+group_by_queue_and_reason([Table]) ->
+    [Table];
 group_by_queue_and_reason(Tables) ->
-    Infos = [Info || {table, Info} <- Tables],
-    Grouped =
-        lists:foldl(fun (Info, Acc) ->
-            Q = x_death_event_key(Info, <<"queue">>, longstr),
-            R = x_death_event_key(Info, <<"reason">>, longstr),
-            Matcher = queue_and_reason_matcher(Q, R),
-            {Matches, _} = lists:partition(Matcher, Infos),
-            case Matches of
-                [X] ->
-                    maybe_append_to_event_group(
-                        ensure_xdeath_event_count(X), Q, R, Acc);
-                [X|_] = Xs when is_list(Xs) ->
-                    maybe_append_to_event_group(
-                        ensure_xdeath_event_count(X, length(Xs)), Q, R, Acc)
-            end
-        end, [], Infos),
-    [{table, Info} || Info <- Grouped].
+    lists:foldl(fun ({table, Info}, Acc) ->
+        Q = x_death_event_key(Info, <<"queue">>, longstr),
+        R = x_death_event_key(Info, <<"reason">>, longstr),
+        Matcher = queue_and_reason_matcher(Q, R),
+        {Matches, _} = lists:partition(Matcher, Tables),
+        {Augmented, N} = case Matches of
+                             [X]        -> {X, 1};
+                             [X|_] = Xs -> {X, length(Xs)}
+                         end,
+        maybe_append_to_event_group(
+            ensure_xdeath_event_count(Augmented, N), Q, R, Acc)
+    end, [], Tables).
 
 update_x_death_header(Info, Headers) ->
     Q = x_death_event_key(Info, <<"queue">>, longstr),
@@ -134,8 +133,8 @@ update_x_death_header(Info, Headers) ->
                 [{table, rabbit_misc:sort_field_table(Info1)} | Others])
     end.
 
-ensure_xdeath_event_count(Info) ->
-    ensure_xdeath_event_count(Info, 1).
+ensure_xdeath_event_count({table, Info}, InitialVal) when InitialVal >= 1 ->
+    {table, ensure_xdeath_event_count(Info, InitialVal)};
 ensure_xdeath_event_count(Info, InitialVal) when InitialVal >= 1 ->
     case x_death_event_key(Info, <<"count">>, long) of
         undefined ->
--- a/src/rabbit_dead_letter.erl
+++ b/src/rabbit_dead_letter.erl
@@ -99,19 +99,19 @@ group_by_queue_and_reason([Table]) ->
 group_by_queue_and_reason(Tables) ->
     {_, Grouped} =
         lists:foldl(fun ({table, Info}, {SeenKeys, Acc}) ->
-            Q = x_death_event_key(Info, <<"queue">>, longstr),
-            R = x_death_event_key(Info, <<"reason">>, longstr),
-            Matcher = queue_and_reason_matcher(Q, R),
-            {Matches, _} = lists:partition(Matcher, Tables),
-            {Augmented, N} = case Matches of
-                                 [X]        -> {X, 1};
-                                 [X|_] = Xs -> {X, length(Xs)}
-                             end,
-            Key = {Q, R},
-            Acc1 = maybe_append_to_event_group(
-                ensure_xdeath_event_count(Augmented, N), Key, SeenKeys, Acc),
-            {sets:add_element(Key, SeenKeys), Acc1}
-        end, {sets:new(), []}, Tables),
+                            Q = x_death_event_key(Info, <<"queue">>, longstr),
+                            R = x_death_event_key(Info, <<"reason">>, longstr),
+                            Matcher = queue_and_reason_matcher(Q, R),
+                            {Matches, _} = lists:partition(Matcher, Tables),
+                            {Augmented, N} = case Matches of
+                                                 [X]        -> {X, 1};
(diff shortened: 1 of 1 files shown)


michaelklishin:
Test case: https://github.com/rabbitmq/rabbitmq-java-client/pull/56.

src/rabbit_dead_letter.erl
```diff
@@ -84,36 +84,85 @@ x_death_event_key(Info, Key, KeyType) ->
...
+        lists:foldl(fun (Info, Acc) ->
+            Q = x_death_event_key(Info, <<"queue">>, longstr),
+            R = x_death_event_key(Info, <<"reason">>, longstr),
+            Matcher = queue_and_reason_matcher(Q, R),
+            {Matches, _} = lists:partition(Matcher, Infos),
+            case Matches of
+                [X] ->
+                    maybe_append_to_event_group(
+                        ensure_xdeath_event_count(X), Q, R, Acc);
+                [X|_] = Xs when is_list(Xs) ->
```
videlalvaro: do you really need to test for is_list here when you are matching already for a list?

src/rabbit_dead_letter.erl
```diff
@@ -84,36 +84,85 @@ x_death_event_key(Info, Key, KeyType) ->
...
+                            increment_xdeath_event_count(M)
                     end,
             rabbit_misc:set_table_value(Headers, <<"x-death">>, array,
-              [{table, rabbit_misc:sort_field_table(Info1)} | Others])
+                [{table, rabbit_misc:sort_field_table(Info1)} | Others])
+    end.
+
+ensure_xdeath_event_count(Info) ->
+    ensure_xdeath_event_count(Info, 1).
+ensure_xdeath_event_count(Info, InitialVal) when InitialVal >= 1 ->
```
videlalvaro: Do you need the guard here? In this line https://github.com/rabbitmq/rabbitmq-server/pull/153/files#diff-e2043d9f095f7dc4abc052dd9fae7599R109 the list passed there must have at least one element for the case to match.

src/rabbit_dead_letter.erl
```diff
@@ -84,36 +84,85 @@ x_death_event_key(Info, Key, KeyType) ->
...
+                <<"count">>, 1, Info,
+                {<<"count">>, long, N + 1})
     end.
 
+queue_and_reason_matcher(Q, R) ->
+  fun({table, Info}) ->
+        x_death_event_key(Info, <<"queue">>, longstr) =:= Q
+          andalso x_death_event_key(Info, <<"reason">>, longstr) =:= R;
+     (Info) when is_list(Info) ->
+        x_death_event_key(Info, <<"queue">>, longstr) =:= Q
```
videlalvaro: Perhaps this could be refactored to not need to duplicate the fun body in both clauses?
Also I hate to make these comments but the indentation here doesn't match what Emacs would give, which I think is what we use all over the place.

src/rabbit_dead_letter.erl
```diff
@@ -84,36 +84,85 @@ x_death_event_key(Info, Key, KeyType) ->
...
+maybe_append_to_event_group(Info, Queue, Reason, Acc) ->
+    case lists:any(queue_and_reason_matcher(Queue, Reason), Acc) of
+        true  -> Acc;
+        false -> [Info | Acc]
+    end.
+
+group_by_queue_and_reason(Tables) ->
+    Infos = [Info || {table, Info} <- Tables],
+    Grouped =
+        lists:foldl(fun (Info, Acc) ->
```
videlalvaro: As the other comment down below, here you can push the `fun` to the next line and then just use Emacs indent like we do here for example https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_msg_store.erl#L2019

videlalvaro: here `fun (Info, Acc) ->` could be `fun ({table, Info}, Acc) ->` and I think you don't need the list comprehension above

src/rabbit_dead_letter.erl
```diff
@@ -84,36 +84,85 @@ x_death_event_key(Info, Key, KeyType) ->
...
+            case Matches of
+                [X] ->
+                    maybe_append_to_event_group(
+                        ensure_xdeath_event_count(X), Q, R, Acc);
+                [X|_] = Xs when is_list(Xs) ->
+                    maybe_append_to_event_group(
+                        ensure_xdeath_event_count(X, length(Xs)), Q, R, Acc)
+            end
+        end, [], Infos),
+    [{table, Info} || Info <- Grouped].
```
videlalvaro: You could make `maybe_append_to_event_group` return `false -> [{table, Info} | Acc]` to remove this comprehension as well.

src/rabbit_dead_letter.erl
```diff
@@ -84,36 +84,85 @@ x_death_event_key(Info, Key, KeyType) ->
...
+
+group_by_queue_and_reason(Tables) ->
+    Infos = [Info || {table, Info} <- Tables],
+    Grouped =
+        lists:foldl(fun (Info, Acc) ->
+            Q = x_death_event_key(Info, <<"queue">>, longstr),
+            R = x_death_event_key(Info, <<"reason">>, longstr),
+            Matcher = queue_and_reason_matcher(Q, R),
+            {Matches, _} = lists:partition(Matcher, Infos),
+            case Matches of
```
videlalvaro: `lists:partition/2` could return an empty list which could cause a `exception error: no case clause matching []` in the `case` there

src/rabbit_dead_letter.erl
```diff
@@ -84,36 +84,85 @@ x_death_event_key(Info, Key, KeyType) ->
...
+    end.
+
+group_by_queue_and_reason(Tables) ->
+    Infos = [Info || {table, Info} <- Tables],
+    Grouped =
+        lists:foldl(fun (Info, Acc) ->
+            Q = x_death_event_key(Info, <<"queue">>, longstr),
+            R = x_death_event_key(Info, <<"reason">>, longstr),
+            Matcher = queue_and_reason_matcher(Q, R),
+            {Matches, _} = lists:partition(Matcher, Infos),
```
videlalvaro: I think you could pass a sorted list of `Infos` to the foldl, so every time you call `lists:partition/2` you can discard a whole bunch of infos, if I understand this correctly, which would remove the need for `maybe_append_to_event_group` which goes over the `Acc` all the time.

michaelklishin: I'm personally not in afvour this implicit behaviour (pre-sorted input required).

src/rabbit_dead_letter.erl
michaelklishin: We take queue name and reason from the current item, so at least one item will match.

src/rabbit_dead_letter.erl
michaelklishin: Currently it takes a proplist and returns a proplist. We could make it accept a pair and return a pair.

src/rabbit_dead_letter.erl
michaelklishin: Unfortunately, even with pre-sorted input we may have duplicate entries (on the two fields we group by).

videlalvaro: You could sort the list inside `group_by_queue_and_reason` before calling the fold. I think currently by calling `lists:any` all the time the complexity of this function could go quite up since you are always traversing the Accumulator for every iteration of the foldl. Sorting the input before calling foldl removes that

michaelklishin:
@videlalvaro ready for another round. I'll re-format with Emacs once we agree on everything else.

videlalvaro:
About the sort, I think lists:partition should take care of partitioning the list by duplicates?

videlalvaro:
Also see the comment related to duplicated fun code in `queue_and_reason_matcher`

michaelklishin:
Sorting is not sufficient: we may have duplicates in the input that `foldl` would still iterate over, so they need to be checked. We can avoid an O(n) check by keeping a set of {queue, reason} pairs in the accumulator.

src/rabbit_dead_letter.erl
```diff
@@ -84,36 +84,90 @@ x_death_event_key(Info, Key, KeyType) ->
...
-                                      {<<"count">>, long, N + 1})
-                            end
+                            increment_xdeath_event_count(M)
                     end,
             rabbit_misc:set_table_value(Headers, <<"x-death">>, array,
-              [{table, rabbit_misc:sort_field_table(Info1)} | Others])
+                [{table, rabbit_misc:sort_field_table(Info1)} | Others])
+    end.
+
+ensure_xdeath_event_count({table, Info}, InitialVal) when InitialVal >= 1 ->
```
videlalvaro: are there cases where `InitialVal` might be less than one? If yes, the caller will get an undef error

michaelklishin: No. If we have an event, the number of times it happened is >= 1.

src/rabbit_dead_letter.erl
```diff
@@ -84,36 +84,90 @@ x_death_event_key(Info, Key, KeyType) ->
...
+            {Matches, _} = lists:partition(Matcher, Tables),
+            {Augmented, N} = case Matches of
+                                 [X]        -> {X, 1};
+                                 [X|_] = Xs -> {X, length(Xs)}
+                             end,
+            Key = {Q, R},
+            Acc1 = maybe_append_to_event_group(
+                ensure_xdeath_event_count(Augmented, N), Key, SeenKeys, Acc),
+            {sets:add_element(Key, SeenKeys), Acc1}
+        end, {sets:new(), []}, Tables),
```
videlalvaro: perhaps since you use a set, why not use the set as the actual accumulator, and then do http://erlang.org/doc/man/sets.html#to_list-1

This is more of a _taste_ comment than a requirement.

videlalvaro:
Added some comments, after those, I think just the indentation ones remain

src/rabbit_dead_letter.erl
videlalvaro: Then what's the point of having those InitialVal guards?

Issue #154, opened 2015-05-13
Support for 'rabbitmq-env.conf' functionality on Windows


Jakauppila:
Users on the Windows side running RabbitMQ are missing out on the functionality of 'rabbitmq-env.conf', this can be replicated by executing another bat where environment variables are defined.

PR #155, opened 2015-05-13
Support for 'rabbitmq-env.conf' functionality on Windows

Jakauppila:
By adding the execution of rabbitmq-env.bat within the existing .bat scripts, we provide an area to define environment variables prior to the execution of the server/ctl.

Fixes [rabbitmq/rabbitmq-server#154](https://github.com/rabbitmq/rabbitmq-server/issues/154)

### DIFF
--- a/scripts/rabbitmq-defaults.bat
+++ b/scripts/rabbitmq-defaults.bat
@@ -1,0 +1,35 @@
+@echo off
+
+REM ### next line potentially updated in package install steps
+REM set SYS_PREFIX=
+
+REM ### next line will be updated when generating a standalone release
+REM ERL_DIR=
+set ERL_DIR=
+
+REM These boot files don't appear to be referenced in the batch scripts
+REM set CLEAN_BOOT_FILE=start_clean
+REM set SASL_BOOT_FILE=start_sasl
+
+REM ## Set default values
+
+if "!RABBITMQ_BASE!"=="" (
+    set RABBITMQ_BASE=!APPDATA!\RabbitMQ
+)
+
+REM CONFIG_FILE=${SYS_PREFIX}/etc/rabbitmq/rabbitmq
+REM LOG_BASE=${SYS_PREFIX}/var/log/rabbitmq
+REM MNESIA_BASE=${SYS_PREFIX}/var/lib/rabbitmq/mnesia
+REM ENABLED_PLUGINS_FILE=${SYS_PREFIX}/etc/rabbitmq/enabled_plugins
+set CONFIG_FILE=!RABBITMQ_BASE!\rabbitmq
+set LOG_BASE=!RABBITMQ_BASE!\log
+set MNESIA_BASE=!RABBITMQ_BASE!\db
+set ENABLED_PLUGINS_FILE=!RABBITMQ_BASE!\enabled_plugins
+
+REM PLUGINS_DIR="${RABBITMQ_HOME}/plugins"
+set PLUGINS_DIR=!TDP0!..\plugins
+
+REM CONF_ENV_FILE=${SYS_PREFIX}/etc/rabbitmq/rabbitmq-env.conf
+if "!RABBITMQ_CONF_ENV_FILE!"=="" (
+    set CONF_ENV_FILE=!APPDATA!\RabbitMQ\rabbitmq-env-conf.bat
+)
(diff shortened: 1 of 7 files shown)


michaelklishin:
Thank you, we will take a look.

Jakauppila:
Thanks, let me know if there's any questions/concerns.

One thing I wasn't 100% sure on was simply the location of the file. It may be worth defining a new environment variable for it's location, if it doesn't exist there, check %RABBITMQ_BASE%, if it doesn't exist there, check the default location !APPDATA!\RabbitMQ.

michaelklishin:
That's the best default value I can think of for Windows. And we don't hear requests to make `rabbitmq-env.conf` overridable on *nix, so it should be alright.

michaelklishin:
This looks good but there is one difference from `rabbitmq-env.conf` on *nix: you can use `NODENAME` instead of `RABBITMQ_NODENAME` in the env file and it will be picked up. I'd like to keep this consistent on Windows.

I'll re-assign to @dumbbell, who has a deeper knowledge of the scripts and how they relate to packaging than I do.

Jakauppila:
Sure, that makes sense. There will just have to be additional logic on every environment variable check:
```batch
if "!RABBITMQ_NODENAME!"=="" (
	if "!NODENAME!"=="" (
		set RABBITMQ_NODENAME=rabbit@!COMPUTERNAME!
	) else (
		set RABBITMQ_NODENAME=!NODENAME!
	)
)
```
If that works, I can continue making the additional changes.

Just another note on the default location, I know on *nix it defaults to `/etc/rabbitmq/rabbitmq-env.conf` so it would make sense to default to `!APPDATA!\RabbitMQ\rabbitmq-env.bat` on Windows.

Since I would like the ability to override it could be done indirectly via `!RABBITMQ_BASE!\rabbitmq-env.bat` like I have above, but that seems a little silly since that environment variable isn't configured with an install. Is this fine since it's more of an advanced feature/config or should it check that location and then the default if it doesn't exist?

dumbbell:
Hi @Jakauppila! Sorry for the delay...

I see the Windows scripts define all the variables themselves (thus, there is duplicated code).

On Unix, we have the following dependency graph:
```
rabbitmq-server, rabbitmqctl, ...
|
`-- rabbitmq-env
    |
    |-- rabbitmq-defaults
    |
    `-- /etc/rabbitmq/rabbitmq-env.conf
```

Here is how it works:

1. `rabbitmqctl` & friends load `rabbitmq-env` which is always installed in the same directory.
2.  `rabbitmq-env` loads `rabbitmq-defaults` who defines a few variables, in particular:

  ```sh
CONF_ENV_FILE=${SYS_PREFIX}/etc/rabbitmq/rabbitmq-env.conf
```

3. `rabbitmq-env` loads `$CONF_ENV_FILE` if it's there.
4. `rabbitmq-env` ensures all variables are initialized.

So, all variables are overridable in `rabbitmq-env.conf`, except its location. This is something we may change in the future (for instance, the package for FreeBSD patches `rabbitmq-defaults` to put `rabbitmq-env.conf` in `/usr/local/etc/rabbitmq`).

On Windows, it would be nice to have the same logic:
* All scripts load `rabbitmq-env.bat` which loads `rabbitmq-defaults.bat`.
* `rabbitmq-defaults.bat` defines `RABBITMQ_BASE` and `RABBITMQ_ENV_CONF_FILE` (the location of `rabbitmq-env-conf.bat`, defaults to `!APPDATA!\RabbitMQ\rabbitmq-env-conf.bat`); possibly allowing these variables to be user-defined.
* `rabbitmq-env.bat` loads `RABBITMQ_ENV_CONF_FILE` and initializes all variables.

What do you think? Does it match common/best practices on Windows or is it too Unix-centric?

Jakauppila:
Hi @dumbbell, sorry for the late response, very extended weekend.

That looks great to me; it'll be nice to have the same logic/functionality on both sides.

As far as putting these changes together, want me to take a stab at it or do you?

michaelklishin:
@Jakauppila please go ahead. The right thing to do is to let the contributor make changes based on feedback :)

Jakauppila:
@michaelklishin Will do, thanks.

Jakauppila:
Sorry for the delay. Finally got some time to work on it and committed the changes.

For the *nix scripts it's mimicking for `rabbitmq-defaults.bat` and `rabbitmq-env.bat`, I commented out those lines and put the batch equivalents below it for easier review.

michaelklishin:
I will take this because there is something closely related that I will test for 3.5.4.

michaelklishin:
Looking into this today as it is a good idea to merge this before #208.

michaelklishin:
Cherry-picked to `stable`. Thank you.

Issue #134, opened 2015-04-29
The file handle cache ignores the memory high watermark


michaelklishin:
From an escalation (`VESC-420`). Manual GC triggering does not change anything. So the issue most likely is an unbounded buffer, e.g. message store overwhelming queue master with messages it cannot relay to the mirror fast enough.

michaelklishin:
To reproduce:

 * Create a 2 node cluster
 * Fill up a durable queue with persistent messages (say, 10+ GB)
 * Stop and reset the mirror
 * Re-add it back
 * Monitor RAM use on the master node

dumbbell:
It appears the memory is consumed by the file handle cache: the master queue's process dictionary is full of `{Ref, fhc_handle} => {handle, ...}` key/value pairs which include the read buffer.

During my test, the queue contains 1000 messages of 10 MiB each. The memory high watermark being set at around 2 GiB, nearly all (if not all) messages are paged to the disk. However, the queue process is happily consuming 5 GiB of memory. This matches what `erlang:process_info(Queue_Process, binary)` returns (in my case, 500 referenced binaries of 10 MiB each).

Neither the file handle cache nor the message store ask the memory monitor how much memory they are allowed to consume. They only pay attention to the number of open file handles.

michaelklishin:
@dumbbell worth delaying 3.5.2 for if you think we can fix this in under a week.

dumbbell:
I agree, let's delay 3.5.2. If I don't have a good solution in a week, we can do the release.

dumbbell:
This is a preliminary commit. There is a debug log message left on purpose for instance; I will remove it after further testing.

dumbbell:
Here are the results of today's testing and benchmark:

I measured the time taken to sync one mirrored queue containing 1000 messages of 10 MiB each (10 GiB worth of data). The two nodes are running on the same host, the disk is an SSD formatted using ZFS (which does heavy caching). Both had a high watermark set to 2.2 GiB.

I did four runs for each of the following versions:
* `stable` which exhibits the bug described here;
* `rabbitmq-server-134` with the explicit garbage collection;
* `rabbitmq-server-134` without the explicit garbage collection.

Results:
```
$ ministat -s -w 88 stable \#134-gc \#134-nogc 
x stable
+ #134-gc
* #134-nogc
+----------------------------------------------------------------------------------------+
|       *                                                                                |
|x  +   *** x            +  x                    x         +                            +|
[...]

PR #160, opened 2015-05-15
Start rabbit_alarm before FHC

### LINKED ISSUE #157: Memory monitor isn't there when FHC assumes it to be
See [this thread](https://groups.google.com/forum/#!topic/rabbitmq-users/ydTzwg7Gjkk). My reading of it is that VM monitor isn't there when FHC assumes it is.

michaelklishin:
This is against `stable`, fixes #157.

Since the FHC now uses memory monitor, we need to make sure it is
running early on. Nothing in rabbit_alarm really depends on the
steps that used to run earlier.

`full` test suite passes.

It may be worth extracting memory and disk monitors into separate
boot steps down the road.

### DIFF
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -40,6 +40,13 @@
                     {requires,    pre_boot},
                     {enables,     external_infrastructure}]}).
 
+%% rabbit_alarm currently starts memory and disk space monitors
+-rabbit_boot_step({rabbit_alarm,
+                   [{description, "alarm handler"},
+                    {mfa,         {rabbit_alarm, start, []}},
+                    {requires,    pre_boot},
+                    {enables,     external_infrastructure}]}).
+
 -rabbit_boot_step({database,
                    [{mfa,         {rabbit_mnesia, init, []}},
                     {requires,    file_handle_cache},
@@ -54,7 +61,8 @@
 -rabbit_boot_step({file_handle_cache,
                    [{description, "file handle cache server"},
                     {mfa,         {rabbit, start_fhc, []}},
-                    {requires,    pre_boot},
+                    %% FHC needs memory monitor to be running
+                    {requires,    rabbit_alarm},
                     {enables,     worker_pool}]}).
 
 -rabbit_boot_step({worker_pool,
@@ -84,12 +92,6 @@
 -rabbit_boot_step({kernel_ready,
                    [{description, "kernel ready"},
                     {requires,    external_infrastructure}]}).
-
--rabbit_boot_step({rabbit_alarm,
-                   [{description, "alarm handler"},
-                    {mfa,         {rabbit_alarm, start, []}},
-                    {requires,    kernel_ready},
-                    {enables,     core_initialized}]}).
 
 -rabbit_boot_step({rabbit_memory_monitor,
                    [{description, "memory monitor"},


videlalvaro:
How does this interact with this code? https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit.erl#L885

videlalvaro:
AFAIK the file handle cache shouldn't refer to any `rabbit_*` code as it does here: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/file_handle_cache.erl#L967

The comment here suggest that: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit.erl#L883

ping @dumbbell

src/rabbit.erl
```diff
@@ -40,6 +40,13 @@
                     {requires,    pre_boot},
                     {enables,     external_infrastructure}]}).
 
+%% rabbit_alarm currently starts memory and disk space monitors
+-rabbit_boot_step({rabbit_alarm,
+                   [{description, "alarm handler"},
+                    {mfa,         {rabbit_alarm, start, []}},
+                    {requires,    pre_boot},
```
videlalvaro: how does moving this earlier in the boot process interacts with the use of `rabbit_log` here: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_alarm.erl#L212 ?

michaelklishin:
`rabbit: start_fhc/0` is used by a different boot step that now depends on the one for `rabbit_alarm`.

michaelklishin:
`rabbit_log` depends on `error_logger`, there is no boot step for it.

michaelklishin:
@videlalvaro that "shouldn't reference any rabbit stuff" comment presumably means that FHC could be a standalone library: nothing in it in theory is RabbitMQ-specific. Unfortunately, this means it completely ignored VM memory watermark, which leads to OOM killer killing the VM when a node with a lot of data on disk boots. So this conceptual purity has a very real drawback.

I will remove the commend.

michaelklishin:
The comment on `rabbit:start_fhc/0` comes from 07ea1d8f2b9e88f965a41b94f8cee2ead78162c0, which was part of `bug24998`. I see no specific reasons for FHC to not be reference "any rabbit stuff" mentioned there, the discussion is largely about the separation of concerns, how to not log the same thing twice, and how alarm handlers should interact with the logger.

videlalvaro:
# About rabbit_alarm and error_logger

`rabbit_log` has no boot step, yes, but `rabbit_alarm` calls into rabbit_log: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_alarm.erl#L212 which in turns call into `error_logger` https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_log.erl#L58

The problem is `rabbit_error_logger` sets itself as a report handler for error_logger: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_error_logger.erl#L45 but `rabbit_error_logger` depends on the exchange infrastructure from being ready. When the exchanges are ready, it means `kernel_ready` is there, that might be the reason why `rabbit_alarm` was depending on `kernel_ready`.

# About "code purity" 

AFAIK, all the files in our repo that don't start with `rabbit_*` should be independent from the rabbit source code, like libraries. Besides the "purity" details, I don't see why the `rabbit_memory_monitor` function required by `file_handle_cache` cannot be passed as a _fun_ as these other things are already done in the `start_fhc` function.

michaelklishin:
> rabbit_alarm calls into rabbit_log

Only when an actual alarm is triggered, which does not normally happen on boot.

> when the exchanges are ready, it means kernel_ready is there, that might be the reason why
> rabbit_alarm was depending on kernel_ready.

We can add separate boot steps for memory and disk monitors.

> AFAIK, all the files in our repo that don't start with rabbit_* should be independent from the rabbit source code

This should have been brought up in #134 and has little to do with this PR

> I don't see why the rabbit_memory_monitor function required by file_handle_cache cannot be passed as a fun

It can be. This PR is about something else.

dumbbell:
> (...) `rabbit_error_logger` depends on the exchange infrastructure from being ready. When the exchanges are ready, it means `kernel_ready` is there, that might be the reason why `rabbit_alarm` was depending on `kernel_ready`.

By "exchange infrastructure", don't you mean "`external_infrastructure`" instead?

Before the boot steps are executed, we have:
```erlang
ok = ensure_working_log_handlers(),
```
So after this point, I understand `rabbit_log` is working (`rabbit_log`'s handler is registered in `error_logger`). @michaelklishin's patch seems correct to me.

Now, regarding source files without the`rabbit_` prefix, I didn't know about the fact they must have no knowledge of RabbitMQ. I will prepare a separate fix once this bug is solved.

videlalvaro:
I meant _exchange infrastructure_, since this: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_error_logger.erl#L108 needs a topic exchange to work. Be aware that 

According to this: https://raw.githubusercontent.com/videlalvaro/rabbit-internals/master/images/boot_steps.png the exchanges are ready once `kernel_ready` has run.

About the file with the `rabbit_` prefix, I think I've read a comment from @simonmacmullen in our Slack channel about that, besides from the comment pointed above about FHC initialisation in `rabbit.erl`.

dumbbell:
Thanks @videlalvaro, I see the problem now.

Here is a new proposal, separated in two commits:
1. In `rabbit_memory_monitor:memory_use/1`, support the fact that `vm_memory_monitor` may not be started yet by catching the `noproc` exception. This avoids any changes to the boot steps ordering.
2. Remove the knowledge of RabbitMQ from `file_handle_cache` by passing an anonymous function to call `rabbit_memory_monitor:memory_use/1`.

If this seems reasonable to you, I will squash everything into a single commit, but I wanted to keep @michaelklishin proposal for the discussion.

videlalvaro:
@dumbbell the purpose of having the FHC into the memory monitor is to not OOM rabbit at startup right?

I think one thing is detecting the problem, another one is to tell the user there's a problem.

As a user I don't want my rabbit crashing (FHC part of the memory monitor feedback system). If it crashes, I want to know why (FHC/memory monitor logging that things went wrong)

videlalvaro:
About the 2. I think that could be a separate PR, unless to fix this issue we have to change that part of the code, then it could be part of this bug

dumbbell:
Only the first few reads before `vm_memory_monitor` is started won't have the check. Once it's started, the memory limit is applied, possibly clearing the cache.

PR #164, opened 2015-05-20
Adds TLS suggestion

### LINKED ISSUE #162: rabbitmqctl should suggest the user might be using TLS for cluster communication
To setup TLS distribution among rabbit nodes one needs to set `RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS` and `RABBITMQ_CTL_ERL_ARGS`. If we forget about the later, `rabbitmqctl` will fail to contact the node but will give wrong feedback:

>rabbit@mynode:
>  * connected to epmd (port 4369) on avidela
>  * epmd reports node 'rabbit' running on port 54269
>  * TCP connection succeeded but Erlang distribution failed
>  * suggestion: hostname mismatch?
>  * suggestion: is the cookie set correctly?

We should add a suggestion about TLS dist

videlalvaro:
Fixes #162

### DIFF
--- a/src/rabbit_nodes.erl
+++ b/src/rabbit_nodes.erl
@@ -137,7 +137,8 @@ dist_broken_diagnostics(Name, Host, NamePorts) ->
                      [{"  * TCP connection succeeded but Erlang distribution "
                        "failed~n"
                        "  * suggestion: hostname mismatch?~n"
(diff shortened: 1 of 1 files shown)


michaelklishin:
This PR is against `master`. Is it meant to be against `stable`?

videlalvaro:
damn, did I fail again? I will resubmit :(

PR #172, opened 2015-05-27
Moves boot steps handling to it's own module

### LINKED ISSUE #163: Add visibility into rabbit boot steps
At the moment it's quite hard to visualise the order in which boot steps are run, both for debugging purposes and for us developers to know what to add where inside RabbitMQ.

It would be nice to be able to export the boot steps graph to latter be used by `dot` or by some management plugin extension showing a graph using d3.js or similar.

videlalvaro:
Fixes #163

### DIFF
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -390,7 +390,7 @@ start_apps(Apps) ->
     app_utils:load_applications(Apps),
     OrderedApps = app_utils:app_dependency_order(Apps, false),
     case lists:member(rabbit, Apps) of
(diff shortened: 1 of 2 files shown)


videlalvaro:
I've made the PR against master since it adds a new API, which should only be used from Erlang, but still not sure this should go on a PATCH release, feel free to rebase and make it part of a PATCH release.

Apart from that, there's no BC breaking changes with the code we had before.

michaelklishin:
Looks good. I'll rebase to `stable` to reduce the distance between `master` and `stable`.

PR #173, opened 2015-05-28
Moves boot steps to its own module

### LINKED ISSUE #163: Add visibility into rabbit boot steps

michaelklishin:
Fixes #163, rebases #172 onto `stable`.


michaelklishin:
OK, giving it a second thought, this may be a pretty unexpected change for plugins. Will rebase back to master.

Issue #62, opened 2015-03-09
Make rabbitmqctl operations support a timeout


michaelklishin:
The smallest in scope solution was submitted as #61 to be considered for `3.5.0`. The team has decided that it is not really an improvement over [timeout(1)](http://linux.die.net/man/1/timeout) (except for working on Windows, perhaps) and ideally the `list_*` commands should list unresponsive/unreachable
items (e.g. channels or queues).

[Mailing list thread](https://groups.google.com/forum/#!searchin/rabbitmq-users/hangs/rabbitmq-users/Ja0iyfF0Szw/MHm8RuqCaacJ).

michaelklishin:
Given ongoing community requests for this feature, we will prioritise this. #61 may be resurrected and included into `3.5.x`.

michaelklishin:
First improvements landed in #181, will be in `3.5.x`.

dumbbell:
We discussed this internaly some time ago, here are more details.

To go beyond what was committed in #181, we want to handle timeous in a finer grain way. Let's take `list_queues` as an example: if there are 10 queues, but one of them seems stuck, the current implementation will timeout globally and may display no information about the responsive queues.

An improved implementation would display information about responsive queues, and indicate a timeout for each stuck element.

videlalvaro:
Perhaps the "one second timeout" could be specified either per element or globally, the later being the default one

michaelklishin:
Yes, this is a good question whether we should have per-element timeouts. I don't think this is what our users would expect. "Global" timeouts are much more common in similar tools.

michaelklishin:
Also note that as far as ops engineers are concerned, it's the total timeout that they care about most.

dumbbell:
I'm not sure a per-element timeout is interesting. Furthermore it would add another option and complicate the UI.

videlalvaro:
I think it's fine as long as we output things as they happen. ie: we print as much stuff as we can during the specified timeout; if the timeout happens, then fine. Stop the whole thing

dumbbell:
It also depends if the output must be sorted or not. I admit I didn't verify if it is the case already for eg. `list_queues`. If we don't need to sort the output, I agree we should display everything as soon as we get the information.

Ayanda-D:
The output is sorted. See `rabbit_control_main:display_info_list`. Question is, do we keep it as is, and display output only after we've received all info and sorted it, or, remove the sorting and display things as they happen?

[...]

PR #181, opened 2015-06-09
Introduce a timeout flag to rabbitmqctl (for some commands)

michaelklishin:
Refeferences #62, same as #61 but against stable.

The goal is to introduce a way to specify a timeout in the next bug fix release, then add riskier changes, such as providing more context about what's timed out.

The default is still `infinity`. We can add more commands to the list of commands with timeouts, including for `3.5.x`.

### DIFF
--- a/src/rabbit_control_main.erl
+++ b/src/rabbit_control_main.erl
@@ -201,7 +201,11 @@ parse_timeout(infinity) ->
 parse_timeout(N) when is_list(N) ->
     try parse_number(N) of
         M ->
-            {ok, round(M) * 1000}
+            Y = case M >= 0 of
+                    true  -> round(M) * 1000;
+                    false -> ?RPC_TIMEOUT
+                end,
+            {ok, Y}
     catch error:badarg ->
         {error, infinity}
     end;
--- a/src/rabbit_control_main.erl
+++ b/src/rabbit_control_main.erl
@@ -21,11 +21,11 @@
 -export([start/0, stop/0, parse_arguments/2, action/5,
          sync_queue/1, cancel_sync_queue/1, become/1]).
 
--import(rabbit_cli, [rpc_call/4]).
+-import(rabbit_cli, [rpc_call/4, rpc_call/5]).
 
 -define(EXTERNAL_CHECK_INTERVAL, 1000).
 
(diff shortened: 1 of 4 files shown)


src/rabbit_control_main.erl
```diff
@@ -160,18 +174,48 @@ print_report0(Node, {Module, InfoFun, KeysFun}, VHostArg) ->
...
-        true  -> action(Command, Node, Args, Opts, Inform)
+        false ->
+            case ensure_app_running(Node) of
+                ok ->
+                    case lists:member(Command, ?COMMANDS_WITH_TIMEOUT) of
+                        true  ->
+                            {T, Suffix} =
+                                case Timeout of
+                                    infinity   -> {"infinity", ""};
+                                    "infinity" -> {"infinity", ""};
```
videlalvaro: why is there the `"infinity"` case?

src/rabbit_control_main.erl
```diff
@@ -160,18 +174,48 @@ print_report0(Node, {Module, InfoFun, KeysFun}, VHostArg) ->
...
-                     ok -> action(Command, Node, Args, Opts, Inform);
-                     E  -> E
-                 end;
-        true  -> action(Command, Node, Args, Opts, Inform)
+        false ->
+            case ensure_app_running(Node) of
+                ok ->
+                    case lists:member(Command, ?COMMANDS_WITH_TIMEOUT) of
+                        true  ->
+                            {T, Suffix} =
```
videlalvaro: Can you move the lines 201-210 to an `inform_timeout` function?

src/rabbit_control_main.erl
```diff
@@ -160,18 +182,67 @@ print_report0(Node, {Module, InfoFun, KeysFun}, VHostArg) ->
...
+                end,
+            {ok, Y}
+    catch error:badarg ->
+        {error, infinity}
+    end;
+parse_timeout(N) ->
+    {ok, N}.
+
+format_timeout(N) when is_number(N) ->
+    float_to_list(N / 1000, [{decimals, 0}]).
```
michaelklishin: `erlang:float_to_list/2` is not available in older Erlang/OTP releases, will get rid of it and re-submit.

PR #184, opened 2015-06-10
Fix set type

weisslj:


michaelklishin:
Cherry-picked to stable.

Thank you, @weisslj.

PR #188, opened 2015-06-11
Update digraph type

weisslj:
The type digraph:digraph/0 got changed to digraph:graph/0 in OTP R16B03.

To support older and newer releases a preprocessor define (with
configure-like check) would be necessary.

### DIFF
--- a/src/rabbit_misc.erl
+++ b/src/rabbit_misc.erl
@@ -217,7 +217,7 @@
         (atom()) -> [{atom(), atom(), [term()]}]).
 -spec(build_acyclic_graph/3 ::
         (graph_vertex_fun(), graph_edge_fun(), [{atom(), [term()]}])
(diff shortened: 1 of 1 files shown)


michaelklishin:
We are not ready to require R16B03 just yet. R13B03 is still the minimum supported version.

weisslj:
OK! Updating the type does not remove support for R13B03. The way it is now all versions >= R16B03 get this warning when calling `make dialyze`:
```
Unknown types:
  digraph:digraph/0
```

After this patch versions < R16B03 would get a Dialyzer warning. To support all versions, a configure-like check for the stdlib version or diagraph type would be required to set a preprocessor define.

Feel free to close the pull request if you want to keep the old type instead.

michaelklishin:
OK, if it's just a warning then it shouldn't affect CI. Let's give it a shot.

weisslj:
Thanks!

PR #192, opened 2015-06-16
adds missing find_prioritisers call

### LINKED ISSUE #191: Throughput regression in 3.4.0
Between 63678f206f5648c64459574ab4b748c5852cf094 and b8cb778430df6279716f9ce4d8b849a518c74474 there's a double digit throughput drop when PerfTest is executed with

```
./runperftest.sh -p -x 1 -y 1 -s 1000 -z 60 -q 1000
```

videlalvaro:
Adds missing call to `find_prioritisers/1`. We need this call since the return from `init/1` on our `gen_server2` behaviour allows the app to swap the `gen_server2` behaviour module. If the module is changed we have to search for message prioritisers. 

Without this change, code like https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_amqqueue_process.erl#L906 won't be called/used.

This bug was affecting both `rabbit_amqqueue_process` and `rabbit_mirror_queue_slave`. So mirroring performance should get enhanced as well.

Fixes #191

### DIFF
--- a/src/gen_server2.erl
+++ b/src/gen_server2.erl
@@ -576,10 +576,11 @@ init_it(Starter, Parent, Name0, Mod, Args, Options) ->
         {ok, State, Timeout, Backoff = {backoff, _, _, _}, Mod1} ->
             Backoff1 = extend_backoff(Backoff),
             proc_lib:init_ack(Starter, {ok, self()}),
-            loop(GS2State #gs2_state { mod           = Mod1,
-                                       state         = State,
-                                       time          = Timeout,
-                                       timeout_state = Backoff1 });
+            loop(find_prioritisers(
+                  GS2State #gs2_state { mod           = Mod1,
+                                        state         = State,
+                                        time          = Timeout,
+                                        timeout_state = Backoff1 }));
         {stop, Reason} ->
             %% For consistency, we must make sure that the
             %% registered name (if any) is unregistered before


dumbbell:
I confirm the patch fixes the performance issue:
* @b8cb778430df6279716f9ce4d8b849a518c74474: 11k/s
* @63678f206f5648c64459574ab4b748c5852cf094: 14k/s
* HEAD of `stable`: 10k/s
* This patch: 13k/s

src/gen_server2.erl
```diff
@@ -576,10 +576,11 @@ init_it(Starter, Parent, Name0, Mod, Args, Options) ->
...
             proc_lib:init_ack(Starter, {ok, self()}),
-            loop(GS2State #gs2_state { mod           = Mod1,
-                                       state         = State,
-                                       time          = Timeout,
-                                       timeout_state = Backoff1 });
+            loop(find_prioritisers(
+                  GS2State #gs2_state { mod           = Mod1,
+                                        state         = State,
+                                        time          = Timeout,
+                                        timeout_state = Backoff1 }));
```
dumbbell: One minor comment: instead of calling `find_prioritisers()` twice, we could:

1. Store the result of the `case catch Mod:init ... end` in a variable
2. Call `find_prioritisers()` + `loop()` after the case.

videlalvaro: The problem is that the `Mod` we pass to the first call to `find_prioritisers/1` is not the same `Mod` passed in the second call. The bug was precisely there, not calling `find_prioritisers/1` for the new `Mod` that was swapped by the `gen_server2` callback.

`rabbit_prequeue` starts as the `gen_server2` callback being `Mod`, then `Mod:init/1` is called, which returns a new module, in our case it could be `rabbit_amqqueue_process`. The module `rabbit_amqqueue_process` will become the new `gen_server2`, so we have to call `find_prioritisers/1` again

rjrizzuto:
Is this change going into 3.5.4?

michaelklishin:
@rjrizzuto it will.

PR #193, opened 2015-06-18
Adds error message for list_user_permissions missing argument

### LINKED ISSUE #183: rabbitmqctl gives cryptic erlang error when list_user_permissions is missing arguments
Run `./scripts/rabbitmqctl list_user_permissions` without passing the `user` option

videlalvaro:
Fixes #183

### DIFF
--- a/src/rabbit_control_main.erl
+++ b/src/rabbit_control_main.erl
@@ -519,6 +519,9 @@ action(list_vhosts, Node, Args, _Opts, Inform, Timeout) ->
     display_info_list(call(Node, {rabbit_vhost, info_all, []}, Timeout),
                       ArgAtoms);
 
+action(list_user_permissions, _Node, _Args = [], _Opts, _Inform, _Timeout) ->
+    {error_string,
+     "list_user_permissions expects a username argument, but none provided."};
 action(list_user_permissions, Node, Args = [_Username], _Opts, Inform, Timeout) ->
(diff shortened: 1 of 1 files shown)


src/rabbit_control_main.erl
```diff
@@ -519,6 +519,9 @@ action(list_vhosts, Node, Args, _Opts, Inform, Timeout) ->
     display_info_list(call(Node, {rabbit_vhost, info_all, []}, Timeout),
                       ArgAtoms);
 
+action(list_user_permissions, _Node, _Args = [], _Opts, _Inform, _Timeout) ->
```
dumbbell: Perhaps `Args = [""]` could be caught here as well?

I'm thinking of scripts calling `rabbitmqctl list_user_permissions "$user"` but `$user` is not set.

videlalvaro: wouldn't the empty string be a problem with most commands actually? I think for that case we get an error like this:

```
Listing permissions for user [] ...
Error: no_such_user: 
```

which at least tells me something about the problem, not like `function_clause` before

michaelklishin: Yes, if we want to handle empty values, it needs to be a separate issue (not sure how many run into that).

dumbbell: Ok, the current error is fine. Disregard my comment.

Issue #196, opened 2015-06-23
Add a way to clear the read cache


dumbbell:
Such a command would clear the read buffers stored in the `file_handle_cache`'s gen_server and in all queues' process dictionary.

PR #199, opened 2015-06-23
file_handle_cache: Add a command to clear all read buffers

dumbbell:
This is meant to be called from `rabbitmqctl`:
```
rabbitmqctl eval 'file_handle_cache:clear_read_cache().'
```

The command sends a message to all queues on all vhosts to make them clear their own read cache. It is asynchronous, so the command will return before the action is complete.

Note: I kept log messages as it could be useful during QA. They should be removed before the merge.

References #196.

### DIFF
--- a/src/file_handle_cache.erl
+++ b/src/file_handle_cache.erl
@@ -148,7 +148,7 @@
          copy/3, set_maximum_since_use/1, delete/1, clear/1]).
 -export([obtain/0, obtain/1, release/0, release/1, transfer/1, transfer/2,
          set_limit/1, get_limit/0, info_keys/0, with_handle/1, with_handle/2,
-         info/0, info/1]).
+         info/0, info/1, clear_read_cache/0]).
 -export([ulimit/0]).
 
 -export([start_link/0, start_link/2, init/1, handle_call/3, handle_cast/2,
@@ -163,6 +163,8 @@
 -define(OBTAIN_LIMIT(LIMIT), trunc((LIMIT * 0.9) - 2)).
 -define(CLIENT_ETS_TABLE, file_handle_cache_client).
 -define(ELDERS_ETS_TABLE, file_handle_cache_elders).
+
+-include("rabbit.hrl"). % For #amqqueue record definition.
 
 %%----------------------------------------------------------------------------
 
@@ -580,6 +582,38 @@ info_keys() -> ?INFO_KEYS.
 
 info() -> info(?INFO_KEYS).
 info(Items) -> gen_server2:call(?SERVER, {info, Items}, infinity).
+
+clear_read_cache() ->
+    gen_server2:cast(?SERVER, clear_read_cache),
+    clear_vhost_read_cache(rabbit_vhost:list()).
+
+clear_vhost_read_cache([]) ->
+    ok;
+clear_vhost_read_cache([VHost | Rest]) ->
+    clear_queue_read_cache(rabbit_amqqueue:list(VHost)),
+    clear_vhost_read_cache(Rest).
+
+clear_queue_read_cache([]) ->
+    ok;
+clear_queue_read_cache([#amqqueue{pid = MPid, slave_pids = SPids} | Rest]) ->
+    Fun = fun(_, State) ->
+        clear_process_read_cache(),
+        State
+    end,
+    rabbit_amqqueue:run_backing_queue(MPid, rabbit_variable_queue, Fun),
+    [rabbit_amqqueue:run_backing_queue(SPid, rabbit_variable_queue, Fun)
+     || SPid <- SPids],
+    clear_queue_read_cache(Rest).
+
+clear_process_read_cache() ->
+    [
+      begin
+          Handle1 = reset_read_buffer(Handle),
+          put({Ref, fhc_handle}, Handle1)
+      end ||
+      {{Ref, fhc_handle}, Handle} <- get(),
+      size(Handle#handle.read_buffer) > 0
+    ].
 
 %%----------------------------------------------------------------------------
 %% Internal functions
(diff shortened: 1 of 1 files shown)


src/file_handle_cache.erl
```diff
@@ -581,6 +583,43 @@ info_keys() -> ?INFO_KEYS.
...
 
+clear_read_cache() ->
+    gen_server2:cast(?SERVER, clear_read_cache),
+    clear_vhost_read_cache(rabbit_vhost:list()).
+
+clear_vhost_read_cache([VHost | Rest]) ->
+    clear_queue_read_cache(rabbit_amqqueue:list(VHost)),
+    clear_vhost_read_cache(Rest);
+clear_vhost_read_cache([]) ->
+    ok.
```
videlalvaro: Just a pet peeve, but can you make the empty list case the first function clause? It seems to be the style favoured in this file as well: https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-196/src/file_handle_cache.erl#L763

Same comment for the rest of the changes

src/file_handle_cache.erl
```diff
@@ -581,6 +583,43 @@ info_keys() -> ?INFO_KEYS.
...
+
+clear_queue_read_cache(
+  [#amqqueue{name = #resource{virtual_host = VHost, name = Name}, pid = Pid}
+  | Rest]) ->
+    rabbit_log:info(
+      "Clear read buffer cache for queue \"~s\" on vhost \"~s\"~n",
+      [Name, VHost]),
+    rabbit_amqqueue:run_backing_queue(Pid, rabbit_variable_queue,
+      fun(_, State) ->
+          clear_process_read_cache(),
```
videlalvaro: Why do you need to run this code in the context of the backing queue? `clear_process_read_cache/0` will call put in the context of the backing queue Pid if I'm not mistake, so that process dictionary will be used.

videlalvaro: Let me guess… a quick glance at the code tells the `file_handle_cache` module is the one doing the `put/2` and what not, but it's being called from the context of the queue process…

dumbbell:
@videlalvaro: I moved the clauses.

dumbbell:
I modified the code to empty the cache from the slave processes as well.

Note that `file_handle_cache.erl` now includes `rabbit.hrl` so it grows a second dependency to RabbitMQ...

videlalvaro:
What if you move the clear_queue_read_cache closer to the queue logic? Is it worth?

dumbbell:
I would like to keep the read/write buffers logic contained in `file_handle_cache.erl` as much as possible because this is something we want to get rid of. I would prefer to keep the queue implementation clean from this feature.

dumbbell:
> Let me guess… a quick glance at the code tells the file_handle_cache module is the one doing the put/2 and what not, but it's being called from the context of the queue process…

Yes, the cache and many other informations are stored in the dictionary of the process opening/reading the file.

src/file_handle_cache.erl
```diff
@@ -581,6 +583,46 @@ info_keys() -> ?INFO_KEYS.
...
+  | Rest]) ->
+    rabbit_log:info(
+      "Clear read buffer cache for queue \"~s\" on vhost \"~s\"~n",
+      [Name, VHost]),
+    Fun = fun(_, State) ->
+        clear_process_read_cache(),
+        State
+    end,
+    rabbit_amqqueue:run_backing_queue(MPid, rabbit_variable_queue, Fun),
+    [rabbit_amqqueue:run_backing_queue(SPid, rabbit_variable_queue, Fun)
```
videlalvaro: Do we need to clear the cache on the slave Pids? Wouldn't that clear caches on a separate node?

src/file_handle_cache.erl
```diff
@@ -581,6 +583,46 @@ info_keys() -> ?INFO_KEYS.
...
+clear_queue_read_cache([]) ->
+    ok;
+clear_queue_read_cache(
+  [#amqqueue{name = #resource{virtual_host = VHost, name = Name},
+             pid = MPid, slave_pids = SPids}
+  | Rest]) ->
+    rabbit_log:info(
+      "Clear read buffer cache for queue \"~s\" on vhost \"~s\"~n",
+      [Name, VHost]),
+    Fun = fun(_, State) ->
```
videlalvaro: can you apply Emacs Erlang's formatting to all these changes?

dumbbell:
> Do we need to clear the cache on the slave Pids? Wouldn't that clear caches on a separate node?

Yes, this will clear the cache on other nodes. In fact, the command as it is clears all read buffers from all queues on all vhosts, so all nodes are affected.

Maybe I should limit it to the queues hosted on the node running the command, what do you think?

videlalvaro:
>Maybe I should limit it to the queues hosted on the node running the command, what do you think?

What impact could this command have on a node that's heavily loaded?

dumbbell:
Hard to say, it depends on the load obviously. I guess that walking through the dictionary's keys and rewriting some of them is cheap, but there could be many many queues with many buffers.

Then, then garbage collector needs to free all those binaries.

Anyway, after more thinking, I would expect this command to limit its action to the node I'm running it on. I will push a new commit.

michaelklishin:
This function will be used only when you want to reduce RAM use of a node. I think invalidating caches on mirrors is OK. The caches may or may not be around next month, too.

src/file_handle_cache.erl
```diff
@@ -581,6 +583,42 @@ info_keys() -> ?INFO_KEYS.
...
+    clear_queue_read_cache(rabbit_amqqueue:list(VHost)),
+    clear_vhost_read_cache(Rest).
+
+clear_queue_read_cache([]) ->
+    ok;
+clear_queue_read_cache([#amqqueue{pid = MPid, slave_pids = SPids} | Rest]) ->
+    %% Limit the action to the current node.
+    Pids = [P || P <- [MPid | SPids], node(P) =:= node()],
+    %% This function is executed in the context of the backing queue
+    %% process because the read buffer is stored in the processo
```
videlalvaro: processo small "spanish" typo :-)

dumbbell:
> Can you apply Emacs Erlang's formatting to all these changes?

Damn, I only see this comment now... I don't know all the rules of the Emacs style. As it doesn't match the style I'm used to, what should I change in this code?

Once we move to a common build system, I should take some time to add an Elvis configuration to our projects.

videlalvaro:
I will push a new file with the changes. In Emacs with the Erlang mode just select the region and apply formatting, on other editors, I have no idea

PR #202, opened 2015-06-29
Be quiet about sockets closed during SSL upgrade

legoscia:
Log such errors on debug level only, since this is probably a TCP healthcheck.


dumbbell:
Thank you!

PR #203, opened 2015-06-29
using os:timestamp/0 since erlang:now/0 is deprecated

marcelog:
Hi guys,

This pull was opened following a suggestion I got in this one [https://github.com/jbrisbin/rabbit_common/pull/15](https://github.com/jbrisbin/rabbit_common/pull/15). It replaces the calls to `erlang:now/0` which is now deprecated ([http://www.erlang.org/news/85](http://www.erlang.org/news/85)) with a call to `os:timestamp/0`.

A CA was sent via email.

Any thoughts?

Thanks in advance.

### DIFF
--- a/src/file_handle_cache.erl
+++ b/src/file_handle_cache.erl
@@ -526,7 +526,7 @@ clear(Ref) ->
       end).
 
 set_maximum_since_use(MaximumAge) ->
-    Now = now(),
+    Now = os:timestamp(),
     case lists:foldl(
            fun ({{Ref, fhc_handle},
                  Handle = #handle { hdl = Hdl, last_used_at = Then }}, Rep) ->
@@ -695,7 +695,7 @@ get_or_reopen(RefNewOrReopens) ->
         {OpenHdls, []} ->
             {ok, [Handle || {_Ref, Handle} <- OpenHdls]};
         {OpenHdls, ClosedHdls} ->
-            Oldest = oldest(get_age_tree(), fun () -> now() end),
+            Oldest = oldest(get_age_tree(), fun () -> os:timestamp() end),
             case gen_server2:call(?SERVER, {open, self(), length(ClosedHdls),
                                             Oldest}, infinity) of
                 ok ->
@@ -731,7 +731,7 @@ reopen([{Ref, NewOrReopen, Handle = #handle { hdl          = closed,
            end,
     case prim_file:open(Path, Mode) of
         {ok, Hdl} ->
-            Now = now(),
+            Now = os:timestamp(),
             {{ok, _Offset}, Handle1} =
                 maybe_seek(Offset, reset_read_buffer(
                                      Handle#handle{hdl              = Hdl,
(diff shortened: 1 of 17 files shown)


michaelklishin:
This looks OK to me. I don't see your CA in `contribute@rabbitmq.com`, can you send it to me directly? `mklishin` in Pivotal domain.

marcelog:
That was a quick :) I just sent you the signed CA. 

Thanks!

michaelklishin:
Got it.

videlalvaro:
The problem I see is that according to these docs: http://www.erlang.org/doc/apps/erts/time_correction.html#Dos_and_Donts yes, using `erlang:now/0` is deprecated, but is not just a job of search/replace for `os:timestamp/0`, as far as I can understand.

Then we have the problem of RabbitMQ supporting quite old Erlang versions, so we might need to do something like what's suggested on section 2.8 of that document.

michaelklishin:
`os:timestamp/0` is [available even in R13B03](http://erldocs.com/R13B03/kernel/os.html?search=os:tim&i=0#timestamp/0), the oldest release we support.

marcelog:
Hi guys,

Thanks for your comments! Both have valid points.

I settled on this change because as Michael states, the function is available in the minimum OTP version required by RabbitMQ, and that seemed like a *reasonable* decision that didn't require to refactor more code. That same document also states that:

> If you want the same format as returned by erlang:now/0, use erlang:timestamp/0.

The idea of the change was to keep the rest of the code "as is", since it depends on the result format. 

Thoughts?

michaelklishin:
@marcelog we are not opposed to this change (in fact, we did exactly that in a few place already), just need to the potential risk here some more. @dumbbell will cast his opinion quite soon.

dumbbell:
Hi @marcelog!

I share @videlalvaro's opinion: `os:timestamp()`, despite returning a time in the same format as `erlang:now()`, is not a drop-in replacement. As far as I understand, time returned by `os:timestamp()` is the OS wall clock, therefore it can go backward, whereas `erlang:now()` increases monotonically.

To fix our (ab)use of `erlang:now()`, all uses need to be audited to determine what we use this function for: a monotonically increasing value, a calculation of elapsed time, a date/time, and so on. Then, I like the advice given in Erlang's documentation of using a wrapper module to be compatible with the old and new APIs. There is such a [module provided in the Erlang source distribution](https://github.com/erlang/otp/blob/maint/erts/example/time_compat.erl)

marcelog:
Hi @dumbbell,

Thanks for stepping in!

It's true that `os:timestamp/0` is not monotonic, that's a good point. It is also true that the behavior of `erlang:now/0` would make the system stall to compensate time differences, and might have a nasty effect too.

I agree then, and it makes sense that perhaps one with a deeper understanding of the code works in a patch for each one of the use cases. Feel free to close this pull if you feel you want to go this way.

Thanks for the enriching discussion :)

PR #204, opened 2015-06-30
add detection for python2.7

ingwinlu:
In python2.7 `simplejson` has been removed from stdlib and replaced
with `json`. `codegen.py` already has a fallback import in place so no
changes needed there.

This way python2.7 can be detected if installed and rabbitmq-server
can be compiled even if python is bound to python3.


michaelklishin:
Thank you!

Will cherry-pick to `stable`.

PR #206, opened 2015-06-30
[WIP] optionally inject user_id into messages

ash-lshift:
Channels currently check the 'user ID' property of messages against the authenticated user: https://www.rabbitmq.com/validated-user-id.html

This PR adds a configuration entry `user_id_behaviour` which can take two values:

* `check` (the default) -- this is the current behaviour; it checks the 'user ID' property and closes the channel if it doesn't match
* `set` -- ignores the property value given with the message and replaces it with the true username.

### DIFF
--- a/src/rabbit_channel.erl
+++ b/src/rabbit_channel.erl
@@ -140,7 +140,9 @@
   %% used by "one shot RPC" (amq.
   reply_consumer,
   %% flow | noflow, see rabbitmq-server#114
-  delivery_flow
+  delivery_flow,
+  %% check | set
+  user_id_behaviour
 }).
 
 
@@ -341,6 +343,7 @@ init([Channel, ReaderPid, WriterPid, ConnPid, ConnName, Protocol, User, VHost,
              true   -> flow;
              false  -> noflow
            end,
+    UserIDBehaviour = rabbit_misc:get_env(rabbit, user_id_behaviour, check),
     State = #ch{state                   = starting,
                 protocol                = Protocol,
                 channel                 = Channel,
@@ -370,7 +373,8 @@ init([Channel, ReaderPid, WriterPid, ConnPid, ConnName, Protocol, User, VHost,
                 trace_state             = rabbit_trace:init(VHost),
                 consumer_prefetch       = 0,
                 reply_consumer          = none,
-                delivery_flow           = Flow},
+                delivery_flow           = Flow,
+                user_id_behaviour       = UserIDBehaviour},
     State1 = rabbit_event:init_stats_timer(State, #ch.stats_timer),
     rabbit_event:notify(channel_created, infos(?CREATION_EVENT_KEYS, State1)),
     rabbit_event:if_enabled(State1, #ch.stats_timer,
@@ -703,6 +707,18 @@ check_user_id_header(#'P_basic'{user_id = Claimed},
                    "'~s'", [Claimed, Actual])
     end.
 
+set_user_id_header(C = #content{properties = Props},
+                   #ch{user = #user{username = Username}}) ->
+    rabbit_binary_generator:clear_encoded_content(
+        C#content{properties = Props#'P_basic'{user_id = Username}}).
+
+check_or_set_user_id_header(Content = #content{properties = Props},
+                            State = #ch{user_id_behaviour = Behaviour}) ->
+    case Behaviour of
+      check -> check_user_id_header(Props, State), Content;
+      set -> set_user_id_header(Content, State)
+    end.
+
 check_expiration_header(Props) ->
     case rabbit_basic:parse_expiration(Props) of
         {ok, _}    -> ok;
@@ -887,9 +903,9 @@ handle_method(#'basic.publish'{exchange    = ExchangeNameBin,
     %% We decode the content's properties here because we're almost
     %% certain to want to look at delivery-mode and priority.
     DecodedContent = #content {properties = Props} =
+      check_or_set_user_id_header(
         maybe_set_fast_reply_to(
-          rabbit_binary_parser:ensure_content_decoded(Content), State),
-    check_user_id_header(Props, State),
+          rabbit_binary_parser:ensure_content_decoded(Content), State), State),
     check_expiration_header(Props),
     DoConfirm = Tx =/= none orelse ConfirmEnabled,
     {MsgSeqNo, State1} =


michaelklishin:
@videlalvaro @dumbbell thoughts? I think we should make it possible to do this in a plugin rather than adding it to the core.

ash-lshift:
FWIW I started off by tinkering with `rabbit_channel_interceptor`: https://github.com/rabbitmq/rabbitmq-server/compare/master...lshift:interceptors?expand=1 with a view to doing this as a plugin. However, it looks like it would require significant changes to the interface. (On the other hand, I am only aware of one plugin which uses it.)

videlalvaro:
I will prefer this to be a plugin as well.
What's the performance impact of this change? Is it doing this check/set for every message passing the broker?
The channel interceptor purposely doesn't let users modify `basic.publish` along with other methods, because they are on the hot path and will impact performance negatively.

ash-lshift:
I haven't done any benchmarking yet but one reason to do it with a core change is to reduce the performance impact. Given that `check_user_id_header` is already on the hot path I am not expecting much impact.

`rabbit_channel_interceptor` also already does a tiny bit of work on the hot path, so my feeling is that it is would be possible to rework it to do what I want with no impact. I have a little empirical data to support this feeling.

Anyway, I'm not wedded to either of these approaches so I look forward to hearing any better ideas you have.

videlalvaro:
The methods singled out here https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_channel_interceptor.erl#L54 AFAIK were actually benchmarked by @rade back when we introduced this feature.

You could try offering your own channel_interceptor which could be plugged in the same way different backing queues are plugged in today, not sure that's a good thing to do, but could be a way to go.

videlalvaro:
BTW, what's the point/use case for modifying the message? I think we avoid that as much as possible

ash-lshift:
Well, the only reason I have at the moment to modify messages is to inject authentication information. The reason for injecting that is so that custom exchanges or consumers can make finer-grained authorization decisions.

ash-lshift:
I've updated the [other branch](https://github.com/rabbitmq/rabbitmq-server/compare/master...lshift:interceptors?expand=1) and implemented this `user_id` thing as a plugin. Somehow my `rabbit_channel_interceptor_new` is slightly faster than the original when no interceptors are loaded. With a plugin on the hot path it is a bit slower, as expected.

What do you think about the other branch? If you like it better than this one I will close this PR and open a PR for the other branch.

videlalvaro:
I think your interceptor is faster because it caches in its state the modules that apply to a certain method as far as I understand the code, while the current interceptor asks for this information every time, so that's a nice improvement.

On the other hand you seem to skip a couple of validations along the way:

Like the one here: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_channel_interceptor.erl#L82 where we check the register module actually exist. 

And here: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_channel_interceptor.erl#L66 where we check that the method returned by the interceptor is the same as the one that was intercepted.

If I understand this correctly your interceptor seems to be compatible with the existing one? 

In any case if your interceptor needs to change the API, then we will need to upgrade the _sharing plugin_ along the way, and also I think your changes would have to be pushed to 3.6.x version since its changed a current public API.

Do you mind elaborating what's the advantage of this new interceptor over the existing one (besides the speed due to caching)?

ash-lshift:
> I think your interceptor is faster because it caches in its state the modules that apply to a certain method as far as I understand the code,

Yes.

> while the current interceptor asks for this information every time, so that's a nice improvement.
> On the other hand you seem to skip a couple of validations along the way:
> Like the one here: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_channel_interceptor.erl#L82 where we check the register module actually exist.
> And here: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_channel_interceptor.erl#L66 where we check that the method returned by the interceptor is the same as the one that was intercepted.

True; I'll add those back if we go ahead with this approach.

> If I understand this correctly your interceptor seems to be compatible with the existing one?

That was the intention. It would be simpler just to change the API though. Which would you prefer?

> In any case if your interceptor needs to change the API, then we will need to upgrade the sharing plugin along the way, and also I think your changes would have to be pushed to 3.6.x version since its changed a current public API.

That would be fine by me.

>Do you mind elaborating what's the advantage of this new interceptor over the existing one (besides the speed due to caching)?

The goal of the new interceptor is to be able to support more tweaks via plugins rather than core changes. For example it might be possible to push the direct reply-to extension to a plugin.

Again, though, I am not wedded to this approach. Another way I have considered is to move message delivery into the exchange. (That is, rather than `rabbit_exchange:route/2` returning a list of destinations and the channel delivering the message into the destinations, `route` returns nothing and does the delivery itself.) I think this might also be a cleaner way of doing 'side-effecty' routing like in https://github.com/rabbitmq/rabbitmq-delayed-message-exchange/blob/master/src/rabbit_exchange_type_delayed_message.erl.

ash-lshift:
(Another example of 'side-effecty routing': https://github.com/rabbitmq/rabbitmq-management-exchange/blob/master/src/rabbit_exchange_type_management.erl)

videlalvaro:
> Another way I have considered is to move message delivery into the exchange.

Exchanges have always been stateless routers in RabbitMQ so I think this is a big change. Also is the channel responsibility to know what happens with messages in order to issue replies to mandatory messages, or send back confirms, so I'm pretty sure this proposal won't advance.

wrt the interceptor: I'm all for having a better API if you can provide examples of use cases of said API. Of course if we introduce a new API, then these changes go to 3.6.x

videlalvaro:
BTW, what would be nice tho is to have a "fixed" current interceptor that does the caching you do in your new one. If the API doesn't change I think we could even include it in a future PATCH release

ash-lshift:
> Exchanges have always been stateless routers in RabbitMQ so I think this is a big change. Also is the channel responsibility to know what happens with messages in order to issue replies to mandatory messages, or send back confirms, so I'm pretty sure this proposal won't advance.

Fair enough.

> BTW, what would be nice tho is to have a "fixed" current interceptor that does the caching you do in your new one. If the API doesn't change I think we could even include it in a future PATCH release

You mean similar to the `backing_queue_module` config entry?

videlalvaro:
So the interceptor has two APIs, the one used in the callbacks by other interceptors, and the one used by the channel.

AFAIK, having an interceptor that caches `applies_to` and so on like yours only needs to modify the API used by the `rabbit_channel`, while at the same time improves performance. So I would welcome a change improving the current interceptor, since we can consider the current non-cached behaviour to be a bug.

ash-lshift:
I'll close this and continue with #214 .

Issue #151, opened 2015-05-07
Autotuning of the number of async I/O threads


michaelklishin:
See [this mailing list thread](https://groups.google.com/d/msg/rabbitmq-users/ijHVImbazFI/cxPtzoxKEUoJ).

I wonder if we can do autotuning that will work reasonably well on all supported platforms.

binarin:
So is default value like ```16 * erlang:system_info(logical_processors_available)``` good enough?

michaelklishin:
@binarin that sounds good but I'd add "but no fewer than 64".

binarin:
Looks like detecting number of CPUs is not as simple as it seems on some OSes. Lev Walkin posted the following snippet in erlang-russian ML:
```
ncpus() ->
    case erlang:system_info(logical_processors_available) of
        unknown -> % Happens on Mac OS X.
            erlang:system_info(schedulers);
        N -> N
    end.
```

michaelklishin:
@binarin combined with the "no fewer than 64" condition above, it sounds good to me. It's unlikely that there will be significantly more schedulers configured than there are logical cores. We might want to cap the maximum number, say, at 512 or 1024, too.

PR #212, opened 2015-07-02
Bump default +A value to 64

### LINKED ISSUE #176: Increase async I/O thread pool default
We are getting reports on the list about people running into busy spinning because they have 8-16 cores, for which our async thread pool size of 30 is not enough.

I've noticed that Riak uses 64 as default value. Should we consider using the same? Values around 90-100 work well for 16 cores, so 64 should cover 8 cores reasonably well, at the risk of having a few unused threads on 4 cores.

michaelklishin:
Based on user feedback, 30 feels a bit low. We have to recommend
users with 8 or more cores per node to bump the value to around 100.

This value is not very scientific and assumes roughly 16
I/O threads per core. This is also the default Riak 2.1 uses.

Fixes #176.

### DIFF
--- a/scripts/rabbitmq-defaults
+++ b/scripts/rabbitmq-defaults
@@ -34,6 +34,6 @@ MNESIA_BASE=${SYS_PREFIX}/var/lib/rabbitmq/mnesia
 ENABLED_PLUGINS_FILE=${SYS_PREFIX}/etc/rabbitmq/enabled_plugins
 
 PLUGINS_DIR="${RABBITMQ_HOME}/plugins"
-IO_THREAD_POOL_SIZE=30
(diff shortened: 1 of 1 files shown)


binarin:
What do you think about applying this (unscientific) guess automatically, if it isn't explicitly configured?

michaelklishin:
@binarin like in https://github.com/rabbitmq/rabbitmq-server/issues/151?

Gsantomaggio:
Should we update also Windows ? 
Currently it uses [`30`](https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-553/scripts/rabbitmq-server.bat#L101)

michaelklishin:
@Gsantomaggio good idea.

PR #214, opened 2015-07-03
change rabbit_channel_interceptor API

ash-lshift:
(This PR is work in progress but I'm hoping you will give feedback as I am working on it.)

At the moment `rabbit_channel_interceptor` can only intercept a subset of the AMQP methods. This restriction was due to performance concerns. This change lifts the restriction.

See #206 for previous discussion.

#206 ("optionally inject `user_id` into messages") was the motivating use case for changing the API but I think the change has a range of other applications. My colleagues and I came up with the following ways it might be useful in a few minutes, but I imagine other people will have other ideas.

* It could be used to implement #206 without further core changes. Consumers might be interested in other connection information than the username, and this could be done too.
* Generally, it might reduce feature creep in the core. For example there are a few extensions on the hot path of message publishing, for example [direct reply-to](https://www.rabbitmq.com/direct-reply-to.html), which could perhaps be moved into plugins.
* It could be used to transparently provide things like encryption/decryption/re-encryption/message signature verification as an alternative to TLS, compression/decompression, normalization of headers/body format, etc.
* It could be used to implement more specialized authorization, eg to allow a client to create but not delete queues, or to enforce some properties of objects the client creates.
* It could be used for more detailed auditing and protocol verification than the existing tracing mechanism.

### DIFF
--- a/src/rabbit_channel.erl
+++ b/src/rabbit_channel.erl
@@ -64,6 +64,7 @@
          prioritise_cast/3, prioritise_info/3, format_message_queue/2]).
 %% Internal
 -export([list_local/0, deliver_reply_local/3]).
+-export([get_vhost/1, get_user/1]).
 
 -record(ch, {
   %% starting | running | flow | closing
@@ -140,7 +141,8 @@
   %% used by "one shot RPC" (amq.
   reply_consumer,
   %% flow | noflow, see rabbitmq-server#114
-  delivery_flow
+  delivery_flow,
+  interceptor_state
 }).
 
 
@@ -182,6 +184,10 @@
 -export_type([channel_number/0]).
 
 -type(channel_number() :: non_neg_integer()).
+
+-export_type([channel/0]).
+
+-type(channel() :: #ch{}).
 
 -spec(start_link/11 ::
         (channel_number(), pid(), pid(), pid(), string(),
@@ -370,12 +376,15 @@ init([Channel, ReaderPid, WriterPid, ConnPid, ConnName, Protocol, User, VHost,
                 trace_state             = rabbit_trace:init(VHost),
                 consumer_prefetch       = 0,
                 reply_consumer          = none,
-                delivery_flow           = Flow},
-    State1 = rabbit_event:init_stats_timer(State, #ch.stats_timer),
-    rabbit_event:notify(channel_created, infos(?CREATION_EVENT_KEYS, State1)),
-    rabbit_event:if_enabled(State1, #ch.stats_timer,
-                            fun() -> emit_stats(State1) end),
-    {ok, State1, hibernate,
+                delivery_flow           = Flow,
+                interceptor_state       = undefined},
+    State1 = State#ch{
+               interceptor_state = rabbit_channel_interceptor:init(State)},
+    State2 = rabbit_event:init_stats_timer(State1, #ch.stats_timer),
+    rabbit_event:notify(channel_created, infos(?CREATION_EVENT_KEYS, State2)),
+    rabbit_event:if_enabled(State2, #ch.stats_timer,
+                            fun() -> emit_stats(State2) end),
+    {ok, State2, hibernate,
      {backoff, ?HIBERNATE_AFTER_MIN, ?HIBERNATE_AFTER_MIN, ?DESIRED_HIBERNATE}}.
 
 prioritise_call(Msg, _From, _Len, _State) ->
@@ -425,14 +434,15 @@ handle_call(_Request, _From, State) ->
 
 handle_cast({method, Method, Content, Flow},
             State = #ch{reader_pid   = Reader,
-                        virtual_host = VHost}) ->
+                        interceptor_state = IState}) ->
     case Flow of
         flow   -> credit_flow:ack(Reader);
         noflow -> ok
     end,
-    try handle_method(rabbit_channel_interceptor:intercept_method(
-                        expand_shortcuts(Method, State), VHost),
-                      Content, State) of
+    Method1 = expand_shortcuts(Method, State),
+    {Method2, Content1} = rabbit_channel_interceptor:intercept_in(
+                            Method1, Content, IState),
+    try handle_method(Method2, Content1, State) of
         {reply, Reply, NewState} ->
             ok = send(Reply, NewState),
             noreply(NewState);
@@ -442,7 +452,7 @@ handle_cast({method, Method, Content, Flow},
             {stop, normal, State}
     catch
         exit:Reason = #amqp_error{} ->
-            MethodName = rabbit_misc:method_record_type(Method),
+            MethodName = rabbit_misc:method_record_type(Method2),
             handle_exception(Reason#amqp_error{method = MethodName}, State);
         _:Reason ->
             {stop, {Reason, erlang:get_stacktrace()}, State}
@@ -1979,3 +1989,7 @@ erase_queue_stats(QName) ->
     [erase({queue_exchange_stats, QX}) ||
         {{queue_exchange_stats, QX = {QName0, _}}, _} <- get(),
         QName0 =:= QName].
+
+get_vhost(#ch{virtual_host = VHost}) -> VHost.
+
+get_user(#ch{user = User}) -> User.
--- a/src/rabbit_channel_interceptor.erl
+++ b/src/rabbit_channel_interceptor.erl
@@ -14,76 +14,102 @@
 %% Copyright (c) 2007-2015 Pivotal Software, Inc.  All rights reserved.
 %%
 
-%% Since the AMQP methods used here are queue related,
-%% maybe we want this to be a queue_interceptor.
-
 -module(rabbit_channel_interceptor).
 
 -include("rabbit_framing.hrl").
 -include("rabbit.hrl").
 
--export([intercept_method/2]).
+-export([init/1, intercept_in/3]).
 
 -ifdef(use_specs).
 
--type(intercept_method() :: rabbit_framing:amqp_method_name()).
+-type(method_name() :: rabbit_framing:amqp_method_name()).
 -type(original_method() :: rabbit_framing:amqp_method_record()).
 -type(processed_method() :: rabbit_framing:amqp_method_record()).
+-type(original_content() :: rabbit_types:maybe(rabbit_types:content())).
+-type(processed_content() :: rabbit_types:maybe(rabbit_types:content())).
+-type(interceptor_state() :: term()).
 
 -callback description() -> [proplists:property()].
-
--callback intercept(original_method(), rabbit_types:vhost()) ->
-    processed_method() | rabbit_misc:channel_or_connection_exit().
-
-%% Whether the interceptor wishes to intercept the amqp method
--callback applies_to(intercept_method()) -> boolean().
+%% Derive some initial state from the channel. This will be passed back
+%% as the third argument of intercept/3.
+-callback init(rabbit_channel:channel()) -> interceptor_state().
+-callback intercept(original_method(), original_content(),
+                    interceptor_state()) ->
+    {processed_method(), processed_content()} |
+    rabbit_misc:channel_or_connection_exit().
+-callback applies_to() -> list(method_name()).
 
 -else.
 
 -export([behaviour_info/1]).
 
 behaviour_info(callbacks) ->
-    [{description, 0}, {intercept, 2}, {applies_to, 1}];
+    [{description, 0}, {init, 1}, {intercept, 3}, {applies_to, 0}];
 behaviour_info(_Other) ->
     undefined.
 
 -endif.
 
-%%----------------------------------------------------------------------------
+init(Ch) ->
+    Mods = [M || {_, M} <- rabbit_registry:lookup_all(channel_interceptor)],
+    check_no_overlap(Mods),
+    [{Mod, Mod:init(Ch)} || Mod <- Mods].
 
-intercept_method(#'basic.publish'{} = M, _VHost) -> M;
-intercept_method(#'basic.ack'{}     = M, _VHost) -> M;
-intercept_method(#'basic.nack'{}    = M, _VHost) -> M;
-intercept_method(#'basic.reject'{}  = M, _VHost) -> M;
-intercept_method(#'basic.credit'{}  = M, _VHost) -> M;
-intercept_method(M, VHost) ->
-    intercept_method(M, VHost, select(rabbit_misc:method_record_type(M))).
+check_no_overlap(Mods) ->
+    check_no_overlap1([sets:from_list(Mod:applies_to()) || Mod <- Mods]).
 
-intercept_method(M, _VHost, []) ->
-    M;
-intercept_method(M, VHost, [I]) ->
-    M2 = I:intercept(M, VHost),
-    case validate_method(M, M2) of
-        true ->
-            M2;
-        _   ->
-            internal_error("Interceptor: ~p expected "
-                                "to return method: ~p but returned: ~p",
-                                [I, rabbit_misc:method_record_type(M),
-                                 rabbit_misc:method_record_type(M2)])
-    end;
-intercept_method(M, _VHost, Is) ->
-    internal_error("More than one interceptor for method: ~p -- ~p",
-                   [rabbit_misc:method_record_type(M), Is]).
+%% Check no non-empty pairwise intersection in a list of sets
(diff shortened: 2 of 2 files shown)


src/rabbit_channel.erl
```diff
@@ -411,6 +415,7 @@ handle_call({info, Items}, _From, State) ->
     end;
 
 handle_call(refresh_config, _From, State = #ch{virtual_host = VHost}) ->
+    % TODO: re-init interceptor_state?
```
videlalvaro: I think the interceptor state only needs to be updated when plugins are enabled/disabled

src/rabbit_channel_interceptor_new.erl
```diff
@@ -0,0 +1,67 @@
...
+  %io:format("R=~p~n", [R]),
+  R.
+
+lookup_all(Class) ->
+  [{Class, M} || {_, M} <- rabbit_registry:lookup_all(Class)].
+
+to_fn(Mods, Ch) ->
+  case Mods of
+    [] -> fun(M, C) -> {M, C} end;
+    [Mod] -> to_fn1(Mod, Ch)
```
videlalvaro: this assumes only one interceptor is defined or?

src/rabbit_channel_interceptor_new.erl
```diff
@@ -0,0 +1,67 @@
...
+-else.
+
+% TODO
+
+-endif.
+
+init(Ch) ->
+  R = (catch begin
+    Mods = lookup_all(?OLD_REGISTRY_CLASS) ++ lookup_all(?NEW_REGISTRY_CLASS),
+    {Hot, Cold} = lists:partition(fun ({_, M}) -> is_hot(M) end, Mods),
```
videlalvaro: I don't understand the point of this partitioning

src/rabbit_channel.erl
ash-lshift: It would currently fail; I will deal with that.

src/rabbit_channel_interceptor_new.erl
ash-lshift: Yes, I am planning to check that no two interceptors apply to the same method.

src/rabbit_channel_interceptor_new.erl
ash-lshift: The aim is to avoid applying 'cold' interceptors to 'hot' methods. If there are only 'cold' interceptors then we only ever apply `fun (M, C) -> {M, C} end` to 'hot' methods/content. This is possibly a premature optimization. I'll do another benchmark without it.

src/rabbit_channel_interceptor_new.erl
```diff
@@ -0,0 +1,67 @@
...
+    [] -> fun(M, C) -> {M, C} end;
+    [Mod] -> to_fn1(Mod, Ch)
+  end.
+
+to_fn1({channel_interceptor, Mod}, Ch) ->
+  VHost = rabbit_channel:get_vhost(Ch),
+  fun (M, C) -> {Mod:intercept(M, VHost), C} end;
+
+to_fn1({channel_interceptor_new, Mod}, Ch) ->
+  St = Mod:init(Ch),
```
videlalvaro: `Mod:init` is an interceptor callback right? What is expected of said function? What state does it tracks?

ash-lshift:
It is proving quite tricky to continue to support the old `rabbit_channel_interceptor` API. What do you think about just changing it? (That is, removing the current `rabbit_channel_interceptor` and renaming `rabbit_channel_interceptor_new` to `rabbit_channel_interceptor`)

videlalvaro:
I think it's fine to replace it, provided there's a migration path for the sharding plugin, and of course then this will go into 3.6.x

src/rabbit_channel.erl
ash-lshift: The comment `refresh_config` is gone.

`rabbit_channel_interceptor` can now handle the case where an interceptor plugin is unloaded while a channel is still open.

src/rabbit_channel_interceptor_new.erl
ash-lshift: It now supports multiple interceptors so long as they don't overlap. (This should be the same as the original behaviour.)

src/rabbit_channel_interceptor_new.erl
ash-lshift: I'll see about removing this now.

src/rabbit_channel_interceptor_new.erl
ash-lshift: Added a comment for that.

src/rabbit_channel_interceptor_new.erl
ash-lshift: Seems to make no difference. Removed.

videlalvaro:
@ash-lshift just ping me when you think I should go for a new review of the code

ash-lshift:
@videlalvaro It's probably ready for another look when you get a moment, thanks.

src/rabbit_channel.erl
```diff
@@ -1979,3 +1987,7 @@ erase_queue_stats(QName) ->
     [erase({queue_exchange_stats, QX}) ||
         {{queue_exchange_stats, QX = {QName0, _}}, _} <- get(),
         QName0 =:= QName].
+
+get_vhost(State = #ch{virtual_host = VHost}) -> VHost.
+
+get_user(State = #ch{user = User}) -> User.
```
videlalvaro: what are these two getters used for? I'd assume on a future interceptor?

src/rabbit_channel.erl
```diff
@@ -425,14 +433,14 @@ handle_call(_Request, _From, State) ->
...
     case Flow of
         flow   -> credit_flow:ack(Reader);
         noflow -> ok
     end,
-    try handle_method(rabbit_channel_interceptor:intercept_method(
-                        expand_shortcuts(Method, State), VHost),
-                      Content, State) of
+    Method1 = expand_shortcuts(Method, State),
+    {Method2, Content1} = rabbit_channel_interceptor:intercept_in(Method1, Content, IState),
+    try handle_method(Method2, Content1, State) of
```
videlalvaro: Now interceptors are allowed to modify the content sent by the user, right?

src/rabbit_channel_interceptor.erl
```diff
@@ -14,73 +14,105 @@
...
+                end,
+                sets:new(),
+                Sets),
+    ok.
+
+intercept_in(#'basic.publish'{} = M, C, S) -> intercept_in1(M, C, S);
+intercept_in(#'basic.ack'{} = M, C, S) -> intercept_in1(M, C, S);
+intercept_in(#'basic.nack'{} = M, C, S) -> intercept_in1(M, C, S);
+intercept_in(#'basic.reject'{} = M, C, S) -> intercept_in1(M, C, S);
+intercept_in(#'basic.credit'{} = M, C, S) -> intercept_in1(M, C, S);
```
videlalvaro: what's the point of all these function clauses if they all end up in `intercept_in1/3`

src/rabbit_channel_interceptor.erl
```diff
@@ -14,73 +14,105 @@
...
+    end.
+
+%% Turn a single interceptor module into a function
+to_fn1(Mod, Ch) ->
+    St = Mod:init(Ch),
+    fun (M, C) ->
+        % this little dance is because Mod might be unloaded at any point
+        case (catch {ok, Mod:intercept(M, C, St)}) of
+            {ok, R = {M2, _}} ->
+                case validate_method(M, M2) of
```
videlalvaro: since now the interceptor modifies the Content as well, don't we need to validate that's the returned value is correct as well? See rabbit_types:content() and rabbit_channel:do/3 spec.

src/rabbit_channel_interceptor.erl
```diff
@@ -14,73 +14,105 @@
...
+                {M1, C1} = Fn1(M, C),
+                Fn2(M1, C1)
+            end
+    end.
+
+%% Turn a single interceptor module into a function
+to_fn1(Mod, Ch) ->
+    St = Mod:init(Ch),
+    fun (M, C) ->
+        % this little dance is because Mod might be unloaded at any point
```
videlalvaro: So instead of refreshing the state cache we just handle the module going away right? I think it's fine since channels might not be so long-lived to have this be a frequent issue.

src/rabbit_channel_interceptor.erl
```diff
@@ -14,73 +14,105 @@
...
+    to_fn(Mods, Ch).
+
+%% Turn a list of interceptor modules into a single function
+to_fn(Mods, Ch) ->
+    case Mods of
+        [] -> fun(M, C) -> {M, C} end;
+        [Mod] -> to_fn1(Mod, Ch);
+        [Mod|Mods1] ->
+            Fn1 = to_fn1(Mod, Ch),
+            Fn2 = to_fn(Mods1, Ch),
```
videlalvaro: Can this be made tail recursive/refactored? I really have a hard time wrapping my head around what's going on on this function

videlalvaro:
Besides the small comments, can you provide a sample interceptor implementation or send a PR to the sharding plugin showing how this works?

src/rabbit_channel.erl
ash-lshift: Yes. That turned out to be necessary to support my original use case.

src/rabbit_channel.erl
ash-lshift: Yes. It was that or move the definition of `ch` to a header.

src/rabbit_channel_interceptor.erl
ash-lshift: Sure.

src/rabbit_channel_interceptor.erl
ash-lshift: Good point!

ash-lshift: (They used to be because the 'hot' methods were handled separately.)

ash-lshift:
Sure.

ash-lshift:
Reimplementation of #206 as a plugin: https://github.com/lshift/rabbitmq-inject-user-id/blob/master/src/rabbitmq_inject_user_id.erl

Changes for sharding plugin: https://github.com/rabbitmq/rabbitmq-sharding/pull/10

videlalvaro:
pretty cool, I'll take a look again at the PR

src/rabbit_channel_interceptor.erl
```diff
@@ -14,77 +14,100 @@
...
 -callback description() -> [proplists:property()].
-
--callback intercept(original_method(), rabbit_types:vhost()) ->
-    processed_method() | rabbit_misc:channel_or_connection_exit().
-
-%% Whether the interceptor wishes to intercept the amqp method
--callback applies_to(intercept_method()) -> boolean().
+%% Derive some initial state from the channel. This will be passed back
+%% as the third argument of intercept/3.
+-callback init(rabbit_channel:channel()) -> any.
```
videlalvaro: what if instead of `any` we make another type called interceptor_state or something like that so in the next callback `intercept/3` we specify we expect said state, it'd be better as documentation in the future.

src/rabbit_channel.erl
```diff
@@ -425,14 +433,14 @@ handle_call(_Request, _From, State) ->
...
+                        interceptor_state = IState}) ->
     case Flow of
         flow   -> credit_flow:ack(Reader);
         noflow -> ok
     end,
-    try handle_method(rabbit_channel_interceptor:intercept_method(
-                        expand_shortcuts(Method, State), VHost),
-                      Content, State) of
+    Method1 = expand_shortcuts(Method, State),
+    {Method2, Content1} = rabbit_channel_interceptor:intercept_in(Method1, Content, IState),
```
videlalvaro: Do you think there could be a reason for an interceptor to update its state as well? by returning `{Method2, Content1, NewIState}` for example?

videlalvaro: OK, considering that now IState is proplist of {Mod, State}, then I don't think my previous comments make sense, without overcomplicating the interceptor

src/rabbit_channel.erl
```diff
@@ -1979,3 +1987,7 @@ erase_queue_stats(QName) ->
     [erase({queue_exchange_stats, QX}) ||
         {{queue_exchange_stats, QX = {QName0, _}}, _} <- get(),
         QName0 =:= QName].
+
+get_vhost(State = #ch{virtual_host = VHost}) -> VHost.
```
videlalvaro: in both these getters `State` is not used and it gives a Warning while compiling

videlalvaro:
Except for the small warning while compiling `rabbit_channel.erl` I think we are ready to merge.

Also take a look at interceptor init spec change

src/rabbit_channel_interceptor.erl
ash-lshift: Good idea.

src/rabbit_channel.erl
ash-lshift: Agreed.

videlalvaro:
once you fix the last couple of things, do you think you could squash all the commits into one?

ash-lshift:
Warning fixed, spec changed, commits squashed.

dumbbell:
Hi @ash-lshift!

I have a few more comments on the style, otherwise it looks good to me too.

Feel free to amend the squashed commit and force-push the new one.

src/rabbit_channel.erl
```diff
@@ -370,12 +376,14 @@ init([Channel, ReaderPid, WriterPid, ConnPid, ConnName, Protocol, User, VHost,
...
                 reply_consumer          = none,
-                delivery_flow           = Flow},
-    State1 = rabbit_event:init_stats_timer(State, #ch.stats_timer),
-    rabbit_event:notify(channel_created, infos(?CREATION_EVENT_KEYS, State1)),
-    rabbit_event:if_enabled(State1, #ch.stats_timer,
-                            fun() -> emit_stats(State1) end),
-    {ok, State1, hibernate,
+                delivery_flow           = Flow,
+                interceptor_state       = undefined},
+    State1 = State#ch{interceptor_state = rabbit_channel_interceptor:init(State)},
```
dumbbell: Please wrap lines longer than 80 columns.

src/rabbit_channel.erl
dumbbell: Here too, this line should be wrapped.

src/rabbit_channel_interceptor.erl
```diff
@@ -14,77 +14,101 @@
...
--callback intercept(original_method(), rabbit_types:vhost()) ->
-    processed_method() | rabbit_misc:channel_or_connection_exit().
-
-%% Whether the interceptor wishes to intercept the amqp method
--callback applies_to(intercept_method()) -> boolean().
+%% Derive some initial state from the channel. This will be passed back
+%% as the third argument of intercept/3.
+-callback init(rabbit_channel:channel()) -> interceptor_state().
+-callback intercept(original_method(), original_content(), interceptor_state()) ->
+    {processed_method(), processed_content()} | rabbit_misc:channel_or_connection_exit().
```
dumbbell: These lines are too long.

ash-lshift:
@dumbbell: I have wrapped the lines.

videlalvaro:
Thanks @ash-lshift

michaelklishin:
Thank you, @ash-lshift!

ash-lshift:
Thanks all for your input.

src/rabbit_channel.erl
```diff
@@ -425,14 +434,15 @@ handle_call(_Request, _From, State) ->
...
-                        virtual_host = VHost}) ->
+                        interceptor_state = IState}) ->
     case Flow of
         flow   -> credit_flow:ack(Reader);
         noflow -> ok
     end,
-    try handle_method(rabbit_channel_interceptor:intercept_method(
-                        expand_shortcuts(Method, State), VHost),
-                      Content, State) of
+    Method1 = expand_shortcuts(Method, State),
```
videlalvaro: `expand_shortcuts` might throw an error, so this here makes the channel crash when no queue name is specified for things like `basic.consume` or `queue.bind` https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_channel.erl#L745

ash-lshift: D'oh, you're right. Want me to fix it?

videlalvaro: I'm on it :)

Issue #121, opened 2015-04-20
Support more queue master-to-node distribution strategies


gsogol:
Per request of @tsaleh:

Having an ability to create queues on different nodes via preconfigured a set of limited strategies would help alleviate a single node getting pounded. Multiple strategies can be given to consumers to pick from, similar to how most load balancers work: 1) # of exchanges/queues already bound on each node 1) by # of messages on each node (so create a queue on a node with least number of queues messages 2) by cpu/memory 3) # of current connections 4) by cpu/memory utilization

Leave the defaults as is, no arrangement but give an ability to configure. In my case, I would choose  the 1st one. I'm not saying do all of them but over time, you add more and more via a pluggable model.

michaelklishin:
Exchanges are just routing tables that are always replicated to all nodes.

So you'd like to control which node queue masters are declared on via strategies? Currently it's the node the client is connected to.

gsogol:
The queue that the client creates should be on a node where the specified strategy meets the criteria.

michaelklishin:
OK, I believe my understanding is correct then. Makes sense, thank you!

gsogol:
Thank you

Ayanda-D:
@michaelklishin, question, for the min-masters policy, do we count & compare bound queues across all vhosts per rabbit node? and not specifically the vhosts of the queue being declared?

michaelklishin:
@Ayanda-D all vhosts.

Ayanda-D:
Okay. Thank you.

michaelklishin:
Fixed by #217, https://github.com/rabbitmq/rabbitmq-test/pull/1, and https://github.com/rabbitmq/rabbitmq-website/pull/32.

michaelklishin:
…and https://github.com/rabbitmq/rabbitmq-test/pull/2.

PR #217, opened 2015-07-07
Introduce queue master distribution strategies

Ayanda-D:
Implemantation of issue https://github.com/rabbitmq/rabbitmq-server/issues/121, currently supporting queue master node selection in a cluster by;

1.  Minimum bound queues
2.  Random selection
3.  Client local node


src/rabbit_queue_location_client_local.erl
```diff
@@ -0,0 +1,47 @@
...
+%%% @end
+%%% Created : 19. Jun 2015
+%%%-------------------------------------------------------------------
+-module(rabbit_queue_location_client_local).
+
+-include("rabbit.hrl").
+
+-behaviour(rabbit_queue_master_locator).
+
+-export([description/0, queue_master_location/1, validate_policy/1]).
```
videlalvaro: We usually add specs to behaviours, see: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_exchange_type.erl#L19

videlalvaro: oops, sorry, I was looking at the wrong file. Please disregard my previous comment

src/rabbit_queue_master_location_misc.erl
```diff
@@ -0,0 +1,144 @@
...
+
+%% API
+-export([ lookup_master/2,
+          lookup_queue/2,
+          get_location/1,
+          get_location_by_config/1,
+          get_location_by_args/1,
+          get_location_by_policy/1,
+          all_nodes/0,
+          delay_ms/1,
```
videlalvaro: this is not used, besides it's the same as http://www.erlang.org/doc/man/timer.html#sleep-1

src/rabbit_queue_location_client_local.erl
```diff
@@ -0,0 +1,47 @@
...
+%%% http://www.mozilla.org/MPL/
+%%%
+%%% Software distributed under the License is distributed on an "AS IS"
+%%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See the
+%%% License for the specific language governing rights and limitations
+%%% under the License.
+%%%
+%%% The Original Code is RabbitMQ.
+%%%
+%%% @author Ayanda Dube <ayanda.dube@erlang-solutions.com>
```
videlalvaro: we don't use author tags in our code, nor the _Created_ one bellow, or the `@doc` one like bellow. We do have doc headers, see `gm.erl` for an example.

Author and Date are stored on git anyway.

src/rabbit_queue_location_client_local.erl
```diff
@@ -0,0 +1,47 @@
...
+%%---------------------------------------------------------------------------
+
+description() ->
+  [{description, <<"Set queue master node as the client local node">>}].
+
+queue_master_location(#amqqueue{}) ->
+  MasterNode = node(),
+  {ok, MasterNode}.
+
+validate_policy(_Args) -> ok.
```
videlalvaro: why do we have this callback if the code seems to return `ok` on all instances?

src/rabbit_queue_location_client_local.erl
Ayanda-D: is this okay for the last few lines?

%%%
%% Developed by Erlang Solutions Ltd on behalf of GoPivotal, Inc.
%% Copyright (c) 2007-2015 GoPivotal, Inc.  All rights reserved.
%%

michaelklishin: I'll ask if our legal team feels strongly about such header. FTR, Pivotal is technically Pivotal Software Inc. these days.

michaelklishin: @Ayanda-D I clarified this, please use the same headers as in other source files.

videlalvaro:
A few comments:

The idea of the `rabbit_queue_master_locator` behaviour is good, but there are a few details, some of them just coding style related, so I hope are easy to fix.

1) The behaviour expects a `validate_policy/1`, but then in all the implementations said callback just returns `ok`. I don't think that's needed, since there's the `rabbit_policy_validator` behaviour that covers `validate_policy/1` already.

2) If I understand correctly, the `rabbit_queue_location_min_masters` strategy counts bindings around nodes and vhosts and from there tries to infer which node is the best one. I don't think this is going to work since a RabbitMQ cluster replicates queue and exchange definitions, together with bindings to all nodes in the cluster, so whichever node you ask for said info (bindings, declared queues, etc), will and should give you the same answer. Also there's no need to `rpc_call` for said info, since each node has it locally.

3) I see that `rabbit_queue_master_location_misc` exports some functions that are not used. If a function is exported, then that function becomes part of the module API, and it's hard to make said function private later since we don't know why it was exported in the first place. 

4) `delay_ms/1` seems to do the same as Erlang's `timer:sleep/1`. Then `policy/1` could have been used here perhaps? https://github.com/rabbitmq/rabbitmq-server/pull/217/files#diff-86750e2c3649c6ea97321ef96f5207d5R88 Please correct other cases that might be similar to these ones.

5) Coding style: as a general rule we use the code indentation Emacs' Erlang mode provides by default. We use _80 chars as max line length_. Apart from that we try to keep case clauses aligned, see: https://github.com/rabbitmq/rabbitmq-server/pull/217/files#diff-86750e2c3649c6ea97321ef96f5207d5R110 if you can't keep them aligned due to line length, then move them to the next line. We try to use the same alignment for function guards as well. See https://github.com/rabbitmq/rabbitmq-server/pull/217/files#diff-86750e2c3649c6ea97321ef96f5207d5R55 vs https://github.com/rabbitmq/rabbitmq-server/pull/217/files#diff-86750e2c3649c6ea97321ef96f5207d5R46 The later is what Emacs suggests. Then small things like binaries, which we usually define like this: `<<"queue-master-location">>` not like `<< "queue-master-location" >>`. with spaces in between. We also don't use spaces before list elements on `-export` declarations.

6) the `rabbit_queue_master_location_misc` module seems to have helper functions and do policy validation as well. Misc modules should just implement helpers, but not do any behaviour things. Please implement the policy validator on its own module.

7) As discussed with @michaelklishin already, please use the same file headers we have in other files of the project. Author, creation date, and so on, are recorded on git already, plus Github will display you under the contributor's tab.

src/rabbit_amqqueue.erl
```diff
@@ -274,9 +274,15 @@ declare(QueueName, Durable, AutoDelete, Args, Owner, Node) ->
...
                                       gm_pids            = [],
                                       state              = live})),
-    Node = rabbit_mirror_queue_misc:initial_queue_node(Q, Node),
+
+    Node1 = case rabbit_queue_master_location_misc:get_location(Q)  of
+              {ok, Node0}  -> Node0;
+              {error, _}   -> Node
+            end,
+
+    Node1 = rabbit_mirror_queue_misc:initial_queue_node(Q, Node1),
```
videlalvaro: Is this an assertion? Otherwise this might crash since Node1 on 283 could not match the value on line 278

src/rabbit_queue_location_client_local.erl
```diff
@@ -0,0 +1,42 @@
...
+%%---------------------------------------------------------------------------
+%% Queue Master Location Callbacks
+%%---------------------------------------------------------------------------
+
+description() ->
+    [{description, <<"Locate queue master node as the client local node">>}].
+
+queue_master_location(#amqqueue{}) ->
+    MasterNode = node(),
+    {ok, MasterNode}.
```
videlalvaro: Perhaps there's no need for a Temp variable and this can just return `{ok, node()}`

src/rabbit_queue_location_min_masters.erl
```diff
@@ -0,0 +1,73 @@
...
+get_bound_queue_masters_per_vhost([], Acc) ->
+    lists:flatten(Acc);
+get_bound_queue_masters_per_vhost([VHost|RemVHosts], Acc) ->
+    Bindings          = rabbit_binding:list(VHost),
+    BoundQueueMasters = get_queue_master_per_binding(VHost, Bindings, []),
+    get_bound_queue_masters_per_vhost(RemVHosts, [BoundQueueMasters|Acc]).
+
+
+get_queue_master_per_binding(_VHost, [], BoundQueueNodes) -> BoundQueueNodes;
+get_queue_master_per_binding(VHost, [#binding{key=QueueName}|RemBindings],
```
videlalvaro: `key` is the routing key used for creating the binding, it's not the queue name.

```erlang
-record(binding, {
    source, %% exchange where the binding was issued
    key, %% routing key for the binding
    destination, %% queue or exchange destination of the binding, see also e2e RabbitMQ extension
    args = [] %% args used during binding creation
}).
```

src/rabbit_queue_location_validator.erl
```diff
@@ -0,0 +1,64 @@
...
+        undefined -> no_location_strategy;
+        Mode      -> module(Mode)
+    end;
+
+module(Strategy) when is_binary(Strategy) ->
+    case rabbit_registry:binary_to_type(Strategy) of
+        {error, not_found} -> no_location_strategy;
+        T ->
+            case rabbit_registry:lookup_module(queue_master_locator, T) of
+                {ok, Module} -> {ok, Module};
```
videlalvaro: on the channel interceptor we also check this: `code:which(M) =/= non_existing` to make sure the module actually exists. Keep in mind plugins might add their own queue locator modules, which might contain errors

src/rabbit_queue_master_location_misc.erl
```diff
@@ -0,0 +1,91 @@
...
+                Reply={ok, _Node} -> Reply
+            end;
+        Reply={ok, _Node} -> Reply
+    end.
+
+get_location_by_args(Queue=#amqqueue{arguments=Args}) ->
+    case proplists:lookup(<<"queue-master-location">> , Args) of
+        {<<"queue-master-location">> , Strategy}  ->
+            case rabbit_queue_location_validator:validate_strategy(Strategy) of
+                {ok, CB} -> CB:queue_master_location(Queue);
```
videlalvaro: wouldn't it be better if these `get_location_*` functions return the module that implements the strategy, and then `CB:queue_master_location/1` is called in just one place?

src/rabbit_amqqueue.erl
rtraschke: I think we need to little bit deeper understanding of what rabbit_mirror_queue_misc:initial_queue_node/2 does. Because we're now passing in a potentially different node.

videlalvaro: I think the key to understand that is the comment here: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_amqqueue.erl#L262 

There are cases where the Master node might change. I suggest to see what's going on on the current version of rabbit vs. this branch when mirroring is in place and the `ha-params` policy is in place

videlalvaro:
@Ayanda-D ping me once you think we should do a new review round

michaelklishin:
@Ayanda-D is this ready for another round of reviews?

Ayanda-D:
@michaelklishin yes it is, and https://github.com/rabbitmq/rabbitmq-test/pull/1.

src/rabbit_amqqueue.erl
michaelklishin: I believe this comment from @videlalvaro wasn't addressed. @rtraschke was this ever investigated?

src/rabbit_queue_master_location_misc.erl
```diff
@@ -0,0 +1,96 @@
...
+                Reply0={ok, _Module} -> Reply0
+             end,
+
+    case Reply1 of
+        {ok, CB} -> CB:queue_master_location(Queue);
+        Error    -> Error
+    end.
+
+get_location_mod_by_args(#amqqueue{arguments=Args}) ->
+    case proplists:lookup(<<"queue-master-location">> , Args) of
```
michaelklishin: This is not the attribute name [our preliminary docs](https://github.com/rabbitmq/rabbitmq-website/pull/32/files) use. I find `queue-master-location` to be incorrect: this is not location that we're defining but location strategy. I'm OK with renaming it to `x-queue-master-locator` instead of `x-queue-master-location-strategy` and updating the docs.

src/rabbit_amqqueue.erl
Ayanda-D: @michaelklishin please also look at my conversations with @videlalvaro on the slack room.

michaelklishin:
@Ayanda-D thanks. I immediately see that the argument name is not what is [in the docs](https://github.com/rabbitmq/rabbitmq-website/pull/32/files). We should rename it to `x-queue-master-location-strategy` or `x-queue-master-locator` ("locator" is already used in a few places in the code).

@dumbbell which one of the above do you prefer?

src/rabbit_amqqueue.erl
michaelklishin: @Ayanda-D can you please tell me the date it took place on?

Ayanda-D: @michaelklishin please have a look at our conversations on 17/07/2015 and 20/07/2015 on my understanding of this. For the changes you're commenting on, @videlalvaro also already had a look at them. Just the test code which hadn't gone through PR.

michaelklishin: @Ayanda-D thanks for the advice but I think I'll figure out how my team does QA and how many people need to review a change before we are happy with it.

Ayanda-D: @michaelklishin cool, no problem. @videlalvaro did mention how critical this part of the system is, so the more the reviews on this, the better. keep us posted.

michaelklishin: OK, so `ha-params` effects on this change have been discussed out-of-band. The conclusion is: `ha-params` with specific nodes listed currently overrides the locator. This is fine: using them at the same time makes little sense. We'll mention that in the docs.

Ayanda-D: Yes, I think thats what I mentioned on 20/07/2015 [6:36 PM], on slack. Just needed confirmation from you guys. If that's what we're concluding on this, then cool.

src/rabbit_queue_master_location_misc.erl
```diff
@@ -0,0 +1,96 @@
...
+        undefined ->  {error, "queue-master-location policy undefined"};
+        Strategy  ->
+            case rabbit_queue_location_validator:validate_strategy(Strategy) of
+                Reply={ok, _CB} -> Reply;
+                Error    -> Error
+            end
+    end.
+
+get_location_mod_by_config(#amqqueue{}) ->
+    case application:get_env(rabbit, queue_master_location) of
```
michaelklishin: The above comment also applies to the app config key, `queue_master_location` is not precise (just in case: we don't want the `x-` prefix there, of course:)).

michaelklishin:
@Ayanda-D this looks good. I will continue with various functional tests but there's one last change we'd like to see: lets rename the `queue-master-location` header to `x-queue-master-locator` and `rabbit.queue_master_location` config key to `rabbit.queue_master_locator`. This is more to the point and also what module names already use.

Thank you.

michaelklishin:
Renaming the header and config key aside, this LGTM.

Ayanda-D:
@michaelklishin thanks for the review and feedback. I've updated the header and config key. Please have a look to make sure everything is okay.

videlalvaro:
@Ayanda-D @michaelklishin as I asked when I was reviewing this PR, what's the point of counting bindings here:

https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_queue_location_min_masters.erl

How is that expected/supposed to work?

Ayanda-D:
@videlalvaro its to ensure we calculate min-masters from bound queues only, across all vhosts.

videlalvaro:
@Ayanda-D so this will filter out queues that have no bindings, the group them by node where they live, then count them?

Ayanda-D:
@videlalvaro you mean this will filter out queues that have bindings, group them by node where they live, then count them? yes.

videlalvaro:
@Ayanda-D go it. I'd assume "filter-out" means "keeps out of the resulting list" but anyway. Thanks for clarifying

Ayanda-D:
@videlalvaro cool. just clarifying the context of "them". but its all good :-)

PR #218, opened 2015-07-07
Fixes #179

### LINKED ISSUE #179: Enable kernel polling and a sensible process limit unconditionally
Currently when overriding VM args it is all too easy to forget to include other default flags. I feel +K and +P should be included unconditionally: I haven't seen them tweaked for a good reason in many years.

hairyhum:
@michaelklishin it is for you.

### DIFF
--- a/scripts/rabbitmq-server.bat
+++ b/scripts/rabbitmq-server.bat
@@ -90,9 +90,9 @@ if "!RABBITMQ_IO_THREAD_POOL_SIZE!"=="" {
 +W w ^
 +A "!RABBITMQ_IO_THREAD_POOL_SIZE!" ^
 +P 1048576 ^
--kernel inet_default_connect_options "[{nodelay, true}]" ^
 !RABBITMQ_LISTEN_ARG! ^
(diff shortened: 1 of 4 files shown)


scripts/rabbitmq-server.bat
```diff
@@ -90,9 +90,9 @@ if "!RABBITMQ_IO_THREAD_POOL_SIZE!"=="" {
 +W w ^
 +A "!RABBITMQ_IO_THREAD_POOL_SIZE!" ^
 +P 1048576 ^
--kernel inet_default_connect_options "[{nodelay, true}]" ^
 !RABBITMQ_LISTEN_ARG! ^
 !RABBITMQ_SERVER_ERL_ARGS! ^
+-kernel inet_default_connect_options "[{nodelay, true}]" ^
```
dumbbell: I don't see any `+K true` flag for the Windows startup script. Is it on purpose?

michaelklishin: Yes, the Erlang runtime does not support kernel polling on Windows.

PR #229, opened 2015-07-22
file_handle_cache: Add a config option to disable buffering

### LINKED ISSUE #226: Add a RabbitMQ option to disable FHC's read/write buffers
This will allow user to test how RabbitMQ behaves with their workload when especially the FHC read buffer is disabled. The results will be used to determine if we can get rid of it.

dumbbell:
To disable read buffering, one has to put the following line in
rabbitmq.config:

```erlang
[
  {rabbit, [
    {fhc_read_buffering, false}
  ]}
].
```

Fixes #226.

### DIFF
--- a/ebin/rabbit_app.in
+++ b/ebin/rabbit_app.in
@@ -28,6 +28,8 @@
          {channel_max, 0},
          {heartbeat, 580},
          {msg_store_file_size_limit, 16777216},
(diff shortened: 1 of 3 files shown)


michaelklishin:
Should we log when buffering is enabled/disabled? For debugging/support purposes.

Issue #239, opened 2015-07-29
Modify the default path of RABBITMQ_CONF_ENV_FILE within rabbitmq-defaults.bat


Jakauppila:
Since the value of RABBITMQ_BASE is already set to `!APPDATA!\RabbitMQ` if not overridden, modify the the default value to be `!RABBITMQ_BASE!\rabbitmq-env-conf.bat` rather than `!APPDATA!\RabbitMQ\rabbitmq-env-conf.bat`

Also, since itself `rabbitmq-env-conf.bat` contains the prefix-less variables, change the value on line 36 of `rabbitmq-defaults.bat` to `RABBITMQ_CONF_ENV_FILE` so the invoke `call !RABBITMQ_CONF_ENV_FILE!` in the other .bats work correctly if not explicitly overridden.

Plorax:
Hi guys, it appears this file 'rabbitmq-env-conf.bat' does not exist anymore.

michaelklishin:
Configuration and environment variable files are not created by default.

PR #240, opened 2015-07-29
Modify the default path of RABBITMQ_CONF_ENV_FILE within rabbitmq-defaults.bat 

Jakauppila:
Modifies the default path of `RABBITMQ_CONF_ENV_FILE` (and adds the RABBITMQ_ prefix). 

Also adds quotes around the call within `rabbitmq-env.bat`.

Fixes https://github.com/rabbitmq/rabbitmq-server/issues/239


michaelklishin:
Will cherry-pick into stable manually.

PR #242, opened 2015-07-29
Fixes a logic problem with NODE_IP_ADDRESS and NODE_PORT in rabbitmq-env.bat.

Jakauppila:
Fixes the logic for NODE_IP_ADDRESS and NODE_PORT. First, check for the existence of the environment variable, then the definition if set via conf file, then proceed with the original logic on setting the IP/Port variables.

As it was configured, RABBITMQ_LISTEN_ARG was not getting configured properly as it was before.

### DIFF
--- a/scripts/rabbitmq-env.bat
+++ b/scripts/rabbitmq-env.bat
@@ -84,21 +84,26 @@ REM       set RABBITMQ_NODE_PORT=5672
 REM    )
 REM )
 
-REM DOUBLE CHECK THIS LOGIC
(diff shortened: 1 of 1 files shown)


michaelklishin:
@Jakauppila if you intend your fixes to be in 3.5.x, make sure the PR is submitted against `stable`.

michaelklishin:
I'm assuming this is intended for `3.5.5` => closing expecting it to be re-submitted against `stable`.

If I'm wrong, just leave a comment.

PR #243, opened 2015-07-29
 Fixes a logic problem with NODE_IP_ADDRESS and NODE_PORT in rabbitmq-env.bat.

Jakauppila:
Fixes the logic for NODE_IP_ADDRESS and NODE_PORT. First, check for the existence of the environment variable, then the definition if set via conf file, then proceed with the original logic on setting the IP/Port variables.

As it was configured, RABBITMQ_LISTEN_ARG was not getting configured properly as it was before.


dumbbell:
Since RabbitMQ 3.5.4.

Issue #268, opened 2015-08-16
gen_server2 can fail to dispatch messages sent by a process to itself


erylee:
If we send several messages to self when a gen_server2 process started and then let `handle_info` return 
```
{norely, State, hibernate}
```
only the first message will be dispatched to callback module,  but others will be queued until next message come into mailbox for the process hibernated...

The gs2issue.erl  will produce this issue:

``` erlang
-module(gs2issue).

-behaviour(gen_server2).

-define(SERVER, ?MODULE).

-export([start_link/0, go/1]).

-export([init/1, handle_call/3, handle_cast/2, handle_info/2,
         terminate/2, code_change/3]).

start_link() ->
    gen_server2:start_link({local, ?SERVER}, ?MODULE, [], []).

go(Pid) ->
    gen_server2:call(Pid, go).

init(Args) ->
    [self() ! X || X <- [a, b, c, d]],
    {ok, Args}.

handle_call(go, _From, State) ->
    {reply, ok, State};

handle_call(_Request, _From, State) ->
    {reply, ok, State}.

handle_cast(_Msg, State) ->
    {noreply, State}.

handle_info(Info, State) ->
    io:format("Got: ~p~n", [Info]),
    {noreply, State, hibernate}.

terminate(_Reason, _State) ->
    ok.

code_change(_OldVsn, State, _Extra) ->
    {ok, State}.
```

The behavihour is different with gen_server.

![gs2issue](https://cloud.githubusercontent.com/assets/11869/9293299/53c744e0-4458-11e5-85c6-ed50cdab0b60.png)

A quick fix for this issue:

```
loop(GS2State = #gs2_state { time          = hibernate,
                             timeout_state = undefined,
                             queue         = Queue }) ->
    case priority_queue:is_empty(Queue) of
        true ->
            pre_hibernate(GS2State);
        false ->
            process_next_msg(GS2State)
    end;

loop(GS2State) ->
    process_next_msg(drain(GS2State)).
```

michaelklishin:
@erylee you can submit a pull request with your change against the `stable` branch. It's not very clear what your proposed change is: having a diff would be better.

erylee:
@michaelklishin, I am forking now, will submit PR minutes later.

videlalvaro:
@erylee the original bug report mentioned version 3.5.4. Can you confirm it only happens since that version of RabbitMQ?

erylee:
[...]

PR #269, opened 2015-08-16
issue #268 - gen_server2 can fail to dispatch messages sent by a process to itself

erylee:
It seems the gen_server2 will hibernate when there are still messages not dispatched in queue. Should we check if the queue is empty before gen_server2 hibernated?


videlalvaro:
Thanks!

PR #274, opened 2015-08-19
Changing sed -E to -r to support older 'sed' versions.

nishan:
Older versions of sed (at-least those earlier than 4.2, I found) does not support the silent "-E" option.  "-r' is perhaps safer option here.

This is tested on CentOS 5.11 and CentOS 6.5.  Please let me know if more testing is required.


nishan:
Closing this. I should check if the target system is Linux based one before using "-r".  Please don't merge this change.

PR #275, opened 2015-08-19
Changing sed -E to -r to support older 'sed' versions.

nishan:
On linux system, use "-r" option for sed (for extended regular expressions). On sed versions earlier than 4.2, the silent "-E" option is not present.


michaelklishin:
@nishan thank you. The `-E` option isn't even in the `sed` man on Linux, although two `4.2.x` version I have at hand do accept it.

Can you please re-submit your PR against the `stable` branch?

michaelklishin:
@dumbbell FYI.

michaelklishin:
Closing because we want to merge this into the stable branch, which then will be merged into master.

PR #276, opened 2015-08-19
On Linux, use "-r" option for sed, since earlier versions of 'sed' does not support "-E" option.

nishan:
rabbitmq-wrapper-script uses "-E" option for sed which is not present in sed versions earlier that 4.2.

On machines with these old sed versions the script broke.  I found atleast CentOS 5.11 where the script broke.


michaelklishin:
Sorry, this is still *submitted* against master.

PR #277, opened 2015-08-19
On Linux, use -r option for 'sed'. -E is not present in versions earlier versions of 'sed'

nishan:
rabbitmq-wrapper-script uses "-E" option for sed which is not present in sed versions earlier that 4.2.

On machines with these old sed versions the script broke. I found atleast CentOS 5.11 where the script broke.

### DIFF
--- a/packaging/common/rabbitmq-script-wrapper
+++ b/packaging/common/rabbitmq-script-wrapper
@@ -16,9 +16,14 @@
 ##
 
 # Escape spaces and quotes, because shell is revolting.
+SED_OPT="-E"
+if [ $(uname -s) == "Linux" ]; then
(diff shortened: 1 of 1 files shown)


michaelklishin:
I'm testing a branch with cherry picked commits. Thank you.

videlalvaro:
Is this a Linux only breakage or we might see this in other platforms as well, say Solaris and so on?

michaelklishin:
I couldn't find what sed version Solaris 10 has.

michaelklishin:
@videlalvaro so since this is a shared packaging script, Solaris is not of concern here. Modern OS X versions have `sed` which supports `-E`.

Issue #227, opened 2015-07-21
Lengthy throughput drop when flushing queue index journal


videlalvaro:
See this thread for more details: https://groups.google.com/forum/#!searchin/rabbitmq-users/raymond/rabbitmq-users/HHrH4I2bVjs/N7IoYB9G4oMJ

videlalvaro:
What needs to be optimized is the `rabbit_queue_index:publish/6` case where `rabbit_queue_index:maybe_flush_journal/2` is called at the end.

The problem here is: we have a use case where the consumers have been stopped for whatever the reason, and messages start accumulating in the queue, with lots of `Journal` entries in the index. At some point the Journal will be flushed to disk, and there performance will drop terribly. We are talking of going from 20k 1kb msgs/sec down to a few hundreds per second.

Yes, we can say: don't publish faster than you can consume, and that's fine, but on the few tests I ran on my mac with SSDs and 16gb of RAM, after two minutes or so, the broker performance drops significantly, because the Journals must be flushed. I think this has to be optimized. Yes, we can't escape flushing stuff that's in memory to disk, but we have to find a strategy that doesn't do this so eagerly and expects the world to stop while doing so.

To test this case I used the following:

`./runjava.sh com.rabbitmq.examples.PerfTest -x 2 -y 0 -u bench -s 1000 -z 360`

As you can see the message size is 1kb (depending who you ask) so messages are embedded in the queue index. See  `queue_index_embed_msgs_below`.

It might be worth investigating the relation between the ingres rate and the `queue_index_max_journal_entries` config setting.

Also be aware that many of these Journal entries/Segments end up stored in the process dictionary when they are passed down to the FHC.

gmr:
I'm thinking this might be even more significant the larger the amount of data. In my production issue today my consumers could not keep up with the publishing velocity. Messages are generally >50KB, but can be larger. Consumers would only be delivered at a rate that was roughly the disk read rate, *but* there was no IOWait on any node in the cluster. So where I might be able to go @1k per sec, when it was ~20GB in size on disk, it was in the 10s per second. When it was ~1GB on disk, it started to go back to normal speeds

Another way to state it was the disk read rate was disproportionately slow, the larger the transient message store became, without any indicators in the OS as to why (other than RAM usage?).

![Big data, slow performance](http://i.imgur.com/z50UcxV.png)

videlalvaro:
[...]

PR #281, opened 2015-08-24
Reduce default QI journal size

### LINKED ISSUE #279: Reduce default QI (queue index) journal size
See #227 for background.

Profiling suggests that the bottleneck in #227 is sequential folding over QI journal. While we will look into parallelising it, the need for reducing I/O by having a larger journal seems a lot less relevant in 2015 than it was ~ 5 years ago.

Benchmarks and monitoring results from multiple people suggests values such as `16K` or `8K` result in much more even throughput.

We've agreed to reduce default journal size from `65536` to `8192`.

michaelklishin:
After some consideration I decided to not make `SEGMENT_ENTRY_COUNT` configurable. It turned out to be hairier than expected and I would rather not have such changes in `3.5.x`. As for `3.6.x`, there are more important issues to spend time on right now.

Fixes #279, see #227 for background info.

### DIFF
--- a/src/rabbit_queue_index.erl
+++ b/src/rabbit_queue_index.erl
@@ -127,7 +127,7 @@
 %% binary generation/matching with constant vs variable lengths.
 
 -define(REL_SEQ_BITS, 14).
--define(SEGMENT_ENTRY_COUNT, 16384). %% trunc(math:pow(2,?REL_SEQ_BITS))).
+-define(SEGMENT_ENTRY_COUNT, 2048).
 
 %% seq only is binary 01 followed by 14 bits of rel seq id
 %% (range: 0 - 16383)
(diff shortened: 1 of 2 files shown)


src/rabbit_queue_index.erl
```diff
@@ -127,7 +127,7 @@
 %% binary generation/matching with constant vs variable lengths.
 
 -define(REL_SEQ_BITS, 14).
--define(SEGMENT_ENTRY_COUNT, 16384). %% trunc(math:pow(2,?REL_SEQ_BITS))).
```
videlalvaro: I don't think we need to remove this comment

michaelklishin:
@videlalvaro comment restore, I think this is ready.

PR #283, opened 2015-08-24
Converts journal entries in place.

### LINKED ISSUE #227: Lengthy throughput drop when flushing queue index journal

videlalvaro:
Instead of folding over the whole journal when we flush it to disk, it
might be better to convert the journal entries to their final disk
format, to avoid folding over 65k journal entries. Of course this
approach uses more memory.

Fixes #227

### DIFF
--- a/src/rabbit_queue_index.erl
+++ b/src/rabbit_queue_index.erl
@@ -178,7 +178,8 @@
                   max_journal_entries, on_sync, on_sync_msg,
                   unconfirmed, unconfirmed_msg}).
 
--record(segment, {num, path, journal_entries, unacked}).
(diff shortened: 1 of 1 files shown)


videlalvaro:
Please thoroughly review this PR since bugs in it might result in message loss/message duplication/missing acks and so on

videlalvaro:
I see is not possible to have two people assigned to the ticket, but @dumbbell if you can please give it a review since changes here are quite critical

PR #286, opened 2015-08-25
Makes CREDIT_DISC_BOUND and IO_BATCH_SIZE configurable

### LINKED ISSUE #284: make IO_BATCH_SIZE and CREDIT_DISC_BOUND tunable by the user
See #227 and #279 for a reference discussion

videlalvaro:
Fixes #284

### DIFF
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -30,6 +30,9 @@
 %% Boot steps.
 -export([maybe_insert_default_data/0, boot_delegate/0, recover/0]).
 
+%% for tests
+-export([validate_msg_store_io_batch_size_and_credit_disc_bound/2]).
+
 -rabbit_boot_step({pre_boot, [{description, "rabbit boot start"}]}).
 
 -rabbit_boot_step({codec_correctness_check,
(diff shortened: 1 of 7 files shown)


src/rabbit.erl
```diff
@@ -848,6 +852,82 @@ warn_if_kernel_config_dubious() ->
         true  -> ok
     end.
 
+warn_if_disc_io_options_dubious() ->
+    %% if these values are not set, it doesn't matter since
+    %% rabbit_variable_queue will pick up the values defined in the
+    %% IO_BATCH_SIZE and CREDIT_DISC_BOUND constants.
+    CreditDiscBound = rabbit_misc:get_env(rabbit, credit_disc_bound, undefined),
```
michaelklishin: Please prefix these with `msg_store_`, e.g. we already have `msg_store_file_size_limit`, `queue_index_embed_msgs_below`, and `queue_index_max_journal_entries`.

src/rabbit.erl
```diff
@@ -30,6 +30,9 @@
 %% Boot steps.
 -export([maybe_insert_default_data/0, boot_delegate/0, recover/0]).
 
+%% for tests
+-export([validate_io_options/2]).
```
michaelklishin: This is a very generic name but we actually validate only 2 settings. I think this needs a more specific name.

michaelklishin:
@videlalvaro except for the two naming remarks, this LGTM.

michaelklishin:
Renaming is done. @videlalvaro please take a look.

PR #292, opened 2015-09-01
fetches IndexMaxSize at BQ init time

### LINKED ISSUE #291: store queue_index_embed_msgs_below in BQ state so it's not read from the env for every publish
For every message that enters the broker, we have to determine if the message will be persisted to the msg_store or to the queue_index. To do that we need to determine what's the configured IndexMaxSize value. At the moment this value is fetch from the application environment for every publish.

See: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_variable_queue.erl#L1473

videlalvaro:
Fixes #291

### DIFF
--- a/src/rabbit_variable_queue.erl
+++ b/src/rabbit_variable_queue.erl
@@ -273,6 +273,7 @@
           msg_store_clients,
           durable,
           transient_threshold,
+          qi_embed_msgs_below,
 
           len,                %% w/o unacked
           bytes,              %% w/o unacked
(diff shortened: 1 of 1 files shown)


src/rabbit_variable_queue.erl
```diff
@@ -273,6 +273,7 @@
           msg_store_clients,
           durable,
           transient_threshold,
+          index_max_size,
```
michaelklishin: I find this name confusing. What about `queue_index_embedding_threshold ` or even `queue_index_embed_msgs_below `?

videlalvaro: I'll change it to `queue_index_embed_msgs_below`

PR #300, opened 2015-09-07
improvements for BQ:purge

### LINKED ISSUE #295: improve queue purge performance
There's a couple of TODOs related to queue purge performance improvements: that we should address on this issue: 

https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_variable_queue.erl#L554

videlalvaro:
Fixes #295

### DIFF
--- a/src/rabbit_queue_index.erl
+++ b/src/rabbit_queue_index.erl
@@ -220,6 +220,7 @@
 -type(shutdown_terms() :: [term()] | 'non_clean_shutdown').
 
 -spec(erase/1 :: (rabbit_amqqueue:name()) -> 'ok').
+-spec(reset_state/1 :: (qistate()) -> qistate()).
 -spec(init/3 :: (rabbit_amqqueue:name(),
                  on_sync_fun(), on_sync_fun()) -> qistate()).
 -spec(recover/6 :: (rabbit_amqqueue:name(), shutdown_terms(), boolean(),
--- a/src/rabbit_queue_index.erl
+++ b/src/rabbit_queue_index.erl
@@ -16,7 +16,7 @@
 
 -module(rabbit_queue_index).
 
--export([erase/1, init/3, recover/6,
+-export([erase/1, init/3, reset_state/1, recover/6,
          terminate/2, delete_and_terminate/1,
          publish/6, deliver/2, ack/2, sync/1, needs_sync/1, flush/1,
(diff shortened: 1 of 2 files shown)


src/rabbit_queue_index.erl
```diff
@@ -257,10 +258,19 @@
 
 erase(Name) ->
     #qistate { dir = Dir } = blank_state(Name),
-    case rabbit_file:is_dir(Dir) of
-        true  -> rabbit_file:recursive_delete([Dir]);
-        false -> ok
-    end.
+    erase_index_dir(Dir).
+
+%% used during variable queue purge when there are no pending acks
```
videlalvaro: we need a _usable_ index state, since after BQ:purge/1 the index state might keep being used by future publishes and so on.

src/rabbit_variable_queue.erl
```diff
@@ -1509,11 +1560,29 @@ remove_pending_ack(false, SeqId, State = #vqstate{ram_pending_ack  = RPA,
     end.
 
 purge_pending_ack(KeepPersistent,
```
videlalvaro: This function got split in 2. The part that collects pending acks, and the part that handles the removal of those pending acks/msgs depending on the `KeepPersistent` flag.

src/rabbit_variable_queue.erl
```diff
@@ -1509,11 +1560,29 @@ remove_pending_ack(false, SeqId, State = #vqstate{ram_pending_ack  = RPA,
...
+    {IndexOnDiskSeqIds, MsgIdsByStore, State1} = purge_pending_ack1(State),
+    case KeepPersistent of
+        true  -> remove_transient_msgs_by_id(MsgIdsByStore, MSCState),
+                 State1;
+        false -> IndexState1 =
+                     rabbit_queue_index:ack(IndexOnDiskSeqIds, IndexState),
+                 remove_msgs_by_id(MsgIdsByStore, MSCState),
+                 State1 #vqstate { index_state = IndexState1 }
+    end.
+
```
videlalvaro: `purge_pending_ack_delete_and_terminate/1` shoud work like the old `purge_pending_ack/2` when `KeepPersistent = false`, but instead of acking/delivering messages to the index, we just call `rabbit_queue_index:delete_and_terminate/1`

PR #303, opened 2015-09-09
Make rabbitmqctl list_* operations display things dynamically, fixes #62

### LINKED ISSUE #62: Make rabbitmqctl operations support a timeout

Ayanda-D:
Changes made for #62, improving on the timeout introduced in #181 for `rabbitmqctl` list_* operations. Information is now returned dynamically and in piecemeal from the rabbit nodes, as things happen, and terminates upon timeout.

### DIFF
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -553,6 +553,7 @@ list_down(VHostPath) ->
 info_keys() -> rabbit_amqqueue_process:info_keys().
 
 map(Qs, F) -> rabbit_misc:filter_exit_map(F, Qs).
+map_fun() -> fun(F, Qs) -> rabbit_misc:filter_exit_map(F, Qs) end.
 
 info(Q = #amqqueue{ state = crashed }) -> info_down(Q, crashed);
 info(#amqqueue{ pid = QPid }) -> delegate:call(QPid, info).
@@ -593,13 +594,13 @@ info_all(VHostPath, Items) ->
         map(list_down(VHostPath), fun (Q) -> info_down(Q, Items, down) end).
 
 info_all(VHostPath, Items, Ref, AggregatorPid) ->
-    rabbit_control_main:emitting_map(
-      AggregatorPid, Ref, fun(Q) -> info(Q, Items) end, list(VHostPath),
+    rabbit_control_main:emitting_map_with_wrapper_fun(
+      AggregatorPid, Ref, fun(Q) -> info(Q, Items) end, map_fun(), list(VHostPath),
       continue),
-    rabbit_control_main:emitting_map(
+    rabbit_control_main:emitting_map_with_wrapper_fun(
       AggregatorPid, Ref,
-      fun(Q) -> info_down(Q, Items) end, list_down(VHostPath)).
-    
+      fun(Q) -> info_down(Q, Items) end, map_fun(), list_down(VHostPath)).
+
 force_event_refresh(Ref) ->
     [gen_server2:cast(Q#amqqueue.pid,
                       {force_event_refresh, Ref}) || Q <- list()],
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -120,7 +120,7 @@
 -spec(info_all/2 :: (rabbit_types:vhost(), rabbit_types:info_keys())
                     -> [rabbit_types:infos()]).
 -spec(info_all/4 :: (rabbit_types:vhost(), rabbit_types:info_keys(),
-                     reference()) -> 'ok').
+                     reference(), pid()) -> 'ok').
 -spec(force_event_refresh/1 :: (reference()) -> 'ok').
 -spec(notify_policy_changed/1 :: (rabbit_types:amqqueue()) -> 'ok').
 -spec(consumers/1 :: (rabbit_types:amqqueue())
@@ -133,8 +133,7 @@
              non_neg_integer(), rabbit_framing:amqp_table()}]).
 -spec(consumers_all/3 ::
         (rabbit_types:vhost(), reference(), pid())
-        -> [{reference(), {name(), pid(), rabbit_types:ctag(), boolean(),
-             non_neg_integer(), rabbit_framing:amqp_table()}}]).
+        -> 'ok').
 -spec(stat/1 ::
         (rabbit_types:amqqueue())
         -> {'ok', non_neg_integer(), non_neg_integer()}).
@@ -633,7 +632,9 @@ consumers_all(VHostPath, Ref, Pid) ->
                                   AckRequired, Prefetch, Args]) ||
                                   {ChPid, CTag, AckRequired, Prefetch, Args}
                                       <- consumers(Q)]}
-          end)).
+          end)),
+    Pid ! {Ref, finished},
+    ok.
 
 stat(#amqqueue{pid = QPid}) -> delegate:call(QPid, stat).
 
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -597,7 +597,8 @@ info_all(VHostPath, Items, Ref, AggregatorPid) ->
       AggregatorPid, Ref, fun(Q) -> info(Q, Items) end, list(VHostPath),
       continue),
     rabbit_control_main:emitting_map(
-      AggregatorPid, Ref, fun(Q) -> info_down(Q, Items) end, list(VHostPath)).
+      AggregatorPid, Ref,
+      fun(Q) -> info_down(Q, Items) end, list_down(VHostPath)).
     
 force_event_refresh(Ref) ->
     [gen_server2:cast(Q#amqqueue.pid,
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -622,8 +622,7 @@ consumers_all(VHostPath, Ref, AggregatorPid) ->
     rabbit_control_main:emitting_map(
       AggregatorPid, Ref,
       fun(Q) -> get_queue_consumer_info(Q, ConsumerInfoKeys) end,
-      list(VHostPath)),
-    ok.
+      list(VHostPath)).
 
 get_queue_consumer_info(Q, ConsumerInfoKeys) ->
     lists:flatten(
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -593,11 +593,11 @@ info_all(VHostPath, Items) ->
         map(list_down(VHostPath), fun (Q) -> info_down(Q, Items, down) end).
 
 info_all(VHostPath, Items, Ref, AggregatorPid) ->
-    map(list(VHostPath), fun (Q) -> AggregatorPid ! {Ref, info(Q, Items)} end) ++
-        map(list_down(VHostPath),
-            fun (Q) -> AggregatorPid ! {Ref, info_down(Q, Items, down)} end),
-    AggregatorPid  ! {Ref, finished},
-    ok.
+    rabbit_control_main:emitting_map(
+      AggregatorPid, Ref, fun(Q) -> info(Q, Items) end, list(VHostPath),
+      continue),
+    rabbit_control_main:emitting_map(
+      AggregatorPid, Ref, fun(Q) -> info_down(Q, Items) end, list(VHostPath)).
     
 force_event_refresh(Ref) ->
     [gen_server2:cast(Q#amqqueue.pid,
@@ -615,27 +615,22 @@ consumers_all(VHostPath) ->
     ConsumerInfoKeys = consumer_info_keys(),
     lists:append(
       map(list(VHostPath),
-          fun (Q) ->
-              [lists:zip(
-                 ConsumerInfoKeys,
-                 [Q#amqqueue.name, ChPid, CTag, AckRequired, Prefetch, Args]) ||
-                  {ChPid, CTag, AckRequired, Prefetch, Args} <- consumers(Q)]
-          end)).
+          fun(Q) -> get_queue_consumer_info(Q, ConsumerInfoKeys) end)).
 
 consumers_all(VHostPath, Ref, AggregatorPid) ->
     ConsumerInfoKeys = consumer_info_keys(),
-    map(list(VHostPath),
-        fun (Q) -> AggregatorPid !
-                       {Ref, get_queue_consumer_info(Q, ConsumerInfoKeys)}
-        end),
-    AggregatorPid ! {Ref, finished},
+    rabbit_control_main:emitting_map(
+      AggregatorPid, Ref,
+      fun(Q) -> get_queue_consumer_info(Q, ConsumerInfoKeys) end,
+      list(VHostPath)),
     ok.
 
 get_queue_consumer_info(Q, ConsumerInfoKeys) ->
-    lists:flatten([lists:zip(ConsumerInfoKeys,
-                             [Q#amqqueue.name, ChPid, CTag,
-                              AckRequired, Prefetch, Args]) ||
-                      {ChPid, CTag, AckRequired, Prefetch, Args} <- consumers(Q)]).
+    lists:flatten(
+      [lists:zip(ConsumerInfoKeys,
+                 [Q#amqqueue.name, ChPid, CTag,
+                  AckRequired, Prefetch, Args]) ||
+          {ChPid, CTag, AckRequired, Prefetch, Args} <- consumers(Q)]).
 
 stat(#amqqueue{pid = QPid}) -> delegate:call(QPid, stat).
 
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -612,7 +612,7 @@ consumers(#amqqueue{ pid = QPid }) -> delegate:call(QPid, consumers).
 consumer_info_keys() -> ?CONSUMER_INFO_KEYS.
 
 consumers_all(VHostPath) ->
-    ConsumerInfoKeys=consumer_info_keys(),
+    ConsumerInfoKeys = consumer_info_keys(),
     lists:append(
       map(list(VHostPath),
           fun (Q) ->
@@ -623,19 +623,19 @@ consumers_all(VHostPath) ->
           end)).
 
 consumers_all(VHostPath, Ref, AggregatorPid) ->
-    ConsumerInfoKeys=consumer_info_keys(),
+    ConsumerInfoKeys = consumer_info_keys(),
     map(list(VHostPath),
-        fun (Q) ->
-                AggregatorPid !
-                    {Ref, [lists:zip(
-                             ConsumerInfoKeys,
-                             [Q#amqqueue.name, ChPid, CTag,
-                              AckRequired, Prefetch, Args]) ||
-                              {ChPid, CTag, AckRequired, Prefetch, Args}
-                                  <- consumers(Q)]}
+        fun (Q) -> AggregatorPid !
+                       {Ref, get_queue_consumer_info(Q, ConsumerInfoKeys)}
         end),
     AggregatorPid ! {Ref, finished},
     ok.
+
+get_queue_consumer_info(Q, ConsumerInfoKeys) ->
+    lists:flatten([lists:zip(ConsumerInfoKeys,
+                             [Q#amqqueue.name, ChPid, CTag,
+                              AckRequired, Prefetch, Args]) ||
+                      {ChPid, CTag, AckRequired, Prefetch, Args} <- consumers(Q)]).
 
 stat(#amqqueue{pid = QPid}) -> delegate:call(QPid, stat).
 
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -624,17 +624,16 @@ consumers_all(VHostPath) ->
 
 consumers_all(VHostPath, Ref, AggregatorPid) ->
     ConsumerInfoKeys=consumer_info_keys(),
-    lists:append(
-      map(list(VHostPath),
-          fun (Q) ->
-                  AggregatorPid !
-                      {Ref, [lists:zip(
-                               ConsumerInfoKeys,
-                               [Q#amqqueue.name, ChPid, CTag,
-                                AckRequired, Prefetch, Args]) ||
-                                {ChPid, CTag, AckRequired, Prefetch, Args}
-                                    <- consumers(Q)]}
-          end)),
+    map(list(VHostPath),
+        fun (Q) ->
+                AggregatorPid !
+                    {Ref, [lists:zip(
+                             ConsumerInfoKeys,
+                             [Q#amqqueue.name, ChPid, CTag,
+                              AckRequired, Prefetch, Args]) ||
+                              {ChPid, CTag, AckRequired, Prefetch, Args}
+                                  <- consumers(Q)]}
+        end),
     AggregatorPid ! {Ref, finished},
     ok.
 
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -24,10 +24,11 @@
          assert_equivalence/5,
          check_exclusive_access/2, with_exclusive_access_or_die/3,
          stat/1, deliver/2, requeue/3, ack/3, reject/4]).
--export([list/0, list/1, info_keys/0, info/1, info/2, info_all/1, info_all/2]).
+-export([list/0, list/1, info_keys/0, info/1, info/2, info_all/1, info_all/2,
+         info_all/4]).
 -export([list_down/1]).
 -export([force_event_refresh/1, notify_policy_changed/1]).
--export([consumers/1, consumers_all/1, consumer_info_keys/0]).
+-export([consumers/1, consumers_all/1,  consumers_all/3, consumer_info_keys/0]).
 -export([basic_get/4, basic_consume/10, basic_cancel/4, notify_decorators/1]).
 -export([notify_sent/2, notify_sent_queue_down/1, resume/2]).
 -export([notify_down_all/2, activate_limit_all/2, credit/5]).
@@ -118,6 +119,8 @@
 -spec(info_all/1 :: (rabbit_types:vhost()) -> [rabbit_types:infos()]).
 -spec(info_all/2 :: (rabbit_types:vhost(), rabbit_types:info_keys())
                     -> [rabbit_types:infos()]).
+-spec(info_all/4 :: (rabbit_types:vhost(), rabbit_types:info_keys(),
+                     reference()) -> 'ok').
 -spec(force_event_refresh/1 :: (reference()) -> 'ok').
 -spec(notify_policy_changed/1 :: (rabbit_types:amqqueue()) -> 'ok').
 -spec(consumers/1 :: (rabbit_types:amqqueue())
@@ -128,6 +131,10 @@
         (rabbit_types:vhost())
         -> [{name(), pid(), rabbit_types:ctag(), boolean(),
              non_neg_integer(), rabbit_framing:amqp_table()}]).
+-spec(consumers_all/3 ::
+        (rabbit_types:vhost(), reference(), pid())
+        -> [{reference(), {name(), pid(), rabbit_types:ctag(), boolean(),
+             non_neg_integer(), rabbit_framing:amqp_table()}}]).
 -spec(stat/1 ::
         (rabbit_types:amqqueue())
         -> {'ok', non_neg_integer(), non_neg_integer()}).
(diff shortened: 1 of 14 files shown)


michaelklishin:
@Ayanda-D thank you! Is this ready for QA or should we wait?

Ayanda-D:
@michaelklishin QA can wait for now, but you can review the updates and give feedback. I'll ping you when ready for QA. Thanks.

src/rabbit_amqqueue.erl
```diff
@@ -609,6 +621,21 @@ consumers_all(VHostPath) ->
                   {ChPid, CTag, AckRequired, Prefetch, Args} <- consumers(Q)]
           end)).
 
+consumers_all(VHostPath, Ref, Pid) ->
```
michaelklishin: I'd rename `Pid` to something that reflects that the pid/process is a result aggregator. `AggregatorPid`, maybe?

Ayanda-D: Okay noted. `AggregatorPid` sounds fine, & appropriate.

src/rabbit_control_main.erl
```diff
@@ -679,6 +693,29 @@ default_if_empty(List, Default) when is_list(List) ->
        true       -> [list_to_atom(X) || X <- List]
     end.
 
+wait_info_messages(Pid, Ref, ArgAtoms, Timeout) ->
```
michaelklishin: Should this be `wait_for_info_messages`? We don't mind longer function names.

Ayanda-D: Okay noted, I'll update & bear that in mind. And yes, purpose is to wait for info massages.

Ayanda-D:
@michaelklishin this is ready for QA. Note, only successful timeout tests have been implemented here.

michaelklishin:
@Ayanda-D thank you!

src/rabbit_amqqueue.erl
```diff
@@ -609,6 +622,21 @@ consumers_all(VHostPath) ->
                   {ChPid, CTag, AckRequired, Prefetch, Args} <- consumers(Q)]
           end)).
 
+consumers_all(VHostPath, Ref, AggregatorPid) ->
+    ConsumerInfoKeys=consumer_info_keys(),
+    map(list(VHostPath),
+        fun (Q) ->
+                AggregatorPid !
+                    {Ref, [lists:zip(
```
videlalvaro: can this be extracted into a function to clarify a bit code intention. After looking at it I do understand that you are matching ConsumerInfoKeys with their values, but still, I can't get that info at first sight without stoping at inspecting the code.

src/rabbit_amqqueue.erl
```diff
@@ -609,6 +622,21 @@ consumers_all(VHostPath) ->
                   {ChPid, CTag, AckRequired, Prefetch, Args} <- consumers(Q)]
           end)).
 
+consumers_all(VHostPath, Ref, AggregatorPid) ->
+    ConsumerInfoKeys=consumer_info_keys(),
```
videlalvaro: don't forget to keep spaces between variables and the `=` sign

src/rabbit_cli.erl
```diff
@@ -223,3 +229,11 @@ rpc_call(Node, Mod, Fun, Args, Timeout) ->
         Time            -> net_kernel:set_net_ticktime(Time, 0),
                            rpc:call(Node, Mod, Fun, Args, Timeout)
     end.
+
+rpc_call(Node, Mod, Fun, Args, Ref, Pid, Timeout) ->
+    case rpc:call(Node, net_kernel, get_net_ticktime, [], Timeout) of
+        {badrpc, _} = E -> E;
+        Time            -> net_kernel:set_net_ticktime(Time, 0),
```
videlalvaro: wouldn't this call here affect other nodes net_ticktime?

src/rabbit_control_main.erl
```diff
@@ -499,62 +499,71 @@ action(purge_queue, Node, [Q], Opts, Inform, Timeout) ->
...
 
 action(list_parameters, Node, [], Opts, Inform, Timeout) ->
     VHostArg = list_to_binary(proplists:get_value(?VHOST_OPT, Opts)),
     Inform("Listing runtime parameters", []),
-    display_info_list(
-      rpc_call(Node, rabbit_runtime_parameters, list_formatted, [VHostArg],
-               Timeout),
-      rabbit_runtime_parameters:info_keys());
+    spawn_link(rabbit_cli, rpc_call, [Node, rabbit_runtime_parameters,
+                                      list_formatted, [VHostArg], Ref=make_ref(),
```
videlalvaro: Please fix variable and `=` spacing

src/rabbit_control_main.erl
```diff
@@ -17,12 +17,12 @@
 -module(rabbit_control_main).
 -include("rabbit.hrl").
 -include("rabbit_cli.hrl").
-
+-compile(export_all).
```
videlalvaro: please remove this line

src/rabbit_control_main.erl
```diff
@@ -499,62 +499,71 @@ action(purge_queue, Node, [Q], Opts, Inform, Timeout) ->
...
+                                      Pid=self(), Timeout]),
+    wait_for_info_messages(Pid, Ref, rabbit_runtime_parameters:info_keys(), Timeout);
 
 action(list_policies, Node, [], Opts, Inform, Timeout) ->
     VHostArg = list_to_binary(proplists:get_value(?VHOST_OPT, Opts)),
     Inform("Listing policies", []),
-    display_info_list(rpc_call(Node, rabbit_policy, list_formatted, [VHostArg],
-                              Timeout),
-                      rabbit_policy:info_keys());
+    spawn_link(rabbit_cli, rpc_call, [Node, rabbit_policy, list_formatted,
```
videlalvaro: is there a reason why we have all these spawn_link calls, and also the new `call/4` down below? Is `call/4` meant to be used everywhere? 

Also there's seems to be a pattern here of `spawn_link` then `wait_for_info_messages`. Perhaps this can be refactored?

src/rabbit_policy.erl
```diff
@@ -170,9 +170,18 @@ list(VHost) ->
 list_formatted(VHost) ->
     order_policies(list0(VHost, fun format/1)).
 
+list_formatted(VHost, Ref, AggregatorPid) ->
+    list0(VHost, fun format/1, Ref, AggregatorPid),
```
videlalvaro: is there a reason why here there's a specialized function for the mapping, while in other places the mapping is done inline?

test/src/rabbit_tests.erl
```diff
@@ -1239,6 +1241,102 @@ test_server_status() ->
 
     passed.
 
+test_list_operations_timeout_pass() ->
```
videlalvaro: can you move the tests to their own file?

src/rabbit_cli.erl
Ayanda-D: @videlalvaro here the `net_ticktime` of the temporary node started when `rabbitmqctl` is executed is being set to match that of the remote node. This is the same logic of the current implementation.

src/rabbit_control_main.erl
Ayanda-D: @videlalvaro there are some `list_*` calls which need only spawn `rabbit_cli:rpc_call` directly, and some which first need to apply `list_to_binary_utf8/1` to their arguments before spawning `rabbit_cli:rpc_call`, which is the purpose of `call/4`. This serves same purpose with `call/3` non-spawned `rabbit_cli:rpc_call`s. So not all `list_*` operations make use of of `call/4`.

Ayanda-D: @videlalvaro for your second question, refactor how exactly, into one function? I kept these two separate based on their unrelatedness, and just to make things a bit more readable.

src/rabbit_cli.erl
videlalvaro: Got it. Can you still refactor the rpc_calls to not have duplicated code?

src/rabbit_control_main.erl
videlalvaro: For me it's kinda weird that everywhere there's the spawn_link, then wait_for_info_messages. This seems like a pattern that could be extracted into its own function. Also `spawn_link` always gets `rabbit_cli, rpc_call`, with the only difference being the arguments. `wait_for_info_messages` seems to always get the same arguments, except the third one, so I would expect a function that takes 2 arguments and abstracts all this `spawn_link/wait_for_info_messages` away.

src/rabbit_cli.erl
```diff
@@ -106,6 +109,9 @@ main(ParseFun, DoFun, UsageMod) ->
             print_error("unable to connect to nodes ~p: ~w", [Nodes, Reason]),
             print_badrpc_diagnostics(Nodes),
             rabbit_misc:quit(2);
+        function_clause ->
+            print_error("invalid parameter: ~p", [Args]),
```
videlalvaro: can you also print here the command that was being executed?

src/rabbit_control_main.erl
videlalvaro: About the need to call `list_to_binary_utf8/1` in some cases, and in some cases not, can this be refactored as well? There seem to be only 3 cases where this special call that handles `list_to_binary_utf8/1` is required, so I don't think it would be difficult to use `call/4` in a way that handles both cases.

michaelklishin:
@Ayanda-D hi, any feedback on @videlalvaro's questions?

src/rabbit_control_main.erl
Ayanda-D: @videlalvaro thanks for the comments, and cool, I'll move `spawn_link/wait_for_info_messages` to the `call/4` function. I'm busy working on this.

michaelklishin:
@Ayanda-D when you think this is ready, please let us know with a comment here.

Ayanda-D:
@michaelklishin this ready.

Ayanda-D:
@michaelklishin this is now ready.

michaelklishin:
@videlalvaro this is ready for another QA round.

videlalvaro:
Yup. I was just handling other fires :)

test/src/timeout_tests.erl
```diff
@@ -0,0 +1,131 @@
...
+
+    %% list bindings
+    ok = info_action(list_bindings, rabbit_binding:info_keys(), true,
+                     ?TIMEOUT_LIST_OPS_PASS),
+
+    %% list connections
+    {H, P} = find_listener(),
+    {ok, C} = gen_tcp:connect(H, P, []),
+    gen_tcp:send(C, <<"AMQP", 0, 0, 9, 1>>),
+    timer:sleep(100),
```
videlalvaro: what's the point of this timer:sleep here?

src/rabbit_amqqueue.erl
```diff
@@ -609,6 +622,21 @@ consumers_all(VHostPath) ->
...
+    ConsumerInfoKeys = consumer_info_keys(),
+    map(list(VHostPath),
+        fun (Q) -> AggregatorPid !
+                       {Ref, get_queue_consumer_info(Q, ConsumerInfoKeys)}
+        end),
+    AggregatorPid ! {Ref, finished},
+    ok.
+
+get_queue_consumer_info(Q, ConsumerInfoKeys) ->
+    lists:flatten([lists:zip(ConsumerInfoKeys,
```
videlalvaro: Why is this code duplicating the code from here? https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-62/src/rabbit_amqqueue.erl#L619

src/rabbit_amqqueue.erl
```diff
@@ -586,6 +592,13 @@ info_all(VHostPath, Items) ->
     map(list(VHostPath), fun (Q) -> info(Q, Items) end) ++
         map(list_down(VHostPath), fun (Q) -> info_down(Q, Items, down) end).
 
+info_all(VHostPath, Items, Ref, AggregatorPid) ->
+    map(list(VHostPath), fun (Q) -> AggregatorPid ! {Ref, info(Q, Items)} end) ++
```
videlalvaro: Compiler warning:

src/rabbit_amqqueue.erl:596: Warning: the result of the expression is ignored (suppress the warning by assigning the expression to the _ variable)

src/rabbit_control_main.erl
```diff
@@ -748,12 +757,24 @@ ensure_app_running(Node) ->
         Other -> Other
     end.
 
+notify_if_timeout(Pid, Ref, Timeout) ->
+    timer:send_after(Timeout, Pid, {Ref, {timeout, Timeout}}).
+
 call(Node, {Mod, Fun, Args}) ->
     rpc_call(Node, Mod, Fun, lists:map(fun list_to_binary_utf8/1, Args)).
 
 call(Node, {Mod, Fun, Args}, Timeout) ->
```
videlalvaro: Compiler warning:

src/rabbit_control_main.erl:766: Warning: function call/3 is unused

src/rabbit_auth_backend_internal.erl
```diff
@@ -321,6 +334,23 @@ list_permissions(Keys, QueryThunk) ->
             %% TODO: use dirty ops instead
             rabbit_misc:execute_mnesia_transaction(QueryThunk)].
 
+list_permissions(Keys, QueryThunk, Ref, AggregatorPid) ->
+    [AggregatorPid ! {Ref, filter_props(Keys, [{user,      Username},
```
videlalvaro: Before I was doing a code review just looking at the diff here on Github. Inspecting the full code tells me all the 
"async" functions just duplicated code from the module. Is my assessment correct?
https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-62/src/rabbit_auth_backend_internal.erl#L324

videlalvaro:
Apart from the comments about duplicated code, and compiler warnings. The code seems to always do the same:

```erlang
    [AggregatorPid ! {Ref, Msg} ||
        Msg <- Generator],
    AggregatorPid ! {Ref, finished},
    ok.
```

I think this has to be abstracted into a function that takes something like a message builder and a generator builder.

videlalvaro:
It's worth pointing out that this code first retrieves all the items to be listed, and then emits them one by one. So the difference with the current implementation on master is that while mapping, items are emitted, vs, emitting them at the end of the mapping. 

I thought this would be using something like Mnesia cursors and so on.

@michaelklishin @dumbbell what do you think?

test/src/timeout_tests.erl
Ayanda-D: @videlalvaro yes its to buy a bit of time for the connection to be established. However `list_connections` proceeds if 100ms had elapsed without any connection established (empty list message replies to the shell are ignored).

I'll update this to verify connection first, after `timer:sleep(100)`, before it proceeds to executing `list_connections`.

src/rabbit_amqqueue.erl
Ayanda-D: @videlalvaro okay i'll update this. had avoided combining any new changes for this PR with the original backend implementations, but will do.

videlalvaro: Why? Who was going to refactor the duplicated code afterwards?

michaelklishin:
The point of this change is to produce output continuously, so that if an item fails to response in time, at least there is some output. Using a cursor wouldn't change much if we want the timeout to be "global" (for all items), which is what the users would expect. So I find this an improvement.

videlalvaro:
By cursos I meant a Mnesia cursos wherever we access data from mnesia. If there are 1000 queues, this  code will first query Mnesia for the 1000 queues, and then it will start providing output. 

I agree this implementation is an improvement over what we already have.

michaelklishin:
I understand. Again, the goal of this improvement is not to make loading more efficient — even if using a cursor would make it so, which may or may not be the case — but to provide a more sensible output to the user.

Ayanda-D:
@videlalvaro question, for

```erlang
    [AggregatorPid ! {Ref, Msg} ||
        Msg <- Generator],
    AggregatorPid ! {Ref, finished},
    ok.
```

are you assuming `Generator` to have been evaluated into a list? Because this wouldn't work as per initial requirement of the issue. Results wouldn't be sent as they happen, but after everything has been evaluated.

michaelklishin:
Lets not change the scope of this, please. Cursors can be added at a later point. We need to factor out as much duplication as reasonable and given that things work as expected, merge this and move on.

videlalvaro:
@Ayanda-D what @michaelklishin said.

You have a thing that produces a list, whether it's obtained from Mnesia or somewhere, then for each message on that list, you send a message to the `AggregatorPid`. Finally, once the iteration over the list is done, you send a `finished` message.

I would like that to be refactored out. Otherwise the smallest change on the current code, would mean modifying the code everywhere. We want to avoid that.

Ayanda-D:
@videlalvaro okay cool. But just to make sure we're on the same wavelength, messages to `AggregatorPid` are only sent as the list is being produced, not after the list is produced. 

But to add some extra level of abstraction, and your idea of passing a `Generator`, I've just added 2 functions to `rabbit_misc`.  Please confirm if you're okay with this;

for example in
```
map(List, fun(E) -> rabbit_misc:send_loop_result(AggregatorPid, Ref, fun() -> generate(E) end ) 
		  end),
rabbit_misc:send_finished(AggregatorPid, Ref).

& in `rabbit_misc.erl`;
```
send_loop_result( AggregatorPid, Ref, Generator ) ->
	AggregatorPid ! {Ref, Generator()}.
	
send_finished(AggregatorPid, Ref) ->
    AggregatorPid ! {Ref, finished}.
```

**Comment by michaelklishin** [comment]
I'd keep those functions in the `ctl` namespace: they are pretty specific to that module's usage patterns, and `rabbit_misc` is already huge and stuffed with all kinds of functions.

**Comment by Ayanda-D** [comment]
@michaelklishin okay, will keep in `ctl` namespace.

**Comment by videlalvaro** [comment]
Could you rebase your branch with the current master? Since right now it's quite hard to test due to error logger breakage since 18.1

**Comment by videlalvaro** [comment]
@Ayanda-D I wasn't thinking about that. There you just move the sending of messages into a couple of functions, but you are not abstracting the repeated pattern.

The pattern I'm talking about is: "Send a message to a pid, tagging it with a Ref, for every item of a list; once done, send a {Ref, finished} message to said Pid".

Which I think can be implemented as something like:

```erlang
emitting_map(Pid, Ref, Fun, List) ->
    [AggregatorPid ! {Ref, Fun(Item)] ||
        Item <- List],
    Pid ! {Ref, finished},
    ok.
```

Depending on each case, will be what `Fun` is doing with each item that's received from the list.

Then you need to see how to integrate the `rabbit_misc:filter_exit_map` into that emitting map, depending on each circumstance.

Ayanda-D:
@videlalvaro okay cool.

Ayanda-D:
@videlalvaro this is ready for another review and QA.

test/src/timeout_tests.erl
Ayanda-D: @videlalvaro I added a check for this, so it ensures that when `list_connections` is executed, a connection already exists. but what could cause the scenario you've mentioned to occur? i mean both the temp client node & rabbit node are started on the same host whenever the tests are run, and/or are likely to have through a number of amqp connection tests by the time they reach this point

src/rabbit_amqqueue.erl
```diff
@@ -586,6 +592,13 @@ info_all(VHostPath, Items) ->
     map(list(VHostPath), fun (Q) -> info(Q, Items) end) ++
         map(list_down(VHostPath), fun (Q) -> info_down(Q, Items, down) end).
 
+info_all(VHostPath, Items, Ref, AggregatorPid) ->
+    rabbit_control_main:emitting_map(
+      AggregatorPid, Ref, fun(Q) -> info(Q, Items) end, list(VHostPath),
+      continue),
+    rabbit_control_main:emitting_map(
+      AggregatorPid, Ref, fun(Q) -> info_down(Q, Items) end, list(VHostPath)).
```
videlalvaro: shouldn't this be mapping `list_down` instead?

src/rabbit_auth_backend_internal.erl
```diff
@@ -333,17 +340,46 @@ list_user_permissions(Username) ->
...
+                                                configure = ConfigurePerm,
+                                                write     = WritePerm,
+                                                read      = ReadPerm}}) ->
+    filter_props(Keys, [{user,      Username},
+                        {vhost,     VHostPath},
+                        {configure, ConfigurePerm},
+                        {write,     WritePerm},
+                        {read,      ReadPerm}]).
+
+internal_user_filter(#internal_user{username = Username, tags = Tags}) ->
```
videlalvaro: how is this function a filter?

videlalvaro:
With the emitting_map this looks better, but we still have lots of duplicated code, like this for example:

```erlang
list_users() ->
    [internal_user_filter(U) ||
        U <- mnesia:dirty_match_object(rabbit_user, #internal_user{_ = '_'})].

list_users(Ref, AggregatorPid) ->
    rabbit_control_main:emitting_map(
      AggregatorPid, Ref,
      fun(U) -> internal_user_filter(U) end,
      mnesia:dirty_match_object(rabbit_user, #internal_user{_ = '_'})).
```

The only difference here, besides the use of `AggregatorPid, Ref` is the mapping function. This problem happens all over the place on this PR.

michaelklishin:
I don't find structural duplication in

``` erlang
list_users() ->
    [internal_user_filter(U) ||
        U <- mnesia:dirty_match_object(rabbit_user, #internal_user{_ = '_'})].

list_users(Ref, AggregatorPid) ->
    rabbit_control_main:emitting_map(
      AggregatorPid, Ref,
      fun(U) -> internal_user_filter(U) end,
      mnesia:dirty_match_object(rabbit_user, #internal_user{_ = '_'})).
```

to be excessive. Chances are, a version without this duplication would be more difficult to understand, not less.

videlalvaro:
Looking at one file of this code might seem right, looking at the whole PR, I'm not sure.

Just in that case, tomorrow someone needs to change the Mnesia query, it has to be changed all over.

Or this other case, from this same PR:

```erlang
list_permissions(Keys, QueryThunk) ->
    [user_permission_filter(Keys, U) ||
        %% TODO: use dirty ops instead
        U <- rabbit_misc:execute_mnesia_transaction(QueryThunk)].

list_permissions(Keys, QueryThunk, Ref, AggregatorPid) ->
    rabbit_control_main:emitting_map(
      AggregatorPid, Ref, fun(U) -> user_permission_filter(Keys, U) end,
      rabbit_misc:execute_mnesia_transaction(QueryThunk)).
```

The TODO comment is gone on `list_permissions/4`. Is the TODO solved? Was it forgotten during code duplication? What happened?

For me, when I start seeing that I have to change the exact same code, for every file I need to modify on this PR, then I start getting suspicious that there has to be a better way.

michaelklishin:
If this is systemic, then I agree.

videlalvaro:
I understand that Erlang doesn't support curry'ing and so on, so things might not be that easy to abstract away in this case, specially since we have AggregatorPid and Ref all over the place. I would prefer even resorting to an Erlang Macro instead of having code that's duplicate the same pattern all over

src/rabbit_control_main.erl
```diff
@@ -748,11 +761,20 @@ ensure_app_running(Node) ->
...
 
+notify_if_timeout(Pid, Ref, Timeout) ->
+    timer:send_after(Timeout, Pid, {Ref, {timeout, Timeout}}).
+
 call(Node, {Mod, Fun, Args}) ->
     rpc_call(Node, Mod, Fun, lists:map(fun list_to_binary_utf8/1, Args)).
 
-call(Node, {Mod, Fun, Args}, Timeout) ->
-    rpc_call(Node, Mod, Fun, lists:map(fun list_to_binary_utf8/1, Args), Timeout).
+call(Node, {Mod, Fun, Args}, InfoKeys, ToBinUtf8, Timeout) ->
```
videlalvaro: can we have `call/4` and `call/5`? In this way, callers to this function don't need to be seeing this `true/false` parameters all over.

With the code as it is, later, once I had forgotten what `call/5` five arguments are doing, I will have to always look at what that `true/false` flag is.

src/rabbit_auth_backend_internal.erl
Ayanda-D: Okay its extracting. How does `extract_user_permission_params/2`, and `extract_internal_user_params` sound?

videlalvaro: :+1:

test/src/timeout_tests.erl
```diff
@@ -0,0 +1,131 @@
...
+%% The Initial Developer of the Original Code is GoPivotal, Inc.
+%% Copyright (c) 2007-2015 Pivotal Software, Inc.  All rights reserved.
+%%
+-module(timeout_tests).
+
+-compile([export_all]).
+
+-export([all_tests/0]).
+
+-import(rabbit_tests,[default_options/0, control_action/2, control_action/3,
```
videlalvaro: please don't import functions that are not going the be used later: `default_options/0` and `expand_options/2` are not used.

test/src/timeout_tests.erl
```diff
@@ -0,0 +1,131 @@
...
+%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
+%% the License for the specific language governing rights and
+%% limitations under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% The Initial Developer of the Original Code is GoPivotal, Inc.
+%% Copyright (c) 2007-2015 Pivotal Software, Inc.  All rights reserved.
+%%
+-module(timeout_tests).
```
videlalvaro: please rename to `rabbit_control_main_timeout_tests` or something appropriate. There are many things that can be interpreted as "timeouts" inside the broker.

src/rabbit_channel.erl
```diff
@@ -325,6 +327,14 @@ info_all() ->
 info_all(Items) ->
     rabbit_misc:filter_exit_map(fun (C) -> info(C, Items) end, list()).
 
+info_all(Items, Ref, AggregatorPid) ->
```
videlalvaro: this code only lists one channel

test/src/timeout_tests.erl
```diff
@@ -0,0 +1,131 @@
...
+                     rabbit_networking:connection_info_keys(),
+		     false,
+                     ?TIMEOUT_LIST_OPS_PASS),
+
+    %% list consumers
+    ok = info_action(list_consumers, rabbit_amqqueue:consumer_info_keys(),
+                     false, ?TIMEOUT_LIST_OPS_PASS),
+
+    %% list channels
+    ok = info_action(list_channels, rabbit_channel:info_keys(), false,
```
videlalvaro: Can you make the tests more thorough? Right now this is just testing that `info_action` runs and that's all. There's a bug right now were channels `info_all/3` only returns the first channel on the list and nothing else. So I would expect first that the code above creates more than one channel, more than one queue, and so on. 

Then instead of asserting that the function `info_action` returns OK, I would assert that the channels I created above are actually listed when I call `list_channels`, just to name an example. The test should be automatic, I shouldn't need to read the screen in order to see if it passed.

Perhaps you have to call this function instead: https://github.com/rabbitmq/rabbitmq-server/pull/303/files#diff-0791814e0c4904b21e083a7378194669R773 in order to make your tests become the aggregator Pid or see how to mock that.

src/rabbit_amqqueue.erl
```diff
@@ -586,6 +592,14 @@ info_all(VHostPath, Items) ->
     map(list(VHostPath), fun (Q) -> info(Q, Items) end) ++
         map(list_down(VHostPath), fun (Q) -> info_down(Q, Items, down) end).
 
+info_all(VHostPath, Items, Ref, AggregatorPid) ->
```
videlalvaro: info_all/2 calls `map/2` which in turns calls `rabbit_misc:filter_exit_map/2`. Why is your code not replicating that behaviour?

src/rabbit_networking.erl
```diff
@@ -440,6 +442,13 @@ connection_info(Pid, Items) -> rabbit_reader:info(Pid, Items).
 connection_info_all() -> cmap(fun (Q) -> connection_info(Q) end).
 connection_info_all(Items) -> cmap(fun (Q) -> connection_info(Q, Items) end).
 
+connection_info_all(Items, Ref, AggregatorPid) ->
```
videlalvaro: this code only lists one connection, same problem as with channels

src/rabbit_policy.erl
```diff
@@ -170,6 +170,11 @@ list(VHost) ->
 list_formatted(VHost) ->
     order_policies(list0(VHost, fun format/1)).
 
+list_formatted(VHost, Ref, AggregatorPid) ->
```
videlalvaro: why is this code not ordering the policies like `list_formatted/1`? I understand that it has to emit things as they come, but at the same time it has to wait for `rabbit_runtime_parameters:list/2` to produce the whole results, so why not order them as well?

videlalvaro:
Closing and re-opening later once the erlang.mk changes have been addressed

PR #309, opened 2015-09-11
Improves dropwhile performance

### LINKED ISSUE #308: improve dropwhile performance
The current of implementation of BQ:dropwhile/2 issues a deliver and an ack for each message that's being removed. The QI deliver/ack functions accept lists of message ids, so we could collect the messages to remove, and then send these lists to QI deliver/ack. The advantage here is similar to what we did for paging to disk, and purge: we avoid as many calls as possible to FHC and related subsystems.

The message store also accepts lists of message ids, so we can take advantage of similar optimisations.

videlalvaro:
Fixes #308

### DIFF
--- a/src/rabbit_variable_queue.erl
+++ b/src/rabbit_variable_queue.erl
@@ -616,16 +616,9 @@ drain_confirmed(State = #vqstate { confirmed = C }) ->
     end.
 
 dropwhile(Pred, State) ->
-    case queue_out(State) of
-        {empty, State1} ->
-            {undefined, a(State1)};
-        {{value, MsgStatus = #msg_status { msg_props = MsgProps }}, State1} ->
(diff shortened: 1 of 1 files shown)


src/rabbit_variable_queue.erl
```diff
@@ -1271,48 +1264,102 @@ msg_size(#msg_status{msg_props = #message_properties{size = Size}}) -> Size.
 
 msg_in_ram(#msg_status{msg = Msg}) -> Msg =/= undefined.
 
-remove(AckRequired, MsgStatus = #msg_status {
```
videlalvaro: What I did here was to try to simplify remove/3 by having two function bodies, one for when `AckRequired=true` the other one for when it's `false`.

PR #314, opened 2015-09-16
fixed support for arguments containing ` characters

ben-page:
When rabbitmqctl extracts it's args it doesn't properly quote them before passing them along to rabbitmq. Therefor you can not create a password with a ` in it.

```shell
sudo rabbitmqctl add_user test '`pass'
sh: 1: Syntax error: EOF in backquote substitution
```

This fixes that.

### DIFF
--- a/ebin/rabbit_app.in
+++ b/ebin/rabbit_app.in
@@ -86,7 +86,13 @@
          {mirroring_flow_control, true},
          {credit_flow_initial_credit, 200},
          {credit_flow_more_credit_after, 50},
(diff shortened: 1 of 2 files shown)


michaelklishin:
@ben-page you submitted this against master, do you intend this to be in `3.6.0` only? If not, please rebase against stable.

michaelklishin:
What if an argument contains a quote? Are we sure we're not trying to be too smart here?

ben-page:
I changed this to commit against stable.
https://github.com/rabbitmq/rabbitmq-server/pull/317

PR #329, opened 2015-09-25
allow initial frame_max to be configured

ash-lshift:
SASL challenge/responses happen before connection tuning, at which point there is a hard-coded maximum frame size of 4 KiB. I ran into a situation where a SASL mechanism wanted to exchange challenges/responses greater than 4 KiB, but the broker would close the connection because the frame size limit was being exceeded. I couldn't see a way to fix it apart from this. The AMQP spec doesn't seem to prescribe an initial limit.

### DIFF
--- a/docs/rabbitmq.config.example
+++ b/docs/rabbitmq.config.example
@@ -161,6 +161,11 @@
    %%
    %% {frame_max, 131072},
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
Makes sense, thank you.

Issue #338, opened 2015-09-30
Permission tags may be missing when using separate authentication and authorization backend modules


rtraschke:
When configuring separate authentication and authorization backends, for example with the LDAP auth backend,

```
{rabbit, [{auth_backends, [{rabbit_auth_backend_ldap, rabbit_auth_backend_internal}]}]}
```

then any permission tags found when applying the authorisation module fail to get propagated into the user record, only permission tags found during authentication are applied.

The upshot of this can be seen using the above configuration when a user is properly configured to have Rabbit MQ login access via LDAP authentication, but the permission tag for the management UI is only stored in the user record within the Rabbit MQ internal database, such a user will not be granted access to the management UI.

michaelklishin:
Fixed and will be in `3.5.6`: that you, @rtraschke!

michaelklishin:
…and @Gsantomaggio.

PR #345, opened 2015-10-05
Collect tags from all authz backends

rtraschke:
Fix for issue https://github.com/rabbitmq/rabbitmq-server/issues/338

Have the authorisation code propagate tags into the user record.

A related change to the rabbitmq-auth-backend-ldap plugin propagates tags from there as well. That change also contains the test code. See https://github.com/rabbitmq/rabbitmq-auth-backend-ldap/pull/17

### DIFF
--- a/src/rabbit_access_control.erl
+++ b/src/rabbit_access_control.erl
@@ -100,7 +100,7 @@ try_authorize(Modules, Username) ->
     lists:foldr(
       fun (Module, {ok, ModsImpls, ModsTags}) ->
               case Module:user_login_authorization(Username) of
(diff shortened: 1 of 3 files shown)


michaelklishin:
So this changes one `authz` function signature but given that without it things cannot really work as advertised in the docs, I guess this can be considered to be a bug fix.

@gotthardp FYI.

rtraschke:
Ah, shall I add the old 2-tuple version back into the callback list?

michaelklishin:
@rtraschke yes, please. That would prevent authz plugins that don't really care/rely on tags from breaking.

rtraschke:
Done.

PR #354, opened 2015-10-13
Allow ERLANG_HOME to be overridden from rabbitmq-env.bat

Jakauppila:
Moving the rabbitmq-env.bat call above the ERLANG_HOME check. This grants the ability to override ERLANG_HOME within rabbitmq-env-conf.bat.


michaelklishin:
This is already possible e.g. in `rabbitmq-server.bat`. Good catch.

Issue #112, opened 2015-04-15
Make memory monitor interval configurable


michaelklishin:
[This thread](https://groups.google.com/forum/#!topic/rabbitmq-users/XwPhZ9sZo_w) makes me wonder if we should make it configurable (currently the value is fixed at 2500 ms). The downside of this is that some people will go to ridiculously low values. However, we can't protect the users from themselves (in most cases) anyway.

michaelklishin:
Fixed in #355.

PR #355, opened 2015-10-13
Read memory monitor interval from configuration

dcorbacho:
Solves #112. @michaelklishin should it be documented somewhere else?


michaelklishin:
@dcorbacho no, we don't expect this to be a commonly used option, so this example is sufficient.

PR #359, opened 2015-10-14
fixes failing test

### LINKED ISSUE #358: failing tests on rabbit_variable_queue:head_message_timestamp/4
Here's the error:

```erlang
"Tests failed
Error: {exit, {error,
 {badrecord,basic_message},
 [{rabbit_variable_queue,
 '-head_message_timestamp/4-lc$^1/1-1-',1,
 [{file,"src/rabbit_variable_queue.erl"},{line,902}]},
 {rabbit_variable_queue,head_message_timestamp,4,
 [{file,"src/rabbit_variable_queue.erl"},{line,900}]},
 {rabbit_tests,'-variable_queue_status/1-lc$^0/1-0-',3,
 [{file,"test/src/rabbit_tests.erl"},{line,3050}]},
 {rabbit_tests,'-variable_queue_status/1-lc$^0/1-0-',3,
 [{file,"test/src/rabbit_tests.erl"},{line,3050}]},
 {rabbit_tests,check_variable_queue_status,2,
 [{file,"test/src/rabbit_tests.erl"},{line,3044}]},
 {rabbit_tests,
 test_variable_queue_partial_segments_delta_thing,1,
 [{file,"test/src/rabbit_tests.erl"},{line,3012}]},
 {rabbit_tests,'-with_fresh_variable_queue/1-fun-0-',3,
 [{file,"test/src/rabbit_tests.erl"},{line,2664}]}]}}
```

videlalvaro:
This is a possible fix for #358 

The problem there is that sometimes a msg_status record doesn't have the msg loaded. That's expected.

We can fix it like on this patch, or read the message from disk like here: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_variable_queue.erl#L616 

Not sure we want to be reading msgs from the file system just to show some stats, so I pushed this fix first in case we decide for it.

Fixes #358

### DIFF
--- a/src/rabbit_variable_queue.erl
+++ b/src/rabbit_variable_queue.erl
@@ -894,7 +894,8 @@ head_message_timestamp(Q3, Q4, RPA, QPA) ->
                        [ get_qs_head([Q4, Q3]),
                          get_pa_head(RPA),
                          get_pa_head(QPA) ],
-                   HeadMsgStatus /= undefined ],
+                   HeadMsgStatus /= undefined,
+                   HeadMsgStatus#msg_status.msg /= undefined ],
 
     Timestamps =
         [Timestamp || HeadMsg <- HeadMsgs,


michaelklishin:
This solution is OK. `head_message_timestamp/4` already doesn't assume we will always have the message loaded, which it accounts for by returning an empty string. Which is a curious choice of a value to say "nothing" but that's a separate issue (and present in other parts of RabbitMQ).

videlalvaro:
As I remember the PR, the empty atom that this thing returns is what's expected by the management plugin

michaelklishin:
Management plugin in master then converts such values to `undefined`, so we'll get rid of empty strings in the future.

Issue #84, opened 2015-03-26
rabbit_mirror_queue_misc:all_nodes/0 belongs to rabbit_nodes


michaelklishin:
I find two things about `rabbit_mirror_queue_misc:all_nodes/0` to be slightly confusing:

 * The module it is in.
 * It returns a list of running nodes, which is not obvious from the name.

I suggest we move it to `rabbit_nodes:all_running/0`.

dcorbacho:
@michaelklishin This is solved in https://github.com/rabbitmq/rabbitmq-server/tree/rabbitmq-server-84
The original function wasn't exported, so it doesn't break any previous API. Do you want the pull request on stable or master?

michaelklishin:
stable. Most  small non-breaking changes can go into stable.

michaelklishin:
Fixed in #360.

PR #360, opened 2015-10-14
Move rabbit_mirror_queue_misc:all_nodes/0 to rabbit_nodes:all_running/0

dcorbacho:
Solves #84


michaelklishin:
@dcorbacho a minor suggestion: when you think that something is ready for a review, assign the PR to someone (me or @videlalvaro, until we move to erlang.mk). That person will be notified and take notice.

Issue #63, opened 2015-03-10
Include net_ticktime into rabbitmqctl status output


michaelklishin:
I suggest that we add `kernel.net_ticktime` to `rabbitmqctl status` and `rabbitmqctl report` output. While not a RabbitMQ configuration setting, it can lead to false positives and false negatives w.r.t. node reachability. Knowing the effective value should make debugging and operations a bit easier.

simonmacmullen:
`rabbitmqctl report` already includes `rabbitmqctl environment` and thus dumps all (configured) environment settings, so if it's not there it's the default.

There may be value to adding it to `rabbitmqctl status`. Should be careful of not adding everything to everything though...

michaelklishin:
Fixed by #361.

PR #361, opened 2015-10-15
Adds net_ticktime to rabbitmqctl status output.

Ayanda-D:
Adding `kernel.net_ticktime` to `rabbitmqctl status` output, which is also included in 
`rabbitmqctl report` output.

References #63


michaelklishin:
Thank you!

PR #363, opened 2015-10-16
Port Fuel fixes for the HA rabbitmq OCF

bogdando:
Sync OCF script changes and fixes from Fuel,
version 8.0dev/9ca5b6709cd18c8134cd8feaccf937ef49a625b4
Changelog brief:
9ca5b67 - Matthew Mosesohn, 8 days ago : Wait for rabbitmq sync before stop/demote actions
00a6a01 - Davanum Srinivas, 4 days ago : Avoid division operation in shell
c1900b4 - Dmitry Mescheryakov, 4 weeks ago : Start RabbitMQ app on notify
00f28b5 - Vladimir Kuklin, 4 weeks ago : Return NOT_RUNNING when beam is not RUNNING
2707a5e - Dmitry Mescheryakov, 7 weeks ago : Make RabbitMQ OCF script tolerate rabbitmqctl timeouts
403b28c - Bogdan Dobrelya, 6 weeks ago : Detect a last man standing for rabbit OCF agent
60d5743 - Victor Sergeyev, 7 weeks ago : Add more logs to rabbitmq get_status function
11c4e4c - Alex Schultz, 9 weeks ago : Fix rabbitmq data restore for large datasets
8ccdfc6 - Stanislav Makar, 6 months ago : Implement the dumping of rabbitMQ definitions
5097d94 - Bogdan Dobrelya, 3 months ago : Fix error return codes for rabbit OCF
b5ae83a - Bogdan Dobrelya, 3 months ago : Fix chowning for rabbit OCF
bf604f8 - Bogdan Dobrelya, 4 months ago : Restart rabbit if can't list queues or found memory alert

The changelog full, with gerrit change-ids and related LP bugs:
commit 9ca5b6709cd18c8134cd8feaccf937ef49a625b4
Author: Matthew Mosesohn <mmosesohn@mirantis.com>
Date:   Mon Oct 5 15:48:33 2015 +0300

    Wait for rabbitmq sync before stop/demote actions
    
    Added new OCF key stop_time (corresponding to start_time)
    Added wait_sync function which tries until start_time/2
    for queues on stopped/demoted node to reach synced state.
    
    Added optional [-t timeout] to su_rabbit_cmd function to
    provide arbitrary timeout
    
    Change-Id: Iae2211b3d477a9603a58d5eacb12e0fba924861a
    Closes-Bug: #1464637

commit 00a6a016d5235de4cbde26cb40db71119967d3c7
Author: Davanum Srinivas <davanum@gmail.com>
Date:   Fri Oct 9 09:42:59 2015 -0700

    Avoid division operation in shell
    
    When the data returned from 'rabbitmqctl list_queues' grows a lot
    and awk sums up all the rows especially for memory calculation it
    returns the sum in scientific notation (example from bug
    was .15997e+09), later when we want to calculate the memory in
    MB instead of bytes, the bash division does not like this string.
    
    We can just avoid the situation by doing the division into MB
    in awk itself. Since we don't need the memory in bytes anyway.
    
    Closes-Bug: #1503331
    Change-Id: I38d25406b84d0f70ed62101d5fb5ba108bcab8bd

commit c1900b49e6ddcfb84bf5c501c75f3fee80903eca
Author: Dmitry Mescheryakov <dmescheryakov@mirantis.com>
Date:   Fri Sep 18 15:05:03 2015 +0300

    Start RabbitMQ app on notify
    
    On notify, if we detect that we are a part of a cluster we still
    need to start the RabbitMQ application, because it is always
    down after action_start finishes.
    
    Closes-Bug: #1496386
    Change-Id: I307452b687a6100cc4489c8decebbc3dccdbc432

commit 00f28b5cc48bdba389922fa460e4a6e6e173589e
Author: Vladimir Kuklin <vkuklin@mirantis.com>
Date:   Tue Sep 15 14:39:08 2015 +0300

    Return NOT_RUNNING when beam is not RUNNING
    
    Change get_status to return NOT_RUNNING when
    beam is not_running. Otherwise, pacemaker
    will get stuck during rabbitmq failover and
    will not attempt to restart the failed resource
    
    Change-Id: I926a3eafa9968abdf07baa5f2d5c22480300fb30
    Closes-bug: #1484280

commit 2707a5ebbff7012a94de77b60fd594f5bcb29e05
Author: Dmitry Mescheryakov <dmescheryakov@mirantis.com>
Date:   Tue Aug 25 17:38:44 2015 +0300

    Make RabbitMQ OCF script tolerate rabbitmqctl timeouts
    
    The change makes OCF script ignore small number of timeouts of rabbitmqctl
    for 'heavy' operations: list_channels, get_alarms and list_queues.
    Number of tolerated timeouts in a row is configured through a new variable
    'max_rabbitmqctl_timeouts'. By default it is set to 1, i.e. rabbitmqctl
    timeouts are not tolerated at all.
    
    Bug #1487517 is fixed by extracting declaration of local variables
    'rc_alarms' and 'rc_queues' from assignment operations.
    
    
    Text for Operations Guide:
    
    If on node where RabbitMQ is deployed
    other processes consume significant part of CPU, RabbitMQ starts
    responding slow to queries by 'rabbitmqctl' utility. The utility is
    used by RabbitMQ's OCF script to monitor state of the RabbitMQ.
    When utility fails to return in pre-defined timeout, OCF script
    considers RabbitMQ to be down and restarts it, which might lead to
    a limited (several minutes) OpenStack downtime. Such restarts
    are undesirable as they cause downtime without benefit. To
    mitigate the issue, the OCF script might be told to tolerate
    certain amount of rabbitmqctl timeouts in a row using the following
    command:
      crm_resource --resource p_rabbitmq-server --set-parameter \
          max_rabbitmqctl_timeouts --parameter-value N
    
    Here N should be replaced with the number of timeouts. For instance,
    if it is set to 3, the OCF script will tolerate two rabbitmqctl
    timeouts in a row, but fail if the third one occurs.
    
    By default the parameter is set to 1, i.e. rabbitmqctl timeout is not
    tolerated at all. The downside of increasing the parameter is that
    if a real issue occurs which causes rabbitmqctl timeout, OCF script
    will detect that only after N monitor runs and so the restart, which
    might fix the issue, will be delayed.
    
    To understand that RabbitMQ's restart was caused by rabbitmqctl timeout
    you should examine lrmd.log of the corresponding controller on Fuel
    master node in /var/log/docker-logs/remote/ directory. Here lines like
    "the invoked command exited 137: /usr/sbin/rabbitmqctl list_channels ..."
    
    indicate rabbitmqctl timeout. The next line will explain if it
    caused restart or not. For example:
    "rabbitmqctl timed out 2 of max. 3 time(s) in a row. Doing nothing for now."
    
    DocImpact: user-guide, operations-guide
    
    Closes-Bug: #1479815
    Closes-Bug: #1487517
    Change-Id: I9dec06fc08dbeefbc67249b9e9633c8aab5e09ca

commit 403b28c2aff8aa9f37125d3ff1ac09861990da7e
Author: Bogdan Dobrelya <bdobrelia@mirantis.com>
Date:   Thu Sep 3 13:53:13 2015 +0200

    Detect a last man standing for rabbit OCF agent
    
    W/o this patch, the race condition is possible
    when there is no running rabbit nodes/resource
    master. The rabbit nodes will start/stop in an
    endless loop as a result introducing full downtime
    for AMQP cluster and cloud control plane.
    
    The solution is:
    * On post-start/post-promote notify, do nothing, if
      either of the following is a true:
      - there is no rabbit resources running or no master
      - the list of rabbit resources being started/promoted
        reported empty
    * For such cases, do not report resource failure and delegate
      recovery, if needed, to the "running out of the cluster"
      monitor's logic.
    * Additionally, report about a last man standing when
      there is no running rabbit resources around.
    
    Closes-bug: #1491306
    
    Change-Id: If1c62fac26b63410636413c49fce55c35e53dc5f

commit 60d5743c41999a43d87374357866cd3c3c02aa64
Author: Victor Sergeyev <vsergeyev@mirantis.com>
Date:   Wed Aug 26 19:13:55 2015 +0300

    Add more logs to rabbitmq get_status function
    
    It's really hard to debug, when get_status() returns $OCF_NOT_RUNNING
    only and looses exit code and error output.
    
    Added more logs to avoid of this situation.
    
    Related-Bug: #1488999
    
    Change-Id: Id0999235d7be688f55799e2952fe22e97b678ce7

commit 11c4e4c3985949f776367dd5045102c6398aef89
Author: Alex Schultz <aschultz@mirantis.com>
Date:   Thu Aug 13 09:31:51 2015 -0500

    Fix rabbitmq data restore for large datasets
    
    Previously we were sending the json backup data on the command line
    which fails when the dataset is large. This change updates the command
    line options for curl to pass the filename directly and let it handle
    the reading of the data.
    
    Change-Id: I37f298279beca06df41fb08e1745602976c6a776
    Closes-Bug: 1383258

commit 8ccdfc62152de8212ba6a0aeed45c1e3677c70a4
Author: Stanislav Makar <smakar@mirantis.com>
Date:   Thu Apr 16 14:32:51 2015 +0000

    Implement the dumping of rabbitMQ definitions
    
    This changes leverages the rabbitmq management plugin to dump
    exchanges, queues, bindings, users, virtual hosts, permissions and
    parameters from the running system. Specifically this change adds the
    following:
    
    * The dumping rabbitMQ definitions (users/vhosts/exchanges/etc) during
      the end of the deployment
    * The possibility to restore definitions to the rabbitmq-server ocf
      script during rabbitMQ startup.
    * Enabled rabbitmq admin plugin, but restricts it to localhost traffic.
      This reverts Ic01c26200f6019a8112b1c5fb04a282e64b3b3e6 but adds
      firewall rules to mitigate the issue.
    
    DocImpact: The dump_rabbit_definitions task can be used to backup the
    rabbitmq definitions and if custom definitions (users/vhosts/etc) are
    created it must be run or the changes may be lost during the rabbitmq
    failover via pacemaker.
    
    Change-Id: I715f7c2ae527f7e105b9f6b7d82c443e8accf178
    Closes-bug: #1383258
    Related-bug: #1450443
    Co-Authored-By: Alex Schultz <aschultz@mirantis.com>

commit 5097d94f5d56fd6126ca9b7c1227961536c94399
Author: Bogdan Dobrelya <bdobrelia@mirantis.com>
Date:   Tue Jul 7 13:32:25 2015 +0200

    Fix error return codes for rabbit OCF
    
    W/o this fix the situation is possible when
    rabbit OCF returns OCF_NOT_RUNNING in the hope of
    future restart of the resource by pacemaker.
    
    But in fact, pacemaker will not trigger restart action
    if monitor returns "not running". This is an issue
    as we want resource restarted.
    
    The solution is to return OCF_ERR_GENERIC instead of
    OCF_NOT_RUNNING when we expect the resource to be restarted
    (which is action stop plus action start).
    
    Closes-bug: #1472230
    
    Change-Id: I10c6e43d92cb23596636d86932674b36864d1595

commit b5ae83a15be191d46d4b0da3d4bc67420311896a
Author: Bogdan Dobrelya <bdobrelia@mirantis.com>
Date:   Tue Jul 7 10:16:08 2015 +0200

    Fix chowning for rabbit OCF
    
    W/o this fix, the list of file names not
    accessible by rabbitmq user will be treated
    as multiple arguments to the if command causing
    it to throw the "too many arguments" error and
    the chown command to be skipped.
    
    This is the problem as it might prevent the rabbitmq
    server from starting because of a bad files ownership.
    
    The solution is to pass the list of files as a single
    argument "${foo}".
    
    Closes-bug: #1472175
    
    Change-Id: I1d00ec3f31cd0f023bd58a4e11e5b31659977229

commit bf604f80d72f69e771152b153973fa38fa83afd8
Author: Bogdan Dobrelya <bdobrelia@mirantis.com>
Date:   Wed Jun 10 13:44:53 2015 +0200

    Restart rabbit if can't list queues or found memory alert
    
    W/o this fix the dead end situation is possible
    when the rabbit node have no free memory resources left
    and the cluster blocks all publishing, by design.
    But the app thinks "let's wait for the publish block have
    lifted" and cannot recover.
    
    The workaround is to monitor results
    of crucial rabbitmqctl commands and restart the rabbit node,
    if queues/channels/alarms cannot be listed or if there are
    memory alarms found.
    This is the similar logic as we have for the cases when
    rabbitmqctl list_channels hangs. But the channels check is also
    fixed to verify if the exit code>0 when the rabbit app is
    running.
    
    Additional checks added to the monitor also require extending
    the timeout window for the monitor action from 60 to 180 seconds.
    
    Besides that, this patch makes the monitor action to gather the
    rabbit status and runtime stats, like consumed memory by all
    queues of total Mem+Swap, total messages in all queues and
    average queue consumer utilization. This info should help to
    troubleshoot failures better.
    
    DocImpact: ops guide. If any rabbitmq node exceeded its memory
    threshold the publish became blocked cluster-wide, by design.
    For such cases, this rabbit node would be recovered from the
    raised memory alert and immediately stopped to be restarted
    later by the pacemaker. Otherwise, this blocked publishing state
    might never have been lifted, if the pressure persists from the
    OpenStack apps side.
    
    Closes-bug: #1463433
    
    Change-Id: I91dec2d30d77b166ff9fe88109f3acdd19ce9ff9

__

Co-authored-by: Matthew Mosesohn <mmosesohn@mirantis.com>
Co-authored-by: Davanum Srinivas <dsrinivas@mirantis.com>
Co-authored-by: Vladimir Kuklin <vkuklin@mirantis.com>
Co-authored-by: Dmitry Mescheryakov <dmescheryakov@mirantis.com>
Co-authored-by: Victor Sergeyev <vsergeyev@mirantis.com>
Co-authored-by: Alex Schultz <aschultz@mirantis.com>
Co-authored-by: Makar Stanislav <smakar@mirantis.com>

### DIFF
--- a/packaging/common/rabbitmq-server-ha.ocf
+++ b/packaging/common/rabbitmq-server-ha.ocf
@@ -30,6 +30,9 @@ OCF_RESKEY_ctl_default="/usr/sbin/rabbitmqctl"
 OCF_RESKEY_debug_default=false
 OCF_RESKEY_username_default="rabbitmq"
 OCF_RESKEY_groupname_default="rabbitmq"
+OCF_RESKEY_admin_user_default="guest"
+OCF_RESKEY_admin_password_default="guest"
+OCF_RESKEY_definitions_dump_file_default="/etc/rabbitmq/definitions"
 OCF_RESKEY_pid_file_default="/var/run/rabbitmq/pid"
 OCF_RESKEY_log_dir_default="/var/log/rabbitmq"
 OCF_RESKEY_mnesia_base_default="/var/lib/rabbitmq/mnesia"
@@ -37,6 +40,7 @@ OCF_RESKEY_node_port_default=5672
 OCF_RESKEY_erlang_cookie_default=false
 OCF_RESKEY_erlang_cookie_file_default="/var/lib/rabbitmq/.erlang.cookie"
 OCF_RESKEY_use_fqdn_default=false
+OCF_RESKEY_max_rabbitmqctl_timeouts_default=1
 
 : ${HA_LOGTAG="lrmd"}
 : ${HA_LOGFACILITY="daemon"}
@@ -45,6 +49,9 @@ OCF_RESKEY_use_fqdn_default=false
 : ${OCF_RESKEY_debug=${OCF_RESKEY_debug_default}}
 : ${OCF_RESKEY_username=${OCF_RESKEY_username_default}}
 : ${OCF_RESKEY_groupname=${OCF_RESKEY_groupname_default}}
+: ${OCF_RESKEY_admin_user=${OCF_RESKEY_admin_user_default}}
+: ${OCF_RESKEY_admin_password=${OCF_RESKEY_admin_password_default}}
+: ${OCF_RESKEY_definitions_dump_file=${OCF_RESKEY_definitions_dump_file_default}}
 : ${OCF_RESKEY_log_dir=${OCF_RESKEY_log_dir_default}}
 : ${OCF_RESKEY_mnesia_base=${OCF_RESKEY_mnesia_base_default}}
 : ${OCF_RESKEY_pid_file=${OCF_RESKEY_pid_file_default}}
@@ -52,11 +59,14 @@ OCF_RESKEY_use_fqdn_default=false
 : ${OCF_RESKEY_erlang_cookie=${OCF_RESKEY_erlang_cookie_default}}
 : ${OCF_RESKEY_erlang_cookie_file=${OCF_RESKEY_erlang_cookie_file_default}}
 : ${OCF_RESKEY_use_fqdn=${OCF_RESKEY_use_fqdn_default}}
+: ${OCF_RESKEY_max_rabbitmqctl_timeouts=${OCF_RESKEY_max_rabbitmqctl_timeouts_default}}
 
 #######################################################################
 
 OCF_RESKEY_start_time_default=$((OCF_RESKEY_CRM_meta_timeout / 6000 + 2))
 : ${OCF_RESKEY_start_time=${OCF_RESKEY_start_time_default}}
+OCF_RESKEY_stop_time_default=${OCF_RESKEY_start_time_default}
+: ${OCF_RESKEY_stop_time=${OCF_RESKEY_start_time_default}}
 OCF_RESKEY_command_timeout_default=""
 : ${OCF_RESKEY_command_timeout=${OCF_RESKEY_command_timeout_default}}
 TIMEOUT_ARG=$((OCF_RESKEY_CRM_meta_timeout / 6000 + 30))
@@ -141,6 +151,30 @@ RabbitMQ group name
 <content type="string" default="${OCF_RESKEY_groupname_default}" />
 </parameter>
 
+<parameter name="admin_user" unique="0" required="0">
+<longdesc lang="en">
+RabbitMQ default admin user for API
+</longdesc>
+<shortdesc lang="en">RabbitMQ admin user</shortdesc>
+<content type="string" default="${OCF_RESKEY_admin_user_default}" />
+</parameter>
+
+<parameter name="admin_password" unique="0" required="0">
+<longdesc lang="en">
+RabbitMQ default admin user password for API
+</longdesc>
+<shortdesc lang="en">RabbitMQ admin password</shortdesc>
+<content type="string" default="${OCF_RESKEY_admin_password_default}" />
+</parameter>
+
+<parameter name="definitions_dump_file" unique="0" required="0">
+<longdesc lang="en">
+RabbitMQ default definitions dump file
+</longdesc>
+<shortdesc lang="en">RabbitMQ definitions dump file</shortdesc>
+<content type="string" default="${OCF_RESKEY_definitions_dump_file}" />
+</parameter>
+
 <parameter name="command_timeout" unique="0" required="0">
 <longdesc lang="en">
 Timeout command arguments for issued commands termination (value is auto evaluated)
@@ -155,6 +189,14 @@ Timeout for start rabbitmq server
 </longdesc>
 <shortdesc lang="en">Timeout for start rabbitmq server</shortdesc>
 <content type="string" default="${OCF_RESKEY_start_time_default}" />
+</parameter>
+
+<parameter name="stop_time" unique="0" required="0">
+<longdesc lang="en">
+Timeout for stopping rabbitmq server
+</longdesc>
+<shortdesc lang="en">Timeout for stopping rabbitmq server</shortdesc>
+<content type="string" default="${OCF_RESKEY_stop_time_default}" />
 </parameter>
 
 <parameter name="debug" unique="0" required="0">
@@ -207,6 +249,16 @@ Either to use FQDN or a shortname for the rabbitmq node
 <content type="boolean" default="${OCF_RESKEY_erlang_cookie_file_default}" />
 </parameter>
 
+<parameter name="max_rabbitmqctl_timeouts" unique="0" required="0">
+<longdesc lang="en">
+If during monitor call rabbitmqctl times out, the timeout is ignored
+unless it is Nth timeout in a row. Here N is the value of the current parameter.
+If too many timeouts happen in a raw, the monitor call will return with error.
+</longdesc>
+<shortdesc lang="en">Fail only if that many rabbitmqctl timeouts in a row occurred</shortdesc>
+<content type="string" default="${OCF_RESKEY_max_rabbitmqctl_timeouts_default}" />
+</parameter>
+
 </parameters>
 
 <actions>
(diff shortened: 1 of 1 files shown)


michaelklishin:
@bogdando so you want this in master only?

bogdando:
This contains major and minor bug fixes and several features, listed in the full change log.

I tested this change manually. I suggest to sync fixes this way, unless we'll get CI here...

michaelklishin:
We don't have CI for OCF and are happy to merge changes from you regularly. However, I suspect you want this to be in 3.5.7? Then this PR is against the wrong branch.

bogdando:
@michaelklishin There would be no harm to backport it as well. There were no any usage of this yet.

bogdando:
okay, will resubmit to stable as well!

michaelklishin:
Thanks, I'll close this one then.

bogdando:
But Is it OK to submit backport ahead of the master?..

michaelklishin:
@bogdando all `stable` changes are merged into `master`. We have scripts that nag us when that's not the case.

michaelklishin:
@bogdando see https://github.com/rabbitmq/rabbitmq-server/commit/98fc9f7e29eb57d1e0765420301db9d321657684, which is in master.

Thank you for your work on this!

bogdando:
Thank you too. More users of this OCF script we will manage to get, the better rabbitmq clustering UX for the community and other projects like OpenStack as well.

PR #365, opened 2015-10-16
Sync lost change from Fuel HA rabbitmq OCF fixes

bogdando:
Fix missed change for the return value from the fix:
5097d94 - Bogdan Dobrelya, 3 months ago : Fix error return codes for rabbit OCF


bogdando:
There was a change lost while I was syncing fixes...

Issue #371, opened 2015-10-20
rabbitmqctl list_queues doesn't support `exclusive` as an info item


michaelklishin:
Instead, it supports `owner_pid`, which makes little sense to beginners.

So the issue is very similar to https://github.com/rabbitmq/rabbitmq-management/issues/40.

michaelklishin:
A fix contributed by @binarin will be in `3.5.7`.

PR #373, opened 2015-10-20
Add `exclusive` item to `rabbitmqctl list_queues`

binarin:
In addition to owner_pid, which makes little sense to beginners.

Just as described in #371

### DIFF
--- a/docs/rabbitmqctl.1.xml
+++ b/docs/rabbitmqctl.1.xml
@@ -1230,6 +1230,11 @@
                   queue is non-exclusive.</para></listitem>
               </varlistentry>
               <varlistentry>
+                <term>exclusive</term>
+                <listitem><para>True if queue is exclusive (i.e. has
+                  owner_pid), false otherwise</para></listitem>
+              </varlistentry>
+              <varlistentry>
                 <term>exclusive_consumer_pid</term>
                 <listitem><para>Id of the Erlang process representing the channel of the
                   exclusive consumer subscribed to this queue. Empty if
--- a/src/rabbit_amqqueue_process.erl
+++ b/src/rabbit_amqqueue_process.erl
@@ -92,7 +92,8 @@
          durable,
          auto_delete,
          arguments,
-         owner_pid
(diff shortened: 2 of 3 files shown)


michaelklishin:
Thank you! This looks good. We have a fairly non-trivial PR landing in the same area in the near future (see #303) but there's less overlap than I expected.

Can you please rebase this against `stable` and re-submit? This can go into `3.5.7`.

videlalvaro:
@michaelklishin why would this go into 3.5.7 considering that it adds new things to CTL? Why not wait for 3.6.0 which anyway is happening soon?

michaelklishin:
@videlalvaro it's not a breaking change and it is very small, why not ship it earlier?

videlalvaro:
Because there are version numbers for a reason? Other wise what's the meaning of them if we don't respect them? We could use just one increasing integer…

Also, what's the rush?

michaelklishin:
@videlalvaro we've been shipping small features in earlier `3.5.x` releases. Please keep your snark to yourself.

michaelklishin:
@dumbbell do you have an opinion on this?

michaelklishin:
@binarin no need to rebase anything: we can cherry-pick to stable if we decide to do so. Thanks again for your contribution.

dumbbell:
@michaelklishin, @videlalvaro: I'm fine with small low-impact new features in a stable branch. We already did that in the past. We can discuss what we mean behind our version numbers outside of this issue though, and in particular if we want to follow "semver" rules or not.

In this specific case, does an old `rabbitmqctl` still work with a new RabbitMQ? And the opposite?

michaelklishin:
@dumbbell we never tried to support cross-version `rabbitmq-server` and `rabbitmqctl`. But this change is hardly breaking: it only adds a new stats info and output column.

michaelklishin:
@binarin thank you. Works like a charm.

michaelklishin:
Cherry-picked to `stable`.

PR #374, opened 2015-10-21
More OCF HA fixes ported from OpenStack Fuel

bogdando:
Minor fixes based on bugs discovered in the Fuel project


michaelklishin:
Thank you!

PR #377, opened 2015-10-21
Implements Lazy Queues

### LINKED ISSUE #351: lazy queues
This bug is to experiment with the idea of introducing "lazy queues". The idea is to support use cases where consumers are disconnected for long periods of time, but where we want to support a constant message ingress rate.

Lazy queues are `rabbit_variable_queues` that can work in two modes: `lazy` and `default`. 

**Default** mode is what RabbitMQ supports today: queues maintain a cache of messages in memory, and from time to time, depending on memory pressure, they page some of those messages to disk. The problem with this mode is that if there are too many messages to page out, the queue can become unresponsive, making it hard to predict the behaviour/performance of a queue.

**Lazy** mode: this mode will force queues to write to disk every message as they enter the queue via `basic.publish`. In this case the variable queue _will not cache messages in memory_. Whenever a consumer requests a message, the variable queue will load messages from disk into memory in batches of QI's `?SEGMENT_ENTRY_COUNT` msgs (16384). 

The modes should be configurable either by queue.declare arguments `x-queue-mode = "lazy"`, assuming "default" in the case when no argument is provided.

Queue modes can also be specified via the `queue-mode` policy.

Queue modes can be switched back and forth by setting/unsetting policies.

videlalvaro:
Fixes #351

This should be merged together with:

https://github.com/rabbitmq/rabbitmq-common/pull/14
https://github.com/rabbitmq/rabbitmq-test/pull/5 
https://github.com/rabbitmq/rabbitmq-website/pull/99

on the test repo you can run the following:

`make standalone-tests FILTER=lazy_queue_test` and
`make lazy-vq-tests`


videlalvaro:
This should be merged together with:

https://github.com/rabbitmq/rabbitmq-website/pull/99 and https://github.com/rabbitmq/rabbitmq-test/pull/5

src/rabbit_variable_queue.erl
```diff
@@ -880,6 +905,51 @@ invoke(      _,   _, State) -> State.
...
+    State;
+set_queue_mode(lazy, State = #vqstate {
+                                target_ram_count = TargetRamCount }) ->
+    %% To become a lazy queue we need to page everything to disk first.
+    State1 = convert_to_lazy(State),
+    %% restore the original target_ram_count
+    a(State1 #vqstate { mode = lazy, target_ram_count = TargetRamCount });
+set_queue_mode(default, State) ->
+    %% becoming a default queue means loading messages from disk like
+    %% whene a queue is recovered.
```
javiereguiluz: typo? `whene` -> `when`

src/rabbit_variable_queue.erl
```diff
@@ -951,9 +1021,16 @@ a(State = #vqstate { q1 = Q1, q2 = Q2, delta = Delta, q3 = Q3, q4 = Q4,
     E4 = ?QUEUE:is_empty(Q4),
     LZ = Len == 0,
 
+    %% if q1 has messsages then q3 cannot be empty. See publish/6.
```
javiereguiluz: typo: `messsages` -> `messages`

src/rabbit_variable_queue.erl
videlalvaro: Looks like it, thanks @javiereguiluz

src/rabbit_variable_queue.erl
```diff
@@ -966,6 +1043,53 @@ a(State = #vqstate { q1 = Q1, q2 = Q2, delta = Delta, q3 = Q3, q4 = Q4,
...
+                     ram_bytes        = RamBytes}) ->
+    E1 = ?QUEUE:is_empty(Q1),
+    E2 = ?QUEUE:is_empty(Q2),
+    ED = Delta#delta.count == 0,
+    E3 = ?QUEUE:is_empty(Q3),
+    E4 = ?QUEUE:is_empty(Q4),
+    LZ = Len == 0,
+    L3 = ?QUEUE:len(Q3),
+
+    %% q1 must always be empty, since q1 only get messages during
```
javiereguiluz: grammar issue: `q1 only get messages` -> `q1 only gets messages`

src/rabbit_variable_queue.erl
```diff
@@ -966,6 +1043,53 @@ a(State = #vqstate { q1 = Q1, q2 = Q2, delta = Delta, q3 = Q3, q4 = Q4,
...
+    %% publish, but for lazy queues messages go straight to delta.
+    true = E1,
+
+    %% q2 only gets messages from q1 when push_alphas_to_betas is
+    %% called for a non empty delta, which won't be the case for a
+    %% lazy queue. This means q2 must always be empty.
+    true = E2,
+
+    %% q4 must always be empty, since q1 only get messages during
```
javiereguiluz: `get messages` -> `gets messages`

test/src/rabbit_tests.erl
```diff
@@ -2699,25 +2721,64 @@ wait_for_confirms(Unconfirmed) ->
...
+     fun test_fetchwhile_varying_ram_duration/1,
+     fun test_variable_queue_ack_limiting/1,
+     fun test_variable_queue_purge/1,
+     fun test_variable_queue_requeue/1,
+     fun test_variable_queue_requeue_ram_beta/1,
+     fun test_variable_queue_fold/1,
+     fun test_variable_queue_batch_publish/1,
+     fun test_variable_queue_batch_publish_delivered/1].
+
+%% same as test_variable_queue_requeue_ram_beta but randomly chaning
```
javiereguiluz: typo: `chaning` -> `changing`

Issue #207, opened 2015-07-01
RAM limit configuration using absolute values


mthenw:
Reporting this based on discussion from mailing list https://groups.google.com/forum/#!topic/rabbitmq-users/ESnCf64vfZ0

Similar to how disk limit can be set, there could be also configuration value for memory limit. This could be helpful when running rabbitmq inside docker container with cgroup memory limitations.

michaelklishin:
Fixed in #378, thank you @dcorbacho!

mthenw:
Awesome, thanks!

PR #378, opened 2015-10-22
Accept absolute values in memory high watermark

dcorbacho:
Solves #207


michaelklishin:
Followed-up with https://github.com/rabbitmq/rabbitmq-server/issues/447.

Issue #369, opened 2015-10-19
queue-master-locator policy has the wrong name


videlalvaro:
See here: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_queue_location_validator.erl#L28

Policies don't have the `x-` prefix. That's only for declare parameters.

michaelklishin:
@videlalvaro so we'd still keep the `queue.declare` arguments key prefixed with an `x-`? So far I've updated the docs to only mention setting locator via policies, to not have to explain this difference.

videlalvaro:
Yes. Arguments have the x- prefix because they are protocol extensions.
Policies don't use that.

michaelklishin:
@videlalvaro https://github.com/rabbitmq/rabbitmq-server/pull/380, https://github.com/rabbitmq/rabbitmq-test/pull/6, https://github.com/rabbitmq/rabbitmq-website/pull/98 are ready. The `queue.declare` argument key is unchanged.

michaelklishin:
#380 is no longer relevant, see https://github.com/rabbitmq/rabbitmq-server/pull/390. https://github.com/rabbitmq/rabbitmq-test/pull/6 has been updated.

michaelklishin:
@videlalvaro friendly reminder that this is waiting QA.

videlalvaro:
@michaelklishin once I'm able to run the tests I will QA this and other PRs

videlalvaro:
Fixed via https://github.com/rabbitmq/rabbitmq-server/pull/390 and related tickets

PR #380, opened 2015-10-22
Correct queue master locator policy key

michaelklishin:
Part of #369.


michaelklishin:
Needs to be resubmitted because of `erlang.mk`.

PR #382, opened 2015-10-23
Recognize help as a valid sub-command for rabbitmqctl and rabbitmq-plugins

lemenkov:
Some people believe that rabbitmqctl and rabbitmq-plugins should provide
a valid help subcommand (instead of throwing an error). So here it is.

This is mostly a cosmetic patch so sorry for the noise in advance.


michaelklishin:
@lemenkov completely agree. Can you please rebase against `stable` and re-submit? Then it would be in `3.5.7`.

Thank you!

michaelklishin:
Closing assuming @lemenkov will re-submit.

PR #383, opened 2015-10-23
Recognize help as a valid sub-command for rabbitmqctl and rabbitmq-plugins

lemenkov:
Some people believe that rabbitmqctl and rabbitmq-plugins should provide a valid help subcommand (instead of throwing an error). So here it is.

This is mostly a cosmetic patch so sorry for the noise in advance. Rebased on top of the stable.


michaelklishin:
:+1:

PR #387, opened 2015-10-23
Disable FHC read cache by default

### LINKED ISSUE #386: Disable FHC read cache by default
As previously discussed, we want to get rid of the cache because the OS (and filesystems) is capable of doing the same job better.

For `3.6.0` we will disable read cache by default. After that, the read cache will go entirely.

michaelklishin:
Fixes #386.


michaelklishin:
Apparently this conflicts with `erlang.mk` changes. Will re-submit.

PR #390, opened 2015-10-25
Rename policy parameters, references #369

michaelklishin:
Supercedes #380, which is no longer relevant after the `erlang.mk` merge.


michaelklishin:
Part of https://github.com/rabbitmq/rabbitmq-server/issues/369.

Issue #117, opened 2015-04-17
prelaunch assumes IPv4


michaelklishin:
``` erlang
gen_tcp:listen(Port, [inet, {reuseaddr, true}])
```

won't work well if `epmd` and RabbitMQ is configured to use IPv6. We should support IPv6-only environments, too.

yottatsa:
I found that https://github.com/yandex/inet64_tcp fixes clustering over IPv6 if you talking about it.

michaelklishin:
Thanks. Not just clustering but also our own code. But it's good to know about that project.

michaelklishin:
Fixed by #391, thank you, @dcorbacho.

PR #394, opened 2015-10-26
Fix spelling of function name

aboroska:
The misspelling was confusing. These local functions are only used here. Introduced by #117.


aboroska:
Thanks!

michaelklishin:
@aboroska thank you!

PR #381, opened 2015-10-23
Fix cluster-wide alarm handling

binarin:
Initial draft for discussion.

- Properly handle clearing alarms from dead nodes - #362.
- Unblock publisher only when all alarms throughout the cluster was
  cleared - #379.
- Disable/enable paging in rabbit_memory_monitor only in response
  to disk resource alarms from current node (I'm not totally sure, but I think this is a bug).
- Add some documentation.

Issue #379, opened 2015-10-22
Blocked connection due to multiple resource alarms may be cleared by a single node in a cluster


noahhaon:
Steps to reproduce:

1. Cause 2 of 3 nodes in a cluster to raise a disk resource alarm
2. Observe publisher connections enter `blocked` state
3. Clear the disk resource alarm on 1 of 2 nodes
4. Observe resource alarm remaining on 1 node
5. Observe publisher connections leaving `blocked` state despite 1 node still in a resource alarm state

videlalvaro:
Can I assume that a memory alarm would cause the same symptoms?

videlalvaro:
>Observe publisher connections leaving blocked state despite 1 node still in a resource alarm state

So what happens if we unblock publishers and then they publish a message that has to be replicated to the node that still has the resource alarm?

michaelklishin:
Any alarm should behave the same way. We should not clear the alarm if there is at least 1 node online that is in alarmed state. That has been the intent all along.

noahhaon:
> Can I assume that a memory alarm would cause the same symptoms?

> So what happens if we unblock publishers and then they publish a message that has to be replicated to the node that still has the resource alarm?

I seem to recall running into this case with earlier versions of RMQ, and it leading to node crashes as the publishers would become unblocked and memory pressure on the affected node would continue to increase. I believe in that case it was for a slave of a mirrored queue.  We still need to verify the recent improvements/fixes to mirrored queue performance, but I suspect those issues were the root cause of the memory exhaustion.

The publishers may enter a flow control state, but that seems like it was incidental to the alarm.

michaelklishin:
PRs mergeable into master:

 * https://github.com/rabbitmq/rabbitmq-server/pull/395
 * https://github.com/rabbitmq/rabbitmq-common/pull/1

michaelklishin:
Fixed in master. Thank you, @binarin.

Issue #362, opened 2015-10-15
Cluster-wide alarm state is not updated when alarmed node goes down


michaelklishin:
See [this thread](https://groups.google.com/forum/#!topic/rabbitmq-users/py4o6Ia9VLk).

binarin:
I think that alarm list is actually being updated in this case. The problem is that in case of alarmed node brutal death, blocked publishers don't receive 'connection.unblocked' frame. Publishers connected after the moment of death don't have any problems.

So the problem is that 'connection.unblocked' is received only when alarm is explicitly cleared on live node.

michaelklishin:
@binarin also a possibility.

carlhoerberg:
According to my observations connections were still marked as "blocked" in the connections mgmt view, if that gives a clue or rules something out..

binarin:
There are 2 problems there:

1. on node_down rabbit_alarm notifies its subscribers with single notification with Source=[]. If instead notification will be sent for every type of alarm from down node, it will solve the case from original message.
2. rabbit_alarm doesn't send information to its subscribers about which node the alarm corresponds to. As a result I was able to find another bug: set disk alarm for 2 of 3 nodes, try to pubsish message and become blocked, and after clearing disk alarm only on one node it becames unblocked. But I think in this case the publisher should be unlocked until alarms an all nodes were cleared. Fixing this will require more work: rabbit_alarm subscription protocol should be updated with node information, and node information should be saved along with alarm type in rabbit_reader `#throttle{}`.

noahhaon:
I have also observed case 2 mentioned by @binarin above. :+1:

michaelklishin:
Please file a separate issue for case two.

noahhaon:
@michaelklishin done - #379

michaelklishin:
@binarin would you be interested in looking into a PR for case 1? We'll handle case 2 because it involves protocol method changes. Both are only reasonable for `3.6.0`.

michaelklishin:
OK, so maybe I've misread case 2. If it doesn't require a client-server protocol change, even better. However, it does change the internal protocol so can break earlier `3.5.x` releases in mixed clusters => still can only go into `3.6.0`.

binarin:
The more I look into rabbit_alarm, the messier it appears )

Actually, I think I want to solve both problems with one patch, because simple fix for the first issue will not be final.

I'll try to make a draft patch tommorow, so there will be something to discuss.

[...]

PR #395, opened 2015-10-27
Fix cluster-wide alarm handling

michaelklishin:
This is #381 by @binarin with updates for master. Still needs evaluation and QA.

From #381:

Initial draft for discussion.

 * Properly handle clearing alarms from dead nodes - #362.
 * Unblock publisher only when all alarms throughout the cluster was
   cleared - #379.
 * Disable/enable paging in rabbit_memory_monitor only in response
   to disk resource alarms from current node (I'm not totally sure, but I think this is a bug).
 * Add some documentation.


src/rabbit_alarm.erl
```diff
@@ -13,6 +13,16 @@
 %% The Initial Developer of the Original Code is GoPivotal, Inc.
 %% Copyright (c) 2007-2015 Pivotal Software, Inc.  All rights reserved.
 %%
+%% @doc There are two types of alarms handled by this module:
```
videlalvaro: as a nitpick, we don't use these kind of `@doc` tags

src/rabbit_memory_monitor.erl
```diff
@@ -86,7 +86,9 @@ report_ram_duration(Pid, QueueDuration) ->
 stop() ->
     gen_server2:cast(?SERVER, stop).
 
-conserve_resources(Pid, disk, Conserve) ->
+%% Paging should be enabled/disabled only in response to disk resource alarms
```
michaelklishin: This is worth discussing. If a node in a cluster runs out of disk space, we currently alarm all nodes. The reasoning behind this is fairly easy to guess: any client can connect to any node and publish in a way that would route messages to the alarmed node.

However, I can see how we may want to be more permissive in this particular case. Unlike running out of RAM, where
any incoming message can put us over the limit, with disk it's less likely to be the case.

binarin: I've understood it that way: when node senses that it has not enough RAM, it notifies all local queues to start pushing data to disk. So suppose we have 2 nodes in cluster that are actively paging data to disk (there is no alarms yet), and first node runs out of free space. All nodes are alarmed at that point and publishers are blocked. But it is safe to continue paging on the node without disk alarm, so we can still relieve RAM pressure there. It needs to be done anyway, and there is no reason to wait for clearing alarms on other nodes.

michaelklishin: It's not just paging that could use the disk. However, RabbitMQ probably tries to be too smart and too conservative at the moment: disk limit is never set to 0 in practice, so writing to disk is not necessarily going to make a node run out of space; we also cannot completely avoid running out of space because it can be another process that's been filling up the disk. The new behaviour is a reasonable compromise between not blocking every other node and not doing anything to avert the problem.

videlalvaro: We have to take into account credit_flow from channels to slaves, back to channels and what would happen if we change things here

michaelklishin: @videlalvaro we can't prevent a node from using disk at all, and even if we could it can't help if another process consumes most space. How exactly messages to flow to the alarmed node doesn't change this. Blocking *all* publishers across all nodes seems excessive to me given that we can't guarantee much as explained above.

videlalvaro: as discussed on slack with @dumbbell what would happen when node A that doesn't have an alarm, receives a publish that has to be routed or replicated to node B (who has a disk alarm for instance). What to do here?

michaelklishin: @videlalvaro currently we block all publishers. I can understand this for RAM alarms because anything that enters a node consumes RAM. For disk alarms this sounds like an overkill and we still cannot guarantee the node won't run out of disk.

michaelklishin: @binarin can you please revert the change that makes only one node alarmed? I failed to convince other team members, even though I completely agree it is a more reasonable behaviour. For 3.7 we would consider making this configurable, with a possible new mode: voluntary node shutdown.

binarin: > as discussed on slack with @dumbbell what would happen when node A that doesn't have an alarm, receives a publish that has to be routed or replicated to node B (who has a disk alarm for instance). What to do here?

Are we all discussing the same thing? ) 

There is no significant change in alarm behaviour, here is the picture with explanation what the guard expression on following line changes:

alarm-seq

michaelklishin: @binarin ok, thanks. I overlooked that `alarm_remote` callers remain the same.

michaelklishin:
OK, this has passed a basic review and I think the changes are reasonable. Now needs some functional QA.

PR #410, opened 2015-11-09
Fix metadata in OCF HA script

binarin:
Looks like copy-paste has gone wrong.


michaelklishin:
@bogdando I assume you folks communicate behind the scenes but JFYI :)

PR #411, opened 2015-11-09
Don't update cookie on every run of HA OCF script

binarin:
Update happens even during no-op commands like 'meta-data' or 'usage'.
During this update there is a short window for a race condition: a shell
redirection truncates the cookie file, and echo writes data there only
after a brief period of time. So erlang may read data from this empty
file and die with the error "Too short cookie string".

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -573,9 +573,22 @@ check_need_join_to() {
 
 # Update erlang cookie, if it has been specified
 update_cookie() {
+    local cookie_file_content
     if [ "${OCF_RESKEY_erlang_cookie}" != 'false' ] ; then
-        echo "${OCF_RESKEY_erlang_cookie}" > "${OCF_RESKEY_erlang_cookie_file}" && \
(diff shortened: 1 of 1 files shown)


scripts/rabbitmq-server-ha.ocf
```diff
@@ -573,9 +573,22 @@ check_need_join_to() {
 
 # Update erlang cookie, if it has been specified
 update_cookie() {
+    local cookie_file_content
     if [ "${OCF_RESKEY_erlang_cookie}" != 'false' ] ; then
-        echo "${OCF_RESKEY_erlang_cookie}" > "${OCF_RESKEY_erlang_cookie_file}" && \
-        chown ${OCF_RESKEY_username}:${OCF_RESKEY_groupname} "${OCF_RESKEY_erlang_cookie_file}" && \
+        if [ -f "${OCF_RESKEY_erlang_cookie_file}" ]; then
+            # First line of cookie file without newline
+            cookie_file_content=$(head -n1 "${OCF_RESKEY_erlang_cookie_file}" | perl -pe chomp)
```
mightydok: Maybe tr -d '\n' better for shell script.

binarin: perl is already used within the script, so I think it don't matter what to use

Issue #396, opened 2015-10-28
Distinct exit codes for rabbitmqctl, rabbitmq-plugins, etc


michaelklishin:
Per discussion with @dims.

It would be helpful to have distinctive (and standard, to the extent to which it is standard across various UNIX-y operating systems) exit codes in case an operation doesn't succeed. The goal is to give those provisioning RabbitMQ a way to retry on certain failures but not all. Currently our exit codes are `0` and `2` (from a few failures I could easily trigger).

This is post `3.6.0`. I think this can go into `3.6.x`: most users who check return codes likely check for `0` and `non-0` as they are random.

dims:
cc @binarin

binarin:
The only attempt to standartization of exit codes that I found is http://www.sbras.ru/cgi-bin/www/unix_help/unix-man?sysexits+3

In the following table I summarized those that I think are relevant. Is this list reasonable and complete?

 Numeric Value | sysexits (if any) | Description                                                       
---------------|-------------------|-------------------------------------------------------------------
             0 | EX_OK             | Self-explanatory                                                  
            64 | EX_USAGE          | Bad command-line args                                             
            65 | EX_DATAERR        | Some parameters do not make sense                                 
            69 | EX_UNAVAILABLE    | Failed to connect to node                                         
            75 | EX_TEMPFAIL       | Temporary failure (e.g. something timed out)                      
            70 | EX_SOFTWARE       | Any other error discovered when running command against live node

dims:
@binarin i see the same in http://fossies.org/dox/glibc-2.22/misc_2sysexits_8h.html so +1 from me to use some of the standard ones from that list

michaelklishin:
The error scenarios/categories make sense
to me.

michaelklishin:
Here's an additional [data point](http://tldp.org/LDP/abs/html/exitcodes.html).

carlhoerberg:
@binarin's suggestion looks good to me! Much better then looking for `/^Error: (.*)/` like we do today..

michaelklishin:
I believe #412 largely addresses this. Feel free to suggest/contribute smaller improvements. Better command line arguments framework will also greatly help make the exit codes more fine grained and sensible.

PR #412, opened 2015-11-09
Distinct exit-codes for CLI utilities

binarin:
Use exit codes from sysexits.h instead of just 0 and 2.

Do I need to write some tests for this issue?


michaelklishin:
It's unfortunate that before we move CLI tools to a sensible CLI library (read: Elixir), `EX_SOFTWARE` would be returned way more often. But from what I understand, #396 is mostly about separating failures such as timeouts and unavailable nodes.

Issue #264, opened 2015-08-12
NSIS installer silent install not supported in Windows upgrades


ryanzink:
I currently have a large number of servers that have v3.2.4 installed. I am trying to automate the upgrade process to v3.5.4 through a scripted install. I tried using the standard NSIS silent parameter by executing the following: 
`.\rabbitmq-server-3.5.4.exe /S`
However, the installer has not been configured to be fully silent... I am facing the following issues:
* The "RabbitMQ is already installed" dialog box appears even when the installer is run in silent mode. NSIS installers can be set up to have a default option for silent installs (examples are here: http://nsis.sourceforge.net/Examples/silent.nsi) to override the message box from appearing.
* Installing in silent mode as an upgrade between versions does not perform the same actions that a non-silent install does. I haven't performed an exhaustive analysis, but right away a major issue is that the RabbitMQ service does not get installed after the previous version is uninstalled.

michaelklishin:
How can we make the installer do what you want?

ryanzink:
To correct the dialog box, it appears the installer needs to have an /SD switch added to the MessageBox so the OK button is selected by default. There's documentation here: http://nsis.sourceforge.net/Docs//Chapter4.html#silent

As far as the service not getting installed, it looks like it is failing to install because the installer detects the environment variable ERLANG_HOME is not present. I see when the uninstall runs it removes ERLANG_HOME variable and when the install runs in non-silent mode it creates the ERLANG_HOME variable, but maybe there is a timing issue that prevents the environment variable from being created when it has been removed in the past second or so as the uninstall/reinstall occurs.

michaelklishin:
@ryanzink we do not know much about NSIS or Windows installer automation. If you'd contribute a PR that does the necessary changes, and a way for us to test it, we would be happy to include your work in 3.5.5.

michaelklishin:
Good point about `ERLANG_HOME`. To alleviate this, we need a way to make the installer aware that this is an upgrade and not an uninstall: is there a standard way to do that?

ryanzink:
@michaelklishin Unfortunately there isn't a standard way to do that. NSIS has a lot of limitations in that department. However I think passing parameters to the uninstall process could work. If there was a parameter that indicated the Uninstall was temporary, it could provide a workaround. I'll see what I can dig up.

ryanzink:
[...]

PR #421, opened 2015-11-13
Rabbitmq server 264

ryanzink:
Fixes to enable silent install using the NSIS /S command and reinstall without removing ERLANG_HOME, as this prevents successful reinstalls.


dumbbell:
References #264.

dumbbell:
As noted in #264, this patch was merged after squashing commits into one and fixing the style.

Thank you!

PR #424, opened 2015-11-16
Add host_ip field

mattymo:
Working with RMQ definitions via management plugin
requires knowing the IP address where it listens.

host_ip parameter will default to 127.0.0.1, but is
configurable.


bogdando:
There is related bug discovered in Fuel https://launchpad.net/bugs/1501731 and this patch addresses that issue.
+1 LGTM for me

Issue #429, opened 2015-11-18
refactor handling of acks from batch_publish_delivered


videlalvaro:
See https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_mirror_queue_sync.erl#L355

PR #430, opened 2015-11-19
Refactors AckTags handling for mirror_queue_sync

videlalvaro:
Needs to be merged together with: https://github.com/rabbitmq/rabbitmq-common/pull/17

Fixes: #429


videlalvaro:
@michaelklishin since this also change the BQ API, it should ship with 3.6.0

Relevant tests:

```bash
make tests FILTER=rabbit_priority_queue_test
make tests FILTER=eager_sync
```

PR #439, opened 2015-11-20
Fix memory reporting for connections

### LINKED ISSUE #435: Memory reporting for listeners is broken
Specifically, calling `rabbit:status().` will always show 0 for listeners.

The issue was introduced in https://github.com/rabbitmq/rabbitmq-server/pull/414/files#r45418469 - or rather the change was not made because of misunderstanding things.

essen:
Before, there was two parts of the supervision tree: one for the
listener/acceptor, and one for connections. The latter was one
big supervisor handling all connections, regardless of the
listener they were coming from.

With Ranch, each listener has its own branch in the supervision
tree. Inside it are then two more branches, one for the
acceptors and one for the connections.

The fix simply adds all Ranch connection supervisors to the list
by fetching them from the ranch_server ETS table.

Fixes #435.

### DIFF
--- a/src/rabbit_vm.erl
+++ b/src/rabbit_vm.erl
@@ -134,11 +134,10 @@ interesting_sups0() ->
     PluginProcs   = plugin_sups(),
     [MsgIndexProcs, MgmtDbProcs, PluginProcs].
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you!

PR #452, opened 2015-11-26
Fix RABBITMQ_CTL_ERL_ARGS variable in rabbitmq-plugins

### LINKED ISSUE #451: rabbitmq-plugins script reading from wrong env variable
Apparent mistake introduced by this commit on Jan 28:

https://github.com/rabbitmq/rabbitmq-server/commit/cc040e95d6ee7e0703547338848c2849b39e59ec#diff-8938c4358ef17da8f0f0f976f727f87c

The variable is supposed to be `RABBITMQ_CTL_ERL_ARGS`  but was added as `RABBITMQ_PLUGINS_ERL_ARGS` in `rabbitmq-plugins`. This variable is never processed in `rabbitmq-env` and thus will be empty.

gigablah:
Fixes #451.

### DIFF
--- a/scripts/rabbitmq-plugins
+++ b/scripts/rabbitmq-plugins
@@ -24,7 +24,7 @@ exec ${ERL_DIR}erl \
     -pa "${RABBITMQ_HOME}/ebin" \
     -noinput \
     -hidden \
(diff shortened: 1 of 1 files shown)


videlalvaro:
Merged, thanks!

PR #462, opened 2015-12-02
Support setting vm_memory_high_watermark and disk_free_limit in units

### LINKED ISSUE #448: Accept RAM and disk watermark values in megabytes
See #207 and #447. Using megabytes makes much more sense in 2015 but we need things to be consistent for RAM and disk watermarks.

Therefore I think RabbitMQ should support values such as `{absolute_in_mb, N}` for both disk and RAM watermarks. Values in bytes will remain for backwards compatibility.

hairyhum:
Fixes #448, per https://github.com/rabbitmq/rabbitmq-server/issues/448#issuecomment-161288678.

Not sure where to put parsing function and default values.

### DIFF
--- a/docs/rabbitmq.config.example
+++ b/docs/rabbitmq.config.example
@@ -196,15 +196,16 @@
    %%
    %% Or you can set absolute value using memory units.
    %%
-   %% {vm_memory_high_watermark, {absolute, "1024MB"}},
+   %% {vm_memory_high_watermark, {absolute, "1024M"}},
    %%
-   %% Memory units rules:
-   %% k, kiB - kibibytes (2^10)
-   %% M, MiB - mebibytes (2^20)
-   %% G, GiB - gibibytes (2^30)
-   %% kB - kilobytes (10^3)
-   %% MB - megabytes (10^6)
-   %% GB - gigabytes (10^9)
+   %% Supported units suffixes:
+   %%
+   %% k, kiB: kibibytes (2^10 bytes)
+   %% M, MiB: mebibytes (2^20)
+   %% G, GiB: gibibytes (2^30)
+   %% kB: kilobytes (10^3)
+   %% MB: megabytes (10^6)
+   %% GB: gigabytes (10^9)
 
    %% Fraction of the high watermark limit at which queues start to
    %% page message out to disc in order to free up memory.
--- a/docs/rabbitmq.config.example
+++ b/docs/rabbitmq.config.example
@@ -193,6 +193,18 @@
    %% Alternatively, we can set a limit (in bytes) of RAM used by the node.
    %%
    %% {vm_memory_high_watermark, {absolute, 1073741824}},
+   %%
+   %% Or you can set absolute value using memory units.
+   %%
+   %% {vm_memory_high_watermark, {absolute, "1024MB"}},
+   %%
+   %% Memory units rules:
+   %% k, kiB - kibibytes (2^10)
+   %% M, MiB - mebibytes (2^20)
+   %% G, GiB - gibibytes (2^30)
+   %% kB - kilobytes (10^3)
+   %% MB - megabytes (10^6)
+   %% GB - gigabytes (10^9)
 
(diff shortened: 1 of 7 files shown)


docs/rabbitmq.config.example
```diff
@@ -211,6 +223,10 @@
    %% listed above for more details.
    %%
    %% {disk_free_limit, 50000000},
+   %%
+   %% Or you can set it using memory units (same as in vm_memory_high_watermark)
+   %% {disk_free_limit, "50MB"},
+   %% {disk_free_limit, "50000kB"},
```
michaelklishin: I think it would make more sense to have an example in gigabytes.

src/rabbit_alarm.erl
```diff
@@ -86,9 +94,56 @@ start() ->
...
     rabbit_sup:start_delayed_restartable_child(
-      rabbit_disk_monitor, [DiskLimit]),
+      rabbit_disk_monitor, [parse_disk_limit(DiskLimit)]),
     ok.
 
+parse_mem_limit({absolute, Limit}) ->
+    case parse_limit(Limit) of
+        {ok, ParsedLimit} -> {absolute, ParsedLimit};
+        {error, parse_error} ->
+            rabbit_log:error("Unable to parse vm_memory_high_watermark value"),
```
michaelklishin: We should log the value that we failed to parse.

src/rabbit_alarm.erl
```diff
@@ -86,9 +94,56 @@ start() ->
...
+            ?DEFAULT_VM_MEMORY_HIGH_WATERMARK
+    end;
+    {absolute, parse_limit(Limit, ?DEFAULT_VM_MEMORY_HIGH_WATERMARK)};
+parse_mem_limit(Relative) when is_float(Relative), Relative < 1 ->
+    Relative;
+parse_mem_limit(_) ->
+    ?DEFAULT_VM_MEMORY_HIGH_WATERMARK.
+
+parse_disk_limit({mem_relative, Relative})
```
michaelklishin: There are similar functions in `rabbit_disk_monitor` — have you considered moving these there?

src/rabbit_alarm.erl
```diff
@@ -86,9 +94,56 @@ start() ->
...
+    case parse_limit(Limit) of
+        {ok, ParsedLimit} -> ParsedLimit;
+        {error, parse_error} ->
+            rabbit_log:error("Unable to parse disk_free_limit value"),
+            ?DEFAULT_DISK_FREE_LIMIT
+    end;
+parse_disk_limit(_) -> ?DEFAULT_DISK_FREE_LIMIT.
+
+parse_limit(MemLim) when is_integer(MemLim) -> {ok, MemLim};
+parse_limit(MemLim) when is_string(MemLim) ->
```
michaelklishin: It would be great to have a few unit tests for this function (in rabbitmq-test).

src/rabbit_alarm.erl
hairyhum: There are `interpret_limit` in both `rabbit_disk_monitor` and `vm_memory_monitor`, maybe it could be merged some way.

hairyhum: Then it would be reasonable to move parse_limit somewhere.

michaelklishin: @hairyhum `rabbit_monitor_misc`, similarly to `rabbit_misc` and `rabbit_control_misc`?

michaelklishin: Hm, maybe `rabbit_resource_monitor_misc` would be clearer, because it's resource monitoring that we are talking about, not, for instance, Erlang process monitoring.

hairyhum:
Also added tests for parse function 
https://github.com/rabbitmq/rabbitmq-test/commit/cd8c9a88c0c16b5e56d6e4fbf771b06451fdb31c

michaelklishin:
@hairyhum thank you, please submit a PR in that repo as well.

Issue #464, opened 2015-12-03
UNIX exit code when rabbitmq-server fails to load its config file is 0


Subito:
```
# /usr/local/sbin/rabbitmq-server
{"could not start kernel pid",application_controller,"error in config file \"/usr/local/etc/rabbitmq/rabbitmq.config\" (none): no ending <dot> found"}
# echo $?
0
```
That makes it unnecessary hard to detect if the startup worked correctly. If the exit-code could indicate an  error, the rc-scripts could respond to that and not just report it started all the time even if it could't even parse the config.

michaelklishin:
The failure above is, unfortunately, in the code that executes before RabbitMQ is started. Perhaps something can be improved in the shell script.

What version do you use?

michaelklishin:
Related: #396.

videlalvaro:
I think the problem might be related to this line: https://github.com/rabbitmq/rabbitmq-server/blob/master/scripts/rabbitmq-server#L164

We should store in a variable the last exit code and probably return that once the wait exits

hairyhum:
`wait` command can wait for specific background job only and will return exit code for it.

dumbbell:
This issue is fixed by #468.

michaelklishin:
Thank you, @hairyhum!

Subito:
Nice, thanks!

PR #467, opened 2015-12-03
Make wait command wait for start_rabbitmq_server command

hairyhum:
If `wait` is waiting for specific command instead of all commands - it returns return status. Related to https://github.com/rabbitmq/rabbitmq-server/issues/464


dumbbell:
Hi!

Looks good to me. Could you please resubmit your patch against the `stable` branch?

PR #478, opened 2015-12-09
Allow users to define poliies and parameters in OCF script for HA

dmitrymex:
By mistake OpenStack policies were copied into the initial version of
the script. The policies are removed by this change.

Instead users are provided with ability to set their own policies in
a new set_rabbitmq_policy.sh file. Also now users can extend
parameters list with their own ones. For now extended parameters can
be used only in the policy file, but later new places to hook into
could appear.

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -13,6 +13,9 @@
 #
 # See usage() function below for more details ...
 #
+# Note that the script uses set_rabbitmq_policy.sh script located in the
+# same directory to setup RabbitMQ policies.
+#
 #######################################################################
 # Initialization:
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
@dmitrymex do you intend this to be in `3.5.7`? Then please rebase against `stable`.

dmitrymex:
@michaelklishin, thank you for suggestion. Yep, I want that change to be in 3.5.7 and so I have ported that pull request to the stable branch - https://github.com/rabbitmq/rabbitmq-server/pull/480

michaelklishin:
Closing, re-submitted as #480.

dmitrymex:
@michaelklishin, a newbie question: will that change be ported to master by someone or how will it get to the next major release?

michaelklishin:
@dmitrymex `stable` is merged into `master` pretty much after every change.

dmitrymex:
@michaelklishin: got it, thank you

Issue #94, opened 2015-04-01
Pluggable logging backends


michaelklishin:
People keep asking for it. We should make Lager dead easy to use with RabbitMQ. Or even consider switching to it entirely.

willejs:
:+1:

willejs:
@michaelklishin are there plans to implement this any time soon? I am very interested in it!

michaelklishin:
@willejs we are pretty busy with other things. It should be very realistic to ship this with `3.6.0`.

michaelklishin:
@willejs that said, #128 is something we can finish by early may. Might give you some options.

willejs:
Cheers @michaelklishin its good enough that this is on the horizon :) . Adding this issue to the 3.6.0 milestone would be cool ;)

willejs:
@michaelklishin sorry to bother you! Is it on a roadmap? Is there anyway to speed this up, apart from learning erlang? ;) I'l donate to the project in any way if i can!

videlalvaro:
There's this already that can help a bit: https://github.com/hyperthunk/rabbitmq-lager I'm using it on a plugin that I'm building

michaelklishin:
@willejs this is not getting any attention yet for one reason: we are a tiny team. It's still on the roadmap for 3.6.0 and we'll see if we can squeeze it into the list of things for July.

willejs:
I understand, hence apologising! July would be great!

willejs:
@michaelklishin me again :) Hows this looking?

michaelklishin:
@willejs it will not be in 3.6. However, now that we've moved to erlang.mk for the build system, introducing new dependencies is trivial, so this becomes easier.

michaelklishin:
There were logging changes in Erlang/OTP 18.1, which 3.7 will likely require, that make me wonder if Lager is really necessary for this to work.

hairyhum:
Lager receives messages from `error_logger`, so it's possible to just start lager application on rabbitmq node and set logging `app_env` and logs will go to lager. (without lagers fancy formatting, which is parse_transform).
So in first approach we can just add log location to `rabbitmq.config` to lager application.

hairyhum:
Do anyone know why is there `append_file` and then `delete` instead of just `rename` in log rotation? https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_error_logger_file_h.erl#L170

michaelklishin:
[...]

PR #491, opened 2015-12-14
Switch to Lager for logging

hairyhum:
Tried to make lager work with all default logging locations and settings.
Log rotation have 2 sec sleep now to let lager finish all writes to old file and open a new one.
Support for configuring lager from `.config`, so it's not overwritten by default rabbit config.
Need review


hairyhum:
Related to #94

hairyhum:
Spend most time trying to make tests with old config working too support same behaviour. 
It should work like nothing is changed. Same log files, same config. But there is a way to customise it with lager `.config` 
Next steps could be better integration with `parse_transform` or adding more sinks for logging different things separately. But this could be a separate tasks.

michaelklishin:
Erlang unit tests fail with

````
OUT=$(RABBITMQ_PID_FILE='/var/folders/6v/nzn6gr5d6k31r39x57fddtpm0000gp/T//rabbitmq-test-instances/rabbit/rabbit.pid' echo "rabbit_tests:all_tests()." | erl_call -sname rabbit -e) ; \
	  echo $OUT ; echo $OUT | grep '^{ok, passed}$' > /dev/null
{ok, "Tests failed
Error: {error, {badmatch,[true,{error,enoent}]}}
Stack trace:
[{rabbit_tests,test_log_management,0,
 [{file,"test/src/rabbit_tests.erl"},{line,742}]},
 {rabbit_tests,all_tests0,0,[{file,"test/src/rabbit_tests.erl"},{line,63}]},
 {rabbit_tests,all_tests,0,[{file,"test/src/rabbit_tests.erl"},{line,34}]},
 {erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,669}]},
 {lib,eval_str,1,[{file,"lib.erl"},{line,106}]},
 {rpc,'-handle_call_call/6-fun-0-',5,[{file,"rpc.erl"},{line,206}]}]
"}
make[1]: *** [run-tests] Error 1
```

Other test suites pass.

hairyhum:
Created pull request for test branch

docs/rabbitmq.config.example
```diff
@@ -621,5 +621,20 @@
    %% Set queries to determine which tags a user has
    %%
    %% {tag_queries, []}
+  ]},
+  {lager, [
+   %% Lager application config to control logs.
+   %% See https://github.com/basho/lager for more documentation
+   %% Log direcrory. Loads authomaticaly from RABBITMQ_LOG_BASE variable.
```
videlalvaro: it should be `automatically`

docs/rabbitmq.config.example
```diff
@@ -621,5 +621,20 @@
...
+   %% {log_root, "RABBITMQ_LOG_BASE"},
+   %% Sasl error log location is in default handlers.
```
videlalvaro: It should be SASL as in the rest of the file

docs/rabbitmq.config.example
```diff
@@ -621,5 +621,20 @@
...
+   %% By default use file handler "rabbitmq.log"
+   %% {handlers, [
+   %%   {lager_file_backend, [{file, "rabbitmq-sasl.log"}, {level, debug}, {date, ""}, {size, 0}]}
+   %% ]},
+   %% Rabbit log should be defined with rabbitmq_lager_event sink
```
videlalvaro: it should be RabbitMQ as in the rest of the file

src/rabbit.erl
```diff
@@ -566,75 +614,135 @@ insert_default_data() ->
...
+                                                                 tty)),
+            SaslHandlers = lager_handlers(application:get_env(rabbit, 
+                                                              sasl_error_logger, 
+                                                              tty)),
+            Sinks = [
+                {?LAGER_SINK, [{handlers, DefaultHandlers}]}
+                 % TODO Waiting for PR https://github.com/basho/lager/pull/303
+                 % ,{error_logger_lager_event, [{handlers, SaslHandlers}]}
+                ],
+                Handlers = SaslHandlers,
```
videlalvaro: what's the purpose of the `Handlers` variables?  Why not just use `SaslHandlers`?

src/rabbit.erl
```diff
@@ -566,75 +614,135 @@ insert_default_data() ->
...
+        kernel ->
+            proplists:get_value(handlers, 
+                proplists:get_value(?LAGER_SINK, 
+                    application:get_env(lager, extra_sinks, [])));
+        sasl ->
+            application:get_env(lager, handlers, undefined)
+    end,
+    case LagerHandlers of
+        undefined -> 
+            throw({error, {cannot_log_to_file, unknown, {lager_handlers_undefined, Type}}, application:get_env(rabbit, sasl_error_logger)});
```
videlalvaro: Lines should be of 80 chars max

michaelklishin:
`make full` fails for me (Java test suites fail, looks like node startup issues) while they all pass when executed individually.

hairyhum:
@michaelklishin I don't have any errors with `make full`. I have tested at MacOS El Capitan and OTP 18

scripts/rabbitmq-server
```diff
@@ -118,6 +118,7 @@ start_rabbitmq_server() {
         -sasl sasl_error_logger "$SASL_ERROR_LOGGER" \
         -rabbit error_logger "$RABBIT_ERROR_LOGGER" \
         -rabbit sasl_error_logger "$RABBIT_SASL_ERROR_LOGGER" \
+        -rabbit log_base "\"$RABBITMQ_LOG_BASE\"" \
```
michaelklishin: if we add this in `rabbitmq-server`, we need to do the same in `rabbitmq-server.bat`.

src/rabbit.erl
```diff
@@ -442,15 +450,52 @@ environment(App) ->
...
-    rabbit_log:info("Rotating logs with suffix '~s'~n", [Suffix]),
-    log_rotation_result(rotate_logs(log_location(kernel),
-                                    Suffix,
-                                    rabbit_error_logger_file_h),
-                        rotate_logs(log_location(sasl),
-                                    Suffix,
-                                    rabbit_sasl_report_file_h)).
+    ok = rotate_lager_handlers(Suffix, application:get_env(lager, handlers)),
+    ok = rotate_lager_sinks(Suffix),
+    timer:sleep(2000),
```
dumbbell: Why is this `timer:sleep/1` needed? And why 2 seconds?

hairyhum: Sleep wait for lager to switch to new files. Nasty solution, but log rotation are called only from rabbitmqctl (should be deprecated maybe?) 
Lage internal log rotation is much nicer because support file handle management and multiple rotated files, size, date and other settings.

michaelklishin: `rabbitmqctl` rotation is used by Linux packages. I don't think we should deprecate it.

src/rabbit.erl
```diff
@@ -442,15 +448,53 @@ environment(App) ->
...
-    log_rotation_result(rotate_logs(log_location(kernel),
-                                    Suffix,
-                                    rabbit_error_logger_file_h),
-                        rotate_logs(log_location(sasl),
-                                    Suffix,
-                                    rabbit_sasl_report_file_h)).
+    ok = rotate_lager_handlers(Suffix, application:get_env(lager, handlers)),
+    ok = rotate_lager_sinks(Suffix),
+    %% give Lager a moment to perform rotation
+    timer:sleep(2000),
```
dumbbell: Why is this `timer:sleep/1` needed? And why 2 seconds?

hairyhum: -- Comment from outdated diff --
Sleep wait for lager to switch to new files. Nasty solution, but log rotation are called only from rabbitmqctl (should be deprecated maybe?) 
Function itself required for backwards compatibility because used in logrotate scripts in Linux packages. 
Lager internal log rotation is much nicer because support file handle management and multiple rotated files, size, date and other settings. Maybe we should recommend to use it instead?

dumbbell: Isn't there a real way to wait for the actual completion of the rotation instead of using a random timeout?

About the relevance of `rabbitmqctl rotate_logs`, can we change it so it drives lager instead of doing its own rotation? I agree that what RabbitMQ does is limited. Lager must provide a much nicer implementation.

dumbbell:
I tested the branch on Windows and it works fine with the default settings, including log rotation.

However, logging to stdout is broken. You can test with:
```
make run-broker RABBITMQ_LOGS=-
```
In `rabbitmq-server`, this becomes:
```
-rabbit error_logger tty
```

Lager complains with the following message:
```
WARNING: Lager's console backend is incompatible with the 'old' shell, not enabling it
```

Note that this depends on a patch I didn't commit yet (`rabbitmq-server.bat` doesn't support `RABBITMQ_LOGS=-`).

This could be specific to Windows though. I will test on FreeBSD once the profiling of lazy queues is finished (erlgrind is already running for almost three hours...)

michaelklishin:
Fixing TTY logging is the only thing that's left, it seems.

dumbbell:
I'm starting to look at this.

dumbbell:
The problem with TTY is only present on Windows. It works as expected on Unix.

dumbbell:
Apparently, this is a known limitation:
https://github.com/basho/lager/blob/master/src/lager_console_backend.erl#L55
https://github.com/basho/lager/blob/master/src/lager_console_backend.erl#L137

It's not a regression as it was impossible to enable console logging on Windows anyway.

I'm testing a few more things but this branchs looks ready to be merged. My concern about the `timer:sleep/1` can be addressed later.

dumbbell:
We can't override the log file locations anymore using the environment variables, if lager is configured in `rabbitmq.config`. Those variables are only considered as the default values. And indeed, with the current `rabbitmq-env` and `rabbitmq-server`, we can't distinguish the default values from user-specified ones. I'm looking for a solution.

hairyhum:
@dumbbell priority of `rabbitmq.config` parameters was selected on purpose, because it is much more flexible way to configure file locations, log levels and rotating.

dumbbell:
I agree it's more flexible and powerful. However, I find it handy if I can just set `RABBITMQ_LOGS` and have this one time log file. Another usecase : I have my complex logging preferences in `rabbitmq.config` but from time to time, I want RabbitMQ sends everything to stdout.

dumbbell:
I still have a few things I would like to fix before merging this branch:
* Lager's documentation recommends to use a parse_transform, `{parse_transform, lager_transform}`, during the build.
* When a crash is logged, the message is even more unreadable than with standard error_logger format, except on stdout:
 ```
# In rabbit-sasl.log
2016-01-06 12:01:41.341 [error] <0.227.0> CRASH REPORT Process <0.227.0> with 0 neighbours exited with reason: no function clause matching rabbit_memory_monitor:init([]) line 113 in gen_server2:init_it/6 line 601
2016-01-06 12:01:41.342 [error] <0.226.0> Supervisor rabbit_memory_monitor_sup had child rabbit_memory_monitor started with rabbit_memory_monitor:start_link() at undefined exit with reason no function clause matching rabbit_memory_monitor:init([]) line 113 in context start_error
2016-01-06 12:01:41.343 [error] <0.161.0> CRASH REPORT Process <0.161.0> with 0 neighbours exited with reason: {error,{{shutdown,{failed_to_start_child,rabbit_memory_monitor,{function_clause,[{rabbit_memory_monitor,init,[[]],[{file,"src/rabbit_memory_monitor.erl"},{line,113}]},{gen_server2,init_it,6,[{file,"src/gen_server2.erl"},{line,559}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,240}]}]}}},{child,undefined,rabbit_memory_monitor_sup,{rabbit_restartable_sup,start_link,[rabbit_memory_monitor_sup,{rabbit_memory_monitor,start_link,[]},false]},transient,infinity,supervisor,[rabbit_restartable_sup]}}} in application_master:init/4 line 134
2016-01-06 12:01:41.343 [info] <0.7.0> Application rabbit exited with reason: {error,{{shutdown,{failed_to_start_child,rabbit_memory_monitor,{function_clause,[{rabbit_memory_monitor,init,[[]],[{file,"src/rabbit_memory_monitor.erl"},{line,113}]},{gen_server2,init_it,6,[{file,"src/gen_server2.erl"},{line,559}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,240}]}]}}},{child,undefined,rabbit_memory_monitor_sup,{rabbit_restartable_sup,start_link,[rabbit_memory_monitor_sup,{rabbit_memory_monitor,start_link,[]},false]},transient,infinity,supervisor,[rabbit_restartable_sup]}}}
```
 ```
# In rabbit.log
2016-01-06 12:01:41.351 [info] <0.2.0> Error description:
   {could_not_start,rabbit,{error,{{shutdown,{failed_to_start_child,rabbit_memory_monitor,{function_clause,[{rabbit_memory_monitor,init,[[]],[{file,"src/rabbit_memory_monitor.erl"},{line,113}]},{gen_server2,init_it,6,[{file,"src/gen_server2.erl"},{line,559}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,240}]}]}}},{child,undefined,rabbit_memory_monitor_sup,{rabbit_restartable_sup,start_link,[rabbit_memory_monitor_sup,{rabbit_memory_monitor,start_link,[]},false]},transient,infinity,supervisor,[rabbit_restartable_sup]}}}}

Log files (may contain more information):
       rabbit.log
       /tmp/rabbitmq-test-instances/rabbit/log/rabbit-sasl.log
```
 ```
# On stdout
BOOT FAILED
===========

Error description:
       {could_not_start,rabbit,
           {error,
               {{shutdown,
                    {failed_to_start_child,rabbit_memory_monitor,
                        {function_clause,
                            [{rabbit_memory_monitor,init,
                                 [[]],
                                 [{file,"src/rabbit_memory_monitor.erl"},
                                  {line,113}]},
                             {gen_server2,init_it,6,
                                 [{file,"src/gen_server2.erl"},{line,559}]},
                             {proc_lib,init_p_do_apply,3,
                                 [{file,"proc_lib.erl"},{line,240}]}]}}},
                {child,undefined,rabbit_memory_monitor_sup,
                    {rabbit_restartable_sup,start_link,
                        [rabbit_memory_monitor_sup,
                         {rabbit_memory_monitor,start_link,[]},
                         false]},
                    transient,infinity,supervisor,
                    [rabbit_restartable_sup]}}}}

Log files (may contain more information):
       rabbit.log
       /tmp/rabbitmq-test-instances/rabbit/log/rabbit-sasl.log

12:01:41.351 [info] Error description:
       {could_not_start,rabbit,{error,{{shutdown,{failed_to_start_child,rabbit_memory_monitor,{function_clause,[{rabbit_memory_monitor,init,[[]],[{file,"src/rabbit_memory_monitor.erl"},{line,113}]},{gen_server2,init_it,6,[{file,"src/gen_server2.erl"},{line,559}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,240}]}]}}},{child,undefined,rabbit_memory_monitor_sup,{rabbit_restartable_sup,start_link,[rabbit_memory_monitor_sup,{rabbit_memory_monitor,start_link,[]},false]},transient,infinity,supervisor,[rabbit_restartable_sup]}}}}

Log files (may contain more information):
       rabbit.log
       /tmp/rabbitmq-test-instances/rabbit/log/rabbit-sasl.log
```

* As shown above, the path to `rabbit.log` is missing.

**Comment by dumbbell** [comment]
Also, I'm wondering if having two separate log files is still relevant. With the default settings, interesting messages are spread accross both `rabbit.log` and `rabbit-sasl.log`:

```
# In rabbit.log
2016-01-06 11:54:45.332 [info] <0.173.0> Disk free limit set to 50MB
2016-01-06 11:54:45.338 [info] <0.180.0> FHC read buffering:  OFF
FHC write buffering: ON
2016-01-06 11:54:45.347 [info] <0.162.0> Priority queues enabled, real BQ is rabbit_variable_queue
2016-01-06 11:54:45.361 [info] <0.235.0> msg_store_transient: using rabbit_msg_store_ets_index to provide index
2016-01-06 11:54:45.364 [info] <0.238.0> msg_store_persistent: using rabbit_msg_store_ets_index to provide index
2016-01-06 11:54:45.379 [info] <0.2.0> Server startup complete; 0 plugins started.
```
```
# In rabbit-sasl.log
2016-01-06 11:54:45.317 [info] <0.167.0> Memory limit set to 953MB of 15897MB total.
2016-01-06 11:54:45.337 [info] <0.179.0> Limiting to approx 470438 file handles (423392 sockets)
2016-01-06 11:54:45.372 [info] <0.263.0> started TCP Listener on [::]:5672
2016-01-06 11:54:45.372 [info] <0.269.0> started TCP Listener on 0.0.0.0:5672
```

Should be use a single file? What do you think?

**Comment by michaelklishin** [comment]
@dumbbell good point. Judging from the above, it makes sense to use a single file. Now that we use Lager and have more flexibility, I like this idea.

**Comment by essen** [comment]
The parse_transform is only useful so that lager can include file and line numbers automatically. This implies that you call lager:error/info/.. functions directly in the source files where you log. This doesn't seem at all compatible with the direction the PR has taken. The parse_transform will probably not do anything, considering you don't call those functions.

**Comment by dumbbell** [comment]
@essen: Yeah, that's what I understand too. But it's harmless, so I believe we should add that now instead of fighting later to understand why a specific feature doesn't work.

In the future, we should probably move to a tighter integration of Lager and use its features to have even better logs.

**Comment by dumbbell** [comment]
About one vs. two files, Lager creates an additional file, `$log_root/log/crash.log` by default, where it stores the "`=CRASH REPORT====`" messages.

So in the current state, messages logged through `rabbit_log` goes to `rabbit.log` (using an extra sink) and those logged through `error_logger` goes to `rabbit-sasl.log` (using the standard Lager sink).

I believe it should simplify log consumption and Lager configuration if we just used the standard sink and let Lager store larger sasl logs to `crash.log`. That was the purpose of `rabbit-sasl.log`, but we now don't need to duplicate Lager's job.

@hairyhum: What do you think about using a single `rabbit.log` through the standard sink?

**Comment by michaelklishin** [comment]
:+1: from me.

**Comment by dumbbell** [comment]
The more I think about this, the more I believe we should go the extra mile and drop `rabbit_log`.

`rabbit_log` was nice to:
* drop messages below a minimum level
* filter by category (though we don't use it a lot).

Lager does both and in a better way.

IMHO, we should use `lager:{info,warning,error}()` directly right now. The benefits are:
* `$file:$line` is included in the message
* We can take this occasion to add a category to all messages

Now, whatever we choose, I'm not entirely satisfied with Lager's default behavior w.r.t. crash logging for messages emitted by Erlang (eg. by `supervisor` or `gen_server`). For example, if a queue crashes (I added an incorrect assertion in `a()`), 31 lines of totally unreadable (for an Erlang developer) mix of text, stacktraces and Erlang terms are written to the log file or stdout. For an end-user, replace "unreadable" by "horrifying" :-)

I don't know yet what we can do about this. The same information is available in `crash.log` in a saner format. Do we have a way to either ignore or reformat those messages in the standard sink? Many users don't know how to parse those Erlang terms/stacktraces, so they are mostly useful to us.

@hairyhum: Do you have an advice?

**Comment by hairyhum** [comment]
Messages like `CRASH REPORT` are formatted by lager here: https://github.com/rabbitmq/lager/blob/master/src/error_logger_lager_h.erl#L256
We could add some custom formatting, but I guess then we will have to make it pluggable and make PR to lager upstream.
Ignoring should be easier by just removing backend, but it will make lager ignore all other error_logger messages.

**Comment by dumbbell** [comment]
I see, so not what we want. Anyway, it's fine for now. In my example, the queue fails to restart, that's why there are in fact multiple crashes logged in a row. This is an extreme (and hopefully rare) situation.

**Comment by dumbbell** [comment]
@hairyhum, @michaelklishin: What do you think about deprecating `rabbit_log` and using the standard sink only through Lager API directly?

**Comment by hairyhum** [comment]
Or as an option we can configure `rabbit_log_lager_event` sink and enable `lager_transform` so lager will  send `rabbit_log:error|debug|info` etc to this sink. At least that's what lager docs say.

**Comment by michaelklishin** [comment]
Using Lager directly will break a lot of plugins. Lets hold off on that and do it for 3.8.

**Comment by michaelklishin** [comment]
As for deprecating `rabbit_log`, plugin authors may find it convenient to have this known way of logging to the broker log file.

**Comment by dumbbell** [comment]
@hairyhum: That's interesting! I will look at that.
@michaelklishin: By deprecating, I mean stop using it internally (and in our plugins). The module can stay (and should probably be moved to `rabbit_common`). It would become a wrapper above Lager API / standard sink.

**Comment by michaelklishin** [comment]
@dumbbell no objections to that.

**Comment by hairyhum** [comment]
@dumbbell it's mentioned in sinks doc section https://github.com/rabbitmq/lager#sinks
It won't support logging with categories, so could be unsafe if categories are used in some plugin code.

**Comment by dumbbell** [comment]
Here is what I propose to fully use Lager and its features, while still maintaining compatibility for those using `rabbit_log`.

* Let's use the default sink to log everything by default. For the end-user, it's easy to configure, because they only have to mess with the main list of handlers:

 ```erlang
% In rabbitmq.config
  {lager, [
      {handlers, [
          {lager_console_backend, [info]}
        ]}
    ]}
```
* By default (ie. if nothing is configured in `rabbitmq.config`), we honor `$RABBITMQ_LOGS`: a handler based on `$RABBITMQ_LOGS` is added to the default sink. If there is something configured, but `$RABBITMQ_LOGS` is forced on the command line (or `rabbitmq-env.conf`), the configured handlers are replaced with a handler based on `$RABBITMQ_LOGS`. This is mainly useful to developers when they want to temporarily redirect messages to stdout by starting `make run-broker RABBITMQ_LOGS=-` for instance.
* We configure an extra sinks (except if it's already configured) called `rabbit_log_lager_event`. In combination with `lager_transform` and the `+'{lager_extra_sinks,[rabbit_log]}` compile option, this makes all calls to `rabbit_log:<level>()` to be transformed into direct calls to Lager. The advantage is that the logged message can contain metadata such as file name and line number.
* To maintain compatibility, the `rabbit_log` module still exports those `<level>()` functions: they call Lager, specifying `rabbit_log_lager_event` as the sink. The result is the same, minus the automatically added metadata.
* I added a `lager_forwarder_backend` module: it acts as a Lager backend which forwards all messages to a specified sink. The goal is to use it for `rabbit_log_lager_event`: if this sink is not configured, a `lager_forwarder_backend` handler is added to this sink and this handler is configured to forward everything to the default Lager sink. The end result is that messages logged with `rabbit_log:<level>()` are written to the same log file than messages logged with error_logger. This was necessary because Lager doesn't allow to have multiple `lager_file_backend` backends to point to the same file.

Let's try to sum up this with a simple drawing of the default configuration:
```
rabbit_log:info() --> rabbit_log_lager_event sink --+
                                                    |
                                          +---------+
                                          |
                                          V
error_logger:info_msg() --> lager_event default sink --> handler based on $RABBITMQ_LOGS
```

Now, about categories. `rabbit_log:log/{3,4}` offers the possibility to specify a category. An admin could then configure per-category minimum log levels. I propose we leverage on extra sinks even further for this purpose:
* `+'{lager_extra_sinks,[rabbit_log]}` becomes `+'{lager_extra_sinks,[rabbit_log,rabbit_connection,rabbit_channel]}` if we take two categories we currentl support, "connection" and "channel".
* This means we can use `rabbit_channel:info()` to log a message to the `rabbit_channel_lager_event` sink.
* Category-specific sinks are configured exactly like `rabbit_log_lager_event`. To be exact, I take the list of extra sinks from the module's compilation options stored in the beam:

 ```erlang
(rabbit@magellan)1> rabbit_lager:module_info(compile).
[{options,[
           ...
           {lager_extra_sinks,[rabbit_log,rabbit_connection,
                               rabbit_channel]},
```
 They are all treated the same.
* This still allows the user to specify per-category minimum log level:

 ```erlang
      {extra_sinks, [
          {rabbit_channel_lager_event, [
              {handlers, [
                  % Do not send messages to $RABBITMQ_LOGS.
                  {lager_forwarder_backend, [lager_event, none]},
                  % Send messages to stdout.
                  {lager_console_backend, info}
                ]}
            ]}
        ]},
```

Let's expand the previous drawing:
```
rabbit_connection:info() --> rabbit_connection_lager_event sink --+
rabbit_channel:info() --> rabbit_channel_lager_event sink --+     |
rabbit_log:info() --> rabbit_log_lager_event sink --+       |     |
                                                    |       |     |
                                          +---------+-------+-----+
                                          |
                                          V
error_logger:info_msg() --> lager_event default sink --> handler based on $RABBITMQ_LOGS
```

Admittedly, this is more complicated to configure than what we have currently, but it's also way more powerful. Lager allows finer-grain control of the log levels: instead of a bare minimum, it supports masks and ranges.

michaelklishin:
So the extra sinks would be used to allow for finer-grained configuration of things we currently allow to configure. That sounds reasonable to me. As does respecting `RABBITMQ_*`.

Should `lager_forwarder_backend` be `rabbit_lager_forwarder_backend`, though, because it is RabbitMQ-specific?

dumbbell:
> Should `lager_forwarder_backend` be `rabbit_lager_forwarder_backend`, though, because it is RabbitMQ-specific?

There is nothing RabbitMQ-specific in it and currently, it lives in rabbitmq-common (where rabbit_log and rabbit_lager should probably go).

michaelklishin:
Fair enough.

dumbbell:
I just pushed the proposed patch so you can read the code and test if you want.

michaelklishin:
I have one test failure:

```
{ok, "Tests failed
Error: {error, undef}
Stack trace:
[{rabbit,log_location,[kernel],[]},
 {rabbit_tests,test_log_management,0,
 [{file,"test/src/rabbit_tests.erl"},{line,732}]},
 {rabbit_tests,all_tests0,0,[{file,"test/src/rabbit_tests.erl"},{line,63}]},
 {rabbit_tests,all_tests,0,[{file,"test/src/rabbit_tests.erl"},{line,34}]},
 {erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,669}]},
 {lib,eval_str,1,[{file,"lib.erl"},{line,106}]},
 {rpc,'-handle_call_call/6-fun-0-',5,[{file,"rpc.erl"},{line,206}]}]
"}
make[1]: *** [run-tests] Error 1
```

(with rabbitmq-test switched to `rabbitmq-server-94`, of course).

dumbbell:
I didn't update the testsuite yet.

dumbbell:
That said, I should commit `lager_forwarder_backend` if you actually want to test :-)

michaelklishin:
@dumbbell I still have the test failure but otherwise no objections to your plan.

dumbbell:
Thanks, I will update the testsuite, then.

dumbbell:
The testsuite is up-to-date so this issue is ready for a final test :-)

michaelklishin:
With a few more tweaks in the `rabbitmq-test` counterpart to this, Lager now looks good to me. I'd be OK with merging this now and updating plugins (primarily management and plugins that extend it) later as needed.

We should also cut a milestone release after the dust from this change settles.

PR #492, opened 2015-12-14
HIPE compilation moved to rabbit_hipe module

### LINKED ISSUE #484: Extract HiPE-related bits into rabbit_hipe
They currently live in `rabbit`, which is one of the modules that could use some splitting into parts.

hairyhum:
Fixes #484


michaelklishin:
Thank you, @hairyhum. We'll leave this for after `3.6.0` (can be `3.6.x`, perhaps).

PR #495, opened 2015-12-14
Install set_rabbitmq_policy.sh as an example script

### LINKED ISSUE #494: Install set_rabbitmq_policy.sh as an example
Currently, `set_rabbitmq_policy.sh` is installed as is: this will overwrite any local modifications. This file should be controled by the user so we must not do that.

Also, there is no need to install it with the execution bit set, as this script is sourced by `rabbitmq-server-ha.ocf`.

dumbbell:
Otherwise, it will overwrite any local modifications.

Fixes #494.


michaelklishin:
@dmitrymex @bogdando any thoughts on this?

dumbbell:
In the near future, I plan to use ucf in the Debian package to manage user-controled files, so the default `set_rabbitmq-policy.sh` will be installed as is on a fresh install. There is probably a similar tool for RPM, I didn't look at it yet.

bogdando:
Looks valid if we want no mirroring queues out of box, though this would have docs impact like: "in order to enable mirroring queues, copy this example file as ..."

dmitrymex:
@michaelklishin : yep, sounds absolutely reasonable to me
@dumbbell, thank you for your fix!

dumbbell:
@bogdando: The default script is empty so I don't see why it's needed to have mirrored queues. Do I miss something?

bogdando:
right, I missed it has all lines commented out :)

PR #510, opened 2015-12-22
Improve error message of rabbitmqctl reset when the node is running

### LINKED ISSUE #167: Improve error message of rabbitmqctl reset when the node is running
We currently have

```
sbin/rabbitmqctl reset
Resetting node hare@mercurio ...
Error: mnesia_unexpectedly_running
```

which could have been clearer.

hairyhum:
Fixes #167 

Maybe something like this could be added to other commands requiring `stop_app`?


michaelklishin:
Feel free to make similar improvements in more places. There aren't that many commands that require RabbitMQ to be stopped, though.

PR #512, opened 2015-12-22
DO NOT MERGE Option to ignore message store file delete errors

### LINKED ISSUE #75: Message store file GC should be more resilient to file deletion failures
See [this mailing list thread](https://groups.google.com/d/msg/rabbitmq-users/2dqZuYc_jRc/ESlLscnqCtwJ).

hairyhum:
Fixes #75, WIP.

### DIFF
--- a/src/rabbit_msg_store.erl
+++ b/src/rabbit_msg_store.erl
@@ -1414,7 +1414,7 @@ safe_file_delete(File, Dir, FileHandlesEts, DiskErrorStrategy) ->
     %% indicates the client has really closed the file.
     case ets:match_object(FileHandlesEts, {{'_', File}, '_'}, 1) of
         {[_|_], _Cont} -> false;
-        _              -> 
+        _              ->
             FileName = form_filename(Dir, filenum_to_name(File)),
             case file:delete(FileName) of
                 ok -> ok;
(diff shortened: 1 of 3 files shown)


src/rabbit_msg_store.erl
```diff
@@ -1400,17 +1404,36 @@ mark_handle_to_close(ClientRefs, FileHandlesEts, File, Invoke) ->
...
+            FileName = form_filename(Dir, filenum_to_name(File)),
+            case file:delete(FileName) of
+                ok -> ok;
+                {error, DeleteErr} ->
+                    rabbit_log:error(
+                        "Unable to delete message store file ~p, reason: ~p", 
+                        [FileName, DeleteErr]),
+                    case DiskErrorStrategy of
+                        ignore -> ok;
+                        crash -> throw({error, DeleteErr});
```
michaelklishin: Are you sure this will shut down the node instead of just the message store? The goal is to make nodes voluntarily stop/leave the cluster if there's a possibility of disk corruption.

hairyhum:
Will it be enough to just stop node? Because I guess it could be restarted again and cause trouble if in cluster. 
So there is an option to leave cluster (if clustered) and stop the node. 
It still require some tests.

michaelklishin:
Leaving the cluster and stopping is what I had in mind.
We considered doing the same as an option for disk alarms after 3.6.0 ships.

I think having a designated named "shutdown coordinator" process that would handle it is the right
way to go about it.

hairyhum:
I like idea of separate coordinator process, but could it be dangerous to allow node to continue working during disk errors while this process detect or collect errors?

michaelklishin:
This feature is mean to be opt-in: you can tell the node to shut down when it encounters a local disk I/O error. Simply terminating message store process doesn't do that, in fact, it worse than what we have today (and will have by default in the future).

michaelklishin:
As described in https://github.com/rabbitmq/rabbitmq-server/issues/75, this PR is a WIP and should not be merged.

michaelklishin:
I am closing this as it misinterprets the idea I had in mind for #75 and now fails to merge anyway.

PR #524, opened 2015-12-30
Ensure rabbit node uptime is reset in the CIB for OCF resource

bogdando:
* Add ocf_run wrappers and info log messages for CIB attribute events
* Move "fast" CIB attribute updates before "heavy" operations like
  start/stop/wait to ensure CIB consistent even if the timeouts
  exceeded for the ops
* Delete master and start time attributes from CIB on action_start
  to ensure the correct rabbit nodes uptime evaluation for new
  master elections for corresponding pacemaker resources
* For post-demote notify and action_demote() delete the master
  attribute from CIB as well.
* For post-start notify, update the start time in the CIB even when
  the node is already clustered. Otherwise it would remain running
  in cluster w/o the start time registered, which affects the new
  master elections badly.

Related Fuel bug https://bugs.launchpad.net/fuel/+bug/1530150

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -475,10 +475,12 @@ now() {
 }
 
 master_score() {
+    local LH="${LL} master_score():"
     local score=$1
     if [ -z $score ] ; then
         score=0
     fi
+    ocf_log info "${LH} Updating master score attribute with ${score}"
     ocf_run crm_master -l reboot -v $score || return $OCF_ERR_GENERIC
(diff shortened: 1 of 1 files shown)


bogdando:
tested the same way, vagrant box...

michaelklishin:
Is this intended to be for `master`?

bogdando:
Misplaced sorry! Should be a stable

bogdando:
Resubmitted to stable as https://github.com/rabbitmq/rabbitmq-server/pull/525

PR #526, opened 2015-12-31
Improve diagnostics in 'rabbitmq-server' script

binarin:
While errors are detected with '-e' shell option in this script,
diagnostics messages leave a lot to be desired.

E.g. when trying to write pid file to full partition, the only message
in log is:

    sh: echo: I/O error

Which is definitely insufficient


michaelklishin:
@binarin is this meant only for master?

binarin:
@michaelklishin  Is 'master' going to become 3.6.1? Then it will suffice.

michaelklishin:
No, master will become `3.7.0` in months. `stable` is used for point releases.

PR #527, opened 2015-12-31
Improve diagnostics in 'rabbitmq-server' script

binarin:
While errors are detected with '-e' shell option in this script,
diagnostics messages leave a lot to be desired.

E.g. when trying to write pid file to full partition, the only message
in log is:

    sh: echo: I/O error

Which is definitely insufficient


dumbbell:
Thank you!

PR #532, opened 2016-01-07
Remove unneeded sleep for a graceful stop by PID

bogdando:
The sleep in not needed according to the
https://www.rabbitmq.com/man/rabbitmqctl.1.man.html
"If a pid_file is specified, also waits for the process
specified there to terminate."

Related Fuel bug https://launchpad.net/bugs/1529897
Related PR
https://github.com/rabbitmq/rabbitmq-server/pull/523


bogdando:
@dmitrymex

dmitrymex:
@bogdando: agree, +1

michaelklishin:
Thank you.

PR #533, opened 2016-01-08
Set deleting exchange status

dcorbacho:
* Avoids race condition between declare and delete

Solves https://github.com/rabbitmq/rabbitmq-federation/issues/7


michaelklishin:
Oops, my comments ended up being on a commit in a fork.

I suggest that we rename `exchange-delete` to `exchange-delete-in-progress`. This is a fairly unexpected use of runtime parameters but no different in practice over using a separate local table, so why not.

Also, have we considered what happens when a node fails right after storing the parameter, then coming up? Runtime parameters are stored in a durable table. Should we perhaps clean up all such parameters on boot?

dcorbacho:
Just renamed the parameter.

Good point in the persistence, you're right, I will look at it. Maybe the runtime-parameters are not such a good idea? I used that because was the most similar structure already in place.

michaelklishin:
It is indeed a similar structure and it is already scoped per vhost, just like what we want. So I'm OK with using it.

michaelklishin:
There are test failures in the Java client (`exchange.delete`, `exchange.declare` tests). I'm investigating.

PR #543, opened 2016-01-12
Fix 'rabbitmqctl rotate_logs' behaviour

binarin:
When 'rabbitmqctl rotate_logs' is called without any parameters, it
clears logs unconditionally. And given that this form is used in
logrotate config files, this could result in data loss.

This could be reproduced with following scenario:
1) 'max_size' is set globally in lograte config
2) One of two rabbitmq logs is greater than that limit
3) Daily logrotate run was already performed today, and now we
   are calling it manually. In this case logrotate will copy only file
   that is bigger than max_size, but 'rabbitmqctl rotate_logs' will
   clear both of them - leading to data loss. 

I've decided to change behaviour of current command (instead of adding new one) because current behaviour doesn't make sense and is actually dangerous.

### DIFF
--- a/docs/rabbitmqctl.1.xml
+++ b/docs/rabbitmqctl.1.xml
@@ -275,10 +275,9 @@
               suffixed files.
             </para>
             <para>
-              When the target files do not exist they are created.
-              When no <option>suffix</option> is specified, the empty
-              log files are simply created at the original location;
-              no rotation takes place.
+              When the target files do not exist they are created. When
+              no <option>suffix</option> is specified, no rotation takes
+              place - log files are just re-opened.
             </para>
             <para role="example-prefix">For example:</para>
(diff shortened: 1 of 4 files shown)


michaelklishin:
@binarin thank you. Note that we are almost done with #94, which basically replaces our logging subsystem.

binarin:
Yep, I'm aware of this - but it will only go to 3.7.0. And I think that new related test case will be useful even for lager logging.

michaelklishin:
Absolutely, just making sure you're not caught by surprise. Once we merge #94, would you be interested in submitting another PR (with just the test) against `master`?

binarin:
Yes, I'll do it.

michaelklishin:
@binarin FYI, #94 is in.

PR #552, opened 2016-01-14
Limit number of unique node names for rabbitmqctl

### LINKED ISSUE #549: Atom table exhaustion due to rabbitmqctl node names
rabbitmqctl generate node names like 'rabbit-cli-&lt;pid&gt;@&lt;host&gt;'. The problem is that node names are stored in atom table and not garbage collected. 

Some numbers:
- Maximum pid number on linux is 32768 by default
- Maximum pid  could be increased to ~4 millions on 64-bit CPU
- Erlang atom table has default limit of 1048576 atoms.

So on linux any long running rabbitmqbroker could be crashed by constantly running rabbitmqctl:
- in cluster of more than 32 nodes (with default pid limit)
- or even on single node, after increasing that limit

While it sounds ridiculous, we've observed such a crash on a production server  )

There should be some way to generate rabbitmqctl nodenames without introducing large quantities of uncollected garbage.

binarin:
It prevents atom table overflow in a long running broker.

Fixes #549

### DIFF
--- a/src/rabbit_cli.erl
+++ b/src/rabbit_cli.erl
@@ -45,11 +45,21 @@
 
 -endif.
 
+ensure_cli_distribution() ->
+    case start_distribution() of
+        {ok, _} ->
+            ok;
+        {error, Error} ->
+            print_error("Failed to initialize erlang distribution: ~p.",
+                        [Error]),
+            rabbit_misc:quit(?EX_TEMPFAIL)
+    end.
+
 %%----------------------------------------------------------------------------
 
 main(ParseFun, DoFun, UsageMod) ->
     error_logger:tty(false),
-    start_distribution(),
+    ensure_cli_distribution(),
     {ok, [[NodeStr|_]|_]} = init:get_argument(nodename),
     {Command, Opts, Args} =
         case ParseFun(init:get_plain_arguments(), NodeStr) of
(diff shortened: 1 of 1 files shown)


src/rabbit_cli.erl
```diff
@@ -125,9 +135,22 @@ main(ParseFun, DoFun, UsageMod) ->
             rabbit_misc:quit(?EX_SOFTWARE)
     end.
 
+start_distribution_anon(0, LastError) ->
+    {error, LastError};
+start_distribution_anon(TriesLeft, _) ->
+    NameCandidate = list_to_atom(rabbit_misc:format("rabbitmq-cli-~2..0b", [rabbit_misc:random(100)])),
```
hairyhum: `rabbit_misc:random` is undefined.

binarin: Yeah, you'll need https://github.com/rabbitmq/rabbitmq-common/pull/48
And also you can't simply run ./scripts/rabbitmqctl from rabbitmq-server checkout - rabbit_common will be missing from code path (but on installed server it will be available through rabbit_common.ez archive).

michaelklishin: One can run `make package-generic-unix UNOFFICIAL_RELEASE=true` from the umbrella to produce a tarball. It should take a couple of minutes.

binarin: Then every sequence of "random" numbers will be the same - and it's exactly the thing that we want to avoid, to reduce contention.

PR #554, opened 2016-01-15
Use "short filenames" on Windows to escape non-ASCII characters

### LINKED ISSUE #493: RabbitMQ on Windows is confused with non-ASCII chars in the path
The most common sympton is when a user account contains non-ASCII characters, the management UI breaks. Therefore `%APPDATA%` expands to eg. `C:\Users\Jean-Sébastien\AppData\Roaming`.

If I understand correctly, Windows stores filenames using UTF-16. However, cmd.exe and batch files do not handle that and filenames are converted to one-byte-wide charsets (Windows-1252/CP 850 in my case). Once passed to Erlang through our scripts (`rabbitmq-server.bat`, `rabbitmqctl.bat`, `rabbitmq-env.bat`, etc.), filenames may be incorrects.

In the case of the management UI, the problem is that the `/api/nodes` endpoint wants to put the path to the configuration file (eg. `C:\Users\Jean-Sébastien\AppData\Roaming\RabbitMQ\rabbitmq.config`) in the JSON payload. The `é` encoded in Windows-1252 is obviously an invalid UTF-8 character and the endpoint returns an error. This endpoint is queried on the management UI homepage so once logged in, the user can't do much.

dumbbell:
Fixes #493.

### DIFF
--- a/scripts/rabbitmq-defaults.bat
+++ b/scripts/rabbitmq-defaults.bat
@@ -23,6 +23,9 @@ if "!RABBITMQ_BASE!"=="" (
     set RABBITMQ_BASE=!APPDATA!\RabbitMQ
 )
 
+REM Make sure $RABBITMQ_BASE contains no non-ASCII characters.
+for /f "delims=" %%F in ("!RABBITMQ_BASE!") do set RABBITMQ_BASE=%%~sF
+
 REM BOOT_MODULE="rabbit"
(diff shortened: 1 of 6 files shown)


dumbbell:
This patch was tested on a French version of Windows 8.1. My username contains `é` on this computer.

dumbbell:
This pull request against master was recreated against stable in #582.

PR #556, opened 2016-01-15
Add optional prefix for RabbitMQ node FQDNs

galanoff:
It would allow to instantiate multiple rabbit clusters constructed
from prefix-based instances of rabbit nodes.
For instance, if prefix is set to 'rabbitmq-' then all nodes names would have changed to: rabbitmq-<node_name-N>
This feature allows to run more than one RabbitMQ instance on each Pacemaker node.


michaelklishin:
Do you expect this to go into `3.7.0` only? If not, perhaps you want to submit this against `stable`.

@dmitrymex @bogdando any comments on this?

bogdando:
Please resubmit to the stable branch

michaelklishin:
Closing per @bogdando's comment.

PR #558, opened 2016-01-18
Add optional prefix for RabbitMQ node FQDNs

galanoff:
It would allow to instantiate multiple rabbit clusters constructed
from prefix-based instances of rabbit nodes.
For instance, if prefix is set to 'rabbitmq-' then all nodes names would have changed to: rabbitmq-
This feature allows to run more than one RabbitMQ instance on each Pacemaker node.


bogdando:
Thank you! LGTM. @dmitrymex ?

dmitrymex:
@bogdando: yep, looks good to me

PR #560, opened 2016-01-19
Reset master score if we decide to restart RabbitMQ on timeout

dmitrymex:
Doing otherwise might not trigger the restart while it is clearly
needed.


dmitrymex:
@bogdando : hey, what do you think regarding this PR?

bogdando:
LGTM, thank you

PR #562, opened 2016-01-20
Fix shell quoting issues in HA OCF script

binarin:
Without double quotes it complains about misplaced '-eq' operator.
With quotes but without default value it sometimes complains about bad integer.

@bogdando, WDYT? Is this fix in the spirit of your initial change?

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -1361,7 +1361,7 @@ get_monitor() {
     local name
     local node
     local nodelist
-    local rc_check
+    local rc_check=$OCF_SUCCESS
     local max
     local our_uptime
     local node_uptime


michaelklishin:
Is this mean to be included into `3.7.0` only?

scripts/rabbitmq-server-ha.ocf
```diff
@@ -1445,7 +1445,7 @@ get_monitor() {
       fi
     fi
 
-    if [ $rc -eq $OCF_ERR_GENERIC -o $rc_check -eq $OCF_ERR_GENERIC ]; then
+    if [ "$rc" -eq "$OCF_ERR_GENERIC" -o "${rc_check:-$OCF_SUCCESS}" -eq "$OCF_ERR_GENERIC" ]; then
```
bogdando: what is the point of this change, please elaborate?

binarin: To remove warnings from script output.
As descriptions says:
- "Without double quotes it complains about misplaced '-eq' operator" - rc_check is not always initialized
- "With quotes but without default value it sometimes complains about bad integer" - empty string is still not a valid integer

binarin: What default value should be used? `$OCF_SUCCESS`? Isn't it conterintuitive, given that default value for `rc` is `$OCF_ERR_GENERIC`?

bogdando: yes, $OCF_SUCCESS

binarin:
Fixed it differently, should go to 3.6.X.

PR #563, opened 2016-01-20
Improve OCF script diagnostics for timed-out 'list_channels'

binarin:
Currently time-out when running 'rabbitmqctl list_channels' is treated
as a sign that current node is unhealthy. But it could not be the
case, as the hanging channel could be actually on some other
node. Given that currently we have seen more than one bug related to
'list_channels', it makes sense to improve diagnostics here.

This patch doesn't change any behaviour, only improves logging after
time-out happens. If time-outs continue to occur (even with latest
rabbitmq versions or with backported fixes), we could switch to this
improved list_channels and kill rabbitmq only if stuck channels are
located on current node. But I hope that all related rabbitmq bugs
were already closed.

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -1501,6 +1501,7 @@ get_monitor() {
     local timeout_alive
     su_rabbit_cmd "${OCF_RESKEY_ctl} list_channels 2>&1 > /dev/null"
     rc_alive=$?
+    [ $rc_alive -eq 137 ] && ocf_log err "${LH} 'rabbitmqctl list_channels' timed out, per-node explanation: $(enhanced_list_channels)"
     check_timeouts $rc_alive "rabbit_list_channels_timeouts" "list_channels"
     timeout_alive=$?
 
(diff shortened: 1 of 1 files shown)


dmitrymex:
Looks good to me, the change only adds more debug output in case list_channels times out, so it should be safe. @bogdando ?

bogdando:
Yes, I liked the idea, @binarin did you test this?

binarin:
@bogdando Yes, I did minimal testing by simulating 'list_channels timeout.

scripts/rabbitmq-server-ha.ocf
```diff
@@ -1501,6 +1501,7 @@ get_monitor() {
     local timeout_alive
     su_rabbit_cmd "${OCF_RESKEY_ctl} list_channels 2>&1 > /dev/null"
     rc_alive=$?
+    [ $rc_alive -eq 137 ] && ocf_log err "${LH} 'rabbitmqctl list_channels' timed out, per-node explanation: $(enhanced_list_channels)"
```
bogdando: there is also 124 exit code we may want to inspect more precisely

bogdando:
I will test this as well, please hold for a while

bogdando:
works for me. It provides things like this:
"per-node explanation: [{rabbit@n2,ok,0},{rabbit@n1,ok,0}]"
Which is: a node, ok-if all channels' status has been checked, how many channels has been checked.
And if reported not ok for a non local node, that would mean that local node has been killed being innocent :)

michaelklishin:
@bogdando should we merge?

bogdando:
yes please

Issue #346, opened 2015-10-07
`time_compat`: Get rid of try/catch


dumbbell:
We can use two module files, one for Erlang 18+, one for Erlang up-to 17, both named `time_compat` internally. The correct module is manually loaded during startup.

As the modules will be in `rabbitmq-common`, we need to determine how to do this load:
* Do we need an application module?
* Can we do this from an `-on_load()` directive? Is this supported by Erlang R16B03?

dumbbell:
The same solution will be applied to `ssl_compat` (#347).

michaelklishin:
Note that we can fix this for `3.7.0` only and thus consider solutions that require Erlang 17.5+.

dumbbell:
I propose the following solution, tested in the following repository:
https://github.com/dumbbell/erlang-time-compat

The idea is to build three modules from the same source file (`time_compat.erl`):
 * `time_compat_old.beam`, using the old Time API (ie. based on `erlang:now/0`);
 * `time_compat_new.beam`, using the new Time API;
 * `time_compat.beam`, a module which replaces itself with one of the other modules.

See `swap_module()` and `do_swap_module()` in `time_compat.erl`.

This example is incomplete because it only exports `unique_integer/0`.

The solution can be used for `ssl_compat` (#347), though the following test needs to be adapted:
https://github.com/dumbbell/erlang-time-compat/blob/master/time_compat.erl#L57

I believe the other tricky part is how to plug this solution to the build system (ie. building three modules out of one source file).

dumbbell:
@michaelklishin: Why not for 3.6.1? The change should be transparent for the consumers of `time_compat` and `ssl_compat`. It should be an implementation detail.

michaelklishin:
`3.6.1` sounds good. I was just mentioning it is a possibility.

dumbbell:
Ok. I think we should target 3.6.x because the basic try/catch can be a performance penalty, though I didn't measure anything.

dcorbacho:
The `-on_load()` directive cannot be used, as doing an external call to the module itself causes a deadlock http://www.erlang.org/doc/reference_manual/code_loading.html#id88557. This happens, at least, when trying to load the new module.

Adding an application behaviour to rabbit_common is a very complex issue, because of rabbit startup. The scripts try to use rabbit_common without trying to initialise the application. Thus, it would require touching the release, packages, etc.
```
[...]

PR #572, opened 2016-01-22
Avoid hipe compile versioned modules

dcorbacho:
Part of #346


michaelklishin:
Merging per discussion in #346 and another round of QA on R16B03 and 18.2.1, with and without HiPE.

PR #574, opened 2016-01-22
Remove mentions of `sd_notify` library

binarin:
Calling shell helper systemd_notify(1) is enough to do the same.

Systemd unit file should specify `NotifyAccess=all`.


michaelklishin:
This is in follow-up to https://github.com/rabbitmq/rabbitmq-server/pull/573, references #52 (where most of the discussion is).

michaelklishin:
Merged too early, there is a missing comma on the changed line.

binarin:
Sorry about that. As always when you think that nothing could go wrong it'll definitely go wrong )

Also I've filed bugs about 'NotifyAccess=all' to RedHat and Debian, so it'll not catch them as a surprise.

Issue #551, opened 2016-01-14
Explore if RabbitMQ can load and validate its config file instead of application_controller


michaelklishin:
This would give us some specific improvements:

 * We can do much more extensive validation
 * We can produce more sensible error messages

This is also a good step towards https://github.com/rabbitmq/rabbitmq-server/issues/550 and should be investigated before it so that we have a more informed opinion.

We should also explore Riak's approach and sub-projects in this area.

michaelklishin:
It would be interesting to know what @essen, @dcorbacho, @Ayanda-D, and @Dzol think of this.

While #550 is going to happen, supporting Erlang term file configs would be *very* nice for backwards compatibility. We don't use any particularly sophisticated data structures so there should be a way to do that by coercing YAML to the Erlang terms we use today.

hairyhum:
As I remember, Riak and Ejabberd use additional `.config` file for variables with too sophisticated structure.

michaelklishin:
So they effectively load two configs and merge them? That sounds reasonable for us.

hairyhum:
Riak is not using YAML, they use this https://github.com/basho/cuttlefish
This library create `.config` every time riak is started from `.conf` file (in sysctl format) and `advanced.config` (erlang config format)

michaelklishin:
OK. Whether or not we want to use the same format and Cuttlefish belongs to #550. Is the generated `.config` file still loaded by application controller or one of Riak apps?

hairyhum:
It is loaded by application controller just like regular `config` file. All validations are being executed during config generation with `cuttlefish`.

michaelklishin:
So in other words, if we are to follow this route (which is quite clever IMO), validation would be handled in #550 and this issue should be simply closed because there's no real need to do this.

Any objections to this approach?

Also, @hairyhum please take a look at how they handle file generation (e.g. where the target file is located — is it next to the source one?) and how we can make it work with our current way of setting config file path (which is via environment variables). This also can complicate upgrades, so we need to consider that.

hairyhum:
Riak use `cuttlefish_rebar_plugin` and shell scripts to generate `app.config` in `.../generated.configs/app.<date>.config` and run `erl` with `-config` pointing to this location.
If we will use cuttlefish env variable could point to `.conf` (e.g. `RABBITMQ_CONF_FILE=rabbitmq.conf`) file and do same thing - create `app.config` in some location and point erlang to it.

[...]

Issue #546, opened 2016-01-13
Bump minimum required Erlang/OTP version for RabbitMQ 3.7.0


michaelklishin:
What the title says. `3.7.0` is scheduled for Q3 or Q4, 2016, so 18.2 doesn't seem particularly aggressive.

However, we'd need to bump standalone OS X release first as it is currently on 17.x.

dumbbell:
I vote for Erlang 18.x because the maps implementation is much better apparently. We could get rid of the `time_compat` and `ssl_compat` modules at the same time.

michaelklishin:
I vote for 18.x for a different reason: fewer obscure cipher suite issues for TLS connections, in particular with ECC.

essen:
In addition to the previous comments:

I vote 18 for maps, because the syntax is more complete, and large maps are fast.

Going to 18 opens the way to eventually switching to Cowboy 2 (currently WIP) which would allow me to delete a lot of code in various parts of RabbitMQ and would also allow supporting HTTP/2.

18.2 would mean having some extra fixes for SSL that makes it easier to avoid it taking insane amounts of memory under specific scenarios. But 18.2 was a little broken on Windows and FreeBSD, so we need to require 18.2.1 at a minimum on those platforms (easy enough for Windows, since it's the last release listed on erlang.org; for FreeBSD we need to check if patches are in).

michaelklishin:
@essen thanks, that's good to know. By the time 3.7.0 ships we'd have 18.4 or 18.5 or maybe even 19.0, so it's `18.x` vs `17.5`.

dumbbell:
@essen: On FreeBSD, the package maintainer does a pretty good job. Erlang 18.2.2 is already available and is what you get if you ask for "erlang" (without specifying a version).

dumbbell:
FYI, our test framework was fixed to work with Erlang 18.2; see rabbitmq/rabbitmq-test#29.

hairyhum:
There are several nasty regressions in `18.3` like https://github.com/rabbitmq/rabbitmq-auth-backend-ldap/pull/42
Maybe it makes sense to set requirement to 18.3.x or 18.4

michaelklishin:
We can bump it to `18.3` for now and `18.4` (or even `19.0`) at a later point. We have months to do that before 3.7.0 ships.

dumbbell:
There are at least two regressions with TLS in 18.3.x:
[...]

PR #575, opened 2016-01-22
Support new configuration format

### LINKED ISSUE #550: Use a more user-friendly format for configuration file
E.g. using YAML is going to be a lot more user-friendly plus easier to generate for tools such as Chef and Puppet.

This is a long standing idea previously discussed with @videlalvaro and @dumbbell.

Why YAML and not JSON? A few reasons:

 * JSON doesn't allow comments
 * YAML works quite well for Cassandra and ElasticSearch
 * @dumbbell maintains an Erlang YAML parser/generator

Alternatively we can consider using [Cuttlefish](https://github.com/basho/cuttlefish), which is used by Riak.

Things to keep in mind: how Cuttlefish's config file generation approach might work with our current way of configuring a `.config` file path; how this can affect upgrades.

hairyhum:
This adds support for the new config format.

Fixes #550, #551.


hairyhum:
@michaelklishin @dumbbell @essen Can you please take a look at `rabbitmq.conf` and `schema` files? Maybe there are more validations or config options required. QA comments require discussion.

Makefile
```diff
@@ -451,3 +451,8 @@ package-rpm-suse package-windows package-standalone-macosx \
 package-generic-unix: $(PACKAGES_SOURCE_DIST_FILE)
 	$(verbose) $(MAKE) -C packaging $@ \
 		SOURCE_DIST_FILE=$(abspath $(PACKAGES_SOURCE_DIST_FILE))
+
+generate-config:
+	rm -rf generated/rabbitmq.*.config
+	./cuttlefish -e . -s ./schema -c rabbitmq.conf.d/rabbitmq.conf -f rabbitmq
```
michaelklishin: This bundles Cuttlefish binary. Is there a way to manage this via `erlang.mk`? @essen, is there a standard way?

hairyhum: This make task should be moved to `rabbitmq-server` script to enable config update. Will there be  `erlang.mk` at this point?

michaelklishin: There won't be but that's not my point. We use `erlang.mk` to manage dependencies. Bundling a binary goes against that. Using `erlang.mk` to manage this dependency and having `cuttlefish` the binary available at runtime are not mutually exclusive.

essen: I need to test it to make sure but downloading/compiling something is a matter of adding it to BUILD_DEPS and setting the appropriate dependent targets. I'm not sure if cuttlefish works well with Erlang.mk though (I know it compiles, I don't know if it produces a working escript). I'll experiment.

essen: Right, it doesn't build a proper escript out of it right now (rebar escript options are ignored). If you don't care too much about having a "cuttlefish" executable you can have it in `BUILD_DEPS` and call the erlang functions directly through `$(erlang ...)`, otherwise I can fix it in Erlang.mk.

michaelklishin: It's OK to keep it for now but by the time we are ready to merge this PR, I'd like the dependency to be managed by `erlang.mk` if we can help it.

hairyhum: Package building should generate escript to be used after installation to update config.

essen: I opened https://github.com/ninenines/erlang.mk/issues/503 to track this. Feel free to remind me if it's not done when it's time to merge, it shouldn't take too long to do.

rabbitmq.conf.d/rabbitmq.conf
```diff
@@ -0,0 +1,722 @@
...
+
+## The default "guest" user is only permitted to access the server
+## via a loopback interface (e.g. localhost).
+## {loopback_users, [<<"guest">>]},
+##
+loopback_user.guest = true
+
+## Uncomment the following line if you want to allow access to the
+## guest user from anywhere on the network.
+loopback_user.guest = false
```
michaelklishin: The line above says "uncomment …" but `loopback_user.guest` is not commented out.

rabbitmq.conf.d/rabbitmq.conf
```diff
@@ -0,0 +1,722 @@
...
+# management.http_log_dir = /path/to/access.log
+
+## Change the port on which the HTTP listener listens,
+## specifying an interface for the web server to bind to.
+## Also set the listener to use SSL and provide SSL options.
+##
+
+# QA: Maybe use IP type like in tcp_listener?
+management.listener.port = 12345
+management.listener.ip   = 127.0.0.1
```
michaelklishin: What would listening on multiple interfaces look like?

rabbitmq.conf.d/rabbitmq.conf
```diff
@@ -0,0 +1,722 @@
...
+## http://www.rabbitmq.com/management.html#fine-stats for more details.
+management.rates_mode = basic
+
+## Configure how long aggregated data (such as message rates and queue
+## lengths) is retained. Please read the plugin's documentation in
+## http://www.rabbitmq.com/management.html#configuration for more
+## details.
+##
+# QA: Some funny syntax
+management.sample_retention_policies.global.60    = 5
```
michaelklishin: We could use `minute`, `hours`, and `24hours` here.

rabbitmq.conf.d/rabbitmq.conf
```diff
@@ -0,0 +1,722 @@
...
+##
+mqtt.vhost = /
+
+## Specify the exchange to which messages from MQTT clients are published.
+##
+mqtt.exchange = amq.topic
+
+## Specify TTL (time to live) to control the lifetime of non-clean sessions.
+##
+mqtt.subscription_ttl = 1800000
```
michaelklishin: It's no longer 30 minutes.

rabbitmq.conf.d/rabbitmq.conf
```diff
@@ -0,0 +1,722 @@
...
+# =======================================
+
+
+## Connections that are not authenticated with SASL will connect as this
+## account. See the README for more information.
+##
+## Please note that setting this will allow clients to connect without
+## authenticating!
+##
+amqp1.default_user = guest
```
michaelklishin: We should use `amqp10`.

rabbitmq.conf.d/rabbitmq.conf
```diff
@@ -0,0 +1,722 @@
...
+
+##
+## Connecting to the LDAP server(s)
+## ================================
+##
+
+## Specify servers to bind to. You *must* set this in order for the plugin
+## to work properly.
+##
+# ldap.servers.myserver = your-server-name-goes-here
```
michaelklishin: As discussed for authentication backend, `myserver` here probably can be just a number. Not because ordering matters but because picking up names for each LDAP server would be a bit silly.

rabbitmq.conf.d/rabbitmq.conf
hairyhum: There is no support for multiple interfaces in management plugin.
https://github.com/rabbitmq/rabbitmq-management/blob/master/src/rabbit_mgmt_app.erl#L28

hairyhum:
> cluster_nodes is (semantically) a list: do we support multiple values for it at the moment?

There is no support for lists. Or you mean using numbers as entry names, like in auth_backends?

michaelklishin:
`rabbit.cluster_nodes` is a collection so we should support it in the new config. I can't think of a scheme better than numbers, although we don't care about ordering.

hairyhum:
Added generation to `rabbitmq-server` scripts. It checks file extension of `RABBITMQ_CONFIG_FILE` and if it's `.conf' generates config with cuttlefish

hairyhum:
There is a problem with windows service. `start-service.bak` script installs service with `erlsrv` utility and when the service is started or stopped it doesn't call any shell scripts.
So using shell script approach to generate config doesn't work in case of windows services. 
Looks like we need to generate it in erlang code, maybe in boot steps.

hairyhum:
Config generation is performed by `cuttlefish` escript which is called from `rabbit_config` module before application load.
After config generation app configuration is updated with `application_controller:change_application_data/2`

To generate config init arguments `-conf`, `-conf_dir`, and `-conf_gen_script` should be passed to `erl` on start.
`-conf` - `.conf` file (without extension)
`-conf_dir` - generated config directory
`-conf_gen_script` - `cuttlefish` script location. `rabbitmq.schema` should be in same direcroty

If `-config` argument was passed to `erl` and the `.config` file exists - nothing will be generated.
If `-conf` argument was passed and `.conf` file exists - try generate `.config` from arguments.
If generation successful - update applications config.

Unix scripts pass `-conf...` arguments only if there is `.conf` file
Windows service script always have `-conf...` arguments if there is no `.config` file, because it's impossible to change start args without service reinstallation.

hairyhum:
Config update will require erlang 17.1

michaelklishin:
@hairyhum that's fine because #546.

hairyhum:
Added support to `additional.config`, which will be merged with generated config.

hairyhum:
To have `.config` support plugins should provide `.schema` file in `priv` directory. But it's not clear how broker should locate this files: copy on plugin install or search in plugin dir when generating config. 
Also when enabling new plugin perhaps broker should generate and load config for it.

scripts/rabbitmq-defaults
```diff
@@ -38,6 +38,8 @@ CONFIG_FILE=${SYS_PREFIX}/etc/rabbitmq/rabbitmq
 LOG_BASE=${SYS_PREFIX}/var/log/rabbitmq
 MNESIA_BASE=${SYS_PREFIX}/var/lib/rabbitmq/mnesia
 ENABLED_PLUGINS_FILE=${SYS_PREFIX}/etc/rabbitmq/enabled_plugins
+GENERATED_CONFIG_DIR=${SYS_PREFIX}/var/lib/rabbitmq/config
+RABBITMQ_CONFIG_ADDITIONAL=${SYS_PREFIX}/etc/rabbitmq/additional
```
michaelklishin: I suggest we rename it to "advanced config" because it is meant to be used for more complex cases, e.g. LDAP query configuration.

michaelklishin:
Some initial thoughts after a quick QA session.

The changes are backwards compatible: I managed to use my `rabbitmq.config` by overrinding `RABBITMQ_CONFIG_FILE` just like I do today.

We need to make sure that `RABBITMQ_GENERATED_CONFIG_DIR` is local to the RabbitMQ "home" directory for generic UNIX tarball and `make run-broker` when running from source, otherwise node fails to start with a fairly obscure error message.

Sysctl format examples need to be easier to find, I think they should be under `docs` just like `rabbitmq.config.example`.

"Additional" config should be "advanced config", because it is meant to be used for more complex cases.

Not sure if `scripts/rabbitmq.schema` belongs under `scripts`. The file is currently treated as binary by git, while it's fairly human-readable.

Using the new format seems nice so far. Will try to replicate more complicated examples and report back.

Lastly, we need to begin porting documentation examples (as found on the website). We can't remove existing Erlang term examples entirely, so they'd have to co-exist for some time. I'm happy to help with that and explaining the difference in various documentation guides.

michaelklishin:
`make generate-config` is currently broken: `cuttlefish` is now under `script`, plus it exists with code 1 and produces no error messages.

michaelklishin:
We currently do not log correct config path. E.g. if I export

```
RABBITMQ_CONFIG_FILE=./etc/rabbitmq/rabbitmq.sysctl
```

and start the node, it will log that effective config path is `./etc/rabbitmq/rabbitmq.sysctl.config` and that it is not found.

We should log the `.conf` path if it's present, and separately log generated config path (say at `debug` level).

michaelklishin:
Related to the comment above: config file paths are displayed in the management UI, we need to make sure those values are correct and what the user would expect as well.

hairyhum:
Looks like plugin support can be harder than it seems.
Plugins are extracted after application been loaded and `.config` file should be already generated, but if there are some plugin specific config parameters in `.conf` file it's impossible to generate config without plugin `.schema`

michaelklishin:
@hairyhum @dumbbell suggested that eliminating plugin extraction entirely should be possible with modern OTP releases. If using `.ez` files directly is possible, can we perhaps scan code path for schema files?

michaelklishin:
or we could include bundled plugins schema in the server repo, then 3rd party plugins would use advanced config. Not ideal but would be an improvement over what we have today.

hairyhum:
It's possible to extract `.schema` files only. 
It's also possible to generate separate "pre_boot.conf" with log and hipe config only to prepare plugins and then generate from normal config with plugins.

hairyhum:
Since `kernel` application cannot be reloaded, its options should be only set using `.config`. I'm starting to think that it would be easier to have windows specific behaviour.

michaelklishin:
@hairyhum it sounds reasonable to me to ask for `kernel` values to be tweaked via advanced config. What kind of Windows-specific behaviour do you have in mind?

hairyhum:
We are always starting broker using `rabbitmq-server` on UNIX, but windows service cannot run bash scripts, so windows should have ability to generate config from working node. What I have in mind right now is either to restart `init` after config generation, or start with kernel in `advanced.config` and with something like `-config advanced -conf_advanced advanced -conf rabbitmq ...` and check if `conf_advanced` and `config` are the same to generate new config.

hairyhum:
Erlang process is started with `advanced.config` if `rabbitmq.config` file is not exist. 
During installation on windows `rabbitmq.config` is being renamed to `advanced.config`

Plugins schemas are extracted on config generation step.

hairyhum:
Looks like it's close to production ready. Issues with windows is fixed. 

How it works:

Env variable `RABBITMQ_CONFIG_FILE` is used to direct to either `.config` or `.conf` file.
If it's set to `.config` file - node is loading this file and works like previous versions.
If it's set to '.conf' - config file is generated in `RABBITMQ_GENERATED_CONFIG_DIR` using `RABBITMQ_CONFIG_ADVANCED` as advanced `.config` file and schemas from `priv` directory of rabbit and plugins. Plugin schemas are being extracted to `priv` directory of `rabbit`.

On UNIX if `RABBITMQ_CONFIG_ADVANCED` exists, it's passed to `erl` as `config` argument and loaded duting boot, so values for `kernel` or `sasl` can be used there. 
On WINDOWS to run service `RABBITMQ_CONFIG_ADVANCED` should exist during service installation to be used. So installer create it or copy old version `rabbitmq.config` if it can be found.

I'm not quite sure if schemas should live in `priv` directory.

What's left:

Config regeneration on plugin enable/disable.
Docs and examples

hairyhum:
Added module to test config files generation. 
Can be used to test examples from website.

michaelklishin:
We need to add schema and examples for Web STOMP and Web MQTT, otherwise this looks good to merge to me.

hairyhum:
Added schemas for web-stomp and web-mqtt and examples for web-stomp.

Issue #559, opened 2016-01-19
Channel Interceptors should be enabled/disabled together with plugins


videlalvaro:
When a plugin that provides an interceptor is disabled, then running channels should reflect the changes.

There could be a function similar to this one: https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_channel.erl#L334 that handles updating the channel state in order to remove the interceptor.

In a similar way, when enabling a plugin, then channels should be notified of new interceptors, if any.

michaelklishin:
As part of this we need to update plugins that provider interceptors. For example, []() fails to activate with  https://github.com/rabbitmq/rabbitmq-server/pull/578 and https://github.com/rabbitmq/rabbitmq-common/pull/49 checked out.

hairyhum:
@michaelklishin what do you mean by 'fails to activate'? I have been testing the feature using this plugin.

michaelklishin:
I'm getting an `undef` when I `make run-broker` with `rabbitmq-server-559` checked out. I will give it another try.

michaelklishin:
I confirm that with manual testing against `rabbitmq_message_timestamp`, disabling the plugin does have effect on newly opened channels, and so does re-enabling. However, test suites in `rabbitmq_test` and `rabbitmq_federation` fail.

hairyhum:
I have set up clean umbrella clone, checked up `rabbit`, `rabbit_common`, `rabbitmq_test`, and `rabbitmq_federation`. Tests passed (except clustering management, which is unrelated).

PR #578, opened 2016-01-25
Support for [de-]registration callbacks in rabbit_registry

hairyhum:
Related to #559

Support for callbacks on register/unregister anything in `rabbit_registry`.
Currently is used to add/remove channel interceptors, but can be used to update other states, related to registry.
Calls `<class_module>:register/2` `<class_module>:unregister/1`, if class module exports such functions.


src/rabbit_registry.erl
```diff
@@ -114,6 +116,29 @@ conditional_unregister({exchange_decorator, Type}) ->
 conditional_unregister(_) ->
     ok.
 
+% update channels to enable/disable interceptors
+registered({{Class, Type}, ModuleName}) ->
+    ClassModule = class_module(Class),
+    case proplists:get_value(registered, ClassModule:module_info(functions)) of
```
videlalvaro: I think this problem deserves a deeper solution. For example by creating a behaviour that's related to the `rabbit_registry`. Also not sure taking the common `registered` and `unregistered` names is a good idea, since many modules might have those functions and not be related to the rabbit registry.

@michaelklishin @dumbbell you should provide input here

michaelklishin: Yes, `registered` is not a very specific name. @videlalvaro what functions do you suggest the behaviour would include?

hairyhum: Agree that a behaviour would be helpful. Not sure what function names should be there. Maybe straightforward `added_to_registry`, `removed_from_registry`?

michaelklishin: Behaviours tend to use short names but function name conflicts in multi-behaviour modules are also a known problem. If we were to use a longer name, I'd go with `added_to_rabbit_registry` and such. At least this way we are as specific as we can :)

PR #579, opened 2016-01-25
DO NOT MERGE: Lazy queue: Use the same conditions for betas->deltas than default queue

### LINKED ISSUE #569: Lazy queues: a RAM duration of 0 causes the queue to be slow
The `test_variable_queue_dynamic_duration_change()` test in `rabbit_tests.erl` is slower when lazy queues are used than when default queues are used. We talk about 4 minutes for the former versus 40 seconds for the latter, when the RAM duration is set to 0. The whole testsuite is also slower.

dumbbell:
Fixes #569.

### DIFF
--- a/src/rabbit_variable_queue.erl
+++ b/src/rabbit_variable_queue.erl
@@ -2318,10 +2318,11 @@ reduce_memory_use(State = #vqstate {
 %% When using lazy queues, there are no alphas, so we don't need to
 %% call push_alphas_to_betas/2.
 reduce_memory_use(State = #vqstate {
-                             mode = lazy,
+                             mode             = lazy,
                              ram_pending_ack  = RPA,
                              ram_msg_count    = RamMsgCount,
-                             target_ram_count = TargetRamCount }) ->
+                             target_ram_count = TargetRamCount,
+                             io_batch_size    = IoBatchSize }) ->
     State1 = #vqstate { q3 = Q3 } =
         case chunk_size(RamMsgCount + gb_trees:size(RPA), TargetRamCount) of
             0  -> State;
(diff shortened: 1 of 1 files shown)


dumbbell:
All tests in `rabbit_tests` pass, including the one testing the conversion from default to lazy queue. `push_betas_to_deltas()` is called in `reduce_memory_use()` for lazy queues during the latter.

In the end, the total runtime for `make unit` is around 20 minutes, where `make lazy-vq-test` takes 10 minutes (it was more like 40 minutes before the change). Note that `make lazy-vq-test` runs only `test_lazy_variable_queue()`, unlike `make unit` which runs `all_tests()`.

dumbbell:
I believe the change to be correct because default and lazy queues now have the same decision process when dealing with beta to delta conversion. It works for default queue and, if I understand correctly, lazy queue doesn't change that part.

src/rabbit_variable_queue.erl
```diff
@@ -2329,13 +2330,22 @@ reduce_memory_use(State = #vqstate {
...
+    %% Q3 could hold more messages than usual. In this case,
+    %% push_betas_to_deltas() will be called.
     State3 =
         case chunk_size(?QUEUE:len(Q3),
                         permitted_beta_count(State1)) of
-            0  ->
-                State1;
-            S2 ->
-                push_betas_to_deltas(S2, State1)
+            S2 when S2 >= IoBatchSize ->
```
videlalvaro: why do you care about the IoBatchSize?

michaelklishin:
@bdshroyer has expressed interest in QA'ing this. We would still highly appreciate feedback from @videlalvaro, of course.

videlalvaro:
@michaelklishin see the line note I've left. The credit flow docs and the variable queue docs should hint at why that might be wrong.

What to keep in mind when reviewing this:

lazy-queues _start_ with every message paged to disk, this can happen when:

- The queue is declared, so every new message will go straight to disk.
- The queue is converted from _default_ mode to _lazy_ mode, in this case every message should be paged to disk. I'm not convinced this patch achieves that.

A lazy queue should only have messages in memory if at any point in its life a consumer fetched messages from it. In that case, around 16k messages will be loaded from disk. See `SEGMENT_ENTRY_COUNT`. That's why I've told @dumbbell that I don't think it make sense to run those slow tests this patch is trying to address with a `duration = 0`, because that's effectively asking the queue to load 16k msgs from disk, flush them, load them, flush them, etc. This is a similar problem to the queue length = 20 reported by some other user, which doesn't make any sense for a lazy-queue. `duration = 0` does't affect a _default_ queue, because a default queue never pages every message to disk, as seen in the functions @dumbbell is modifying.

michaelklishin:
I'll close this because there were changes in relevant areas and new information suggests there may be other improvements worth considering first.

Issue #343, opened 2015-10-04
Batch writes to index when requeueing


videlalvaro:
This area could also batch operations when writing to the index.

It's quite critical to have this as fast as possible, since slave promotion into master does requeue'ing

michaelklishin:
@videlalvaro is this still relevant?

videlalvaro:
it could be improved when many rejects are sent at the same time, but I'm not sure the gains will be too big

michaelklishin:
`basic.nack` with `multiple` set to `true` is probably the only realistic scenario.

essen:
Just found out about https://github.com/rabbitmq/rabbitmq-server/commit/d85792bd6fe579b291533deb35cc7166510cac40 (there's already a rabbitmq-server-343 branch it seems). The good news is that what I did is basically the same so I guess I'm on the right tracks. I need to experiment with calling update index twice though. I'll push/open a PR tonight but I'll want to do more testing next week.

PR #585, opened 2016-01-27
Batch writes to the index on requeue

essen:
Flagged "do not merge" because I want to run more tests. But if you do and find out the performances are better or equal, I won't get mad if you merge it.

Fixes https://github.com/rabbitmq/rabbitmq-server/issues/343.


essen:
I have updated to current stable, pushed and this can be merged.

I have done a few more tests, in particular in a cluster with a HA queue, and was not able to see much of a difference before/after this patch. Still this brings this part of the code in line with the rest of the module and should have no negative impact.

Issue #273, opened 2015-08-19
RabbitMQ-wrapper-script breaks if "sed" does not have "-E" option.


nishan:
rabbitmq-wrapper-scripts break if sed does not have "-E" option.

Not all machines have a version of "sed" which supports "-E" option. Perhaps "-r" is a safer option to use.

I found this on a CentOS 5.11 machine.  I tested on CentOS 6 and Debian 7, where I found the same problem.

michaelklishin:
Kudos to @nishan for submitting #275. The change looks good, so it's a matter of small technicalities to get it in.

nishan:
Thanks @michaelklishin. I have raised a new pull request on stable branch.

https://github.com/rabbitmq/rabbitmq-server/pull/276

michaelklishin:
@nishan I will pull your changes manually in a bit.

nishan:
Oops! Sorry.  Let's try one more time?

https://github.com/rabbitmq/rabbitmq-server/pull/277

michaelklishin:
Note: this was in a packaging script and since recent OS X versions have a modern `sed`, this only needed testing on Linux.

dumbbell:
Since RabbitMQ 3.5.2.

PR #593, opened 2016-01-31
Use -r with sed on Linux

### LINKED ISSUE #592: scripts/rabbitmq-env doesn't work with GNU sed 4.2.2
The BSD-style -E option is not recognized, tested in SLES 12 with GNU sed 4.2.2.  The equivalent -r option works (and is recognized in current BSD releases).

```
--- /usr/lib/rabbitmq/lib/rabbitmq_server-3.6.0/sbin/rabbitmq-env.old   2016-01-29 22:24:22.000000000 -0600
+++ /usr/lib/rabbitmq/lib/rabbitmq_server-3.6.0/sbin/rabbitmq-env       2016-01-29 22:24:29.000000000 -0600
@@ -106,7 +106,7 @@
rmq_normalize_path() {
     local path=$1
-    echo "$path" | sed -E -e 's,//+,/,g' -e 's,(.)/$,\1,'
+    echo "$path" | sed -r -e 's,//+,/,g' -e 's,(.)/$,\1,'
}
 rmq_normalize_path_var() {
```

michaelklishin:
Fixes #592.

We previously did the same change in #273 (PR: #275), the file changed there is now at `scripts/rabbitmq-script-wrapper`.

Note that #592 recommends using `-r` unconditionally but that option
is not recognised by sed which ships with OS X.

### DIFF
--- a/scripts/rabbitmq-env
+++ b/scripts/rabbitmq-env
@@ -103,10 +103,15 @@ fi
 
 ##--- Set environment vars RABBITMQ_<var_name> to defaults if not set
 
+SED_OPT="-E"
+if [ $(uname -s) = "Linux" ]; then
+    SED_OPT="-r"
+fi
+
 rmq_normalize_path() {
     local path=$1
 
-    echo "$path" | sed -E -e 's,//+,/,g' -e 's,(.)/$,\1,'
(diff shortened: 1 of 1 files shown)


michaelklishin:
@dumbbell FYI.

michaelklishin:
Submitted against the wrong branch.

PR #599, opened 2016-02-02
Fix uninitialized status_master

bogdando:
Fix multiple nodes may be reported in logs as the running master

Related Fuel bug https://bugs.launchpad.net/bugs/1540936


bogdando:
https://github.com/rabbitmq/rabbitmq-server/pull/600 is the correct one

PR #602, opened 2016-02-03
Streamline checking for cluster partitioning

dmitrymex:
Move check if we are current cluster master to earlier place in code.
That way we will avoid unnecessary operations for master case.


dmitrymex:
@bogdando : hey, what do you think about it?

bogdando:
please hold for a while, I'm testing this

bogdando:
LGTM, thank you!

PR #611, opened 2016-02-04
Auto deleted queue log

### LINKED ISSUE #590: Explicitly log when auto-delete and exclusive queues are deleted
We see users claiming they had message loss while in reality their apps relied on auto-delete and exclusive queues.

Now that we have a revamped logging subsystem, there is no reason why we should not log this at least as `debug` messages, possibly introducing a new logging category, `queues`.

hairyhum:
Fixes #590

Added some functionality to logging:
- Support for old format log to category. 
- Separate sink for `error_logger` messages. Works same way as `rabbit_` sinks.
- Removed rabbitmq lager fork from dependency

### DIFF
--- a/src/rabbit_amqqueue_process.erl
+++ b/src/rabbit_amqqueue_process.erl
@@ -1361,9 +1361,9 @@ handle_pre_hibernate(State = #q{backing_queue = BQ,
 format_message_queue(Opt, MQ) -> rabbit_misc:format_message_queue(Opt, MQ).
 
 log_auto_delete(Reason, Owner, State = #q{ q = #amqqueue{ name = Name } }) ->
-    rabbit_queue:log("Queue deleted automatically: Queue name: ~p Reason: " 
-                         ++ Reason ++ " ~p",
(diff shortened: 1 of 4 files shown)


michaelklishin:
This needs some work, as explained on Slack. It currently uses not particularly clear wording and doesn't distinguish between exclusive and auto-delete queues, which have different removal reasons/events.

src/rabbit_amqqueue_process.erl
```diff
@@ -142,6 +142,8 @@ init_it(Recover, From, State = #q{q = #amqqueue{exclusive_owner = Owner}}) ->
                  {_, Terms} = recovery_status(Recover),
                  BQS = bq_init(BQ, Q, Terms),
                  %% Rely on terminate to delete the queue.
+                 log_auto_delete("Exclusive query connection closed",
```
michaelklishin: You probably mean "queue", not "query". Also, this conflates exclusive and auto-delete queues. That's confusing, please use separate functions with clearer wording (as discussed on Slack).

PR #615, opened 2016-02-09
Use basic regular expressions for sed portability

cbandy:
More maintainable (imo) solution to a barrage of issues about `sed -E` vs `sed -r`. See #594.

On El Capitan and Ubuntu Trusty:
```
$ echo 'a//b/c///d/e//' | sed -e 's#/\{2,\}#/#g' -e 's#/$##'
a/b/c/d/e
```

### DIFF
--- a/scripts/rabbitmq-env
+++ b/scripts/rabbitmq-env
@@ -103,15 +103,11 @@ fi
 
 ##--- Set environment vars RABBITMQ_<var_name> to defaults if not set
 
-SED_OPT="-E"
-if [ $(uname -s) = "Linux" ]; then
-    SED_OPT="-r"
-fi
-
 rmq_normalize_path() {
     local path=$1
 
-    echo "$path" | sed $SED_OPT -e 's,//+,/,g' -e 's,(.)/$,\1,'
+    # Remove redundant slashes and strip a trailing slash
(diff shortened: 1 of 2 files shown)


michaelklishin:
Looks reasonable, thank you.

michaelklishin:
This can (should?) go into `stable`, too.

cbandy:
Shall I open another PR?

michaelklishin:
@cbandy no need, this was for our team members. We will cherry-pick Thank you.

michaelklishin:
Tested this on OS X and Ubuntu (with a Debian package), found no issues with it. @dumbbell FYI.

dumbbell:
Sorry, I didn't receive any notification from GitHub and missed this issue.

I was using `sed -E` (`sed -r` was only for a user of really old GNU sed) because extended POSIX regular expressions are portable, unlike `sed` without `-E`. At least in my experience.

I didn't test this yet, but I would not be surprised if it's broken on *BSD and some OS X releases.

cbandy:
- http://www.freebsd.org/cgi/man.cgi?query=re_format&sektion=7
- https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man7/re_format.7.html

PR #617, opened 2016-02-10
rabbit_mirror_queue_sync: Pause the syncer process if memory is low

### LINKED ISSUE #616: Mirrored queues can cause a node to run out of memory
When a slave queue is syncing messages from a master queue, the sync process continues to send messages even if the node is above its memory high watermark. This can lead to a crash of the slave node.

dumbbell:
If any node in the cluster is running out of memory, we pause the syncer process. If we don't do this, this alarmed node could crash.

Fixes #616.

### DIFF
--- a/src/rabbit_mirror_queue_sync.erl
+++ b/src/rabbit_mirror_queue_sync.erl
@@ -18,7 +18,7 @@
 
 -include("rabbit.hrl").
 
--export([master_prepare/4, master_go/8, slave/7]).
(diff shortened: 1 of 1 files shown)


michaelklishin:
I confirm that with this change, triggering a memory alarm on a mirror node that's undergoing a sync will pause, and then continue when the alarm clears. I used a queue with 2M messages enqueued while the node that hosted a mirror was shut down.

PR #630, opened 2016-02-16
Exit waiting loop once node has unjoined

dmitrymex:
Without the break we always wait for 50 seconds, even if we don't need
to wait at all.


dmitrymex:
@bogdando: please take a look
@michaelklishin: am I right that stable will go into 3.6.1? If yes, then I suggest postpone merging the fix until 3.6.1 is release. The fix is not that important.

michaelklishin:
Correct. Well, it's not much effort to merge this for 3.6.1
but we'll do as you say :)

bogdando:
Fuel CI tests looks good, so LGTM, thank you!

Issue #625, opened 2016-02-15
Windows installer for 3.6.0 and 3.6.1 RC1 fails to correctly install the Windows service


jcasale:
With Windows Server 2008 R2, both installers create a HKEY_LOCAL_MACHINE\SOFTWARE\Ericsson\Erlang\ErlSrv\1.1\RabbitMQ key with Args and Env values missing all data. The service starts but does nothing.

The solution is to start a RabbitMQ Command Prompt and remove then install the service after installation.

hairyhum:
Maybe it's related to fixes in https://github.com/rabbitmq/rabbitmq-server/pull/575 
Look at.
https://github.com/rabbitmq/rabbitmq-server/commit/bc3d5247c38e14e938bf8461889c52b15d188aaa
and
https://github.com/rabbitmq/rabbitmq-server/commit/0f5111cab23e900ccc6794b53f220de594a11c80
And other non log or config related fixes in rabbitmq-service.bat

Gsantomaggio:
Hi @jcasale 
Thank you for reporting. 

I retried the `Windows installer 3.6.1 RC1` on Windows 7, just to be sure, and it works correctly:

 ![reg_windows_7](https://cloud.githubusercontent.com/assets/386987/13078455/517ad9d4-d4bf-11e5-9139-fc0a79e56296.png)

I'm going to try with `Windows 2008 R2`.

hairyhum:
@Gsantomaggio you can also check installation when epmd is not running.

Gsantomaggio:
@hairyhum Yes, thank you! I will.

Gsantomaggio:
Ok I have the same problem.

I noticed that there is and error during the setup :
![error](https://cloud.githubusercontent.com/assets/386987/13083919/5e72e09c-d4d7-11e5-847d-74b3d5532551.png)

I think that the installation didn't finish correctly and the registry is not correct:
![reg_error](https://cloud.githubusercontent.com/assets/386987/13084121/56bd1cae-d4d8-11e5-80a8-12cbf9f5bbde.png)




I stopped the service and put it as `manual` and I restarted the machine.

I got the same error using `rabbitmq-server.bat` as: 

```
C:\Program Files\RabbitMQ Server\rabbitmq_server-3.6.0.901\sbin>rabbitmq-server.
bat
The filename, directory name, or volume label syntax is incorrect.
The filename, directory name, or volume label syntax is incorrect.
[...]

PR #631, opened 2016-02-17
Added erl call to make sure epmd is started

Gsantomaggio:
Fixes https://github.com/rabbitmq/rabbitmq-server/issues/625

got form master commit https://github.com/rabbitmq/rabbitmq-server/commit/0f5111cab23e900ccc6794b53f220de594a11c80


michaelklishin:
I will produce a build for the reporter to try. Thank you, @hairyhum @Gsantomaggio.

PR #637, opened 2016-02-19
Windows: upper case inferred hostname

### LINKED ISSUE #620: 3.6.0 uses different hostname case in RABBITMQ_NODENAME default
Windows (obviously), 64 bit, using 32 bit erlang 18.1.0 and 18.2.1.  I am using %% to surround something to indicate it is an environment variable in this issue.

In at least version 3.5.6 and previous, if %RABBITMQ_NODENAME% was not set, rabbitmq-env.bat would set it to the value of rabbit@%COMPUTERNAME%.  In 3.6.0, it is now being set by shelling out to erlang.  Where %COMPUTERNAME% was returning uppercase, erlang is returning lowercase, so this changes the value.

The impact of this is felt (at least) when running rabbitmq-service.bat install with %RABBITMQ_NODENAME% unset, as it calls rabbitmq-env.bat, which sets %RABBITMQ_NODENAME% to the lowercase version, which sets HKLM\Software\[Perhaps Wow6432Node]\Ericsson\Erlang\ErlSrv\[Service Name]\SName to that value, which then 'breaks' a previously existing node database due to the casing difference in node name with the error message "Cluster upgrade needed" even though there was not a clustered node.

More details as to exactly how this prevents the node from starting can be found here:
https://groups.google.com/d/msg/rabbitmq-users/ykoK-bepR6g/m-qaZIlyBAAJ

michaelklishin:
Prior to 3.6.0, hostname was calculated as
%COMPUTERNAME%, which is uppercase. This has changed in https://github.com/rabbitmq/rabbitmq-server/commit/1fb451090a2557f35a2800f9d04c476ff1ad6a22.

See https://github.com/rabbitmq/rabbitmq-server/issues/620
for details.

Using uppercase isn't inherently better than lowercase but this brings things in line with the `3.5.x` behaviour. We will have to mention this in the change log for those upgrading (overriding `RABBITMQ_NODENAME` avoids the problem entirely).

Fixes #620.

@dumbbell @hairyhum WDYT?

### DIFF
--- a/scripts/rabbitmq-env.bat
+++ b/scripts/rabbitmq-env.bat
@@ -79,7 +79,7 @@ if "!RABBITMQ_NODENAME!"=="" (
         REM !COMPUTERNAME! and Erlang may return different results.
 	REM Start erl with -sname to make sure epmd is started.
 	call "%ERLANG_HOME%\bin\erl.exe" -A0 -noinput -boot start_clean -sname rabbit-prelaunch-epmd -eval "init:stop()." >nul 2>&1
-        for /f "delims=" %%F in ('call "%ERLANG_HOME%\bin\erl.exe" -A0 -noinput -boot start_clean -eval "net_kernel:start([list_to_atom(""rabbit-gethostname-"" ++ os:getpid()), %NAMETYPE%]), [_, H] = string:tokens(atom_to_list(node()), ""@""), io:format(""~s~n"", [H]), init:stop()."') do @set HOSTNAME=%%F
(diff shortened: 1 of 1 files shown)


michaelklishin:
Now that I think of it, perhaps lowercasing was the point of https://github.com/rabbitmq/rabbitmq-server/commit/1fb451090a2557f35a2800f9d04c476ff1ad6a22 and thus the solution for those upgrading from `3.5.x` should be to set `RABBITMQ_NODENAME` explicitly. @dumbbell would know.

michaelklishin:
Closing in favour of the current behaviour, see #620 for details.

PR #639, opened 2016-02-19
Private attributes usage in rabbitmq script

lefremova:
There are three types of rabbitmq attributes for pacemaker nodes:
	-'rabbit-master'
	-'rabbit-start-time'
	- timeouts:
		-'rabbit_list_channels_timeouts'
		-'rabbit_get_alarms_timeouts'
		-'rabbit_list_queues_timeouts'

Attributes with names 'rabbit-master' and 'rabbit-start-time' should be
public because we monitor this attributes in cycle for all nodes in our
script.

All timeouts attributes were changed to private to avoid unnecessary
transitions.

Also, --lifetime and --node options were removed for attrd_updater as
'lifetime' for this command is always 'reboot' and 'node' default value
is local one.

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -1321,18 +1321,18 @@ is_master() {
 # separately. The second argument is used to distingush them.
 check_timeouts() {
     local op_rc=$1
-    local crm_attr_name=$2
+    local timeouts_attr_name=$2
     local op_name=$3
 
     if [ $op_rc -ne 124 -a $op_rc -ne 137 ]; then
-        ocf_run crm_attribute -N $THIS_PCMK_NODE -l reboot --name $crm_attr_name --update 0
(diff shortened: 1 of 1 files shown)


michaelklishin:
@dmitrymex @bogdando can I have your blessing, please? This is the first PR submitted by @lefremova .

bogdando:
Waiting for the patch test results from Fuel CI infra.
Just wondering, could we think of the separate Travis CI job here, see an example job here https://groups.google.com/forum/#!original/rabbitmq-users/BnoIQJb34Ao/eZmrONnpBQAJ

michaelklishin:
@bogdando we would consider it. I can't immediately say where this belongs. Probably `rabbitmq_test` but that eliminates the benefits of having Travis CI set up for pull requests (because it's a different repo).

michaelklishin:
Merged and will be in `3.6.1`, thank you @lefremova.

lefremova:
Ok. Thank you!

Issue #591, opened 2016-01-29
Plugin version constraints


michaelklishin:
All plugins only support so many RabbitMQ server versions, often a single series, e.g. `3.6.x`. Currently there is no mechanism for RabbitMQ broker to check if a plugin is compatible or not, which leads to obscure issues when incompatible plugins are deployed, enabled, and then used.

We should introduce a mechanism for plugin authors (including our own team) to indicate what versions a plugin is compatible with. Incompatible plugins must be logged and ignored.

michaelklishin:
@gotthardp @gmr would love to hear your thoughts on this.

hairyhum:
Since plugins are erlang applications, the most obvious idea is to use application versions and format them in some way. But it can be not enough to determine obsolete plugins.

michaelklishin:
We cannot determine if a plugin is compatible. We should give plugin authors the tools to explicitly say what version range(s) they support, then filter out those that do not match.

hairyhum:
Then we can rely on application version. If it is formatted to some expression, that broker can parse. Maybe not the best idea, because it can violate some standards of application versioning.

hairyhum:
It's possible to define custom application key in `.app.src` with something like `{rabbitmq_versions, ["3.5.6", "3.6.0"]}` wth minimal minor version for each major (major is "3.6"). And then check using `rabbit_misc:version_minor_equivalent` and `rabbit_misc:version_compare`.
This approach is based on assumption that we will not break any plugins by changing minor version.

gmr:
@hairyhum I think that's a good way to go.

PR #642, opened 2016-02-22
Validate plugin compatibility with version ranges

hairyhum:
Proposal for https://github.com/rabbitmq/rabbitmq-server/issues/591

In plugin `.app.src` rabbitmq versions will be defined like `{rabbitmq_versions, ["3.5.4", "3.6.0"]}`, which means that this plugin valid for versions "3.5.4" and higher "3.5.x" and all "3.6.x" versions.
The same format is used for plugin dependencies e.g:
```
{application, rabbitmq_federation_management,[
    ...
    {plugins_versions, [{rabbitmq_management, ["3.7.2"]}, {rabbitmq_federation, ["3.7.0"]}]}
]}.
```
Requires `rabbitmq_management` dependency version to be `3.7.2` or higher for `3.7.x` series and `rabbitmq_federation` version to be any of `3.7.x` series.


hairyhum:
In plugin `.app.src` rabbitmq versions will be defined like `{rabbitmq_versions, ["3.5.4", "3.6.0"]}`, which means that this plugin valid for versions "3.5.4" and higher "3.5.x" and all "3.6.x" versions.

michaelklishin:
Lets start with some automated test examples that'd demonstrate how it works and what kind of edge cases there may be.

hairyhum:
Should we continue with this?

michaelklishin:
@hairyhum I haven't looked at it properly yet, sorry.

hairyhum:
Maybe it makes sense to include version checks for plugins depending on plugins.

hairyhum:
Here are possible error messages:

Rabbitmq version mismatch: 

`[warning] Problem reading some plugins: plugin1: version mismatch, required: ["3.5.0","3.6.0"], got "3.7.0"`

Rabbitmq version is not specified:

`[warning] Some plugins doesn't specify required RabbitMQ version: plugin2, plugin3`

Plugin dependency version do not match: (specified required version for `dep1` and `dep2`, dep1 doesn't exist, `dep2` version do not match requirement)

```
[warning] Some plugin versions do not match:
    plugin1: 
        missing dependency: dep1
        dependency version mismatch: dep2 required ["3.5.6", "3.6.0"], got "3.7.0"
```

Plugin will not be loaded in case of version mismatch. But will be loaded if required version is not specified. There should be some way to inform about it.

@michaelklishin @dumbbell @bdshroyer any ideas on improving messages?

michaelklishin:
"Problem loading a plugin…"
"Some plugin do not specify…"

I also don't think that using a pair of two versions will be very clear to the user.

hairyhum:
Error message format:
```
Failed to enable some plugins: 
    rabbitmq_federation_management:
        Dependency is missing or invalid: rabbitmq_management
    rabbitmq_management:
        Broker version is invalid. Current version: "3.7.0" Required: ["3.6.0"]
```

When running `rabbitmq-plugins enable` or `rabbitmq-plugins set` invalid plugins will not be added to `enabled_plugins` and error is printed.
When plugins (or broker) are updated and server restarts with existing list of plugins - it will print message to log and start without this plugins.

src/rabbit.app.src
```diff
@@ -1,7 +1,7 @@
 {application, rabbit,           %% -*- erlang -*-
  [{description, "RabbitMQ"},
   {id, "RabbitMQ"},
-  {vsn, "0.0.0"},
+  {vsn, "3.7.0"},
```
michaelklishin: This value is updated at build time. @dumbbell do you agree with this default value bump?

hairyhum: Sorry, changed accidentally.

src/rabbit_plugins.erl
```diff
@@ -245,19 +253,100 @@ prepare_plugins(Enabled) ->
...
+        end,
+        [],
+        RequiredVersions),
+    case Problems of
+        [] -> ok;
+        _  -> {error, Problems}
+    end.
+
+version_support(_Version, []) -> ok;
+version_support(Version, ExpectedVersions) ->
```
michaelklishin: I'd recommend renaming this to `is_version_supported` or `check_version_compatibility`.

src/rabbit_plugins.erl
```diff
@@ -245,19 +253,103 @@ prepare_plugins(Enabled) ->
     AllPlugins = list(PluginsDistDir),
     Wanted = dependencies(false, Enabled, AllPlugins),
     WantedPlugins = lookup_plugins(Wanted, AllPlugins),
-
+    {ValidPlugins, Problems} = validate_plugins(WantedPlugins),
+    %TODO: error message formatting
```
michaelklishin: Should we address this?

michaelklishin:
@hairyhum this is on the finish line but there are still two things to do:

 * I had to update the copy. Please check that I haven't made any mistakes.
 * When I configure `rabbitmq_management` to validate `rabbitmq_management_agent` version so that I can trigger a dependency version mismatch error, I get effective version reported incorrectly.

Specifically, with the following added to `rabbitmq_management.app.src`:

```
{dependency_version_requirements, [{rabbitmq_management_agent, ["3.7.0"]}]},
```

and attempt to start the broker with `make run-broker PLUGINS='rabbitmq_management',
the effective version reports doesn't seem correct:

```
Error: Failed to enable some plugins:
    rabbitmq_management:
        Version '[]' of dependency 'rabbitmq_management_agent' is unsupported. Version ranges supported by the plugin: ["3.7.0"]
```

michaelklishin:
Scratch the above, it's due to the fact that in development plugin version was an empty string. When I set it to a realistic value, I get what I expect:

```
Error: Failed to enable some plugins:
    rabbitmq_management:
        Version '"3.4.0"' of dependency 'rabbitmq_management_agent' is unsupported. Version ranges supported by the plugin: ["3.7.0"]
```

So yeah, this seems to be mostly about copy review and a bit more testing, e.g. with a release build.

hairyhum:
I guess we can skip validations if plugin version is empty, assuming it's development version. Or just warn users that there are empty versions, which won't be checked.

michaelklishin:
@hairyhum OK, that sounds like a good idea. Since this is in master, we should also add debug logging for what plugins were checked against what version ranges.

src/rabbit_plugins.erl
```diff
@@ -245,19 +253,105 @@ prepare_plugins(Enabled) ->
...
+                        || #plugin{name = Name, version = Vsn} <- AllPlugins],
+    Problems = lists:foldl(
+        fun({Name, Versions}, Acc) ->
+            case proplists:get_value(Name, ExistingVersions) of
+                undefined -> [{missing_dependency, Name} | Acc];
+                Version   ->
+                    case is_version_supported(Version, Versions) of
+                        true  -> Acc;
+                        false ->
+                            [{{version_mismatch, Version, Versions}, Name} | Acc]
```
michaelklishin: We should rename this to `dependency_version_mismatch` to be more specific.

PR #646, opened 2016-02-24
[OCF HA] Do not check cluster health if master is not elected

dmitrymex:
Doing otherwise causes node to restart when get_monitor is called
within action_promote - it does not find a master and assumes that
it is running out of cluster.

Also, code is refactored a little bit - a new function returning
current master is created and is used in the changed code.

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -752,6 +752,24 @@ get_alive_pacemaker_nodes_but()
     fi
 }
 
+# Get current master. If a parameter is provided,
+# do not check node with that name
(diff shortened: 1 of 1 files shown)


dmitrymex:
@bogdando hey, could you please take a look at this?

michaelklishin:
Is this intentionally submitted for `master`?

dmitrymex:
@michaelklishin: oh, good point, thank you! You are right, that should be in stable

dmitrymex:
Created a new pull request for stable here - https://github.com/rabbitmq/rabbitmq-server/pull/647/files

PR #647, opened 2016-02-24
[OCF HA] Do not check cluster health if master is not elected

dmitrymex:
Doing otherwise causes node to restart when get_monitor is called
within action_promote - it does not find a master and assumes that
it is running out of cluster.

Also, code is refactored a little bit - a new function returning
current master is created and is used in the changed code.


dmitrymex:
@bogdando: please take a look at this

dmitrymex:
I have created a replacement for that PR - #654

PR #648, opened 2016-02-24
Restart option for windows service.

### LINKED ISSUE #645: Restart action for windows service
Related to [this thread](https://groups.google.com/forum/#!topic/rabbitmq-users/1W-zFNOuitk).
There are `OnFail` argument to `erlsrv` which defines the restart behaviour of windows service. Default is `ignore` but we could use some better parameter like `restart`, which will restart the service with maximum restart frequency 10 sec.

hairyhum:
FIxes #645


michaelklishin:
Lets wait until `3.6.1` is released before merging this.

Gsantomaggio:
Tried by killing `erl.exe` and the process restart correctly. 


Just for more clarification:
1.  `OnFail` flag ignores the `Windows service recovery setting` , right?

2.  By default now  `rabbitmq-service.bat` uses:

```
if "!RABBITMQ_SERVICE_RESTART!"=="" (
     set RABBITMQ_SERVICE_RESTART=restart
)
```

Which are the other values for `RABBITMQ_SERVICE_RESTART` ? 
We should document it here https://www.rabbitmq.com/configure.html

michaelklishin:
Lets wait and see if this default works well. I have a feeling very few people would use something other than "restart".

hairyhum:
@Gsantomaggio yes, it ignores service recovery settings. They seem like not working when `erl` crashes anyway.
Other values for restart flag are described [in erlsrv doc](http://erlang.org/doc/man/erlsrv.html)
```
OnFail: This can be either of reboot, restart, restart_always or ignore (the default). In case of reboot, the NT system is rebooted whenever the emulator stops (a more simple form of watchdog), this could be useful for less critical systems, otherwise use the heart functionality to accomplish this. The restart value makes the Erlang emulator be restarted (with whatever parameters are registered for the service at the occasion) when it stops. If the emulator stops again within 10 seconds, it is not restarted to avoid an infinite loop which could completely hang the NT system. restart_always is similar to restart, but does not try to detect cyclic restarts, it is expected that some other mechanism is present to avoid the problem. The default (ignore) just reports the service as stopped to the service manager whenever it fails, it has to be manually restarted.
```

Gsantomaggio:
Thank you @hairyhum

michaelklishin:
@Gsantomaggio if I understood your comment correctly, this PR works as expected. Merging, feel free to revert if that's not the case.

Issue #398, opened 2015-10-30
Add rabbitmqctl command for node health check


binarin:
In Vancouver Michael has mentioned one-shot command for performing health-check on rabbit node.

I don't know what was the idea then, but here is the list of checks from OCF script:
- BEAM is running, 'kernel' and 'rabbit' OTP applications are running.
- Checks (with timeout) that `rabbitmqctl list_channels` works and returns something.
- Checks whether memory alarm is set for this node.
- Checks (with timeout) if queues can be listed with `rabbitmqctl list_queues`.

So the new command should collect at least the same date, but in a less kludgy way. And maybe extended with additional checks, if somebody would come up with them )

michaelklishin:
@binarin that sound OK but should perhaps be combined with what the aliveness check HTTP API resource does: opens a connection, declares a queue, publishes one messages and consumes it with `basic.get`.

This would be a decent health check for a single node.

jiajunsu:
I think it's better adding a rabbitmqctl cli instead the management aliveness check.

If server's health-check depend on management plugin, when management got crash for an internal error, we could not know whether the server is fine.

michaelklishin:
I'm growing convinced that there should be two improvements made:

 * Single node health check improvements plus a `rabbitmqctl` command
 * Cluster state health check (the specifics are not yet clear), which should be a separate issue

dcorbacho:
I implemented a first version of the rabbitmqctl command in: https://github.com/rabbitmq/rabbitmq-server/tree/rabbitmq-server-398.

However, there is an issue if we want to include what the aliveness check HTTP API does, as it requires the amqp_client application. Neither rabbit nor rabbit_common depend on it, and I don't think we want to add that dependency.

@michaelklishin what are your thoughts now?

michaelklishin:
I'd rather avoid the dependency. Declaring a topology and publishing messages using the internal RabbitMQ API is easy, I am less sure about consuming messages.

@binarin's list of features for this check does not involve publishing messages, and many people don't find that particular check important (at least the way it works today). So we can even leave publishing out if it proves to be difficult without a client library.

binarin:
Yes, I no longer think that just publishing something is a useful test. Especially without taking into account a lot of site-specific thing, like whether clustering or HA is used.

And problems are usually more subtle and involve existing queues/channels.

dcorbacho:
This change is made of 3 PRs:
[...]

PR #652, opened 2016-02-25
Add rabbitmqctl command for node health check

dcorbacho:
Part of #398

### DIFF
--- a/src/rabbit_control_main.erl
+++ b/src/rabbit_control_main.erl
@@ -25,7 +25,7 @@
 -import(rabbit_cli, [rpc_call/4, rpc_call/5, rpc_call/7]).
 
 -define(EXTERNAL_CHECK_INTERVAL, 1000).
--define(ALIVENESS_TIMEOUT, 70000).
+-define(NODE_HEALTH_CHECK_TIMEOUT, 70000).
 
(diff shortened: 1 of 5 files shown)


src/rabbit_control_main.erl
```diff
@@ -547,6 +549,16 @@ action(eval, Node, [Expr], _Opts, _Inform) ->
...
     io:format("~s", [rabbit_ctl_usage:usage()]);
 
+action(node_health_check, Node, _Args, _Opts, Inform) ->
+    Inform("Health check of node ~p", [Node]),
+    try
+        rabbit_health_check:node(Node),
+        io:format("Node ~p is up and running~n", [Node])
+    catch
+        {node_is_ko, ErrorMsg} ->
+            io:format("~s~nProblems encountered in node ~p~n", [ErrorMsg, Node])
```
michaelklishin: @binarin what UNIX return code should we use when a check fails?

binarin: Currently there is only one possible reason - `node_is_ko`. EX_SOFTWARE (70) suits it the best.

And is it possible to add separate reason with exit code EX_NOHOST (68) for cases when a node is unavailable?

michaelklishin: @binarin good idea

dcorbacho:
@michaelklishin Just added the unix error codes to rabbitmqctl

PR #653, opened 2016-02-26
[OCF HA] Log process id in RabbitMQ OCF script

dmitrymex:
Several OCF calls might run simultaneously. For example, it often
happens that two monitor calls intersect. Logging current process id
for each line helps distinguish logs of different calls.

Also aligned get_status() logging with format used in all other
parts of the script.

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -613,7 +613,7 @@ rmq_setup_env() {
         fi
     done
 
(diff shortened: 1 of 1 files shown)


dmitrymex:
@bogdando: please take a look

michaelklishin:
@bogdando @dmitrymex should we merge this?

bogdando:
LGTM, good point, thank you!

PR #657, opened 2016-02-29
[OCF HA] On neighbor promotion do nothing if we are already clustered

dmitrymex:
[OCF HA] On neighbor promotion do nothing if we are already clustered

 + extracted function checking if we are in the same cluster with
   given node

 + made post-promote ignore promotion of self. Previously it was
   done inside jjj_join, but now we need to do that before the
   new check.

 + now we write "post-promote end" log entry at the very
   end of post-promote, not somewhere in the middle.

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -769,6 +769,13 @@ get_master_name_but()
     done
 }
 
+# Returns 0 if we are clustered with provideded node
+is_clustered_with()
+{
+    get_running_nodes | grep -q $(rabbit_node_name $1);
+    return $?
+}
+
 
 check_need_join_to() {
     local join_to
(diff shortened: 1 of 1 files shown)


dmitrymex:
please note that this pull request is based on #654, which must be merged first

michaelklishin:
@dmitrymex @bogdando if you want this as well as #653, #654 to make it into `3.6.1`, we have to get a go-ahead from you and merge them today.

dmitrymex:
@bogdando : could you please take a look?

dmitrymex:
@michaelklishin : Thanks for letting us know. These are non-critical enhancements, so it is not that important for them to get into 3.6.1

dmitrymex:
Update: that PR is based on #658 , the previous PR - #654 was made to master

bogdando:
LGTM

Issue #443, opened 2015-11-22
unit test failure introduced in #439


michaelklishin:
Running unit tests now sporadically fails with, which seems directly related to #439.

```
Error: {error, {badmatch,
 {badrpc,
 {'EXIT',
 {badarg,
 [{ets,match,
 [ranch_server,{{conns_sup,'_'},'$1'}],
 []},
 {rabbit_vm,conn_sups,0,
 [{file,"src/rabbit_vm.erl"},{line,138}]},
 {rabbit_vm,interesting_sups,0,
 [{file,"src/rabbit_vm.erl"},{line,129}]},
 {rabbit_vm,memory,0,
 [{file,"src/rabbit_vm.erl"},{line,37}]},
 {rabbit,status,0,
 [{file,"src/rabbit.erl"},{line,440}]},
 {rpc,local_call,3,
 [{file,"rpc.erl"},{line,330}]},
 {rabbit_control_main,display_call_result,2,
 [{file,"src/rabbit_control_main.erl"},
 {line,738}]},
 {rabbit_tests,control_action,4,
 [{file,"test/src/rabbit_tests.erl"},
 {line,1841}]}]}}}}}
```

I cannot reproduce this with Java or Bunny test suites, or manually.

essen:
Guess it can happen if the function is called before the table was created, I'll check on Tuesday.

Issue #433, opened 2015-11-19
In Windows RABBITMQ_PLUGINS_DIR default value is wrong if RABBITMQ_BASE is set


ilmirajat:
I tried to install RabbitMQ to windows server so that RABBITMQ_BASE was set to non-default folder (in this case "D:\ProgramFiles\RabbitMQServer-Data"). The rest of the RabbitMQ was installed to the default folder in this case "C:\Program Files (x86)\RabbitMQ Server\rabbitmq_server-3.5.6".

Attempt to install management plugin failed. Rabbitmq-service.bat did not expand correctly ez-files to the expanded-folder under Mnesia-folder.

Solution was to set RABBITMQ_PLUGINS_DIR environment variable explicitly. (In this case I also copied plugins to another folder, because instructions warned that in some case windows folders names causes problems.) 

The problem seems to be in rabbitmq-env.bat:

```
if "!RABBITMQ_PLUGINS_DIR!"=="" (
	if "!PLUGINS_DIR!"=="" (
		set RABBITMQ_PLUGINS_DIR=!RABBITMQ_BASE!\plugins
	) else (
		set RABBITMQ_PLUGINS_DIR=!PLUGINS_DIR!
	)
)
```

If RABBITMQ_PLUGINS_DIR or PLUGINS_DIR is not set, then the path would be RABBITMQ_BASE\plugins i.e. in my case D:\ProgramFiles\RabbitMQServer-Data\plugins"). However it should be "C:\Program Files (x86)\RabbitMQ Server\rabbitmq_server-3.5.6\plugins".

Probably line 3 above should be:

```set RABBITMQ_PLUGINS_DIR=!RABBITMQ_HOME!\plugins```

(Because in the beginning of the file ```set RABBITMQ_HOME=%SCRIPT_DIR%..```)

Currently, the service I needed to install works as it should and I don't have time to do pull request right now. I posted this issue and solution, just in case someone else have the same problem. If some one else have not fixed this, I'll do pull request once I had have time to test this solution in a clean environment.

michaelklishin:
Your understanding [seems to be correct](https://github.com/rabbitmq/rabbitmq-server/blob/master/scripts/rabbitmq-defaults#L42), thank you.

michaelklishin:
Fixed in #444.

PR #660, opened 2016-02-29
Create rabbitmq.config on windows installation

### LINKED ISSUE #659: Config is not used by windows service if not present during installation.
If `rabbitmq.config` file doesn't exist during windows service installation, it will not be used when created, even after service restart.  
It makes sense to create file during installation if it doesn't exist.

hairyhum:
Fixes #659

### DIFF
--- a/packaging/windows-exe/rabbitmq_nsi.in
+++ b/packaging/windows-exe/rabbitmq_nsi.in
@@ -140,6 +140,15 @@ Section "RabbitMQ Server (required)" Rabbit
   ; ...And put the example config file there
   File "rabbitmq_server-%%VERSION%%\etc\rabbitmq.config.example"
 
+  IfFileExists rabbitmq.config 0 +2
+    Goto config_written
+    ClearErrors
+    FileOpen $0 rabbitmq.config w
+    IfErrors config_written
+    FileWrite $0 "[]."
+    FileClose $0
+  config_written:
+
   ; Write the installation path into the registry
   WriteRegStr HKLM "SOFTWARE\VMware, Inc.\RabbitMQ Server" "Install_Dir" "$INSTDIR"
 
--- a/packaging/windows-exe/rabbitmq_nsi.in
+++ b/packaging/windows-exe/rabbitmq_nsi.in
@@ -140,15 +140,6 @@ Section "RabbitMQ Server (required)" Rabbit
   ; ...And put the example config file there
   File "rabbitmq_server-%%VERSION%%\etc\rabbitmq.config.example"
 
-  IfFileExists rabbitmq.config 0 +2
-    Goto config_written
-    ClearErrors
-    FileOpen $0 rabbitmq.config w
-    IfErrors config_written
-    FileWrite $0 "[]."
-    FileClose $0
-  config_written:
-
   ; Write the installation path into the registry
   WriteRegStr HKLM "SOFTWARE\VMware, Inc.\RabbitMQ Server" "Install_Dir" "$INSTDIR"
 
--- a/scripts/rabbitmq-service.bat
+++ b/scripts/rabbitmq-service.bat
@@ -131,6 +131,12 @@ if ERRORLEVEL 3 (
     set RABBITMQ_DIST_ARG=-kernel inet_dist_listen_min !RABBITMQ_DIST_PORT! -kernel inet_dist_listen_max !RABBITMQ_DIST_PORT!
 )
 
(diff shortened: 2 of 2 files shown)


Gsantomaggio:
I made two tests

1. Setup using the standard configuration and the `rabbitmq.config`is located inside the folder `C:\Users\gabriele\AppData\Roaming\RabbitMQ\`.  **OK**


2.  removed all and reinstalled the setup using `RABBITMQ_BASE`as `c:\test_rmq` and the file `rabbitmq.config` has been created inside  `C:\Users\gabriele\AppData\Roaming\RabbitMQ\`

```
=INFO REPORT==== 3-Mar-2016::16:14:25 ===
node           : rabbit@windowsdev
home dir       : C:\Windows
config file(s) : c:/test_rmq/rabbitmq.config (not found)
```
In this case the file should be created inside `c:/test_rmq/`, isn't it ?

hairyhum:
@Gsantomaggio installer has no information about `RABBITMQ_BASE`. Maybe file creation should be moved to `rabbitmq-service.bat`?

Gsantomaggio:
@hairyhum Yes I think it should be moved.

I'd add that when `RABBITMQ_BASE`is configured the directory `AppData\Roaming\RabbitMQ` shouldn't be created.

Even if `RABBITMQ_BASE`is configured,  the setup creates:

```
Directory di C:\Users\gabriele\AppData\Roaming\RabbitMQ

03/03/2016  16:14    <DIR>          .
03/03/2016  16:14    <DIR>          ..
03/03/2016  16:14    <DIR>          db
03/03/2016  16:14    <DIR>          log
```
The directories are empty.

Gsantomaggio:
@hairyhum what do you think ?

hairyhum:
@Gsantomaggio  In theory it's possible to consider `RABBITMQ_BASE` during installation, but it's not about this issue, I guess.
It still can be discussed here: https://github.com/rabbitmq/rabbitmq-server/issues/626

Gsantomaggio:
@hairyhum I got it, thanks. 
I'd move the file creation to `rabbitmq-service.bat` to close the this issue. 

For the other points I will refer to the #626

Gsantomaggio:
`echo []. > !RABBITMQ_CONFIG_FILE!.config` raises `The system cannot find the path specified.` if the variable `RABBITMQ_CONFIG_FILE`  points to a directory that does not exist.

I tried leaving the default configuration and setting  `RABBITMQ_CONFIG_FILE=C:\test_pull660_1\test_file_conf`, where  `C:\test_pull660_1\` does not exist. I got the error during the setup, and the error inside the log:
`config file(s) : c:/test_pull660_1/test_file_conf.config (not found)`

I read the [comment](https://github.com/rabbitmq/rabbitmq-server/commit/5d0c0c7932cc43dc8387baccdf9a7d5046ee8ac2#diff-d0fc4e42f493273ee25c52fd15d7149bR135), I want to be sure that this is the expected behavior or we should try to create the directory as we do for [RABBITMQ_BASE](https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-659/scripts/rabbitmq-env.bat#L56)

hairyhum:
I thought that file directory should exist for default file location. We can try to create directory for this file, but I don't know how to get directory for non-existent file in CMD.

Gsantomaggio:
> `RABBITMQ_CONFIG_FILE`  is The path to the configuration file.

I think that most of the users use the default configuration or they change `RABBITMQ_BASE`.

So, we have two options:

1. Ignore this case, and we suppose that who changed this setting knows what is doing.
2. Find a way to create the directory form. 

Except for this particular situation, the setup works correctly now.

hairyhum:
The initial idea was to make default installation work with default config file location, so I think we can ignore this case.

dls314:
@Gsantomaggio, @hairyhum do you know if there is/was another issue about having the installer use environment values for RABBITMQ_HOME?

I'm glad for this discussion regardless, I just spent a bit finding out/confirming that it doesn't :-\

Gsantomaggio:
@dls314 do you mean  `RABBITMQ_BASE` in Windows ?

maybe this https://github.com/rabbitmq/rabbitmq-server/issues/433

dls314:
@Gsantomaggio I do mean in Windows. Eventually I installed the service by hand to fix my problems and unchecked those options (services and links in the start menu) from the installer.

I'll spin up a new issue I think, #443 seems to be more about the plugins and batch files, and they all seem to work correctly.

PR #661, opened 2016-02-29
Update rabbitmqctl manual (-p)

deadtrickster:
Commands: `sync_queue`, `cancel_sync_queue`, `purge_queue` also accept vhostpath.

Also question about `list_connections` -
looks like it [ignores options ](https://github.com/rabbitmq/rabbitmq-server/blob/023fd3f48a0eb74b9322a7687c2db282350640a3/src/rabbit_control_main.erl#L622) but defined as `{list_connections, [?VHOST_DEF]}`. Am I missing something?

And website probably could be updated too :-).

### DIFF
--- a/docs/rabbitmqctl.1.xml
+++ b/docs/rabbitmqctl.1.xml
@@ -535,7 +535,7 @@
           </listitem>
         </varlistentry>
         <varlistentry>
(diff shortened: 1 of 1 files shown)


michaelklishin:
Yeah, that might be an unintentional issue with `list_connections`.

Please submit this PR against `stable`.

PR #662, opened 2016-02-29
Added the code to list the nodes that are down via the rabbitmqctl sc…

epretha:
…ript


michaelklishin:
Closing per request from @Ayanda-D.

PR #666, opened 2016-03-02
Systemd improvements

### LINKED ISSUE #664: Pass correct OS PID to systemd-notify
See the discussion in https://github.com/rabbitmq/rabbitmq-server/pull/573.

binarin:
After some investigation I think that even with `sd_notify` MAINPID
detection didn't work 100% percent correctly.

Foreground broker was started through a shell process that was
converting signals into graceful shutdown commands. And it's pid of that
proxy that was to be communicated to systemd. Automatic pid inference
worked in case epmd was already running on machine, otherwise it was epmd
who was chosen as main process.

This commit introduces new environment variable to prevent creation of
such a proxy process. And this var should be specified in unit
file (example added to docs/). New variable is needed due to following
reasons:
- No way to detect whether we are running under systemd or not
- I was thinking about adding command-line switch instead, but there is
  no way to work with "$@" in portable shell script.

Another improvement is writing message to stderr about graceful
shutdown, just to be sure that shutdown was indeed graceful. In case of
systemd it will automatically appear in the journal.

Fixes #664

### DIFF
--- a/docs/rabbitmq-server.service.example
+++ b/docs/rabbitmq-server.service.example
@@ -1,0 +1,17 @@
+# systemd unit example
+[Unit]
+Description=RabbitMQ broker
+After=syslog.target network.target
+
+[Service]
+Type=notify
+User=rabbitmq
+Group=rabbitmq
+NotifyAccess=all
+TimeoutStartSec=3600
+WorkingDirectory=/var/lib/rabbitmq
+ExecStart=/usr/lib/rabbitmq/bin/rabbitmq-server
+ExecStop=/usr/lib/rabbitmq/bin/rabbitmqctl stop
+
+[Install]
+WantedBy=multi-user.target
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -322,6 +322,15 @@ sd_notify_legacy() ->
             false
     end.
 
+%% socat(1) is the most portable way the sd_notify could be
+%% implemented in erlang, without introducing some NIF. Currently the
+%% following issues prevent us from implementing it in a more
+%% reasonable way:
+%% - systemd-notify(1) is unstable for non-root users
+%% - erlang doesn't support unix domain sockets.
+%%
+%% Some details on how we ended with such a solution:
+%%   https://github.com/rabbitmq/rabbitmq-server/issues/664
 sd_notify_socat() ->
     case sd_current_unit() of
         {ok, Unit} ->
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -374,7 +374,14 @@ sd_current_unit() ->
     end.
 
 sd_wait_activation(Port, Unit) ->
-    sd_wait_activation(Port, Unit, 10).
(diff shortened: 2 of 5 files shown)


michaelklishin:
@binarin this no longer merges cleanly, can you please rebase? Thank you.

docs/rabbitmq-server.service.example
```diff
@@ -0,0 +1,17 @@
...
+[Unit]
+Description=RabbitMQ broker
+After=syslog.target network.target
+
+[Service]
+Type=notify
+User=rabbitmq
+Group=rabbitmq
+NotifyAccess=all
+TimeoutStartSec=3600
```
dumbbell: I see you added a

src/rabbit.erl
```diff
@@ -284,16 +284,94 @@ broker_start() ->
...
+            sd_notify_socat(Unit);
+        _ ->
+            false
+    end.
+
+sd_notify_socat(Unit) ->
+    case catch open_port(
+                 {spawn_executable, os:find_executable("socat")},
+                 [{args, ["unix-sendto:" ++ os:getenv("NOTIFY_SOCKET"), "STDIO"]},
+                  exit_status, use_stdio, out]) of
```
dumbbell: You never call `port_close/1`, how is socat(1) termianted?

src/rabbit.erl
```diff
@@ -284,16 +284,94 @@ broker_start() ->
...
+sd_notify_socat() ->
+    case sd_current_unit() of
+        {ok, Unit} ->
+            io:format(standard_error, "systemd unit for activation check: \"~s\"~n", [Unit]),
+            sd_notify_socat(Unit);
+        _ ->
+            false
+    end.
+
+sd_notify_socat(Unit) ->
```
michaelklishin: Please add a comment linking to #664. It's not obvious why we have to use `socat` and other things that are going on here.

docs/rabbitmq-server.service.example
binarin: Recently I'm very suspicious of `infinity` timeouts. I don't remember who said something like: 
> When I see a timeouf of `inifinity`, I suggest to replace it with some insane value like 30 years. And if the reply to this is that 30 years doesn't make sense, I reply that infinity also makes no sense.

With `infinity` timeout some problem with startup could go unnoticed for a long period of time. The question is what is the sane timeout value - in my opininon it's no more than several hours.

src/rabbit.erl
binarin: I saw that it terminated and never asked that question. This happens unintentionally due to controlling process death. I'm going to make it explicit.

docs/rabbitmq-server.service.example
dumbbell: I discussed this with @michaelklishin and he agrees with you, so we'll keep this timeout of one hour.

dumbbell:
Thank you @binarin!

Issue #671, opened 2016-03-02
Move rabbit_types, rabbit_auth_backend_* modules to rabbitmq-common


bdshroyer:
These modules are good candidates for rabbitmq-common because they are referenced in multiple repos, including rabbitmq-common itself in some cases. Moving them to rabbitmq-common would also have the side effect of making them easier to import into the new CLI.

PR #672, opened 2016-03-02
Move auth_backend and types modules to rabbitmq-common.

bdshroyer:
* Delete these modules from rabbitmq-server and add them to rabbitmq-common.

Addresses issue [#671](https://github.com/rabbitmq/rabbitmq-server/issues/671).


michaelklishin:
Please submit this against `stable`.

PR #674, opened 2016-03-03
Verify that mandatory environment variables are (pre)set

### LINKED ISSUE #553: Scripts should verify presence of mandatory environment variables
Related to this [mailing list thread](https://groups.google.com/forum/#!topic/rabbitmq-users/1OgdpOgsLdc)

Some variables required by start script are being set from defaults if empty. But if defaults are empty as well - there are no checks and validations. So `erl` process can fail with unrelated exception. 

Maybe there should be validation of required variables just before running `erl` and reporting to user?

hairyhum:
Fixes #553 
Exits with `78` if required parameters are empty.


scripts/rabbitmq-server
```diff
@@ -151,6 +154,24 @@ stop_rabbitmq_server() {
...
+    check_not_empty "RABBITMQ_BOOT_MODULE" $RABBITMQ_BOOT_MODULE
+    check_not_empty "RABBITMQ_NAME_TYPE" $RABBITMQ_NAME_TYPE
+    check_not_empty "RABBITMQ_NODENAME" $RABBITMQ_NODENAME
+    check_not_empty "SASL_BOOT_FILE" $SASL_BOOT_FILE
+    check_not_empty "RABBITMQ_IO_THREAD_POOL_SIZE" $RABBITMQ_IO_THREAD_POOL_SIZE
+}
+
+check_not_empty() {
+    if [ "x" = "x$2" ]; then
+        echo "Error: ENV variable should be defined: $1.
```
michaelklishin: So this check is for us, or the user? If it's the latter, we should change this message to mention `rabbitmq-env.conf` and environment variables.

michaelklishin: OK, I assume this is for us.

hairyhum: It's not only for us. I think mentioning the env variable possible source is a good idea.

scripts/rabbitmq-server
```diff
@@ -151,6 +154,24 @@ stop_rabbitmq_server() {
...
 
+check_start_params() {
+    check_not_empty "RABBITMQ_BOOT_MODULE" $RABBITMQ_BOOT_MODULE
+    check_not_empty "RABBITMQ_NAME_TYPE" $RABBITMQ_NAME_TYPE
+    check_not_empty "RABBITMQ_NODENAME" $RABBITMQ_NODENAME
+    check_not_empty "SASL_BOOT_FILE" $SASL_BOOT_FILE
+    check_not_empty "RABBITMQ_IO_THREAD_POOL_SIZE" $RABBITMQ_IO_THREAD_POOL_SIZE
+}
+
+check_not_empty() {
```
binarin: It's possible to avoid duplicating variable name at a call site with a function body like this:
```shell
local name="${1:?}"
local value
eval value=\$$name
if [ -z "$value" ]; then
# error 
....
```

scripts/rabbitmq-service.bat
```diff
@@ -188,6 +188,16 @@ set ERLANG_SERVICE_ARGUMENTS= ^
 set ERLANG_SERVICE_ARGUMENTS=!ERLANG_SERVICE_ARGUMENTS:\=\\!
 set ERLANG_SERVICE_ARGUMENTS=!ERLANG_SERVICE_ARGUMENTS:"=\"!
 
+set ENV_OK=true
```
Gsantomaggio: I had  to move this code [here](https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-659/scripts/rabbitmq-service.bat#L106) (Didn't commit yet). Because [here](https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-553/scripts/rabbitmq-service.bat#L191) the service is already installed.
On my tests work better, Makes sense for you?

scripts/rabbitmq-service.bat
```diff
@@ -188,6 +188,16 @@ set ERLANG_SERVICE_ARGUMENTS= ^
 set ERLANG_SERVICE_ARGUMENTS=!ERLANG_SERVICE_ARGUMENTS:\=\\!
 set ERLANG_SERVICE_ARGUMENTS=!ERLANG_SERVICE_ARGUMENTS:"=\"!
 
+set ENV_OK=true
+CALL :check_not_empty "RABBITMQ_BOOT_MODULE" "!RABBITMQ_BOOT_MODULE!" 
+CALL :check_not_empty "RABBITMQ_NAME_TYPE" "!RABBITMQ_NAME_TYPE!"
+CALL :check_not_empty "RABBITMQ_NODENAME" "!RABBITMQ_NODENAME!"
```
Gsantomaggio: I removed the double quotes also here, for the same reason I did [here] (https://github.com/rabbitmq/rabbitmq-server/commit/053e798b312a01ada29c17cc101fa140f3ae9753)

hairyhum:
Removed debug echo from shell script and added unquoting to bat scripts.

Gsantomaggio:
looks good to me

PR #680, opened 2016-03-09
Disable erlang break handler in rabbitmqctl

binarin:
So rabbitmq operator will not be baffled by erlang break handler when he
tries to interrupt some long-running `rabbitmqctl` command by `Ctrl-C`.


michaelklishin:
@binarin thank you!

PR #683, opened 2016-03-10
Listing items in parallel

binarin:
I need some feedback before proceeding further. This is proof-of-concept, only `list_channels` is working currently. If this is an acceptable change, I'll going to fix all the other commands.

Current implementation of various `list_XXX` commands require cross-node
roundtrip for every processed item - because `rabbitmqctl` target node
is responsible for gathering global list of all items of
interest (channels etc.) and then processing them one by one.

For example, listing 10000 channels evenly distributed across 3 nodes
where network has 1ms delay takes more than 10 seconds on my
machine. And with the proposed change listing will take almost the same
time as it'll take to gather this info locally. E.g. in the case above
listing now takes 0.7 second on the same machine with same 1ms delay.

It works by invoking emitting_map on every node, where it should send
info about only local items to aggregator, in an async fashion - as no
reply from aggregator is needed.

### DIFF
--- a/src/rabbit_control_main.erl
+++ b/src/rabbit_control_main.erl
@@ -23,7 +23,7 @@
          sync_queue/1, cancel_sync_queue/1, become/1,
          purge_queue/1]).
 
--import(rabbit_misc, [rpc_call/4, rpc_call/5, rpc_call/7]).
+-import(rabbit_misc, [rpc_call/4, rpc_call/5]).
 
 -define(EXTERNAL_CHECK_INTERVAL, 1000).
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Speeding up `list_*` operations would be great.

src/rabbit_control_main.erl
```diff
@@ -841,16 +848,30 @@ call(Node, {Mod, Fun, Args}, InfoKeys, ToBinUtf8, Timeout) ->
...
 call(Node, {Mod, Fun, Args}, InfoKeys, ToBinUtf8, Timeout, IsEscaped) ->
-    Args0 = case ToBinUtf8 of
-                true  -> lists:map(fun list_to_binary_utf8/1, Args);
-                false -> Args
-            end,
+    Args0 = prepare_call_args(Args, ToBinUtf8),
+    Ref = make_ref(),
+    Pid = self(),
+    spawn_emitter_caller(Node, Mod, Fun, Args0, Ref, Pid, Timeout),
+    rabbit_control_misc:wait_for_info_messages(
```
dcorbacho: Note that this call is now broken

dcorbacho:
@binarin The change seems a good idea, I didn't verify the times but I'll trust you on that. 

The calls that determine if the data will be emitted from one or several nodes are indeed in the remote node. Any rabbitmqctl command that doesn't need this optimisation, would be the equivalent to your changes with Chunks == 1. I think that you can do all that work with minimal changes to the code:

- rabbit_control_main:call_emitter is not needed if rabbit_control_main:call is expanded to accept one more parameter that is the list of nodes. It would default to the current target node, so any other command would still work.
- Then, it would only require a function that  appends Pid, Ref and Nodes to the Args if length(nodes) > 1, just before call rpc_cli:rpc_call/7.
- rabbit_control_misc:wait_for_info_messages receives always the number of chunks (1-N)

In this way, we only have to maintain one set of `rabbit_control_main:call` functions

binarin:
I've decided to introduce `call_emitter` due to following reasons:
- Less confusion, it took me some time to discover that emitting versions of `rabbit_channel:info_all` etc. are indeded used. And because `rabbit_control_main:call/2` is completely different function, which should be called something like `call_utf8_args` - but I'm leaving it for some later PR.
- Less confusion, again. Functions with more than 3 or 4 arguments are usually very confusing.

So my current plan is replace all calls to `rabbit_control_main:call/4`, `call/5` and `call/6` with a call to the new `call_emitter/4`.

binarin:
Done.

I have my own test-suite for this issue at https://github.com/binarin/rabbit_destructive_tests - but it requires linux and password-less `sudo` for `tc` command. Still, it allows anybody to see that promised performance improvements were indeed achived. And I have some distant plans on re-doing this tests without `sudo`/`tc` dependency, using only pure erlang - then those tests could be incorporated into `rabbitmq-test`.

michaelklishin:
@binarin do you have any numbers to share from that test? I'm mostly being curious, no objections to making hot code paths of the `list_*` commands parallel.

binarin:
I have following test setup:
- single physical machine: i5 at 3.3 GHz, 16 GB RAM
- 3-node cluster connected through loopback interface
- 15000 queues evenly distributed across those 3 nodes
- 10000 channels on 50 connections evenly distributed across those 3 nodes.
- `delayed` in the table below means introducing 1ms delay on loopback interface.
- all results are in seconds

| command | original, w/o delay | original, delayed | patched, w/o delay | patched, delayed |
| ------ | ------ | ------ | ------- | -------- |
| list_channels | 1.2 | 15.5 | 0.78 | 0.82 |
| list_queues | 3.6 |  24.4 | 0.94 | 1.07 |

binarin:
WIP again, I've found some issues with current version of this code.

binarin:
Now I believe it's really done )

Solution is somewhat complex, but performance gain is worth it - there will be significantly less false positives when list_XXX commands are used as a health checks.

hairyhum:
@binarin can `wait_for_info_messages` be a `fold` function to collect values, instead of just calling `DisplayFun`?
This will make testing easier and provide a way to use different output formats in future.

binarin:
Different output formats already can be supported by providing different display fun.

This patch is primarily targeted at a cases where there are thousands and thousands of items to be listed. And there are several reason to display items immediately, without collection:
- Overall run-time - with intermediate collection step, rabbitmqctl will be sitting mostly idle initially, and will start formatting (i.e. actively using CPU) only after receiving everything. With immediate display this load will be evenly distributed, and rabbitmqctl will exit almost immediately after receiving last item.
- Memory consumption - collection of tens of thousands info items can be relatively costly in terms of RAM
- Interactively running lines will make ops people happy - there will be no way to suspect that something is stuck when you see some running lines on your terminal :smile:

hairyhum:
Fold function should allow to use current logic as well, if accumulator returned from `DisplayFun` will always be `ok` for example.

binarin:
Aha, got it. Will amend it in several hours.

binarin:
I've updated the patches.

hairyhum:
Unit tests are failing with
```
{badrpc, {'EXIT', {{nocatch, {bad_argument, dummy}}, [{rabbit_amqqueue, info, 2, [{file, "src/rabbit_amqqueue.erl"}, {line, 599}]}, {rabbit_misc, with_exit_handler, 2, [{file, "src/rabbit_misc.erl"}, {line, 501}]}, {rabbit_control_misc, step_with_exit_handler, 4, [{file, "src/rabbit_control_misc.erl"}, {line, 70}]}, {rabbit_control_misc, '-emitting_map0/5-lc$^0/1-0-', 5, [{file, "src/rabbit_control_misc.erl"}, {line, 62}]}, {rabbit_control_misc, emitting_map_with_exit_handler, 5, [{file, "src/rabbit_control_misc.erl"}, {line, 58}]}, {rabbit_control_misc, emitting_map_with_exit_handler, 4, [{file, "src/rabbit_control_misc.erl"}, {line, 53}]}]}}}
```

Command with invalid info keys never completes.

hairyhum:
Running command with invalid info keys from command line
```
$ scripts/rabbitmqctl list_queues dummy
Listing queues ...
Error: {{nocatch,{bad_argument,dummy}},
        [{rabbit_amqqueue,info,2,
                          [{file,"src/rabbit_amqqueue.erl"},{line,599}]},
         {rabbit_misc,with_exit_handler,2,
                      [{file,"src/rabbit_misc.erl"},{line,501}]},
         {rabbit_control_misc,step_with_exit_handler,4,
                              [{file,"src/rabbit_control_misc.erl"},
                               {line,70}]},
         {rabbit_control_misc,'-emitting_map0/5-lc$^0/1-0-',5,
                              [{file,"src/rabbit_control_misc.erl"},
                               {line,62}]},
         {rabbit_control_misc,emitting_map_with_exit_handler,5,
                              [{file,"src/rabbit_control_misc.erl"},
                               {line,58}]},
         {rabbit_control_misc,emitting_map_with_exit_handler,4,
                              [{file,"src/rabbit_control_misc.erl"},
                               {line,53}]}]}
```

binarin:
Somewhat unexpected side effect of making wait_for_info_messages/5 fold-like. I intentionally converted returning error to exit at https://github.com/binarin/rabbitmq-common/commit/5aae3c274ac8c41c143520a0d9450481c9c34df1#diff-b3c3007fa0d605476c3d9646fb7956b9R166
- to be completely like fold, which either return an accumulator or barfs.

So I need to:
- Add code that will unwrap this `nocatch` thing in `rabbit_cli:main/3`, so it'll be reported in simple form `{bad_argument, dummy}` (as it happens in previous version).
- Change unit-test so it'll expect the control action to throw error instead of returning it.

Does this sound reasonable?

binarin:
BTW, what is expected error output in this case? I've checked both in 3.6.1 and in master, they both print
```
Listing queues ...
Error: {bad_argument,dummy}
```

Should it be handled in [this clause](https://github.com/rabbitmq/rabbitmq-server/blob/8981049b7b9b463e4056c5c9352014a0393f88ff/src/rabbit_cli.erl#L100) ? I've tried to add similar clause that will catch the same error in 3.6.1, but it looks ugly due to the big usage message.

hairyhum:
@binarin I think error formatting is out of scope of this issue.

binarin:
Changed PR so that error messaged during interactive use is the same and the unit tests are passing.

hairyhum:
Looks like this implementation doesn't support `--online` and `--offline` arguments.
```
$ scripts/rabbitmqctl list_queues --offline name
Listing queues ...
Error: undef
```

binarin:
Can't reproduce - in my checkout `--online`/`--offline` work flawlessly.

hairyhum:
Ok. It works with fresh `master`

michaelklishin:
@binarin thank you, this is a very nice improvement!

PR #694, opened 2016-03-18
"rabbitmqctl stop" exits with 0 if it fails to contact the node

### LINKED ISSUE #693: "rabbitmqctl stop" fails if the node is not running
When one runs `rabbitmqctl stop` while the node is already stopped, it complains vehemently and exits with `EX_UNAVAILABLE` (69):
```
# rabbitmqctl stop
Stopping and halting node 'rabbit@rabbit-jpedron-debian1' ...
Error: unable to connect to node 'rabbit@rabbit-jpedron-debian1': nodedown

DIAGNOSTICS
===========
(...)

# echo $?
69
```

Obviously, the command should be quiet and exit with 0 in that case.

In the case of systemd, it looks scary:
```
# service rabbitmq-server status
   Loaded: loaded (/lib/systemd/system/rabbitmq-server.service; disabled)
   Active: failed (Result: exit-code) since Thu 2016-03-17 17:52:24 GMT; 12s ago
  Process: 4112 ExecStop=/usr/lib/rabbitmq/bin/rabbitmqctl stop (code=exited, status=69)
  Process: 3758 ExecStart=/usr/lib/rabbitmq/bin/rabbitmq-server (code=exited, status=0/SUCCESS)
 Main PID: 3758 (code=exited, status=0/SUCCESS)
   Status: "Initialized"
```

dumbbell:
If rabbitmqctl fails to contact the node to ask it to stop, it still displays an error message. However, now it exits with 0; it was `EX_UNAVAILABLE` before.

There are two main reasons it can happen:

* The node is stopped, so that's ok to exit with 0 because the end result is the same.
* The Erlang cookie is incorrect. The error message is here to help but this situation is quite rare compared to the other one.

There is no way to distinguish both, so let's assume the common situation.

This helps in the case of systemd in particular: if one stops RabbitMQ using `rabbitmqctl stop` instead of `service rabbitmq-server stop`, systemd tries to call `rabbitmqctl stop` again. Before this change, it
failed and put the service in a "failed" state.

Fixes #693.

### DIFF
--- a/src/rabbit_cli.erl
+++ b/src/rabbit_cli.erl
@@ -117,7 +117,10 @@ main(ParseFun, DoFun, UsageMod) ->
                 _ ->
                     print_error("unable to connect to node ~w: ~w", [Node, Reason]),
                     print_badrpc_diagnostics([Node]),
-                    rabbit_misc:quit(?EX_UNAVAILABLE)
+                    case Command of
+                        stop -> rabbit_misc:quit(?EX_OK);
+                        _    -> rabbit_misc:quit(?EX_UNAVAILABLE)
+                    end
             end;
         {badrpc_multi, Reason, Nodes} ->
             print_error("unable to connect to nodes ~p: ~w", [Nodes, Reason]),


bdshroyer:
Looks good to me. Those diagnostic messages are great!

PR #697, opened 2016-03-21
Put the RabbitMQ OCF RA policy to /usr/sbin

bogdando:
Fix failing pcs resource list command

Related Fuel bug: https://bugs.launchpad.net/fuel/+bug/1558627

### DIFF
--- a/docs/README-for-packages
+++ b/docs/README-for-packages
@@ -18,3 +18,7 @@ An example configuration file is provided in the same directory as
 this README. Copy it to /etc/rabbitmq/rabbitmq.config to use it. The
 RabbitMQ server must be restarted after changing the configuration
 file.
+
+An example policy file for HA queues is provided in the same directory
+as this README. Copy and chmod +x it to
+/usr/local/sbin/set_rabbitmq_policy to use it with the Pacemaker OCF RA.
--- a/docs/set_rabbitmq_policy.sh.example
+++ b/docs/set_rabbitmq_policy.sh.example
@@ -2,4 +2,3 @@
 # cluster start up. It is a convenient place to set your cluster
 # policy here, for example:
 # ${OCF_RESKEY_ctl} set_policy ha-all "." '{"ha-mode":"all", "ha-sync-mode":"automatic"}' --apply-to all --priority 0
-
--- a/packaging/RPMS/Fedora/rabbitmq-server.spec
+++ b/packaging/RPMS/Fedora/rabbitmq-server.spec
@@ -26,7 +26,6 @@ RabbitMQ is an open source multi-protocol messaging broker.
 %define _rabbit_server_ocf scripts/rabbitmq-server.ocf
 %define _plugins_state_dir %{_localstatedir}/lib/rabbitmq/plugins
 %define _rabbit_server_ha_ocf scripts/rabbitmq-server-ha.ocf
-%define _set_rabbitmq_policy_sh scripts/set_rabbitmq_policy.sh
 
 
 %define _maindir %{buildroot}%{_rabbit_erllibdir}
@@ -51,7 +50,6 @@ mkdir -p %{buildroot}%{_localstatedir}/log/rabbitmq
 install -p -D -m 0755 %{S:1} %{buildroot}%{_initrddir}/rabbitmq-server
 install -p -D -m 0755 %{_rabbit_server_ocf} %{buildroot}%{_exec_prefix}/lib/ocf/resource.d/rabbitmq/rabbitmq-server
 install -p -D -m 0755 %{_rabbit_server_ha_ocf} %{buildroot}%{_exec_prefix}/lib/ocf/resource.d/rabbitmq/rabbitmq-server-ha
-install -p -D -m 0644 %{_set_rabbitmq_policy_sh} %{buildroot}%{_exec_prefix}/lib/ocf/resource.d/rabbitmq/set_rabbitmq_policy.sh.example
 install -p -D -m 0644 %{S:2} %{buildroot}%{_sysconfdir}/logrotate.d/rabbitmq-server
 
 mkdir -p %{buildroot}%{_sysconfdir}/rabbitmq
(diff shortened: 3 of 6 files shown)


bogdando:
@dmitrymex

bogdando:
-1 WIP, I forgot packaging changes

bogdando:
ready for review

bogdando:
@dumbbell please take a look

bogdando:
I'm not sure which default path to use. Could be as well a relative path, like just a "...=set_rabbitmq_policy". But I'm not a fan of relative paths for binaries :/ So, what do you think?

docs/set_rabbitmq_policy.sh.example
```diff
@@ -1,5 +1,4 @@
 # This script is called by rabbitmq-server-ha.ocf during RabbitMQ
 # cluster start up. It is a convenient place to set your cluster
 # policy here, for example:
-# ${OCF_RESKEY_ctl} set_policy ha-all "." '{"ha-mode":"all", "ha-sync-mode":"automatic"}' --apply-to all --priority 0
-
+${OCF_RESKEY_ctl} set_policy ha-all "." '{"ha-mode":"all", "ha-sync-mode":"automatic"}' --apply-to all --priority 0
```
mattymo: Why is this related to the patch to relocate the policy script?

packaging/RPMS/Fedora/rabbitmq-server.spec
```diff
@@ -125,6 +125,7 @@ done
 %doc LICENSE*
 %doc README
 %doc docs/rabbitmq.config.example
+%doc docs/set_rabbitmq_policy.sh.example
```
mattymo: %doc docs/examples/set_rabbitmq_policy.sh.example

scripts/rabbitmq-server-ha.ocf
```diff
@@ -46,6 +45,7 @@ OCF_RESKEY_erlang_cookie_file_default="/var/lib/rabbitmq/.erlang.cookie"
 OCF_RESKEY_use_fqdn_default=false
 OCF_RESKEY_fqdn_prefix_default=""
 OCF_RESKEY_max_rabbitmqctl_timeouts_default=3
+OCF_RESKEY_policy_file_default="/usr/local/sbin/set_rabbitmq_policy.sh"
```
bogdando: this ^^ so , which path? or just OCF_RESKEY_policy_file_default="set_rabbitmq_policy" ?

scripts/rabbitmq-server-ha.ocf
```diff
@@ -13,8 +13,7 @@
 #
 # See usage() function below for more details ...
 #
-# Note that the script uses set_rabbitmq_policy.sh script located in the
-# same directory to setup RabbitMQ policies.
+# Note that the script uses an external file to setup RabbitMQ policies.
```
mattymo: Maybe include that this file is distributed with rabbitmq-server package?

scripts/rabbitmq-server-ha.ocf
mattymo: /usr/sbin/set_rabbitmq_policy.sh is a better path since most people will automate this via a package. But personally I would prefer to see this file as "set_rabbitmq_policy" with no suffix and without explicit path

scripts/rabbitmq-server-ha.ocf
bogdando: It is not, it is distributed as an example in docs

scripts/rabbitmq-server-ha.ocf
bogdando: how to update the [ -f foo ] test then?

docs/set_rabbitmq_policy.sh.example
bogdando: I think the example file must do something if copied to /usr/sbin ?

scripts/rabbitmq-server-ha.ocf
bogdando: Okay, if folks want package it, they may use OCF param instead.
So, let's keep the change simple, which is an absolute path with a simple -f test instead of complex PATH lookups

docs/README-for-packages
```diff
@@ -18,3 +18,7 @@ An example configuration file is provided in the same directory as
 this README. Copy it to /etc/rabbitmq/rabbitmq.config to use it. The
 RabbitMQ server must be restarted after changing the configuration
 file.
+
+An example policy file for HA queues is provided in the same directory
```
dmitrymex: If you change that PR one more time, I suggest to put the exact example name here: set_rabbitmq_policy.sh.example, so that people know what to search

BTW since you removed .sh suffix from the script, it might be removed from the example as well

dmitrymex:
@bogdando: I've left a nit suggestion, but I am not insisting on changing that. Otherwise LGTM

packaging/debs/Debian/debian/rules
```diff
@@ -49,8 +50,9 @@ override_dh_auto_install:
 		$(DEB_DESTDIR)$(PREFIX)/lib/ocf/resource.d/rabbitmq/rabbitmq-server
 	install -p -D -m 0755 scripts/rabbitmq-server-ha.ocf \
 		$(DEB_DESTDIR)$(PREFIX)/lib/ocf/resource.d/rabbitmq/rabbitmq-server-ha
-	install -p -D -m 0644 scripts/set_rabbitmq_policy.sh \
-		$(DEB_DESTDIR)$(PREFIX)/lib/ocf/resource.d/rabbitmq/set_rabbitmq_policy.sh.example
+
+	install -p -D -m 0644 docs/set_rabbitmq_policy.sh.example \
+		$(DEB_DESTDIR)$(RMQ_DOCS_DIR)/set_rabbitmq_policy.sh.example
```
dumbbell: This file should be listed in `debian/rabbitmq-server.docs` instead of being manually installed by `debian/rules`.

bogdando: thank you! fixed

PR #700, opened 2016-03-23
Make slaves wait timeout configurable instead of hardcoded 15000

lemenkov:
This slightly improves rabbitmq/rabbitmq-server#584

### DIFF
--- a/src/rabbit_mirror_queue_master.erl
+++ b/src/rabbit_mirror_queue_master.erl
@@ -43,7 +43,8 @@
                  backing_queue_state,
                  seen_status,
                  confirmed,
-                 known_senders
+                 known_senders,
+                 wait_timeout
                }).
 
 -ifdef(use_specs).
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you. Did you mean this to go into `3.6.x`? (we can cherry-pick, since it's a single commit)

lemenkov:
@michaelklishin yes, I believe this should be applied to 3.6.x.

michaelklishin:
@lemenkov OK, I will cherry-pick, no need to re-submit. Thank you.

michaelklishin:
@lemenkov also, what lead you to making the change? Is 15s not high enough for your case? Should we bump the default somewhat?

lemenkov:
Nope, nothing in particular. We just feel uncomfortable w/o ability to change that setting fast enough. Frankly speaking, "15 seconds" sounds too arbitrary, so we're certain there will be situations when we have to change it.

michaelklishin:
Amended the PR a bit, cherry-picked both commits to `stable`. Thank you, @lemenkov!

PR #702, opened 2016-03-23
Fix `rabbitmqctl list_consumers`

### LINKED ISSUE #701: `rabbitmqctl list_consumers` is unreliable
- When queue has multiple consumers, only one of them is reported. This also happens with `rabbitmqctl report`
- When queue has no consumers at all, `list_consumers` prints strange `[] [] [] [] []` there.

binarin:
As we are mapping over queues in search of their consumers, actual
result for every queue could be zero or more consumers. We need to take
that into account while printing.

Closes #701


michaelklishin:
Thank you!

PR #711, opened 2016-03-29
Support file extension in the RABBITMQ_CONFIG_FILE env var values

### LINKED ISSUE #691: Support config file(s) path values with and without file extensions
Currently we add the `.config`/`.conf` (once #550 is merged) suffix to `RABBITMQ_CONFIG_FILE` before attempting to do anything with it. This is fairly counter-intuitive and makes some of our documentation examples look confusing. It would be a lot more user friendly to support values with and without file extensions.

hairyhum:
Fixes #691


michaelklishin:
This breaks support for paths without a filename extension. The problematic line seems to be
the one with `"${RABBITMQ_CONFIG_FILE%.*}"`:

```
export RABBITMQ_CONFIG_FILE=./etc/rabbitmq/rabbitmq
echo "${RABBITMQ_CONFIG_FILE%.*}"

export RABBITMQ_CONFIG_FILE=./etc/rabbitmq/rabbitmq.config
echo "${RABBITMQ_CONFIG_FILE%.*}"
./etc/rabbitmq/rabbitmq
export RABBITMQ_CONFIG_FILE=./etc/rabbitmq/rabbitmq.sysctl
echo "${RABBITMQ_CONFIG_FILE%.*}"
./etc/rabbitmq/rabbitmq
export RABBITMQ_CONFIG_FILE=./etc/rabbitmq/rabbitmq.sysctl.conf
echo "${RABBITMQ_CONFIG_FILE%.*}"
./etc/rabbitmq/rabbitmq.sysctl
```

michaelklishin:
I've pushed something that fixes this particular problem by using `dirname` and `basename`. I'm hardly great at shell script, so @hairyhum and @dumbbell, please double check.

We also need to work on supporting both `advanced` and `advanced.config`.

hairyhum:
Or we can just change conditions, because extension is being checked there anyway. So we can use any extension trimming mechanism.

michaelklishin:
@hairyhum this passes the acceptance criteria for `rabbitmq.conf` and `rabbitmq.config` but not `advanced.config`. I think we should support both options for the `advanced.config` as well.

hairyhum:
Added support for `advanced.config`. Haven't checked windows scripts yet.

michaelklishin:
This looks good on OS X (zsh) and Ubuntu (dash). Haven't had a chance to test on Windows yet.

scripts/rabbitmq-service.bat
```diff
@@ -171,8 +178,27 @@ if exist "!RABBITMQ_CONFIG_FILE!.config" (
                                 -conf_advanced "!RABBITMQ_ADVANCED_CONFIG_FILE!" ^
                                 -config "!RABBITMQ_ADVANCED_CONFIG_FILE!"
     )
+) else (
+    if exist "!RABBITMQ_CONFIG_FILE!.config" (
+        set RABBITMQ_CONFIG_ARG=-config "!RABBITMQ_CONFIG_FILE!"
+    ) else if exist "!RABBITMQ_CONFIG_FILE!.config" (
```
Gsantomaggio: this `else if`  should be `if exist "!RABBITMQ_CONFIG_FILE!.conf"` as [`rabbitmq-server.bat`](https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-691/scripts/rabbitmq-server.bat#L93)

scripts/rabbitmq-service.bat
```diff
@@ -126,6 +126,11 @@ set RABBITMQ_EBIN_ROOT=!RABBITMQ_HOME!\ebin
 
 set RABBITMQ_CONFIG_FILE="!RABBITMQ_CONFIG_FILE!"
```
Gsantomaggio: I am not sure this `set` is necessary. 
In [`rabbitmq-server.bat`](https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-691/scripts/rabbitmq-server.bat#L43) is not present.

Issue #698, opened 2016-03-21
Investigate support for mixed Erlang versions in a cluster


michaelklishin:
We are constrained by the compatibility of internal OTP libraries that we currently use, namely `mnesia` and `global`, between major OTP releases. It would be great if we could support at least a range of versions, e.g. 17.x and 18.x.

hairyhum:
We have started cluster with erlang versions `17.0` and `18.3`.
We did tests on `mnesia` and `global` apps.
Then we have started RabbitMQ cluster of two nodes (17.0 and 18.3) and run java functional and server tests.
Also tested a bit using management and mirrored queues.
Everything worked for this setup.

michaelklishin:
Thanks. We need to find out if Mnesia has an inter-node communication protocol versioning.

Given that 3.6.x supports R16B03, it's worth testing with that version instead of 17.0, too. Or all 3 of them.

Also note that `ant test-functional` in the Java client repo does not use mirrored queues, there is a separate set suite/setup scenario for that and it's going to be tricky to coerce it into using different VMs.

hairyhum:
Same tests work with `R16B03`.

hairyhum:
According to `mnesia` [changelog](http://erlang.org/doc/apps/mnesia/notes.html), it can have incompatible versions, but changelog is only way to know it.

michaelklishin:
Surely we can spend some time (a couple of days, for example) to
investigate Mnesia internals to see if they have any internal communication
versioning that we can track. Change logs are great but they are rarely
written well, or have 100% of the changes listed.

On Tue, Mar 22, 2016 at 6:19 PM, Daniil Fedotov <notifications@github.com>
wrote:

> According to mnesia changelog
> <http://erlang.org/doc/apps/mnesia/notes.html>, it can have incompatible
> versions, but changelog is only way to know it.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/rabbitmq/rabbitmq-server/issues/698#issuecomment-199861899>
>



-- 
MK

Staff Software Engineer, Pivotal/RabbitMQ

hairyhum:
Looks like `mnesia` have `protocol_version`, that can be accessed using `mnesia:system_info(protocol_version)`. 
Mnesia supports communication with two versions, so called `current` and `previous`. 
[...]

PR #719, opened 2016-03-31
Replace OTP version consistency with mnesia protocol consistency

hairyhum:
To be able to run RabbitMQ cluster on different versions of Erlang/OTP, OTP consistency was replaced by mnesia consistency, which checks if mnesia can negotiate protocol with remote node.
Related to #698


hairyhum:
Maybe it should go to `stable`. I'm not sure.

michaelklishin:
We primarily would like it in stable.

Issue #717, opened 2016-03-30
Unable to set_disk_free_limit mem_relative to 1.0 or higher


bdshroyer:
When `rabbitmqctl set_disk_free_limit mem_relative` is invoked with a value of 1.0 or higher, you get the following error message:

```
Error: {{function_clause,
            [{rabbit_resource_monitor_misc,parse_information_unit,
                 [{mem_relative,1.0}],
                 [{file,"src/rabbit_resource_monitor_misc.erl"},{line,29}]},
             {rabbit_disk_monitor,interpret_limit,1,
                 [{file,"src/rabbit_disk_monitor.erl"},{line,241}]},
             {rabbit_disk_monitor,set_disk_limits,2,
                 [{file,"src/rabbit_disk_monitor.erl"},{line,188}]},
             {rabbit_disk_monitor,handle_call,3,
                 [{file,"src/rabbit_disk_monitor.erl"},{line,145}]},
             {gen_server,try_handle_call,4,
                 [{file,"gen_server.erl"},{line,629}]},
             {gen_server,handle_msg,5,[{file,"gen_server.erl"},{line,661}]},
             {proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,240}]}]},
        {gen_server,call,
            [rabbit_disk_monitor,
             {set_disk_free_limit,{mem_relative,1.0}},
             infinity]}}
```

michaelklishin:
Looks like this was introduced in https://github.com/rabbitmq/rabbitmq-server/commit/829890be2a664a00205a3cbf04cdf00e18b022a0, and is a misunderstanding of what the value is.

PR #721, opened 2016-03-31
Allow memory-relative disk-free limits greater than 1.0

bdshroyer:
Fixes https://github.com/rabbitmq/rabbitmq-server/issues/717.


bdshroyer:
Tested in [this PR](https://github.com/rabbitmq/rabbitmq-test/pull/34).

PR #723, opened 2016-04-01
Replace OTP version consistency with mnesia protocol consistency

### LINKED ISSUE #698: Investigate support for mixed Erlang versions in a cluster

hairyhum:
To be able to run RabbitMQ cluster on different versions of Erlang/OTP, OTP consistency was replaced by mnesia consistency, which checks if mnesia can negotiate protocol with remote node.

Fixes #698.

### DIFF
--- a/src/rabbit_mnesia.erl
+++ b/src/rabbit_mnesia.erl
@@ -423,6 +423,7 @@ cluster_status(WhichNodes) ->
 
 node_info() ->
     {rabbit_misc:otp_release(), rabbit_misc:version(),
+     mnesia:system_info(protocol_version),
      cluster_status_from_mnesia()}.
 
 node_type() ->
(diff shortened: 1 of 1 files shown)


src/rabbit_mnesia.erl
```diff
@@ -785,6 +786,49 @@ check_nodes_consistency(Node, RemoteStatus = {RemoteAllNodes, _, _}) ->
...
+        stopping ->
+            ensure_mnesia_not_running(),
+            no;
+        starting ->
+            ensure_mnesia_running(),
+            yes;
+        Other    -> Other
+    end,
+    case MnesiaRunning of
+        yes -> Fun();
```
michaelklishin: Can we please use `true` and `false` here?

src/rabbit_mnesia.erl
```diff
@@ -785,6 +786,49 @@ check_nodes_consistency(Node, RemoteStatus = {RemoteAllNodes, _, _}) ->
...
+        no  ->
+            {ok, MnesiaDir} = application:get_env(mnesia, dir),
+            application:unset_env(mnesia, dir),
+            mnesia:start(),
```
michaelklishin: I'm curious what will the Mnesia directory value be in this case (or well, by default)?

src/rabbit_mnesia.erl
hairyhum: There will be 
```
    stopping -> ..., false;
    starting -> ..., true;
    yes      -> true;
    no       -> false
```
Does it look better?

michaelklishin: I find them roughly equally descriptive but `true` and `false` a much more standard ;)

src/rabbit_mnesia.erl
hairyhum: `{CWD}/Mnesia.{NODENAME}`
But it won't be created until schema with `disc_copies` is created. By default schema will have only `ram_copies`.

michaelklishin: Excellent, that sounds pretty safe.

src/rabbit_mnesia.erl
```diff
@@ -596,21 +597,21 @@ check_cluster_consistency(Node, CheckNodesConsistency) ->
     case rpc:call(Node, rabbit_mnesia, node_info, []) of
         {badrpc, _Reason} ->
             {error, not_found};
-        {_OTP, _Rabbit, {error, _}} ->
+        {_OTP, Rabbit, Hash, _Status} when is_binary(Hash) ->
+            %% delegate hash checking implies version mismatch
```
michaelklishin: I don't understand this comment. Can you please elaborate?

hairyhum: In previous versions (long time ago), `rabbit_mnesia:node_info` was returning 4-element tuple. 3rd element was `delegate_beam_hash` - beam md5 hash of `delegate` module. 
This commit adds 4-element tuple back, where 3rd element would be protocol consistency. Type match is here to avoid confusing one with another.

src/rabbit_mnesia.erl
```diff
@@ -593,27 +594,35 @@ check_cluster_consistency() ->
...
+            case check_consistency(Node, OTP, Rabbit, Protocol) of
                 {error, _} = E -> E;
                 ok             -> {ok, Status}
             end
     end.
 
+remote_node_info(Node) ->
+    case rpc:call(Node, rabbit_mnesia, node_info, []) of
+        {badrpc, _} = Error   -> Error;
+        {OTP, Rabbit, Status} -> {OTP, Rabbit, unsupported, Status};
```
hairyhum: Will check OTP consistency if mnesia consistency is unsupported.

PR #731, opened 2016-04-07
Stop a rabbitmq pacemaker resource when monitor fails

bogdando:
Related Fuel bug https://bugs.launchpad.net/fuel/+bug/1567355


bogdando:
@dmitrymex

dmitrymex:
@bogdando : frankly, I don't like us to perform any action in monitor. But since we don't have better idea how to fix this bug, we will have to go with that workaround. So +1 from me.

michaelklishin:
Thanks, folks.

Issue #744, opened 2016-04-13
Concurrent updates to policies can trigger a `function_clause` exception


dumbbell:
Discovered while working on #725, concurrent executions of `rabbitmqctl set_policy` my cause the following exception:

```
{error,
 {function_clause,
  [{rabbit_policy,'-match/2-lc$^0/1-0-',
    [{'EXIT',
      {aborted,
       {cyclic,'rmq-ct-unit_inbroker_SUITE-21000@localhost',
        {rabbit_runtime_parameters,'______WHOLETABLE_____'},
        read,read,       
        {tid,145,<9838.1856.0>}}}}],
    [{file,"src/rabbit_policy.erl"},{line,292}]},
   {rabbit_policy,match,2,
    [{file,"src/rabbit_policy.erl"},{line,292}]},
   {rabbit_policy,update_exchange,2,
    [{file,"src/rabbit_policy.erl"},{line,259}]},
   {rabbit_policy,'-update_policies/1-lc$^2/1-1-',2,
    [{file,"src/rabbit_policy.erl"},{line,248}]},
   {rabbit_policy,'-update_policies/1-fun-0-',1,
    [{file,"src/rabbit_policy.erl"},{line,248}]},
   {mnesia_tm,apply_fun,3,
    [{file,"mnesia_tm.erl"},{line,833}]},
   {mnesia_tm,execute_transaction,5,
    [{file,"mnesia_tm.erl"},{line,808}]},
   {rabbit_misc,'-execute_mnesia_transaction/1-fun-0-',1,
    [{file,"src/rabbit_misc.erl"},{line,542}]}]}}
```

The problem comes from this code in `rabbit_policy.erl`:

```erlang
    {Xs, Qs} = rabbit_misc:execute_mnesia_transaction(
                 fun() ->
                         [mnesia:lock({table, T}, write) || T <- Tabs], %% [1]
                         case catch list(VHost) of
                             {error, {no_such_vhost, _}} ->
                                 ok; %% [2]
                             Policies ->
                                 {[update_exchange(X, Policies) ||
                                      X <- rabbit_exchange:list(VHost)],
                                  [update_queue(Q, Policies) ||
                                      Q <- rabbit_amqqueue:list(VHost)]}
                         end
                 end),
```

`list(VHost)` calls `rabbit_runtime_parameters:list(VHost), <<"policy">>)`, which calls `rabbit_misc:dirty_read_all()`, which finally calls `mnesia:dirty_select()`.

[...]

PR #745, opened 2016-04-13
rabbit_policy: Fix several issues with policy updates

dumbbell:
Thus, any transaction abort are properly forwarded. This fixes a `function_clause` because the code called `update_exchange()` or `update_queue()` with the caught exit signal.

References #744.
[#117522069]

### DIFF
--- a/src/rabbit_policy.erl
+++ b/src/rabbit_policy.erl
@@ -242,6 +242,8 @@ update_policies(VHost) ->
                  fun() ->
                          [mnesia:lock({table, T}, write) || T <- Tabs], %% [1]
                          case catch list(VHost) of
+                             {'EXIT', Exit} ->
+                                 exit(Exit);
                              {error, {no_such_vhost, _}} ->
                                  ok; %% [2]
                              Policies ->
--- a/src/rabbit_policy.erl
+++ b/src/rabbit_policy.erl
@@ -242,10 +242,10 @@ update_policies(VHost) ->
                  fun() ->
                          [mnesia:lock({table, T}, write) || T <- Tabs], %% [1]
                          case catch list(VHost) of
+                             {'EXIT', {throw, {error, {no_such_vhost, _}}}} ->
+                                 {[], []}; %% [2]
                              {'EXIT', Exit} ->
                                  exit(Exit);
-                             {error, {no_such_vhost, _}} ->
-                                 {[], []}; %% [2]
                              Policies ->
                                  {[update_exchange(X, Policies) ||
                                       X <- rabbit_exchange:list(VHost)],
--- a/src/rabbit_policy.erl
+++ b/src/rabbit_policy.erl
@@ -245,7 +245,7 @@ update_policies(VHost) ->
                              {'EXIT', Exit} ->
                                  exit(Exit);
                              {error, {no_such_vhost, _}} ->
-                                 ok; %% [2]
+                                 {[], []}; %% [2]
                              Policies ->
                                  {[update_exchange(X, Policies) ||
                                       X <- rabbit_exchange:list(VHost)],


src/rabbit_policy.erl
```diff
@@ -242,6 +242,8 @@ update_policies(VHost) ->
                  fun() ->
                          [mnesia:lock({table, T}, write) || T <- Tabs], %% [1]
                          case catch list(VHost) of
+                             {'EXIT', Exit} ->
+                                 exit(Exit);
```
hairyhum: should it be `mnesia:abort/1` instead of `exit/1`?

hairyhum:
Can you add some tests, that will check that exchanges and queues policy settings correspond to policy database after errors in this transaction?

src/rabbit_policy.erl
dumbbell: It should be `exit()` because:
* If we caught an exit signal triggered by `mnesia:abort()`, the tuple is already of the form expected by Mnesia (`{aborted, Some_Internal_Record}`). So here, we just forward it.
* If we caught a "normal" exit signa, we also forward it as is and let Mnesia handle the difference between an abortion and a plain exit signal.

dumbbell:
I'm going to add tests but only in the common_test branch because our current framework doesn't do parallel tests.

dumbbell:
Ok, I added tests which check if a policy was properly updated by looking at its properties.

hairyhum:
@dumbbell but does it check that queues and exchanges are being updated after policy update? If policy had been updated, but `update_policies`, which is updating queues and exchanges parameters to correspond with policies, will fail?

dumbbell:
The tests don't check that, but the Mnesia transaction is completed successfully; otherwise an exception is thrown.

hairyhum:
As far as I can see from code, this transaction is not updating policy, but only exchanges and queues. Policy is being updated in another function https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_runtime_parameters.erl#L142. So if policy is updated and this function fails there will be inconsistency.

dumbbell:
In the case of the testsuite and with this patch, the transaction function fails but Mnesia restarts the transaction which succeeds eventually.

In `rabbit_cpolicy.erl`, the transaction is expected to succeed:
```erlang
{Xs, Qs} = rabbit_misc:execute_mnesia_transaction(
```

If the transaction is aborted, `rabbit_misc` throws an exception.

So the difference between without and with the patch is that the transaction is restarted with the patch. Without the patch the transaction function crashes (so not a transaction abort).

However, with or without the patch, `rabbit_policy.erl` may not ensure consistency between policy updates and queue/exchange updates. This is something I didn't check. If this is true, this is another bug which this patch doesn't try to address.

While here, I see another problem:
```erlang
    {Xs, Qs} = rabbit_misc:execute_mnesia_transaction(
                 fun() ->
                         [mnesia:lock({table, T}, write) || T <- Tabs], %% [1]
                         case catch list(VHost) of
                             % ...
                             {error, {no_such_vhost, _}} ->
                                 ok; %% [2]
                               % ^^
```

The code expects the transaction to return a tuple, not `ok`.

hairyhum:
I've tried to reproduce mnesia abort by adding conditional abort to `rabbit_runtime_parameters:list` and it doesn't look like restarting, are you sure it's restarting in error case?

hairyhum:
I'm going to file additional issue then.

dumbbell:
In my tests, without the patch, the transaction function crashes 30% of the time while running the testsuite on policies. With the patch, it always succeeds and thus returns the expected tuple. I'm not 100% sure the transaction is restarted, however, that's the only possibility I see.

dumbbell:
I filed #755 to address the second problem in the transaction function and I will add a patch to it in this pull request.

hairyhum:
I've tested with additional logs and it looks like not restarting transaction.

hairyhum:
Looks like it detects `cyclic` errors and retries. But if we fail before this transaction there still be inconsistency mentioned in #756

hairyhum:
@dumbbell will you add tests for #755 ?

dumbbell:
@hairyhum: Yes, I will add one in the common_test branch.

hairyhum:
If VHost is being deleted when set_policy is running there are `{throw,{error,{no_such_vhost,<<"...">>}}}` error. This error is being caught and propagated like `{'EXIT', ...}` instead of being matched by `{error, {no_such_vhost, _}}`

dumbbell:
Yes, I see the same thing.

dumbbell:
What I get is really weird. Here is what I do from a test case (not manually):
1. Create a vhost
2. Create a policy on this vhost
3. Delete the vhost.

Currently, I'm doing this synchronously, so I'm not testing the removal of the vhost during the policy update. However, deleting the vhost fails with the same error you get, even though a call to `list_vhosts` clearly shows it. So far, I have no idea why this happens. I must misinterpret what I'm seeing.

dumbbell:
The same test from the CLI confirms there is something broken with vhost removal:
```
$ ./scripts/rabbitmqctl add_vhost /bla
Creating vhost "/bla" ...

$ ./scripts/rabbitmqctl delete_vhost /bla
Deleting vhost "/bla" ...
```
```
$ ./scripts/rabbitmqctl add_vhost /bla
Creating vhost "/bla" ...

$ ./scripts/rabbitmqctl set_policy -p /bla pouet ".*" "{\"ha-mode\":\"all\"}"
Setting policy "pouet" for pattern ".*" to "{\"ha-mode\":\"all\"}" with priority "0" ...

$ ./scripts/rabbitmqctl list_vhosts
Listing vhosts ...
/bla
/

$ ./scripts/rabbitmqctl list_policies -p /bla
Listing policies ...
/bla	pouet	all	.*	{"ha-mode":"all"}	0

$ ./scripts/rabbitmqctl delete_vhost /bla
Deleting vhost "/bla" ...
Error: {throw,{error,{no_such_vhost,<<"/bla">>}}}

$ ./scripts/rabbitmqctl list_vhosts
Listing vhosts ...
/
```

`dbg` shows that the vhost is removed, then its existence is checked during `update_policies`:
```
(<0.947.0>) call rabbit_vhost:with(<<"/vhost_removed_while_updating_policy-vhost">>,#Fun<rabbit_runtime_parameters.2.50904568>)
(<0.947.0>) returned from rabbit_vhost:with/2 -> #Fun<rabbit_vhost.4.112283716>
(<0.947.0>) call rabbit_vhost:'-with/2-fun-0-'(#Fun<rabbit_runtime_parameters.2.50904568>,<<"/vhost_removed_while_updating_policy-vhost">>)
(<0.947.0>) call rabbit_runtime_parameters:'-mnesia_clear/3-fun-0-'(<<"vhost_removed_while_updating_policy-policy">>,<<"policy">>,<<"/vhost_removed_while_updating_policy-vhost">>)
(<0.947.0>) returned from rabbit_runtime_parameters:'-mnesia_clear/3-fun-0-'/3 -> ok
(<0.947.0>) returned from rabbit_vhost:'-with/2-fun-0-'/2 -> ok
(<0.947.0>) returned from rabbit_runtime_parameters:mnesia_clear/3 -> ok
(<0.947.0>) returned from rabbit_runtime_parameters:clear_any/3 -> #Fun<rabbit_runtime_parameters.1.50904568>
(<0.947.0>) returned from rabbit_policy:delete/2 -> #Fun<rabbit_runtime_parameters.1.50904568>
(<0.947.0>) call rabbit_vhost:'-internal_delete/1-lc$^2/1-2-'([],<<"/vhost_removed_while_updating_policy-vhost">>)
(<0.947.0>) returned from rabbit_vhost:'-internal_delete/1-lc$^2/1-2-'/2 -> []
(<0.947.0>) returned from rabbit_vhost:'-internal_delete/1-lc$^2/1-2-'/2 -> [#Fun<rabbit_runtime_parameters.1.50904568>]
(<0.947.0>) returned from rabbit_vhost:internal_delete/1 -> [#Fun<rabbit_runtime_parameters.1.50904568>]
(<0.947.0>) returned from rabbit_vhost:'-delete/1-fun-1-'/1 -> [#Fun<rabbit_runtime_parameters.1.50904568>]
(<0.947.0>) returned from rabbit_vhost:'-with/2-fun-0-'/2 -> [#Fun<rabbit_runtime_parameters.1.50904568>]
(<0.908.0>) call rabbit_vhost:'-delete/1-lc$^2/1-2-'([#Fun<rabbit_runtime_parameters.1.50904568>])
...
(<0.908.0>) call rabbit_policy:update_policies(<<"/vhost_removed_while_updating_policy-vhost">>)
(<0.948.0>) call rabbit_policy:'-update_policies/1-fun-0-'(<<"/vhost_removed_while_updating_policy-vhost">>)
(<0.948.0>) call rabbit_policy:'-update_policies/1-lc$^0/1-0-'([rabbit_queue,rabbit_durable_queue,rabbit_exchange,rabbit_durable_exchange])
(<0.948.0>) call rabbit_policy:'-update_policies/1-lc$^0/1-0-'([rabbit_durable_queue,rabbit_exchange,rabbit_durable_exchange])
(<0.948.0>) call rabbit_policy:'-update_policies/1-lc$^0/1-0-'([rabbit_exchange,rabbit_durable_exchange])
(<0.948.0>) call rabbit_policy:'-update_policies/1-lc$^0/1-0-'([rabbit_durable_exchange])
(<0.948.0>) call rabbit_policy:'-update_policies/1-lc$^0/1-0-'([])
(<0.948.0>) returned from rabbit_policy:'-update_policies/1-lc$^0/1-0-'/1 -> []
(<0.948.0>) returned from rabbit_policy:'-update_policies/1-lc$^0/1-0-'/1 -> []
(<0.948.0>) returned from rabbit_policy:'-update_policies/1-lc$^0/1-0-'/1 -> []
(<0.948.0>) returned from rabbit_policy:'-update_policies/1-lc$^0/1-0-'/1 -> []
(<0.948.0>) returned from rabbit_policy:'-update_policies/1-lc$^0/1-0-'/1 -> []
(<0.948.0>) call rabbit_policy:list(<<"/vhost_removed_while_updating_policy-vhost">>)
(<0.948.0>) call rabbit_policy:list0(<<"/vhost_removed_while_updating_policy-vhost">>,#Fun<rabbit_policy.1.39015468>)
(<0.948.0>) call rabbit_runtime_parameters:list(<<"/vhost_removed_while_updating_policy-vhost">>,<<"policy">>)
(<0.948.0>) call rabbit_runtime_parameters:'-list/2-fun-0-'(<<"policy">>,<<"/vhost_removed_while_updating_policy-vhost">>)
(<0.948.0>) call rabbit_vhost:assert(<<"/vhost_removed_while_updating_policy-vhost">>)
(<0.948.0>) call rabbit_vhost:exists(<<"/vhost_removed_while_updating_policy-vhost">>)
(<0.948.0>) returned from rabbit_vhost:exists/1 -> false
(<0.948.0>) exception_from {rabbit_vhost,assert,1} {throw,{error,{no_such_vhost,<<"/vhost_removed_while_updating_policy-vhost">>}}}
(<0.948.0>) exception_from {rabbit_runtime_parameters,'-list/2-fun-0-',2} {throw,{error,{no_such_vhost,<<"/vhost_removed_while_updating_policy-vhost">>}}}
(<0.948.0>) exception_from {rabbit_runtime_parameters,list,2} {exit,{throw,{error,{no_such_vhost,<<"/vhost_removed_while_updating_policy-vhost">>}}}}
(<0.948.0>) exception_from {rabbit_policy,list0,2} {exit,{throw,{error,{no_such_vhost,<<"/vhost_removed_while_updating_policy-vhost">>}}}}
(<0.948.0>) exception_from {rabbit_policy,list,1} {exit,{throw,{error,{no_such_vhost,<<"/vhost_removed_while_updating_policy-vhost">>}}}}
(<0.948.0>) exception_from {rabbit_policy,'-update_policies/1-fun-0-',1} {exit,{throw,{error,{no_such_vhost,<<"/vhost_removed_while_updating_policy-vhost">>}}}}
(<0.908.0>) exception_from {rabbit_policy,update_policies,1} {throw,
    {error,
        {throw,
            {error,
                {no_such_vhost,
                    <<"/vhost_removed_while_updating_policy-vhost">>}}}}}
```

So there is even no need for concurrent accesses to trigger the problem.

hairyhum:
It should be [this throw](https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-744/src/rabbit_vhost.erl#L142)

dumbbell:
Yes

dumbbell:
After changing the case clause to `{'EXIT', {throw, {error, {no_such_vhost, _}}}}`, I get the bad match with `ok`. I'm going to file a third issue.

dumbbell:
Here it is: #759.

I will add a test in the common_test branch which will test both #755 and #759 in fact.

PR #757, opened 2016-04-19
Stop process when rabbit is running but is not connected to master.

relusek:
It's should goes down due to avoid split brain.


michaelklishin:
If you intend this to go into a `3.6.x` release, please re-submit against the `stable` branch.

@bogdando @dmitrymex can I get your blessing?

bogdando:
LGTM, but yes please submit to the stable

PR #758, opened 2016-04-19
Stop process when rabbit is running but is not connected to master.

relusek:
It's should goes down due to avoid split brain.

Related Fuel bug: https://bugs.launchpad.net/fuel/+bug/1541471


michaelklishin:
This branch is based on `master` but submitted against `stable`.

PR #760, opened 2016-04-19
Match on types deleted and more efficient match (x-headers).

sylvainhubsch:
Hi,

Why I propose to no more match on types on headers match ?
**1.** firstly because all began with a guess (<< ... it's not specified, but I assume that a mismatched type causes a mismatched value >>) and types have already been relaxed one time (https://github.com/rabbitmq/rabbitmq-server/commit/000551380173f54e7b9c7f8a8db4081ce3127974)
**2.** because I personnaly did not see in AMQP specs such requirements : << ...if the field in the bind arguments has a value and a field of the same name exists in the message headers and has that same value >> (rather vague, but "value" is written, not "type")
**3.** because the only case I see it could impact a customer is one who has defined bindings with "float" and match data with "integer" values (or vice-versa) which is false today anyway... This is why I think it is pretty much unlikely to impact customer (but yes, sometimes people do weird things :)
(I do not see other case other than "float vs int" comparing values via "==" gives true between different value types..)
**4.** because it is simplier to not match at all on types, letting code be more readable !

I added too earlier stop function that could be much more efficient in some cases :
<<
headers_match(_, _, false, _, all) -> false;
headers_match(_, _, _, true, any) -> true;
>>

And finally, I added some comments to make source code more readable.

Sincerly,
Sylvain.


michaelklishin:
You claim that this version is more efficient. Do you have any benchmarks that compare the two versions?

sylvainhubsch:
I did, but I don't have a summary because there are many cases (clearly more efficient on bindings with long list of headers). Maybe more substantiation would be better, I understand; so I will work on producing some benchmarks results in the meantime.
Thanks you.

michaelklishin:
@sylvainhubsch any numbers you can share? It'd help us prioritise this issue. Thank you.

sylvainhubsch:
Sorry I did not have much time these days; will post some stats on monday/tuesday (main problem is with big bindings which take "long time" to be retrived from mnesia so that stats seems not relevant).
Thank you.

sylvainhubsch:
Hi,


Here are values from my tests; maybe it's a bit messy.. sorry about that.


At my business Cie, we work a lot with headers exchanges; what I've done
too is a "deep match" on headers (not only first level) through table and
array to catch values of x-death messages to stop the loop of recycling
messages definitively not handled by consumers (rerouted via DLX).
To achieve this, x-death messages (the 3 types) should be separated (new
header not named "x-*" ??) and in a mid time I incorporated comparing
header values with message when binding key starts with ">= " or "< " etc...

So in a word, I prepare another PR that will modify that code (but I don't
know exactly when)...and I would not blame you if you refuse my current PR.
But I'm pretty sure that comparing relaxed-types is "overkill" and that it
missed the two lines added in this PR :)


Long life to the Rabbit and thanks for all your work!

Sylvain.


2016-06-01 0:49 GMT+02:00 Michael Klishin <notifications@github.com>:

michaelklishin:
Thank you, that looks promising.

michaelklishin:
Cherry-picked to `stable` so it will be in `3.6.3`. Thanks again.

PR #765, opened 2016-04-21
Do not move rabbitmq.conf to rabbitmq-env.conf during RPM and DEB install

### LINKED ISSUE #763: Invalid RPM spec file %post section
For some reason master contains an RPM spec %post section as follows:

    if [ -f %{_sysconfdir}/rabbitmq/rabbitmq.conf ] && [ ! -f %{_sysconfdir}/rabbitmq/rabbitmq-env.conf ]; then
        mv %{_sysconfdir}/rabbitmq/rabbitmq.conf %{_sysconfdir}/rabbitmq/rabbitmq-env.conf
    else 
        touch %{_sysconfdir}/rabbitmq/rabbitmq-env.conf
    fi

I suspect that is an error, I am not clear why the new style configuration gets renamed to an environment file which of course results in an invalid configuration.

hairyhum:
Fixes #763
Since `rabbitmq.conf` name is used in new configuration format, we should not assume it's an old env configuration. 
If someone will upgrade from the old env config (`2.3.0` which is unlikely), rabbitmq will just fail to generate a new `.config` and will crash.

### DIFF
--- a/packaging/RPMS/Fedora/rabbitmq-server.spec
+++ b/packaging/RPMS/Fedora/rabbitmq-server.spec
@@ -91,11 +91,7 @@ fi
 
 %post
 /sbin/chkconfig --add %{name}
-if [ -f %{_sysconfdir}/rabbitmq/rabbitmq.conf ] && [ ! -f %{_sysconfdir}/rabbitmq/rabbitmq-env.conf ]; then
-    mv %{_sysconfdir}/rabbitmq/rabbitmq.conf %{_sysconfdir}/rabbitmq/rabbitmq-env.conf
-else 
-    touch %{_sysconfdir}/rabbitmq/rabbitmq-env.conf
-fi
+touch %{_sysconfdir}/rabbitmq/rabbitmq-env.conf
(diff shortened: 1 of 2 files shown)


packaging/RPMS/Fedora/rabbitmq-server.spec
```diff
@@ -91,11 +91,7 @@ fi
 
 %post
 /sbin/chkconfig --add %{name}
-if [ -f %{_sysconfdir}/rabbitmq/rabbitmq.conf ] && [ ! -f %{_sysconfdir}/rabbitmq/rabbitmq-env.conf ]; then
-    mv %{_sysconfdir}/rabbitmq/rabbitmq.conf %{_sysconfdir}/rabbitmq/rabbitmq-env.conf
-else 
-    touch %{_sysconfdir}/rabbitmq/rabbitmq-env.conf
-fi
+touch %{_sysconfdir}/rabbitmq/rabbitmq-env.conf
```
dumbbell: I believe you can get rid of the creation of `rabbit-env.conf`.

PR #768, opened 2016-04-22
Improve notifications about damaged I/O subsystem

binarin:
Follow-up to https://github.com/rabbitmq/rabbitmq-server/pull/747

It's the best thing that can be done. Without `exit` anything inside
rabbit could hang - most prone to it is the `start_app` action. Crashing
loudly will be helpful, because during 2016 there were at least 2
production investigations about that issue.

And here is the script that I was using for reproduction:
https://gist.github.com/binarin/edd4102b7262d53e2cf09c76d5ca8e9e

### DIFF
--- a/src/rabbit_log.erl
+++ b/src/rabbit_log.erl
@@ -96,7 +96,7 @@ with_local_io(Fun) ->
     Node = node(),
     case node(GL) of
         Node -> Fun();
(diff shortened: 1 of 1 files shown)


dumbbell:
@binarin: I cherry-picked your commit onto `stable` and created a new pull request: #771.

PR #769, opened 2016-04-22
Don't use hardcoded path to `df`

binarin:
Search the `PATH` instead, because e.g. NixOS has no `/bin/df`.


hairyhum:
@binarin can you please resubmit this against `stable`?

dumbbell:
@binarin: I cherry-picked your commit and created a new pull request: #770.

Issue #772, opened 2016-04-27
More verbose logging for node partition detection


Gsantomaggio:
Add log information during the network_partition.

It is useful to add a `warning` log during the partition in the [`rabbit_node_monitor` module](https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-705/src/rabbit_node_monitor.erl#L417)

Something like:

     _  -> rabbit_log:warning("Unexpected running node, the node ~p  is still running.",[Node]),
     cast(Rep, {partial_partition,Node, node(), RepGUID})

This will help in the future to understand network_partition situations

PR #773, opened 2016-04-27
Add log information during the network_partition

Gsantomaggio:
Fixes https://github.com/rabbitmq/rabbitmq-server/issues/772

Output for the new log:

```
=WARNING REPORT==== 27-Apr-2016::11:19:51 ===
Unexpected running node: rabbit1@mac is still running
```

### DIFF
--- a/src/rabbit_node_monitor.erl
+++ b/src/rabbit_node_monitor.erl
@@ -415,8 +415,9 @@ handle_cast({check_partial_partition, Node, Rep, NodeGUID, MyGUID, RepGUID},
                            case rpc:call(Node, rabbit, is_running, []) of
                                {badrpc, _} -> ok;
                                _           ->  
-				   rabbit_log:warning("Unexpected running node:"
-						      " ~p is still running ~n",
+				   rabbit_log:warning("Received a 'DOWN' message" 
+						      " from ~p but still can" 
+						      " communicate with it ~n",
 						      [Node]),
 				   cast(Rep, {partial_partition,
                                                          Node, node(), RepGUID})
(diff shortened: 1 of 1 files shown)


src/rabbit_node_monitor.erl
```diff
@@ -414,7 +414,11 @@ handle_cast({check_partial_partition, Node, Rep, NodeGUID, MyGUID, RepGUID},
                    fun () ->
                            case rpc:call(Node, rabbit, is_running, []) of
                                {badrpc, _} -> ok;
-                               _           -> cast(Rep, {partial_partition,
+                               _           ->  
+				   rabbit_log:warning("Unexpected running node:"
```
michaelklishin: I don't think this message is very descriptive. It lacks context entirely. What it should say (for example):

```
Received a 'DOWN' message from X but still can communicate with it
```

michaelklishin:
@Gsantomaggio I'm afraid this message is more confusing than helpful. This needs some more work.

src/rabbit_node_monitor.erl
Gsantomaggio: Yes, as first choice we wrote a message similar to  yours:

> Disconnection from node ~p but the node is still running

I decided to change it, because this call checks if the node `x` is running.  It does not receive the `DOWN` message. 

BTW maybe for the user is most clear another kind of the message. I'm going to change it

Gsantomaggio:
Changed with:

```
=WARNING REPORT==== 27-Apr-2016::16:17:07 ===
Received a 'DOWN' message from rabbit1@mac but still can communicate with it
```

PR #774, opened 2016-04-27
Ignore non-array x-dead header

### LINKED ISSUE #767: Dead-lettering a message second time crashes the queue
Hi,

The use case is we dead-letter message after several retries, if it is because of some external dependency of consumer, we replay those dead-letter messages to the actual queue again. But consider that the external dependency again goes down, and the consumer tries to dead-letter those messages again, the queue crashes.

I am attaching java client used to simulating the scenario.

Steps to follow to simulate scenario
1. create a dead-letter queue
2. Run the java client to publish and consume message.

I am also attaching the error from rabbitmq logs

[rabbitmq-files.zip](https://github.com/rabbitmq/rabbitmq-server/files/231086/rabbitmq-files.zip)

hairyhum:
Fixes #767

### DIFF
--- a/src/rabbit_dead_letter.erl
+++ b/src/rabbit_dead_letter.erl
@@ -139,7 +139,16 @@ update_x_death_header(Info, Headers) ->
                     end,
             rabbit_misc:set_table_value(
               Headers, <<"x-death">>, array,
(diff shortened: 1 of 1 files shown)


hairyhum:
Accidental PR to master

Issue #735, opened 2016-04-09
Specify broker version requirements for community plugins maintained by Pivotal


michaelklishin:
Now that #591 is merged, we can make community plugins maintained by our team specify broker version constraints.

When producing a release, we should specify the current release series. Note that for preview releases, the format is `a.b.c.d` and [the tests we have only cover `a.b.c`](https://github.com/rabbitmq/rabbitmq-test/pull/28/files).

hairyhum:
Given release process we have right now, we can do this:
- for plugins that ship with rabbitmq and automatically versioned, we can automatically set required broker version to current version during build.
- for plugins that ship independently, we can set `3.7.0` and `3.6.0` as basic supported versions and update them manually if there are some incompatibilities or new rabbitmq minor version.

As the result, there will be broker requirement for exact same version in plugins, shipped with rabbitmq and custom requirements in independently shipped plugins.

dumbbell:
Thanks a lot! This feature was badly needed!

PR #776, opened 2016-04-27
set broker requirements during package gen

hairyhum:
Part of #735
When building a release, replace empty `broker_version_requirements` with the release version.
Only empty list of requirements are being replaced.
Only plugins shipped with rabbitmq are affected.


dumbbell:
Currently, there is no plugin having the `broker_requirement` key in their `.app` file. Also, the commit message suggests the key would be added if it's missing, but the Makefile only replace an empty existing value.

Is it normal?

hairyhum:
Key will not be added. Only empty list will be replaced. I will submit PRs for plugins with empty requirements

dumbbell:
I'm going to merge the two `for` loops because they both modify the same files with sed, and I will merge the branch.

Issue #690, opened 2016-03-15
On Windows, ERLANG_HOME missing on upgrade


skizunov:
On Windows, if RabbitMQ is already installed (in my case I had 3.5.6). Then you install a newer RabbitMQ (I installed 3.6.1). The important `ERLANG_HOME` environment variable will be removed. This will prevent certain functionality from working correctly. The expected behavior is that the `ERLANG_HOME` environment variable be updated (in case ERLANG has also been replaced with a newer version).

I believe the problem lies in:
https://github.com/rabbitmq/rabbitmq-server/blob/master/packaging/windows-exe/rabbitmq_nsi.in

This logic is not correctly avoiding the delete of `ERLANG_HOME` in the upgrade case:

```
  ; If reinstalling immediately (e.g. program update) don't remove ERLANG_HOME environment variable
  ${If} $ISREINSTALL == ""
    DeleteRegValue ${env_hklm} ERLANG_HOME
    SendMessage ${HWND_BROADCAST} ${WM_WININICHANGE} 0 "STR:Environment" /TIMEOUT=5000
  ${EndIf}
```

Gsantomaggio:
hi @	skizunov 
Thank you for reporting, yes I think there is a bug here: https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-553/packaging/windows-exe/rabbitmq_nsi.in#L265 

We could:

1. Remove always the `ERLANG_HOME` key and re-setup it each time, by removing this code  https://github.com/rabbitmq/rabbitmq-server/blob/rabbitmq-server-553/packaging/windows-exe/rabbitmq_nsi.in#L243  and make sure to call again `findErlang` 

2. Avoid to remove the key during the setup 

The first option should be the best.

skizunov:
I like option 1 as well. The problem is that the remove ERLANG_HOME happens after the add ERLANG_HOME in the upgrade scenario. This is because the add ERLANG_HOME logic is in "Function findErlang" and this is called before the uninstall logic. To properly implement option 1, you will problem need to split "Function findErlang" into 2 functions, the second being something like "Function setErlangHome" which will be invoked after the uninstall logic.

PR #778, opened 2016-04-28
Windows, reconfigure ERLANG_HOME during the update

Gsantomaggio:
Fixes https://github.com/rabbitmq/rabbitmq-server/issues/690

Behaviour:

- Install: `ERLANG_HOME` is configured.
- Update: `ERLANG_HOME` is removed by the uninstall, and then re-configured.
- Uninstall: `ERLANG_HOME` is removed.

Tests:
- New installation  
- Update from `3.6.0`
- Unistall


dumbbell:
I removed the `un.StrContains` macro which now unused.

Issue #718, opened 2016-03-31
Signature of deb repo is weak


rpadovani:
Since the release of APT 1.2.7 keys which use SHA1 are considered weak and give a warning on `apt update`.

https://juliank.wordpress.com/2016/03/15/clarifications-and-updates-on-apt-sha1/

`W: http://www.rabbitmq.com/debian/dists/testing/InRelease: Signature by key F78372A06FF50C80464FC1B4F7B8CEA6056E8E56 uses weak digest algorithm (SHA1)`

michaelklishin:
Changing a signing key is a very big deal. Since this is a warning, I don't think it's worth the change for now.

noahhaon:
I think you misunderstand, the message digest is weak, you can keep your key you just need to sign using SHA2.

noahhaon:
Also the latest LTS version of Ubuntu will not install packages signed with SHA1 keys.

noahhaon:
If you can tell me what build tools you use to sign your debs, I would be happy to investigate this further.

michaelklishin:
@noahhaon thank you, I believe [we use gpg](https://github.com/rabbitmq/rabbitmq-public-umbrella/blob/f2320f2a42e703ffa3e01d0c7c70137b96e4f75a/Makefile#L495). @dumbbell has more info on the release toolchain.

shernshiou:
+1

GMTA:
As @noahhaon stated, Ubuntu 16.04 LTS users will be unable to install packages. AFAIK there is no workaround, apart from downloading and installing the package and all its dependencies by hand.

noahhaon:
Hi, sorry I didn't get an earlier notification on this.  I have some ideas and will do some digging today.

noahhaon:
@michaelklishin Unfortunately I think this change is most easily handled in your build environment.  You would need to add the following to the the `.gnupg/gpg.conf` file in the build user's homedir:

`digest-algo sha256` 

Disappointingly, `reprepro` appears to have no options to passing arguments to gpg as it uses an embedded version, `libgpgme`.  However, you can set the environment variable `GNUPGHOME` to point to the location of the directory housing `gpg.conf`, and drop the modified config file via `make` or similar.  Not ideal, but at least something you could get into git.

Relevant bits for build the deb repo appear to live here:

https://github.com/rabbitmq/rabbitmq-server/tree/master/packaging/debs/apt-repository

michaelklishin:
[...]

PR #792, opened 2016-05-12
Deb repository: Use SIGNING_KEY to select the repository signing key

dumbbell:
By default, honor the default key; usually it is specified in `gpg.conf`.

References #718.
[#118296861]


dumbbell:
This must be merged with rabbitmq/rabbitmq-public-umbrella#40.

PR #794, opened 2016-05-12
Private attributes usage in rabbitmq script

lefremova:
Required Pacemaker >= 1.1.13.
(The command 'attrd_updater' have '-p' option only since this version).

There are three types of rabbitmq attributes for pacemaker nodes:
	-'rabbit-master'
	-'rabbit-start-time'
	- timeouts:
		-'rabbit_list_channels_timeouts'
		-'rabbit_get_alarms_timeouts'
		-'rabbit_list_queues_timeouts'

Attributes with names 'rabbit-master' and 'rabbit-start-time' should be
public because we monitor this attributes in cycle for all nodes in our
script. All timeouts attributes were changed to private to avoid
unnecessary transitions.

Also, --lifetime and --node options were removed for attrd_updater as
'lifetime' for this command is always 'reboot' and 'node' default value
is local one.

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -1358,18 +1358,18 @@ is_master() {
 # separately. The second argument is used to distingush them.
 check_timeouts() {
     local op_rc=$1
-    local crm_attr_name=$2
+    local timeouts_attr_name=$2
     local op_name=$3
 
     if [ $op_rc -ne 124 -a $op_rc -ne 137 ]; then
-        ocf_run crm_attribute -N $THIS_PCMK_NODE -l reboot --name $crm_attr_name --update 0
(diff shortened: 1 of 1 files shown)


michaelklishin:
@dmitrymex can I get your blessing?

bogdando:
LGTM, I will take a TODO item for me as well, which is update docs, with something like : The "rabbitmqctl" control plane's consequent timeouts can be detected by setting the max_rabbitmqctl_timeouts>0 (default is 3). Requires a Pacemaker >=1.1.14, otherwise it does noting. Set the max_rabbitmqctl_timeouts=0 to enable it in a simplest detection mode for older versions. Note: A simplest mode makes monitors to report failed resources after a very first encountered rabbitmqctl timeout.

michaelklishin:
We are approaching the `3.6.2` release and docs can be updated independently. So I'll merge this tomorrow.

PR #801, opened 2016-05-16
Add explicit epmd dependency in systemd example unit

binarin:
Our testing shows that you can't rely on socket activation alone.

### DIFF
--- a/docs/rabbitmq-server.service.example
+++ b/docs/rabbitmq-server.service.example
@@ -1,7 +1,8 @@
 # systemd unit example
 [Unit]
 Description=RabbitMQ broker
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you. Just in time for `3.6.2`.

lemenkov:
@michaelklishin I think you should remove syslog.target (like we did already). Neither rabbitmq nor any dependent services in chain actually relies on syslog, so why keep it?

michaelklishin:
@lemenkov please submit a PR?

lemenkov:
@michaelklishin done. See PR #804.

PR #804, opened 2016-05-18
Remove systemd dependency on syslog target

lemenkov:
Neither rabbitmq nor any dependent services in chain actually relies on
syslog, so why keep it?

This tested in Fedora since 2016-01. See this commit:

http://pkgs.fedoraproject.org/cgit/rpms/rabbitmq-server.git/commit/?id=1477671

Nobody complained so far.

See also PR #801 

### DIFF
--- a/docs/rabbitmq-server.service.example
+++ b/docs/rabbitmq-server.service.example
@@ -1,7 +1,7 @@
 # systemd unit example
 [Unit]
 Description=RabbitMQ broker
-After=syslog.target network.target epmd@0.0.0.0.socket
+After=network.target epmd@0.0.0.0.socket
 Wants=network.target epmd@0.0.0.0.socket
 
 [Service]


michaelklishin:
@binarin WDYT?

@lemenkov should this also be applied to `stable`?

lemenkov:
@michaelklishin applying t to stable isn't absolutely necessary. Removing additional service startup dependencies will only decrease startup time, won't fix any issues. So this is just a minor improvement. Feel free to apply it there though but I'm not insisting :)

lemenkov:
Ok, I'm taking my words back :)
@michaelklishin please push it to stable as well (upcoming 3.6.2 I suppose, right?) - this would save us one additional patch (or even the entire service-file) from shipping with src.rpm.

michaelklishin:
@lemenkov no problem. I'm waiting for @binarin's opinion on this.

binarin:
LGTM. Especially given that there is no portable way to use syslog from erlang

michaelklishin:
Will cherry-pick to stable.

Issue #795, opened 2016-05-13
Infinity loop in synchronisation when priorities are higher than maximum in slave restart


dcorbacho:
When pushing messages with a priority higher than the maximum, mirror node process(es) will crash when a slave node is restarted. That causes an infinity loop as synchronization is not able to finish.

```
=ERROR REPORT==== 12-May-2016::17:04:02 ===
** Generic server <0.30173.0> terminating  
** Last message in was {'$gen_cast',
                           {sync_start,#Ref<8977.0.524290.86024>,
                               <8977.24530.3>}}
** When Server state == {state,
                         {amqqueue,
                          {resource,<<"/">>,queue,<<"p_2">>},
                          true,false,none,
                          [{<<"x-max-priority">>,long,3}],
                          <8977.18865.3>,
                          [<0.30135.0>,<8978.24713.6>],
                          [],
                          ['rabbit@t-srv-rabbit06','rabbit@t-srv-rabbit05'],
                          [{vhost,<<"/">>},
                           {name,<<"ha.all">>},

.......
                         {dict,0,16,16,8,80,48,
                          {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                          {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]}}},
                         {dict,0,16,16,8,80,48,
                          {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                          {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]}}},
                         {state,
                          {dict,1,16,16,8,80,48,
                           {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                           {{[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                             [[<8977.22136.3>|{delegate_12,<8977.22136.3>}]],
                             []}}},
                          delegate},
                         9462}
** Reason for termination ==
** {not_found,5}
```

Found while analyzing #687

michaelklishin:
@dcorbacho @Gsantomaggio is there anything left in #687 after this?

michaelklishin:
Nevermind my earlier comment, I now see a couple of recent comments in #687.

Issue #687, opened 2016-03-12
Priority queue fails to synchronize after node restart.


bharris47:
We are running a RabbitMQ cluster (3 nodes) with 3 mirrored, durable, priority queues. Restarting any of the nodes causes the priority queues to fall into an endless sync loop. 

The management console shows the queue synchronization progress going from 0% to complete. After a few seconds, the synchronization starts again.

We have tried disabling automatic sync and manually triggering sync, but experience the same behavior.

We initially believed they were failing to sync because there were still messages being sent to the queue. When we stop the producer, the behavior stops momentarily, but resumes immediately upon re-enabling the producer.

Our issue seems very similar to #366, but that is claimed to have been fixed in an earlier version.

Non-priority, durable queues do not exhibit this behavior.

Any help would be much appreciated.

Technical Details
- OS: CentOS Linux release 7.2.1511
- RabbitMQ: 3.6.1
- Erlang 18.2

Crash Details
```
** Reason for termination ==
** {{badmatch,-1},
    [{rabbit_mirror_queue_slave,update_delta,2,
                                [{file,"src/rabbit_mirror_queue_slave.erl"},
                                 {line,984}]},
     {rabbit_mirror_queue_slave,process_instruction,2,
                                [{file,"src/rabbit_mirror_queue_slave.erl"},
                                 {line,917}]},
     {rabbit_mirror_queue_slave,handle_cast,2,
                                [{file,"src/rabbit_mirror_queue_slave.erl"},
                                 {line,254}]},
     {gen_server2,handle_msg,2,[{file,"src/gen_server2.erl"},{line,1049}]},
     {proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,240}]}]}
```

Morthy:
Also occurring for us, with nearly the same setup/config but OS is Debian 7 Linux 3.6.6-xenu-ig7

epretha:
Hi,
We were not able to reproduce the issue with the information provided. It would be really helpful if you could kindly share with us the configuration applied and any script that you might have to reproduce the issue.

Could you also kindly send us the full log files for all the nodes in the cluster?
This will help us to understand the context under which the problem occurred.

Thank you very much in advance.

bharris47:
@epretha do you guys have any private channels where we could send the logs and other details as opposed to this public Github forum.

michaelklishin:
@bharris47 michael at RabbitMQ domain is one way.

[...]

PR #805, opened 2016-05-19
Use max priority for higher priorities when returning backing queue state

dcorbacho:
Fixes the slave synchronization, otherwise messages with priorities higher than max-priority are dropped during synchronization.

This causes a similar crash to #687, a `badmatch` with a positive value triggered by the [`depth`]( https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_mirror_queue_slave.erl#L940) request.

Related to #795, which fixed part of the support for priorities exceeding the maximum. See https://www.rabbitmq.com/priority.html.


michaelklishin:
No separate issue for this one => assigning milestone to the PR.

PR #808, opened 2016-05-20
Allow caching HiPE-compilation results

binarin:
That way HiPE compilation can be performed during package installation
and will not waste time during every startup.

rabbit_hipe is refactored to support both modes of compilation - during
every server startup or separately with caching in the filesystem.


src/rabbit_control_main.erl
```diff
@@ -383,6 +384,16 @@ action(rotate_logs, Node, Args = [Suffix], _Opts, Inform) ->
...
     call(Node, {rabbit, rotate_logs, Args});
 
+action(hipe_compile, _Node, [TargetDir], _Opts, _Inform) ->
+    ok = application:load(rabbit),
+    case rabbit_hipe:can_hipe_compile() of
+        true ->
+            {ok, _, _} = rabbit_hipe:compile_to_directory(TargetDir),
+            ok;
+        false ->
+            "HiPE compilation is not supported"
```
michaelklishin: Should this be `{error, Msg}`?

binarin: Yes, you're right. While printed result will be the same, it's better to be explicit.

PR #811, opened 2016-05-26
Switch testsuite to common_test

### LINKED ISSUE #725: Switch testsuite to common_test
The switch was already made in the Erlang client library in rabbitmq/rabbitmq-erlang-client#39. We will use `rabbit_ct_helpers` created during this first part. That's why it needs to be moved to rabbitmq-common.

dumbbell:
Fixes #725.
[#116526487]


dumbbell:
I still see transient failures on the Jenkins slave, but those existed before.

michaelklishin:
oops, this is against stable. Then merging in `master` wasn't a very good idea.

PR #835, opened 2016-06-09
Significantly bump max restarts intensity for worker_pool_sup

### LINKED ISSUE #834: Increase worker_pool_sup max restarts intensity
Currently it tolerates 10 restarts in 10 seconds. For worker processes that's way too low. Consider a system with 128 pool workers doing something involving network connections, e.g. talking to a LDAP server. When network goes down, under load it is easy to get 100s of unhandled exception in operations submitted to the pool. That means `worker_pool_sup` with current settings will terminate for good, failing all authn and authz operations until node is restarted.

To reproduce from a running node with a REPL:

```
supervisor:which_children(worker_pool_sup).
[worker_pool:submit_async(fun() -> 1 = 2 end) || N <- lists:seq(1, 500)].
supervisor:which_children(worker_pool_sup).
```

It would be more reasonable to tolerate up to 1K worker process crashes per second.

michaelklishin:
Fixes #834. The number is picked to tolerate up to 1K worker process failures per second. Per discussion with @dcorbacho.

### DIFF
--- a/src/worker_pool_sup.erl
+++ b/src/worker_pool_sup.erl
@@ -52,7 +52,7 @@ init([WCount, PoolName]) ->
     %% e.g. when a large worker pool used for network connections
     %% encounters a network failure. This is the case in the LDAP authentication
     %% backend plugin.
-    {ok, {{one_for_one, 10 * 1000, 10},
+    {ok, {{one_for_one, 1000, 1},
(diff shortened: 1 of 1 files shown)


dcorbacho:
@michaelklishin why not use `{one_for_one, 1000, 1}`? Otherwise, it is allowing 10k in 1s

michaelklishin:
@dcorbacho done.

Issue #802, opened 2016-05-17
Infinite loop in synchronization of priority queues


dcorbacho:
This crash generates an infinity loop in the synchronization of priority queues

Found while reproducing #687 

```
=INFO REPORT==== 17-May-2016::14:29:35 ===
Mirrored queue 'pri_q2_0' in vhost '/': Adding mirror on node 'rabbit@t-srv-rabbit05': <8977.27711.1>

=INFO REPORT==== 17-May-2016::14:29:35 ===
Mirrored queue 'pri_q2_0' in vhost '/': Adding mirror on node 'rabbit@t-srv-rabbit06': <8978.26069.1>

=ERROR REPORT==== 17-May-2016::14:29:35 ===
** Generic server <0.302.0> terminating
** Last message in was {init,{<0.148.0>,
                              [non_clean_shutdown,non_clean_shutdown,
                               non_clean_shutdown,non_clean_shutdown,
                               non_clean_shutdown,non_clean_shutdown,
                               non_clean_shutdown,non_clean_shutdown,
                               non_clean_shutdown,non_clean_shutdown,
                               non_clean_shutdown,non_clean_shutdown,
                               non_clean_shutdown,non_clean_shutdown,
.....
                               non_clean_shutdown,non_clean_shutdown,
                               non_clean_shutdown]}}
** When Server state == {q,{amqqueue,
                               {resource,<<"/">>,queue,<<"pri_gb2_2">>},
                               true,false,none,
                               [{<<"x-max-priority">>,signedint,256}],
                               <0.302.0>,[],[],[],
                               [{vhost,<<"/">>},
                                {name,<<"ha-all">>},
                                {pattern,<<>>},
                                {'apply-to',<<"all">>},
                                {definition,
                                    [{<<"ha-mode">>,<<"all">>},
                                     {<<"ha-sync-mode">>,<<"automatic">>}]},
                                {priority,0}],
                               [],[],live},
                           none,false,undefined,undefined,
                           {state,
                               {queue,[],[],0},
                               {active,-576460741042383,1.0}},
                           undefined,undefined,undefined,undefined,
                           {state,fine,5000,undefined},
                           {0,nil},
                           undefined,undefined,undefined,
                           {state,
                               {dict,0,16,16,8,80,48,
                                   {[],[],[],[],[],[],[],[],[],[],[],[],[],[],
                                    [],[]},
                                   {{[],[],[],[],[],[],[],[],[],[],[],[],[],
[...]

PR #857, opened 2016-06-24
Infinity loop in synchronisation of priority queues

dcorbacho:
This solves two of the bugs in #802, the third one can only be fixed in `master` as detecting a stale slave requires changes on a mnesia table.

The unit tests try to trigger the bugs in a multinode test. The `invoke` crash seems to be triggered consistently, but it cannot be guaranteed on any environment. Thus, the suite also contains an individual test case to force it.


michaelklishin:
I'm getting test failures, shared details on Slack.

Issue #863, opened 2016-06-27
Ensure old incarnations of mirrors are stopped before new ones start


michaelklishin:
Follow-up to #802, see that thread for discussion.

PR #864, opened 2016-06-27
Ensure old incarnations of mirrors are stopped before new ones start

michaelklishin:
This is a follow-up #802 that can only go into `3.7.0`. Fixes https://github.com/rabbitmq/rabbitmq-server/issues/863.


michaelklishin:
Closed in favour of https://github.com/rabbitmq/rabbitmq-server/issues/863.

Issue #612, opened 2016-02-05
Tune scheduling flags for Erlang VM


sega-yarkin:
By default the Erlang VM doesn't pin their threads to vCPU and when multi-core server has big load we have huge context switching (8.3k per vCPU in our case). But the Erlang VM gives to us ability to configure [scheduler bind type](http://erlang.org/doc/man/erl.html#id176181), this can helps to reduce context switching. The flag name is `+stbt`, in our installation we're using `+stbt ts`, it reduced context switching for 4 times (2k per vCPU).

It looks like easy tuning which gives a nice result for reducing unnecessary CPU context switching and I think it can be considered for including to default configuration. But `ts` value might be not the best choice for every platforms, it should be tested before (we've tasted only VMware virtual machines).

https://groups.google.com/forum/?hl=en#!topic/rabbitmq-users/WIprYZF5T8Q

michaelklishin:
@dumbbell @hairyhum @essen @bdshroyer any objections?

hairyhum:
I think we can make `db` bind type default and introduce environment variable to change it.

michaelklishin:
I think @sega-yarkin's point is that it should be `ts` by default. Introducing an environment variable for easy overriding is a good idea.

hairyhum:
According to docs `db` and `ts` will have different behaviour only on NUMA nodes. But `db` is called default, so it's more expected.

michaelklishin:
I don't think many users even know you can tune scheduler-to-CPU bindings, so there are no particular expectations. @sega-yarkin provides a realistic example where the default is clearly sub-optimal.

Perhaps @dcorbacho @Dzol and @Ayanda-D have an opinion on this?

dcorbacho:
I think it is a good idea to add that option. In a particular case, I found a good performance improvement in a load testing tool by using `+sbt db`. However, care must be taken on the implementation of the flag in the rabbit side, because the default may not be optimal in some platforms (add a warning?) or the option not allowed (if my memory doesn't fail, OS X doesn't accept the `+sbt db` flag and fails on startup).

hairyhum:
@dcorbacho `+stbt` should be more safe. `+sbt` will fail to start VM if runtime cannot support scheduler binding or cpu topology, while `+stbt` will just ignore it.

dcorbacho:
Thanks @hairyhum, I hadn't notice that option before!

sega-yarkin:
[...]

PR #873, opened 2016-07-04
Tune scheduling bind flags for Erlang VM

dcorbacho:
Pending final decision on the default flag in #612.


scripts/rabbitmq-env
```diff
@@ -62,8 +62,11 @@ RABBITMQ_HOME="$(rmq_realpath "${RABBITMQ_SCRIPTS_DIR}/..")"
 ## Set defaults
 . ${RABBITMQ_SCRIPTS_DIR}/rabbitmq-defaults
 
+DEFAULT_SCHEDULING_BIND_TYPE_FLAG="tnnps"
+[ "x" = "x$RABBITMQ_SCHEDULING_BIND_TYPE_FLAG" ] && RABBITMQ_SCHEDULING_BIND_TYPE_FLAG=${DEFAULT_SCHEDULING_BIND_TYPE_FLAG}
+
 ## Common defaults
-SERVER_ERL_ARGS="+P 1048576"
+SERVER_ERL_ARGS="+P 1048576 +stbt $RABBITMQ_SCHEDULING_BIND_TYPE_FLAG "
```
michaelklishin: I'd rename this to `RABBITMQ_SCHEDULER_BIND_TYPE`. That's how the docs describe this option (whether that's the clearest name from the English grammar point of view). I don't think we use the `_FLAG` prefix for other values.

scripts/rabbitmq-server.bat
```diff
@@ -112,6 +112,10 @@ if "!ENV_OK!"=="false" (
     EXIT /b 78
 )
 
+if "!RABBITMQ_SCHEDULING_BIND_TYPE_FLAG!"=="" (
```
michaelklishin: Any reason why this didn't go into [rabbitmq-env.bat](https://github.com/rabbitmq/rabbitmq-server/blob/master/scripts/rabbitmq-env.bat)?

dcorbacho: Not that I'm aware, but I set it here the same way `IO_THREAD_POOL_SIZE` is defined just [above](https://github.com/rabbitmq/rabbitmq-server/pull/873/files#diff-bf0f372f52d6c544a7ae5214db1c5f21L100)

PR #878, opened 2016-07-07
Improve logging performance

binarin:
Noticed dobule atom_to_list/list_to_atom roundtrip while profiling. Not
too significant, but it still caught my eye because it was happening for
every line in log.

I think this a missing part for
https://github.com/rabbitmq/rabbitmq-common/commit/46c78094c69a769ba35bea86ed55bf6c48e8a163

And with unrolling predefined sinks it should avoid
atom_to_list/list_to_atom roundtrip completely.


lemenkov:
Nice catch :)

PR #883, opened 2016-07-14
Make `node_health_check` completely node-local

### LINKED ISSUE #818: node_health_check should list local queues only
Because the command is node-scoped by definition.

@binarin WDYT?

binarin:
Changes to rabbitmq-common - https://github.com/rabbitmq/rabbitmq-common/pull/118

Fixes #818

Also:
- added test-suite
- fix queues health-checking - `rabbit_amqqueue:info_all` was expecting
      vhost argument, but was passed `[pid]` instead. So no health check was
      actually performed for queues. Proper way is to iterate over all vhosts.
- delegate all timeout/error handling to `rabbit_cli`, so no more magic
      `70` and `68` and no more `case`-ing on `badrpc` that is already
      present in other places.
-  Make rabbitmqctl timeout arg handling more consistent - global timeout `?RPC_TIMEOUT` was not used, because default value of
      `"infinity"` was always introduced via `?TIMEOUT_DEF`. Now `infinity` is
      used for commands without timeout support, and `?RPC_TIMEOUT`
      otherwise.
- `?COMMANDS_WITH_TIMEOUT` now can contain per-command default values
      for timeout, using tuple `{Command, DefaultTimeoutInMilliSeconds}`
      instead of just `Command`.

### DIFF
--- a/include/rabbit_cli.hrl
+++ b/include/rabbit_cli.hrl
@@ -34,7 +34,7 @@
 -define(NODE_DEF(Node), {?NODE_OPT, {option, Node}}).
 -define(QUIET_DEF, {?QUIET_OPT, flag}).
 -define(VHOST_DEF, {?VHOST_OPT, {option, "/"}}).
--define(TIMEOUT_DEF, {?TIMEOUT_OPT, {option, "infinity"}}).
+-define(TIMEOUT_DEF, {?TIMEOUT_OPT, {option, use_default}}).
 
 -define(VERBOSE_DEF, {?VERBOSE_OPT, flag}).
 -define(MINIMAL_DEF, {?MINIMAL_OPT, flag}).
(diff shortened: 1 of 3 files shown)


michaelklishin:
Corrected `end_per_testsuite` a bit and merged. Thank you.

michaelklishin:
@binarin this fails to merge into master cleanly. Would you have a chance to resolve conflicts?

binarin:
Should I just merge current `stable` into `master`? It also pushes some commits that are not authored by me.

binarin:
Here it is - https://github.com/rabbitmq/rabbitmq-server/pull/884

michaelklishin:
@binarin yup, simply merge stable into master and resolve conflicts. Thank you.

michaelklishin:
@binarin just in case: for both server and common :)

binarin:
Done.

Issue #627, opened 2016-02-15
Force close connections when vhost is deleted


michaelklishin:
We currently remove all resources but do not close connections. With connection tracking introduced in #500, that become a possibility, so we should do it.

Issue #607, opened 2016-02-04
Per-user connection limit


johnfoldager:
It would be really great if it was possible to limit the number of connections/channels for a specific user.

michaelklishin:
You can do this via HTTP API today (poll and force close connections). However, I expected to see this after #500 and they have a lot of similarities.

We won't be providing a channel limit per user but you can limit max number of channels per connection via `rabbit.channel_max`.

michaelklishin:
@johnfoldager in your opinion, should this limit be global or scoped per vhost? We see a lot of interest in per-vhost connection limit (already done, #500) but there is no consensus on this one.

michaelklishin:
We'd like to understand the use cases better before we spend more than a spike worth amount of time on this => clearing milestone.

wmoran-uatc:
With the current setup, an application with a memory leak can crash RabbitMQ when it runs out of file descriptors. If a per-user connection limit could be enforced, it could be guaranteed that one application with a connection leak won't crash Rabbit (thus causing problems for all applications that use it). This would provide a more stable backend for systems that are developed rapidly enough that bugs occasionally sneak through.

michaelklishin:
@wmoran-uatc we have introduced per-vhost connection limit in #500. However, we now fail to come up with a compelling use case for this per-user limit and there's no time to fit it into the `3.7.0`, too. So it is on the back burner but the benefits of connection limits are clear to our team :)

wmoran-uatc:
While per-vhost is an excellent feature, it still fails to provide protection in the case where multiple teams are creating software that connects to the same vhost. With per-vhost, a single buggy app could eat up all the connections for a particular vhost, thus negatively impacting other applications that need to use that vhost. With per-user limits, each application could be limited independently of any other.

michaelklishin:
@wmoran-uatc we will consider this feature for a future release, thank you.

cwhsu1984:
I would love to see rabbitmq to have this feature, I currently plan to let users connect to rabbitmq through Web STOMP. However, I just want to have only one connection per user to rabbitmq to prevent users from draining the resources(cpu, memory) on the rabbitmq node. Although, I might be able to do this by closing user connection when a user tries to login to our server, still, it would be great if rabbitmq support this feature.

jesferman:
Is there any planning to include this feature in some RabbitMQ release?

michaelklishin:
@jesferman not into 3.7.0, then we will see.

[...]

Issue #501, opened 2015-12-15
Per-vhost limit on the number of queues


michaelklishin:
See #498 for background and #500 for a related issue.This should perhaps be configured via a policy to be more flexible. Note that this means extending policies to vhosts, so using `effort-medium` as it can take more than a couple of days.

michaelklishin:
Moving to `3.7.0` as it shares changes with #500.

PR #891, opened 2016-07-22
Keep track of connections, introduce per-vhost limits

### LINKED ISSUE #500: Per-vhost connection limit
See #498 for background. We already have a channel limit but this should perhaps be configured via a policy to be more flexible. Note that this means extending policies to vhosts, so using `effort-medium` as it can take more than a couple of days.

michaelklishin:
This introduces (client) connection tracking that's aware of connection

 * vhost
 * username
 * protocol

and so on across the cluster. We already collect much of this information
in node-local state but that's not sufficient for #500, #607, and related features.

This changes internal database schema and thus can only go into `3.7.0`.

This also makes sure that when a vhost is deleted, we force close all connections in it.

This is a 3rd iteration on the design. All connections and counters are tracked using
node-specific tables which are *only* updated on the node itself but read by all cluster nodes.
This avoids concurrent counter updates without going all the way to CRDTs.

The implementation isn't perfect but I run out of practical test cases to add. Points of improvement for the future:

 * Parallel folding over per-node connection tracking tables
 * Management plugin extensions related to limit management
 * `rabbitmqctl status` and `report` entries related to limits
 * As with many other things, replacing Mnesia would open new possibilities but it's a long term goal

Fixes #500, #627.
Lays ground work for #501, #607.

Depends on https://github.com/rabbitmq/rabbitmq-common/pull/121.


michaelklishin:
Feel free to QA but I have one more test case to add => do not merge yet.

michaelklishin:
Also: this has a `rabbitmqctl(1)` entry but still needs a doc guide.

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,229 @@
...
+
+register_connection(#tracked_connection{vhost = VHost, id = ConnId} = Conn) ->
+    rabbit_misc:execute_mnesia_transaction(
+      fun() ->
+        %% upsert
+              case mnesia:dirty_read(?TABLE, ConnId) of
+                  []    ->
+                    mnesia:write(?TABLE, Conn, write),
+                    mnesia:dirty_update_counter(
+                      rabbit_tracked_connection_per_vhost, VHost, 1);
```
dumbbell: You probably want to use `?PER_VHOST_COUNTER_TABLE` here.

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,229 @@
...
+-spec on_node_down(node()) -> ok.
+
+on_node_down(Node) ->
+  case lists:member(Node, nodes()) of
+        false ->
+          Cs = list_on_node(Node),
+          rabbit_log:info(
+            "Node ~p is down, unregistering ~p client connections~n",
+            [Node, length(Cs)]),
+          [unregister_connection(Id) || #tracked_connection{id = Id} <- Cs],
```
dumbbell: It would be more efficient to unregister all connections in a single transaction.

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,229 @@
...
+        true  -> rabbit_log:info(
+                   "Keeping ~s connections: the node is already back~n", [Node])
+    end.
+
+-spec on_node_up(node()) -> ok.
+on_node_up(Node) ->
+  rabbit_connection_tracker:reregister(Node),
+  ok.
+
+-spec is_over_connection_limit(rabbit_types:vhost()) -> boolean().
```
dumbbell: The spec should be:
```erlang
-spec is_over_connection_limit(rabbit_types:vhost()) -> {true, non_neg_integer()} | false.
```

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,229 @@
...
+
+is_over_connection_limit(VirtualHost) ->
+    ConnectionCount = count_connections_in(VirtualHost),
+    case rabbit_vhost_limit:connection_limit(VirtualHost) of
+        undefined   -> false;
+        {ok, Limit} -> case {ConnectionCount, ConnectionCount >= Limit} of
+                           %% 0 = no limit
+                           {0, _}     -> false;
```
dumbbell: What about using -1 as the "no limit" value or even a more meaninful value such as `undefined` or `false`? 0 could be used to block any connections to a vhost.

Sorry if I missed a previous discussion about this.

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,229 @@
...
+-spec count_connections_in(rabbit_types:vhost()) -> non_neg_integer().
+
+count_connections_in(VirtualHost) ->
+    try
+        case mnesia:transaction(
+               fun() ->
+                       case mnesia:dirty_read(
+                              {?PER_VHOST_COUNTER_TABLE,
+                               VirtualHost}) of
+                           []    -> 0;
```
dumbbell: The "no limit" value could be a macro so we understand its meaning everywhere.

src/rabbit_connection_tracking_handler.erl
```diff
@@ -0,0 +1,101 @@
...
+%% on its local node. The primary goal here is to decouple connection
+%% tracking from rabbit_reader in rabbit_common.
+%%
+%% Events from other nodes are ignored.
+
+%% This module keeps track of connection creation and termination events
+%% on its local node. The primary goal here is to decouple connection
+%% tracking from rabbit_reader in rabbit_common.
+%%
+%% Events from other nodes are ignored.
```
dumbbell: Comments are duplicated.

src/rabbit_control_main.erl
```diff
@@ -544,6 +547,18 @@ action(clear_policy, Node, [Key], Opts, Inform) ->
     Inform("Clearing policy ~p", [Key]),
     rpc_call(Node, rabbit_policy, delete, [VHostArg, list_to_binary(Key)]);
 
+action(set_vhost_limits, Node, [Defn], Opts, Inform) ->
+    Msg = "Setting vhost limits for vhost ~p",
+    VHostArg = list_to_binary(proplists:get_value(?VHOST_OPT, Opts)),
+    Inform(Msg, [VHostArg]),
```
dumbbell: `Msg` is being used once, you can probably get rid of it.

src/rabbit_upgrade_functions.erl
```diff
@@ -88,9 +91,36 @@
...
+
+tracked_connection_per_vhost() ->
+    create(tracked_connection_per_vhost, [{record_name, tracked_connection_per_vhost},
+                                          {attributes, [vhost, connection_count]}]).
+
+%% replaces vhost.dummy (used to avoid having a single-field record
+%% which Mnesia doesn't like) with vhost.limits (which is actually
+%% used)
+vhost_limits() ->
+    io:format("vhost_limits vhost_limits vhost_limits~n"),
```
dumbbell: Is this a debugging message?

test/per_vhost_connection_limit_SUITE.erl
```diff
@@ -0,0 +1,660 @@
...
+all() ->
+    [
+     {group, cluster_size_1},
+     {group, cluster_size_2}
+    ].
+
+groups() ->
+    [
+      {cluster_size_1, [], [
+          most_basic_single_node_connection_count_test,
```
dumbbell: Please do not use the '_test' suffix to not confuse the testcases with EUnit tests.

test/per_vhost_connection_limit_SUITE.erl
```diff
@@ -0,0 +1,660 @@
...
+    Config1 = rabbit_ct_helpers:set_config(Config, [
+                                                    {rmq_nodes_count, 1},
+                                                    {rmq_nodename_suffix, Suffix}
+      ]),
+    rabbit_ct_helpers:run_steps(Config1,
+      rabbit_ct_broker_helpers:setup_steps() ++
+      rabbit_ct_client_helpers:setup_steps());
+init_per_group(cluster_size_2, Config) ->
+    init_per_multinode_group(cluster_size_2, Config, 2);
+init_per_group(partition_handling, Config) ->
```
dumbbell: There is no `partition_handling` group.

test/per_vhost_connection_limit_SUITE.erl
```diff
@@ -0,0 +1,660 @@
...
+
+init_per_group(cluster_size_1, Config) ->
+    Suffix = rabbit_ct_helpers:testcase_absname(Config, "", "-"),
+    Config1 = rabbit_ct_helpers:set_config(Config, [
+                                                    {rmq_nodes_count, 1},
+                                                    {rmq_nodename_suffix, Suffix}
+      ]),
+    rabbit_ct_helpers:run_steps(Config1,
+      rabbit_ct_broker_helpers:setup_steps() ++
+      rabbit_ct_client_helpers:setup_steps());
```
dumbbell: You could call `init_per_multinode_group()` instead of duplicating the code.

test/per_vhost_connection_limit_SUITE.erl
```diff
@@ -0,0 +1,660 @@
...
+    VHost2 = <<"vhost2">>,
+
+    set_up_vhost(Config, VHost1),
+    set_up_vhost(Config, VHost2),
+
+    ?assertEqual(0, count_connections_in(Config, VHost1)),
+    ?assertEqual(0, count_connections_in(Config, VHost2)),
+
+    Conn1 = open_unmanaged_connection(Config, 0, VHost1),
+    ?assertEqual(1, count_connections_in(Config, VHost1)),
```
dumbbell: Connection counts should be verified for both virtual hosts.

src/rabbit_connection_tracker.erl
```diff
@@ -0,0 +1,99 @@
...
+
+code_change(_OldVsn, State, _Extra) ->
+  {ok, State}.
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+reregister_connection(Conn) ->
+  try
```
Ayanda-D: @michaelklishin message sending to a Pid never fails, so that error will never occur. You're probably better off validating with `is_process_alive(Conn)` instead, prior to attempting to reregister, if you want to log any error here.

michaelklishin: Thank you.

michaelklishin:
This approach suffers from the lost updates problem when connections from a failed node are removed concurrently. We have evaluated a couple of approaches and the most prominent so far seems to be:

 * Have a pair of connection tracking tables per node
 * Every node updates its own tables
 * Every node reads from all tables in order to come up with a total count

So, one writer, many readers. This should also make it possible to eliminate the connection re-registration mechanism in place today.

src/rabbit_connection_tracking_handler.erl
```diff
@@ -0,0 +1,101 @@
...
+        {proplists:get_value(node, Details),
+         proplists:get_value(name, Details)}),
+    {ok, State};
+handle_event(#event{type = vhost_deleted, props = Details}, State) ->
+    VHost = proplists:get_value(name, Details),
+    rabbit_log_connection:info("Closing all connections in vhost '~s' because it's being deleted", [VHost]),
+    case rabbit_connection_tracking:list(VHost) of
+      [] -> {ok, State};
+      Cs ->
+        [rabbit_networking:close_connection(Pid, rabbit_misc:format("vhost '~s' is deleted", [VHost])) || #tracked_connection{pid = Pid} <- Cs],
```
Ayanda-D: is a case statement is necessary here? if `Cs = []`, then nothing to evaluate in the list comprehension, and you can just return `{ok, State}`.

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,311 @@
...
+    Node = proplists:get_value(node, EventDetails),
+    #tracked_connection{id           = {Node, Name},
+                        name         = Name,
+                        node         = Node,
+                        vhost        = proplists:get_value(vhost, EventDetails),
+                        username     = proplists:get_value(user, EventDetails),
+                        connected_at = proplists:get_value(connected_at, EventDetails),
+                        pid          = proplists:get_value(pid, EventDetails),
+                        peer_host    = proplists:get_value(peer_host, EventDetails),
+                        peer_port    = proplists:get_value(peer_port, EventDetails)}.
```
Ayanda-D: this might not lie on the critical path, but use of `proplists:get_value/2` seems quite extensive. how about using `rabbit_misc:pget/2` instead, considering that it was optimised in https://github.com/rabbitmq/rabbitmq-common/pull/51

michaelklishin: Good idea, thank you.

src/rabbit_connection_tracking.erl
michaelklishin: OK, that makes sense.

src/rabbit_table.erl
```diff
@@ -297,9 +306,24 @@ definitions() ->
...
+     %% All data in this table is transient.
+     {rabbit_tracked_connection,
+      [{record_name, tracked_connection},
+       {attributes, record_info(fields, tracked_connection)},
+       {match, #tracked_connection{_ = '_'}}]},
+
+     {rabbit_tracked_connection_per_vhost,
+      [{record_name, tracked_connection_per_vhost},
+       {attributes, record_info(fields, tracked_connection_per_vhost)},
+       {match, #tracked_connection_per_vhost{_ = '_'}}]}
```
dumbbell: Are those two Mnesia tables still used by the patch? I can't find where.

michaelklishin: No, they can be deleted.

michaelklishin:
I've missed one more obvious improvement: when a node is deleted from the cluster, its tables should be deleted, too.

dumbbell:
The connection tracking tables are only created when a cluster is formed. Don't we need those tables for single-node instances?

michaelklishin:
@dumbbell there is a boot step that does it for the starting node (`rabbit.erl:181`).

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,320 @@
...
+
+-spec ensure_tracked_connections_table_for_node(node()) -> ok.
+
+ensure_tracked_connections_table_for_node(Node) ->
+    TableName = tracked_connection_table_name_for(Node),
+    case mnesia:create_table(TableName, [{record_name, tracked_connection},
+                                       {attributes, [id, node, vhost, name,
+                                                     pid, protocol,
+                                                     peer_host, peer_port,
+                                                     username, connected_at]}]) of
```
dumbbell: Any reason for not using `record_info(fields, ...)` for those two table creations?

michaelklishin: Perhaps here we can use it, unlike in upgrade_functions.

michaelklishin:
This affects connection management over HTTP API. Investigating.

michaelklishin:
@dumbbell this is ready for another round.

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,345 @@
...
+                      [Val] -> Val#tracked_connection_per_vhost.connection_count
+                 end
+               end) of
+                 {atomic,  Val}     -> Val;
+                 {aborted, _Reason} -> 0
+            end
+          catch _ -> 0
+          end
+        end, rabbit_mnesia:cluster_nodes(running)),
+      lists:foldl(fun(X, Acc) -> Acc + X end, 0, Ns)
```
Ayanda-D: @michaelklishin you're going through 2 lists just to calculate number of connections? imagine how much this will impact performance (per connection open request) as cluster size increases? can you use `lists:foldl/3` only please, with connection count as your accumulator. there's no need for `lists:map/2` here.

michaelklishin: Sorry but you are worrying about the wrong thing here. What really needs to happen is doing a parallel map but for small clusters it will likely be less efficient so I decided that it can be done at a later point.

It's not common to see clusters with more than maybe 7-9 nodes and it's unlikely to be a major contributor to how many connections/second can be accepted. Plus if you don't have any limits configured, none of this code will be executed.

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,340 @@
...
+
+-spec list(rabbit_types:vhost()) -> [rabbit_types:tracked_connection()].
+
+list(VHost) ->
+  Chunks = lists:map(
+    fun (Node) ->
+      Tab = tracked_connection_table_name_for(Node),
+      mnesia:dirty_match_object(Tab, #tracked_connection{vhost = VHost, _ = '_'})
+    end, rabbit_mnesia:cluster_nodes(running)),
+  lists:foldl(fun(Chunk, Acc) -> Acc ++ Chunk end, [], Chunks).
```
Ayanda-D: you can also make `Chunks` your accumulator in `lists:foldl/3` for both your listing functions (unless you were planning something else)

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,340 @@
...
+-spec is_over_connection_limit(rabbit_types:vhost()) -> {true, non_neg_integer()} | false.
+
+is_over_connection_limit(VirtualHost) ->
+    case rabbit_vhost_limit:connection_limit(VirtualHost) of
+        %% no limit configured
+        undefined                                            -> false;
+        %% with limit = 0, no connections are allowed
+        {ok, 0}                                              -> {true, 0};
+        {ok, Limit} when is_integer(Limit) andalso Limit > 0 ->
+          ConnectionCount = count_connections_in(VirtualHost),
```
Ayanda-D: on coding style, can we stick to our general indentation rule, of the default Erlang mode for Emacs? your indentation also seems to differ from function to function. for example `boot/0` and `ensure_tracked_connections_table_for_node/1` appear to use different indentation.

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,340 @@
...
+
+-spec delete_per_vhost_tracked_connections_table_for_node(node()) -> ok.
+
+delete_per_vhost_tracked_connections_table_for_node(Node) ->
+    TableName = tracked_connection_per_vhost_table_name_for(Node),
+    case mnesia:delete_table(TableName) of
+      {atomic, ok}              -> ok;
+      {aborted, {no_exists, _}} -> ok;
+      {aborted, Error} ->
+        rabbit_log:error("Failed to delete a per-vhost tracked connection table for node ~p: ~p", [Node, Error]),
```
Ayanda-D: also on coding style, please stick to the max 80 chars per line convention.

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,340 @@
...
+      {aborted, Error} ->
+        rabbit_log:error("Failed to create a tracked connection table for node ~p: ~p", [Node, Error]),
+        ok
+    end.
+
+
+-spec ensure_per_vhost_tracked_connections_table_for_node(node()) -> ok.
+
+ensure_per_vhost_tracked_connections_table_for_node(Node) ->
+    TableName = tracked_connection_per_vhost_table_name_for(Node),
```
Ayanda-D: consider referencing your table name creating functions using macros. would make code a bit more easier to read.

src/rabbit_connection_tracking.erl
```diff
@@ -0,0 +1,340 @@
...
+    case mnesia:create_table(TableName, [{record_name, tracked_connection_per_vhost},
+                                          {attributes, [vhost, connection_count]}]) of
```
Ayanda-D: please also use `record_info` for specifying `tracked_connection_per_vhost` table attributes. otherwise if ever we decide to add more attributes to the table/record, then we'd have to hard code them here as well.

src/rabbit_connection_tracking.erl
michaelklishin: We are no longer living in the age of 80 character wide IBM punch cards and terminals.

src/rabbit_connection_tracking.erl
michaelklishin: Sorry but I don't see how using a macro instead of an actual function would make anything clearer.

Ayanda-D: Okay, your preference. But usually, using macros for table names, or directly specifying the table name atom, is general practice. I think `?TRACKED_CONN_TABLE(N)` would read better.

src/rabbit_connection_tracking.erl
Ayanda-D: then why are we still retaining this rule in our header comments? and besides, most of the code base still follows this convention.

PR #892, opened 2016-07-22
Fix longname-mode on hosts without detectable FQDN

### LINKED ISSUE #890: rabbitmq-server and rabbitmqctl broken with long names and some bizarre DNS configurations
When you have a strangely configured system where `hostname -f` returns name without a dot in it (hello, k8s), `erl -name test` fails to start. Some rabbit code expects that this magic will work, i.e. the server prelaunch process and the rabbitmqctl. 

Simplest fix is to unconditionally add `@localhost` when long names are used: in prelaunch and other places. Will it be accepted to `stable`?

Another option is to add yet another environment variable, where you can provide this suffix in case where you know that your name resolving is somewhat broken.

WDYT?

binarin:
Server startup and CLI tools fail in longnames-mode if erlang is not
able to determine host FQDN (with at least one dot in it).

E.g. this can happen when you want to assemble a cluster using only
IP-addresses, and you completely don't care about FQDNs.

And it was not possible to alleviate this situation using any options
from http://erlang.org/doc/apps/erts/inet_cfg.html

Fixes #890

### DIFF
--- a/scripts/rabbitmq-server
+++ b/scripts/rabbitmq-server
@@ -62,6 +62,17 @@ RABBITMQ_EBIN_ROOT="${RABBITMQ_HOME}/ebin"
 
 set +e
 
+# `net_kernel:start/1` will fail in `longnames` mode when erlang is
+# unable to determine FQDN of a node (with a dot in it). But `erl`
+# itself has some magic that still allow it to start when you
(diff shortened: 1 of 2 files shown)


dumbbell:
Thank you @binarin! Sorry for the delay...

PR #896, opened 2016-07-28
Bump default VM atom table limit to 5M

### LINKED ISSUE #895: Bump default Erlang VM atom table limit
We have fixed a couple of cases that could lead to atom table exhaustion in the last few months but also discovered legitimate cases where a very long running node can be driven out of available atoms.

Resolving all of them does not seem realistic but we can reduce the probability of this issue occurring by bumping the limit, even though 99% of users won't ever go over the default value (~ 1M).

Per discussion with the team.

michaelklishin:
See #895 for background and reasoning.

Fixes #895.


kjnilsson:
Ok I’ve started the service on windows with that setting and verified it using: `rabbitmqctl eval 'string:tokens(binary_to_list(erlang:system_info(info)),"\n").'`

In the output I found this:

```32665\ndepth: 10\n=i",
 "dex_table:atom_tab\nsize: 32768\nlimit: 5000000\ne"```

I think I got the escaping of the token delimiter on windows a bit wrong but it does show the `atom_tab` limit to be of the correct value.

PR #900, opened 2016-07-29
fixed typo in config example comment

johnlonganecker:


michaelklishin:
Thank you.

michaelklishin:
OK, so this is in the lager section => no need to cherry-pick to `stable`.

Issue #851, opened 2016-06-21
Add more filtering options for `list_queues`


binarin:
We believe it'd be useful to have the ability to list only queues whose master is on the current node, specifically for monitoring, where every host has some collection agent running. It doesn't make sense to collect duplicate statistics from global `list_queues` result.

michaelklishin:
@binarin with #911 merged, should this be closed?

binarin:
I don't know. Regexp filtering is not implemented yet and it can be useful for some use-cases. But we are no longer interested in this feature.

michaelklishin:
OK, let's edit issue description and close it then.

PR #911, opened 2016-08-09
Add support for listing only local queues

binarin:
Partially implements https://github.com/rabbitmq/rabbitmq-server/issues/851

- Made old `--online`/`--offline` options mutually exclusive between
  themselves and the new `--local` option
- Added documentation both for the old and the new option
- Fixed some ugly indentation in generated usage (only `set_policy`
  wrapped line remains unfixed)
- Added integration test suite for `rabbitmqctl list_queues`

### DIFF
--- a/docs/rabbitmqctl.1.xml
+++ b/docs/rabbitmqctl.1.xml
@@ -749,9 +749,7 @@
 
         <varlistentry>
           <term>
-          <cmdsynopsis>
-            <command>authenticate_user</command> <arg choice="req"><replaceable>username</replaceable></arg> <arg choice="req"><replaceable>password</replaceable></arg>
-          </cmdsynopsis>
+          <cmdsynopsis><command>authenticate_user</command> <arg choice="req"><replaceable>username</replaceable></arg> <arg choice="req"><replaceable>password</replaceable></arg></cmdsynopsis>
           </term>
           <listitem>
             <variablelist>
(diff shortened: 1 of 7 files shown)


docs/rabbitmqctl.1.xml
```diff
@@ -1237,14 +1235,51 @@
...
+              <varlistentry>
+                <term><cmdsynopsis><arg choice="opt">--online</arg></cmdsynopsis></term>
+                <listitem>
+                  <para>
+                    List queues that are currently live.
+                  </para>
+                </listitem>
+              </varlistentry>
+              <varlistentry>
+                <term><cmdsynopsis><arg choice="opt">--offline</arg></cmdsynopsis></term>
```
michaelklishin: This seems to be a copy-and-paste artifact?

binarin: Yes, here should be `--local` instead. Fixed.

test/unit_SUITE.erl
```diff
@@ -135,26 +135,32 @@ check_parse_arguments(ExpRes, Fun, As) ->
...
+mutually_exclusive_flags_parsing(_Config) ->
+    Spec = [{"--online", online}
+           ,{"--offline", offline}
+           ,{"--local", local}],
+    Default = all,
+    Cases =[{["--online"], {ok, online}}
+           ,{[], {ok, Default}}
+           ,{["--offline"], {ok, offline}}
+           ,{["--local"], {ok, local}}
+           ,{["--offline", "--local"], {error, "Options '--offline' and '--local' are mutually exclusive"}}
```
dcorbacho: These error checks are too closely tied to the implementation by verifying the exact message. For long term maintenance, it would be better to match only the `error` part in the tuple.

docs/rabbitmqctl.1.xml
```diff
@@ -1237,14 +1235,51 @@
...
+            <cmdsynopsis><command>list_queues</command> <arg choice="opt">-p <replaceable>vhost</replaceable></arg> <group choice="opt"><arg>--offline</arg><arg>--online</arg><arg>--local</arg></group> <arg choice="opt" role="usage-option-list"><replaceable>queueinfoitem</replaceable> ...</arg></cmdsynopsis>
+          </term>
           <listitem>
             <para>
               Returns queue details. Queue details of the <command>/</command> virtual host
               are returned if the "-p" flag is absent. The "-p" flag can be used to
               override this default.
             </para>
             <para>
+              Displayed queues can be filtered by their status or
```
dcorbacho: `offline`/`online`/`local` are not strictly `queueinfoitems`, but are listed as such in the help. What if we add those details just below [usage](https://github.com/rabbitmq/rabbitmq-server/pull/911/files#diff-b96d4e7ccd3d984656b8ab3534e6b460R1230) ?

```
  list_queues [-p <vhost>] [--offline|--online|--local] [<queueinfoitem> ...]
...
<queueinfoitem> must be a member of the list [[--offline], [--online],
[--local], name, durable, auto_delete, arguments, policy, pid, owner_pid,
exclusive, exclusive_consumer_pid, exclusive_consumer_tag, messages_ready,
```

binarin: Moved this block of documentation inside the `<para>`, so the dreaded `usage.xsl` will skip them when generating this list.

test/unit_SUITE.erl
binarin: Fixed.

lemenkov:
Cool! Thanks, @binarin for that, and thanks @dcorbacho for merging it finally!

michaelklishin:
Not clear whether #851 should be closed now or more things are coming => assigning milestone to this PR.

PR #913, opened 2016-08-09
Add Fedora/RHEL/CentOS systemd service integration

harlowja:
Add support for systemd usage (vs sysvinit) on versions of fedora
and supporting versions of centos/rhel (newer than or equal to 7).

### DIFF
--- a/packaging/RPMS/Fedora/rabbitmq-server.service
+++ b/packaging/RPMS/Fedora/rabbitmq-server.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=RabbitMQ broker
+After=syslog.target network.target
+
+[Service]
+Type=notify
+User=rabbitmq
+Group=rabbitmq
+WorkingDirectory=/var/lib/rabbitmq
+ExecStart=/usr/sbin/rabbitmq-server
+ExecStop=/usr/sbin/rabbitmqctl stop
+NotifyAccess=all
+TimeoutStartSec=3600
+
+[Install]
+WantedBy=multi-user.target
--- a/packaging/RPMS/Fedora/rabbitmq-server.spec
+++ b/packaging/RPMS/Fedora/rabbitmq-server.spec
@@ -1,4 +1,5 @@
 %define debug_package %{nil}
+%define erlang_minver R16B-03
 
 Name: rabbitmq-server
 Version: %%VERSION%%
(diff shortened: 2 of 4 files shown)


packaging/RPMS/Fedora/rabbitmq-server.spec
```diff
@@ -105,20 +137,48 @@ if [ $1 = 0 ]; then
   # We do not remove /var/log and /var/lib directories
   # Leave rabbitmq user and group
 fi
+%endif
```
dcorbacho: This `endif` doesn't match any `if`. Could you please remove it?

dcorbacho:
@harlowja Thanks for your PR. I verified it in Fedora 24 and CentOS 7.

Once the stale `if` is removed, it is ready to merge.

packaging/RPMS/Fedora/rabbitmq-server.spec
harlowja: Ah, sure will fix that :)

Wonder how that snuck in :-P

harlowja: Yup, will do

Gsantomaggio:
Currently the `rabbitmq-server-xxx.noarch.rpm` works under `Centos 6` and `Centos 7`. 
After merged this PR, will it work on `Centos 6` ?  Since `systemd`  is only for `Centos 7`. 

I made a quick test:
```
cat /etc/redhat-release
CentOS release 6.8 (Final)

sudo rpm -i rabbitmq-server-3.6.5.777-1.el7.centos.noarch.rpm
error: Failed dependencies:
	systemd is needed by rabbitmq-server-3.6.5.777-1.el7.centos.noarch
```

Am I missing something ?
Thanks

harlowja:
Hmmm, I wouldn't try installing a el7 package on el6,

The %if and such in the package should make it still work on older versions, bu I wouldn't recommend installing a package for a newer distro on an older one.

michaelklishin:
Yes, CentOS 6 compatibility is a must for us, so we should be careful with any potentially breaking changes.

Gsantomaggio:
@harlowja you are right, I created the package using Centos 6 and works correctly.

At this point, should we have two `RPM` version ? 

`rabbitmq-server-3.6.5.888-1.el6.noarch.rpm`

and

`sudo rpm -i rabbitmq-server-3.6.5.888-1.el7.noarch.rpm`

?

michaelklishin:
@Gsantomaggio hmm, we can consider that but we can consider it. @dumbbell @dcorbacho WDYT?

harlowja:
Seems like it yes, u'll need a el7 and a el6 one; though I am pretty sure this is common practice.

jerrykuch:
Oh hey, Joshua Harlow, nice to see you... we met when I visited Yahoo for RabbitMQ-related stuff a couple of times back when you were working on Annie's team there...

harlowja:
U turned into a cat! Hi there :-P

harlowja:
My first rabbitmq commit ftw :-P

jerrykuch:
Welcome, etc.!

michaelklishin:
I suggest that re-publish @harlowja's branch into this repo, close this PR and split our RPM package into two for `3.6.6`. `make package-rpm` then will produce both versions, for CentOS 6 and 7.

Thoughts?

harlowja:
A `make package-rpm` to build both is going to require a little bit more work IMHO, it will likely require something like `mock` but this starts to push the boundaries of my understanding of rpmbuild.

https://fedoraproject.org/wiki/Mock?rd=Subprojects/Mock

harlowja:
I can try to work on said thing as a followup if we want (it shouldn't be impossible). Thoughts?

michaelklishin:
@harlowja it's in the top 3 things for someone on our team, we hope to get to it next week. Our build and release system changes are probably quite difficult for outside contributors to make sense of. Thank you for the offer, though!

harlowja:
Cools.

packaging/RPMS/Fedora/rabbitmq-server.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=RabbitMQ broker
+After=syslog.target network.target
+
+[Service]
```
bogdando: Perhaps we should add EnvironmentFile=/etc/rabbitmq/rabbitmq-env.conf ?

binarin: No, rabbitmq-env.conf is interpreted by rabbitmq scripts themselves, and variables here should be given without `RABBITMQ_` prefix

packaging/RPMS/Fedora/rabbitmq-server.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
```
binarin: Shouldn't the epmd dependency be specified here using `Requires` (or at least `Wants`)?

kjnilsson:
I've republished this branch into [rabbitmq-server-932](https://github.com/rabbitmq/rabbitmq-server/tree/rabbitmq-server-932) to contain the additional work required so will close this PR.

harlowja:
Cool thx for letting me know, I hope rabbitmq-server-932 gets merged soon.

lemenkov:
Please before merging address the issue mentioned in my comment:

rabbitmq/rabbitmq-server@c49e2949451178175987702cfb59b515113940f0#commitcomment-18758501

PR #916, opened 2016-08-10
Use new rabbitmqctl features for monitoring

binarin:
@dmitrymex @bogdando WDYT? I haven't tested it yet, but if you are OK with overall shape of this patch, I'll start polishing and testing it.

To stop wasting network bandwidth during health checks (e.g. list_queues
in 3-node cluster with 10k queues costs on average 12 megabytes of
traffic and 27k TCP packets).

Features are disabled by default to preserve compatibility, but they
SHOULD be enabled when following patches are present in currently used
rabbitmq version:
- https://github.com/rabbitmq/rabbitmq-server/pull/915
- https://github.com/rabbitmq/rabbitmq-server/pull/911
- https://github.com/rabbitmq/rabbitmq-server/pull/883

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -47,6 +47,8 @@ OCF_RESKEY_use_fqdn_default=false
 OCF_RESKEY_fqdn_prefix_default=""
 OCF_RESKEY_max_rabbitmqctl_timeouts_default=3
 OCF_RESKEY_policy_file_default="/usr/local/sbin/set_rabbitmq_policy"
+OCF_RESKEY_rmq_feature_health_check_default=true
+OCF_RESKEY_rmq_feature_local_list_queues_default=true
 
 : ${HA_LOGTAG="lrmd"}
 : ${HA_LOGFACILITY="daemon"}
@@ -68,6 +70,8 @@ OCF_RESKEY_policy_file_default="/usr/local/sbin/set_rabbitmq_policy"
 : ${OCF_RESKEY_fqdn_prefix=${OCF_RESKEY_fqdn_prefix_default}}
 : ${OCF_RESKEY_max_rabbitmqctl_timeouts=${OCF_RESKEY_max_rabbitmqctl_timeouts_default}}
 : ${OCF_RESKEY_policy_file=${OCF_RESKEY_policy_file_default}}
+: ${OCF_RESKEY_rmq_feature_health_check=${OCF_RESKEY_rmq_feature_health_check_default}}
+: ${OCF_RESKEY_rmq_feature_local_list_queues=${OCF_RESKEY_rmq_feature_local_list_queues_default}}
 
 #######################################################################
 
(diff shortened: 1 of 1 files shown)


scripts/rabbitmq-server-ha.ocf
```diff
@@ -1537,7 +1571,45 @@ get_monitor() {
...
+    su_rabbit_cmd "${OCF_RESKEY_ctl} node_health_check -t $timeout"
+    rc=$?
+
+    check_timeouts $rc "rabbit_node_health_check_timeouts" "node_health_check"
+    rc_timeouts=$?
+
+    if [ "$rc_timeouts" -eq 2 ]; then
+        master_score 0
+        return $OCF_ERR_GENERIC
+    elif [ $rc -ne 0 ]; then
```
dmitrymex: Here should be something like (rc_timeouts == 0 AND rc !- 0) because if (rc_timeouts == 1 AND rc == 137), that is a timeout which should be ignored

scripts/rabbitmq-server-ha.ocf
```diff
@@ -1537,7 +1571,45 @@ get_monitor() {
...
+    fi
+}
+
+node_health_check_local() {
+    local LH="${LH} node_health_check_local():"
+    local rc
+    local rc_timeouts
+
+    local timeout=$((TIMEOUT_ARG - 2)) # give node_health_check some time to handle timeout by itself
+    su_rabbit_cmd "${OCF_RESKEY_ctl} node_health_check -t $timeout"
```
dmitrymex: Is there a reason to use timeout built-into rabbitmqctl? I think we can use only /usr/bin/timeout as we do right now just for simplicity.

dmitrymex: And BTW if you agree here, there is no need to check "$op_rc -ne 75" in check_timeouts above

scripts/rabbitmq-server-ha.ocf
```diff
@@ -298,6 +302,26 @@ A path to the shell script to setup RabbitMQ policies
...
+</parameter>
+
+<parameter name="rmq_feature_local_list_queues" unique="0" required="0">
+<longdesc lang="en">
+For rabbit version that implements --local flag for list_queues, this
+can greatly reduce network overhead in cases when node is
+stopped/demoted.
+</longdesc>
+<shortdesc lang="en">Use --local option for list_queues</shortdesc>
+<content type="string" default="${OCF_RESKEY_rmq_feature_local_list_queues_default}" />
```
dmitrymex: Here and above you can use type="boolean", like it is done for 'debug' parameter, for example. Though I am not sure if it will affect anything at all.

dmitrymex:
@binarin: overall patch looks good to me, it looks like it should remove load done by our current monitoring.

scripts/rabbitmq-server-ha.ocf
binarin: Because it's possible to implement better diagnostic messages inside `rabbitmqctl`. It's not implemented yet, but is definitely possible.

michaelklishin:
What's the conclusion on this? Should we merge it?

bogdando:
-1 Please fix bashisms:

```
possible bashism in /tmp/tmp.ixnMM0S650 line 1708 (printf %q|%b):
            q_c=`printf "%b\n" "${queues}" | wc -l`
possible bashism in /tmp/tmp.ixnMM0S650 line 1710 (printf %q|%b):
            mem=`printf "%b\n" "${queues}" | awk -v sum=0 '{sum+=$1} END {print (sum/1048576)}'`
possible bashism in /tmp/tmp.ixnMM0S650 line 1712 (printf %q|%b):
            mes=`printf "%b\n" "${queues}" | awk -v sum=0 '{sum+=$2} END {print sum}'`
possible bashism in /tmp/tmp.ixnMM0S650 line 1714 (printf %q|%b):
            c_u=`printf "%b\n" "${queues}" | awk -v sum=0 -v cnt=${q_c} '{sum+=$3} END {print (sum+1)/(cnt+1)}'`
sh syntax error in scripts/rabbitmq-server-ha.ocf
```

binarin:
@michaelklishin Please don't merge it yet, I'm still testing.

michaelklishin:
This currently doesn't merge cleanly.

binarin:
Rebased and tested again. Now I'm happy with this patch.

michaelklishin:
Thank you!

PR #919, opened 2016-08-15
Update iptables calls with --wait

mwhahaha:
If iptables is currently being called outside of the ocf script, the
iptables call will fail because it cannot get a lock. This change
updates the iptables call to include the -w flag which will wait until
the lock can be established and not just exit with an error.


michaelklishin:
Thank you. If this is intended for the `3.6.x` series, please re-submit (or submit another PR) against the `stable` branch.

@binarin @dmitrymex @bogdando can I have your blessing, please? :)

mwhahaha:
I'll submit another against the stable branch as well. Thanks.

binarin:
LGTM

PR #924, opened 2016-08-18
Fix bashisms in rabbitmq OCF RA

bogdando:
Change "printf %b" to be passing the checkbashisms.


bogdando:
Submitted to stable

PR #929, opened 2016-08-22
[OCF HA] Change master score computation & split-brain detection logic

dmitrymex:
The proposed two commits fix the following 2 bugs opened in Fuel downstream:
    https://bugs.launchpad.net/fuel/+bug/1559136
    https://bugs.launchpad.net/mos/+bug/1561894

For the explanation, please see the commit messages.

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -882,12 +882,21 @@ action_validate() {
     return $OCF_SUCCESS
 }
 
+update_rabbit_start_time_if_rc() {
+    local nowtime
+    local rc=$1
+    if [ $rc -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


dmitrymex:
That is a version of commits which passed through several iterations in downstream Fuel in https://review.openstack.org/#/c/324646/ and https://review.openstack.org/#/c/324647/.

@bogdando, @binarin : guys, what do you think?

bogdando:
We had a long discussion on that topic, I checked the changes against network partitions (by Jepsen's Nemesis) and for all of the cases the cluster was auto recovered. Dmitry also prepared docs update (UML flow changes). So nothing to add from my side, LGTM. Thank you for a hard work, folks!

michaelklishin:
Thank you!

PR #936, opened 2016-08-25
Docs: Default AMQP heartbeat delay should be 60 seconds in rabbitmq.config.example

### LINKED ISSUE #935: Docs: Default AMQP heartbeat delay should be 60 seconds in rabbitmq.config.example
The default AMQP heartbeat delay value in `/docs/rabbitmq.config.example` is `600` seconds. According to the offical [docs](https://www.rabbitmq.com/configure.html#configuration-file), the value should be `60`.

PR coming...

```erlang
 %%
   %% Additional network and protocol related configuration
   %% =====================================================
   %%

   %% Set the default AMQP heartbeat delay (in seconds).
   %%
%% {heartbeat, 600},
```

niclic:
It is `600` but should be `60`.

PR made on branch `stable`.

Fixes #935

### DIFF
--- a/docs/rabbitmq.config.example
+++ b/docs/rabbitmq.config.example
@@ -161,7 +161,7 @@
 
    %% Set the default AMQP heartbeat delay (in seconds).
    %%
-   %% {heartbeat, 600},
+   %% {heartbeat, 60},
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you!

PR #939, opened 2016-08-26
Perform partition checks from OCF HA script

binarin:
Partitioned nodes are ordered to restart by master. It may sound like
    `autoheal`, but the problem is that OCF script and `autoheal` are not
    compatible because concepts of master in pacemaker and winner in
    autoheal are completely unrelated.

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -812,10 +812,45 @@ get_master_name_but()
     done
 }
 
+erl_eval() {
+    local fmt="${1:?}"
+    shift
+
+    ${OCF_RESKEY_ctl} eval "$(printf "$fmt" "$@")"
+}
+
 # Returns 0 if we are clustered with provideded node
 is_clustered_with()
 {
-    get_running_nodes | grep -q $(rabbit_node_name $1);
+    local LH="${LH}: is_clustered_with: "
+    local node_name
+    local rc
+    node_name=$(rabbit_node_name $1)
+
+    local seen_as_running
+    seen_as_running=$(erl_eval "lists:member('%s', rabbit_mnesia:cluster_nodes(running))." "$node_name")
+    rc=$?
+    if [ "$rc" -ne 0 ]; then
+        ocf_log err "${LH} Failed to check whether '$node_name' is considered running by us"
+        # XXX Or should we give remote node benefit of a doubt?
+        return 1
+    elif [ "$seen_as_running" != true ]; then
+        ocf_log info "${LH} Node $node_name is not running, considering it not clustered with us"
+        return 1
+    fi
+
+    local seen_as_partitioned
+    seen_as_partitioned=$(erl_eval "lists:member('%s', rabbit_node_monitor:partitions())." "$node_name")
+    rc=$?
+    if [ "$rc" -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


michaelklishin:
@dmitrymex @bogdando waiting for a go-ahead from you :)

scripts/rabbitmq-server-ha.ocf
```diff
@@ -812,10 +812,45 @@ get_master_name_but()
...
 is_clustered_with()
 {
-    get_running_nodes | grep -q $(rabbit_node_name $1);
+    local LH="${LH}: is_clustered_with: "
+    local node_name
+    local rc
+    node_name=$(rabbit_node_name $1)
+
+    local seen_as_running
+    seen_as_running=$(erl_eval "lists:member('%s', rabbit_mnesia:cluster_nodes(running))." "$node_name")
```
bogdando: Shall we as well use this approach in the check_need_join_to()? Your change seems like if get_running_nodes() is not good :(

bogdando: Or may be we need update get_nodes__base()'s eval "mnesia:system_info(${infotype})." as well?

bogdando:
I'm not sure which problem you solve by replacing  ```eval "mnesia:system_info(running)."``` to ```"lists:member('%s', rabbit_mnesia:cluster_nodes(running))."```. And how to test this.

scripts/rabbitmq-server-ha.ocf
```diff
@@ -812,10 +812,45 @@ get_master_name_but()
...
+    rc=$?
+    if [ "$rc" -ne 0 ]; then
+        ocf_log err "${LH} Failed to check whether '$node_name' is considered running by us"
+        # XXX Or should we give remote node benefit of a doubt?
+        return 1
+    elif [ "$seen_as_running" != true ]; then
```
bogdando: is it != "true" ?

binarin: Yes, that's correct. I'm checking here that given node is missing from the list of running nodes.

binarin:
> I'm not sure which problem you solve by replacing eval "mnesia:system_info(running)." to "lists:member('%s', rabbit_mnesia:cluster_nodes(running)).". And how to test this.

This are actually equivalent. Functions return the same values and can be used interchangeably, and wrapping result in `lists:member/2` allows us to get rid of `grep` call in script itself, we can just compare with `true` or `false` using `if` statement.

scripts/rabbitmq-server-ha.ocf
binarin: By approach do you mean using `lists:member/2` or checking both running and partitioned states? 

I've added `lists:member/2` because I needed to add another check and did it just for unification with new piece of code. It's slightly cleaner IMO, but it's not worth changing it everywhere. And for partition check it's just way more harder to implement it using grep.

binarin: As for `cheek_need_join_to()` it got me thinking. Is the check for `running` state is a correct check here? Or should we check list of known nodes instead?

michaelklishin:
Is this ready to be merged?

bogdando:
LGTM and the standard check I do with jepsen (https://github.com/bogdando/rabbitmq-cluster-ocf-vagrant) works for me

PR #942, opened 2016-09-06
[OCF HA] Do not suggest to run the second monitor action

dmitrymex:
Right now we suggest to users to run the second monitor for slaves
with depth=30. It made sense previously, when there was an additional
check at that depth. Right now we don't have any depth-specific
checks and hence it does not make sense to run the second monitor.
Moreover, removing the second monitor fixes an issue with Pacemaker
not reacting on failing monitor if it takes more than a minute. For
details see Fuel bug https://launchpad.net/bugs/1618843


dmitrymex:
@bogdando: WDYT ^^?

bogdando:
:+1:

PR #946, opened 2016-09-07
[OCF HA] Delete Mnesia schema on mnesia reset

dmitrymex:
Not doing so leads to RabbitMQ node being half-stuck in cluster. As a
result, it can't clearly join back and constantly fails. Details could
be found in the following Fuel bug:
https://bugs.launchpad.net/fuel/+bug/1620649

### DIFF
--- a/scripts/rabbitmq-server-ha.ocf
+++ b/scripts/rabbitmq-server-ha.ocf
@@ -39,6 +39,7 @@ OCF_RESKEY_definitions_dump_file_default="/etc/rabbitmq/definitions"
 OCF_RESKEY_pid_file_default="/var/run/rabbitmq/pid"
 OCF_RESKEY_log_dir_default="/var/log/rabbitmq"
 OCF_RESKEY_mnesia_base_default="/var/lib/rabbitmq/mnesia"
+OCF_RESKEY_mnesia_schema_base_default="/var/lib/rabbitmq"
 OCF_RESKEY_host_ip_default="127.0.0.1"
 OCF_RESKEY_node_port_default=5672
 OCF_RESKEY_erlang_cookie_default=false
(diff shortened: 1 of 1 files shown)


dmitrymex:
@bogdando: could you please review it?

bogdando:
:+1:

bogdando:
I've reproduced this error having the patch in place:
```
Mnesia(rabbit@n2): ** ERROR ** (core dumped to file: "/var/lib/rabbitmq/MnesiaCore.rabbit@n2_1473_324370_784463")
 ** FATAL ** Failed to merge schema: Bad cookie in table definition rabbit_user_permission: rabbit@n2 = {cstruct,rabbit_user_permission,set,[],[rabbit@n2,rabbit@n1],[],[],0,read_write,false,[],[],false,user_permission,[user_vhost,permission],[],[],[],{{1473324142071425597,-576460752303421855,1},rabbit@n2},{{3,2},{rabbit@n2,{1473,324194,33515}}}}, rabbit@n1 = {cstruct,rabbit_user_permission,set,[],[rabbit@n1],[],[],0,read_write,false,[],[],false,user_permission,[user_vhost,permission],[],[],[],{{1473324350438460524,-576460752303423231,1},rabbit@n1},{{2,0},[]}}
```
shall it clean up ```/var/lib/rabbitmq/MnesiaCore``` as well?

scripts/rabbitmq-server-ha.ocf
```diff
@@ -711,7 +721,9 @@ reset_mnesia() {
     if $make_amnesia ; then
         kill_rmq_and_remove_pid
         ocf_run rm -rf "${MNESIA_FILES}"
-        ocf_log warn "${LH} Mnesia files appear corrupted and have been removed from ${MNESIA_FILES}."
+        mnesia_schema_location="${OCF_RESKEY_mnesia_schema_base}/Mnesia.$(rabbit_node_name $(get_hostname))"
```
bogdando: Could you capitalize to MNESIA_SCHEMA_FILES? Please also take into account the ./MnesiaCore path as well

bogdando:
Ok, it should be working with the 3.6.6 only, yet released, or with the MOS downstream 3.6.5, so I ran tests against wrong version. I hope it works for you folks! :+1:

scripts/rabbitmq-server-ha.ocf
dmitrymex: I would prefer to keep mnesia_schema_location lowercase - it is just a local variable and we have them in lowercase everywhere. I'd rather make MNESIA_FILES a local variable later and make it lowercase as well.

Regarding files like MnesiaCore.rabbit@n2_1473_324370_784463, I don't think we should remove them. These are core dumps from Mnesia and they might help us to perform RCA in case Mnesia fails. In my case just removing schema directory helped to restore RabbitMQ cluster.

dmitrymex:
@bogdando: I have performed 10 Jepsen runs as you requested, as well as rebooted the whole cluster a couple times. Each time cluster was restored back successfully, except one time when a tricky partitioning occurred which OCF script failed to fix. The problem is unrelated to the current fix and I will investigate it separately.

michaelklishin:
So, is this good to go?

bogdando:
:+1:

PR #968, opened 2016-09-19
Provide a way to limit max number of queues in a vhost

### LINKED ISSUE #501: Per-vhost limit on the number of queues

michaelklishin:
Fixes #501.


src/rabbit_vhost_limit.erl
```diff
@@ -53,6 +54,55 @@ notify_clear(VHost, <<"vhost-limits">>, <<"limits">>) ->
...
+    end.
+
+
+-spec is_over_queue_limit(rabbit_types:vhost()) -> {true, non_neg_integer()} | false.
+
+is_over_queue_limit(VirtualHost) ->
+    case queue_limit(VirtualHost) of
+        %% no limit configured
+        undefined                                            -> false;
+        %% with limit = 0, no connections are allowed
```
kjnilsson: comment copy and paste ;)

Issue #979, opened 2016-09-29
Provide a way to encrypt credentials in the config file


michaelklishin:
Some users need a way to encrypt credentials in the config file. We'll largely do what [Tomcat does in that area](http://tcserver.docs.pivotal.io/docs-tcserver/topics/manual.html#obfusc). The private key used for decryption will be loaded from a local filesystem.

Note: **we have no illusions about whether that makes the system meaningfully more secure**: it does not. But there are regulations in various countries that require this security theater. It will be an opt-in feature.

michaelklishin:
To clarify why this would not make the system meaningfully more secure: if an attacker has access to the config file, more often than not she has access to the entire file system or can eventually obtain it, and thus access the private key. This may change if the key was loaded differently but key management in distributed systems is [hard and there are no alien technologies](https://www.youtube.com/watch?v=MvPIthr4kXA) to make it straightforward, so that is outside of scope of this issue.

PR #982, opened 2016-09-29
DO NOT MERGE Add optional password based encryption of configuration values

essen:
Opening so we don't forget.

See #979.

So far we got the PBE up and running, it works on binaries or arbitrary Erlang terms, and the next step is to decrypt values at the appropriate place (some hints in the comments commit).

Early review welcome.


essen:
Closing because bad target branch.

Issue #987, opened 2016-10-04
Cluster node format in the new config format is confused


michaelklishin:
The translation schema looks like this:

```
%% Make clustering happen *automatically* at startup - only applied
%% to nodes that have just been reset or started for the first time.
%% See http://www.rabbitmq.com/clustering.html#auto-config for
%% further details.
%%
%% {cluster_nodes, {['rabbit@my.host.com'], disc}},

{mapping, "cluster_nodes.disc.$node", "rabbit.cluster_nodes",
    [{datatype, atom}]}.

{mapping, "cluster_nodes.ram.$node", "rabbit.cluster_nodes",
    [{datatype, atom}]}.

{translation, "rabbit.cluster_nodes",
fun(Conf) ->
    DiskNodes = [ V || {_, V} <- cuttlefish_variable:filter_by_prefix("cluster_nodes.disc", Conf)],
    RamNodes  = [ V || {_, V} <- cuttlefish_variable:filter_by_prefix("cluster_nodes.ram", Conf)],

    case {DiskNodes, RamNodes} of
        {_,  []} -> {DiskNodes, disc};
        {[],  _} -> {RamNodes, ram}
    end
end}.
```

so it assumes that we specify a set of disc or RAM nodes, while in the classic config the 2nd tuple argument is **our own** node type.

michaelklishin:
Discovered while working on #486.

Issue #487, opened 2015-12-11
More flexible cluster formation


michaelklishin:
See #486 for background.

This issue is about rolling rabbitmq-clusterer's ideas into the core but not forcing it on everybody. [autocluster](https://github.com/aweber/rabbitmq-autocluster/) has its own take on cluster formation and
it should be possible for them to continue doing the same thing. Hence the "pluggable" part.

Unlike #486, where the scope and implementation are quite clear and proven by the autocluster plugin,
this has open-ended questions.

michaelklishin:
After a long in person discussion of the scope of this we have a list of things to focus on. This is therefore an umbrella issue now.

michaelklishin:
This issue has become a placeholder for all open ended questions and references around cluster formation improvement in 3.7.0. I believe we now have all the answers and no further changes are expected. This can be considered done.

PR #989, opened 2016-10-04
Switch config-driven autoclustering to rabbit_peer_discovery_backend

### LINKED ISSUE #486: Pluggable cluster member discovery
Currently we support two ways to discover cluster members:

 * None (manual clustering)
 * `rabbit.cluster_nodes` ("autoclustering")

The excellent [autocluster](https://github.com/aweber/rabbitmq-autocluster/) plugin provides an API and several more implementations (DNS A record-based discovery, Consul, etcd). We need to integrate this behaviour into the core and refactor our own "autoclustering" to use it. A lot of users would be interested in EC2 API-based discovery, à la ElasticSearch AWS Cloud plugin.

Note that this issue is only about discovering nodes. How clusters are formed is a separate concern, and autocluster is different from rabbitmq-clusterer, for example. We need to find a way to make both possible but that belongs to a separate issue: #487.

michaelklishin:
This switches the classic automatic clustering feature for virgin nodes to the newly introduced `rabbit_peer_discovery_backend` behaviour pulled in from https://github.com/aweber/rabbitmq-autocluster.

Fixes #486. Depends on #987. References #487.


michaelklishin:
This depends on #996 (or rather, makes a lot more sense to merge after it) but otherwise seems to be ready for evaluation.

michaelklishin:
@gmr FYI.

src/rabbit_peer_discovery.erl
```diff
@@ -0,0 +1,59 @@
...
+%%
+%% Software distributed under the License is distributed on an "AS IS"
+%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
+%% the License for the specific language governing rights and
+%% limitations under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% The Initial Developer of the Original Code is Pivotal Software, Inc.
+%% Copyright (c) 2007-2015 Pivotal Software, Inc.  All rights reserved.
```
kjnilsson: 2016?

michaelklishin:
Ready for another round.

acogoluegnes:
I tested by listing cluster nodes in the configuration file (the old way) and it works as before. LGTM.

PR #1008, opened 2016-10-25
DNS peer discovery backend

### LINKED ISSUE #988: Provide a DNS cluster peer discovery backend in the core
Follow-up to #486. Since the DNS backend in rabbitmq-autocluster doesn't have any dependencies outside of Erlang/OTP, we can move it into the core.

This is a separate issue to keep #486 as focussed as possible. For example, we'll have to come up with a config syntax for this after #987 is resolved. This should not hold most of the work in #486 back.

michaelklishin:
This ports [the DNS peer discovery backend from rabbitmq-autocluster](https://github.com/aweber/rabbitmq-autocluster/commit/cf71f5759bdabdb67285f94298ac5ab2cf299c67) with a few changes:

 * Node prefix (e.g. `rabbit` in `rabbit@eng.megacorp.com`) is assumed to be the same across all nodes. The node that performs discovery will attach its own prefix to all hostnames discovered via DNS. In `rabbitmq-autocluster` it is configurable.
 * Long vs. short name option is also not configurable and determined from `net_kernel` on the current node (using `net_kernel:longnames/0`, which seems to be undocumented :/)
 * We don't do the [IP address massaging](https://github.com/aweber/rabbitmq-autocluster/blob/master/src/autocluster_util.erl#L210) that `rabbitmq-autocluster` does because I am not sure I understand the intent. We need to clarify that with @gmr :)

While at it, since node type won't vary from backend to backend we should make it backend agnostic.

Documentation for these changes as well as #486 can be found in https://github.com/rabbitmq/rabbitmq-website/pull/298.

Fixes #988.

### DIFF
--- a/src/rabbit_peer_discovery_dns.erl
+++ b/src/rabbit_peer_discovery_dns.erl
@@ -37,9 +37,8 @@ list_nodes() ->
             undefined -> io:format("rabbit.autocluster.peer_discovery_dns~n", []), {[], disc};
             Proplist  ->
                 Hostname = rabbit_data_coercion:to_list(proplists:get_value(hostname, Proplist)),
-                NodeType = proplists:get_value(node_type, Autocluster, disc),
 
-                {discover_nodes(Hostname, net_kernel:longnames()), NodeType}
+                {discover_nodes(Hostname, net_kernel:longnames()), rabbit_peer_discovery:node_type()}
         end
     end.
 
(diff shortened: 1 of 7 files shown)


src/rabbit_peer_discovery_dns.erl
```diff
@@ -0,0 +1,82 @@
...
+
+%%
+%% API
+%%
+
+-spec list_nodes() -> {ok, Nodes :: list()} | {error, Reason :: string()}.
+
+list_nodes() ->
+    case application:get_env(rabbit, autocluster) of
+      undefined         -> io:format("rabbit.autocluster~n", []), {[], disc};
```
dcorbacho: @michaelklishin I guess this `io:format` slipped through from dev, but maybe an info log would be useful? The same below.

michaelklishin:
Ready for another round.

PR #1016, opened 2016-10-28
Support multiple plugins directories

### LINKED ISSUE #1001: Simplify external plugins packaging
Currently there is no easy way to package external plugin to a
.deb-package, because plugins directory is versioned (like
`/usr/lib/rabbitmq/lib/rabbitmq_server-3.6.5.905`). It's too tight
coupling for a plugin.

There should be a way to use some common plugins directory (like
`/usr/lib/rabbitmq/plugins`). I see two possible ways to implement
this:
- Use dpkg triggers to make rabbitmq-server package populate its
  plugins store with symlinks to plugins from common
  directory. Plugin .deb's will notify rabbitmq-server package, so it
  can refresh symlinks there.
- Transform `RABBITMQ_PLUGINS_DIR` into colon-separed list of paths.

First option is less intrusive, but only debian packagers will benefit
from it. Second one is more generic, but will require some relatively
clever way to properly populate `RABBITMQ_PLUGINS_DIR`.

WDYT? What option would you prefer for me to implement?

binarin:
And provide some packaging-friendly defaults on Unix.

Closes #1001.

### DIFF
--- a/scripts/rabbitmq-defaults
+++ b/scripts/rabbitmq-defaults
@@ -41,4 +41,15 @@ ENABLED_PLUGINS_FILE=${SYS_PREFIX}/etc/rabbitmq/enabled_plugins
 
 PLUGINS_DIR="${RABBITMQ_HOME}/plugins"
 
+# RABBIT_HOME can contain a version number, so default plugins
+# directory can be hard to find if we want to package some plugin
+# separately. When RABBITMQ_HOME points to a standard location where
+# it's usally being installed by package managers, we add
+# "/usr/lib/rabbitmq/plugins" to plugin search path.
+case "$RABBITMQ_HOME" in
+    /usr/lib/rabbitmq/*)
+        PLUGINS_DIR="/usr/lib/rabbitmq/plugins:$PLUGINS_DIR"
+        ;;
+esac
+
 CONF_ENV_FILE=${SYS_PREFIX}/etc/rabbitmq/rabbitmq-env.conf
--- a/src/rabbit_plugins.erl
+++ b/src/rabbit_plugins.erl
@@ -84,42 +84,16 @@ active() ->
             lists:member(App, InstalledPlugins)].
 
 %% @doc Get the list of plugins which are ready to be enabled.
-list(PluginsDir) ->
-    list(PluginsDir, false).
+list(PluginsPath) ->
+    list(PluginsPath, false).
 
-list(PluginsDir, IncludeRequiredDeps) ->
-    EZs = [{ez, EZ} || EZ <- filelib:wildcard("*.ez", PluginsDir)],
-    FreeApps = [{app, App} ||
-                   App <- filelib:wildcard("*/ebin/*.app", PluginsDir)],
-    %% We load the "rabbit" application to be sure we can get the
-    %% "applications" key. This is required for rabbitmq-plugins for
-    %% instance.
-    application:load(rabbit),
-    {ok, RabbitDeps} = application:get_key(rabbit, applications),
-    {AvailablePlugins, Problems} =
-        lists:foldl(fun ({error, EZ, Reason}, {Plugins1, Problems1}) ->
-                            {Plugins1, [{EZ, Reason} | Problems1]};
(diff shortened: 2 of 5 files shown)


scripts/rabbitmq-defaults
```diff
@@ -41,4 +41,20 @@ ENABLED_PLUGINS_FILE=${SYS_PREFIX}/etc/rabbitmq/enabled_plugins
 
 PLUGINS_DIR="${RABBITMQ_HOME}/plugins"
 
+# Search for another `plugins` directory towards the filesystem
+# root. This is useful when we want to have some common place to put
+# pre-packaged plugins. And $RABBITMQ_HOME usually contains RabbitMQ
+# version somewhere in the middle, making plugin packaging very
+# inconvinient.
+plugins_test_path="$(dirname $(dirname "$PLUGINS_DIR"))"
+while test "$plugins_test_path" != "/"; do
```
hairyhum: If there are multiple rabbitmq versions installed, plugins from both will be added to this variable. 
Should there be just predefined "common plugins directory" e.g /usr/lib/rabbitmq/plugins instead?

binarin: No, plugins from other versions will be ignored, because we are moving towards fs root and `find` is called with `-maxdepth 1`. - there is now way we can reach sibling directories that way. Using predefined path is simplier, but it'll fail for other standard fs layouts, e.g.:

    /usr/local/lib/rabbitmq/lib/rabbitmq-3.6.6/plugins/
    /usr/local/lib/rabbitmq/plugins/

hairyhum: Ok, I've missed that. But I still don't think the script should look in directories outside of possible install prefix and predefined paths.

michaelklishin: I share the above sentiment.

src/rabbit_plugins.erl
```diff
@@ -84,34 +84,14 @@ active() ->
...
-                    end, {[], []},
-                    [plugin_info(PluginsDir, Plug) || Plug <- EZs ++ FreeApps]),
+list(PluginsPath) ->
+    list(PluginsPath, false).
+
+list(PluginsPath, IncludeRequiredDeps) ->
+    PluginsDirs = split_path(PluginsPath),
+    EZs = list_ezs(PluginsDirs),
+    FreeApps = list_free_apps(PluginsDirs),
+    {AvailablePlugins, Problems} = collect_plugins_info(EZs ++ FreeApps, IncludeRequiredDeps),
```
hairyhum: If there are several plugins with different versions in the plugins path, smallest version will be picked (because of file sorting). It's inconsistent with standard erlang code loading behaviour and is not what user would expect when installing newer version of plugin.

scripts/rabbitmq-defaults
binarin: What about check like this - use `/usr/lib/rabbitmq/plugins`  additionally but only if we are installed below `/usr/lib/rabbitmq/lib`?

hairyhum:
@binarin any update on the comments above?

binarin:
Yes, I'm working on this.

binarin:
Done

michaelklishin:
@binarin thank you. @hairyhum is away from GitHub for a few days, we'll review once he's back.

hairyhum:
When I add a new plugin version to one of the plugins directories, this version is displayed by `rabbitmq-plugins list` for the enabled plugin, while the actual application version `application:get_key(<plugin>, vsn).` can be lower. This can be confusing.
For example:
There is the `rabbitmq_management` plugin with version `3.6.6`enabled.
I add the  `rabbitmq_management-3.6.7.ez` plugin archive to the common plugins directory.
`rabbitmq-plugins list` displays
```
[E*] rabbitmq_management               3.6.7
```
While the application version is `3.6.6`

binarin:
What is the preferred output in such case? 2 different lines for single plugin? Or something like:

    [E*] rabbitmq_management                  3.6.6 (3.6.7 after restart)

michaelklishin:
```
[E*] rabbitmq_management                  3.6.6 (pending upgrade to 3.6.7)
```

PR #1021, opened 2016-11-03
New RabbitMQ CLI integration

hairyhum:
Add `rabbitmq_cli` as a dependency.
Copy escript from rabbitmq_cli to `/escript` dir on make
Run escript from `rabbitmqctl` and `rabbitmq-plugins` shell and CMD scripts.
Related to https://github.com/rabbitmq/rabbitmq-cli/pull/109


michaelklishin:
Hooray, several months worth of work are finally integrated! 🎆

PR #1025, opened 2016-11-08
Switch JSON library to JSX

essen:
Part of https://github.com/rabbitmq/rabbitmq-common/issues/46


src/rabbit_policy.erl
```diff
@@ -212,11 +212,11 @@ parse_set(Type, VHost, Name, Pattern, Definition, Priority, ApplyTo) ->
     end.
 
 parse_set0(Type, VHost, Name, Pattern, Defn, Priority, ApplyTo) ->
-    case rabbit_json:try_decode(Defn) of
+    case rabbit_json:try_decode(list_to_binary(Defn)) of
```
michaelklishin: In the new CLI all Elixir strings are binaries, so I don't think we can get a list here.

PR #1032, opened 2016-11-18
Move rabbit_pbe module to rabbitmq-common

acogoluegnes:
So new CLI can use it.

References rabbitmq/rabbitmq-cli#118


michaelklishin:
This will have to wait until after `3.6.6` is tagged :(

PR #1039, opened 2016-11-28
Make it possible to disable background GC

### LINKED ISSUE #1026: Make background GC runner more configurable
See [this mailing list patch](https://groups.google.com/forum/#!topic/rabbitmq-users/gdqLpdWbop0) and its effect observations in a production system.

We should

 * Make the "ideal interval" configurable
 * Make disabling background GC runner possible

michaelklishin:
and make the interval configurable via config file.

Kudos to Tim Stewart and his team for contributing
a patch that was included with minor modifications:
https://groups.google.com/forum/#!topic/rabbitmq-users/gdqLpdWbop0.

Note that in master we should also update new config format schema to support these new
values. I'll file a separate issue for that.

Fixes #1026.

### DIFF
--- a/docs/rabbitmq.config.example
+++ b/docs/rabbitmq.config.example
@@ -312,7 +312,15 @@
    %% Size in bytes below which to embed messages in the queue index. See
    %% http://www.rabbitmq.com/persistence-conf.html
    %%
-   %% {queue_index_embed_msgs_below, 4096}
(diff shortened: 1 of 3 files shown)


michaelklishin:
Updated the description to give credit to Tim Stewart, per Kyle's request.

PR #1044, opened 2016-11-29
Properly handle systemd notify errors

binarin:
Original intention was to call `open_port` protected by `catch`, but I've forgot about it. So instead of nice error message it resulted in a stack-trace like https://groups.google.com/forum/#!topic/rabbitmq-users/JsBTioFtZwI


michaelklishin:
Thank you!

PR #1045, opened 2016-11-30
Add function returning active plugins and versions

acogoluegnes:
For rabbitmq/rabbitmq-cli#131

### DIFF
--- a/src/rabbit_plugins.erl
+++ b/src/rabbit_plugins.erl
@@ -215,7 +215,8 @@ is_loadable(App) ->
 %% List running plugins along with their version.
 -spec running_plugins() -> [{atom(), Vsn :: string()}].
 running_plugins() ->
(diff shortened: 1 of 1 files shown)


src/rabbit_plugins.erl
```diff
@@ -211,6 +211,12 @@ is_loadable(App) ->
         _                            -> false
     end.
 
+
+%% List running plugins along with their version.
+-spec running_plugins() -> [{atom(), Vsn :: string()}].
+running_plugins() ->
+    {ok, [{App, Vsn} || {App, _ , Vsn} <- rabbit_misc:which_applications(), lists:member(App, active())]}.
```
michaelklishin: This is hardly a serious concern for this function given that it will be used only by CLI tools but I'd cache the return value of `active/0` instead of calling it N times.

PR #1060, opened 2016-12-16
Bump the default ets tables limit

### LINKED ISSUE #1059: Bump max ETS tables limit
RabbitMQ heavily uses ETS as a storage for many components and processes. Default erlang ETS tables limit is just 1400, which can be not enough. The limit should be increased.

hairyhum:
Set `ERL_MAX_ETS_TABLES` to 50000.
Fixes #1059

### DIFF
--- a/scripts/rabbitmq-server
+++ b/scripts/rabbitmq-server
@@ -123,6 +123,11 @@ else
     RABBIT_SASL_ERROR_LOGGER='{file,"'${RABBITMQ_SASL_LOGS}'"}'
 fi
 
+# Bump ETS table limit to 50000
+if ["x" = "x$ERL_MAX_ETS_TABLES"]; then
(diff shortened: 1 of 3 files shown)


scripts/rabbitmq-server
```diff
@@ -123,6 +123,11 @@ else
     RABBIT_SASL_ERROR_LOGGER='{file,"'${RABBITMQ_SASL_LOGS}'"}'
 fi
 
+# Bump ETS table limit to 50000
+if ["x" = "x$ERL_MAX_ETS_TABLES"]; then
```
michaelklishin: This should be `if [ "x" = "x$ERL_MAX_ETS_TABLES" ]` (note the spaces), otherwise shell will attempt to execute a command. Fixed in `stable`.

PR #1068, opened 2016-12-25
Fixed typos in code comments

AndrewDryga:
There was number of typos in docs and comments, this PR fixes some of them.

PR #1069, opened 2016-12-25
Fixed various typos across code comments

AndrewDryga:
Retry #1068 with taking into account GB accent.

### DIFF
--- a/src/rabbit_alarm.erl
+++ b/src/rabbit_alarm.erl
@@ -16,7 +16,7 @@
 %% There are two types of alarms handled by this module:
 %%
 %% * per-node resource (disk, memory) alarms for the whole cluster. If any node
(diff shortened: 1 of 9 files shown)


src/rabbit_alarm.erl
```diff
@@ -16,7 +16,7 @@
 %% There are two types of alarms handled by this module:
 %%
 %% * per-node resource (disk, memory) alarms for the whole cluster. If any node
-%%   has an alarm, then all publishing should be disabled througout the
+%%   has an alarm, then all publishing should be disabled throughput the
```
michaelklishin: this is mean to say "throughout", not "throughput."

michaelklishin:
There is still one case where your spell checker didn't recognize a valid word. I will correct it manually.

Issue #964, opened 2016-09-15
High memory consumption with lazy queues in flow control


Gsantomaggio:
When the lazy queue gets in `flow` status, the msg_store seems not fast enough to save the messages, thus the messages are stored in memory.

 `reduce_memory_use` function should be the point to start.

```
reduce_memory_use(State = #vqstate {
                             mode = lazy,
                             ram_pending_ack  = RPA,
                             ram_msg_count    = RamMsgCount,
                             target_ram_count = TargetRamCount }) ->
    State1 = #vqstate { q3 = Q3 } =
        case chunk_size(RamMsgCount + gb_trees:size(RPA), TargetRamCount) of
            0  -> State;
            S1 -> {_, State2} =  limit_ram_acks(S1, State),
                  State2
        end,

    State3 =
        case chunk_size(?QUEUE:len(Q3),
                        permitted_beta_count(State1)) of
            0  ->
                State1;
            S2 -> push_betas_to_deltas(S2, State1) 
        end,
    garbage_collect(),
    State3.
```

diaolanshan:
We should wait for the next release for this or add a parameter can solve the issue?

Gsantomaggio:
@diaolanshan The new parameter `lazy_queue_explicit_gc_run_operation_threshold` will not solve this issue, but part of it.
Please see https://github.com/rabbitmq/rabbitmq-server/pull/978.

Gsantomaggio:
we looked in deep into the issue and just to clarify: 

The memory **does not grow up** always when the lazy queues get in flow status, but only under certain conditions. 

Here is two tests that explain in detail the problem:

**Test1 - 5 Connections - 10 Channels -  Body size 51200 bytes**
```
➜  scripts git:(rabbitmq-server-973) ✗ ./rabbitmqctl list_queues message_bytes message_bytes_ready message_bytes_ram name state policy
Listing queues ...
9145088000  9145088000  0 lazy.t_2  flow  Lazy
9145088000  9145088000  0 lazy.t_0  flow  Lazy
9145088000  9145088000  0 lazy.t_1  flow  Lazy
```
```
➜  scripts git:(rabbitmq-server-973) ✗ ./rabbitmqctl list_connections name state | grep -v running
Listing connections ...
127.0.0.1:61419 -> 127.0.0.1:5672 flow
127.0.0.1:61420 -> 127.0.0.1:5672 flow
127.0.0.1:61422 -> 127.0.0.1:5672 flow
[...]

PR #1072, opened 2016-12-30
Configure memory reduction for all queue types

dcorbacho:
Might alleviate the problem in #964, as memory reduction is very expensive in CPU usage. If the garbage collection (VM and data) happens less often, the throughput increases and is able to process all queued requests a bit faster.

Could not reproduce the memory increase during the queue's flow control (as connections and channels get into flow first), so it's not fully verified.

### DIFF
--- a/src/rabbit_variable_queue.erl
+++ b/src/rabbit_variable_queue.erl
@@ -434,14 +434,20 @@
 
 
 %% we define the garbage collector threshold
-%% it needs to tune the GC calls inside `reduce_memory_use`
-%% see: rabbitmq-server-973 and `maybe_execute_gc` function
--define(DEFAULT_EXPLICIT_GC_RUN_OP_THRESHOLD, 250).
--define(EXPLICIT_GC_RUN_OP_THRESHOLD,
+%% it needs to tune the `reduce_memory_use` calls. Thus, the garbage collection.
+%% see: rabbitmq-server-973 and rabbitmq-server-964
+-define(DEFAULT_EXPLICIT_GC_RUN_OP_THRESHOLD, 1000).
+-define(EXPLICIT_GC_RUN_OP_THRESHOLD(Mode),
     case get(explicit_gc_run_operation_threshold) of
         undefined ->
-            Val = rabbit_misc:get_env(rabbit, lazy_queue_explicit_gc_run_operation_threshold,
-                ?DEFAULT_EXPLICIT_GC_RUN_OP_THRESHOLD),
+            Val = case Mode of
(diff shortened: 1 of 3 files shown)


src/rabbit_variable_queue.erl
```diff
@@ -434,14 +434,20 @@
...
-            Val = rabbit_misc:get_env(rabbit, lazy_queue_explicit_gc_run_operation_threshold,
-                ?DEFAULT_EXPLICIT_GC_RUN_OP_THRESHOLD),
+            Val = case Mode of
+                      lazy -> rabbit_misc:get_env(rabbit,
+                                                  lazy_queue_explicit_gc_run_operation_threshold,
+                                                  ?DEFAULT_EXPLICIT_GC_RUN_OP_THRESHOLD);
+                      _ -> rabbit_misc:get_env(rabbit,
+                                               queue_explicit_gc_run_operation_threshold,
+                                               ?DEFAULT_EXPLICIT_GC_RUN_OP_THRESHOLD)
+                  end,
```
Ayanda-D: @dcorbacho minor thoughts here; how about a single function which takes desired env variable and calls `rabbit_misc:get_env/3`, to avoid what appears like code replication in both case clauses. Also, could be worth adding documentation for `queue_explicit_gc_run_operation_threshold` as part of this PR (unless `queue_explicit_gc_run_operation_threshold` configuration audience is not intended for general rabbit users).

michaelklishin: I agree with @Ayanda-D (thanks for the review, by the way).

michaelklishin:
This branch halves `unit_inbroker` execution time. This is very encouraging.

michaelklishin:
```
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
rabbit_ct_broker_helpers:rpc failed on line 897
Reason: {error,{badmatch,true},[{...}|...]}
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Testing deps.rabbit.unit_inbroker_SUITE: *** FAILED test case 196 ***

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
rabbit_ct_broker_helpers:rpc failed on line 897
Reason: {error,{badmatch,true},[{...}|...]}
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Testing deps.rabbit.unit_inbroker_SUITE: *** FAILED test case 198 ***
```

PR #1075, opened 2017-01-06
Re-introduce rabbit_table:wait_for_replicated/0

michaelklishin:
for backwards compatibility with plugins.

Fixes rabbitmq/rabbitmq-clusterer#19.


gerhard:
Closes rabbitmq/rabbitmq-clusterer#19

PR #1084, opened 2017-01-17
Modify new scripts to use /bin/sh

gerhard:
While at it, make them POSIX.1-2008 compliant (a.k.a fix all shellchecks)

Prefer parameter expansion and manipulation over command invocation:

    # OK-ish
    `dirname $0`

    # SPOT ON
    "${0%/*}"

Guard against empty ESCRIPT_DIR var


gerhard:
Not merging into stable since it's new rabbitmqctl related.

PR #1091, opened 2017-01-26
Ignore pre-release version part in plugin version requirement validation

### LINKED ISSUE #1090: Plugin version requirements failing to validate alpha versions
While testing a pre-release version, rabbitmq-clusterer `3.6.5` fails to install on RabbitMQ `3.7.0-alpha.62` with the following error: 
``
rabbitmq_clusterer: Plugin doesn't support current server version. Actual broker version: "3.7.0-alpha.62", supported by the plugin: ["3.6.0", "3.7.0"]
```

Pre-release versions should be matched as release versions (3.7.0-alpha.62 as 3.7.0)

### PR DESCRIPTION
**Pull request by hairyhum**
Fixes #1090 

Pre-release version is considered to be less then its basic version by semver semantics (e.g. `3.7.0-alpha.68 < 3.7.0`)
But plugins should be valid for pre-release versions if version requirements contain release version.

Alpha version part is removed during plugin version check to avoid this inconsistency.

### DIFF
--- a/src/rabbit_plugins.erl
+++ b/src/rabbit_plugins.erl
@@ -340,7 +340,10 @@ check_plugins_versions(PluginName, AllPlugins, RequiredVersions) ->
 is_version_supported("", _)        -> true;
 is_version_supported("0.0.0", _)   -> true;
 is_version_supported(_Version, []) -> true;
-is_version_supported(Version, ExpectedVersions) ->
+is_version_supported(VersionFull, ExpectedVersions) ->
+    %% Pre-release version should be supported in plugins,
+    %% therefore alpha part should be removed
+    Version = remove_version_alpha_part(VersionFull),
     case lists:any(fun(ExpectedVersion) ->
                        rabbit_misc:version_minor_equivalent(ExpectedVersion, Version)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/rabbit_plugins.erl [hunk 1]
```diff
@@ -340,7 +340,10 @@ check_plugins_versions(PluginName, AllPlugins, RequiredVersions) ->
 is_version_supported("", _)        -> true;
 is_version_supported("0.0.0", _)   -> true;
 is_version_supported(_Version, []) -> true;
-is_version_supported(Version, ExpectedVersions) ->
+is_version_supported(VersionFull, ExpectedVersions) ->
+    %% Pre-release version should be supported in plugins,
+    %% therefore alpha part should be removed
+    Version = remove_version_alpha_part(VersionFull),
```
michaelklishin: Let's use "preview" instead of "alpha."

hairyhum: It's called `alpha_info` in `rabbit_semver`

michaelklishin: A chance to fix it there. Not all preview versions are alphas.

PR #1096, opened 2017-02-02
Support error messages returned from `check_X_access` in authz backend

hairyhum:
Check functions now can return `{error_message, string()}` to
report a problem to clients in protocol error.

Part of https://github.com/rabbitmq/rabbitmq-auth-backend-uaa/pull/5
https://github.com/rabbitmq/rabbitmq-auth-backend-uaa/issues/3

### DIFF
--- a/src/rabbit_access_control.erl
+++ b/src/rabbit_access_control.erl
@@ -178,21 +178,18 @@ check_access(Fun, Module, ErrStr, ErrArgs) ->
     check_access(Fun, Module, ErrStr, ErrArgs, access_refused).
 
 check_access(Fun, Module, ErrStr, ErrArgs, ErrName) ->
-    Allow = case Fun() of
-                {error, E}  ->
-                    rabbit_log:error(ErrStr ++ " by ~s: ~p~n",
-                                     ErrArgs ++ [Module, E]),
-                    {false, ""};
-                {error_message, Message} ->
-                    rabbit_log:error(ErrStr ++ " by ~s: ~p~n",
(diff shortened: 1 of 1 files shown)


src/rabbit_access_control.erl
```diff
@@ -178,17 +178,18 @@ check_access(Fun, Module, ErrStr, ErrArgs) ->
...
         true ->
             ok;
         false ->
-            rabbit_misc:protocol_error(ErrName, ErrStr, ErrArgs)
+            rabbit_misc:protocol_error(ErrName, ErrStr, ErrArgs);
+        {error, E}  ->
+            rabbit_log:error(ErrStr ++ " by ~s: ~p~n",
+                             ErrArgs ++ [Module, E]),
+            rabbit_misc:protocol_error(ErrName, ErrStr, ErrArgs);
+        {error_message, Message} ->
```
kjnilsson: woudn't returning `{error, Message}` pretty much do the same thing? Do we need a special `error_message`?

hairyhum: `{error, Message}` can return an arbitrary error reason and will only log it. `{error_message, Message}` will also add it to the protocol error sent to a client.

michaelklishin: I'm somewhat on the fence about this. Can we at least rename it to `client_reported_error` for clarify?

michaelklishin:
We agreed to not do this and propagate all errors to clients. Closing per discussion with @hairyhum @kjnilsson.

PR #1098, opened 2017-02-05
Disable background GC by default; bump credit flow defaults

michaelklishin:
Quite a bit of evidence from different kinds of workloads suggest
that background GC does not contribute meaningfully to reducing
node RAM usage on most of them. It does, however, on each run
produce a massive spike in VM and CPU context switches,
causing latency spikes.

Some users report that running without background GC cuts
their latency to a half or even down to a third.

Multi-hour stress tests that max out all node CPU cores
suggest that our current credit_flow default can still be bumped
safely, reducing the amount of time publishers are throttled
and making publisher latency and node's observed message throughput
more predictable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ define PROJECT_ENV
 	    %% setting has no effect because credit_flow is not used when
 	    %% writing to the queue index. See the setting
 	    %% queue_index_embed_msgs_below above.
-	    {msg_store_credit_disc_bound, {2000, 500}},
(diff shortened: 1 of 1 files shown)


javaforfun:
does credit_flow affect consume speed?  if there are a few consumers and small prefetch_count.

michaelklishin:
Yes but I wouldn't expect any significant throughput 
changes. Since credit flow settings can be configured already, you can give it a try with your workloads and see for yourself.

PR #1101, opened 2017-02-08
Makefile: Use rabbitmq-cli's Makefile `install` target

dumbbell:
Before this change, we used cp(1) and its `-r` flag to copy the source directory recursively. Pretty standard.

We are using hardlinks in the source directory because all escripts are the same code: the program uses the filename to determine the behavior. Pretty standard too. We settled on hardlinks, not symlinks, because on Windows, you need elevated privileges to use symlinks apparently.

Unfortunately, `cp -r` treats those hardlinks as separate files. thus the destination directory contains three distinct copies of the same file:

```
$ ls -l escript/
-rwxr-xr-x  1 dumbbell  dumbbell  4545805 Feb  8 17:25 rabbitmq-diagnostics
-rwxr-xr-x  1 dumbbell  dumbbell  4545805 Feb  8 17:25 rabbitmq-plugins
-rwxr-xr-x  1 dumbbell  dumbbell  4545805 Feb  8 17:25 rabbitmqctl
            ^---- Single references to distinct files

$ du -sh escript/
13M    escript/
```

I first tried with `-a` (cp(1) archive mode) instead of `-r`, but the behavior is inconsistent accross platforms. GNU coreutils' cp(1) preserves hardlinks as I would expect, but BSD cp does not. Indeed `-a`
isn't a standard option. BSD cp's manpage documents the behavior and suggests to use tar(1), cpio(1) or pax(1) when hardlinks are meant to be preserved.

tar(1) and cpio(1) are unavailable by default in MSYS2 (used on Windows), but pax(1) is. So now, we use pax(1) to copy the escript source directory and hardlinks are preserved on all platforms, including
Windows.

```
$ ls -l escript/
-rwxr-xr-x  3 dumbbell  dumbbell  4545805 Feb  8 13:53 rabbitmq-diagnostics
-rwxr-xr-x  3 dumbbell  dumbbell  4545805 Feb  8 13:53 rabbitmq-plugins
-rwxr-xr-x  3 dumbbell  dumbbell  4545805 Feb  8 13:53 rabbitmqctl
            ^---- Three references to a single file

$ du -sh escript/
4.4M   escript/
```

**v2**: The comment above is now obsolete. We use the new `install` target in rabbitmq-cli's `Makefile`. See rabbitmq/rabbitmq-cli#175 for more details.

[#138925175]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,10 +139,6 @@ EXTRA_SOURCES += $(USAGES_ERL)
 .DEFAULT_GOAL = all
 $(PROJECT).d:: $(EXTRA_SOURCES)
 
-copy-escripts:
-	cp -r ${DEPS_DIR}/rabbitmq_cli/escript ./
-
-
 DEP_PLUGINS = rabbit_common/mk/rabbitmq-build.mk \
 	      rabbit_common/mk/rabbitmq-dist.mk \
 	      rabbit_common/mk/rabbitmq-run.mk \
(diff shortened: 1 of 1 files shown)


dumbbell:
With the second commit to that branch, the `Makefile` recipe was moved to rabbitmq-cli. Thus it can be reused by package building scripts.

See rabbitmq/rabbitmq-cli#175 and rabbitmq/rabbitmq-server-release#19.

PR #1105, opened 2017-02-09
Quote -conf_dir/-conf_schema_dir path variables

### LINKED ISSUE #1104: Quote -conf_dir/-conf_schema_dir path variables in server/service.bat
There are a few path variables within rabbitmq-server.bat and rabbitmq-service.bat that aren't quoted, so if there is a space within the path, it would break execution.

Jakauppila:
Fixes #1104


michaelklishin:
Thank you!

Jakauppila:
Do you like to have associated issues for PRs even when they're something small like this? Or is just the PR fine?

michaelklishin:
They are perfectly fine.

michaelklishin:
Err, a PR alone would have been just fine :)

Issue #838, opened 2016-06-10
Disk-based message store index plugin


michaelklishin:
Message store index is currently ETS-based. This has one obvious downside: it can only grow so long. Since the interface is quite small, we should provide an alternative version. The downsides of using e.g. LevelDB are also predictable:
- Lower throughput for messages that go into message store and not embedded into QI
- Shipping native code will require build system and distribution changes, opens a can of worms in terms of support

But for some users a HanoiDB or LevelDB/RocksDB-backed option could be worth using.

Related: #740.

hairyhum:
Addressed by https://github.com/rabbitmq/rabbitmq-msg-store-index-eleveldb

PR #1123, opened 2017-02-23
Rename msg_store_index:delete_by_file to index_cleanup_undefined_file

hairyhum:
Refactoring related to #838 
The `delete_by_file` function is used only once with `undefined` parameter.
For key-value store implementations (e.g. eleveldb) it's much easier to
implement, than maintaining a filename index. ETS index also can be optimised.

### DIFF
--- a/src/rabbit_msg_store.erl
+++ b/src/rabbit_msg_store.erl
@@ -1537,9 +1537,9 @@ index_update_fields(Key, Updates, #msstate { index_module = Index,
 index_delete(Key, #msstate { index_module = Index, index_state = State }) ->
     Index:delete(Key, State).
 
-index_delete_by_file(File, #msstate { index_module = Index,
+index_cleanup_undefined_file(#msstate { index_module = Index,
                                       index_state  = State }) ->
-    Index:delete_by_file(File, State).
+    Index:cleanup_undefined_file(State).
 
 %%----------------------------------------------------------------------------
 %% shutdown and recovery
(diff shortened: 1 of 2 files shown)


src/rabbit_msg_store_ets_index.erl
```diff
@@ -68,8 +68,8 @@ delete_object(Obj, State) ->
     true = ets:delete_object(State #state.table, Obj),
     ok.
 
-delete_by_file(File, State) ->
-    MatchHead = #msg_location { file = File, _ = '_' },
+cleanup_undefined_file(State) ->
```
michaelklishin: I'd rename this to `clean_up_entries_without_file`.

src/rabbit_msg_store.erl
```diff
@@ -1537,9 +1537,9 @@ index_update_fields(Key, Updates, #msstate { index_module = Index,
 index_delete(Key, #msstate { index_module = Index, index_state = State }) ->
     Index:delete(Key, State).
 
-index_delete_by_file(File, #msstate { index_module = Index,
+index_cleanup_undefined_file(#msstate { index_module = Index,
```
michaelklishin: I'd rename this to `index_clean_up_entries_without_file`.

src/rabbit_msg_store_ets_index.erl
hairyhum: How about `clean_up_entries_with_undefined_file`, to clarify that the `file` field should be `undefined`?

michaelklishin: I think that's an implementation detail that can change over time. "without_file" is entirely generic.

hairyhum: The function itself is an implementation detail. And I would like to point it out. The better name would be `clean_up_temporary_reference_count_entries_without_file`

michaelklishin: Right, then either `clean_up_entries_with_undefined_file` or `clean_up_temporary_reference_count_entries_without_file` is good with me.

luchiano1: 1

src/rabbit_msg_store.erl
luchiano1: 1

luchiano1:
1

Issue #1124, opened 2017-02-24
Exclude applications that are not RabbitMQ plugins from the list of plugins


dcorbacho:
Currently, this is the list of plugins when RabbitMQ starts:

```
  rabbitmq_recent_history_exchange
  rabbitmq_management_agent
  cowlib
  cowboy
  rabbitmq_web_dispatch
  amqp_client
  rabbitmq_management
  rabbitmq_mqtt
  rabbitmq_web_mqtt
  rabbitmq_amqp1_0
  rabbitmq_stomp
  sockjs
  rabbitmq_web_stomp
  rabbitmq_tracing
  rabbitmq_trust_store
  rabbitmq_consistent_hash_exchange
  rabbitmq_web_mqtt_examples
  rabbitmq_top
  rabbitmq_auth_backend_ldap
  rabbitmq_federation
  rabbitmq_shovel
  rabbitmq_shovel_management
  rabbitmq_auth_mechanism_ssl
  rabbitmq_sharding
  rabbitmq_jms_topic_exchange
  rabbitmq_event_exchange
  rabbitmq_federation_management
  rabbitmq_web_stomp_examples
```
sockjs, cowboy, cowlib should not be present in this list. All applications that do not depend on `rabbit` should be removed from this list.

dumbbell:
This new requirement should be communicated clearly in the release notes as well as on the [plugins development documentation](http://www.rabbitmq.com/plugin-development.html). Probably even a mail on rabbitmq-users@.

I believe many third-party plugins do not depend on `rabbit` in their `Makefile` or `$plugin.app` because it was not possible a year ago, probably less (ie. before the release build was fixed to get rid of the broker depending on its plugins). So if third-party plugin developers do not modify their plugin, they won't be listed anymore.

PR #1128, opened 2017-02-27
Exclude apps that are not RabbitMQ plugins from the list of plugins

dcorbacho:
Part of #1124

### DIFF
--- a/src/rabbit_plugins.erl
+++ b/src/rabbit_plugins.erl
@@ -22,6 +22,7 @@
 -export([ensure/1]).
 -export([extract_schemas/1]).
 -export([validate_plugins/1, format_invalid_plugins/1]).
+-export([is_strict_plugin/1, strict_plugins/2, strict_plugins/1]).
 
 % Export for testing purpose.
 -export([is_version_supported/2, validate_plugins/2]).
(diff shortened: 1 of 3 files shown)


src/rabbit_plugins.erl
```diff
@@ -123,6 +127,24 @@ dependencies(Reverse, Sources, AllPlugins) ->
     true = digraph:delete(G),
     Dests.
 
+%% Filter real plugins from application dependencies
+is_strict_plugin(#plugin{extra_dependencies = ExtraDeps}) ->
```
michaelklishin: I'd suggest renaming this to either `is_strictly_plugin/1` or `is_actual_plugin/1`.

src/rabbit_plugins.erl
```diff
@@ -123,6 +127,24 @@ dependencies(Reverse, Sources, AllPlugins) ->
...
+    lists:member(rabbit, ExtraDeps).
+
+strict_plugins(Plugins, AllPlugins) ->
```
michaelklishin: Same as above, `strictly_plugins/2` or `actual_plugins/2`.

PR #1130, opened 2017-03-01
Update metrics gc to support gc marker

kjnilsson:
Part of https://github.com/rabbitmq/rabbitmq-management/issues/353


michaelklishin:
This was incorrectly submitted against master. I will cherry-pick.

PR #1134, opened 2017-03-03
Provide some defaults for embedding

binarin:
Sorry, erroneously submitted to `master` the first time.

This commit provides defaults for some paths (similar to the thing the
mnesia does with its `dir` parameter) that will allow us to start
broker in embedded mode without any additional configuration. The
minimal test-case is to issue the following commands from
`rabbitmq-server` checkout:

    make shell "SHELL_OPTS=-sname embed-friendly@localhost"
    > rabbit:start().


binarin:
Fixed

Issue #1136, opened 2017-03-06
Split rabbitmq-server tests


dcorbacho:
The rabbitmq-server job is the longest job to execute. At the same time, the testing of all the plugins depends on it. The proposal is to split it in two jobs, moving all the long-running tests to the end of pipeline, so it won't delay testing of plugins.

Possible suites to move:
- Split `unit_inbroker_SUITE`
- Move `cluster_rename_SUITE`
- Move `clustering_management_SUITE`
- Move `dynamic_ha_SUITE`
- Move `eager_sync_SUITE`
- Move `health_check_SUITE`
- Move `partitions_SUITE `

PR #1137, opened 2017-03-06
Split test suites into fast and slow Makefile targets

dcorbacho:
Part of #1136


erlang.mk
```diff
@@ -6120,6 +6120,8 @@ CT_OPTS ?=
 ifneq ($(wildcard $(TEST_DIR)),)
 ifndef CT_SUITES
 CT_SUITES := $(sort $(subst _SUITE.erl,,$(notdir $(call core_find,$(TEST_DIR)/,*_SUITE.erl))))
+SLOW_CT_SUITES := $(sort cluster_rename clustering_management dynamic_ha eager_sync health_check partitions priority_queue simple_ha queue_master_location unit_inbroker_backing_queue)
+FAST_CT_SUITES := $(filter-out $(SLOW_CT_SUITES),$(CT_SUITES))
```
dumbbell: You can't modify `erlang.mk` because it's a generated file which will get overwritten with the next update of Erlang.mk (`make erlang-mk`).

Those changes should go the `Makefile`.

test/unit_inbroker_cluster_SUITE.erl
```diff
@@ -0,0 +1,363 @@
...
+%%
+%% Software distributed under the License is distributed on an "AS IS"
+%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See the
+%% License for the specific language governing rights and limitations
+%% under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% The Initial Developer of the Original Code is GoPivotal, Inc.
+%% Copyright (c) 2011-2016 Pivotal Software, Inc.  All rights reserved.
```
dumbbell: The copyright year should be updated to 2017.

test/unit_inbroker_cluster_SUITE.erl
```diff
@@ -0,0 +1,363 @@
...
+%%
+
+-module(unit_inbroker_cluster_SUITE).
```
dumbbell: You can get rid of the `unit_inbroker` prefix for most of them. That name only made "sense" because unrelated testcases were in the same testsuite.

erlang.mk
dumbbell: The implementation of `ct-slow` and `ct-fast` can also go to one of our Erlang.mk plugins so we could define `$(CT_SLOW)` and `$(CT_FAST)` in other projects too.

test/backing_queue_SUITE.erl
```diff
@@ -0,0 +1,1689 @@
...
+      [
+       {backing_queue_embed_limit_0, [], ?BACKING_QUEUE_PARALLEL},
+       {backing_queue_embed_limit_1024, [], ?BACKING_QUEUE_PARALLEL}
+      ]},
+     {backing_queue_tests, [], [
+          msg_store,
+          {backing_queue_embed_limit_0, [], ?BACKING_QUEUE_TESTCASES},
+          {backing_queue_embed_limit_1024, [], ?BACKING_QUEUE_TESTCASES}
+        ]}
+    ].
```
dumbbell: Is there a reason why you splitted that into two groups with the same name? Is it to get a dedicated broker/cluster for each top-level group?

dumbbell: I mean, did having the `variable_queue_*` groups inside `?BACKING_QUEUE_TESTCASES` prevent them from running successfully in parallel for instance?

dcorbacho: Well spotted, not anymore. I split them originally to parallelize more groups, but that didn't work (or became slower). I should have reverted also this part.

PR #1142, opened 2017-03-09
DO NOT MERGE Pluggable message stores experiment

hairyhum:
In order to be able to use different implementation for message store,
the message store module is not called directly from rabbit_variable_queue,
instead it can be configured with application environment and will
be saved in a message store supervisor.

The message store has a bit more complex behaviour when confirming messages to be written on disk synchronously. It still needs to be described in the behaviour module comments.

This is a part of an experiment to replace a message store. 
The replacement module using ElevelDB can be found in [this repo](https://github.com/hairyhum/rabbitmq-msg-store-eleveldb)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ define PROJECT_ENV
 	    {memory_monitor_interval, 2500},
 	    {disk_free_limit, 50000000}, %% 50MB
 	    {msg_store_index_module, rabbit_msg_store_ets_index},
+	    {msg_store_module, rabbit_msg_store},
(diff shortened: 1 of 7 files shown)


hairyhum:
Missing parts:

Migration between message stores.
Tests for pluggability.

hairyhum:
Migrations between message stores are disabled by checking the old store name when starting the new one. If it's different - message store will fail to start.

Need rebase after https://github.com/rabbitmq/rabbitmq-server/pull/1158

michaelklishin:
Closing as this PR is 18 months old and we have a new/different next gen message store plan.

Issue #1146, opened 2017-03-13
Per-vhost supervision tree for queues and VHosts


hairyhum:
Queues and VHosts reside in different supervisors since #567 and when a message store process crashes, queues are still alive and expect it to be running. This can cause issues with publisher confirms and error messages hard do decrypt.
When message store is crashed it should cause all its VHost queues to restart, which can be achieved using supervisors.
A supervisor tree per VHost should run queues and message stores recovery steps on startup, so it can be restarted without crashing a node.
Another solution can be to disable message stores restart and make them crash a node.

PR #1147, opened 2017-03-13
Do not restart per-vhost message stores.

hairyhum:
Minimal solution to #1146
Set max_restarts to 0 for per-vhost message stores.
Crashed message store will terminate a node.


michaelklishin:
We introduced per-vhost message stores to avoid the entire node being hosed should a message store fail for any reason.

hairyhum:
So we need to proceed with per-vhost supervisors then.

PR #1156, opened 2017-03-22
Fix lager initialization code

binarin:
Setting env var to 'undefined' is different from not setting it at
all, and lager is sensitive to this distinction in its 'log_root'
parameter. There will be a crash when rabbit is started without the
help of startup scripts (i.e. from `make shell` or in the embedded
mode).

### DIFF
--- a/src/rabbit_lager.erl
+++ b/src/rabbit_lager.erl
@@ -185,7 +185,7 @@ configure_lager() ->
             %% Setting env var to 'undefined' is different from not
             %% setting it at all, and lager is sensitive to this
             %% difference.
-            case application:get_env(rabbit, lager_log_root, undefined) of
+            case application:get_env(rabbit, lager_log_root) of
                 {ok, FromRabbitConfig} ->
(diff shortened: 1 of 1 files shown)


src/rabbit_lager.erl
```diff
@@ -182,9 +182,15 @@ configure_lager() ->
...
     case application:get_env(lager, log_root) of
         undefined ->
-            application:set_env(lager, log_root,
-                                application:get_env(rabbit, lager_log_root,
-                                                    undefined));
+            %% Setting env var to 'undefined' is different from not
+            %% setting it at all, and lager is sensitive to this
+            %% difference.
+            case application:get_env(rabbit, lager_log_root, undefined) of
+                {ok, FromRabbitConfig} ->
```
hairyhum: This line will never match. `application:get_env/3` returns value itself or a default value.
Did you wanted to use `application:get_env(rabbit, lager_log_root)` instead?

binarin: Yes, my mistake.

PR #1158, opened 2017-03-23
Per-vhost supervisors.

### LINKED ISSUE #1146: Per-vhost supervision tree for queues and VHosts

hairyhum:
Fixes #1146 

Message stores and queues for a specific vhost are moved to per-vhost supervision trees.
`rabbit_vhost_sup_sup` is a `simple_one_for_one` supervisor, containing a supervision tree per vhost.
Recovery mechanism was changed to recover a specific vhost when it's restarting. So message store will be started with actual durable queues and the queues will be recovered the same way as on rabbit start.
Recovery terms are moved to per-vhost directories.
Supports migration from 3.6 and 3.5 was tested.

Because restarting logic is per-vhost, a vhost supervisor cannot restart it's children (even using `one_for_all`), so every vhost supervisor is a child of another "wrapper" supervisor. THe wrapper supervisor defines how many times a vhost supervisor can be restarted (currently 1 in 1000 sec).

When vhost fails to restart, it will either crash the `rabbit` application, or gives up and stops. This behaviour can be configured using `vhost_restart_strategy` environment variable. The variable can be either `stop_rabbit` or `give_up` (`stop_rabbit` by default).


hairyhum:
Upgrade from `3.4` requires old versions of some functions. We should require upgrade to `3.6` and then to `3.7`

michaelklishin:
I think that's reasonable.

src/rabbit_vhost.erl
```diff
@@ -42,6 +43,36 @@
...
+    ok.
+
+recover(VHost) ->
+    VHostDir = rabbit_vhost:msg_store_dir_path(VHost),
+    rabbit_log:info("Making sure data directory '~s' for vhost '~s' exists~n",
+                    [VHostDir, VHost]),
+    VHostStubFile = filename:join(VHostDir, ".vhost"),
+    ok = rabbit_file:ensure_dir(VHostStubFile),
+    ok = file:write_file(VHostStubFile, VHost),
+rabbit_log:info("Starting vhost ~p~n", [VHost]),
```
dcorbacho: Indentation, also other logs below

PR #1169, opened 2017-03-31
Add more info during rabbitmqctl shutdown

gerhard:
We want to know at which shutdown stage the command is at, and get
confirmation of successful shutdown.

Inform is used all over the place, is it OK to remove " ..." from all
rabbitmqctl commands that use it?

[#142699191]


hairyhum:
I think `...` is supposed to mean "command is not finished yet".

michaelklishin:
@hairyhum is this an explanation or are you requesting a change?

hairyhum:
Just an explanation. Dots after each log do not make much sense.

PR #1170, opened 2017-03-31
rabbit_epmd_monitor: call register_node on Mod, not its default

srenatus:
The module-to-be-called is stored in `init/1` in case it was overridden, so we should also call that possibly-overridden module, and not the default, to `register_node/2`.

(trivial change)

### DIFF
--- a/src/rabbit_epmd_monitor.erl
+++ b/src/rabbit_epmd_monitor.erl
@@ -92,6 +92,6 @@ check_epmd(#state{mod  = Mod,
                     "epmd does not know us, re-registering ~s at port ~b~n",
                     [Me, Port]),
                   rabbit_nodes:ensure_epmd(),
-                  erl_epmd:register_node(Me, Port);
+                  Mod:register_node(Me, Port);
         _      -> ok
     end.


michaelklishin:
@srenatus I love the branch name :) Should this be backported to `3.6.x`?

srenatus:
@michaelklishin that's up to you -- I don't need it urgently 😃

michaelklishin:
@srenatus understood, we will backport it. May I ask how/what for do you override `epmd_module`?

michaelklishin:
@srenatus the reason why I'm asking is that we've seen one case for doing that so far, would be great to know if there may be others.

srenatus:
@michaelklishin I was just poking around the code when investigating various options around controlling epmd better. One option was getting rid of epmd and going the route [outlined here](https://www.erlang-solutions.com/blog/erlang-and-elixir-distribution-without-epmd.html), which would have made this necessary.

Is that "other case" you've seen before documented publicly? I'd be curious about that, too 😃

michaelklishin:
@srenatus some companies are paranoid about having services that don't require authentication, even if they are seriously limited (like modern epmd is). The only solution in some of those cases is to avoid using epmd.

srenatus:
@michaelklishin thanks for sharing :+1:

michaelklishin:
Thank you for contributing!

I will cherry-pick to `stable.`

PR #1175, opened 2017-04-05
Register with peer discovery backend on start and stop

### LINKED ISSUE #1143: Register (and unregister) with peer discovery backend
This is a follow-up to #487. To truly port the ideas from rabbitmq-autocluster we need to finish node registration. This slipped through the cracks because our 2 built-in backends don't support peer registration:

 * Classic config
 * DNS

This is quite straightforward for joins but less so for when a node leaves the cluster.

michaelklishin:
Much like rabbitmq-autocluster does it. Fixes #1143.

This needs a bit more testing e.g. with `rabbitmqctl stop_app` and pausing partition handling modes. Also it would be nice to port rabbitmq-autocluster's Consul module (this is something we are quite familiar with by now) to master before we can consider this complete.


src/rabbit.erl
```diff
@@ -472,6 +474,7 @@ stop() ->
     rabbit_log:info("RabbitMQ is asked to stop...~n", []),
     Apps = ?APPS ++ rabbit_plugins:active(),
     stop_apps(app_utils:app_dependency_order(Apps, true)),
+    rabbit_peer_discovery:maybe_unregister(),
```
hairyhum: Is that a duplicate call? `prep_stop` is supposed to be called by `stop_apps`.

michaelklishin:
Let's keep this as `DO NOT MERGE` until we have some experience with a new backend-specific plugin (such as Consul) on top of this.

PR #1200, opened 2017-04-27
[stable] RabbitMQ-server does not quit with Ctrl-C

sodre:
This is the same PR as #1192 but against `stable`. 

/bin/sh may point to dash which has a bug with the trap calls in
lines 290-291. In that case "control-c" will kill the wait
command and leave rabbitmq-server still running in the
background.

The fix is to add a noop around the wait command.

Signed-off-by: Patrick Sodré <sodre@sodre.co>


dumbbell:
The patch from #1192 was cherry-picked to `stable` so we can close this pull request.

See #1192 for the entire discussion around the solution. Thank you again @sodre!

PR #1201, opened 2017-04-27
docs: Convert manpages from DocBook to mdoc(7)

dumbbell:
Writing those manpages directly in the target markup language allows to use its full expressiveness, unlike the generated output from DocBook.

To create their HTML versions, we use mandoc(1), plus a small awk script to adapt the generated HTML to our needs. I tried groff(1) and man2html(1) but the former transforms nearly everything to simple `<p>` and the later tries to reproduce the rendering of a manpage in a terminal.

Manpages in section 1 are moved to section 8 because they are commands to manage a system service, not general purpose commands.

This commits requires a change in the website to integrate the new generated HTML manpages.

[#143563295]


michaelklishin:
Should this and https://github.com/rabbitmq/rabbitmq-website/pull/375, https://github.com/rabbitmq/rabbitmq-public-umbrella/pull/49 be backported to `stable`?

PR #1204, opened 2017-04-30
Ensure ERL_LIBS starts with a valid path

ebelew:
Works around empty value for `ERL_LIBS` on first call to `:filter_path` which was causing erl.exe to crash on any rabbitmq command.

### DIFF
--- a/scripts/rabbitmq-env.bat
+++ b/scripts/rabbitmq-env.bat
@@ -417,7 +417,12 @@ set paths=
 exit /b
 
 :filter_path
-set ERL_LIBS=%ERL_LIBS%;%~dps1%~n1%~x1
+REM Ensure ERL_LIBS begins with valid path
+IF [%ERL_LIBS%] EQU [] (
+    set ERL_LIBS=%~dps1%~n1%~x1
+) else (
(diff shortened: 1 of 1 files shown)


michaelklishin:
@ebelew thank you. Can you please provide more details?

ebelew:
Before this patch, the env was getting set with the following command:

    set ERL_LIBS=;C:\PROGRA~1\RABBIT~1\RABBIT~1.6\plugins

After many trials and tribulations involving writing a custom batch script using the output of `set` and testing various variables, I found that if I manually set the `ERL_LIBS` value with:

  set ERL_LIBS=C:\PROGRA~1\RABBIT~1\RABBIT~1.6\plugins

`rabbitmqctl.bat`, `rabbitmq-server.bat` and others ran correctly without causing the error `erl.exe has stopped`. Therefor, I added logic to remove the empty item at the beginning of the list.

 While this is most likely technically a bug in erlang, this workaround allowed me to complete setup and get RMQ Server running on my local windows server.

michaelklishin:
Makes sense. Thank you!

michaelklishin:
@kjnilsson we should cherry-pick to `stable`.

PR #1205, opened 2017-05-01
Inject a randomized startup delay

### LINKED ISSUE #1202: Introduce a randomized startup delay for peer discovery mechanisms that support registration
As of #486 and #1143 RabbitMQ has a pluggable peer discovery mechanism. This unifies an "autoclustering" feature the broker has had for a while with some ideas from @gmr's [rabbitmq-autocluster](https://github.com/rabbitmq/rabbitmq-autocluster).

However, we haven't adopted the randomized delay feature from the plugin. Consider a cluster of 5 nodes that is provisioned from scratch on AWS. Nodes 1 and 2 boot in parallel and discover no peers, therefore register as ones. This is a natural race condition. Other nodes discover 1 (or 2) and join them. As a result, due to the race condition there are 2 clusters instead of 1. Provision 1 cluster, get 1 for free!

This problem is present in automated cluster provisioning of some other tools and there are two common workarounds:

 * Provision a "seed node" first, then the rest
 * Inject a random delay to work around the race condition (in practice)

The latter approach may sound silly but it works well in practice and was adopted in some shape or form in several systems and distributed algorithms (Raft is a popular recent example). It also worked reasonably well for rabbitmq-autocluster.

Therefore I suggest that we introduce a similar delay, e.g. in the range of 5 to 90 seconds by default, for backends that support peer registration. Backends that rely on node config file or pre-configured DNS records are not affected by the above problem and therefore need no delay.

michaelklishin:
…for peer discovery backends that support registration.

Fixes #1202.

### DIFF
--- a/src/rabbit_peer_discovery.erl
+++ b/src/rabbit_peer_discovery.erl
@@ -22,7 +22,8 @@
 
 -export([discover_cluster_nodes/0, backend/0, node_type/0,
          normalize/1, format_discovered_nodes/1, log_configured_backend/0,
-         register/0, unregister/0, maybe_register/0, maybe_unregister/0]).
+         register/0, unregister/0, maybe_register/0, maybe_unregister/0,
+         maybe_inject_randomized_delay/0]).
 -export([append_node_prefix/1, node_prefix/0]).
 
 -define(DEFAULT_BACKEND,   rabbit_peer_discovery_classic_config).
@@ -31,6 +32,9 @@
 -define(DEFAULT_NODE_TYPE, disc).
 %% default node prefix to attach to discovered hostnames
 -define(DEFAULT_PREFIX, "rabbit").
+%% default randomized delay range, in seconds
+-define(DEFAULT_STARTUP_RANDOMIZED_DELAY, {5, 60}).
+
 -define(NODENAME_PART_SEPARATOR, "@").
 
 
@@ -96,6 +100,56 @@ maybe_unregister() ->
     false ->
       rabbit_log:info("Peer discovery backend ~s does not support registration, skipping unregistration.", [Backend]),
       ok
+  end.
+
+
+-spec maybe_inject_randomized_delay() -> ok.
+maybe_inject_randomized_delay() ->
+  Backend = backend(),
+  case Backend:supports_registration() of
+    true  ->
+      rabbit_log:info("Peer discovery backend ~s supports registration.", [Backend]),
+      inject_randomized_delay();
+    false ->
+      rabbit_log:info("Peer discovery backend ~s does not support registration, skipping randomized startup delay.", [Backend]),
+      ok
+  end.
+
+-spec inject_randomized_delay() -> ok.
+
+inject_randomized_delay() ->
+    {Min, Max} = case randomized_delay_range() of
+                     {A, B} -> {A, B};
+                     [A, B] -> {A, B}
+                 end,
+    case {Min, Max} of
+        %% When the max value is set to 0, consider the delay to be disabled.
+        %% In addition, `rand:uniform/1` will fail with a "no function clause"
+        %% when the argument is 0.
+        {_, 0} ->
+            rabbit_log:info("Randomized delay range's upper bound is set to 0. Considering it disabled."),
+            ok;
+        {_, N} when is_number(N) ->
+            rand:seed(exsplus),
+            RandomVal  = rand:uniform(round(N)),
+            rabbit_log:debug("Randomized startup delay: configured range is from ~p to ~p, PRNG pick: ~p...", [Min, Max, RandomVal]),
+            Effective  = case RandomVal < Min of
+                             true  -> Min;
+                             false -> RandomVal
+                         end,
+            rabbit_log:info("Will wait for ~p seconds before proceeding with regitration...", [Effective]),
+            timer:sleep(Effective * 1000),
+            ok
+    end.
+
+-spec randomized_delay_range() -> {integer(), integer()}.
+
+randomized_delay_range() ->
+    case application:get_env(rabbit, autocluster) of
+    {ok, Proplist} ->
+      proplists:get_value(randomized_startup_delay_range, Proplist, ?DEFAULT_STARTUP_RANDOMIZED_DELAY);
+    undefined      ->
+      ?DEFAULT_STARTUP_RANDOMIZED_DELAY
   end.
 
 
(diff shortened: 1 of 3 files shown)


src/rabbit_peer_discovery.erl
```diff
@@ -99,6 +103,56 @@ maybe_unregister() ->
...
+      rabbit_log:info("Peer discovery backend ~s does not support registration, skipping randomized startup delay.", [Backend]),
+      ok
+  end.
+
+-spec inject_randomized_delay() -> ok.
+
+inject_randomized_delay() ->
+    {Min, Max} = case randomized_delay_range() of
+                     {A, B} -> {A, B};
+                     [A, B] -> {A, B}
```
hairyhum: What is the purpose of supporting lists here?

src/rabbit_peer_discovery.erl
```diff
@@ -99,6 +103,56 @@ maybe_unregister() ->
   end.
 
 
+-spec maybe_inject_randomized_delay() -> ok.
+maybe_inject_randomized_delay() ->
+  Backend = backend(),
+  case Backend:supports_registration() of
```
Gsantomaggio: Should we add `supports_registration()` to [rabbit_peer_discovery_backend](
https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_peer_discovery_backend.erl) ?

src/rabbit_peer_discovery.erl
michaelklishin: Some users can't tell the difference between tuples and lists.

src/rabbit_peer_discovery.erl
michaelklishin: Oh, I thought I did that as part of https://github.com/rabbitmq/rabbitmq-server/issues/1143 but looks like that change was lost. Adding it now.

michaelklishin: Done.

src/rabbit_peer_discovery.erl
```diff
@@ -99,6 +103,56 @@ maybe_unregister() ->
...
+                         end,
+            rabbit_log:info("Will wait for ~p seconds before proceeding with regitration...", [Effective]),
+            timer:sleep(Effective * 1000),
+            ok
+    end.
+
+-spec randomized_delay_range() -> {integer(), integer()}.
+
+randomized_delay_range() ->
+    case application:get_env(rabbit, autocluster) of
```
dcorbacho: you could use `get_env/3`

michaelklishin: Then it'd have to return a proplist for default value, which seems pretty weird to me.

michaelklishin:
Merging per verbal approval from @dcorbacho.

binarin:
This should be accompanied by a BIG WARNING somewhere in the documentation that external split-brain monitoring is 100% necessary. We were doing a lot of manual and automated testing, and the rate of failures was unacceptable even with high delays.

michaelklishin:
@binarin well, how high is high? I have used the same approach with e.g. Cassandra more than once in the past and 60-90 second delays worked just fine. I can't say I was running cluster formation in a loop, of course.

michaelklishin:
The docs will be rewritten to cover peer discovery, so there will be a chance to dedicate a subsection to it.

binarin:
Even a single failure is too much for a production system ) And I've seen this several times with timeouts ranging from 30 to 60 seconds while developing. But maybe I'm just extremely unlucky.

michaelklishin:
@binarin I am questioning that but distributed locking also has a natural race condition and is known to be a much harder problem than it seems, so we'll see how it goes for inclusion into the core.

src/rabbit_peer_discovery.erl
```diff
@@ -99,6 +103,56 @@ maybe_unregister() ->
...
+        {_, N} when is_number(N) ->
+            rand:seed(exsplus),
+            RandomVal  = rand:uniform(round(N)),
+            rabbit_log:debug("Randomized startup delay: configured range is from ~p to ~p, PRNG pick: ~p...", [Min, Max, RandomVal]),
+            Effective  = case RandomVal < Min of
+                             true  -> Min;
+                             false -> RandomVal
+                         end,
+            rabbit_log:info("Will wait for ~p seconds before proceeding with regitration...", [Effective]),
+            timer:sleep(Effective * 1000),
```
binarin: @michaelklishin I've started to think about a possible reason for my extreme unluckiness, and I believe I've found it. Random number should be taken from a  range specified in milliseconds right from the start, and not by multiplying by 1000 afterwards.

src/rabbit_peer_discovery.erl
```diff
@@ -99,6 +103,56 @@ maybe_unregister() ->
...
+                 end,
+    case {Min, Max} of
+        %% When the max value is set to 0, consider the delay to be disabled.
+        %% In addition, `rand:uniform/1` will fail with a "no function clause"
+        %% when the argument is 0.
+        {_, 0} ->
+            rabbit_log:info("Randomized delay range's upper bound is set to 0. Considering it disabled."),
+            ok;
+        {_, N} when is_number(N) ->
+            rand:seed(exsplus),
```
binarin: `rabbit_misc:random/1` is already responsible for PRNG seeding (or non-seeding for erlang > 18.3)

src/rabbit_peer_discovery.erl
michaelklishin: This makes sense, I will change this.

src/rabbit_peer_discovery.erl
michaelklishin: Thanks.

michaelklishin: Hm, perhaps in stable but not in master? https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_misc.erl#L1139

binarin: It was a problem for old `random` module, thus there is some code to work around this in `stable`. But modern `rand`  (i.e. `rand:uniform/1`) will initialize seed automatically and thus all this support code was dropped from `rabbit_misc`.

michaelklishin: I will test again without the explicit seeding.

michaelklishin:
Err, just noticed a typo in a comment above that made it look like I'm questioning @binarin's feeback. I'm definitely not, the randomization is a good enough solution but not a guarantee. Unfortunately neither is locking :(

PR #1208, opened 2017-05-05
Use plugins expand dir with default fallback value

binarin:
In addition to #1134


michaelklishin:
Please re-submit against the `stable` branch.

binarin:
`strictly_plugins` is only in `master`, should I split this into 2 separate PRs?

michaelklishin:
ahem, then yes

PR #1226, opened 2017-05-19
Don't expand plugins that are already unpacked

binarin:
This prevents creating unnecessary files when running in embedded
mode, and in development mode it keeps code path clean and compatible
with different interactive tools.

### DIFF
--- a/src/rabbit_plugins.erl
+++ b/src/rabbit_plugins.erl
@@ -262,12 +262,12 @@ prepare_plugin(#plugin{type = ez, name = Name, location = Location}, ExpandDir) 
                 [PluginAppDescPath|_] ->
                     prepare_dir_plugin(PluginAppDescPath);
                 _ ->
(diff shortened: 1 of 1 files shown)


hairyhum:
Looks good. Though I'm a bit concerned about using `plugins_dist_dir` in the `active/0` function, because it's unzipping an `.app` file every time. On other hand, the function is not widely used (only in plugin operation), so it should be fine.

binarin:
I've updated the commit message.

michaelklishin:
We will postpone this until after 3.6.10 GA ships.

michaelklishin:
@binarin I updated a few error messages and tuple tags in https://github.com/rabbitmq/rabbitmq-server/pull/1227. Can you please update this PR accordingly? Thank you.

binarin:
I've cherry-picked fb17eed here.

PR #1227, opened 2017-05-19
Don't expand plugins that are already unpacked - master branch

binarin:
In addition to #1226 to make it compatible with `master`


michaelklishin:
@binarin I changed the wording and error tuple tags a bit. Thank you.

PR #1252, opened 2017-06-07
Improve documentation around new default for background garbage collection

FabianPonce:
2f3fb8023ba1f535b13f4deb97106c08b78bc1d9 (v3.6.7) altered the default setting for `background_gc_enabled` to false, however, the default configuration files seem to imply that the default value for this is still set to true. 

This pull request corrects the default configuration and makes a small change to background_gc.erl so that the defaults which are set in the Makefile are clearly communicated to anyone viewing the source code as well.

### DIFF
--- a/src/background_gc.erl
+++ b/src/background_gc.erl
@@ -74,7 +74,7 @@ interval_gc(State = #state{last_interval = LastInterval}) ->
     State#state{last_interval = Interval}.
 
 gc() ->
-    Enabled = rabbit_misc:get_env(rabbit, background_gc_enabled, true),
+    Enabled = rabbit_misc:get_env(rabbit, background_gc_enabled, false),
     case Enabled of
         true ->
             [garbage_collect(P) || P <- processes(),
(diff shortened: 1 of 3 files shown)


src/background_gc.erl
```diff
@@ -74,7 +74,7 @@ interval_gc(State = #state{last_interval = LastInterval}) ->
     State#state{last_interval = Interval}.
 
 gc() ->
-    Enabled = rabbit_misc:get_env(rabbit, background_gc_enabled, true),
```
michaelklishin: `rabbit.background_gc_enabled` has an [application-level default](https://github.com/rabbitmq/rabbitmq-server/blob/master/Makefile#L117), so this should not break anything.

michaelklishin:
@FabianPonce thank you. Can you please submit the same change against `stable`? (note: there won't be `rabbitmq.example.conf` in that branch).

FabianPonce:
See #1253 for PR against stable branch. Thanks for the quick review, guys!

Issue #1257, opened 2017-06-08
Implement locking support in peer discovery


dcorbacho:
Originally part of the `rabbitmq_autocluster` plugin: https://github.com/rabbitmq/rabbitmq-autocluster/pull/6

michaelklishin:
`effort-medium` seems reasonable here because this involves 3 different engineers, multiple plugins and quite a bit of testing (on top of what's already been done in https://github.com/rabbitmq/rabbitmq-autocluster/pull/6).

michaelklishin:
There is no PR for `rabbitmq-common` where [peer_discovery_backend](https://github.com/rabbitmq/rabbitmq-common/blob/rabbitmq-server-1257/src/rabbit_peer_discovery_backend.erl) resides.

PR #1258, opened 2017-06-09
Locking support in cluster formation

dcorbacho:
Adapted from `rabbitmq_autocluster` to avoid race conditions on startup

Part of #1257

### DIFF
--- a/src/rabbit_peer_discovery.erl
+++ b/src/rabbit_peer_discovery.erl
@@ -23,8 +23,9 @@
 -export([discover_cluster_nodes/0, backend/0, node_type/0,
          normalize/1, format_discovered_nodes/1, log_configured_backend/0,
          register/0, unregister/0, maybe_register/0, maybe_unregister/0,
-         maybe_inject_randomized_delay/0]).
--export([append_node_prefix/1, node_prefix/0]).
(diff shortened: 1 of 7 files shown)


src/rabbit_peer_discovery.erl
```diff
@@ -183,6 +204,34 @@ unregister() ->
       ok
   end.
 
+-spec lock() -> ok | {ok, Data :: term()} | not_supported | {error, Reason :: string()}.
+
+lock() ->
+    Backend = backend(),
+    rabbit_log:info("Will try to lock with peer discovery backend ~s", [Backend]),
+    case Backend:lock(node()) of
```
michaelklishin: Should the backend indicate whether locking is supported, like we do for registration? Should `lock/1` and `unlock/1` be a part of https://github.com/rabbitmq/rabbitmq-common/blob/rabbitmq-server-1257/src/rabbit_peer_discovery_backend.erl?

michaelklishin: There was a missing PR: https://github.com/rabbitmq/rabbitmq-common/pull/204/files.

The `lock/1` return value of `unsupported` is not very consistent with `supports_registration/0` but
it's consistent with rabbitmq-autocluster :/. Arguably this is not a show stopper for #1258 and 3.7.0 release plans.

PR #1262, opened 2017-06-13
The function in term_to_binary_compat to format stomp queue name hashes.

hairyhum:
Part of fix for rabbitmq/rabbitmq-stomp#115

It is unsafe to use `term_to_binary` to generate IDs like queue names,
because the format can change in future erlang versions.
STOMP plugin uses 2-element tuples of strings and binaries to
generate queue names (for example for topic subscriptions).


hairyhum:
Replaced by #1268

Issue #1243, opened 2017-06-01
Messages can be lost on upgrade from Erlang/OTP 19 (or earlier) to Erlang/OTP 20


hairyhum:
OTP-20 has different format for `term_to_binary` format, probably changed in [OTP-14337](https://github.com/erlang/otp/commit/82e849adee6e2fd20e2a3faa6ecb463cc2c7256e).
Because queue index directory names are generated using `term_to_binary` they are different in OTP-20 and pre-20, so any data created before erlang upgrade will be lost because RabbitMQ deletes all unknown queue index directories.

The milestone says `3.7.0` but together with some backwards-compatibility restoring changes by the the Erlang/OTP team, [we hope to support OTP 20 in 3.6.11 as well](https://groups.google.com/forum/#!topic/rabbitmq-users/_imbAavBYjY).

See [RabbitMQ Erlang/OTP 20 compatibility](https://groups.google.com/forum/#!topic/rabbitmq-users/_imbAavBYjY) thread on rabbitmq-users for more details and updates.

michaelklishin:
It's not clear if we can make Erlang upgrades from 19 to 20 work for `3.6.x` (that is, without breaking changes in RabbitMQ itself), so milestone => `3.7.0`.

hairyhum:
In that case there is not much difference between 3.6 and 3.7, but limiting fix to 3.7 still makes sense.

The problem here is in this line https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_queue_index.erl#L694 
We get `term_to_binary` from a record with atoms inside and then calculate MD5 hash of it.
OTP-20 can decode binaries encoded in OTP-19, but it doesn't encode them the same way. And because we do MD5 we cannot decode it.
I can see 3 options for us:
  - Migrate dir names by generating the old binary format ourselves
  - Require OTP-19 to migrate them
  - Convince Erlang core devs to return backwards-compatibility option.

Third option can be hard, due to RC process, but could make sense not only for us, but other Erlang users as well.

hairyhum:
Discussion in erlang bugtracker https://bugs.erlang.org/browse/ERL-431

hairyhum:
OTP team does not want to ad options to `term_to_binary`, recommending us to hand-craft the "old format" values.

hairyhum:
The critical parts are queues and vhosts names. We also have several places across plugins which use `term_to_binary`, but they could be safe (needs verification)

[...]

PR #1268, opened 2017-06-16
Make term_to_binary_compat generate binaries using minor_version 1.

hairyhum:
Fix for #1243 for 3.6.x

term_to_binary now has `minor_version:2` to always generate utf-8 atoms,
while default is `1`, so binaries generated in OTP-20 should be
the same as in OTP-19 for non-unicode atoms.


hairyhum:
Requires rabbitmq/rabbitmq-server#1268

PR #1274, opened 2017-06-23
Pass in extra arguments to authentication backend

acogoluegnes:
Those extra arguments are extracted from an external module,
the convention being "rabbit_%protocol%_connection_info" for the module name.
This allows to pass in plugin-specific authentication arguments,
e.g. client ID in the case of MQTT.

Part of rabbitmq/rabbitmq-mqtt#139


src/rabbit_direct.erl
```diff
@@ -101,6 +105,45 @@ connect(Creds, VHost, Protocol, Pid, Infos) ->
...
+extract_extra_auth_props(Creds, VHost, Pid, Infos) ->
+    case extract_protocol(Infos) of
+        undefined ->
+            [];
+        Protocol ->
+            maybe_call_connection_info_module(Protocol, Creds, VHost, Pid, Infos)
+    end.
+
+extract_protocol(Infos) ->
+    case proplists:get_value(protocol, Infos, undefined) of
```
michaelklishin: We should test that all protocols we currently support either don't have this entry or have a connection info module.

acogoluegnes: Sorry, I don't get it. Here we get the protocol on the fly from the connection info (e.g. `{'MQTT', '3.1.1'}` and then build the module name from it (e.g. `rabbit_mqtt_connection_info`) and check if the module exists.

michaelklishin: OK, that takes care of what I had in mind (protocol plugins that we do not maintain).

PR #1283, opened 2017-07-01
Handle concurrent vhost creation and deletion better

### LINKED ISSUE #1280: Multiple processes tripped up by a non-existent vhost leads to node termination
Accidentally reproduced while working on an issue in CLI tools:

```
2017-06-30 10:45:37.649 [error] <0.505.0> Supervisor {<0.505.0>,rabbit_vhost_sup_wrapper} had child rabbit_vhost_sup started with rabbit_vhost_sup_wrapper:start_vhost_sup(<<"test_vhost1">>) at undefined exit with reason {'EXIT',{{badmatch,{error,{no_such_vhost,<<"test_vhost1">>}}},[{rabbit_recovery_terms,start,1,[{file,"src/rabbit_recovery_terms.erl"},{line,51}]},{rabbit_queue_index,start,2,[{file,"src/rabbit_queue_index.erl"},{line,502}]},{rabbit_variable_queue,start,2,[{file,"src/rabbit_variable_queue.erl"},{line,482}]},{rabbit_priority_queue,start,2,[{file,"src/rabbit_priority_queue.erl"},{line,92}]},{rabbit_amqqueue,recover,1,[{file,"src/rabbit_amqqueue.erl"},{line,236}]},{rabbit_vhost,recover,1,[{file,...},...]},...]}} in context start_error
2017-06-30 10:45:37.649 [info] <0.505.0> Making sure data directory '/var/folders/b3/6thx2ls900z6h456z2xb658m0000gn/T/rabbitmq-test-instances/rabbit/mnesia/rabbit/msg_stores/vhosts/BPY12RLD3HXBADXVD0QCEE7DC' for vhost 'test_vhost1' exists
2017-06-30 10:45:37.650 [error] <0.505.0> Supervisor {<0.505.0>,rabbit_vhost_sup_wrapper} had child rabbit_vhost_sup started with rabbit_vhost_sup_wrapper:start_vhost_sup(<<"test_vhost1">>) at undefined exit with reason {'EXIT',{{badmatch,{error,{no_such_vhost,<<"test_vhost1">>}}},[{rabbit_recovery_terms,start,1,[{file,"src/rabbit_recovery_terms.erl"},{line,51}]},{rabbit_queue_index,start,2,[{file,"src/rabbit_queue_index.erl"},{line,502}]},{rabbit_variable_queue,start,2,[{file,"src/rabbit_variable_queue.erl"},{line,482}]},{rabbit_priority_queue,start,2,[{file,"src/rabbit_priority_queue.erl"},{line,92}]},{rabbit_amqqueue,recover,1,[{file,"src/rabbit_amqqueue.erl"},{line,236}]},{rabbit_vhost,recover,1,[{file,...},...]},...]}} in context start_error
2017-06-30 10:45:37.650 [error] <0.505.0> Supervisor {<0.505.0>,rabbit_vhost_sup_wrapper} had child rabbit_vhost_sup started with rabbit_vhost_sup_wrapper:start_vhost_sup(<<"test_vhost1">>) at {restarting,<0.506.0>} exit with reason reached_max_restart_intensity in context shutdown
2017-06-30 10:45:37.650 [error] <0.328.0> Supervisor rabbit_vhost_sup_sup had child rabbit_vhost started with rabbit_vhost_sup_wrapper:start_link(<<"test_vhost1">>) at <0.505.0> exit with reason shutdown in context child_terminated
2017-06-30 10:45:37.650 [error] <0.328.0> Supervisor rabbit_vhost_sup_sup had child rabbit_vhost started with rabbit_vhost_sup_wrapper:start_link(<<"test_vhost1">>) at <0.505.0> exit with reason reached_max_restart_intensity in context shutdown
2017-06-30 10:45:37.650 [info] <0.338.0> Stopping message store for directory '/var/folders/b3/6thx2ls900z6h456z2xb658m0000gn/T/rabbitmq-test-instances/rabbit/mnesia/rabbit/msg_stores/vhosts/9FYW91Z090TQ8GC4PKLCNIK12/msg_store_persistent'
2017-06-30 10:45:37.650 [error] <0.257.0> Supervisor rabbit_sup had child rabbit_vhost_sup_sup started with rabbit_vhost_sup_sup:start_link() at <0.328.0> exit with reason shutdown in context child_terminated
2017-06-30 10:45:37.650 [info] <0.371.0> Stopping message store for directory '/var/folders/b3/6thx2ls900z6h456z2xb658m0000gn/T/rabbitmq-test-instances/rabbit/mnesia/rabbit/msg_stores/vhosts/628WB79CIFDYO9LJI6DKMI09L/msg_store_persistent'
2017-06-30 10:45:37.650 [error] <0.257.0> Supervisor rabbit_sup had child rabbit_vhost_sup_sup started with rabbit_vhost_sup_sup:start_link() at <0.328.0> exit with reason reached_max_restart_intensity in context shutdown
```

To reproduce this, run an integration test that creates a vhost, does something very quickly
(e.g. in our CLI suite, runs a command that fails validation immediately) and deletes the vhost.
As a result, one of the vhost supervision tree processes can trip up because the record no longer
exists due to a concurrent deletion.

Such cases should be logged as errors and ignored otherwise.

michaelklishin:
In the case of a vhost creation and immediate deletion those
operations can end up executing concurrently. This changes
a number of things to be more resilient:

 * Default vhost restart strategy changes to "continue" (previosly "ignore").
   It's a tough default to pick but we believe the damage of a single vhost terminating
   the entire node that may have a lot of users is greater than the need to set up an alternative
   strategy for environments where only one vhost is used. Note that the event of
   message store or vhost supervision tree termination should be very rare.

 * Virtual host watcher process now safely shuts down vhost supervision tree
    instead of existing and triggering a cascading shutdown that reaches all the way
    to the `rabbit` app.

 * `rabbit_vhost:add/2` now propagates errors to the caller.

 * For vhost [message store] recovery, we now log "no_such_vhost" errors
   in a sensible way instead of a bunch of process terminations.

 * Max restart intensity settings for `rabbit_vhost_sup_wrapper` are
   more forgiving.

There are also drive by doc changes.

How to reproduce: in the CLI repo, run `./test/list_vhost_limits_command_test.exs` multiple times in a row, e.g.:

```
for i in {1..100}; do gmake tests TEST_FILE=./test/list_vhost_limits_command_test.exs; done
```

or, from a node REPL:

``` erlang
Xs = lists:seq(1, 250), lists:foreach(fun(_) -> spawn(fun() -> rabbit_vhost:add(<<"123">>, <<"internal">>) end), spawn(fun () -> rabbit_vhost:delete(<<"123">>, <<"internal">>) end) end, Xs).
```

By the end you may observe errors logged but the node should not terminate because
one vhost was detected to be toast.

Fixes #1280.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,9 +122,8 @@ define PROJECT_ENV
 	    {disk_monitor_failure_retries, 10},
 	    {disk_monitor_failure_retry_interval, 120000},
 	    %% either "stop_node" or "continue".
-	    %% by default we choose to not restart the entire node if one
-	    %% vhost had to shut down e.g. because of a message store exception
-	    %% see server#1158 and server#1280
+	    %% by default we choose to not terminate the entire node if one
+	    %% vhost had to shut down, see server#1158 and server#1280
 	    {vhost_restart_strategy, continue}
 	  ]
 endef
--- a/Makefile
+++ b/Makefile
@@ -112,17 +112,20 @@ define PROJECT_ENV
 	                            {passphrase, undefined}
 	                           ]},
 
-	    %% rabbitmq-server-973
+	    %% rabbitmq-server#973
 	    {queue_explicit_gc_run_operation_threshold, 1000},
 	    {lazy_queue_explicit_gc_run_operation_threshold, 1000},
 	    {background_gc_enabled, false},
 	    {background_gc_target_interval, 60000},
-	    %% rabbitmq-server-589
+	    %% rabbitmq-server#589
 	    {proxy_protocol, false},
 	    {disk_monitor_failure_retries, 10},
 	    {disk_monitor_failure_retry_interval, 120000},
-	    %% either "stop_node" or "ignore"
-	    {vhost_restart_strategy, stop_node}
+	    %% either "stop_node" or "continue".
+	    %% by default we choose to not restart the entire node if one
+	    %% vhost had to shut down e.g. because of a message store exception
+	    %% see server#1158 and server#1280
+	    {vhost_restart_strategy, continue}
 	  ]
 endef
 
--- a/priv/schema/rabbit.schema
+++ b/priv/schema/rabbit.schema
@@ -976,11 +976,11 @@ end}.
 {mapping, "proxy_protocol", "rabbit.proxy_protocol",
     [{datatype, {enum, [true, false]}}]}.
 
-%% Whether to stop the rabbit application if VHost data
-%% cannot be recovered.
+%% Whether to stop the rabbit application if a vhost message
+%% store or recovery fails (or another unrecoverable issue is detected).
 
 {mapping, "vhost_restart_strategy", "rabbit.vhost_restart_strategy",
-    [{datatype, {enum, [stop_node, ignore]}}]}.
+    [{datatype, {enum, [stop_node, continue, transient, persistent]}}]}.
 
 % ==========================
 % Lager section
--- a/priv/schema/rabbit.schema
+++ b/priv/schema/rabbit.schema
@@ -976,8 +976,8 @@ end}.
 {mapping, "proxy_protocol", "rabbit.proxy_protocol",
     [{datatype, {enum, [true, false]}}]}.
 
-%% Whether to stop the rabbit application if a vhost message
-%% store or recovery fails (or another unrecoverable issue is detected).
+%% Whether to stop the rabbit application if a vhost has
+%% to terminate for any reason.
 
 {mapping, "vhost_restart_strategy", "rabbit.vhost_restart_strategy",
     [{datatype, {enum, [stop_node, continue, transient, persistent]}}]}.
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -234,8 +234,14 @@ recover(VHost) ->
     %% for further processing in recover_durable_queues.
     {ok, OrderedRecoveryTerms} =
         BQ:start(VHost, [QName || #amqqueue{name = QName} <- Queues]),
-    {ok, _} = rabbit_amqqueue_sup_sup:start_for_vhost(VHost),
-    recover_durable_queues(lists:zip(Queues, OrderedRecoveryTerms)).
+    case rabbit_amqqueue_sup_sup:start_for_vhost(VHost) of
+        {ok, _} ->
+            recover_durable_queues(lists:zip(Queues, OrderedRecoveryTerms));
+        %% a failure
+        _       ->
+            rabbit_log:error("Failed to start queue supervisors for vhost '~s', skipping queue recovery", [VHost]),
+            []
+    end.
 
 stop(VHost) ->
     ok = rabbit_amqqueue_sup_sup:stop_for_vhost(VHost),
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -235,12 +235,11 @@ recover(VHost) ->
     {ok, OrderedRecoveryTerms} =
         BQ:start(VHost, [QName || #amqqueue{name = QName} <- Queues]),
     case rabbit_amqqueue_sup_sup:start_for_vhost(VHost) of
-        {ok, _} ->
+        {ok, _}         ->
             recover_durable_queues(lists:zip(Queues, OrderedRecoveryTerms));
-        %% a failure
-        _       ->
-            rabbit_log:error("Failed to start queue supervisors for vhost '~s', skipping queue recovery", [VHost]),
-            []
+        {error, Reason} ->
+            rabbit_log:error("Failed to start queue supervisor for vhost '~s': ~s", [VHost, Reason]),
+            throw({error, Reason})
     end.
 
 stop(VHost) ->
--- a/src/rabbit_amqqueue_sup_sup.erl
+++ b/src/rabbit_amqqueue_sup_sup.erl
@@ -63,7 +63,7 @@ find_for_vhost(VHost, Node) ->
         Result -> {error, {queue_supervisor_not_found, Result}}
     end.
 
--spec start_for_vhost(rabbit_types:vhost()) -> {ok, pid()} | ok | {error, term()}.
+-spec start_for_vhost(rabbit_types:vhost()) -> {ok, pid()} | {error, term()}.
 start_for_vhost(VHost) ->
     case rabbit_vhost_sup_sup:vhost_sup(VHost) of
         {ok, VHostSup} ->
@@ -72,8 +72,8 @@ start_for_vhost(VHost) ->
               {rabbit_amqqueue_sup_sup,
                {rabbit_amqqueue_sup_sup, start_link, []},
                transient, infinity, supervisor, [rabbit_amqqueue_sup_sup]});
-        %% we can get here if a vhost is created and immediately
-        %% deleted, e.g. some integration tests do it
+        %% we can get here if a vhost is added and removed concurrently
+        %% e.g. some integration tests do it
         {error, {no_such_vhost, VHost}} ->
             rabbit_log:error("Failed to start a queue process supervisor for vhost ~s: vhost no longer exists!",
                              [VHost]),
--- a/src/rabbit_amqqueue_sup_sup.erl
+++ b/src/rabbit_amqqueue_sup_sup.erl
@@ -63,17 +63,32 @@ find_for_vhost(VHost, Node) ->
         Result -> {error, {queue_supervisor_not_found, Result}}
     end.
 
--spec start_for_vhost(rabbit_types:vhost()) -> {ok, pid()} | {error, term()}.
+-spec start_for_vhost(rabbit_types:vhost()) -> {ok, pid()} | ok | {error, term()}.
 start_for_vhost(VHost) ->
-    {ok, VHostSup} = rabbit_vhost_sup_sup:vhost_sup(VHost),
-    supervisor2:start_child(
-        VHostSup,
-        {rabbit_amqqueue_sup_sup,
-         {rabbit_amqqueue_sup_sup, start_link, []},
-         transient, infinity, supervisor, [rabbit_amqqueue_sup_sup]}).
+    case rabbit_vhost_sup_sup:vhost_sup(VHost) of
+        {ok, VHostSup} ->
+            supervisor2:start_child(
+              VHostSup,
+              {rabbit_amqqueue_sup_sup,
+               {rabbit_amqqueue_sup_sup, start_link, []},
+               transient, infinity, supervisor, [rabbit_amqqueue_sup_sup]});
+        %% we can get here if a vhost is created and immediately
+        %% deleted, e.g. some integration tests do it
+        {error, {no_such_vhost, VHost}} ->
+            rabbit_log:error("Failed to start a queue process supervisor for vhost ~s: vhost no longer exists!",
+                             [VHost]),
+            ok
+    end.
 
 -spec stop_for_vhost(rabbit_types:vhost()) -> ok.
 stop_for_vhost(VHost) ->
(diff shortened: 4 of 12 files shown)


src/rabbit_vhost_sup_wrapper.erl
```diff
@@ -26,15 +32,9 @@
...
-%% provide exactly once restart.
-
-%% rabbit_vhost_sup supervisor children are added dynamically,
-%% so one_for_all strategy cannot be used.
-
 init([VHost]) ->
-    %% Two restarts in 1 hour. One per message store.
-    {ok, {{one_for_all, 2, 3600000},
+    %% 2 restarts in 5 minutes. One per message store.
+    {ok, {{one_for_all, 2, 5000},
```
dcorbacho: This is 83 min (5000 seconds), 5 min would be 300

hairyhum:
Should we consider adding VHosts synchronously?

hairyhum:
It seems like it's actually added concurrently, but when it's deleted the watcher detects deletion before it's stopped, so it takes the supervisor down abnormally instead of waiting for it to stop normally.
Disabling watcher before deleting a vhost from the mnesia db should fix this.

hairyhum:
After investigation, it seems like the watcher is doing it's job correctly and it is not a problem that it's taking down the supervisor when the VHost is deleted. But the `permanent` restart strategy is wrong. Instead of `permanent` vs `transient` it should be `transient` vs `temporary`, because `transient` will still take down entire node if supervisor exits abnormally (e.g. when data is corrupted), but the `permanent` strategy was taking down the node even if the supervisor exits normally.

hairyhum:
Disabling the watcher makes not much sense, making it stop the supervisor using `rabbit_vhost_sup_sup` API instead of restarts should do the trick.

michaelklishin:
@hairyhum in my case the sequence of events was `created-then-deleted` — I could tell from the crashes in the log. But you have a point. Feel free to submit a PR that stops the watcher first, it makes sense.

src/rabbit_vhost_sup_sup.erl
```diff
@@ -169,3 +168,14 @@ vhost_sup_pid(VHost) ->
...
+vhost_restart_strategy() ->
+    %% This assumes that a single vhost termination should not shut down nodes
+    %% unless the operator opts in.
+    case application:get_env(rabbit, vhost_restart_strategy, continue) of
+        continue  -> transient;
+        stop_node -> permanent;
+        transient -> transient;
+        permanent -> permanent;
+        %% old name of continue
+        ignore    -> transient
```
hairyhum: I don't think we need `ignore` anymore, since it's not released yet.

Makefile
```diff
@@ -112,17 +112,20 @@ define PROJECT_ENV
...
+	    %% rabbitmq-server#589
 	    {proxy_protocol, false},
 	    {disk_monitor_failure_retries, 10},
 	    {disk_monitor_failure_retry_interval, 120000},
-	    %% either "stop_node" or "ignore"
-	    {vhost_restart_strategy, stop_node}
+	    %% either "stop_node" or "continue".
+	    %% by default we choose to not restart the entire node if one
+	    %% vhost had to shut down e.g. because of a message store exception
+	    %% see server#1158 and server#1280
```
hairyhum: Is this comment still relevant?

priv/schema/rabbit.schema
```diff
@@ -976,11 +976,11 @@ end}.
 {mapping, "proxy_protocol", "rabbit.proxy_protocol",
     [{datatype, {enum, [true, false]}}]}.
 
-%% Whether to stop the rabbit application if VHost data
-%% cannot be recovered.
+%% Whether to stop the rabbit application if a vhost message
+%% store or recovery fails (or another unrecoverable issue is detected).
```
hairyhum: Message store recovery failures on freshly started node or when vhost is just added will not cause application to stop. How can we mention that?

src/rabbit_amqqueue.erl
```diff
@@ -234,8 +234,14 @@ recover(VHost) ->
...
         BQ:start(VHost, [QName || #amqqueue{name = QName} <- Queues]),
-    {ok, _} = rabbit_amqqueue_sup_sup:start_for_vhost(VHost),
-    recover_durable_queues(lists:zip(Queues, OrderedRecoveryTerms)).
+    case rabbit_amqqueue_sup_sup:start_for_vhost(VHost) of
+        {ok, _} ->
+            recover_durable_queues(lists:zip(Queues, OrderedRecoveryTerms));
+        %% a failure
+        _       ->
+            rabbit_log:error("Failed to start queue supervisors for vhost '~s', skipping queue recovery", [VHost]),
+            []
```
hairyhum: Should we raise the exception here? It's not safe to continue recovery process if the vhost supervisor is not found.

src/rabbit_amqqueue_sup_sup.erl
```diff
@@ -63,17 +63,32 @@ find_for_vhost(VHost, Node) ->
...
+              VHostSup,
+              {rabbit_amqqueue_sup_sup,
+               {rabbit_amqqueue_sup_sup, start_link, []},
+               transient, infinity, supervisor, [rabbit_amqqueue_sup_sup]});
+        %% we can get here if a vhost is created and immediately
+        %% deleted, e.g. some integration tests do it
+        {error, {no_such_vhost, VHost}} ->
+            rabbit_log:error("Failed to start a queue process supervisor for vhost ~s: vhost no longer exists!",
+                             [VHost]),
+            ok
```
hairyhum: Should there be error? `rabbit_amqqueque:recover/1` matches `_`, but it's still not right.

src/rabbit_variable_queue.erl
```diff
@@ -496,15 +496,26 @@ stop(VHost) ->
...
-    {ok, _} = rabbit_vhost_msg_store:start(VHost,
-                                           ?PERSISTENT_MSG_STORE,
-                                           Refs,
-                                           StartFunState),
-    rabbit_log:info("Message stores for vhost '~s' are started~n", [VHost]).
+    safely_start_msg_store(VHost, ?TRANSIENT_MSG_STORE, undefined, ?EMPTY_START_FUN_STATE),
+    safely_start_msg_store(VHost, ?PERSISTENT_MSG_STORE, Refs, StartFunState),
+    ok.
+
+safely_start_msg_store(VHost, Type, Refs, StartFunState) ->
```
hairyhum: Is it OK to call a function, that an exit `safely_...`?

src/rabbit_vhost_msg_store.erl
```diff
@@ -23,17 +23,33 @@
...
-         transient, ?WORKER_WAIT, worker, [rabbit_msg_store]}).
+    case rabbit_vhost_sup_sup:vhost_sup(VHost) of
+        {ok, VHostSup} ->
+            VHostDir = rabbit_vhost:msg_store_dir_path(VHost),
+            supervisor2:start_child(VHostSup,
+                                    {Type, {rabbit_msg_store, start_link,
+                                            [Type, VHostDir, ClientRefs, StartupFunState]},
+                                     transient, ?WORKER_WAIT, worker, [rabbit_msg_store]});
+        %% we can get here if a vhost is created and immediately
+        %% deleted, e.g. some integration tests do it
```
hairyhum: We can get here only if a vhost is added and deleted concurrently.

src/rabbit_vhost_sup_sup.erl
```diff
@@ -83,7 +82,10 @@ start_vhost(VHost) ->
                     case supervisor2:start_child(?MODULE, [VHost]) of
                         {ok, _}                       -> ok;
                         {error, {already_started, _}} -> ok;
-                        Error                         -> throw(Error)
+                        Error                         ->
+                            rabbit_log:error("Could not start process tree "
+                                             "for vhost '~s': the vhost no longer exists!", [VHost]),
```
hairyhum: Do we know that vhost no longer exists at this point?

src/rabbit_vhost_sup_wrapper.erl
```diff
@@ -14,6 +14,12 @@
 %% Copyright (c) 2017 Pivotal Software, Inc.  All rights reserved.
 %%
 
+%% This module is a wrapper around vhost supervisor to
+%% provide exactly once restart semantics.
+%%
+%% rabbit_vhost_sup supervisor children are added dynamically,
+%% and the one_for_all strategy would be questionable here
```
hairyhum: It would be questionable in `rabbit_vhost_sup`, but not here.

hairyhum:
Few more questions. Hope this review will not be lost.
Another thing to be tested is how resilient the `transient` strategy to data corruption.

michaelklishin:
@hairyhum do you have any particular scenarios in mind?

hairyhum:
@michaelklishin when vhost data directory disappears, unaccessible or when disk is full.

michaelklishin:
I conducted a few tests per @hairyhum's suggestion.

Here's one:

 * Start a node
 * Create 3 vhosts, say, `vhost11`, `vhost12`, `vhost13`
 * Run a seed script [similar to this one](https://gist.github.com/michaelklishin/d912e3de212df170098a21adfdeec82c)
 * Observe that messages are there using `rabbitmqctl list_queues -p vhost11` and so on
 * Delete vhost data directory (`rabbit_vhost:delete_storage/1` is a convenient way of doing it)
 * Re-run the seed script
 * Observe that messages are there
 * Restart the node
 * Observe that messages are there

The test can be modified to delete data directory manually, it still recovers.

michaelklishin:
Another test: similar to the above but the script is modified to produce a constant load.

Outcome: the vhost is intact, queue processes terminate and do not always recover (which is expected and beyond the scope of this PR); newly declared queues are functional.

src/rabbit_amqqueue_sup_sup.erl
```diff
@@ -63,17 +63,32 @@ find_for_vhost(VHost, Node) ->
         Result -> {error, {queue_supervisor_not_found, Result}}
     end.
 
--spec start_for_vhost(rabbit_types:vhost()) -> {ok, pid()} | {error, term()}.
+-spec start_for_vhost(rabbit_types:vhost()) -> {ok, pid()} | ok | {error, term()}.
```
hairyhum: The function cannot return `ok` anymore

src/rabbit_amqqueue_sup_sup.erl
```diff
@@ -63,17 +63,32 @@ find_for_vhost(VHost, Node) ->
...
-         {rabbit_amqqueue_sup_sup, start_link, []},
-         transient, infinity, supervisor, [rabbit_amqqueue_sup_sup]}).
+    case rabbit_vhost_sup_sup:vhost_sup(VHost) of
+        {ok, VHostSup} ->
+            supervisor2:start_child(
+              VHostSup,
+              {rabbit_amqqueue_sup_sup,
+               {rabbit_amqqueue_sup_sup, start_link, []},
+               transient, infinity, supervisor, [rabbit_amqqueue_sup_sup]});
+        %% we can get here if a vhost is created and immediately
```
hairyhum: Should it be
```
%% we can get here if a vhost is added and removed concurrently
%% e.g. some integration tests do it
```
?

src/rabbit_recovery_terms.erl
```diff
@@ -48,20 +48,35 @@
...
-         [?MODULE]}),
+    case rabbit_vhost_sup_sup:vhost_sup(VHost) of
+      {ok, VHostSup} ->
+            {ok, _} = supervisor2:start_child(
+                        VHostSup,
+                        {?MODULE,
+                         {?MODULE, start_link, [VHost]},
+                         transient, ?WORKER_WAIT, worker,
+                         [?MODULE]});
+        %% we can get here if a vhost is created and immediately
```
hairyhum: Same as in `rabbit_amqqueue_sup_sup`

michaelklishin:
@hairyhum addressed.

PR #1293, opened 2017-07-18
Close connections when vhost is unavailable (supervision tree is down)

hairyhum:
Follow up to #1158

When vhost is down, we close all the connections to the vhost on the node.
When a client tries to connect to the vhost on the node - connection is refused with `internal_error`.

There are changes in supervision tree, which makes sure that vhost restarts are performed in order without concurrent message store processes running on the same vhosts.

Recovery is performed in the `vhost_process` process, which is former `vhost_sup_watcher`.
On termination the vhost process will notify that vhost is down, so the vhost connections can be closed.

A vhost "aliveness" status can be checked by checking aliveness of the vhost process.

### DIFF
--- a/src/rabbit_vhost_process.erl
+++ b/src/rabbit_vhost_process.erl
@@ -14,10 +14,21 @@
 %% Copyright (c) 2017 Pivotal Software, Inc.  All rights reserved.
 %%
 
-%% This module implements a watcher process which should stop
-%% the parent supervisor if its vhost is missing from the mnesia DB
+%% This module implements a vhost identity process.
 
--module(rabbit_vhost_sup_watcher).
+%% On start this process will try to recover the vhost data and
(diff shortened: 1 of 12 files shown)


src/rabbit_vhost_process.erl
```diff
@@ -14,10 +14,21 @@
...
+%% processes structure (queues and message stores).
+%% If recovered successfully, the process will save it's PID
+%% to vhost process registry. If vhost process PID is in the registry and the
+%% process is alive - the vhost is considered running.
+
+%% On termination, the ptocess will notify of vhost going down.
+
+%% The process will also check periodically if the vhost still
+%% present in mnesia DB and stop the vhost supervision tree when it
+%% desapears.
```
kjnilsson: spelling

PR #1309, opened 2017-07-26
Set queue state to 'stopped' when terminating.

### LINKED ISSUE #1303: Queues stopped by a vhost supervisor are not shown in list_queues.
When vhost message store is failing and the vhost supervisor is stopped, this vhost queues are still `live`. But since their PIDs are not running it's impossible to list them using `rabbitmqctl list_queues`.
`rabbitmqctl list_queues --offline` only lists the queues that are not in `rabbit_amqqueue`
`rabbitmqctl list_queues` lists crashed queues only if their state is `crashed`

hairyhum:
If a queue process is stopped by a vhost supervision tree, it should be visible in the management UI and the `list_queues` command output. 
Setting the queue state to `stopped` is easier to reason about than checking the vhost aliveness status on a remote node. 
Should queue be restarted or migrated to a different node, the status will be set to `live`. 
If a node was not stopped normally the queues will have `live` state. If we cannot recover the queues they should be marked `stopped` to be reported to rabbitmqctl and management UI 

Fixes #1303 
[#148409695]


src/rabbit_amqqueue.erl
```diff
@@ -452,6 +462,12 @@ with(Name, F, E, RetriesLeft) ->
             E({absent, Q, timeout});
         {ok, Q = #amqqueue{state = crashed}} ->
             E({absent, Q, crashed});
+        {ok, Q = #amqqueue{state = stopped}} ->
```
hairyhum: Right. We should not.

PR #1316, opened 2017-07-31
Start queue mirrors after vhost recovery instead of node start

### LINKED ISSUE #1314: Queue mirror is not added when vhost is restarted
When vhost supervisor is restarted, HA queue slaves for the vhost are not started.
Both when master was on a different node, or when it's migrated to a different node during the restart.

hairyhum:
Vhost supervisors can crash and restart without crashing the
node, so the slave queues on this vhosts should be started
after the vhsot recovery instead of node boot process.

Fixes #1314
[#149484151]

### DIFF
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -168,12 +168,6 @@
                     {requires,    recovery},
                     {enables,     routing_ready}]}).
 
--rabbit_boot_step({mirrored_queues,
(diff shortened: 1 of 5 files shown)


hairyhum:
Test failures should be fixed

lukebakken:
FWIW, tests still pass

works-on-my-machine-badge-small

PR #1341, opened 2017-08-23
Do not GC channel-queue metrics on mirror migration

### LINKED ISSUE #1340: GC of HA queue metrics causes negative delivery and ack rates
The change in https://github.com/rabbitmq/rabbitmq-server/pull/1278 introduced the garbage collection of queue data on migration of HA queues. This change forces to also GC the channel-queue stats, which should remain on the node holding the channel.

Consumers and producers can connect to any node on a RMQ cluster, which might not have mirror replicas on every single node. If the connections are to one of the node not holding a mirror, channel-queue stats are still generated that should be maintained on the origin node until the channel is closed.

The GC of these causes the negative rates reported in https://groups.google.com/forum/#!topic/rabbitmq-users/_wjzUhcKzxY

dcorbacho:
Closes #1340

### DIFF
--- a/src/rabbit_core_metrics_gc.erl
+++ b/src/rabbit_core_metrics_gc.erl
@@ -70,10 +70,17 @@ gc_channels() ->
     ok.
 
 gc_queues() ->
+    gc_local_queues(),
+    gc_global_queues().
+
+gc_local_queues() ->
     Queues = rabbit_amqqueue:list_local_names(),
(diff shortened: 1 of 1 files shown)


michaelklishin:
Tested this locally and deployed into our long-running environment. So far looking promising, no negative stats in over 1 hour (we had a setup that reproduced it every 2 minutes — so, after each stats GC ru — before).

PR #1366, opened 2017-09-19
Sync rabbitmq.config.example with PROJECT_ENV

gerhard:
I want to reconcile the existing discrepancies between rabbitmq.config.example & [PROJECT_ENV](https://github.com/rabbitmq/rabbitmq-server/blob/stable/Makefile#L10-L115). I want to review all config properties and make sure that they are all present and accurate in this canonical rabbitmq.config

Let's discuss which changes need to happen where so that rabbitmq.config.example is accurate & in sync with PROJECT_ENV

### DIFF
--- a/docs/rabbitmq.config.example
+++ b/docs/rabbitmq.config.example
@@ -424,7 +424,7 @@
    %% Whether or not to enable background GC.
    %%
    %% {background_gc_enabled, false},
-   %%
+
    %% Interval (in milliseconds) at which we run background GC.
    %%
    %% {background_gc_target_interval, 60000},
--- a/docs/rabbitmq.config.example
+++ b/docs/rabbitmq.config.example
@@ -310,6 +310,12 @@
    %%
    %% {mirroring_sync_batch_size, 4096},
 
+   %% Controls flow control between queue mirrors
+   %% Be careful with this configuration, it's controversial
+   %% https://github.com/rabbitmq/rabbitmq-server/pull/133
+   %%
+   %% {mirroring_flow_control, true},
+
    %% To announce custom properties to clients on connection:
    %%
    %% {server_properties, []},
@@ -406,6 +412,14 @@
    %% Number of lazy queue operations required to trigger an explicit garbage collection
    %%
    %% {lazy_queue_explicit_gc_run_operation_threshold, 1000},
+
+   %% Number of times to retry querying disk space usage
+   %%
+   %% {disk_monitor_failure_retries, 10},
+
+   %% Milliseconds to wait for disk space usage values
+   %%
+   %% {disk_monitor_failure_retry_interval, 120000},
 
    %% Whether or not to enable background GC.
    %%
--- a/docs/rabbitmq.config.example
+++ b/docs/rabbitmq.config.example
@@ -52,7 +52,7 @@
    %% connection. Hostnames will then be shown instead of IP addresses
    %% in rabbitmqctl and the management plugin.
    %%
-   %% {reverse_dns_lookups, true},
+   %% {reverse_dns_lookups, false},
 
    %%
    %% Security / AAA
@@ -124,11 +124,16 @@
    %%
    %% To use the SSL cert's CN instead of its DN as the username
    %%
-   %% {ssl_cert_login_from, common_name},
+   %% {ssl_cert_login_from, distinguished_name},
 
    %% SSL handshake timeout, in milliseconds.
    %%
    %% {ssl_handshake_timeout, 5000},
+
+   %% Whether to allow SSH POODLE attack,
+   %% as is the case of Erlang versions that make it impossible to disable SSLv3
+   %%
+   %% {ssl_allow_poodle_attack, false},
 
    %% Password hashing implementation. Will only affect newly
    %% created users. To recalculate hash for an existing user
@@ -200,7 +205,7 @@
    %% Set the max permissible number of channels per connection.
    %% 0 means "no limit".
    %%
-   %% {channel_max, 128},
+   %% {channel_max, 0},
 
    %% Customising Socket Options.
    %%
@@ -335,9 +340,13 @@
    %%
    %% {collect_statistics_interval, 5000},
 
+   %% Enables vhosts tracing.
+   %%
+   %% {trace_vhosts, []},
+
    %% Explicitly enable/disable HiPE compilation.
    %%
-   %% {hipe_compile, true},
+   %% {hipe_compile, false},
 
    %% Number of delegate processes to use for intra-cluster communication.
    %% On a machine which has a very large number of cores and is also part of a cluster,
@@ -644,8 +653,12 @@
 
    %% TCP/Socket options (as per the broker configuration).
    %%
-   %% {tcp_listen_options, [{backlog,   128},
-   %%                       {nodelay,   true}]}
+   %% {tcp_listen_options, [
+   %%                         {backlog,   128},
+   %%                         {nodelay,       true},
+   %%                         {linger,        {true, 0}},
+   %%                         {exit_on_close, false}
+   %%                      ]},
   ]},
 
  %% ----------------------------------------------------------------------------
--- a/docs/rabbitmq.config.example
+++ b/docs/rabbitmq.config.example
@@ -321,7 +321,7 @@
    %% {server_properties, []},
 
    %% How to respond to cluster partitions.
-   %% See http://www.rabbitmq.com/partitions.html for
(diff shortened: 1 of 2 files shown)


gerhard:
Once the missing properties are added to rabbitmq.config.example, I was thinking that we should order all config properties alphabetically. It's the ordering that requires the least cognitive overhead, both when adding or looking for properties. What do y'all think?

docs/rabbitmq.config.example
```diff
@@ -338,6 +372,26 @@
    %%
    %% {queue_index_embed_msgs_below, 4096},
 
+   %% The credits that a queue process is given by the message store.
```
michaelklishin: These paragraphs belong to a guide on rabbitmq.com. Comments to configuration keys can be brief descriptions with a link to more detailed docs.

michaelklishin: Here is [a good candidate](http://www.rabbitmq.com/persistence-conf.html). Might need some rearranging and editing but it tries to cover the same set of settings.

gerhard: I believe that contextual information is highly valuable when dealing with configuration properties. I agree that guides are essential for explaining how multiple configuration properties work together, but I strongly believe that we should keep essential details alongside configuration properties, in the config file context, without requiring users to jump out.

When I look at a canonical config example, I'm thinking:

1. What configs exist?
1. What do they mean?
1. Why would I want to change a config?

[The Cassandra config file example](https://github.com/apache/cassandra/blob/trunk/conf/cassandra.yaml) captures my sentiment well, [as does this less-known nginx config](https://github.com/gchef/nginx-cookbook/blob/master/attributes/default.rb).

I would really like to know what others think re the above.

Given the big picture, I will reduce the scope of this PR and just sync [PROJECT_ENV](https://github.com/rabbitmq/rabbitmq-server/blob/stable/Makefile#L10-L115) with `rabbitmq.config.example`. Single lines are in order, especially since I am yet to understand what every configuration property does.

docs/rabbitmq.config.example
```diff
@@ -338,6 +372,26 @@
...
+   %%
+   %% By default, a queue process is given 4000 message store credits,
+   %% and then 800 for every 800 messages that it processes.
+   %%
+   %% A queue process persists AMQP messages by sending them 1 by 1 to the message store process.
+   %% If the queue process sent 4000 AMQP messages and the message store process didn't manage to persist 800,
```
hairyhum: It does not have to persist messages, just process them. Credit flow ack is sent before message store handles a message, it will take some time to actually persist messages.

docs/rabbitmq.config.example
michaelklishin: @gerhard Cassandra's config does not explain what Hinted Handoff is or why it exists in several paragraphs, it says "See http://wiki.apache.org/cassandra/HintedHandoff". A few of sentences per key makes sense but config examples should not end up being small guides. There are 2 config examples in master, would you duplicate all those "micro guides"? That's my point.

hairyhum: There are number of guides already and most of config variables are documented in those guides, so we can keep doing that and add links here.

gerhard:
This is now ready to review & merge

docs/rabbitmq.config.example
```diff
@@ -288,19 +297,37 @@
...
+   %%
+   %% {queue_master_locator, <<"client-local">>},
+
+   %% Batch size of messages to synchronise between queue mirrors
+   %% See https://www.rabbitmq.com/ha.html#batch-sync
+   %%
+   %% {mirroring_sync_batch_size, 4096},
+
+   %% Controls flow control between queue mirrors
+   %% Be careful with this configuration, it's controversial
```
michaelklishin: It is not controversial and disabling it is dangerous (as is not having any flow control) with well know outcomes: masters outpacing mirrors at some point, making them consume more and more RAM.

docs/rabbitmq.config.example
```diff
@@ -319,9 +346,19 @@
...
+   %%
+   %% {trace_vhosts, []},
+
+   %% Explicitly enable/disable HiPE compilation.
    %%
-   %% {hipe_compile, true},
+   %% {hipe_compile, false},
+
+   %% Number of delegate processes to use for intra-cluster communication.
+   %% On a machine which has a very large number of cores and is also part of a cluster,
```
michaelklishin: How large is "very large"? I'd say "more than 16 cores and plenty of bandwidth".

docs/rabbitmq.config.example
```diff
@@ -333,18 +370,76 @@
...
+   %%
+   %% This value MUST be higher than the initial msg_store_credit_disc_bound value,
+   %% otherwise paging performance may worsen.
+   %%
+   %% {msg_store_io_batch_size, 4096},
+
+   %% Number of credits that a connection, channel or queue are given
+   %% By default, every connection, channel or queue is given 400 credits,
+   %% and then 200 for every 200 messages that it processes
+   %% See https://www.rabbitmq.com/blog/2015/10/06/new-credit-flow-settings-on-rabbitmq-3-5-5/
```
michaelklishin: We should mention that increasing these values may help with throughput but is dangerous (very high values are no different from having any flow control).

docs/rabbitmq.config.example
```diff
@@ -572,8 +667,12 @@
 
    %% TCP/Socket options (as per the broker configuration).
    %%
-   %% {tcp_listen_options, [{backlog,   128},
-   %%                       {nodelay,   true}]}
+   %% {tcp_listen_options, [
+   %%                         {backlog,   128},
+   %%                         {nodelay,       true},
```
michaelklishin: `nodelay` is forced by Ranch even if you disable it. I see no reason to do that, by the way.

docs/rabbitmq.config.example
```diff
@@ -572,8 +667,12 @@
 
    %% TCP/Socket options (as per the broker configuration).
    %%
-   %% {tcp_listen_options, [{backlog,   128},
-   %%                       {nodelay,   true}]}
+   %% {tcp_listen_options, [
```
michaelklishin: Any reason why there is no link to http://www.rabbitmq.com/networking.html here? It's a pretty extensive guide.

docs/rabbitmq.config.example
```diff
@@ -333,18 +370,76 @@
...
    %%
    %% {background_gc_enabled, false},
-   %%
+
    %% Interval (in milliseconds) at which we run background GC.
    %%
-   %% {background_gc_target_interval, 60000}
+   %% {background_gc_target_interval, 60000},
+
+   %% File size in bytes used for persisting messages
```
michaelklishin: Please change it to "Message store operations are stored in a sequence of files called segments. This controls max size of a segment file. Increasing this value may speed up (sequential) disk writes but will slow down segment GC process. DO NOT CHANGE THIS for existing installations."

docs/rabbitmq.config.example
```diff
@@ -333,18 +370,76 @@
...
+
+   %% Number of lazy queue operations required to trigger an explicit garbage collection
+   %%
+   %% {lazy_queue_explicit_gc_run_operation_threshold, 1000},
+
+   %% Number of times to retry querying disk space usage
+   %%
+   %% {disk_monitor_failure_retries, 10},
+
+   %% Milliseconds to wait for disk space usage values
```
michaelklishin: This is an interval of time disk monitor waits before retrying, not a query timeout.

docs/rabbitmq.config.example
```diff
@@ -120,12 +124,17 @@
...
    %% To use the SSL cert's CN instead of its DN as the username
    %%
-   %% {ssl_cert_login_from, common_name},
+   %% {ssl_cert_login_from, distinguished_name},
 
    %% SSL handshake timeout, in milliseconds.
    %%
    %% {ssl_handshake_timeout, 5000},
 
+   %% Whether to allow SSL POODLE attack,
```
michaelklishin: This makes it sound like anyone wants to enable an attack.

"Makes RabbitMQ accept SSLv3 client connections by default. DO NOT DO THIS IF YOU CAN HELP IT."

docs/rabbitmq.config.example
gerhard: Should `nodelay` be removed from [PROJECT_ENV](https://github.com/rabbitmq/rabbitmq-server/blob/stable/Makefile#L60)?

gerhard:
@michaelklishin how about now?

PR #1374, opened 2017-09-26
An option to discard messages if queue is full.

### LINKED ISSUE #995: Nack messages that cannot be deposited to all queues due to max length reached
Under publisher confirms, messages are nacked when the broker could not process them. This issue requests to optionally treat "at least one queue is full" as one of the reasons when the broker could not process a message, instead of discarding old messages, which is the current behavior that may remain the default. 

See also: https://groups.google.com/forum/#!topic/rabbitmq-users/Fzsdx1CEiWA

hairyhum:
If a queue is to be overflowed by a delivery it can reject
the delivery or drop messages from the head.
To reject delivery overflow can be configured to `reject_publish`,
to drop head it's `drop_head` (default setting).

Messages which will be rejected should still confirm being routed,
so mandatory expectations are not accumulated on the channel side.

Slave nodes will only confirm if a message was published or discarded.
To drop confirms from slaves, all confirms for a message are cleared
when the message is rejected.

When promoting a new master, left-behind deliveries should
be rejected if the queue is full, just like normal deliveries.

Fixes #995
[#151294447]


src/rabbit_policies.erl
```diff
@@ -104,7 +105,13 @@ validate_policy0(<<"queue-mode">>, <<"default">>) ->
 validate_policy0(<<"queue-mode">>, <<"lazy">>) ->
     ok;
 validate_policy0(<<"queue-mode">>, Value) ->
-    {error, "~p is not a valid queue-mode value", [Value]}.
+    {error, "~p is not a valid queue-mode value", [Value]};
+validate_policy0(<<"overflow">>, <<"drop_head">>) ->
```
lukebakken: The other policy names use dash characters instead of underscores. Should these be `drop-head` and `reject-publish` to be consistent? They could still be used as atoms elsewhere if single-quoted.

hairyhum: Good find. Will change.

src/rabbit_mirror_queue_slave.erl
```diff
@@ -302,6 +302,7 @@ handle_cast({deliver, Delivery = #delivery{sender = Sender, flow = Flow}, true},
     %% We are acking messages to the channel process that sent us
     %% the message delivery. See
     %% rabbit_amqqueue_process:handle_ch_down for more info.
+    %% TODO: reject publishes
```
lukebakken: Just to confirm ... this is for future work and not necessary here?

src/rabbit_amqqueue_process.erl
```diff
@@ -421,6 +426,12 @@ init_max_bytes(MaxBytes, State) ->
     {_Dropped, State1} = maybe_drop_head(State#q{max_bytes = MaxBytes}),
     State1.
 
+init_overflow(undefined, State) ->
+    State;
+init_overflow(Overflow, State) ->
+    %% TODO maybe drop head
```
lukebakken: This is for future work and not necessary here?

src/rabbit_channel.erl
```diff
@@ -1874,21 +1895,24 @@ send_nacks(_MXs, State = #ch{state = closing}) -> %% optimisation
...
+        ok      -> ConfirmMsgSeqNos =
                        lists:foldl(
                          fun ({MsgSeqNo, XName}, MSNs) ->
                                  ?INCR_STATS(exchange_stats, XName, 1,
                                              confirm, State),
                                  [MsgSeqNo | MSNs]
                          end, [], lists:append(C)),
-                   send_confirms(MsgSeqNos, State#ch{confirmed = []});
+                   State1 = send_confirms(ConfirmMsgSeqNos, State#ch{confirmed = []}),
+                   %% TODO: msg seq nos, same as for confirms. Need to implement
```
dcorbacho: is this a future change or pending for this PR?

test/unit_inbroker_parallel_SUITE.erl
```diff
@@ -1018,6 +1041,161 @@ set_vm_memory_high_watermark_command1(_Config) ->
...
+    {_Conn, Ch} = rabbit_ct_client_helpers:open_connection_and_channel(Config, 0),
+    QName = <<"max_length_drop_head_queue">>,
+    QNameDefault = <<"max_length_default_drop_head_queue">>,
+    QNameBytes = <<"max_length_bytes_drop_head_queue">>,
+    QNameDefaultBytes = <<"max_length_bytes_default_drop_head_queue">>,
+
+    MaxLengthArgs = [{<<"x-max-length">>, long, 1}],
+    MaxLengthBytesArgs = [{<<"x-max-length-bytes">>, long, 100}],
+    OverflowArgs = [{<<"x-overflow">>, longstr, <<"drop-head">>}],
+    amqp_channel:call(Ch, #'queue.delete'{queue = QName}),
```
dcorbacho: If cleanup is properly implemented, these `queue.delete` shouldn't be needed at the beginning of the test. Or am I missing something? I think instead `end_per_group` these changes require `end_per_testcase`.

test/unit_inbroker_parallel_SUITE.erl
```diff
@@ -1018,6 +1041,161 @@ set_vm_memory_high_watermark_command1(_Config) ->
...
+    check_max_length_rejects(Config, QNameBytes, Ch, Payload1, Payload2, Payload3).
+
+max_length_drop_publish(Config) ->
+    {_Conn, Ch} = rabbit_ct_client_helpers:open_connection_and_channel(Config, 0),
+    QName = <<"max_length_drop_publish_queue">>,
+    QNameBytes = <<"max_length_bytes_drop_publish_queue">>,
+    MaxLengthArgs = [{<<"x-max-length">>, long, 1}],
+    MaxLengthBytesArgs = [{<<"x-max-length-bytes">>, long, 100}],
+    OverflowArgs = [{<<"x-overflow">>, longstr, <<"reject-publish">>}],
+    amqp_channel:call(Ch, #'queue.delete'{queue = QName}),
```
dcorbacho: Same here

src/rabbit_channel.erl
hairyhum: We have no `nack` rates. So it's a future change. I'll create a new issue.

michaelklishin:
Measuring throughput impact of this (in both overflow modes) is what's left to QA.

dcorbacho:
@michaelklishin This PR doesn't introduce any (negative) impact on the throughput, in fact the speed up is over 3x! See below, first rate line belongs to a queue with `drop-head` and a length of 300. Second line is a queue with `reject-publish` and length of 300.
screen shot 2017-11-13 at 10 46 21 am

michaelklishin:
Initial set of local tests do not demonstrate a meaningful deviation from master. We are going to set up a long running environment to see if this is true for runs that span hours/days in a more realistic deployment environment.

gerhard:
Having run this PR against 3.7.0-rc.1 for [ndq-1kb-autoack](https://github.com/rabbitmq/use-cases/tree/master/ndq-1kb-autoack), we can see an 18% improvement (50660 msg/s vs 42930 msg/s) when looking at the average throughput over 4 hours (purple is this PR, blue is 3.7.0-rc.1):

screen shot 2017-11-15 at 14 44 05

uvzubovs:
Thank you so much for getting this done. Is this available for testing?

michaelklishin:
@uvzubovs will be in 3.7.0-rc.2 later this week.

PR #1384, opened 2017-10-04
Integrate looking_glass

gerhard:
[An Erlang/Elixir/BEAM profiler tool](https://github.com/rabbitmq/looking_glass)

Signed-off-by: Loïc Hoguin <loic@rabbitmq.com>

### DIFF
--- a/src/rabbit_looking_glass.erl
+++ b/src/rabbit_looking_glass.erl
@@ -1,0 +1,46 @@
+%% The contents of this file are subject to the Mozilla Public License
+%% Version 1.1 (the "License"); you may not use this file except in
+%% compliance with the License. You may obtain a copy of the License
+%% at http://www.mozilla.org/MPL/
+%%
+%% Software distributed under the License is distributed on an "AS IS"
+%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
+%% the License for the specific language governing rights and
+%% limitations under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% The Initial Developer of the Original Code is GoPivotal, Inc.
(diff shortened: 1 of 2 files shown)


src/rabbit_looking_glass.erl
```diff
@@ -0,0 +1,46 @@
...
+
+boot() ->
+    case os:getenv("RABBITMQ_TRACER") of
+        false ->
+            ok;
+        Value ->
+            Input = parse_value(Value),
+            rabbit_log:info("Enabling Looking Glass profiler, input value: ~p", [Input]),
+            {ok, _} = application:ensure_all_started(looking_glass),
+            lg:trace(Input, lg_file_tracer, "traces.lz4", #{
```
dumbbell: Maps are not supported by Erlang R16B03 so this won't compile at all.

src/rabbit_looking_glass.erl
```diff
@@ -0,0 +1,55 @@
...
+            rabbit_log:info(
+                "Enabling Looking Glass profiler, input value: ~p",
+                [Input]
+            ),
+            {ok, _} = application:ensure_all_started(looking_glass),
+            lg:trace(
+                Input,
+                lg_file_tracer,
+                "traces.lz4",
+                maps:from_list([
```
dumbbell: The `maps` module isn't available in Erlang R16B03 either:
```
$ gmake xref
(...)
src/rabbit_looking_glass.erl:24 `rabbit_looking_glass:boot/0` calls undefined function `maps:from_list/1`
maps:from_list/1` is not defined as a function
gmake: *** [erlang.mk:6755: xref] Error 1
```

You could add another `-ignore_xref()` directive as we won't run this code on R16B03 anyway.

PR #1388, opened 2017-10-10
Flush messages to disk in batches.

hairyhum:
If messages should be embedded to a queue index, there will
be no credit flow limit, so message batches can be too big
and block the queue process.
Limiting the batch size allows consumer to make progress while
publishers are blocked by the paging-out process.

[#151614048]

### DIFF
--- a/src/rabbit_variable_queue.erl
+++ b/src/rabbit_variable_queue.erl
@@ -2363,45 +2363,79 @@ reduce_memory_use(State = #vqstate {
                                                 out     = AvgEgress,
                                                 ack_in  = AvgAckIngress,
                                                 ack_out = AvgAckEgress } }) ->
-    State1 = #vqstate { q2 = Q2, q3 = Q3 } =
+    {CreditDiscBound, _} =rabbit_misc:get_env(rabbit,
+                                              msg_store_credit_disc_bound,
+                                              ?CREDIT_DISC_BOUND),
+    {NeedResumeA2B, State1} = {_, #vqstate { q2 = Q2, q3 = Q3 }} =
         case chunk_size(RamMsgCount + gb_trees:size(RPA), TargetRamCount) of
-            0  -> State;
+            0  -> {false, State};
             %% Reduce memory of pending acks and alphas. The order is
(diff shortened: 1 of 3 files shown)


src/rabbit_variable_queue.erl
```diff
@@ -2363,45 +2363,79 @@ reduce_memory_use(State = #vqstate {
...
+        {false, false} -> ok;
+        %% Planning next batch
+        {false, true}  ->
+            %% We don't want to use self-credit-flow, because it's harder to
+            %% reason about. So the process sends a (prioritised) message to
+            %% itself and sets a waiting_bump value to keep the message box clean
+            case get(waiting_bump) of
+                true -> ok;
+                _    -> self() ! bump_reduce_memory_use,
+                        put(waiting_bump, waiting)
```
lukebakken: Instead of `waiting`, I think the value here should be `true` otherwise the skip-case won't be hit on future calls that are waiting for the message. Also see #1393

hairyhum: Right. Missed that.

michaelklishin:
Here are some (fairly basic) benchmark results in a constraint environment, both with and without consumer rate limiting.

When consumer throughput is really constraint, we see that publishers are throttled more aggressively but consumers are never completely blocked — stdev for consumer latency is significantly lower on this branch.

On other workloads the difference is fairly small but surprisingly this branch demonstrates a slight higher overall throughput (which I wasn't expecting) and lower consumer latency (which was easier to foresee).

[perf_test_runs_batched_vs_stable.zip](https://github.com/rabbitmq/rabbitmq-server/files/1402148/perf_test_runs_batched_vs_stable.zip)

gerhard:
Great improvement!

Page out behaviour before this change:
page-out-1-3 6 12
page-out-2-3 6 12

Page out behaviour after this change:
page-out-3 6 13-dev 1388

This is what used to happen when filling & draining a queue (no paging to disk):
fill-3 6 13-dev 1388
drain-3 6 12

This is what happens now when filling & draining a queue (no paging to disk):
fill-3 6 12
drain-3 6 13-dev 1388

Issue #1216, opened 2017-05-09
Consider making `rabbit` application transient.


hairyhum:
OTP applications support `restart_type` option, which defines a VM behaviour when application is stopped. It can be `permanent`, `temporary` or `transient` http://erlang.org/doc/apps/kernel/application.html#start-2

Since the `rabbit` application is essential in most of deployments, it makes sense to make it `transient` by default, so the VM will be stopped when the application fails abnormally. That would be much less confusing for an operator, than an erlang VM running without RabbitMQ.

The current `temporary` restart type can be made optional.

michaelklishin:
@binarin FYI, I hope it doesn't go against your embedding efforts.

PR #1406, opened 2017-10-26
Start `rabbit` application as temporary in tests.

hairyhum:
Some tests output can break when `rabbit` application failure
stops the node. Make it possible to configure restart type.
This is a part of a backport of https://github.com/rabbitmq/rabbitmq-server/issues/1216


gerhard:
Without this, killing rabbit_sup app results in 500 errors on the Management UI. It also leaves the Erlang node running, heart-beating as if nothing happened. This definitely addresses the issues that we've been seeing in the wild.

michaelklishin:
@hairyhum @gerhard `gmake ct-clustering_management` fails for me on this branch.

PR #1407, opened 2017-10-26
Re-introduce consumer bias & set to 2.0

hairyhum:
Revert #1378 because it leads to performance degradation for publishers.

Set consumer bias to 2.0 to prioritise consumers.


gerhard:
Following #1378, we have observed that when using PerfTest, always prioritising consumers results in publishers being blocked most of the time. This is less than ideal. We have settled on making consumers 2x as fast as publishers.

## No consumer bias

no-consumer-bias-1

no-consumer-bias-2

## 2x consumer bias

2x-consumer-bias-1

2x-consumer-bias-2

PR #1413, opened 2017-10-30
Make rabbit_networking:close_connection/2 more benign

michaelklishin:
## Proposed Changes

The function relies on a fanout to all running cluster nodes
which in turn check their local connection process group.
All errors that might arise are ignored.

There are at least two scenarios where the current behavior makes
little sense:

 * A connection process has just terminated (but is still present in the stats DB)
 * A node could not respond (or respond in time) to the fanout

In both cases claiming that connection pid is "not a connection pid"
is misleading and likely will leave the user puzzled. This change
makes connection_close/2 more benign: it simply logs a warning and returns
instead of throwing.

There are only two call sites that use it: rabbitmqctl and an HTTP API
handler, so the chances of non-connection pids passed in are pretty slim.

While at it, only log that a connection was closed when it actually was.

Per discussion with @gerhard.## Types of Changes

## Checklist

### DIFF
--- a/src/rabbit_networking.erl
+++ b/src/rabbit_networking.erl
@@ -305,10 +305,16 @@ connection_info_all(Items, Ref, AggregatorPid) ->
       connections()).
 
 close_connection(Pid, Explanation) ->
-    rabbit_log:info("Closing connection ~p because ~p~n", [Pid, Explanation]),
     case lists:member(Pid, connections()) of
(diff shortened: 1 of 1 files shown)


michaelklishin:
Note: `test/close_connection_command_test.exs` in rabbitmq-cli must be updated if/when this is merged into `master`.

gerhard:
🎉

PR #1419, opened 2017-11-03
DRAFT Set fullsweep_after to 0 for queue processes.

hairyhum:
The setting tends to increase performance of a queue in
publish mode. Although can use more memory if queue contains
more messages when performing GC.

Need to investigate performance impact of this change on other workloads.


michaelklishin:
The title says "set to 0" but the code sets to 10. So what's the target value at the moment?

dumbbell:
Is this pull request still relevant?

michaelklishin:
It is, it's an unfinished experiment (spike).

PR #1436, opened 2017-11-29
ID processes in Looking Glass and trace more by default

essen:
Companion PR: https://github.com/rabbitmq/rabbitmq-common/pull/241

This is mostly a cosmetic change for Looking Glass output, processes will now be identified as "reader", "channel" and so on and that makes for a lot more comfortable reading of some of these outputs. This should have zero impact when tracing is not enabled.

Tracing all modules by default is a better idea because Erlang will do a lot of operations via the erlang and a few other modules and we generally want to have those in the output for them to make sense. It also makes the output make more sense when other modules like lists are involved in multiple calls in one function.

Finally the process_dump option is necessary for flame graph output and it doesn't hurt much to have it by default here since this code is only used for development. HOWEVER please run the following test. Yesterday I had the surprise that enabling this option made messages slow to a crawl (4 msgs/s). I couldn't figure it out. Today the issue resolved itself for reasons I do *not* understand. It probably was a temporary problem in my environment, but just in case you should run the following and make sure it doesn't happen elsewhere:

- Apply this and the rabbit_common patch
- Run the broker in rabbit `cd deps/rabbit; make run-broker`
- Use perftest to check the message rate `bin/runjava com.rabbitmq.perf.PerfTest`

If you get a few thousands, good! If you get, like, 4, not good.


pivotal-issuemaster:
@essen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-server&pullRequestId=1436)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-server&pullRequestId=1436) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@essen Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-server&pullRequestId=1436)!

essen:
Done requested changes in other PR and rebased to current master.

PR #1466, opened 2018-01-05
Internal authN backend: make it impossible to successfully log in with a blank password (for 3.6.x)

michaelklishin:
## Proposed Changes

This disables logins with a blank **provided** password in the internal [authN backend](http://rabbitmq.com/access-control.html). There should be no reason to use passwordless users
and password-based authentication (the PLAIN authentication mechanism). Blank passwords
are only useful when they are, ahem, not used because authentication is handled out of band, e.g. [via x509 certificates](http://github.com/rabbitmq/rabbitmq-auth-mechanism-ssl).

It is still possible to create a user with a blank password or clear a password on an existing user.
When the EXTERNAL authentication mechanism is used, the internal backend is not involved at all.

Note that this change is limited to the internal backend only. External backends such as LDAP
are considered to not be controlled by the broker.

## Types of Changes

- [x] Bug fix
- [x] Documentation (correction or otherwise): https://github.com/rabbitmq/rabbitmq-website/pull/472

## Checklist

## Further Comments

[#153435857]

### DIFF
--- a/src/rabbit_auth_backend_internal.erl
+++ b/src/rabbit_auth_backend_internal.erl
@@ -94,9 +94,9 @@ hashing_module_for_user(#internal_user{
         rabbit_password:hashing_mod(ModOrUndefined).
 
 -define(BLANK_PASSWORD_REJECTION_MESSAGE,
-         "user '~s' attempted to log in with a blank password, which is prohibited by the internal authN backend. "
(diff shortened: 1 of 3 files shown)


michaelklishin:
The message will be a single line in the log even if it's a multi-line Erlang string. In other words, they are actually concatenated by `erlc`.

PR #1478, opened 2018-01-22
Declare a quorum queue using the queue.declare method

dcorbacho:
## Proposed Changes

Part of the new high availability implementation using raft as a consensus algorithm

Uses the x-queue-type argument to introduce the type: quorum/classic
Defaults to classic

## Types of Changes

## Checklist


michaelklishin:
This is a 4.0-specific change, which means we either have to roll it back or create a branch for 3.8 and reconfigure Concourse and so on.

PR #1504, opened 2018-02-11
rabbit_router:match_bindings/2 performance improvement

sylvainhubsch:
This is a performance improvement only; it directly impacts modules using rabbit_router:match_bindings/2 (such as rabbit_exchange_type_headers and rabbitmq_jms_topic_exchange found in umbrella repo).

TL;DR
router_match_bindings_optimize
(data come from rabbitmq-perf-test; no headers in messages sent so that there is no routing process that could disturb measures; example of config file used : https://gist.github.com/sylvainhubsch/9aef5c0214e600acb630772390c14d36)

That PR is the first one to bring rabbit_exchange_type_headers really useful; however, if more details are needed I will do my best to explain how I get those results.

### DIFF
--- a/src/rabbit_router.erl
+++ b/src/rabbit_router.erl
@@ -36,17 +36,14 @@
 
 %%----------------------------------------------------------------------------
 
-%% TODO: Maybe this should be handled by a cursor instead.
-%% TODO: This causes a full scan for each entry with the same source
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you! We'll backport once this change passes our entire pipeline successfully.

michaelklishin:
Backported, will be included into `3.7.4`.

PR #1520, opened 2018-02-20
DO NOT MERGE YET Avoid O(n^2) complexity on acknowledgment

acogoluegnes:


src/rabbit_queue_consumers.erl
```diff
@@ -179,7 +182,7 @@ erase_ch(ChPid, State = #state{consumers = Consumers}) ->
             All = priority_queue:join(Consumers, BlockedQ),
             ok = erase_ch_record(C),
             Filtered = priority_queue:filter(chan_pred(ChPid, true), All),
-            {[AckTag || {AckTag, _CTag} <- queue:to_list(ChAckTags)],
+            {[AckTag || {AckTag, _CTag} <- maps:to_list(ChAckTags)],
```
michaelklishin: This has to traverse the entire collection. If we can avoid this, it should make a difference.

src/rabbit_queue_consumers.erl
```diff
@@ -245,16 +249,22 @@ deliver_to_consumer(FetchFun,
...
 record_ack(ChPid, LimiterPid, AckTag) ->
     C = #cr{acktags = ChAckTags} = ch_record(ChPid, LimiterPid),
-    update_ch_record(C#cr{acktags = queue:in({AckTag, none}, ChAckTags)}),
+    assert_not_in_map(AckTag, ChAckTags),
+    update_ch_record(C#cr{acktags = ChAckTags#{AckTag => none}}),
     ok.
 
+assert_not_in_map(Key, Map) ->
+    case maps:is_key(Key, Map) of
+        true -> error("ack map already contains key");
```
michaelklishin: We should mention what key that is for our own debugging sanity.

src/rabbit_queue_consumers.erl
```diff
@@ -270,6 +280,57 @@ subtract_acks(ChPid, AckTags, State) ->
...
+% subtract_acks_impl removes the acks specified in the first parameter,
+% the second and third paramters should be empty array and empty map
+% the fourth paramter is the ack queue
+% it returns {map(CTag => Count), AckQueue}
+subtract_acks_impl([], CTagCounts, AckQ) ->
+    {CTagCounts, AckQ};
+subtract_acks_impl([T | TL], CTagCounts, AckQ) ->
+    case maps:take(T, AckQ) of
+      {CTag, AckQ1} ->
+        subtract_acks_impl(TL, maps:update_with(CTag, fun (Old) -> Old + 1 end, 1, CTagCounts), AckQ1);
```
michaelklishin: Do we need `maps:update_with/3` here? Looks like we can safe a function call by using `maps:update/3` since we're only incrementing a counter.

michaelklishin:
It would be interesting to see some profiling data. Ack subtraction search/sorting and conversion to lists seem to be the most costly operations after a very quick look (this is a guess, of course).

michaelklishin:
We concluded that without switching to a different set of data structures we cannot address this without seriously affecting other code paths. Closing as we won't be looking into this in the medium term.

PR #1524, opened 2018-02-22
Support kernel.net_ticktime in Cuttlefish configuration (master)

### LINKED ISSUE #1522: kernel.net_ticktime is not recognized in rabbitmq.conf (sysctl format)
* RabbitMQ version: 3.7.3
* Docker image: [3.7.3](https://github.com/docker-library/rabbitmq/blob/b5543df2789c16d8ea1c2119484378d4c23baa6c/3.7/debian/Dockerfile)

`kernel.net_ticktime` is not recognized as a valid option when it is set in rabbitmq.conf ([sysctl format](https://github.com/rabbitmq/rabbitmq-server/blob/v3.7.3/docs/rabbitmq.conf.example#L455)):
* rabbitmq.conf:
```
kernel.net_ticktime = 60
```
* RabbitMQ server log:
```
Config file generation failed 09:25:07.272 [error] You've tried to set kernel.net_ticktime, but there is no setting with that name.
09:25:07.272 [error]   Did you mean one of these?
09:25:07.310 [error]     collect_statistics
09:25:07.310 [error]     connection_max
09:25:07.310 [error]     default_vhost
09:25:07.310 [error] Error generating configuration in phase transform_datatypes
09:25:07.310 [error] Conf file attempted to set unknown variable: kernel.net_ticktime
BOOT FAILED
===========
Error description:
    rabbit:start_it/1 line 448
    rabbit:boot_error/2 line 848
    rabbit_lager:log_locations/0 line 59
    rabbit_lager:ensure_lager_configured/0 line 160
    rabbit_lager:lager_configured/0 line 167
    lager:list_all_sinks/0 line 317
    lager_config:get/2 line 71
    ets:lookup(lager_config, {'_global',handlers})
exit:generate_config_file
Log file(s) (may contain more information):
{"init terminating in do_boot",generate_config_file}
init terminating in do_boot (generate_config_file)

Crash dump is being written to: /var/log/rabbitmq/erl_crash.dump...done
```
It is recognized when it is set in advanced.config (Erlang format):
* advanced.config:
```
[{kernel, [{net_ticktime, 60}]}].
```

acogoluegnes:
`net_ticktime`, if present in the `.conf` file, is set up with `net_kernel:set_net_ticktime/1` before the configuration of the `.conf` file is applied to the other applications. The `kernel` application is already running when the configuration is applied, so net_ticktime is set in a specific way.

Fixes #1522

### DIFF
--- a/docs/rabbitmq.conf.example
+++ b/docs/rabbitmq.conf.example
@@ -459,7 +459,7 @@
 # Kernel section
 # ======================================
 
-# kernel.net_ticktime = 60
+# net_ticktime = 60
 
 ## ----------------------------------------------------------------------------
 ## RabbitMQ Management Plugin
--- a/priv/schema/rabbit.schema
+++ b/priv/schema/rabbit.schema
@@ -1150,6 +1150,14 @@ end}.
     {datatype, {enum, [debug, info, notice, warning, error, critical, alert, emergency, none]}}
 ]}.
 
+% ==========================
+% Kernel section
+% ==========================
+
+{mapping, "net_ticktime", "kernel.net_ticktime",[
+    {datatype, [integer]}
+]}.
+
 % ===============================
 % Validators
 % ===============================
--- a/src/rabbit_config.erl
+++ b/src/rabbit_config.erl
@@ -72,13 +72,10 @@ update_app_config(ConfigFile) ->
     %% For application config to be updated, applications should
     %% be unloaded first.
     %% If an application is already running, print an error.
-    lists:foreach(fun({App, _Config}) ->
+    lists:foreach(fun({App, AppConfig}) ->
         case lists:member(App, RunningApps) of
             true ->
-                io:format(standard_error,
-                          "~nUnable to update config for app ~p from *.conf file."
-                          " App is already running. Use advanced.config instead.~n",
-                          [App]);
+                maybe_print_warning_for_running_app(App, AppConfig);
             false ->
                 case lists:member(App, LoadedApps) of
                     true  -> application:unload(App);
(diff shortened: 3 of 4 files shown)


michaelklishin:
I discovered the following during manual testing: since the transition period by default takes 60 seconds, should a CLI tool contact a running node with `net_ticktime` in the `.conf` file with this PR, it will fail with something like this:

```
Stopping and halting node rabbit@warp10 ...
Error:
function_clause
 Stacktrace [{net_kernel,set_net_ticktime,
                         [{ongoing_change_to,20},0],
                         [{file,"net_kernel.erl"},{line,195}]},
             {rabbit_misc,rpc_call,5,
                          [{file,"src/rabbit_misc.erl"},{line,1175}]},
             {'Elixir.RabbitMQCtl',maybe_run_command,3,
                                   [{file,"lib/rabbitmqctl.ex"},{line,113}]},
             {'Elixir.RabbitMQCtl','-exec_command/2-fun-0-',5,
                                   [{file,"lib/rabbitmqctl.ex"},{line,80}]},
             {'Elixir.RabbitMQCtl',main,1,
                                   [{file,"lib/rabbitmqctl.ex"},{line,43}]},
             {'Elixir.Kernel.CLI','-exec_fun/2-fun-0-',3,
                                  [{file,"lib/kernel/cli.ex"},{line,76}]}]
```

because CLI tools temporarily tweak node's net_ticktime for the two values to be in sync.

We can try a transition period of 0 but I don't know what complications can stem out of that. I don't think the change is worth the risk since it's a minor convenience improvement.

michaelklishin:
A correction: it's not just CLI tools, it's any code that uses [rabbit_misc:rpc_call/5](https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_misc.erl#L11720) but it should not be used on any hot code path.

Updating that function to support transition periods should be sufficient and won't affect many cases.

michaelklishin:
According to https://github.com/rabbitmq/rabbitmq-common/pull/60, that function is used by management and possibly other plugins.

PR #1556, opened 2018-03-15
Do not try to start a vhost supervisors on not fully booted nodes.

hairyhum:
Sometimes when several nodes are started at the same time, add_vhost
can try to start a remote vhost supervisor on a node, which does
not have a rabbit_vhost_sup_sup process yet, resulting in `{error,rabbit_vhost_sup_sup_not_running}`
error.

Filter only fully booted nodes to start remote vhost supervisors.

### DIFF
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -27,7 +27,7 @@
 -export([start/2, stop/1, prep_stop/1]).
 -export([start_apps/1, start_apps/2, stop_apps/1]).
 -export([log_locations/0, config_files/0, decrypt_config/2]). %% for testing and mgmt-agent
--export([is_booted/0, is_booted/1, is_booting/0]).
+-export([is_booted/1]).
 
 -ifdef(TEST).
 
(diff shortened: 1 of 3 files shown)


src/rabbit_vhost_sup_sup.erl
```diff
@@ -61,7 +61,9 @@ init([]) ->
 
 start_on_all_nodes(VHost) ->
     NodesStart = [ {Node, start_vhost(VHost, Node)}
-                   || Node <- rabbit_nodes:all_running() ],
+                   || Node <- rabbit_nodes:all_running(),
+                      %% Do not try to start a vhost on booting nodes.
```
michaelklishin: This must include our own node, however.

michaelklishin:
@hairyhum please review my changes from this morning.

src/rabbit.erl
```diff
@@ -745,6 +752,16 @@ is_running() -> is_running(node()).
 
 is_running(Node) -> rabbit_nodes:is_process_running(Node, rabbit).
 
+is_booted(Node) ->
+    case rpc:call(Node, rabbit, is_booted, []) of
```
michaelklishin: Technically this won't be available on pre-3.7.4 nodes. But given that vhosts are not created during rolling upgrades in practice, this might be OK.

hairyhum: Right. We can check for processes directly instead.

PR #1549, opened 2018-03-14
Force-delete queues that have no promotable master (or candidates)

hairyhum:
Fixes #1501
[#155801556]
If a queue is configured to not be promoted (via ha-promote-on-shutdown: when-synced)
queue.delete can hang. Make it check for process existense first and
force-delete if no master of slave processes are running.
Do not force-delete if if_empty is set, since there is no
way to check that the queue is empty.

PR #1558, opened 2018-03-15
Delete dead queue 3.7

hairyhum:
#1549  for 3.7


michaelklishin:
Sorry, we now cherry-pick from master into v3.7.x instead of submitting separate PRs. I'm almost done with that.

PR #1567, opened 2018-03-28
Revert "Reduce lock contention when nodes restart (master)"

michaelklishin:
Reverts rabbitmq/rabbitmq-server#1527.

`ets:select_replace/2` is not available in OTP 19.3 😢.


gerhard:
Great catch, we totally missed this. Back at it next Tuesday ⛑

PR #1578, opened 2018-04-18
Policy key to not promote unsynchronised queues.

hairyhum:
This new policy controls if unsynchronised slaves should be promoted
after master crash. If set to `when-synced`, unsynchronised slaves
will not be promoted, keeping the state of the queue, but making it
unavailable until master node returns.

This change is supposed to make the cluster shutdown safier,
because queues can fail or be killed on shutdown.

The queues without master will be available from the management UI
and can be deleted and redeclared, but will not automatically loose
messages.

Trying to declare or passively declare the queue will result in a
timeout error. Same way as if the master was gracefully stopped with
ha-promote-on-shutdown: when-synced

This setting probably can be backported to 3.6. 

[#156811690]

### DIFF
--- a/src/rabbit_mirror_queue_coordinator.erl
+++ b/src/rabbit_mirror_queue_coordinator.erl
@@ -365,7 +365,12 @@ handle_cast({gm_deaths, DeadGMPids},
                                                    DeadPids),
             {stop, shutdown, State};
         {error, not_found} ->
-            {stop, normal, State}
+            {stop, normal, State};
+        {error, {not_synced, _}} ->
+            rabbit_log:error("Mirror queue ~p in unexpected state after gm_deaths."
+                             " Promoted to master but already a master.",
+                             [QueueName]),
+            error(unexpected_mirrored_state)
     end;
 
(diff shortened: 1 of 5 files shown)


src/rabbit_mirror_queue_coordinator.erl
```diff
@@ -365,7 +365,12 @@ handle_cast({gm_deaths, DeadGMPids},
                                                    DeadPids),
             {stop, shutdown, State};
         {error, not_found} ->
-            {stop, normal, State}
+            {stop, normal, State};
+        {error, {not_synced, _}} ->
+            rabbit_log:error("Mirror queue ~p in unexpected state after gm_deaths."
```
michaelklishin: This error message is too much inside baseball.

hairyhum: Can you clarify?

michaelklishin: "gm_deaths" is an implementation detail that makes no sense to the user.

PR #1597, opened 2018-05-12
Hard cap for maximum priorities

### LINKED ISSUE #1590: Hard cap for max queue priorities
We had an issue report today that took a while to figure out but it came down to an overflow/underflow kind of error that resulted in an `x-max-priorities` argument to have the value of `2,147,483,647`. Needless to say that completely deprived the node of RAM and scheduler resources.

We should introduce a hard cap for max priorities, which I think would make sense even for a future implementation that doesn't use one backing queue/process per priority.

The value can be configurable (it won't be used on the hot code path and no default will be right for everyone). I suggest 100 as the initial value (and even that is generous).

Per discussion with @jerrykuch.

michaelklishin:
## Proposed Changes

This [re]introduces a hard cap on the maximum number of priorities supported by priority queues. See #1590 for details.

It also makes it possible to define priorities using a policy.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #1590)

## Checklist

## Further Comments

Closes #1590.

### DIFF
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -613,9 +613,9 @@ check_message_ttl_arg({Type, Val}, Args) ->
 
 check_max_priority_arg({Type, Val}, Args) ->
     case check_non_neg_int_arg({Type, Val}, Args) of
-        ok when Val =< 255 -> ok;
-        ok                 -> {error, {max_value_exceeded, Val}};
-        Error              -> Error
+        ok when Val =< ?MAX_SUPPORTED_PRIORITY -> ok;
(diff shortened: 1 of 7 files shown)


michaelklishin:
How to QA priority definition via policies (a drive-by change in this feature):

 * Set up a policy
 * Declare a matching queue
 * Publish 200 messages with random priorities (I used 1 through 5)
 * Add a consumer that prints message priority and observe that they are consumed in priority order

Java client tests PR focuses on the limits and should be self-explanatory.

michaelklishin:
We found out why policies were not supported earlier: they can change while certain message store aspects of priority queues are not (must be fixed). So I will undo some portions of this PR and docs.

PR #1601, opened 2018-05-18
Add client properties to connection.closed events

### LINKED ISSUE #1596: Add user-provided connection name to connection.closed events
That can be useful, in particular since connection process PIDs can't/shouldn't be used by non-BEAM programming languages.

See [this rabbitmq-users thread](https://groups.google.com/d/msg/rabbitmq-users/QHcZpMeTe4E/l2999FW8BAAJ).

If no user-provided connection name is available, the key should probably be excluded: this would make most sense to rabbitmq-event-exchange consumers.

acogoluegnes:
This commit adds client properties to connection.closed events. The
original need was to have only the optional user-provided connection
name added to correlate connections between created and closed events,
but all the client properties are finally added for the sake of
consistency between the 2 events.

This commit uses the process dictionary to convey the client properties,
as they're not yet available in the connection state when the call to
send the closed event is made (in the after block, just after the
network connection has been established).

Fixes #1596

[#157500358]

image

image

### DIFF
--- a/src/rabbit_reader.erl
+++ b/src/rabbit_reader.erl
@@ -134,7 +134,7 @@
         peer_cert_validity, auth_mechanism, ssl_protocol,
         ssl_key_exchange, ssl_cipher, ssl_hash, protocol, user, vhost,
         timeout, frame_max, channel_max, client_properties, connected_at,
-        node, user_who_performed_action, connection_user_provided_name]).
+        node, user_who_performed_action]).
 
 -define(INFO_KEYS, ?CREATION_EVENT_KEYS ++ ?STATISTICS_KEYS -- [pid]).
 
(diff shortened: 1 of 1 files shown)


src/rabbit_reader.erl
```diff
@@ -387,10 +387,17 @@ start_connection(Parent, HelperSup, Deb, Sock) ->
         %% socket w/o delay before termination.
         rabbit_net:fast_close(RealSocket),
         rabbit_networking:unregister_connection(self()),
-	rabbit_core_metrics:connection_closed(self()),
+        rabbit_core_metrics:connection_closed(self()),
+        ClientProperties = case get(client_properties) of
```
lukebakken: Rather than use the process dictionary, could we instead use [`sys:get_status(self())`](http://erlang.org/doc/man/sys.html#get_status-1) to get the current process state and extract `client_properties` from that?

hairyhum: `sys:get_status/1` will send a system call message to the process. This doesn't work with `self()`

acogoluegnes:
With the new `connection_user_provided_name`:

image

image

acogoluegnes:
Key is now `user_provided_name` and only added if necessary:

image

image

src/rabbit_reader.erl
lukebakken: Thanks for the explanation!

PR #1604, opened 2018-05-21
Use more optimal default memory allocators where available (Erlang 20.2.3 or later)

essen:
## Proposed Changes

See [this rabbitmq-users thread](https://groups.google.com/d/msg/rabbitmq-users/LSYaac9frYw/LNZDZUlrBAAJ) for background.

We have settled on the following configuration for memory
allocators after testing many different combinations:

+MBas ageffcbf +MHas ageffcbf +MBlmbcs 512 +MHlmbcs 512 +MMmcs 30

They reduce the memory usage and help RabbitMQ reclaim memory,
at the cost of a slight decrease in performance due to an
increased number of memory operations.

We need to start Erlang with these values in order to figure
out whether they are supported. The allocator strategies we
recommend were introduced in Erlang/OTP 20.2.3. The values
can be overriden using RABBITMQ_SERVER_ERL_ARGS.

cc @gerhard

## Types of Changes

- [x] Optimization (peak and median RAM consumption)

## Checklist

### DIFF
--- a/scripts/rabbitmq-server
+++ b/scripts/rabbitmq-server
@@ -120,6 +120,25 @@ if [ ! -f "${RABBITMQ_SCHEMA_DIR}/rabbit.schema" ]; then
     cp "${RABBITMQ_HOME}/priv/schema/rabbit.schema" "${RABBITMQ_SCHEMA_DIR}"
 fi
 
+# The default allocation strategy RabbitMQ is using was introduced
+# in Erlang/OTP 20.2.3. Earlier Erlang versions fail to start with
+# this configuration. We therefore need to ensure that erl accepts
+# these values before we can use them.
+#
+# The defaults are meant to reduce RabbitMQ's memory usage and help
+# it reclaim memory at the cost of a slight decrease in performance
+# (due to an increase in memory operations). These defaults can be
+# overriden using the RABBITMQ_SERVER_ERL_ARGS variable.
+RABBITMQ_DEFAULT_ALLOC_ARGS="+MBas ageffcbf +MHas ageffcbf +MBlmbcs 512 +MHlmbcs 512 +MMmcs 30"
+
+${ERL_DIR}erl ${RABBITMQ_DEFAULT_ALLOC_ARGS} \
+    -boot "${CLEAN_BOOT_FILE}" \
+    -noinput -eval 'halt(0)' 2>/dev/null
+
+if [ $? != 0 ] ; then
+    RABBITMQ_DEFAULT_ALLOC_ARGS=
+fi
+
 set -e
 
 RABBITMQ_CONFIG_FILE_NOEX="${RABBITMQ_CONFIG_FILE%.*}"
@@ -213,6 +232,7 @@ start_rabbitmq_server() {
         ${RABBITMQ_CONFIG_ARG} \
         +W w \
         +A ${RABBITMQ_IO_THREAD_POOL_SIZE} \
+        ${RABBITMQ_DEFAULT_ALLOC_ARGS} \
         ${RABBITMQ_SERVER_ERL_ARGS} \
         +K true \
         -kernel inet_default_connect_options "[{nodelay,true}]" \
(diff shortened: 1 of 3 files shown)


essen:
@dumbbell Please give it a small look and tell me if we're going the wrong way about this.

I also have not been able to test on Windows (I wrote a small script testing for memory allocators, tested it and then incorporated it into rabbitmq-server.bat which I then didn't test, I'm hoping you have a CI environment for Windows).

michaelklishin:
I tested it on 19.3 and 20.3, it works as expected. The effects of these allocators flag on actual workloads needs more research and user feedback.

dumbbell:
The patch looks good and yes, we have Windows VMs in CI. However, I need to verify which version of Erlang we use there.

dumbbell:
We use Erlang 20.0 on all Windows VMs. I need to figure out a way to test with at least two versions of Erlang, like we do for Linux.

dumbbell:
My concern is mostly how I integrate that into Concourse configuration :-) Otherwise, we use a new Windows VM each time we do a Windows installer test and throw it away, so installing Erlang is not an issue.

hairyhum:
Should it also be in `rabbitmq-service.bat`?

gerhard:
Started testing this against various workloads, all having the following in common:

* [n1-highcpu-8](https://cloud.google.com/compute/pricing)
* Ubuntu 14.04.5 LTS  4.4.0-124-generic x86_64
* Erlang/OTP 20.3.6
* RabbitMQ v3.7.5
* perf-test v2.1.2
* 5 minute averages

## [ndq-autoack](https://github.com/rabbitmq/workloads/tree/master/ndq-autoack)

|  | Before | After |
|-|-|-|
| Messages/s | 55k | 56k |
| Memory | 81 MiB | 81 MiB |
| GC/s | 27K | 33K |

screen shot 2018-05-21 at 18 39 49 3

## [ndq-multiack](https://github.com/rabbitmq/workloads/tree/master/ndq-multiack)

## [ndq-10q-multiack](https://github.com/rabbitmq/workloads/tree/master/ndq-10q-multiack)

## [dq-3n-confirm-multiack](https://github.com/rabbitmq/workloads/tree/master/dq-3n-confirm-multiack)
 
## [lqs-ha-confirm-multiack ](https://github.com/rabbitmq/workloads/tree/master/lqs-ha-confirm-multiack )

essen:
I have added the changes to `rabbitmq-service.bat`.

Should I work on a separate PR to include it in 3.6.16?

michaelklishin:
@essen if we are certain those things are safe for R16B03 and 19.3 users (yes, we technically still allow running on R16B03, even though the docs say it is unsupported), please do. I produced an RC of 3.6.16 earlier today but we can include this into RC2.

gerhard:
Even though we haven't benchmarked the impact of this on durable, mirrored or lazy queues, are the above benchmarks against non-durable queues sufficient to make a call whether this change is a good default?

michaelklishin:
We haven't identified any meaningful (statistically significant) regressions with these new defaults and there's some empirical evidence on the mailing list that the new defaults make a big difference for some users. So I think we should adopt these changes.

essen:
Woo!!

gerhard:
Are we planning to merge this into 3.7.6?

How about 3.6.16? #1612

michaelklishin:
It was cherry-picked to 3.7.6. We haven't done any 3.6.x testing yet. Also, our 3.7.x LRE runs OTP 20.1.7, so it would be cool to have that upgraded to 20.3.6 before 3.7.6 GA.

gerhard:
I can look into that on Monday, the LRE deployments are very outdated at this point. Does anyone want to join the fun or should I Hans Solo it?

michaelklishin:
If you bring your Millennium Falcon someone might join you :)

gerhard:
I was wondering whether it would be worth mentioning in the release notes how to go back to the previous defaults, `+MBas aoffcbf +MHas aoffcbf +MBlmbcs 5120 +MHlmbcs 5120 +MMmcs 10`

michaelklishin:
@gerhard good idea.

michaelklishin:
#1612 was merged, currently a backport of this PR is on track to be included into 3.6.16.

PR #1612, opened 2018-05-29
Use better default memory allocators

essen:
The v3.6.x equivalent of https://github.com/rabbitmq/rabbitmq-server/pull/1604


essen:
This was tested against R16B03-1 on Linux only.

PR #1649, opened 2018-07-23
erlang/otp 21 deprecated erlang:get_stacktrace/0 function, use the fu…

getong:
work with rabbitmq/rabbitmq-common#271
copy from [meck](https://github.com/eproxus/meck/blob/master/src/meck_code_gen.erl#L186-L192)

### DIFF
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -456,8 +456,8 @@ start_it(StartFun) ->
                         false -> StartFun()
                     end
                 catch
-                    Class:Reason ->
-                        boot_error(Class, Reason)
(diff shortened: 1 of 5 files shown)


michaelklishin:
Thank you for your time but our team has a different plan. As of https://github.com/rabbitmq/rabbitmq-server/pull/1615/files#diff-f8aa00d902505f5bd2fd015ff0e32444R19 we silenced the warning and by the time 3.8.0 goes GA we will require Erlang 21 and switch to the new syntax. The proposed change doesn't seem to be necessary, or at least I'm not sure what build environment would need it given that https://github.com/rabbitmq/rabbitmq-server/pull/1615 and similar PRs are in place.

Issue #1648, opened 2018-07-23
Concurrent virtual host state recovery


michaelklishin:
Currently in environments with a lot of virtual hosts their recovery can take a long time, even if each has a moderate amount of data to recover.

This can be attributed to 3 things:

 * As of https://github.com/rabbitmq/rabbitmq-server/issues/567 each virtual host has separate messages stores to recover.
 * As of https://github.com/rabbitmq/rabbitmq-server/pull/1608 virtual host startup is synchronised with its counterparts on all cluster nodes.
 * Virtual host recovery on an individual node [is performed sequentially](https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_vhost.erl#L59).

The first two items above are by design and their penalty is a trade-off we are willing to accept. The sequential recovery part is not really a design decision and can be revisited.

This issue is about investigating concurrent virtual host recovery and its "unknown unknowns". Just like with other areas of node data recovery we want to limit the concurrency rate, e.g. by using a separate fixed size work pool.

There are currently two steps to recovery: "init", which interacts with the virtual host process supervisor (and possibly cannot be done concurrently) and "recover", which does most of the work (and is closer in what we already do concurrently to recover queue state). Making the latter step concurrent is our current area of focus.

Per discussion with @kjnilsson.

kjnilsson:
An alternative to changing the `init` sequence is to use a permanent  and configurable pool of vhost supervisors instead. Then almost all code can stay the same apart from some additional logic to choose a suitable pool member for a newly created vhost and such.

michaelklishin:
#1650 makes this problem significantly less severe (at least on a few workloads we tested with).

mkuratczyk:
2025 update: empty vhost recovery performance is mostly limited by mnesia transcations. Using Khepri, the problem pretty much goes away. Here are the numbers for [1000 vhosts](https://github.com/rabbitmq/sample-configs/blob/main/vhosts/1000-vhosts.json) on a single node:

Mnesia:
* `rabbitmqctl import_definitions`: 185s
* `rabbitmqctl stop_app`: 5s
* `rabbitmqctl start_app`: 360s

Khepri:
* `rabbitmqctl import_definitions`: 22s
* `rabbitmqctl stop_app`: 3s
* `rabbitmqctl start_app`: 19s

Issue #567, opened 2016-01-20
Per-vhost message store


michaelklishin:
This idea has been floating around since at least late 2012: we should use separate message stores for different vhosts, e.g. two (for persistent, transient messages) per vhost.

Some of the benefits are:

 * Better isolation of physical resources between hosts (e.g. in the spirit of #498, #500, #501)
 * (Potentially) improved availability should a message store terminate or significantly fall behind
 * (At least potentially) Less contention, more parallelism
 * Opens the door to moving different vhosts to different disks

michaelklishin:
`effort-high` because migration of existing data can take a while to cover well with tests. Otherwise this is `effort-medium`.

hairyhum:
What if we enable per vhost message stores for new vhosts only, and recommend to shovel messages for migration?

michaelklishin:
I'm afraid that's not gonna fly with hosted RabbitMQ providers, and at least some Cloud Foundry users. It would be great to automatically migrate messages on boot (if we can).

hairyhum:
Initial research:
Thanks to `rabbit_msg_store` architecture enabling per-vhost grouping is pretty easy by starting message stores in subdirectories and controlling them with supervisor.  
Migration of old data can still be a problem, since there is no vhost information in message records.

hairyhum:
Migration mechanism proposal:
```
move_messages_to_vhost_store() ->
    Queues = get_durable_queues(),
    OldStore = run_old_persistent_store(),
    Migrations = spawn_for_each(fun(Queue) -> 
                                    migrate_queue(Queue, OldStore) 
                                end, Queues),
    wait_for_completion(Migrations),
    delete_old_store(OldStore).

migrate_queue(Queue, OldStore) ->
    OldStoreClient = get_client(OldStore),
    NewStore       = ensure_new_store_exist(get_vhost(Queue)),
    NewStoreClient = get_client(NewStore),
    walk_queue_index(
        fun(MessageIdInStore) ->
            Msg = get_msg_from_store(OldStoreClient),
            put_message_to_store(Msg, NewStoreClient)
        end,
        Queue).
```

michaelklishin:
@hairyhum given that upgrades perform a backup of the node data dir, that sounds reasonable.

PR #1650, opened 2018-07-24
Prepare semi-durable binding table once instead of each time a virtual host is recovered

hairyhum:
Before #567 all binding recover functions were executed once on
node restart.
Executing table_filter can be expensive with high number of vhosts
and bindings and effectively should be called only once.
Moved to a separate function, which should be called from the global
recovery function (`rabbit_vhost:recover/0`).

This problem exacerbated what's described in #1648.


hairyhum:
Didn't run all tests yet, but the behaviour should not be different.

PR #1677, opened 2018-08-17
WIP: Headers' type exchanges on steroids

sylvainhubsch:
AMQP headers' type exchanges are underused and barely unknown for the most people because most MQ systems routing logics are "topics ruled" only. Reality is that headers routing are much more powerful and may evolve without causing breaking changes. Until now, they were not really useful (equal comparison only) nor performing..

**So whats's new here :**
* new headers match operators : less than, more than or equal to, not equal to, on existing keys etc..
* using "list type" in binding rule to do some "is in" operators
* new routing facilities such as "when binding matches false, route to multi queues.." etc..
* improve performances when reading headers exchange bindings (see graph below)
* some other (little) benefits (by example you can now match keys header beginning with "x-")

--> improve your global process by moving some works made by consumers into some more specific routing logics in RabbitMQ
--> simplify your routing rules with less exchanges and less bindings
--> improve messages throughput

TL;DR about performances :
image

**Is this compatible with AMQP 0.9.1 protocol and is that PR compatible with my existing bindings ?**
Mainly yes.. as long as you don't specify explicitly some binding rules beginning by "x-?<= " or "x-?< " or "x-?= " or "x-?!= " or "x-?> " or "x-?>= " and so on... But why you would do that until now as those rules (all the ones beginning by "x-") are not taken into account ?!

**So, to be clear, is that kind of a new proposal for protocol "AMQP 0.9.2" ? Else, for now is that an RabbitMQ extension ?**
Yes, I would love that ! :)

**Is that PR really ready to merge ?**
Not really, there are 2 points missing I think :
--> **Please please please**, I DO ask community for some helps about writing tests... I know how to write tests (I write tests mainly for webservices those days....), but I know little about ct module (which seems to be amazing by the way) and how tests are managed/ruled by RabbitMQ in general. So, maybe one can help me to write the first test, I will contribute to them to test the whole new rules! Thanks in advance.
--> Documentation follows..... sorry about that..
(this PR depends too of those on rabbitmq_management and rabbit_common for the same branch..)

I think documentation had to be written before making this PR, but I wanted to know if my code was really bad before :)

## Types of Changes
- [X] Perfs <-- yes, this one is my add-on :)

## Checklist

## Further Comments
.. are welcome :)


sylvainhubsch:
The big change is that now a specific table (rabbit_headers_bindings) handles headers' type exchange bindings instead of the generic one (rabbit_route). When bindings are added, they are processed before being written so that we do not more job when routing message... Secondly, when a message goes in exchange, a single read is done to read all bindings related to the current exchange..
Those 2 facts make perfs better.

michaelklishin:
This looks promising, thank you. We would definitely consider such compatible (within reason) improvements to the protocol and the headers exchange certainly could use some. We'll get to review this some time next week. Thanks a lot!

michaelklishin:
If bindings are stored in a separate table that's by definition a major breaking change that cannot go into 3.7.x.

michaelklishin:
In fact, if durable bindings of headers exchanges are not migrated with an upgrade function this PR cannot be accepted even for 3.8.

sylvainhubsch:
Thanks Michael for this first feedback.
I thought tables' upgrade would be made via add_binding function, but this effectively can't be the case ! I will try to add upgrade funs for that point; may you consider this PR not to be rejected for now ?
Thanks again.

michaelklishin:
We are not going to close it. Simply add existing durable binding migration via an upgrade function that follows the one you've added and we'll continue with QA. It's perfectly fine for a PR to evolve a bit and go through several rounds of review. I will add a prefix our team usually uses for something that's known to be a WIP. Thanks.

sylvainhubsch:
I've just added function rabbit_exchange_type_headers:upgrade_headers_bindings/0 (which call add_binding/2 itself) but I don't know if I can call it from rabbit_upgrade_functions:headers_bindings/0... Actually I really don't know how to test my code by triggering some 'real' migration processes, that's so bad.. :/
Again, thanks for your help and your time.

michaelklishin:
@sylvainhubsch it should be a separate upgrade function that runs after the `headers_bindings` one. They way we test upgrade functions is usually

 * Start a GA release node with data directory at a certain path
 * Stop it
 * Start a node from source with `gmake run-broker RABBITMQ_ALLOW_INPUT=1 RABBITMQ_MNESIA_DIR=…` pointing at the same directory
 * Drop into a REPL if you have to or observe the migration any other way

We have separate upgrade tests but they operate at the package level. We can extend them to declare some headers exchanges with bindings. I don't think they are open source since our pipeline code isn't.

michaelklishin:
@sylvainhubsch is this ready for another round of QA or are more changes coming?

michaelklishin:
@sylvainhubsch ping. Should we expect more changes or this is ready for QA? Thank you.

sylvainhubsch:
@michaelklishin : I'm terribly sorry for the delay.. really.
We can cancel this merge request, I've worked on something more "useful" than just improve headers's exchange; I will give details later, this week I hope (the main idea is the same, but even better I think). I'm about to submit a totally new merge request that will be immediatly ready for QA.
Once again, sorry for the delay; I would like to have more time to contribute to RabbitMQ and MOM systems in general..
Thank you.

michaelklishin:
@sylvainhubsch no worries and thank you so much for your ongoing contributions!

Issue #1699, opened 2018-09-14
pg_local:member_died/1 should be more resilient


michaelklishin:
See [this rabbitmq-users thead](https://groups.google.com/forum/#!topic/rabbitmq-users/cjoNsDwv4Sc).

If due to any reason a bookkeeping table behind `pg_local` groups doesn't have a record, `pg_local:member_died/1` should be more resilient (a no-op) or use the pid information from the `DOWN` event that triggered it. Further cascading failures due to a `badmatch` won't be productive in any way.

PR #1700, opened 2018-09-15
Make pg_local:member_died/2 more resilient

michaelklishin:
## Proposed Changes

This makes `pg_local:member_died/2` more defensive and resilient, avoiding failed assertions where they aren't productive or helpful in troubleshooting.

See #1699 for background.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #1699)

## Checklist

## Further Comments

[#160530707]

### DIFF
--- a/src/pg_local.erl
+++ b/src/pg_local.erl
@@ -162,9 +162,7 @@ member_died(Ref, Pid) ->
         [{{ref, Ref}, Pid}] ->
             leave_all_groups(Pid);
         %% in case the key has already been removed
-        %% we can perform the lookup using the DOWN message pid
-        [] ->
-            leave_all_groups(Pid);
+        %% we can perform the lookup using the valuee from the DOWN message
         _  ->
             leave_all_groups(Pid)
     end,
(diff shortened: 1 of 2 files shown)


src/pg_local.erl
```diff
@@ -148,46 +157,57 @@ terminate(_Reason, _S) ->
 %%% {{pid, Pid, Name}}
 %%%    Pid is a member of group Name.
 
-member_died(Ref) ->
-    [{{ref, Ref}, Pid}] = ets:lookup(pg_local_table, {ref, Ref}),
+member_died(Ref, Pid) ->
+    case ets:lookup(?TABLE, {ref, Ref}) of
```
dcorbacho: Do we really need this lookup if we always call `leave_all_groups` with the same `Pid`? The three `case` clauses do exactly the same. Or am I missing something?

michaelklishin: This code is many years old so I can only guess the intent here. I think it tries to assert that the ref/pid combination is known. I'd leave it as is for now, this is not a hot code path event.

michaelklishin: The clauses simply demonstrate what 3 scenarios are possible. We can combine the latter two.

Issue #1690, opened 2018-08-30
Binding lookup before removal is very inefficient (requires a linear table scan)


michaelklishin:
See #1566 and #1589 for background.

The least efficient part of queue deletion is binding deletion because there is currently no way to load bindings of a queue (or exchange) without a full scan on one of the binding tables. This unfortunate design can be mitigated with secondary indices but this would be a breaking schema change and therefore can only go into 3.8.0.

hairyhum:
The main issue was not the linear scan, because match_object optimise on key scans, but the fact that entire table had to be locked on every deletion, which made concurrent queue deletion transactions block each other. 
Also queues without any bindings were bound to default exchange, those bindings were never read during routing (default exchange has it's own routing logic) and were slowing down the cleanup.
The #1715 PR contains some optimisations.

acogoluegnes:
This has been backported to 3.7.x: b2097218c564e859f7d957a0e869621d16200593.

PR #1715, opened 2018-09-28
Bindings optimisation

hairyhum:
A follow-up to #1589
Adresses #1690

A number of optimisations around bindings. Apparently there is not much sense in adding table indexes to binding tables, as index read locks entire table. Indexes could slightly improve performance when there are many bindings and queues, but will require a breaking change in the schema.

The changes in this branch provide performance improvement comparable to introduction of indexes without reworking the schema:

- Replaced `delete_object` with `delete` as bindings are value-objects and there are no bag tables
- In some lookups during cleanup `match_object` is replaced with `dirty_match_object` to not lock the entire bindings table. This change was reverted in d0423f958599cba5148864997aa8446c873dad00 because it was creating inconsistencies with `dirty_delete`. Current branch does not contain dirty deletes. Still requires more testing.
- Default bindings are removed. The default exchange does direct lookup of queues without actually loading the bindings and default bindings are not unbindable and only displayed in `list_bindings`. They are not used, but slow down queue deletion and creation.

Benchmark data will follow.

### DIFF
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -366,8 +366,7 @@ internal_declare(Q = #amqqueue{name = QueueName}, false) ->
                           not_found           -> Q1 = rabbit_policy:set(Q),
                                                  Q2 = Q1#amqqueue{state = live},
                                                  ok = store_queue(Q2),
-                                                 B = add_default_binding(Q1),
-                                                 fun () -> B(), Q1 end;
+                                                 fun () -> Q1 end;
                           {absent, _Q, _} = R -> rabbit_misc:const(R)
                       end;
                   [ExistingQ] ->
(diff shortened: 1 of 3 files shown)


hairyhum:
Exclusive queue cleanup benchmark:

Time to cleanup queues on connection:

| 100K queues | No explicit bindings | 1 binding | 2 bindings |
|--------------|---------------------|----------|------------|
| Before            | 57.7 sec                     | 77.9 sec  | 107.1 sec   |
| After               | 20.9 sec                    | 25.7 sec  | 30.5 sec   |

hairyhum:
This can be backported to 3.7 if accepted.

hairyhum:
The interesting thing about default bindings and min-master locator is that it's counting queue masters looking at bindings, but all queues were bound. Originally it was counting bindings and selecting a node which min bindings to the queues on the node. Then it was changed in https://github.com/rabbitmq/rabbitmq-server/commit/1d413eac80c2adb6844b82c8ce57e21a04a55b04 to count multiple bindings for the same queue as one. The assumption was that we should not account for unbound queues. But  any queue had at least 1 binding (default).
I'm going to make it count queues, because it will be the same behaviour.

hairyhum:
The test now passes.

hairyhum:
So default bindings are shown in the management API. I've missed that somehow. I still don't think we should store them in the database. Maybe we could fake them for the management API
I'm going to create a separate PR with default bindings removed.

hairyhum:
Moved default bindings removal to a separate PR since most of the failures are about that change. It might be even for 3.8.

michaelklishin:
@hairyhum we can append them in HTTP API handlers and `rabbitmqctl list_bindings` in CLI.

michaelklishin:
…or functions that list bindings, that'd cover more cases with a single change?

hairyhum:
Merged to 3.7

michaelklishin:
#1721, a spin-off from this issue, was merged for 3.8.

PR #1721, opened 2018-10-02
Remove default bindings

hairyhum:
Split from #1715

Default bindings slow down queue creation and deletion and are not used in the actual routing logic. They are returned in some HTTP API endpoints though, therefore to merge this PR there should be some changes done in the management plugin.

There is no information in bindings which cannot be extracted from queues, which makes it possible to fake results for HTTP API. The question is should we do that or make the bindings result empty if there are no explicit bindings?

The definitions result does not contain default bindings BTW.

### DIFF
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -29,7 +29,7 @@
 -export([list/0, list/1, info_keys/0, info/1, info/2, info_all/1, info_all/2,
          emit_info_all/5, list_local/1, info_local/1,
 	 emit_info_local/4, emit_info_down/4]).
--export([list_down/1, count/1, list_names/0, list_local_names/0]).
+-export([list_down/1, count/1, list_names/0, list_names/1, list_local_names/0]).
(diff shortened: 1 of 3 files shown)


hairyhum:
It seems like `rabbit_binding:list` is used in bindings load in both list_bindings and HTTP API. Also `rabbit_binding:list_for_source`, `rabbit_binding:list_for_destination` and `rabbit_binding:list_for_source_and_destination`. We can add fake bindings there for `amq.default` and queues.

michaelklishin:
So far I see two types of failures:

 * `GET /bindings/{vhost}/e/amq.default/q/{queue}/{key}` returns 404
 * Some tests expect names or resource names but seem to be getting full queue records/objects

hairyhum:
Fake bindings should have fixed that. I'll take a look.

michaelklishin:
This was backported to `v3.7.x` in a backwards compatible way plus converted to a feature flag for further mixed version cluster compatibility in master.

PR #1741, opened 2018-10-21
DO NOT MERGE Reply `reject_publish` when queue marks a message as duplicate

noxdafox:
The `rabbit_backing_queue` behaviour defines a `is_duplicate` callback
to signal if the message about to be published is a
duplicate. Nevertheless, the broker logic does not take any action if
so.

This becomes problematic if the client enables publisher confirmation
as neither a confirm nor a reject is provided as response in case
a message gets dropped as duplicate.

This patch ensures the client receives a rejection if a queue marks
the delivered message as duplicate.

Signed-off-by: Matteo Cafasso <noxdafox@gmail.com>

## Types of Changes

## Checklist

## Further Comments
Few users reported [issues](https://github.com/noxdafox/rabbitmq-message-deduplication/issues/21) when using the message deduplication plugin together with publisher confirmation.

It appears  the broker does not deliver any notification to the publisher when a backing queue `is_duplicate` callback returns `true`. This causes clients to hang indefinitely waiting for a confirmation/rejection to come.

This patch tackles the issue making sure a `nack` is delivered in case a `rabbit_backing_queue` implementation provides message deduplication.

To reproduce the issue, you can install the deduplication plugin from the [releases](https://github.com/noxdafox/rabbitmq-message-deduplication/releases) tab and follow the instruction provided in this [issue](https://github.com/noxdafox/rabbitmq-message-deduplication/issues/21).

### DIFF
--- a/src/rabbit_amqqueue_process.erl
+++ b/src/rabbit_amqqueue_process.erl
@@ -658,43 +658,48 @@ attempt_delivery(Delivery = #delivery{sender  = SenderPid,
                             State#q{consumers = Consumers})}
     end.
 
-maybe_deliver_or_enqueue(Delivery, Delivered, State = #q{overflow = Overflow}) ->
+maybe_deliver_or_enqueue(Delivery = #delivery{message = Message},
+                         Delivered,
+                         State = #q{overflow            = Overflow,
+                                    backing_queue       = BQ,
+                                    backing_queue_state = BQS}) ->
     send_mandatory(Delivery), %% must do this before confirms
-    case {will_overflow(Delivery, State), Overflow} of
-        {true, 'reject-publish'} ->
-            %% Drop publish and nack to publisher
-            send_reject_publish(Delivery, Delivered, State);
-        _ ->
-            %% Enqueue and maybe drop head later
-            deliver_or_enqueue(Delivery, Delivered, State)
+
+    {IsDuplicate, BQS1} = BQ:is_duplicate(Message, BQS),
+    State1 = State#q{backing_queue_state = BQS1},
+
+    if IsDuplicate -> send_reject_publish(Delivery, Delivered, State1);
+       true        -> case {will_overflow(Delivery, State1), Overflow} of
+                          {true, 'reject-publish'} ->
+                              %% Drop publish and nack to publisher
+                              send_reject_publish(Delivery, Delivered, State1);
+                          _ ->
+                              %% Enqueue and maybe drop head later
+                              deliver_or_enqueue(Delivery, Delivered, State1)
+                      end
     end.
 
 deliver_or_enqueue(Delivery = #delivery{message = Message,
                                         sender  = SenderPid,
                                         flow    = Flow},
-                   Delivered, State = #q{backing_queue       = BQ,
-                                         backing_queue_state = BQS}) ->
+                   Delivered, State = #q{backing_queue = BQ}) ->
     {Confirm, State1} = send_or_record_confirm(Delivery, State),
     Props = message_properties(Message, Confirm, State1),
-    {IsDuplicate, BQS1} = BQ:is_duplicate(Message, BQS),
-    State2 = State1#q{backing_queue_state = BQS1},
-    case IsDuplicate orelse attempt_delivery(Delivery, Props, Delivered,
-                                             State2) of
-        true ->
+
+    case attempt_delivery(Delivery, Props, Delivered, State1) of
+        {delivered, State2} ->
             State2;
-        {delivered, State3} ->
-            State3;
         %% The next one is an optimisation
-        {undelivered, State3 = #q{ttl = 0, dlx = undefined,
+        {undelivered, State2 = #q{ttl = 0, dlx = undefined,
                                   backing_queue_state = BQS2,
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@noxdafox Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-server&pullRequestId=1741)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-server&pullRequestId=1741) to manually synchronize the status of this Pull Request.

michaelklishin:
Can we take a step back and explain why the current behavior has to change?

noxdafox:
Uhm... there seems to be some issue with GitHub and this issue ticket as @michaelklishin comment seems to appear/disappear?

@michaelklishin I explain what the issue is in both the commit message and in the "further comments". More information is also available at the linked [issue](https://github.com/noxdafox/rabbitmq-message-deduplication/issues/21). Can you please help me understanding what shall I clarify more?

michaelklishin:
@noxdafox under what circumstances did the client not receive a nack? The commit message doesn't explain that. Why is this a good idea to nack such messages? What kind of applications can be affected?

michaelklishin:
So the context is in https://github.com/noxdafox/rabbitmq-message-deduplication/issues/21. I'm not convinced this is what RabbitMQ users who don't use that plugin want.

michaelklishin:
The backing queue behavior docs do not suggest if duplicates should be rejected but it does [mention](https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_backing_queue.erl#L256) that the caller might consider "dropping" the message.

[rabbit_variable_queue:is_duplicate/2](https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_variable_queue.erl#L990) always returns `false` and `rabbit_priority_queue:is_duplicate/2` [passes through](https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_priority_queue.erl#L437) to the next BQ in the chain (e.g. the variable one).

It gets a lot more interesting in the classic [mirrored queue implementation](https://github.com/rabbitmq/rabbitmq-server/blob/4da1431d27336f31d5f04c23c8326680e6d25d5c/src/rabbit_mirror_queue_master.erl#L439). Mirrors can see a message more than once due to how the `gm` module (the "transport" and current "consensus" implementation) works and handlers failures. A duplicate in that case is not a
reason to unconditionally communicate a nack to the client.

noxdafox:
According to the `rabbit_backing_queue` behaviour, the [`is_duplicate`](https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_backing_queue.erl#L259) callback shall signal the queue process that the message about to be published is actually a duplicate. In such case, the queue process [will not `publish` the message] in the queue(https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_amqqueue_process.erl#L683). 

Nevertheless, the logic does not take this scenario into account. This becomes problematic if a client enables publish confirmation as the client is left hanging indefinitely. 

I did not take into account the mirrored queues implementation actually. I see how the raised concern is valid. 
I'll see if we can find a better way to tackle this. We don't want to send ack/nack back if the duplicate is the result of an internal duplication. Nevertheless, we don't want to let the client hang if a message is not enqueued because of some policy applied by the user.

michaelklishin:
We'd welcome a different PR for 3.8 (master) that would make it possible to address https://github.com/noxdafox/rabbitmq-message-deduplication/issues/21 without breaking changes.
I can't immediately think of a way to do this for mirrored queues but perhaps with quorum queues, when they are merged, there can be a way.

michaelklishin:
@noxdafox I can see why the current behavior in `amqqueue_process` is problematic for your plugin.

As @kjnilsson and I concluded in a 1-on-1 conversation, the callback was originally added to introduce some of the idiosyncrasies and flaws in the `gm` module and thus classic mirrored queues, not to actually let plugins deduplicate :(

kjnilsson:
Sadly the dynamic nature of RabbitMQ queues mean that many of the abstractions aren't as intuitive as they should be.

Could we extend the `is_duplicate` callback with a return value that tells the queue process to reject the message and not just drop it silently? like `{true, drop} | {true, reject} | false`

noxdafox:
I see, the behaviour documentation did not mention the callback was for internal use only.

May I suggest to change the callback interface or add a new one? This would still enable the Use Case without the need of affecting mirrored queues implementation.

For example, the `is_duplicate` callback could return a tuple `{bool, symbol}` where the symbol could instruct the process what to do. Something like `{true, ignore}` or `{true, reject}`.

The change would be minimal within the broker logic but I am not sure how flexible we can be when it comes to change behaviour APIs.

michaelklishin:
@noxdafox we can make any internal interface changes for 3.8.0. As long as on disk data and client protocol are not affected, we'd consider it ;)

noxdafox:
I will start by providing a RFC PR changing the `is_duplicate` behaviour and the related logic. 

We can later discuss whether to introduce `nack` for duplicated messages and what are the implications.

Issue #1743, opened 2018-10-22
Allow passive concurrent consumers to an active exclusive one


acogoluegnes:
In AMQP 0-9-1 the "exclusive consumer" flag will throw a protocol error if more than one exclusive consumer is attempted to be added. In some cases it would be more convenient to allow for passive concurrent consumers (they'd get no deliveries).

This is the "one-group case" of [message grouping](#262).

PR #1744, opened 2018-10-22
Introduce an option to have only one concurrent consumer active

acogoluegnes:
This commit introduces exclusive consumer. A queue is declared with the
x-exclusive-consumer argument set to true and it will dispatch messages
to only one consumer at a time. This exclusive consumer is typically
the first one to consume from the queue. Other consumers are simply
ignored. If the exclusive consumer is cancelled, another consumer is
picked up randomly.

Subsequent commits will handle other cases than simple cancellation
(e.g. the owning channel of the exclusive consumer is closed).

[#161090309]

References #1743


michaelklishin:
We already have an exclusive consumer feature (in AMQP 0-9-1). The naming has to change to indicate the difference from #1743.

`x-single-active-consumer` is the best name I can think of.

hairyhum:
Should exclusive consumer be internally renamed to single active consumer? It can be confusing to read the code otherwise.

acogoluegnes:
@hairyhum @michaelklishin Renaming makes sense. I'll resolve the conflicts ASAP.

acogoluegnes:
@hairyhum @michaelklishin OK, renaming and merging done. It's ready for QA. Note there may be some follow-up stories: implement single active consumer in quorum queues, policy support, and maybe changes in the management UI, but I wouldn't mind some feedback about what's been done so far.

src/rabbit_amqqueue_process.erl
```diff
@@ -998,13 +1023,15 @@ i(effective_policy_definition,  #q{q = Q}) ->
         undefined -> [];
         Def       -> Def
     end;
-i(exclusive_consumer_pid, #q{exclusive_consumer = none}) ->
+i(exclusive_consumer_pid, #q{active_consumer = none}) ->
```
hairyhum: I understand that single active consumer feature is reusing the field for exclusive consumer, but should we reuse the info item? If there is a single ective consumer it will be reported as exclusive, which is close, but can be confusing.

acogoluegnes: This is fixed, single active consumer don't use exclusive consumer info items anymore.

michaelklishin:
Superseded by #1802.

PR #1758, opened 2018-11-02
Call rabbit:stop_apps in reverse order

dcorbacho:
`rabbit:stop_apps` already reverts the app order

Tested in the management app. Disabling web dispatch has to disable first management, if it doesn't the http listener won't be removed from mnesia and will still be reported in `rabbit:status()`.

## Types of Changes

## Checklist


michaelklishin:
Could reproduce with 3.7.8 and master without the patch, plus https://github.com/rabbitmq/rabbitmq-management/pull/626 fails without it. All server suites pass 👍.

PR #1770, opened 2018-11-20
VHost recovery memory

kjnilsson:
After vhost recovery this change triggers an explicit gc to reclaim any memory allocated during recovery as after the init function has completed this process does very little else and may never allocate enough to trigger another gc.

Mailing list link: https://groups.google.com/d/msg/rabbitmq-users/ZbTNp6YqZeo/53JLiy2zAQAJ


kjnilsson:
For a newly created vhost this processes uses ~100kB before this change and ~16kB after.

michaelklishin:
Thank you, this is an interesting find.

PR #1782, opened 2018-11-30
Quorum queue confirm availability

kjnilsson:
This addresses two issues:

There was a bug in the channel that resulted in confirms not being sent after a quorum queue leader change.

A scenario where a confirm could never be issued after a quorum queue leader change.

There are also a few miscellaneous fixes and refactorings.


kjnilsson:
This change works best with: https://github.com/rabbitmq/ra/pull/30

PR #1792, opened 2018-12-06
Ignore error when dead letter exchange cannot be found

dcorbacho:
[#162411672]

## Types of Changes

## Checklist

### DIFF
--- a/src/rabbit_quorum_queue.erl
+++ b/src/rabbit_quorum_queue.erl
@@ -559,9 +559,14 @@ args_policy_lookup(Name, Resolve, Q = #amqqueue{arguments = Args}) ->
 dead_letter_publish(undefined, _, _, _) ->
     ok;
 dead_letter_publish(X, RK, QName, ReasonMsgs) ->
-    {ok, Exchange} = rabbit_exchange:lookup(X),
-    [rabbit_dead_letter:publish(Msg, Reason, Exchange, RK, QName)
-     || {Reason, Msg} <- ReasonMsgs].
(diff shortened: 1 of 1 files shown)


src/rabbit_quorum_queue.erl
```diff
@@ -559,9 +559,14 @@ args_policy_lookup(Name, Resolve, Q = #amqqueue{arguments = Args}) ->
...
 dead_letter_publish(X, RK, QName, ReasonMsgs) ->
-    {ok, Exchange} = rabbit_exchange:lookup(X),
-    [rabbit_dead_letter:publish(Msg, Reason, Exchange, RK, QName)
-     || {Reason, Msg} <- ReasonMsgs].
+    case rabbit_exchange:lookup(X) of
+        {ok, Exchange} ->
+            [rabbit_dead_letter:publish(Msg, Reason, Exchange, RK, QName)
+             || {Reason, Msg} <- ReasonMsgs];
+        {error, not_found} ->
+            rabbit_log:warning("Dead lettering for quorum queue ~p failed: "
```
michaelklishin: If we log this for every message we can swamp the log very quickly.

dcorbacho: should we ignore it then?

kjnilsson: Probably but on the other hand this is useful information. What do classic queues do?

dcorbacho: Nothing, they only drop the messages from the queue

PR #1797, opened 2018-12-10
New x-open exchange : pure AMQP message routing on data and more

sylvainhubsch:
## TL;DR

Exchange types defined by AMQP 0.9 are tied to really closed rules such as AMQP topics or equality on headers keys/values for example; this new exchange type permits you to extend those rules and to mix matches on all AMQP data like you ~~want~~ need to. It may also greatly simplify your global process by moving some works made by consumers into some more specific routing logics in RabbitMQ.

## Types of Changes

## Checklist

- [X] Any dependent changes have been merged and published in related repositories

## Further Comment

__What are new functionalities :__
* new headers match operators : on keys (exist / not exist), on values : numeric comparison (<, <=, >=, >), non-equality, regex matching
* match on routing key : non-equality, regex matching
* match on UTC or local datetime the message is processed
* ordering bindings, define some gotos depending on binding's evaluation
* route message to multiple destinations in one binding (by name or regex match), can be set via binding only and/or by producer (ala "CC/BCC")

By mixing those simple rules, __you can now build some others__; 3 simple examples :
* mimic all other exchange types in one place
* feed some queues on monday between 7am to 17pm when some header key's value is greater than 20 and/or routing key have "foo" pattern
* route message to queues name defined via regex excluding those having "private" word

Beyond beeing a pure data routing exchange, I think it is a new way of thinking the AMQP protocol; a process may now have 3 roles : Producer, Consumer and Binder. By combining declaration of bindings, producers and consumers can work together (but without knowing eachother) to achieve some very specific routing rules in a more decoupling way.
I think this is really compatible with the "AMQP's spirit"; this is why I consider this exchange not to be a plugin, but should be in RabbtiMQ's core.
In a way, I'm pretty sure this new exchange's type may introduce new ways to build systems (especially for IT people who don't know AMQP at all..).

Yet another thing, this PR replaces the cancelled #1677 but has the same performance gain on headers match only compared to the native headers exchange type :
image

__Regarding Q/A about code__, I'm well aware that there are many questions or things to discuss before merging :
* should the table be copied in RAM for each local node ?
* is it ok to use process dict to hold some data during one message processing ?... I know it is used in other RabbitMQ's.. so I did the same to gain some perf
* surely many others things..

Also, __many extends may be considered before beeing merged__ :
* regexes can really hurt servers, but from my point of view we must offer them
--> We could consider adding some limitations on regex's engine iterations via exchange's arguments ? (something like "x-open-regex-iter-max" ?)
* that exchange may skip some bindings due to routing facilities (contrary to all other exchange's types)
--> From my point of view, this would be a "must have" : doing some restrictions/authorisations on some binding's arguments (the Binder's role) ?
* filtering on datetimes currently hurts performances (mainly because it is based on calendar module?)
--> We should rewrite that as a NIF module
* listing all queues may be a performance problem (this is used in regexes destinations routing facility)
--> Maybe there is a neater solution via rabbitmq events, to hold some cache ?
* use "phash2.." instead of "crypto:hash(md5..." in case of backports ?
* source code is almost a mess, and this should no be !
--> Surely we should split validations / other in different modules ?
--> We may consider having one function per usage used by binding's validation AND creation
--> Usage of some "ZZZ" vars : I should rename them !
* etc..

And finally, there are many improvements I would love to implement but I don't currently have time :
* save some "states" between bindings which we can match on to considerably improve and simplify bindings's declaration (and/or/goto logics)
* include reverse AMQP topic exchange
* build some match on key names via regexes with minimal complexity and match on exact list of key names (regexes or not)
* implement the "set" match type (like "all" but without binding's headers beeing mandatory)
* capability of producers to sort headers locally before sending them, so that it would not be done prior to process message by exchange (that one is fun :) )
* make some match on (start of) payload : lots of questions here (we should constraints on max message's size etc..)
* make some match on user properties such as content-type etc...
* etc..

Yes, I'm now in the process to do some documentations and demos..

Thanks to have read that line as it is the last :)

### DIFF
--- a/src/rabbit_exchange_type_open.erl
+++ b/src/rabbit_exchange_type_open.erl
@@ -1,0 +1,1022 @@
+%% The contents of this file are subject to the Mozilla Public License
+%% Version 1.1 (the "License"); you may not use this file except in
+%% compliance with the License. You may obtain a copy of the License
+%% at http://www.mozilla.org/MPL/
+%%
+%% Software distributed under the License is distributed on an "AS IS"
+%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
+%% the License for the specific language governing rights and
+%% limitations under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% The Initial Developer of the Original Code is GoPivotal, Inc.
+%% Copyright (c) 2007-2017 Pivotal Software, Inc.  All rights reserved.
+%%
+
+-module(rabbit_exchange_type_open).
+-include("rabbit.hrl").
+-include("rabbit_framing.hrl").
+
+-behaviour(rabbit_exchange_type).
+
+-export([description/0, serialise_events/0, route/2]).
+-export([validate/1, validate_binding/2, info/1, info/2,
+         create/2, delete/3, policy_changed/2, add_binding/3,
+         remove_bindings/3, assert_args_equivalence/2]).
+-export([init_db/0]).
+
+-rabbit_boot_step({?MODULE, [ {description, "exchange type x-open : registry"}
+  ,{mfa,      {rabbit_registry, register, [exchange, <<"x-open">>, ?MODULE]}}
+  ,{cleanup,  {rabbit_registry, unregister, [exchange, <<"x-open">>]}}
+  ,{requires, rabbit_registry}
+  ,{enables,  kernel_ready}
+  ]}).
+-rabbit_boot_step({rabbit_exchange_type_open_mnesia, [ {description, "exchange type x-open : mnesia"}
+  ,{mfa,      {?MODULE, init_db, []}}
+  ,{requires, database}
+  ,{enables,  external_infrastructure}
+  ]}).
+
+description() ->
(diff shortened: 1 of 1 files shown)


michaelklishin:
@sylvainhubsch we really appreciate the amount of time you are putting into optimizing the headers exchange (or finding an optimal alternative). We historically have chosen to develop plugins for new exchanges instead of bundling them with RabbitMQ. I think it's the right thing to do here, too.

I don't know if `x-open` is a very descriptive name but I also don't have any suggestions right now. Please wrap this into a plugin and our team would be happy to take a look. Note that it's currently a bit hard to do without at least some high level test cases.

michaelklishin:
Note that as a plugin this exchange can

 * Target RabbitMQ 3.7.x
 * Evolve independently and at the pace you choose both in the near and long term
 * Be evaluated by the community and battle tested by your own systems with very little risk (compared to trying to get it into RabbitMQ core)

So I think it's in everyone's interest right now to proceed with this being a separate plugin.

sylvainhubsch:
@michaelklishin I perfectly understand. Thanks for your feedback.

PR #1802, opened 2018-12-14
Single active consumer

### LINKED ISSUE #1799: Support single active consumer in quorum queues
See #1743.

acogoluegnes:
Fixes #1799, supersedes #1744.

Depends on https://github.com/rabbitmq/rabbitmq-common/pull/288.


src/rabbit_fifo.erl
```diff
@@ -698,6 +705,54 @@ num_checked_out(#state{consumers = Cons}) ->
                 end, 0, maps:values(Cons)).
 
 cancel_consumer(ConsumerId,
+                {Effects0, #state{single_active_consumer_on = false} = S0}) ->
+    %% general case, single active consumer off
+    cancel_consumer0(ConsumerId, {Effects0, S0});
+cancel_consumer(ConsumerId,
+                {Effects0, #state{single_active_consumer_on = true,
+                                  waiting_consumers = WaitingConsumers  } = S0}) when length(WaitingConsumers) == 0 ->
```
kjnilsson: match waiting consumers with `[]` rather than taking the length should be more efficient.

src/rabbit_fifo.erl
```diff
@@ -698,6 +705,54 @@ num_checked_out(#state{consumers = Cons}) ->
...
+cancel_consumer(ConsumerId,
+                {Effects0, #state{consumers = Cons0,
+                                  single_active_consumer_on = true,
+                                  waiting_consumers = WaitingConsumers0  } = State0}) ->
+    %% single active consumer on, consumers are waiting
+    case maps:take(ConsumerId, Cons0) of
+        {_CurrentActiveConsumer, _} ->
+            % The active consumer is to be removed
+            % Cancel it
+            {Effects1, State1} = case maps:take(ConsumerId, Cons0) of
```
kjnilsson: Formatting: line up `case` with `end`

src/rabbit_fifo.erl
```diff
@@ -698,6 +705,54 @@ num_checked_out(#state{consumers = Cons}) ->
...
+                {#consumer{checked_out = Checked0}, _} ->
+                    S = return_all(State0, Checked0),
+                    Effects = cancel_consumer_effects(ConsumerId, S, Effects0),
+                    {Effects, State0};
+                error ->
+                    {Effects0, State0}
+            end,
+            % Take another one from the waiting consumers and put it in consumers
+            {NewActiveConsumerId, NewActiveConsumer} = lists:nth(1, WaitingConsumers0),
```
kjnilsson: Rather than using `nth` and `delete` you could simply match the head in the function clause `[{_, _} | RemainingWaitingConsumers]`

src/rabbit_fifo.erl
```diff
@@ -698,6 +705,54 @@ num_checked_out(#state{consumers = Cons}) ->
...
+            {Effects1, State2};
+        error ->
+            % The cancelled consumer is not the active one
+            % Just remove it from idle_consumers
+            case lists:keytake(ConsumerId, 1, WaitingConsumers0) of
+                {value, {ConsumerId, #consumer{checked_out = Checked0}}, WaitingConsumers1} ->
+                    S = return_all(State0, Checked0),
+                    Effects = cancel_consumer_effects(ConsumerId, S, Effects0),
+                    {Effects, #state{waiting_consumers = WaitingConsumers1}};
+                false ->
```
kjnilsson: The prior function clause when it matches on the empty list should make this clause impossible.

src/rabbit_fifo.erl
```diff
@@ -232,7 +232,12 @@
          prefix_msg_counts = {0, 0} :: {Return :: non_neg_integer(),
                                         PrefixMsgs :: non_neg_integer()},
          msg_bytes_enqueue = 0 :: non_neg_integer(),
-         msg_bytes_checkout = 0 :: non_neg_integer()
+         msg_bytes_checkout = 0 :: non_neg_integer(),
+         %% whether single active consumer is on or not for this queue
+         single_active_consumer_on = false :: boolean(),
```
kjnilsson: I'm wondering if a `consumer_stategy = default :: default | single_active` or similar would provide a bit more future proofing than a bool.

src/rabbit_fifo.erl
```diff
@@ -232,7 +232,12 @@
...
+         %% waiting consumers, one is picked active consumer is cancelled or dies
+         %% used only when single active consumer is on
+         waiting_consumers = [] :: list()
```
kjnilsson: should this be `waiting_consumers = [] :: [{consumer_id(), consumer()}]` ?

src/rabbit_fifo.erl
```diff
@@ -268,9 +273,11 @@ update_config(Conf, State) ->
     DLH = maps:get(dead_letter_handler, Conf, undefined),
     BLH = maps:get(become_leader_handler, Conf, undefined),
     SHI = maps:get(shadow_copy_interval, Conf, ?SHADOW_COPY_INTERVAL),
+    SingleActiveConsumerOn = maps:get(single_active_consumer_on, Conf, false),
```
kjnilsson: `single_active_consumer_on` would need to be added to the `config()` type above.

src/rabbit_fifo.erl
acogoluegnes: Done in the body of the function, the function clause was getting beefy.

src/rabbit_fifo.erl
acogoluegnes: Good idea, done.

src/rabbit_fifo.erl
acogoluegnes: Done. This needed to declare a new `consumer` type, just after the record declaration.

src/rabbit_fifo.erl
acogoluegnes: Done.

PR #1807, opened 2018-12-19
Do not include x-delivery-count header on first delivery.

kjnilsson:
[#162755631]


BhupenPal:
@kjnilsson @michaelklishin Why this change has been done?

BhupenPal:
@michaelklishin  Yes, I understand. Thanks for replying, I have a contrary viewpoint but I guess it's a discussion for some other day

kjnilsson:
> @michaelklishin Yes, I understand. Thanks for replying, I have a contrary viewpoint but I guess it's a discussion for some other day

There is no need to introduce unnecessary overhead for the most common delivery (i.e the first delivery). that is why it isn't included and almost certainly never will be. Cheers.

PR #1812, opened 2018-12-28
Introduce a configurable limit to message size

hairyhum:
## Proposed Changes

Motivation: very large messages can be problematic for Erlang distribution. The distribution message can only be up to 2GB, but there will be some additional data, so the message body should be even smaller.
Another issue is failure detection, which is based on heartbeats. If a message is taking too long to transfer, the cluster nodes can miss the heartbeat and break the cluster.

Things like garbage collection and general performance can also suffer with a bigger message sizes.

Add `max_message_size` configuration to configure the limit in bytes.
If message is bigger - channel exception will be thrown, just like it was before for messages bigger than 2GB

Default limit is 128MB (134217728 bytes).
There is still a hard limit of 521MB (536870912), see https://github.com/rabbitmq/rabbitmq-common/pull/289.

## Types of Changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


hairyhum:
Should we consider adding some human-friendly configuration format (for example `128MB`), as it is for memory limits?

hairyhum:
@michaelklishin no, it's using ets tables https://github.com/erlang/otp/blob/master/lib/kernel/src/application_controller.erl#L332 
And it does not need to get a group leader, because appliction name is explicit in this case.

michaelklishin:
Even then, why not make it a constant in the channel state and eliminate several new extra calls that would effectively always return the same value?

hairyhum:
Fair enough. The value can technically change during a channel lifetime, but that's not a public API, so we can move it to the channel state.

PR #1823, opened 2019-01-10
Improve quorum queue deletion

kjnilsson:
Only delete the amqqueue record after successful Ra cluster deletion.

Force delete Ra servers when not able to achieve consensus.

### DIFF
--- a/src/rabbit_fifo.erl
+++ b/src/rabbit_fifo.erl
@@ -609,7 +609,8 @@ state_enter(recovered, #state{prefix_msg_counts = PrefixMsgCounts})
   when PrefixMsgCounts =/= {0, 0} ->
     %% TODO: remove assertion?
     exit({rabbit_fifo, unexpected_prefix_msg_counts, PrefixMsgCounts});
-state_enter(eol, #state{enqueuers = Enqs, consumers = Custs0}) ->
+state_enter(eol, #state{enqueuers = Enqs,
+                        consumers = Custs0}) ->
     Custs = maps:fold(fun({_, P}, V, S) -> S#{P => V} end, #{}, Custs0),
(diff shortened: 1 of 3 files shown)


kjnilsson:
Those tables won’t be cleared up until the wal rolls over. They will get
cleared up so not a leak per se unless they are still there after a roll
over. High churn ra clusters isn’t something that’s been optimised for.

michaelklishin:
Fair enough, quorum queues make little sense for transient data, so this is acceptable.

kjnilsson:
Yes it still tries all the servers in the cluster and logs if any of them fail which is sort of expected. A bit simplistic perhaps, we could possibly filter out the server that timed out to get rid of that error message.

kjnilsson:
Actually that won't work at all as it doesn't know which nodes are down so it needs to try hare anyway. I think this succeeded in the end but that the management timeout was too low.

michaelklishin:
At the very least the code that force deletes should handle down nodes?

kjnilsson:
@michaelklishin each call to force delete a server is inside a `catch` expression, if it errors we log a message to say we couldn't delete that node (as unavailable) which means the ra data dir could not be removed and may need manual clean up. Servers that were removed successfully are not logged (`rabbit` it the scenario above).

I'm not sure how we can make this a bit more intuitive. Perhaps we shouldn't automatically fall back on force deletion after a consensus server deletion times out and instead provide cli tool to do it or similar?

michaelklishin:
@kjnilsson I strongly oppose "more manual intervention required" kind of solutions for such a basic operation. How hard is deleting a queue?

michaelklishin:
We concluded that another test is needed here on my end since the goal of this PR is to avoid manual intervention.

PR #1832, opened 2019-01-15
Add an upgrade function to remove all default bindings.

hairyhum:
Versions before 3.8 create default bindings for queues in the database.
Since #1721 they are replaced with placeholders only for list operations.
We need to cleanup the bindings to improve binding performance with
older queues.

[#163224049]

### DIFF
--- a/src/rabbit_upgrade_functions.erl
+++ b/src/rabbit_upgrade_functions.erl
@@ -658,9 +658,10 @@ exchange_options(Table) ->
        operator_policy, decorators, options]).
 
 default_bindings() ->
-    %% The bindings are not used anyway,
-    %% and are replaced by placeholders.
-    %% It should be safe to remove them dirty.
+    %% Default exchange bindings are now implicit
+    %% (not stored in the route tables).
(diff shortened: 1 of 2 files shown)


src/rabbit_upgrade_functions.erl
```diff
@@ -654,6 +657,22 @@ exchange_options(Table) ->
...
        operator_policy, decorators, options]).
 
+default_bindings() ->
+    %% The bindings are not used anyway,
+    %% and are replaced by placeholders.
+    %% It should be safe to remove them dirty.
+    Queues = mnesia:dirty_all_keys(rabbit_queue),
+    [begin
+        Binding = rabbit_binding:implicit_for_destination(Q),
+        mnesia:dirty_delete(rabbit_route, Binding),
```
michaelklishin: This tries to delete bindings from a table that stores routes.

hairyhum:
Make it load durable queues and remove durable bindings. When moving to feature flags we will need to load all the queues and remove all the bindings.

PR #1833, opened 2019-01-15
v3.7.x: backport implicit bindings for forward-compatibility with 3.8.

hairyhum:
Backport of c7d107dd9fc1a336246f5105574f6c6a5acc2593
10013f55c653f8fa06e581368fe236a1ad1f894d
e01dc9882a637cdb5b8b9415947d7d704b8dde91
7ba04a6a681e3e3f805b582a7f0aa3f8361c248a

In 3.8 queues will not create default bindings.
Because default bindings are only used in list function, it should be
safe to run in a mixed mode with 3.8.
List functions should show implicit bindings for queues created in 3.8.

Min master locator logic is simplified and with or without default
bindings will have the same behaviour as before.

Follow-up to #1721


michaelklishin:
Management plugin suite revealed that we missed https://github.com/rabbitmq/rabbitmq-server/commit/76acfa47cc7da7c1b9072747a0a2bb14b51d3094.

PR #1856, opened 2019-01-30
Configure Ra to log using lager

kjnilsson:
[#163518117]

Requires https://github.com/rabbitmq/ra/pull/49 and https://github.com/rabbitmq/rabbitmq-common/pull/296


michaelklishin:
If `rabbit_log` isn't suitable we can extend it. I doubt we really need more logging modules.

kjnilsson:
`rabbit_log` already has a log/4 function but with a different signature but I could make it delegate to `rabbit_log:log/4` perhaps?

michaelklishin:
I see. I'd say let's make it a new OTP logger shim not specific to Ra. How about `rabbit_otp_logger` in `rabbitmq-common`?

kjnilsson:
We could do that although I'm not sure what other uses it may have at this point. Thinking about it though should logs coming from ra have their own category / sink or just use the default lager sink?

michaelklishin:
We can consider a separate category. My point is that all logging modules should probably be colocated.

PR #1871, opened 2019-02-07
QQ: Repair amqqrecord leader info on tick

kjnilsson:
It is possible that the initial update could fail and then the leader
would never be updated. Here we check periodically if the amqqueue
record pid field is set to the current leader, if not we update it.

Also changes the Ra internal tick frequency to every 5s.

[#163554015]

This PR should work without https://github.com/rabbitmq/ra/pull/55
but there is a fix in that PR that makes it less likely to trigger the repair process shortly after a leader change.


hairyhum:
It probably makes sense to test to which extent RabbitMQ stays usable in general after partitions.

Issue #1873, opened 2019-02-10
Possible binding/route table update inconsistency affects queue operations


michaelklishin:
## Overview

This is for something that's been reported every few weeks on the list for a few years now (e.g. [one example](https://groups.google.com/forum/#!topic/rabbitmq-users/3tlWiaeRBDc) that we never managed to reproduce it, even if others [supposedly can](https://groups.google.com/d/msg/rabbitmq-users/ble31gjFGUE/-rYfRf1JDAAJ). Filing it here so that there is an issue number that's public.

The issue leads to certain operations on a queue failing with a log message that says something like this

```
no binding {queue} between exchange {exchange} in vhost {vhost}
```

and they persist until the queue is deleted and redeclared.

## Current Hypothesis and Findings

One working theory is that it's fundamentally something similar to #1691 in a different place: schema data store transaction retries can affect binding/route table consistency in unexpected ways. That can lead to other parts of the code finding stray or missing bindings they do not expect.

Another is that there's a race condition between node monitor operations on different nodes
which can delete queues and their dependent bindings concurrently in separate transactions.

The log entry is reported by a function that's delegated to by `{queue, exchange}.{bind, unbind}` operations in [rabbit_channel](https://github.com/rabbitmq/rabbitmq-server/blob/ed3dd4d6257df0925ce78ad94de099304346da4c/src/rabbit_channel.erl#L1850). The eventually go through `rabbitmq_binding:add/3` and `rabbit_binding:remove/3` that [perform a `rabbit_route` read](https://github.com/rabbitmq/rabbitmq-server/blob/ed3dd4d6257df0925ce78ad94de099304346da4c/src/rabbit_binding.erl#L165).

## Relevant Changes Merged

#1878, #1884

## Known Ways to Reproduce

#1884 review notes [contain a synthetic way of reproducing](https://github.com/rabbitmq/rabbitmq-server/pull/1884#issuecomment-464277810). If you have a way to reproduce at least semi-reliably
without directly messing with schema tables, **please share it on the mailing list**.

Somewhat relevant here: #1832.

michaelklishin:
#1872 mentions that one way of reproducing is to stop RabbitMQ on all cluster nodes at the same time. While this is not a scenario we find realistic and "solvable", it can be tried to gain improved understanding.

michaelklishin:
Not scheduling for any milestone because we are not confident that #1878/#1879 address every reported case.

michaelklishin:
[...]

PR #1878, opened 2019-02-12
Recover bindings for all durable queues including failed to recover.

hairyhum:
If a queue fails to recover it may still be restarted by the supervisor
and eventually start. After that some bindings may be in rabbit_durable_route
but not rabbit_route. This can cause binding not found errors.

If bindings are recovered for failed queues, the behaviour will be
the same as for the crashed queues. (which is currently broken
but needs to be fixed separately)

Addresses #1873
[#163919158]

This will reqruire a separate PR for 3.7


hairyhum:
Separate PR for 3.7

Ayanda-D:
ah, solving this weird case https://groups.google.com/forum/#!topic/rabbitmq-users/-X19dq8cYaY thank you! 👍

michaelklishin:
@Ayanda-D we don't yet know and would not claim it is sufficient because there was no reliable way to introduce any of those reports :(

Ayanda-D:
@michaelklishin ok, this was reported to us as well, but haven't been able to to reproduce it either :/ seemed the error handling could be improved as well? When binding [lookup](https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_binding.erl#L165-L166)  in `rabbit_route` is done and not found, but exists in [rabbit_durable_route](https://github.com/rabbitmq/rabbitmq-server/blob/ed3dd4d6257df0925ce78ad94de099304346da4c/src/rabbit_binding.erl#L181-L182), before immediately returning [not_found](https://github.com/rabbitmq/rabbitmq-server/blob/ed3dd4d6257df0925ce78ad94de099304346da4c/src/rabbit_binding.erl#L193) to the channel, maybe do/attempt a [re-sync](https://github.com/rabbitmq/rabbitmq-server/blob/ed3dd4d6257df0925ce78ad94de099304346da4c/src/rabbit_binding.erl#L424) from `rabbit_durable_route`, to update `rabbit_route` and `rabbit_semi_durable_route`? And only return `not_found` if binding still doesn't exist after a "failed re-synch" attempt (if such a case can ever occur)? coz if its in `rabbit_durable_route` then surely the binding exists, and returning `not_found` here doesn't sound right. Some thoughts, guess this PR resolving the issue on recovery then maybe this case may never be hit on runtime. We'll see 😅

hairyhum:
@Ayanda-D this PR will fix `binding not found` errors only if there were failed queues. You can check that in logs searching for `Queue <...> failed to initialise` in the logs.

The option to allow recovery of bindings does also make sense. I was thinking of letting `queue.bind` to create bindings even if there is an inconsistency, because that's what clients use to recover the topologies.

Ayanda-D:
Ok, sounds good 👍

PR #1879, opened 2019-02-12
Recover bindings for all durable queues including failed to recover. 3.7

hairyhum:
If a queue fails to recover it may still be restarted by the supervisor
and eventually start. After that some bindings may be in rabbit_durable_route
but not rabbit_route. This can cause binding not found errors.

If bindings are recovered for failed queues, the behaviour will be
the same as for the crashed queues. (which is currently broken
but needs to be fixed separately)

This is #1878 for version 3.7.x

Addresses #1873
[#163919158]


michaelklishin:
Setting milestone on the PR here because we are not yet ready to consider #1873 resolved (and thus can't know what milestone it should use).

PR #1889, opened 2019-02-18
Poison handling in quorum queues

dcorbacho:
## Proposed Changes

Drop messages that exceed the configured number of delivery attempts.
If a DLX is configured, dead letter them.

Part of #502.

## Types of Changes

## Checklist


kjnilsson:
Looks generally fine but it needs to be tested through `rabbit_fifo_props` which I am pretty sure will fail as we need to persist the delivery count in the dehydrated state as well message bytes. Perhaps we should add the bytes to the headers map and use this as our per message dehydrated data?

PR #1893, opened 2019-02-19
Make sure that publishes to dead or unaccessible queues return nack

hairyhum:
The problem:

If a queue is crashed or disconnects during a publish - the client still gets `ack`, while the message is not enqueued.
Sonce it was routed successfully - there will be no error even if `mandatory` is set.

Current behaviour:

Channel is counting unacked messages in a remove-only data structure `dtree`.
Each published message id is associated with a list of queues where it was routed to.
On confirm of queue failures queues were removed from the list
As soon as there are no queues in the list - the message can be confirmed.

This meant that if all queues fail with "not abnormal" reasons - the message may be
confirmed, but not enqueued.

Fix:

This patch adds confirmation tracking for publishes, so if a message was confirmed
by at least one queue it was routed to, it can be confirmed as soon as all queues
confirm or fail. If any of the queues fail it will be rejected.

[#163952410]


hairyhum:
UPDATE:

Changed the behaviour to track partial confirms per queue to handle mirrored queues failures only.

Also tolerate queue deletion separately by confirming on `normal` exit reason.

kjnilsson:
Sorry about being super slow on reviewing this.

I think the approach seems fine although I find mirrored queue failure semantics super hard to reason about which makes me a bit worried. Also I cannot see how the HA queues are handled differently from other classic queues. Can you explain?

hairyhum:
How it worked before:
Channel sends publishes to all mirrors and waits for a response from each of them (by pid). If one fails - the pid is removed from the "waiting ack" list for the message. When the last queue pid is removed from the list - the message was confirmed.
As I understand, this behaviour is supposed to tolerate mirror failure, but it acts the same way for the non-mirrored queues if the message was routed to multiple. Also it does not have a limit on how many queues/mirrors may fail.
Effectively this logic did not distinguish between queue pid failures (except "abnormal") and confirms. 

The new behaviour:
Channels keep track of which queues confirmed a message in addition to pids. If a queue process fails and there are no more pids to wait for a confirmation from - the messages to be confirmed are checked if they were confirmed by the failing queue. Messages which weren't are rejected.

This makes me think that it's not entirely correct still. Because if the last queue to fail had confirmation from other mirrors the message will still be confirmed.
Seems like we need to redesign the dtree data structure. It makes reasoning about this logic too hard.

hairyhum:
Old commits were removed because they may be confusing.
New behaviour:
On confirm messages, track which queues (by name) have confirmed a message.
Messges are confirmed by all pids in the queue (for mirrored queues there are may pids)
When there are not pids to confirm - send either confirm or reject. Confirm is sent if all the queues (by names) had at least one pid, which confirmed the message.

PR #1903, opened 2019-02-28
rabbit_fifo: change single active consumer on noconnection

kjnilsson:
This ensures progress is made when e.g. a node hosting the currently active consumer channel is disconnected or shut down.

Also includes a bunch of other fixes to ensure progress when consumers are cancelled but still have unacked messages outstanding and consumer that receive credit through the manual credit process (as used with AMQP 1.0)

### DIFF
--- a/src/rabbit_fifo.erl
+++ b/src/rabbit_fifo.erl
@@ -195,7 +195,8 @@ apply(Meta, #return{msg_ids = MsgIds, consumer_id = ConsumerId},
 apply(Meta, #credit{credit = NewCredit, delivery_count = RemoteDelCnt,
                     drain = Drain, consumer_id = ConsumerId},
       #?MODULE{consumers = Cons0,
-             service_queue = ServiceQueue0} = State0) ->
+               service_queue = ServiceQueue0,
+               waiting_consumers = Waiting0} = State0) ->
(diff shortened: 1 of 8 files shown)


src/rabbit_fifo.erl
```diff
@@ -294,27 +310,64 @@ apply(#{index := RaftIdx}, #purge{},
...
+    %% select a new consumer from the waiting queue and run a checkout
+    State2 = State1#?MODULE{waiting_consumers = WaitingConsumers},
+    {State, Effects1} = activate_next_consumer(State2, Effects0),
+
+    %% mark any enquers as suspected
+    Enqs = maps:map(fun(P, E) when node(P) =:= Node ->
+                            E#enqueuer{status = suspected_down};
+                       (_, E) -> E
+                    end, Enqs0),
+    Effects = [{monitor, node, Node} | Effects1],
```
acogoluegnes: Don't the effects need to contain `{aux, inactive}` if no new active consumer could have been set up? (same behavior as the next clause)

kjnilsson: ok I've addressed this. I think the whole aux active/inactive could do with a review as currently is seems more like art than science to get that working well with qqs.

Issue #1869, opened 2019-02-06
Consider starting protocol listeners later


michaelklishin:
Per discussion with @Gsantomaggio and Mat Burton's excellent feedback in [this rabbitmq-users thread](https://groups.google.com/forum/#!topic/rabbitmq-users/ctLroWGmqGQ).

When a node, in particular a new node, joins a cluster it becomes "eligible" for client operations as soon as

 * It is known to be a cluster member
 * It starts TCP/TLS listeners for all enabled protocols

So sometimes we'd see clients, directly or via remote nodes acting as coordinators, to begin executing operations on a node that is not 100% ready for that, e.g.

 * Virtual hosts are still initializing
 * Stateful plugins are still starting (e.g. the consistent hash exchange plugin has its ring state to recover)

and so on. This can cause client operations to fail. Depending on the resiliency of the application this can result in a technical operations incident.

If we oversimplify, a node can be considered a cluster member as early as the schema database sync (without it there isn't much you can do on a node successfully) and is hard to tweak due to various OTP dependencies using a very simplistic logic when checking if a node is available/ready. This part is hard to change.

When listener are started is a piece of node-local state and something the node itself controls fully. This part is trivial to change.

By starting listeners later we reduce the probability (although not to 0, of course) of clients attempting to perform operations on a node that's not really 100% ready even though it is booted and has enabled plugins and as far as the service manager or orchestration tool is concerned, generally ready to serve clients.

If plugin activation and TCP/TLS listener startup could coordinate with virtual host initialisation,
it would be an even bigger improvement but also a non-trivial change to the boot [step execution] process.

michaelklishin:
#1913 moves listener startup later. I still does not coordinate with every other component; that would be a major refactoring task (and a fairly high risk one since there's a fair number of components and plugins that initialise concurrently).

I considered adding a "sleep" boot step but expect it to be ill received by other team members :P

PR #1913, opened 2019-03-13
Move TCP and TLS listener startup to the last boot step "stage"

michaelklishin:
## Proposed Changes

This moves TCP and TLS listener startup to the last boot step "stage". This makes it less likely for clients to connect to and perform operations on a node that's still initialising. See #1869 for background.

## Types of Changes

- [x] Non-breaking change which fixes issue #1869)

## Checklist


michaelklishin:
I backed out a couple of unrelated boot step ordering changes after seeing how they can potentially be unsafe. I no longer can reproduce one curious startup race condition after that. Let's get this through the pipeline as all the tests I tried locally pass.

PR #1917, opened 2019-03-15
Quorum Queues: resend messages if first batch is lost

### LINKED ISSUE #1906: Messages ready is 0 for quorum queues when publishing starts before queue declare
Starting a publisher on the default exchange before a quorum queue is created causes ready number of messages to be `0` when the queue is declared, and this value stays like that while publishing takes place. The total number of messages reports the expected growing value in both `coarse_queue_metrics` table and the management UI. Once the publisher is stopped, the ready number is properly updated and matches the total.

See `empty-qq` below
screen shot 2019-03-04 at 3 16 23 pm

The same happens to `message_bytes` metrics in `queue_metrics`.

This issue only affects quorum queues.

dcorbacho:
`last_applied` is now initialised to -1 and resend always happens if we have an out of order sequence

Fixes #1906

## Types of Changes

## Checklist

### DIFF
--- a/src/rabbit_fifo_client.erl
+++ b/src/rabbit_fifo_client.erl
@@ -49,6 +49,8 @@
 -define(TIMER_TIME, 10000).
 
 -type seq() :: non_neg_integer().
+%% last_applied is initialised to -1
+-type maybe_seq() :: integer().
 -type action() :: {send_credit_reply, Available :: non_neg_integer()} |
                   {send_drained, CTagCredit ::
                    {rabbit_fifo:consumer_tag(), non_neg_integer()}}.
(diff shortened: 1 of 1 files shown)


src/rabbit_fifo_client.erl
```diff
@@ -63,7 +65,7 @@
                 servers = [] :: [ra_server_id()],
                 leader :: maybe(ra_server_id()),
                 next_seq = 0 :: seq(),
-                last_applied :: maybe(seq()),
+                last_applied = -1 :: maybe_seq(),
```
michaelklishin: Should we document inline why it is the case?

dcorbacho: done

acogoluegnes:
Reproduced and tested the fix, works as expected.

PR #1972, opened 2019-04-08
Consumer timeouts

kjnilsson:
Implements consumer timeouts for channels, i.e. to ensure that a faulty consumer that consumes but never acks will eventually be cancelled by the channel _and_ the queue and all pending messages will be returned to the queue for re-delivery.

This implementation piggy-backs on the (formely named) queue_cleanup timer so these timeouts are very coarse grained which should be ok as the consumer timeouts should be quite large to avoid false positives. The timer has been renamed a more generic: `queue_tick_timer` for periodic queue actions.

For basic.get and consumers that don't support the channel initiated basic.cancel mechanism the channels is instead closed.

### DIFF
--- a/src/rabbit_channel.erl
+++ b/src/rabbit_channel.erl
@@ -81,92 +81,96 @@
 %% Mgmt HTTP API refactor
 -export([handle_method/6]).
 
--record(ch, {
-  %% starting | running | flow | closing
-  state,
-  %% same as reader's protocol. Used when instantiating
-  %% (protocol) exceptions.
-  protocol,
-  %% channel number
-  channel,
-  %% reader process
-  reader_pid,
-  %% writer process
-  writer_pid,
-  %%
-  conn_pid,
-  %% same as reader's name, see #v1.name
-  %% in rabbit_reader
-  conn_name,
-  %% channel's originating source e.g. rabbit_reader | rabbit_direct | undefined
-  %% or any other channel creating/spawning entity
-  source,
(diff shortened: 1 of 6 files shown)


src/rabbit_channel.erl
```diff
@@ -805,34 +821,45 @@ handle_info({'DOWN', _MRef, process, QPid, Reason}, State) ->
...
-                            QName =  rabbit_quorum_queue:queue_name(QS),
-                            [] /= rabbit_amqqueue:lookup(QName)
+                            QName = rabbit_quorum_queue:queue_name(QS),
+                            [] /= rabbit_amqqueue:lookup([QName])
                     end, QueueStates0),
-    noreply(init_queue_cleanup_timer(State#ch{queue_states = QueueStates}));
-
-handle_info({channel_source, Source}, State = #ch{}) ->
-    noreply(State#ch{source = Source}).
+    case evalauate_consumer_timeout(State0#ch{queue_states = QueueStates1}) of
```
dcorbacho: typo `evalauate`

kjnilsson: done

vikinghawk:
What is the default timeout? And is that configurable server side or client side?

michaelklishin:
It is configurable via `rabbit.consumer_timeout` (no new style config key at the moment). The default is "no timeout". This is needed for Raft log to not grow overly long when buggy consumers consume and never (or "never" for a given time frame) acknowledge deliveries.

PR #1977, opened 2019-04-10
Status data structures update for 3.8

michaelklishin:
for CLI tools to format them.

Part of rabbitmq/rabbitmq-cli#340.


michaelklishin:
Merging per verbal approval in https://github.com/rabbitmq/rabbitmq-cli/pull/341.

Issue #2000, opened 2019-05-03
Message store GC race condition with readers.


hairyhum:
Reported in https://groups.google.com/forum/#!topic/rabbitmq-users/FOPHWSyO1aU

Message store GC is a separate process, which is communicates with
message store using a shared ETS table. This table contains information
about message store files.

Then the message store file is marked for compaction, the GC process will
read the file summary and process it as soon as there are no readers (queues).

At the same time queues requesting read from the message store may open this 
file. They increase readers counter in file summary to block it from being deleted.

This two operations may conflict, causing file to be read and processed by GC at 
the same time. The read operation handles that by catching exception.

The GC process calls has_readers function to filter only files without readers, but when performing a combine or delete action it reads from ETS again and asserts that there is still no readers.
This badmatch failure crashes the message store.

PR #2001, opened 2019-05-03
Move check for active readers to message store GC action function

hairyhum:
## Proposed Changes

Message store GC postpones processing of file, which have readers.
When performing an action, it asserts that there are no readers.

Check for readers may race with readers update by a queue, crashing
the message store.

Make check and assert work with the same lookup to reduce failure rate.

In case of races the queue process should handle exception instead.

It's quite hard to reproduce the race without adding sleeps in the code, unfortunately.
I was not able to produce a test case.

Addresses #2000
[#165755203]

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #2000)

## Checklist

### DIFF
--- a/src/rabbit_msg_store.erl
+++ b/src/rabbit_msg_store.erl
@@ -2068,6 +2068,8 @@ do_combine_files(SourceSummary, DestinationSummary,
               {#file_summary.file_size,        TotalValidData}]),
 
     Reclaimed = SourceFileSize + DestinationFileSize - TotalValidData,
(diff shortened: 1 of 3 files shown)


michaelklishin:
This looks reasonable. I'm in the process of adding a backing queue test that would trigger a compaction.

michaelklishin:
I'm still working on a test that would specifically trigger compaction. Had to change my approach as communicated during the sync-up.

michaelklishin:
Will backport after the pipeline passes.

michaelklishin:
Backported to `v3.7.x`.

PR #2017, opened 2019-05-28
Confirm messages on ack.

hairyhum:
Make it possible to configure queues to confirm messages either on enqueue (default) or on ack.
This will make it possible to communicate to the publisher that the message was processed by consumer.
For example in federation this will mean that the message is successfully published on the downstream (1 level).

Configurations:
queue argument `x-confirm-on`
policy `confirm-on`

Values:
`enqueue` - old behaviour, confirm on enqueue (default)
`ack` - new behaviour, confirm when message is acked or dropped

TODO:
- [x] There may still be some ways for a message to leave the queue
without acking. Need to find them all.

- [x] Dead lettering and dropping of the message should probably not
confirm, but reject the message.

- [x] When changing the configuration using policies to `enqueue`,
all messages will be confirmed. Need to implement some sort of
enqueued messages tracker to only confirm enqueued messages.

- [ ] Quorum queues

- [ ] Mirrored queues

- [x] rabbitmqctl info item

### DIFF
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -693,7 +693,8 @@ declare_args() ->
      {<<"x-queue-mode">>,              fun check_queue_mode/2},
      {<<"x-single-active-consumer">>,  fun check_single_active_consumer_arg/2},
      {<<"x-queue-type">>,              fun check_queue_type/2},
-     {<<"x-quorum-initial-group-size">>,     fun check_default_quorum_initial_group_size_arg/2}].
+     {<<"x-quorum-initial-group-size">>,     fun check_default_quorum_initial_group_size_arg/2},
+     {<<"x-confirm-on">>,              fun check_confirm_on/2}].
 
 consume_args() -> [{<<"x-priority">>,              fun check_int_arg/2},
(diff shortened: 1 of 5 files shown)


hairyhum:
Supporing this feature for mirrored queues correctly is channelnging, because mirroring is coupled with confirms (`rabbit_mirror_queue_slave` is a separate module, which does not support policies).
We may change the mirroring protocol to support that, but this will be a breaking change.

In the other hand, in most of use cases for confirm-on-ack queues it does not make sense to mirror them, for data safety at least.
Mirroring still may be useful for queue migration between nodes though. Eventually this should be addressed.

Changes in this PR may be used for non-mirrored queues. The only problem is to prevent it's being used with mirrored queues, because the setting may be configured in a policy and using parameters.
The setting may be limited to policy only and the policy may be validated to not intersect with HA to enforce non-mirrored queues.

hairyhum:
For the quorum queues this feature makes as much sense as for mirrored and may be addressed later.

hairyhum:
Added validations to make sure this setting is only configured with policy and that this policy conflicts with `ha-mode`

hairyhum:
requires https://github.com/rabbitmq/rabbitmq-common/pull/330

PR #2031, opened 2019-06-12
Update log configuration tests for rabbitmq/rabbitmq-common#328

michaelklishin:
To match a new sink introduced in https://github.com/rabbitmq/rabbitmq-common/pull/328.

Part of https://github.com/rabbitmq/rabbitmq-shovel/pull/61.


michaelklishin:
Backported to `v3.7.x`.

Issue #2039, opened 2019-06-20
NaN (non existing) Exclusive Queues caused by queue-master-locator argument


luos:
Hi,

Exclusive queues are affected by the global (config file), policy, and queue argument setting of `queue-master-locator`. When a client declares an exclusive queue it may happen that in this case the queue is not `client-local`, for example because the locator is set as `random` or `min-masters`. 

When a node hosting a non-client-local exclusive queue is restarted the queue dies, but the cleanup method is missing check if the queue pid is alive, something like:

```erlang
is_dead_exclusive(#amqqueue{exclusive_owner = OwnerPid, pid = QPid}) when is_pid(OwnerPid), is_pid(QPid) ->
    not (rabbit_mnesia:is_process_alive(OwnerPid)).
```

For this reason the still alive nodes think that the queue exists and do not allow the client to redeclare it. 

Two solutions came to our mind:
- If a queue is exclusive, it should be always client local. This may not be that good solution for users who rely on the fact that exclusive queues are also affected by these settings 
- Extend the is_dead_exclusive call with the following:
```not (rabbit_mnesia:is_process_alive(OwnerPid) andalso rabbit_mnesia:is_process_alive(QPid)).```
in this case there is some race condition between a client trying to re-declare the queue and it being actually deleted (i am not sure about this)

I'll open a PR to the `3.7.x` branch shortly with my proposed changes, please review it and merge it if you think it's OK.

I noticed that the `is_dead_exclusive`  changed on the `master` branch, but seems like it's now the opposite, it misses the Owner check? Maybe it was refactored to be different. I have not tested it with the new branch.

https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_amqqueue.erl#L1618-L1622

Thanks,
Lajos

michaelklishin:
There is a race condition between client actions and exclusive queue cleanup.

The option where exclusive queues are declared as `client-local` regardless of the locator policy makes sense to me. Adding a clause could be OK, too. @kjnilsson @hairyhum @gerhard @acogoluegnes thoughts?

michaelklishin:
@luos please submit the same PR against master, we will either accept it or work on an alternative that limits exclusive queues to the client local node.

velimir:
[...]

PR #2040, opened 2019-06-20
fix cleanup of exclusive queues in case of restart or partition

luos:
## Proposed Changes

When using different than client-local queues exclusive ones can
be placed on a different host than the client. When there is a partition
or restart these queues are stuck because they are not cleaned up in
the nodes as the connection is still alive when the node goes offline,
however the queue is dead in the partition. These queues can not be
redeclared or consumed.
 
This change fixes the test to test if the queue process is alive.
 
Fixes: #2039

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #2039)

## Checklist

### DIFF
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -1114,8 +1114,8 @@ is_mirrored(Q) ->
 
 is_dead_exclusive(#amqqueue{exclusive_owner = none}) ->
     false;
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you. Pull requests should be submitted against master.

michaelklishin:
So the implementation in master is different enough to warrant to different PRs and an exception to our standard workflow.

michaelklishin:
Thank you!

PR #2046, opened 2019-06-27
Pass all members to synchronous Ra operations

kjnilsson:
As this will make these more work reliably when leadership is in flux.

Also ensure that servers started and added after the initial declare
uses the same configuration.

requires https://github.com/rabbitmq/ra/pull/100.

### DIFF
--- a/src/rabbit_fifo.erl
+++ b/src/rabbit_fifo.erl
@@ -64,7 +64,9 @@
          make_credit/4,
          make_purge/0,
          make_purge_nodes/1,
-         make_update_config/1
+         make_update_config/1,
+
+         from_log/2
         ]).
 
 %% command records representing all the protocol actions that are supported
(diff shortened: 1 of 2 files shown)


gerhard:
Looks good, ship it. How to we stress-test this in LREs? Would this help? https://github.com/rabbitmq/rabbitmq-server-boshrelease/tree/master/jobs/periodic_shutdown

kjnilsson:
I think it's tricky to test this specific thing but lots of queue declare during leader elections would be the way I'd approach it.

gerhard:
If nodes get periodically shut-down and consumers get forcefully disconnect, when consumers reconnect they should first try to declare queues before consuming from them. Would that work?

michaelklishin:
@kjnilsson @dcorbacho FYI, this has a legitimate conflict with https://github.com/rabbitmq/rabbitmq-server/pull/2053.

PR #2052, opened 2019-07-02
Add rabbit_queue_type behaviour

kjnilsson:
First cut queue type behaviour

Uses the implementing module as the value of the amqqueue record
`type` field. This will allow for easy dispatch to the queue type
implementation.

Provides a stateful API around queue interactions. Currently the classic queue type will need some special handling when implementing processes that interact with queues but this should simplify this as well as the development of new queue types.


lukebakken:
Great idea :+1:

src/rabbit_channel.erl
```diff
@@ -1846,30 +1783,32 @@ consumer_monitor(ConsumerTag,
...
              delivering_queues = case NoAck of
                                      true  -> DQ;
                                      false -> sets:add_element(QRef, DQ)
                                  end}.
 
 handle_publishing_queue_down(QPid, Reason,
                              State = #ch{unconfirmed = UC,
-                                         queue_names = QNames})
+                                         queue_states = QStates})
   when ?IS_CLASSIC(QPid) ->
```
dcorbacho: These macros are not generic either. Should we have a `rabbit_queue_type:needs_handle_publishing_queue_down` or something similar?

The same applies to all other ocurrences of this and the `?IS_QUORUM` macro. Now that we have types they should disappear

src/amqqueue.erl
```diff
@@ -551,18 +551,16 @@ set_recoverable_slaves(#amqqueue{} = Queue, Slaves) ->
 set_recoverable_slaves(Queue, Slaves) ->
     amqqueue_v1:set_recoverable_slaves(Queue, Slaves).
 
-% quorum_nodes (new in v2)
+% type_state (new in v2)
 
--spec get_quorum_nodes(amqqueue()) -> [node()].
+-spec get_type_state(amqqueue()) -> map().
+get_type_state(#amqqueue{type_state = TState}) -> TState;
+get_type_state(_)                               -> [].
```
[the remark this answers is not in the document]
kjnilsson: nice catch - that is what it returned when it was just quorum nodes - the value is separated by so much whitespace it's quite hard to read :) - will fix

src/rabbit_channel.erl
```diff
@@ -737,72 +735,26 @@ handle_cast({reject_publish, MsgSeqNo, _QPid}, State = #ch{unconfirmed = UC}) ->
...
-                    erase_queue_stats(QName),
-                    noreply_coalesce(
-                      State3#ch{queue_states = maps:remove(Name, QueueStates),
-                                queue_names = maps:remove(Name, QNames)})
-            end;
-        _ ->
-            %% the assumption here is that the queue state has been cleaned up and
-            %% this is a residual Ra notification
-            noreply_coalesce(State0)
+handle_info({ra_event, From, _} = Evt,
```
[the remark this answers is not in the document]
kjnilsson: no it isn't - we can consider at some point make all queues send messages with a particular format - we'd have to make a change in Ra to allow custom ra event formatting to do so though

src/rabbit_channel.erl
kjnilsson: I agree and to do that we'd have to move all the confirm accounting inside the queue type module. I change I found a bit scary at this point.

kjnilsson: So having pondered this a bit more the current scope of the queue type abstraction isn't complete, it would need to handle confirms, downs etc as well to be considered so. Taht said I think at this point it achieves one goal: it makes it easier for us to deliver new queue types. The secondary goal of completing the queue type abstraction and making it easier for us to write new channel processes is one we can choose to defer at this point.

src/rabbit_queue_type.erl
```diff
@@ -0,0 +1,491 @@
...
+                 queue_state()) ->
+    {ok, queue_state(), actions()} | {error, term()}.
+
+%% any async events returned from the queue system should be processed through
+%% this
+-callback handle_event(Event :: term(),
+                       queue_state()) ->
+    {ok, queue_state(), actions()} | {error, term()} | eol.
+
+-callback deliver([{amqqueue:amqqueue(), queue_state()}],
```
dumbbell: After a bit more thoughts, I find this callback (and all others working on a list of queues) a bit more weird in fact compared to last week:
1. As I said, from an API PoV, this looks inconsistent: some callbacks take a list of queues, others a single queue, and the name doesn't reflect this change of behavior.
2. Should the queues in that list share the same queue type? What if a queue type implementation finds a queue from a different queue type in the middle of that list?

It seems to me that this feature should be outside of the new behavior and it should call an actual `deliver()` callback which takes a queue as argument. Again, my comment applies to other callbacks such as `recover()`.

src/rabbit_queue_type.erl
```diff
@@ -0,0 +1,491 @@
...
+-callback reject(rabbit_types:ctag(), Requeue :: boolean(),
+                 MsgIds :: [non_neg_integer()], queue_state()) ->
+    queue_state().
+
+-callback credit(rabbit_types:ctag(),
+                 non_neg_integer(), Drain :: boolean(), queue_state()) ->
+    queue_state().
+
+-callback dequeue(NoAck :: boolean(), LimiterPid :: pid(),
+                  rabbit_types:ctag(), queue_state()) ->
```
dumbbell: Why is the ctag third here?

include/amqqueue.hrl
```diff
@@ -51,16 +51,16 @@
...
           ?amqqueue_v1_field_state(Q) =:= State))).
 
--define(amqqueue_v1_type, classic).
+-define(amqqueue_v1_type, rabbit_classic_queue).
 
 -define(amqqueue_is_classic(Q),
         ((?is_amqqueue_v2(Q) andalso
-          ?amqqueue_v2_field_type(Q) =:= classic) orelse
+          ?amqqueue_v2_field_type(Q) =:= rabbit_classic_queue) orelse
          ?is_amqqueue_v1(Q))).
```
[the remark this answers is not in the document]
lukebakken: My goal was to put all "knowledge" of how to determine if an `amqqueue` is classic into this macro. This way, when the identification of a classic queue changes, we only have to change this macro and not all the places where it is used.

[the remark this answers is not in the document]
lukebakken: "Tell Don't Ask"

kjnilsson:
More work is needed to make this PR fit for purpose. The confirm logic needs to be moved inside the queue type abstraction and DOWN events need to be passed down to each queue type so that they can mutate their state and run logic as appropriate. 

#2062 has been created to ensure the breaking changes to the amqqueue record are included in 3.8 allowing us to work on this further and include it in a future release.

PR #2075, opened 2019-08-12
Make it possible to pre-configure cluster name via config

michaelklishin:
## Proposed Changes

This introduces a new boot step that [pre-]configures cluster name so that it is available to plugins that might use it, such as `rabbitmq_prometheus`.

## Types of Changes

## Checklist

## Further Comments

Per discussion with @gerhard.

### DIFF
--- a/docs/rabbitmq.conf.example
+++ b/docs/rabbitmq.conf.example
@@ -200,6 +200,10 @@
 # ssl_handshake_timeout = 5000
 
 
+## Cluster name
+##
+# cluster_name = dev3.eng.megacorp.local
+
 ## Password hashing implementation. Will only affect newly
 ## created users. To recalculate hash for an existing user
 ## it's necessary to update her password.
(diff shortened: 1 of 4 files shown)


gerhard:
- [ ] build custom generic unix
- [ ] add cluster name templating in `rabbitmq.config` in the BOSH release
- [ ] BOSH new deploy to GCP
- [ ] metrics from all nodes are tagged with the correct cluster name
- [ ] delete deployment
- [ ] repeat previous 3 steps a couple of times, ensure the results are consistent

gerhard:
Would it be worth adding this to the config examples?

* https://github.com/rabbitmq/rabbitmq-server/blob/master/docs/rabbitmq.conf.example
* https://github.com/rabbitmq/rabbitmq-server/blob/master/docs/rabbitmq.config.example

michaelklishin:
@gerhard done.

gerhard:
Would it be worth cherry-picking this to 3.7.x?

michaelklishin:
Since it's mostly needed by the new Prometheus exporter, are there any reason to cherry-pick?

gerhard:
> Since it's mostly needed by the new Prometheus exporter, are there any reason to cherry-pick?

I was thinking to back-port all rabbitmq_prometheus changes to prometheus_rabbitmq_exporter, so that users migrating from that plugin have a seamless upgrade path. Let's defer a cherry-pick until someone requests / complains about it.

Context: https://github.com/deadtrickster/prometheus_rabbitmq_exporter/pull/80#issuecomment-490987181

Issue #2012, opened 2019-05-16
A way for operators to attach a virtual host description


michaelklishin:
Currently in deployments with a lot of virtual hosts that do not follow any naming conventions
it is quite difficult to map a virtual host to an app/pod/set of apps/environment/what have you.

It would be useful to have a way to provide a description for a virtual host, and possibly even tags.

Per discussion with @mkuratczyk.

michaelklishin:
This turned out to be moderately difficult because it changes the schema of a distributed table, so we have to use a feature flag and a new virtual host field acceess module, à la `amqqueue` for queues. If this takes longer than a two day spike we'll close this issue as a `wontfix`.

PR #2080, opened 2019-08-14
Add virtual host description and tags

dcorbacho:
Part of https://github.com/rabbitmq/rabbitmq-server/issues/2012

Introduces 'description' and 'tags' fields on vhost record. The record has been made versioned and behind a feature flag for mixed version cluster compatibility.

## Types of Changes

## Checklist

### DIFF
--- a/include/vhost.hrl
+++ b/include/vhost.hrl
@@ -1,0 +1,6 @@
+-include("vhost_v1.hrl").
+-include("vhost_v2.hrl").
+
+-define(is_vhost(V),
+        (?is_vhost_v2(V) orelse
+         ?is_vhost_v1(V))).
--- a/include/vhost_v1.hrl
+++ b/include/vhost_v1.hrl
@@ -1,0 +1,4 @@
+-define(is_vhost_v1(V), is_record(V, vhost, 3)).
+
+-define(vhost_v1_field_name(V), element(2, V)).
+-define(vhost_v1_field_limits(V), element(3, V)).
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -894,7 +894,7 @@ total_queue_count() ->
     lists:foldl(fun (VirtualHost, Acc) ->
                   Acc + rabbit_amqqueue:count(VirtualHost)
                 end,
-                0, rabbit_vhost:list()).
+                0, rabbit_vhost:list_names()).
 
 %% TODO this only determines if the rabbit application has started,
 %% not if it is running, never mind plugins. It would be nice to have
@@ -1098,7 +1098,7 @@ insert_default_data() ->
     DefaultWritePermBin = rabbit_data_coercion:to_binary(DefaultWritePerm),
     DefaultReadPermBin = rabbit_data_coercion:to_binary(DefaultReadPerm),
 
-    ok = rabbit_vhost:add(DefaultVHostBin, ?INTERNAL_USER),
+    ok = rabbit_vhost:add(DefaultVHostBin, <<"Default virtual host">>, [], ?INTERNAL_USER),
     ok = lager_exchange_backend:maybe_init_exchange(),
     ok = rabbit_auth_backend_internal:add_user(
         DefaultUserBin,
--- a/src/rabbit_core_ff.erl
+++ b/src/rabbit_core_ff.erl
@@ -17,7 +17,8 @@
 -module(rabbit_core_ff).
 
 -export([quorum_queue_migration/3,
-         implicit_default_bindings_migration/3]).
+         implicit_default_bindings_migration/3,
+         virtual_host_metadata_migration/3]).
 
 -rabbit_feature_flag(
    {quorum_queue,
@@ -33,6 +34,13 @@
                        "being stored in the database",
       stability     => stable,
       migration_fun => {?MODULE, implicit_default_bindings_migration}
+     }}).
+
+-rabbit_feature_flag(
+   {virtual_host_metadata,
+    #{desc          => "Virtual host metadata (description, tags, etc)",
+      stability     => stable,
+      migration_fun => {?MODULE, virtual_host_metadata_migration}
      }}).
 
 %% -------------------------------------------------------------------
@@ -95,3 +103,19 @@ remove_explicit_default_bindings(FeatureName, Queues) ->
     [rabbit_binding:remove_default_exchange_binding_rows_of(Q)
      || Q <- Queues],
     ok.
+
+
+%% -------------------------------------------------------------------
+%% Virtual host metadata.
+%% -------------------------------------------------------------------
+
+virtual_host_metadata_migration(_FeatureName, _FeatureProps, enable) ->
+    Tab = rabbit_vhost,
+    rabbit_table:wait([Tab], _Retry = true),
+    Fun = fun(Row) -> vhost:upgrade_to(vhost_v2, Row) end,
+    case mnesia:transform_table(Tab, Fun, vhost:fields(vhost_v2)) of
+        {atomic, ok}      -> ok;
+        {aborted, Reason} -> {error, Reason}
+    end;
+virtual_host_metadata_migration(_FeatureName, _FeatureProps, is_enabled) ->
+    mnesia:table_info(rabbit_vhost, attributes) =:= vhost:fields(vhost_v2).
--- a/src/rabbit_vhost.erl
+++ b/src/rabbit_vhost.erl
@@ -17,16 +17,20 @@
 -module(rabbit_vhost).
 
 -include_lib("rabbit_common/include/rabbit.hrl").
-
-%%----------------------------------------------------------------------------
+-include("vhost.hrl").
 
 -export([recover/0, recover/1]).
--export([add/2, delete/2, exists/1, list/0, count/0, with/2, with_user_and_vhost/3, assert/1, update/2,
-         set_limits/2, limits_of/1, vhost_cluster_state/1, is_running_on_all_nodes/1, await_running_on_all_nodes/2]).
+-export([add/2, add/4, delete/2, exists/1, with/2, with_user_and_vhost/3, assert/1, update/2,
+         set_limits/2, vhost_cluster_state/1, is_running_on_all_nodes/1, await_running_on_all_nodes/2,
+        list/0, count/0, list_names/0, all/0, parse_tags/1]).
 -export([info/1, info/2, info_all/0, info_all/1, info_all/2, info_all/3]).
 -export([dir/1, msg_store_dir_path/1, msg_store_dir_wildcard/0]).
 -export([delete_storage/1]).
 -export([vhost_down/1]).
+
+%%
+%% API
+%%
 
 recover() ->
     %% Clear out remnants of old incarnation, in case we restarted
@@ -42,11 +46,11 @@ recover() ->
     %% So recovery will be run every time a vhost supervisor is restarted.
     ok = rabbit_vhost_sup_sup:start(),
 
-    [ok = rabbit_vhost_sup_sup:init_vhost(VHost) || VHost <- rabbit_vhost:list()],
+    [ok = rabbit_vhost_sup_sup:init_vhost(VHost) || VHost <- list_names()],
     ok.
 
 recover(VHost) ->
-    VHostDir = rabbit_vhost:msg_store_dir_path(VHost),
+    VHostDir = msg_store_dir_path(VHost),
     rabbit_log:info("Making sure data directory '~ts' for vhost '~s' exists~n",
                     [VHostDir, VHost]),
     VHostStubFile = filename:join(VHostDir, ".vhost"),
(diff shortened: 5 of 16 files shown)


michaelklishin:
This looks reasonable but we need to test it in a mixed version cluster scenario.

include/vhost.hrl
```diff
@@ -0,0 +1,30 @@
...
+        (?is_vhost_v2(V) orelse
+         ?is_vhost_v1(V))).
+
+-define(vhost_field_name(Q),
+        case ?is_vhost_v2(Q) of
+            true  -> ?vhost_v2_field_name(Q);
+            false -> case ?is_vhost_v1(Q) of
+                         true -> ?vhost_v1_field_name(Q)
+                     end
+        end).
```
dumbbell: In `amqqueue.hrl`, the macros are meant to be used in guard expressions. A `case` statement won't work in this case. Here are examples from that header:

```erlang
-define(amqqueue_is_durable(Q),
        ((?is_amqqueue_v2(Q) andalso
          ?amqqueue_v2_field_durable(Q) =:= true) orelse
         (?is_amqqueue_v1(Q) andalso
          ?amqqueue_v1_field_durable(Q) =:= true))).

-define(amqqueue_exclusive_owner_is(Q, Owner),
        ((?is_amqqueue_v2(Q) andalso
          ?amqqueue_v2_field_exclusive_owner(Q) =:= Owner) orelse
         (?is_amqqueue_v1(Q) andalso
          ?amqqueue_v1_field_exclusive_owner(Q) =:= Owner))).
```

Furthermore, function calls should be used otherwise:
1. That's the point of having a wrapper module, we can change that module and old code calling it still works. If people use macros to access the record (i.e. compile-time syntax like records), they are still screwed: they will have to recompile their code.
2. It's usually better to let the compiler decide what should be inlined or not.

include/vhost_v1.hrl
```diff
@@ -0,0 +1,5 @@
+-define(is_vhost_v1(V), is_record(V, vhost, 3)).
+
+-define(vhost_v1_field_name(V), element(2, V)).
+-define(vhost_v1_field_limits(V), element(2, V)).
+-define(vhost_v1_field_metadata(V), #{}).
```
dumbbell: This header represents exactly the record fields. `metadata` isn't part of that: compilation should fails if we try to access this field in an older vhost record.

include/vhost_v1.hrl
```diff
@@ -0,0 +1,5 @@
+-define(is_vhost_v1(V), is_record(V, vhost, 3)).
+
+-define(vhost_v1_field_name(V), element(2, V)).
+-define(vhost_v1_field_limits(V), element(2, V)).
```
dumbbell: `limits` has the wrong index here.

I believe we should re-use the same code to generate the header as it was done in the `amqqueue*` modules:

```erlang
macros() ->
    io:format(
      "-define(is_~s(Q), is_record(Q, amqqueue, ~b)).~n~n",
      [?record_version, record_info(size, amqqueue)]),
    %% The field number starts at 2 because the first element is the
    %% record name.
    macros(record_info(fields, amqqueue), 2).

macros([Field | Rest], I) ->
    io:format(
      "-define(~s_field_~s(Q), element(~b, Q)).~n",
      [?record_version, Field, I]),
    macros(Rest, I + 1);
macros([], _) ->
    ok.
```

src/rabbit.erl
```diff
@@ -466,6 +466,7 @@ start_it(StartFun) ->
                         false -> StartFun()
                     end
                 catch Class:Reason ->
+                    io:format("CATCH Class = '~s', Reason = '~p'", [Class, Reason]),
```
dumbbell: Debug printf left over?

src/rabbit_core_ff.erl
```diff
@@ -95,3 +103,19 @@ remove_explicit_default_bindings(FeatureName, Queues) ->
...
+virtual_host_metadata_migration(_FeatureName, _FeatureProps, enable) ->
+  Tab = rabbit_vhost,
+  rabbit_table:wait([Tab]),
+  Fun = fun vhost:upgrade/1,
+  case mnesia:transform_table(Tab, Fun, [virtual_host, limits, metadata]) of
+      {atomic, ok}      -> ok;
+      {aborted, Reason} -> {error, Reason}
+  end;
+virtual_host_metadata_migration(_FeatureName, _FeatureProps, is_enabled) ->
+  lists:member(metadata, mnesia:table_info(rabbit_vhost, attributes)).
```
dumbbell: Just a style comment: the rest of RabbitMQ uses 4-space indentation.

src/rabbit_core_ff.erl
```diff
@@ -95,3 +103,19 @@ remove_explicit_default_bindings(FeatureName, Queues) ->
...
     ok.
+
+
+%% -------------------------------------------------------------------
+%% Virtual host metadata.
+%% -------------------------------------------------------------------
+
+virtual_host_metadata_migration(_FeatureName, _FeatureProps, enable) ->
+  Tab = rabbit_vhost,
+  rabbit_table:wait([Tab]),
```
dumbbell: We must use the `rabbit_table:wait([Tab], _Retry = true)` (`Retry` defauls to `false`), otherwise the behavior is inconsistent during RabbitMQ startup: the general table wait already done by RabbitMQ retries. So if it takes time for the other nodes to come up, the migration will fail (no retry), but the general table wait will succeed (after retry), leading to inconsistent  state.

src/rabbit_core_ff.erl
```diff
@@ -95,3 +103,19 @@ remove_explicit_default_bindings(FeatureName, Queues) ->
...
+  Fun = fun vhost:upgrade/1,
```
dumbbell: We must be explicit with the target version of the record (in this migration function, we only know and care about version 2), so please use `vhost:upgrade_to(vhost_v2, Vhost)`. See `migrate_to_amqqueue_with_type()` for an example.

src/rabbit_core_ff.erl
```diff
@@ -95,3 +103,19 @@ remove_explicit_default_bindings(FeatureName, Queues) ->
...
+  case mnesia:transform_table(Tab, Fun, [virtual_host, limits, metadata]) of
```
dumbbell: Please use something like `amqqueue:fields(amqqueue_v2)` to query the list of fields, instead of hard-coding it.

src/rabbit_core_ff.erl
dumbbell: Likewise, instead of hard-coding the knowledge of `metadata`, please query the list of record fields and compare it to the table fields. Here is an example for the quorum queue migration:
```erlang
Fields = amqqueue:fields(amqqueue_v2),
mnesia:table_info(rabbit_queue, attributes) =:= Fields
```

src/rabbit_vhost.erl
```diff
@@ -42,7 +46,7 @@ recover() ->
     %% So recovery will be run every time a vhost supervisor is restarted.
     ok = rabbit_vhost_sup_sup:start(),
 
-    [ok = rabbit_vhost_sup_sup:init_vhost(VHost) || VHost <- rabbit_vhost:list()],
+    [ok = rabbit_vhost_sup_sup:init_vhost(VHost) || VHost <- rabbit_vhost:list_names()],
```
dumbbell: I suppose the compiler would see that, but we don't need to specify `rabbit_vhost:` here as it is a local function call.

src/rabbit_vhost.erl
```diff
@@ -61,37 +65,47 @@ recover(VHost) ->
...
+-spec add(vhost:name(), rabbit_types:username()) -> rabbit_types:ok_or_error(any()).
 
 add(VHost, ActingUser) ->
     case exists(VHost) of
         true  -> ok;
-        false -> do_add(VHost, ActingUser)
+        false -> do_add(VHost, <<"">>, <<"">>, ActingUser)
+    end.
+
+-spec add(vhost:name(), rabbit_types:username(), string(), [atom()]) -> rabbit_types:ok_or_error(any()).
```
dumbbell: Should tags be atoms? It would be dangerous and I don't see the need to use atoms here. Strings or binaries would be more practicle IMHO.

dumbbell: The order of arguments in the `-spec()` doesn't match the order in the function definition.

src/rabbit_vhost.erl
```diff
@@ -61,37 +65,47 @@ recover(VHost) ->
...
+    case exists(Name) of
+        true  -> ok;
+        false -> do_add(Name, Description, Tags, ActingUser)
     end.
 
-do_add(VHostPath, ActingUser) ->
-    rabbit_log:info("Adding vhost '~s'~n", [VHostPath]),
-    R = rabbit_misc:execute_mnesia_transaction(
+do_add(Name, Description, Tags, ActingUser) ->
+    rabbit_log:info("Adding vhost '~s' (description: '~s', tags: '~s')", [Name, Description, Tags]),
```
dumbbell: The type for tags in the `-spec()` doesn't match the type expected by this format string (list of atoms vs. a string), though I suppose `io:format()` will take this as an `iolist()` and silently concatenate all atoms.

src/rabbit_vhost.erl
```diff
@@ -213,92 +229,95 @@ assert_benign({error, {absent, Q, _}}, ActingUser) ->
...
+-spec list_names() -> [vhost:name()].
 
--spec list() -> [rabbit_types:vhost()].
+list_names() -> mnesia:dirty_all_keys(rabbit_vhost).
 
-list() ->
-    mnesia:dirty_all_keys(rabbit_vhost).
+-spec list() -> [vhost:name()].
+
+list() -> mnesia:dirty_match_object(rabbit_vhost, vhost:pattern_match_all()).
```
dumbbell: Can we keep the `list()` function unchanged, and add another function to list records? For the sake of keeping the API stable between 3.7.x and 3.8.x.

src/rabbit_vhost.erl
```diff
@@ -213,92 +229,95 @@ assert_benign({error, {absent, Q, _}}, ActingUser) ->
...
+limits_of(VHost) when is_binary(VHost) ->
+    assert(VHost),
+    case mnesia:dirty_read({rabbit_vhost, VHost}) of
         [] ->
-            mnesia:abort({no_such_vhost, VHostPath});
-        [#vhost{limits = Limits}] ->
-            Limits
+            mnesia:abort({no_such_vhost, VHost});
+        [VHost] ->
+            ?vhost_field_limits(VHost)
```
dumbbell: As said earlier in this review, please use a function call to the `vhost` module, not inlined code.

This comment is valid for all the uses of those macros in function bodies.

src/rabbit_vhost.erl
```diff
@@ -321,32 +339,38 @@ msg_store_dir_base() ->
...
-i(cluster_state, VHost) -> vhost_cluster_state(VHost);
-i(Item, _)        -> throw({bad_argument, Item}).
-
--spec info(rabbit_types:vhost()) -> rabbit_types:infos().
-
-info(VHost)        -> infos(?INFO_KEYS, VHost).
-
--spec info(rabbit_types:vhost(), rabbit_types:info_keys())
-                -> rabbit_types:infos().
+i(name,    VHost) -> ?vhost_field_name(VHost);
```
dumbbell: This is confusing:
* The old `i(name, VHost)` returned its second argument, meaning it was the name of the vhost, not the record. This is confirmed by the `-spec()` of the old `infos()` function which took a `rabbit_types:vhost()`.
* The new `i()` takes a record now.

src/vhost.erl
```diff
@@ -0,0 +1,167 @@
...
+-define(record_version, vhost_v2).
+
+-type(name() :: binary()).
+
+-type vhost() :: vhost_v1:vhost_v1() | vhost_v2().
+
+-type vhost_v2() :: #vhost{
+                          virtual_host :: rabbit_types:vhost(),
+                          limits :: list(),
+                          metadata :: map() | undefined
```
dumbbell: You can define it further by specifyfing the type of `description` and `tags` and mark them as optional.

I would define a dedicated type for `metadata` to be able to export it and use it in this file.

src/vhost.erl
```diff
@@ -0,0 +1,167 @@
...
+  get_tags/1,
+  set_limits/2
+]).
+
+-record(vhost, {
+          %% name as a binary
+          virtual_host,
+          %% proplist of limits configured, if any
+          limits,
+          metadata}).
```
dumbbell: The record should also be type-speced to cover the needs of `vhost_v2()` and `vhost_v2_pattern()`.

src/vhost.erl
```diff
@@ -0,0 +1,167 @@
...
+-export_type([name/0,
+              vhost/0,
+              vhost_v2/0,
+              vhost_pattern/0,
+              vhost_v2_pattern/0]).
+
+-spec new(name(), list()) -> vhost().
+new(Name, Limits) ->
+  case record_version_to_use() of
+      ?record_version ->
```
dumbbell: Style: this file uses a mix of 2-space and 4-space indentation.

src/vhost.erl
```diff
@@ -0,0 +1,167 @@
...
+    end.
+
+fields(?record_version) -> record_info(fields, vhost);
+fields(Version)         -> vhost_v1:fields(Version).
+
+info_keys() ->
+    case record_version_to_use() of
+        %% note: this reports description and tags separately even though
+        %% they are stored in the metadata map. MK.
+        ?record_version -> [name, description, tags, metadata, tracing, cluster_state];
```
dumbbell: Should we instead use `metadata` as the info key and let the caller take whatever it wants from the returned map?

include/vhost.hrl
michaelklishin: It was adopted from `amqqueue.hrl`:

``` erlang
-define(amqqueue_field_name(Q),
        case ?is_amqqueue_v2(Q) of
            true  -> ?amqqueue_v2_field_name(Q);
            false -> case ?is_amqqueue_v1(Q) of
                         true -> ?amqqueue_v1_field_name(Q)
                     end
        end).
```

The above approach returns a boolean, which is not the case with these macros. I have replaced them with functions.

src/vhost.erl
michaelklishin: We do report the metadata key but we also report `description` and `tags` so that formatting is easy on the CLI side.

src/rabbit_vhost.erl
michaelklishin: Done.

src/rabbit_vhost.erl
michaelklishin: User tags are atoms.

michaelklishin:
This is ready for another round.

src/vhost_v1.erl
```diff
@@ -0,0 +1,117 @@
...
+%%
+%% As of 3.6.x, the second argument is vhost limits,
+%% which is actually used and has the same default.
+%% Nonetheless, this required a migration, see rabbit_upgrade_functions.
+
+-record(vhost, {
+  %% name as a binary
+  virtual_host,
+  %% proplist of limits configured, if any
+  limits}).
```
dumbbell: Like in `vhost.erl`, the record type specs are missing.

src/vhost.erl
dumbbell: Record type specs are missing.

src/vhost.erl
```diff
@@ -0,0 +1,195 @@
...
+get_description(VHost) ->
+    vhost_v1:get_description(VHost).
+
+-spec get_tags(vhost()) -> [atom()].
+get_tags(#vhost{} = VHost) ->
+    maps:get(tags, get_metadata(VHost), undefined);
+get_tags(VHost) ->
+    vhost_v1:get_tags(VHost).
+
+-spec parse_tags(binary() | string() | atom()) -> [atom()].
```
dumbbell: This `parse_tags()` should be moved to e.g. `rabbit_vhost.erl`. The `vhost.erl` module, which could become `vhost_v2.erl` in the future if we introduce a newer version of the record, has a limited scope: it only provides functions to read and set the record's properties.

src/vhost.erl
```diff
@@ -0,0 +1,195 @@
...
+    case record_version_to_use() of
+      ?record_version ->
+        VHost#vhost{limits = Value};
+      _ ->
+        vhost_v1:set_limits(VHost, Value)
+    end.
+
+-spec trim_tag(list() | binary() | atom()) -> atom().
+trim_tag(Val) ->
+    rabbit_data_coercion:to_atom(string:trim(rabbit_data_coercion:to_list(Val))).
```
dumbbell: Why atoms? Do we absolutely need to use this type?

src/rabbit_vhost.erl
dumbbell: This error is still present. Furthermore, the CLI now passes a list of tags.

src/rabbit_vhost.erl
```diff
@@ -61,37 +65,47 @@ recover(VHost) ->
...
--define(INFO_KEYS, [name, tracing, cluster_state]).
-
--spec add(rabbit_types:vhost(), rabbit_types:username()) -> rabbit_types:ok_or_error(any()).
+-spec add(vhost:name(), rabbit_types:username()) -> rabbit_types:ok_or_error(any()).
 
 add(VHost, ActingUser) ->
     case exists(VHost) of
         true  -> ok;
-        false -> do_add(VHost, ActingUser)
+        false -> do_add(VHost, <<"">>, <<"">>, ActingUser)
```
dumbbell: The default empty tag doesn't match other uses: a binary vs. a list of tags.

src/vhost.erl
michaelklishin: User tags use atoms. I don't think there are major risks to that. We can switch to any type but I'd like a decision to be made first.

dcorbacho: I'm fine with the atoms as that is consistent with user tags. If that hasn't been a problem until now, this shouldn't be either - even if it has potential to be.

dumbbell: Ok for atoms then.

michaelklishin: Also note that users often have tags because that's how they gain management UI and monitoring API access. Virtual host tags are entirely advisory. I wouldn't be surprised if they weren't used in many environments.

michaelklishin:
Ready for  another round.

michaelklishin:
Initial mixed version cluster testing was a success.

PR #2088, opened 2019-08-28
rabbitmq-env.bat: Quote the expansion of %1 in `filter_path`

dumbbell:
Otherwise, paths containing parenthesis such as `C:/Program Files (x86)/...` get mis-interpreted and the function fails.

In my testing, the error was:
```
\Elixir\lib\elixir was unexpected
```

... based on the fact that Elixir is installed in `C:\Program Files (x86)\Elixir`.

While here, explicitely calls `rabbitmqctl.bat`, not `rabbitmqctl`. The latter might be available and picked in a Git clone.

### DIFF
--- a/scripts/rabbitmq-env.bat
+++ b/scripts/rabbitmq-env.bat
@@ -477,9 +477,9 @@ goto :eof
 
 :filter_path
 IF "%ERL_LIBS%"=="" (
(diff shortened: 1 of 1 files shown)


dumbbell:
Thank you :)

PR #2089, opened 2019-08-28
scripts/rabbitmq-server.bat: Use `werl.exe` when user wants a shell

dumbbell:
Compared to `erl.exe`, `werl.exe` has proper line editing and commands history, like Unix' `erl`.


dumbbell:
Thank you!

PR #2091, opened 2019-08-30
Remove use of ra.hrl

kjnilsson:
As this is an internal header now.
Needs: https://github.com/rabbitmq/ra/pull/115

### DIFF
--- a/src/rabbit_fifo.hrl
+++ b/src/rabbit_fifo.hrl
@@ -1,4 +1,6 @@
 
+-type option(T) :: undefined | T.
+
 -type raw_msg() :: term().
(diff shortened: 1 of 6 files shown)


src/rabbit_fifo.hrl
```diff
@@ -1,4 +1,6 @@
 
+-type maybe(T) :: undefined | T.
```
michaelklishin: We [already have](https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_types.erl#L36) this in `rabbit_types`.

kjnilsson: `maybe/1` in `rabbit_types` is declared `T | none` which is different

PR #2122, opened 2019-09-26
List possibles queue types for rebalance command

acogoluegnes:


michaelklishin:
I believe this effectively makes the line say that there are two arguments, `type` and `<all | quorum | classic>`. I suggest that we add a paragraph explaining what the accepted `type` values are instead.

michaelklishin:
We found no standard way to express an enumeration in `mdoc(7)`. So we will do what `git log` does for `--pretty`: a separate paragraph that explains what the accepted values are.

PR #2126, opened 2019-09-30
rabbitmqctl & rabbitmq-diagnostics manpage polishing

acogoluegnes:


michaelklishin:
Backported to `v3.8.x`.

PR #2127, opened 2019-09-30
fix health check returning ok in case of partition

luos:
Health check assumed if the result of partitions() is a list then everything
is ok, this is not the case the result is always a list.
rabbit_node_monitor:partitions() returns a non empty list in case there
is an mnesia_partition.

## Proposed Changes

The check is incorrect, I added a fix to check if the list is non-empty. 

I could not write a test case as in normal operation it's quite hard to introduce a partition, however for example the known suspend vm issue can cause it (or if it's left on ignore), for this reason this check is a must have.
If you have some pointers I'd gladly implement a test

Let me know what you think.

It returns errors like: 
```
{"status":"failed","reason":"cluster partition in effect: ['rabbit@rmq1.source.host',\n                             'rabbit@rmq3.source.host']"}
```

```
{"status":"failed","reason":"cluster partition in effect: ['rabbit@rmq2.source.host']"}
```

```
root@rmq1:/# rabbitmqctl --node rabbit@rmq2.source.host node_health_check
Timeout: 70 seconds ...
Checking health of node rabbit@rmq2.source.host ...
Error: healthcheck failed. Message: cluster partition in effect: ['rabbit@rmq1.source.host',
                             'rabbit@rmq3.source.host']
```

## Types of Changes

## Checklist

### DIFF
--- a/src/rabbit_health_check.erl
+++ b/src/rabbit_health_check.erl
@@ -62,8 +62,11 @@ node_health_check(list_queues) ->
 
 node_health_check(rabbit_node_monitor) ->
     case rabbit_node_monitor:partitions() of
-        L when is_list(L) ->
-            ok
+        [] ->
+            ok;
+        L when is_list(L), length(L) > 0 ->
+            ErrorMsg = io_lib:format("cluster partition in effect: ~p", [L]),
+            {error_string, ErrorMsg}
     end;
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you.

luos:
Thanks!

PR #2129, opened 2019-10-02
Adjust quorum queue flow control

kjnilsson:
Such that enqueuing channels will be put into flow slightly earlier
whilst aligning the quorum_commands_soft_limit with the credit flow
default will allow consuming channels to send acks slightly earlier.

Also slight state refactoring to separate the mutable from static parts of the rabbit_fifo_state.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,7 @@ define PROJECT_ENV
 	    %% see rabbitmq-server#143,
 	    %% rabbitmq-server#949, rabbitmq-server#1098
 	    {credit_flow_default_credit, {400, 200}},
-	    {quorum_commands_soft_limit, 256},
+	    {quorum_commands_soft_limit, 400},
 	    {quorum_cluster_size, 5},
 	    %% see rabbitmq-server#248
 	    %% and rabbitmq-server#667
--- a/src/rabbit_fifo_client.erl
+++ b/src/rabbit_fifo_client.erl
@@ -45,7 +45,7 @@
 
 -include_lib("rabbit_common/include/rabbit.hrl").
 
--define(SOFT_LIMIT, 256).
+-define(SOFT_LIMIT, 400).
 -define(TIMER_TIME, 10000).
 
 -type seq() :: non_neg_integer().
@@ -61,8 +61,14 @@
 -record(consumer, {last_msg_id :: seq() | -1,
                    delivery_count = 0 :: non_neg_integer()}).
 
--record(state, {cluster_name :: cluster_name(),
+-record(conf, {cluster_name :: cluster_name(),
                 servers = [] :: [ra:server_id()],
+                soft_limit = ?SOFT_LIMIT :: non_neg_integer(),
+                block_handler :: fun(() -> term()),
+                unblock_handler :: fun(() -> ok),
+                timeout :: non_neg_integer()}).
+
+-record(state, {conf :: #conf{},
                 leader :: undefined | ra:server_id(),
                 next_seq = 0 :: seq(),
                 %% Last applied is initialise to -1 to note that no command has yet been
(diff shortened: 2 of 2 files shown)


gerhard:
With the following PerfTest config `--producers 10 --consumers 1 --queue-args x-queue-type=quorum --auto-delete false --flag persistent --qos 500 --multi-ack-every 100` the  publishers are not slowed down until there are about 1 mil messages in the queue, at which point the memory alarm gets triggered and the consumer starts getting blocked intermittently and it never makes sufficient progress on the message backlog. It looks like this:

image

* [RabbitMQ Overview metrics](https://grafana.gcp.rabbitmq.com/dashboard/snapshot/zpvlffE2RjAXF7ASLeyyctlVw7I4hRDd?orgId=1)
* [RabbitMQ Raft metrics](https://grafana.gcp.rabbitmq.com/dashboard/snapshot/W4DCVS5XZ1aHPvA3t9eSHjUaeHolEkM3?orgId=1)

Will try the same without this PR.

gerhard:
We have a similar behaviour without the changes in this PR (basically 3.8.0), same PerfTest config as before:

image

* [RabbitMQ Overview metrics](https://grafana.gcp.rabbitmq.com/dashboard/snapshot/o4fYA3m01OdZGKAClMjI99lWWlYwdebK)
* [RabbitMQ Raft metrics](https://grafana.gcp.rabbitmq.com/d/f1Mee9nZz/rabbitmq-raft?orgId=1&var-rabbitmq_cluster=pr-2129&from=1570475571617&to=1570476320356)

This is an interesting perspective from the Erlang Distribution, Network, CPU & PerfTest viewpoint: [Erlang-Distribution-Compare metrics](https://grafana.gcp.rabbitmq.com/dashboard/snapshot/n0GSTk8oQjuGbOW7EicN6535Ddn2Rsku).

image

kjnilsson:
I never promised it will make a significant difference or change how flow control works (or doesn't). My testing showed that for _some_ scenarios it may be beneficial as in it pushed the number of publishers to consumers ratio up slightly before overrun became a reality.

gerhard:
With 5 publishers instead of 10 (everything else is the same), I am noticing a slower growth of message backlog, which is a good thing (blue is 3.8.0, green is this PR):

image

5 fast publishers are still able to overpower the 1 consumer. Trying with 3 publishers next.

gerhard:
With 3 publishers, the message backlog growth is slower, but they still overpower a single consumer:

image

Trying 2 publishers next.

gerhard:
2 publishers do not overpower a single consumer, which is good. Message backlog grows but remains stable:

image

The surprising part is that 3.8.0 without the PR behaves exactly the same:

image

I need to double-check that the 2 versions which I am comparing are indeed different, because as it currently stands, both generic-unix builds that I am running behave the same when PerfTest runs using the following config: --producers 2 --consumers 1 --queue-args x-queue-type=quorum --auto-delete false --flag persistent --qos 500 --multi-ack-every 100``

Vanlightly:
I just ran a test that I had previously run, except that each step lasted 30 minutes instead of 2-5 minutes as previous executions had. Using longer test time significantly reduced the publisher counts which have triggered the drop in throughput.

The test is to slowly increase publisher count, one-by-one. Add a publisher every 30 minutes. Ends at 10 publishers - a 5 hour test. Three parallel runs.

3.8.0 succumbed at publisher counts 3, 5 and 5
3.9.0-pr2129 succumbed at publisher counts 4, 7 and 7

So I definitely see an improvement with this PR, at the cost of the max throughput.

michaelklishin:
This optimizes for the pessimistic case. Should we leave things as is until there's more evidence this workload is common in the wild?

kjnilsson:
This PR has some good refactoring that could do with being merged
 - should I change it such that it no longer adjust the `quorum_command_soft_limit` dynamically and reset the default to 256 as before and then we can merge it just for the refactoring?

gerhard:
Yes, that sounds reasonable @kjnilsson

michaelklishin:
@kjnilsson sounds good. Note that this has been sitting unfinished for over a month.

kjnilsson:
@Vanlightly is doing some work around this which supersedes this work.

PR #2136, opened 2019-10-10
Update core metrics with down state

dcorbacho:
As the processes responsible to update `queue_metrics` are the queues
themselves, if they crash nothing else updates their metric state to down.
Thus, showing as running in the UI. This change uses the existing gc process
to check the state of the local queues and update it if required, while
it is doing the normal gc scanning. It covers all types of queues.

We consider a queue as down if the process/master/leader does no answer to
requests or it is dead. There could be other situations where a queue is
functionally down, but that is not covered here.

## Types of Changes

## Checklist


michaelklishin:
Backported to `v3.8.x`.

michaelklishin:
Reverted due to test and `gmake xref` failures on Concourse.

PR #2138, opened 2019-10-11
Optimise QQ memory use

kjnilsson:
Take fewer release cursor snapshots points as the message backlog grows.

Also introduces a compacted form of the internal message header map
where initially it is only an integer representing the size of the
message body. Later when additional keys need to be added it is expanded
into a full map. This avoid creating and holding many individial maps
with just a size element.

[#169064158]

### DIFF
--- a/src/rabbit_fifo.erl
+++ b/src/rabbit_fifo.erl
@@ -143,7 +143,14 @@ update_config(Conf, State) ->
                                competing
                        end,
     Cfg = State#?MODULE.cfg,
-    State#?MODULE{cfg = Cfg#cfg{release_cursor_interval = SHI,
+    SHICur = case State#?MODULE.cfg of
+                 #cfg{release_cursor_interval = {_, C}} ->
+                     C;
+                 #cfg{release_cursor_interval  = C} ->
+                     C
+             end,
+
+    State#?MODULE{cfg = Cfg#cfg{release_cursor_interval = {SHI, SHICur},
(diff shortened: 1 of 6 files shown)


src/rabbit_fifo.erl
```diff
@@ -974,26 +976,48 @@ append_to_master_index(RaftIdx,
     State#?MODULE{ra_indexes = Indexes}.
 
 
+incr_enqueue_count(#?MODULE{cfg = #cfg{release_cursor_interval = C} = Cfg}
+                   = State0)
+  when is_integer(C) ->
+    %% conversion to new release cursor interval format
+    State = State0#?MODULE{cfg = Cfg#cfg{release_cursor_interval = {C, C}}},
+    incr_enqueue_count(State);
```
dumbbell: This function clause could go after the next one:
* the most common clause appears first, which could be more readable for developpers
* this is a small optimization (the most common clause is evaluated first)

dumbbell: Likewise for other function clauses which do the conversion.

kjnilsson: Hmm yes that may actually apply to pretty much all "overloaded" functions. I'll update.

kjnilsson: That said the compiler might do this optimisation for us.

michaelklishin:
This was backported to `v3.8.x`.

PR #2141, opened 2019-10-15
Reserve file handles for quorum queues

dcorbacho:
These should be taken into account into the limits, but always be granted.
Three file handles are reserved on startup for the wal, segment writer and a DETS table,
the rest must be reserved by the queues themselves using `set_reservation/0` or
`set_reservation/1`.

Depends on https://github.com/rabbitmq/rabbitmq-common/pull/336

## Types of Changes

## Checklist

### DIFF
--- a/src/rabbit_fifo.erl
+++ b/src/rabbit_fifo.erl
@@ -574,8 +574,9 @@ state_enter(eol, #?MODULE{enqueuers = Enqs,
     [{send_msg, P, eol, ra_event}
      || P <- maps:keys(maps:merge(Enqs, AllConsumers))] ++
         [{mod_call, rabbit_quorum_queue, file_handle_release_reservation, []}];
-state_enter(_, #?MODULE{cfg = #cfg{resource = _Resource} }) ->
-    [{mod_call, rabbit_quorum_queue, file_handle_other_reservation, []}];
+state_enter(State, #?MODULE{cfg = #cfg{resource = _Resource}}) when State =/= leader ->
+    FHReservation = {mod_call, rabbit_quorum_queue, file_handle_other_reservation, []},
+    [FHReservation];
  state_enter(_, _) ->
     %% catch all as not handling all states
     [].
--- a/src/rabbit_fifo.erl
+++ b/src/rabbit_fifo.erl
@@ -541,6 +541,7 @@ state_enter(leader, #?MODULE{consumers = Cons,
                              enqueuers = Enqs,
                              waiting_consumers = WaitingConsumers,
                              cfg = #cfg{name = Name,
+                                        resource = Resource,
                                         become_leader_handler = BLH},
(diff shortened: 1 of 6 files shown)


michaelklishin:
@dcorbacho @kjnilsson I have noticed two test failures in `rabbit_fifo_SUITE` and tried to address them. However, one clause in `rabbit_fifo:state_enter/2` does not make sense to me, so my changes to it badly need a second pair of 👀.

src/rabbit_fifo.erl
```diff
@@ -574,8 +574,9 @@ state_enter(eol, #?MODULE{enqueuers = Enqs,
     [{send_msg, P, eol, ra_event}
      || P <- maps:keys(maps:merge(Enqs, AllConsumers))] ++
         [{mod_call, rabbit_quorum_queue, file_handle_release_reservation, []}];
-state_enter(_, #?MODULE{cfg = #cfg{resource = _Resource} }) ->
-    [{mod_call, rabbit_quorum_queue, file_handle_other_reservation, []}];
+state_enter(State, #?MODULE{cfg = #cfg{resource = _Resource}}) when State =/= leader ->
```
dcorbacho: This should be the catch all clause, as `leader` and `eol` are already handled. Why do we need the next one? Am I missing something?

michaelklishin: only because not  having  a catch-all  clause seemed suspicious.

michaelklishin: It turned out that the clause was used by `state_enter(recover, State)`. Without it all `ct-dead_lettering` QQ tests fail, for example. @kjnilsson @dcorbacho I've left it in, not sure if explicit handling would be worth doing.

kjnilsson: Right so at the recover stage the State is `undefined` because, well, it hasn't recovered the state yet! Incidentally it will need a file handle reservation for recover so perhaps we should handle this case explicitly. I have to admit I didn't realise the state was `undefined` at recover. This should be clarified in the callback type.

kjnilsson: oh but I realise we can't reserve at recover as we do not have a queue name available.

kjnilsson: Ah looking closer it _should_ have the state available at that point as it does read the snapshot or generates the initial state earlier. Ok will need to dig further

kjnilsson: I just ran ct-dead_lettering without the catch-all and it passed ok so perhaps something else is going on.

dcorbacho: Same here, not failing either

kjnilsson: Apart from that this PR looks fine to me

michaelklishin:
Backported to `v3.8.x`.

PR #2154, opened 2019-10-30
Improve error handling and logging around quorum queues

dcorbacho:
Some quorum queue queries might return an error or timeout, which leads to a crash of the channel. While this is the expected behaviour, it can be handled better by returning an internal error instead of crashing with a `badmatch` error which polutes the logs and is harder to understand for users.

## Types of Changes

## Checklist

### DIFF
--- a/src/rabbit_amqqueue.erl
+++ b/src/rabbit_amqqueue.erl
@@ -1592,7 +1592,7 @@ basic_consume(Q,
             {ok, maps:put(Name, QState, QStates)};
         {error, Reason} ->
             rabbit_misc:protocol_error(internal_error,
-                                       "Cannot consume a message from quorum queue '~s': ~p",
+                                       "Cannot consume a message from quorum queue '~s': ~w",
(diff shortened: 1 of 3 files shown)


src/rabbit_amqqueue.erl
```diff
@@ -1582,13 +1582,19 @@ basic_consume(Q,
...
+                                           ConsumerPrefetchCount,
+                                           ConsumerTag,
+                                           ExclusiveConsume, Args,
+                                           ActingUser,
+                                           OkMsg, QState0) of
+        {ok, QState} ->
+            {ok, maps:put(Name, QState, QStates)};
+        {error, Reason} ->
+            rabbit_misc:protocol_error(internal_error,
+                                       "Cannot consume a message from quorum queue '~s': ~p",
```
kjnilsson: I would use `~W` for the reason here rather than `~p`

michaelklishin:
Backported to `v3.8.x`.

PR #2155, opened 2019-10-31
rabbit_reader: Format `Explanation` before passing it

dumbbell:
... to `rabbit_misc:amqp_error()`.

`Explanation` can be a user input. Therefore, we don't want to pass it as a format string to `rabbit_misc:amqp_error()`, as we can't trust its content.

Now, we pass our own format string (`"~s"`) and `Explanation` becomes the argument to that format string. This ensures we don't interpret untrusted user input.

Discussed with: @dcorbacho, @essen and @michaelklishin.


michaelklishin:

PR #2162, opened 2019-11-13
Set virtual host metadata defaults on upgrade

### LINKED ISSUE #2161: Virtual host metadata are created empty on upgrade
When upgrading from 3.7 to 3.8 and enabling the `virtual_host_metadata`, all the existing virtual hosts records are upgraded with an empty metadata map. The system assume at least 2 keys inside the metadata map: `description` and `tags`.

Not providing default values for these 2 can provoke weird issues anywhere a value is assumed. For example, [Hop would fail](https://github.com/rabbitmq/hop/commit/1119a9abba71839993f50addc9494a950e40f0e4) because it expects an array for tags and it gets `"undefined"` after an upgrade.

To reproduce (the easiest is to use the umbrella from `master`, in the `rabbitmq_server_release` directory):

```
export RABBITMQ_FEATURE_FLAGS=
make run-broker PLUGINS='rabbitmq_management'
...
../rabbit/scripts/rabbitmqctl enable_feature_flag virtual_host_metadata
./rabbit/scripts/rabbitmqctl list_vhosts name,description,tags        
Listing vhosts ...
name	description	tags
/	undefined	undefined
../rabbit/scripts/rabbitmqctl add_vhost test
Adding vhost "test" ...
../rabbit/scripts/rabbitmqctl list_vhosts name,description,tags                         
Listing vhosts ...
name	description	tags
/	undefined	undefined
test		[]
```

The http://localhost:15672/api/vhosts endpoint returns:

```json
[
  {
    "cluster_state": {
      "rabbit@acogoluegnes-inspiron": "running"
    },
    "description": "undefined",
    "metadata": {
      
    },
    "name": "/",
    "tags": "undefined",
    "tracing": false
  },
  {
    "cluster_state": {
      "rabbit@acogoluegnes-inspiron": "running"
    },
    "description": "",
    "metadata": {
      "description": "",
      "tags": [
        
      ]
    },
    "name": "test",
    "tags": [
      
    ],
    "tracing": false
  }
]
```

Note the `undefined` in the upgraded `/` virtual host, for both the console output and the JSON output.

A more appropriate behavior would be to have the same empty metadata values than a newly created virtual host.

acogoluegnes:
[#169700677]

Fixes #2161

### DIFF
--- a/src/vhost.erl
+++ b/src/vhost.erl
@@ -117,7 +117,7 @@ upgrade(OldVHost)         -> upgrade_to(record_version_to_use(), OldVHost).
 upgrade_to(?record_version, #vhost{} = VHost) ->
     VHost;
 upgrade_to(?record_version, OldVHost) ->
-    Fields = erlang:tuple_to_list(OldVHost) ++ [#{}],
+    Fields = erlang:tuple_to_list(OldVHost) ++ [#{description => <<"">>, tags => []}],
     #vhost{} = erlang:list_to_tuple(Fields);
 upgrade_to(Version, OldVHost) ->
     vhost_v1:upgrade_to(Version, OldVHost).


michaelklishin:
Backported to `v3.8.x`.

Issue #2168, opened 2019-11-20
MQTT, STOMP performance degradation due to sync call to the connection process


velimir:
Hi,

MQTT Plugin (as well as STOMP and others) uses direct channels to communicate with rabbit.

Authorisation introduces synchronous call to the connection process for every message published on a channel to obtain amqp parameters used to connect to the broker (`rabbit_amqp_connection:amqp_params/2`): [here](https://github.com/rabbitmq/rabbitmq-server/blob/0991bbf81cbdb04e0ea91d019a82c70d96e90213/src/rabbit_channel.erl#L1334) and [here](https://github.com/rabbitmq/rabbitmq-server/blob/0991bbf81cbdb04e0ea91d019a82c70d96e90213/src/rabbit_channel.erl#L1337).

Original performance:

<img width="1327" alt="Screenshot 2019-11-19 at 16 06 44" src="https://user-images.githubusercontent.com/709160/69230733-be51b780-0b7f-11ea-8e4a-580da0c5dd3f.png">

Performance w/o sync call:

<img width="1334" alt="Screenshot 2019-11-19 at 16 41 47" src="https://user-images.githubusercontent.com/709160/69230771-c7428900-0b7f-11ea-8e0a-8a2c9b5a452f.png">

Would it be feasible to provide connection parameters (variable map) as a required parameter for the `rabbit_channel` to void the call to the connection process?
Would you be able advise on why it can't be done w/o call to the connection process?

michaelklishin:
It can be done w/o a call to the connection process. The sync call has bitten us elsewhere (e.g. #1886) and something we should replace. It sounds doable but this API area is sensitive for cross-version Federation.

michaelklishin:
PRs contributed by @velimir so far:

 * https://github.com/rabbitmq/rabbitmq-server/pull/2169
 * https://github.com/rabbitmq/rabbitmq-erlang-client/pull/127 (this is the potentially breaking part for Federation)
 * https://github.com/rabbitmq/rabbitmq-management/pull/755

velimir:
@michaelklishin thanks for looking into this. That's great. I was afraid that there's something that I'm missing and the call is unavoidable.

 I should be able to add tests and fix broken ones later next week.

michaelklishin:
I tested Federation with 3.6 and 3.7 upstreams as well as 3.6 and 3.7 downstreams. No issues with exchange federation which uses direct client.

[...]

PR #2169, opened 2019-11-20
avoid synchronous call to a connection process from a channel process

velimir:
Performance optimisation, see #2168 and #1886 for background.

## Proposed Changes

Pass amqp parameters for a direct channel as part of rabbit_channel start up arguments.

## Types of Changes

- [X] Breaking change (fix or feature that would cause an observable behavior change in existing systems)

## Checklist

## Further Comments

No tests except performance/smoke tests were performed. PR just to clarify a proposed change/issue: #2168

### DIFF
--- a/src/rabbit_channel.erl
+++ b/src/rabbit_channel.erl
@@ -55,7 +55,7 @@
 
 -behaviour(gen_server2).
 
--export([start_link/11, do/2, do/3, do_flow/3, flush/1, shutdown/1]).
+-export([start_link/12, do/2, do/3, do_flow/3, flush/1, shutdown/1]).
 -export([send_command/2, deliver/4, deliver_reply/2,
          send_credit_reply/2, send_drained/2]).
 -export([list/0, info_keys/0, info/1, info/2, info_all/0, info_all/1,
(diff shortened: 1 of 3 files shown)


velimir:
Feels like `#conf.source` is no longer needed in the state of the channel.

michaelklishin:
Channel source was introduced in https://github.com/rabbitmq/rabbitmq-server/pull/1886. I have removed it and its related functions and running various test suites now. @velimir @acogoluegnes  can you please double check that this change does not undo what #1886 was introduced to address? A couple more reviewers would not hurt.

velimir:
My understanding that #1886 was introduced mainly to avoid call to the connection process when channel is instantiated by a network client: https://github.com/rabbitmq/rabbitmq-server/pull/1886/files#diff-98224c82bfd7f4172da6a5adff2c42b0R1026

Since this PR removes the call for all connections (not only those which are spawned by `rabbit_reader`) it covers the case of the mentioned PR, however I would run performance tests for AMQP clients just to be sure.

michaelklishin:
Superseded by #2172.

PR #2172, opened 2019-11-23
A continuation to 2169

michaelklishin:
This is a continuation to #2169 by @velimir. It avoids a synchronous call from channels to their connection which introduces unnecessary overhead (#2168).

Previously this was partially solved in #1886.


michaelklishin:
Backported to `v3.8.x`.

PR #2179, opened 2019-12-05
rabbit_{amqueue,quorum_queue}: add several listing and sampling functions

michaelklishin:
So that node-local queues that satisfy certain criteria
can be listed easily.

Mob: @kjnilsson and several other Pivots.

Part of rabbitmq/rabbitmq-cli#389.


michaelklishin:
Will resubmit with a different branch name.

PR #2182, opened 2019-12-11
make garbage collector threshold configurable

nyczol:
This PR gives possibility to configure gc threshold - how often gc will be run (if any).

Please also check:
https://github.com/rabbitmq/rabbitmq-common/pull/343


michaelklishin:
As mentioned in https://github.com/rabbitmq/rabbitmq-common/pull/343, we'd like to see some data on the effects of this on both workloads with a lot of small messages and fewer really large (say, 120-128 MB) ones.

michaelklishin:

PR #2194, opened 2019-12-26
Broker can resume work when "recovery.dets" is damaged.

tomyouyou:
When the file is damaged, for example, because the disk is full, the service fails to return to normal after the disk space is enough.
The PR is to delete the file,try to recover all queues without the use of the recovery hints, and regain availability after that.

To reproduce the issue:

[root@pc4 test]# python msg_test.py create_queue 10.43.178.220 'queue=q11;durable=true'
connect to 10.43.178.220 ok
2019-12-26 09:18:10.418 create_queue q11
2019-12-26 09:18:10.421 to close connection
[root@pc4 test]# python msg_test.py create_exchange 10.43.178.220 'exchange=ex11;exchange_type=topic'
connect to 10.43.178.220 ok
2019-12-26 09:18:10.604 create_exchange ex11 topic
2019-12-26 09:18:10.606 to close connection
[msg_test.txt](https://github.com/rabbitmq/rabbitmq-server/files/4001130/msg_test.txt)

[root@pc4 test]# python msg_test.py bind 10.43.178.220 'exchange=ex11;queue=q11;routing_key=exq11'
connect to 10.43.178.220 ok
2019-12-26 09:18:10.778 bind ex11 q11 r_key:exq11
2019-12-26 09:18:10.782 to close connection
[root@pc4 test]# python msg_test.py publish 10.43.178.220 'exchange=ex11;routing_key=exq11;msg="_unique_id:123-10";persistent=true;count=1'
connect to 10.43.178.220 ok
2019-12-26 09:18:11.605 publish ex11 r_key:exq11, 0
2019-12-26 09:18:12.645 to close connection

[root@pc4 SOURCES]# rabbitmqctl list_queues
Timeout: 60.0 seconds ...
Listing queues for vhost / ...
name    consumers       messages_ready  messages_unacknowledged
q11     0       1       0

[root@pc4 home]# systemctl stop rabbitmq-server
%%%%%% to destroy the "recovery.dets" by vim tool
[root@pc4 home]#  vim /var/lib/rabbitmq/mnesia/rabbit@pc4/msg_stores/vhosts/628WB79CIFDYO9LJI6DKMI09L/recovery.dets
[root@pc4 home]# systemctl start rabbitmq-server

%%%%% crash log in broker:
[...]

PR #2195, opened 2019-12-26
Handle corrupt or empty virtual host recovery term files 

michaelklishin:
This is a reworked version of #2194 by @tomyouyou.


michaelklishin:

PR #2197, opened 2019-12-29
do not check msg size for undefined gc threshold

nyczol:
Simple improvment to
https://github.com/rabbitmq/rabbitmq-server/pull/2182
For undefined gc threshold we do not need to call rabbit_basic:maybe_gc_large_msg
This PR should be merged into master and v3.8.x branch


michaelklishin:
#2199 is more confusing, there's already a clause where `GCThreshold` is `undefined`.

michaelklishin:
Applied manually, backported to `v3.8.x`.

nyczol:
Still no commit in v3.8.x

PR #2212, opened 2020-01-16
rabbit_plugins: Filter non-plugins out

dumbbell:
When the module discovers plugins in the plugins directories, it considers all applications there. There are actual RabbitMQ plugins, their dependencies, and possibly other applications which are neither
plugins nor dependencies.

So far, this was the case of `syslog` which could be used by Lager if the user configures it. To make sure `rabbit_plugins` or rabbitmq-plugins(8) didn't mess with `syslog`, there was a hard-coded filtering function which excluded this specific application.

In RabbitMQ Enterprise Edition, we want to add another application of this kind: `inet_tcp_compress_dist`. This application implements an Erlang distribution protocol, so it is effectively unrelated to RabbitMQ. However, to keep packaging simple, it is provided in the plugins
directory.

We could add it to the hard-coded filtering function, but instead this function is replaced by the following logic to filter plugins:
  * Applications which depend on `rabbit` are considered actual plugins.
  * Plugins' dependencies, direct and indirect, are considered plugins as well.

Therefore, `rabbit_plugins` will ignore:
  * Applications which don't depend on `rabbit` and no plugins depend on them (again, directly or indirectly)
  * Applications which are part of Erlang/OTP.

This means the behavior slightly changes: a plugin which didn't declare its dependency to `rabbit` will not be considered a plugin anymore. I consider it is a bug in the plugin in the first place.

The code will log the following debug messages to list ignored applications:

```
2020-01-14 13:18:41.468 [debug] <0.941.0> Plugins discovery: ignoring getopt, not a RabbitMQ plugin
2020-01-14 13:18:41.468 [debug] <0.941.0> Plugins discovery: ignoring inet_tcp_compress_dist, not a RabbitMQ plugin
2020-01-14 13:18:41.468 [debug] <0.941.0> Plugins discovery: ignoring lz4, not a RabbitMQ plugin
2020-01-14 13:18:41.468 [debug] <0.941.0> Plugins discovery: ignoring syslog, not a RabbitMQ plugin
2020-01-14 13:18:41.468 [debug] <0.941.0> Plugins discovery: ignoring zstd, not a RabbitMQ plugin
```

### DIFF
--- a/src/rabbit_plugins.erl
+++ b/src/rabbit_plugins.erl
@@ -225,14 +225,6 @@ is_plugin_provided_by_otp(#plugin{name = eldap}) ->
 is_plugin_provided_by_otp(_) ->
     false.
 
-is_skipped_plugin(#plugin{name = syslog}) ->
-    % syslog is shipped as an .ez file, but it's not an actual plugin and
-    % it's not a direct dependency of the rabbit application, so we must
-    % skip it here
-    true;
-is_skipped_plugin(_) ->
-    false.
-
 %% Make sure we don't list OTP apps in here, and also that we detect
 %% missing dependencies.
 ensure_dependencies(Plugins) ->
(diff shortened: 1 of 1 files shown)


dumbbell:
Once merged and CI'd, this change should be backported to 3.8.x and 3.7.x to keep the behavior consistent accross branches.

gerhard:
I have already tested this in the context of `rabbitmq-enterprise-3.8.3-r2`, part of https://github.com/rabbitmq/inet_tcp_compress_dist/pull/5. I agree that we should backport this. Will have a go at cherry-picking this merge into 3.8.x & 3.7.x.

cc @michaelklishin for visibility

dumbbell:
Thank you!

dumbbell:
Backported to `v3.7.x` and `v3.8.x`.

Issue #2180, opened 2019-12-05
Move rabbitmq-server(8) scripts to Erlang code


dumbbell:
Today, to start RabbitMQ, we **depend on several Bourne shell scripts** (for Unix-based OSes) **and Batch scripts** (for Microsoft Windows). They have several downsides:
* The code is duplicated between the two languages used for those scripts.
* Maintenance is obviously increased.
* Debugging what happens before the Erlang VM is launched is difficult.
* The Batch version for Windows gets less testing and may not have the same features as the Bourne shell version.

From an Erlang/OTP point of view, **RabbitMQ is not a regular Erlang/OTP application** because it relies on several things happening before the actual `rabbit` application can be started. This complicates developers' life because RabbitMQ does not embrace Erlang/OTP system principles and a few things are reinvented internally.

We want to move most of the scripts to plain Erlang code and change RabbitMQ so it is built as a standard Erlang/OTP application.

The [initial description of the problem was written down in a gist](https://gist.github.com/dumbbell/b568d5ca70bb28873af683b4926d562d).

dumbbell:
All pull requests were QA's and merged.

michaelklishin:
This was backported for `3.8.4`.

PR #2219, opened 2020-01-23
rabbit: Change start_apps/2 to call application:ensure_all_started/2

dumbbell:
.. instead of reinventing about the same logic.

start_apps/2 is not called to start the `rabbit` application anymore (it is started as a normal Erlang application now). It means it's only used to start plugins which are also supposed to be regular Erlang
applications.

It runs boot steps unconditionally as well (because this function never starts `rabbit).

One difference with the previous code is that the previous code started all loaded applications. This seemed unnecessary and even caused issues with testsuites where some applications were loaded like `rabbitmqctl` but were not meant to be started.

Another difference is that before, all boot steps were executed at once, then all plugins were started. Now, to improve consistency and make sure that a dependency is fully ready before a plugin which depends on it starts, for each application we execute the boot steps then start it, before moving to the next.

### DIFF
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -599,21 +599,27 @@ start_apps(Apps) ->
 -spec start_apps([app_name()],
                  #{app_name() => restart_type()}) -> 'ok'.
 
+%% TODO: start_apps/2 and start_loaded_apps/2 are now specific to
+%% plugins. Those function should be moved over `rabbit_plugins`, along
+%% with top_apps/1, once the latter stops using app_utils as well.
+
 start_apps(Apps, RestartTypes) ->
     app_utils:load_applications(Apps),
     ok = rabbit_feature_flags:refresh_feature_flags_after_app_load(Apps),
     start_loaded_apps(Apps, RestartTypes).
 
 start_loaded_apps(Apps, RestartTypes) ->
+    false = lists:member(rabbit, Apps), %% Assertion.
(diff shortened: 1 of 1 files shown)


dumbbell:
The failure I mention in the commit happens all the time in Travis CI for instance. I couldn't reproduce it locally however. Perhaps there is a timing issue somewhere, I don't know... Anyway, I looked at the history of that code starting applications back to May 2012 but no commits explain the reason behind it.

dumbbell:
Currently, the `start_apps/2` function (both before and after this patch) runs all boot steps it finds, then starts all applications.

We want to change that to a loop where, for each application, it runs the boot steps for this application, then starts it (and continues with the next application). This should be more consistent w.r.t. what a particular application can expect. For instance, a plugin's boot steps could rely on the fact dependencies will always be running. This is not the case currently, depending on how dependencies are started.

Also, we found an issue, introduced with rabbitmq/rabbitmq-server#2180, with plugins' boot steps executed as part of RabbitMQ startup: a plugin's boot steps are executed after the plugin is started.

michaelklishin:
@dumbbell this does not backport cleanly, I'll assume this is because the `rabbit_env` and `rabbit_prelaunch` transition hasn't been backported so far, so those changes have to be backported together.

dumbbell:
Yes, that's correct. This patch is linked to the "master" issue (#2180) so we remember to backport it as well.

Thank you for merging it!

PR #2231, opened 2020-02-05
Convert systemd notification to prelaunch steps

pjk25:
Refactor to allow an arbitrary number of listeners to boot state changes, in a manner where each can maintain their own private state if necessary. systemd integration is the first - support is checked on startup, and it exits quietly if not, and afterward dropped from the list of listeners.

### DIFF
--- a/apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
+++ b/apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
@@ -1,0 +1,84 @@
+%%%-------------------------------------------------------------------
+%% The contents of this file are subject to the Mozilla Public License
+%% Version 1.1 (the "License"); you may not use this file except in
+%% compliance with the License. You may obtain a copy of the License
+%% at https://www.mozilla.org/MPL/
+%%
+%% Software distributed under the License is distributed on an "AS IS"
+%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
+%% the License for the specific language governing rights and
+%% limitations under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% Copyright (c) 2019-2020 Pivotal Software, Inc.  All rights reserved.
+%%
+
+-module(rabbit_boot_state).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-export([get/0,
+         set/1,
+         wait_for/2]).
+
+-define(PT_KEY_BOOT_STATE,    {?MODULE, boot_state}).
+
+-type boot_state() :: 'stopped' | 'booting' | 'ready' | 'stopping'.
+
+-export_type([boot_state/0]).
+
+-spec get() -> boot_state().
+get() ->
+    persistent_term:get(?PT_KEY_BOOT_STATE, stopped).
+
+-spec set(boot_state()) -> ok.
+set(BootState) ->
+    rabbit_log_prelaunch:debug("Change boot state to `~s`", [BootState]),
+    ?assert(is_valid(BootState)),
+    case BootState of
+        stopped -> persistent_term:erase(?PT_KEY_BOOT_STATE);
+        _       -> persistent_term:put(?PT_KEY_BOOT_STATE, BootState)
+    end,
+    rabbit_boot_state_sup:notify_boot_state_listeners(BootState).
+
+-spec wait_for(boot_state(), timeout()) -> ok | {error, timeout}.
+wait_for(BootState, infinity) ->
+    case is_reached(BootState) of
+        true  -> ok;
+        false -> Wait = 200,
+                 timer:sleep(Wait),
+                 wait_for(BootState, infinity)
+    end;
+wait_for(BootState, Timeout)
+  when is_integer(Timeout) andalso Timeout >= 0 ->
+    case is_reached(BootState) of
+        true  -> ok;
+        false -> Wait = 200,
+                 timer:sleep(Wait),
+                 wait_for(BootState, Timeout - Wait)
+    end;
+wait_for(_, _) ->
+    {error, timeout}.
+
+boot_state_idx(stopped)  -> 0;
+boot_state_idx(booting)  -> 1;
+boot_state_idx(ready)    -> 2;
+boot_state_idx(stopping) -> 3.
+
+is_valid(BootState) ->
+    is_integer(boot_state_idx(BootState)).
+
+is_reached(TargetBootState) ->
+    is_reached(?MODULE:get(), TargetBootState).
+
+is_reached(CurrentBootState, CurrentBootState) ->
+    true;
+is_reached(stopping, stopped) ->
+    false;
+is_reached(_CurrentBootState, stopped) ->
+    true;
+is_reached(stopped, _TargetBootState) ->
+    true;
+is_reached(CurrentBootState, TargetBootState) ->
+    boot_state_idx(TargetBootState) =< boot_state_idx(CurrentBootState).
--- a/apps/rabbitmq_prelaunch/src/rabbit_boot_state_sup.erl
+++ b/apps/rabbitmq_prelaunch/src/rabbit_boot_state_sup.erl
@@ -1,0 +1,46 @@
+%%%-------------------------------------------------------------------
+%% The contents of this file are subject to the Mozilla Public License
+%% Version 1.1 (the "License"); you may not use this file except in
+%% compliance with the License. You may obtain a copy of the License
+%% at https://www.mozilla.org/MPL/
+%%
+%% Software distributed under the License is distributed on an "AS IS"
+%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
+%% the License for the specific language governing rights and
+%% limitations under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% Copyright (c) 2020 Pivotal Software, Inc.  All rights reserved.
+%%
+
+-module(rabbit_boot_state_sup).
+-behaviour(supervisor).
+
+-export([start_link/0,
+         init/1]).
+
+-export([notify_boot_state_listeners/1]).
+
+start_link() ->
+    supervisor:start_link({local, ?MODULE}, ?MODULE, []).
+
+init([]) ->
+    SystemdSpec = #{id => rabbit_boot_state_systemd,
+                    start => {rabbit_boot_state_systemd, start_link, []},
+                    restart => transient},
+    {ok, {#{strategy => one_for_one,
+            intensity => 1,
+            period => 5},
+          [SystemdSpec]}}.
+
+-spec notify_boot_state_listeners(rabbit_boot_state:boot_state()) -> ok.
+notify_boot_state_listeners(BootState) ->
+    lists:foreach(
+      fun
+          ({_, Child, _, _}) when is_pid(Child) ->
+              gen_server:cast(Child, {notify_boot_state, BootState});
+          (_) ->
+              ok
+      end,
+      supervisor:which_children(?MODULE)).
--- a/apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
+++ b/apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
@@ -1,0 +1,183 @@
+%%%-------------------------------------------------------------------
+%% The contents of this file are subject to the Mozilla Public License
+%% Version 1.1 (the "License"); you may not use this file except in
+%% compliance with the License. You may obtain a copy of the License
+%% at https://www.mozilla.org/MPL/
+%%
+%% Software distributed under the License is distributed on an "AS IS"
+%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
+%% the License for the specific language governing rights and
+%% limitations under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% Copyright (c) 2015-2020 Pivotal Software, Inc.  All rights reserved.
+%%
+
+-module(rabbit_boot_state_systemd).
+
+-behaviour(gen_server).
+
+-export([start_link/0]).
+
+-export([init/1,
+         handle_call/3,
+         handle_cast/2,
+         terminate/2,
+         code_change/3]).
+
+-record(state, {mechanism,
+                sd_notify_module,
+                socket}).
+
+-define(LOG_PREFIX, "Boot state/systemd: ").
+
+start_link() ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+init([]) ->
+    case os:type() of
+        {unix, _} ->
+            case code:load_file(sd_notify) of
+                {module, sd_notify} ->
+                    {ok, #state{mechanism = legacy,
+                                sd_notify_module = sd_notify}};
+                {error, _} ->
+                    case os:getenv("NOTIFY_SOCKET") of
+                        false ->
+                            ignore;
+                        "" ->
+                            ignore;
+                        Socket ->
+                            {ok, #state{mechanism = socat,
+                                        socket = Socket}}
+                    end
+            end;
+        _ ->
+            ignore
+    end.
+
+handle_call(_Request, _From, State) ->
+    {noreply, State}.
+
+handle_cast({notify_boot_state, BootState}, State) ->
+    notify_boot_state(BootState, State),
+    {noreply, State}.
+
+terminate(normal, _State) ->
+    ok.
+
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+
+%%% Private
+
+notify_boot_state(ready = BootState,
+                  #state{mechanism = legacy, sd_notify_module = SDNotify}) ->
+    rabbit_log_prelaunch:debug(
+      ?LOG_PREFIX "notifying of state `~s` (via native module)",
+      [BootState]),
+    sd_notify_legacy(SDNotify);
+notify_boot_state(ready = BootState,
+                  #state{mechanism = socat, socket = Socket}) ->
+    rabbit_log_prelaunch:debug(
+      ?LOG_PREFIX "notifying of state `~s` (via socat(1))",
+      [BootState]),
+    sd_notify_socat(Socket);
+notify_boot_state(BootState, _) ->
+    rabbit_log_prelaunch:debug(
+      ?LOG_PREFIX "ignoring state `~s`",
+      [BootState]),
+    ok.
+
+sd_notify_message() ->
+    "READY=1\nSTATUS=Initialized\nMAINPID=" ++ os:getpid() ++ "\n".
+
+sd_notify_legacy(SDNotify) ->
+    SDNotify:sd_notify(0, sd_notify_message()).
+
+%% socat(1) is the most portable way the sd_notify could be
+%% implemented in erlang, without introducing some NIF. Currently the
+%% following issues prevent us from implementing it in a more
+%% reasonable way:
+%% - systemd-notify(1) is unstable for non-root users
+%% - erlang doesn't support unix domain sockets.
+%%
+%% Some details on how we ended with such a solution:
+%%   https://github.com/rabbitmq/rabbitmq-server/issues/664
+sd_notify_socat(Socket) ->
+    case sd_current_unit() of
+        {ok, Unit} ->
+            rabbit_log_prelaunch:debug(
+              ?LOG_PREFIX "systemd unit for activation check: \"~s\"~n",
+              [Unit]),
+            sd_notify_socat(Socket, Unit);
+        _ ->
+            ok
+    end.
+
+sd_notify_socat(Socket, Unit) ->
+    try sd_open_port(Socket) of
+        Port ->
+            Port ! {self(), {command, sd_notify_message()}},
+            Result = sd_wait_activation(Port, Unit),
+            port_close(Port),
+            Result
+    catch
+        Class:Reason ->
+            rabbit_log_prelaunch:debug(
+              ?LOG_PREFIX "Failed to start socat(1): ~p:~p~n",
+              [Class, Reason]),
+            false
+    end.
+
+sd_current_unit() ->
+    CmdOut = os:cmd("ps -o unit= -p " ++ os:getpid()),
+    Ret = (catch re:run(CmdOut,
+                        "([-.@0-9a-zA-Z]+)",
+                        [unicode, {capture, all_but_first, list}])),
+    case Ret of
+        {'EXIT', _}     -> error;
+        {match, [Unit]} -> {ok, Unit};
+        _               -> error
+    end.
+
+socat_socket_arg("@" ++ AbstractUnixSocket) ->
+    "abstract-sendto:" ++ AbstractUnixSocket;
+socat_socket_arg(UnixSocket) ->
+    "unix-sendto:" ++ UnixSocket.
+
+sd_open_port(Socket) ->
+    open_port(
+        {spawn_executable, os:find_executable("socat")},
+        [{args, [socat_socket_arg(Socket), "STDIO"]},
+            use_stdio, out]).
+
+sd_wait_activation(Port, Unit) ->
+    case os:find_executable("systemctl") of
+        false ->
+            rabbit_log_prelaunch:debug(
+              ?LOG_PREFIX "systemctl(1) unavailable, falling back to sleep~n"),
+            timer:sleep(5000),
+            ok;
+        _ ->
+            sd_wait_activation(Port, Unit, 10)
+    end.
+
+sd_wait_activation(_, _, 0) ->
+    rabbit_log_prelaunch:debug(
+      ?LOG_PREFIX "service still in 'activating' state, bailing out~n"),
+    ok;
+sd_wait_activation(Port, Unit, AttemptsLeft) ->
+    Ret = os:cmd("systemctl show --property=ActiveState -- '" ++ Unit ++ "'"),
+    case Ret of
+        "ActiveState=activating\n" ->
+            timer:sleep(1000),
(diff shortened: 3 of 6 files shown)


apps/rabbitmq_prelaunch/src/rabbit_boot_state_sup.erl
```diff
@@ -0,0 +1,38 @@
...
+-export([start_link/0,
+         init/1]).
+
+-export([child_pids/0]).
+
+start_link() ->
+    supervisor:start_link({local, ?MODULE}, ?MODULE, []).
+
+init([]) ->
+    Procs = [#{id => rabbit_boot_state_systemd,
```
michaelklishin: Since this is a standard supervisor I'd use maps for everything, including the restart strategy.

pjk25: Sounds good.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
```diff
@@ -0,0 +1,97 @@
...
+%%
+%% Software distributed under the License is distributed on an "AS IS"
+%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
+%% the License for the specific language governing rights and
+%% limitations under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% The Initial Developer of the Original Code is GoPivotal, Inc.
+%% Copyright (c) 2007-2020 Pivotal Software, Inc.  All rights reserved.
```
dumbbell: You can change 2007 into 2019 because that code was written last year.

 I believe you can also remove the "Initial Developer" line above the copyright because we were already Pivotal Software at that point.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
```diff
@@ -0,0 +1,97 @@
...
+%%
+
+-module(rabbit_boot_state).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-export([get_boot_state/0,
+         set_boot_state/1,
+         wait_for_boot_state/1]).
```
dumbbell: What do you think of renaming those functions to `get/0`, `set/1` and `wait_for/1`? The module name indicates the context of what is to be set/get/waited.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
```diff
@@ -0,0 +1,97 @@
...
+-define(PT_KEY_BOOT_STATE,    {?MODULE, boot_state}).
+
+-type boot_state() :: 'stopped' | 'booting' | 'ready' | 'stopping'.
+
+-export_type([boot_state/0]).
+
+-spec get_boot_state() -> boot_state().
+get_boot_state() ->
+  persistent_term:get(?PT_KEY_BOOT_STATE, stopped).
+
```
dumbbell: About the style, we use 4-space indentation usually.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
```diff
@@ -0,0 +1,97 @@
...
+-spec set_boot_state(boot_state()) -> term().
```
dumbbell: Thank you for adding specifications!

In this case, it should return `ok`, not `term()`.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
```diff
@@ -0,0 +1,97 @@
...
+set_boot_state(BootState) ->
+  rabbit_log_prelaunch:debug("Change boot state to `~s`", [BootState]),
+  ?assert(is_boot_state_valid(BootState)),
+  case BootState of
+    stopped -> persistent_term:erase(?PT_KEY_BOOT_STATE);
+    _ -> persistent_term:put(?PT_KEY_BOOT_STATE, BootState)
```
dumbbell: Style about the style, I don't know what other member of the team do, but I tend to align arrows when the statements following then is on the same line. This isolates a bit better the term being matched (before the arrow).

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
```diff
@@ -0,0 +1,97 @@
...
+  end,
+  notify_boot_state_listeners(BootState).
+
+-spec wait_for_boot_state(boot_state()) -> term().
```
dumbbell: Here, it should return `ok | {error, timeout}`.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
```diff
@@ -0,0 +1,97 @@
...
+is_boot_state_reached(stopping, stopped) ->
+  false;
+is_boot_state_reached(_CurrentBootState, stopped) ->
+  true;
+is_boot_state_reached(stopped, _TargetBootState) ->
+  true;
+is_boot_state_reached(CurrentBootState, TargetBootState) ->
+  boot_state_idx(TargetBootState) =< boot_state_idx(CurrentBootState).
+
+notify_boot_state_listeners(BootState) ->
```
dumbbell: This entire function can be moved to `rabbit_boot_state_sup` I believe. It is responsible for its children.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_sup.erl
```diff
@@ -0,0 +1,41 @@
...
+%%
+%% Software distributed under the License is distributed on an "AS IS"
+%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
+%% the License for the specific language governing rights and
+%% limitations under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% The Initial Developer of the Original Code is GoPivotal, Inc.
+%% Copyright (c) 2007-2020 Pivotal Software, Inc.  All rights reserved.
```
dumbbell: Copyright starting this year for this code.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
```diff
@@ -0,0 +1,160 @@
...
+%%
+%% Software distributed under the License is distributed on an "AS IS"
+%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
+%% the License for the specific language governing rights and
+%% limitations under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% The Initial Developer of the Original Code is GoPivotal, Inc.
+%% Copyright (c) 2007-2020 Pivotal Software, Inc.  All rights reserved.
```
dumbbell: The systemd notification was introduced in 2015, already under Pivotal Software as well.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
```diff
@@ -0,0 +1,160 @@
...
+
+start_link() ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+init([]) ->
+    case code:load_file(sd_notify) of
+        {module, sd_notify} ->
+            {ok, #state{sd_notify_module = sd_notify}};
+        {error, _} ->
+            case rabbit_prelaunch:get_context() of
```
dumbbell: The `rabbit_boot_state_sup` tree is started before `rabbit_prelaunch` runs, so there is a chance this process also runs before the context is initialized. Now that we have that, we should perhaps revert part of one of my previous commit: this module should check `$NOTIFY_SOCKET` itself to make it independent (and RabbitMQ itself doesn't care about this variable after the systemd support is properly isolated). The revert can be handled later.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
```diff
@@ -0,0 +1,160 @@
...
+handle_cast({notify_boot_state, ready}, #{sd_notify_module := SDNotify} = State) ->
+    rabbit_log_prelaunch:debug("notifying systemd of readiness via native module"),
+    sd_notify_legacy(SDNotify),
+    {noreply, State};
+handle_cast({notify_boot_state, ready}, #{socket := Socket} = State) ->
+    rabbit_log_prelaunch:debug("notifying systemd of readiness via socat"),
+    sd_notify_socat(Socket),
+    {noreply, State};
+handle_cast({notify_boot_state, _BootState}, State) ->
+    {noreply, State}.
```
dumbbell: The handling of `{notify_boot_state, _}` should probably be moved to its own function (which would take the `State` as argument). It's often easier to read code outside of the `handle_cast()`/`handle_call()` callbacks.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
```diff
@@ -0,0 +1,160 @@
...
+                    {ok, #state{socket = Socket}};
+                _ ->
+                    ignore
+            end
+    end.
+
+handle_call(_Request, _From, State) ->
+    {noreply, State}.
+
+handle_cast({notify_boot_state, ready}, #{sd_notify_module := SDNotify} = State) ->
```
dumbbell: You mix the syntax of maps and records here.

To match the `#state` record, you have to do:
```
handle_cast({notify_boot_state, ready}, #state{sd_notify_module = SDNotify} = State) ->
```

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
```diff
@@ -0,0 +1,160 @@
...
+    rabbit_log_prelaunch:debug("notifying systemd of readiness via native module"),
+    sd_notify_legacy(SDNotify),
+    {noreply, State};
+handle_cast({notify_boot_state, ready}, #{socket := Socket} = State) ->
```
dumbbell: Same comment as above w.r.t. state matching.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
```diff
@@ -0,0 +1,160 @@
...
+    rabbit_log_prelaunch:debug("notifying systemd of readiness via socat"),
+    sd_notify_socat(Socket),
+    {noreply, State};
+handle_cast({notify_boot_state, _BootState}, State) ->
+    {noreply, State}.
+
+handle_info(Msg, State) ->
+    io:format(standard_error, "Unexpected message: ~p~n",[Msg]),
```
dumbbell: I suppose this is a debug message. We almost never write directly to stdout/stderr. You can log a debug message for this one if you want, but you need to put more context because the message won't tell which module/process issued it.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
```diff
@@ -0,0 +1,160 @@
...
+    {noreply, State}.
+
+terminate(normal, _State) ->
+    io:format(standard_error, "~p terminating.~n", [?MODULE]),
```
dumbbell: This one is not needed, except if you want to log more details about the termination. Erlang can take care of logging this kind of event if we enable it.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
```diff
@@ -0,0 +1,160 @@
...
+    ok.
+
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+
+%%% Private
+
+sd_notify_message() ->
+  "READY=1\nSTATUS=Initialized\nMAINPID=" ++ os:getpid() ++ "\n".
```
dumbbell: As earlier, please use 4-space indentation.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
pjk25: seems reasonable to me since the exports of that module are small.

pjk25: It does mean that there is one place in this module where we call the function where it will become `?MODULE:get()` to avoid ambiguity with the bif get, but I think that is acceptable.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
```diff
@@ -0,0 +1,97 @@
...
+  wait_for_boot_state(BootState, infinity).
+
+wait_for_boot_state(BootState, Timeout) ->
+  ?assert(is_boot_state_valid(BootState)),
+  wait_for_boot_state1(BootState, Timeout).
+
+wait_for_boot_state1(BootState, infinity = Timeout) ->
+  case is_boot_state_reached(BootState) of
+    true  -> ok;
+    false -> wait_for_boot_state1(BootState, Timeout)
```
pjk25: @dumbbell is this intended to wait in a tight loop? Also, since the only public wait_for uses a wait of infinity, it seems we will always follow this path. Should we not just expose the function with timeout, and invoke with the BOOT_FINISH_TIMEOUT when this is called in rabbit.erl?

dumbbell: Oops, you're right, it lacks a `timer:sleep(...)` like the clause with an integer timeout.

I agree for the use of `?BOOT_FINISH_TIMEOUT`.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: This crossed my mind, but in practice the context seemed to have been initialized. Another option that I considered was to start cleanly in init, and gen_server:call another init message to the listeners in prelaunch. This could occur after prelaunch logging is set up, and the boot state listeners could stop cleanly then.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: the original systemd code logs in this fashion, which I agree isn't ideal. I believe this would only occur in the case that the boot state listener contract is not fulfilled, and this is more a developer than a runtime error. Is it common practice to raise an exception in such a situation?

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: done.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: I didn't realize that. Removed.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
dumbbell: I agree.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: thanks. I guess erlang syntax can be a little complicated

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
dumbbell: I think we should just read `$NOTIFY_SOCKET` value and avoid any race conditions. This is not a RabbitMQ configurable variable anyway. I should not have added it to `rabbit_env`.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: I guess the other messages can use the prelaunch logger, as they won't occur until that logger is ready.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: I might have waited to do so until there was more than one message type to handle, but I have made an attempt nonetheless. Let me know what you think.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: you know your history. fixed.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
pjk25: done

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
pjk25: agreed.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
pjk25: it has been tightened up to `ok`

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
pjk25: done.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
dumbbell: Yes. You should use pattern matching if possible and let the code crash if there is an invalid input. This module is internal (not exposed to user input), so it makes sense to crash when the input is invalid.

Here, you have to implement `handle_info()` so you can't really use pattern matching here because we expect no message here.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
```diff
@@ -0,0 +1,87 @@
...
+wait_for_boot_state(_, _) ->
+    {error, timeout}.
+
+boot_state_idx(stopped)  -> 0;
+boot_state_idx(booting)  -> 1;
+boot_state_idx(ready)    -> 2;
+boot_state_idx(stopping) -> 3;
+boot_state_idx(_)        -> undefined.
+
+is_boot_state_valid(BootState) ->
```
dumbbell: We can probably rename `is_boot_state_valid()` and `is_boot_state_reached()` to `is_valid()` and `is_reached()` respectively. What do you think?

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
```diff
@@ -0,0 +1,87 @@
...
+        _       -> persistent_term:put(?PT_KEY_BOOT_STATE, BootState)
+    end,
+    rabbit_boot_state_sup:notify_boot_state_listeners(BootState).
+
+-spec wait_for(boot_state()) -> ok | {error, timeout}.
+wait_for(BootState) ->
+    ?assert(is_boot_state_valid(BootState)),
+    wait_for_boot_state(BootState, infinity).
+
+wait_for_boot_state(BootState, infinity = Timeout) ->
```
dumbbell: This function being the same as `wait_for/1`, you should also rename it to `wait_for/2`.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
```diff
@@ -0,0 +1,161 @@
...
+
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+
+%%% Private
+
+handle_notify_boot_state(ready, #{sd_notify_module = SDNotify} = State) ->
+    rabbit_log_prelaunch:debug("notifying systemd of readiness via native module"),
+    sd_notify_legacy(SDNotify),
+    {noreply, State};
```
dumbbell: The function should be `notify_boot_state/2` (preferrably (to me) with the state as first argument) and it should return `ok` (and perhaps an error in the future if we want error handling).

The `handle_cast/2` should take care of the `{noreply, ...}` return value.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
dumbbell: Ok, I added a comment

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: I think under the circumstances it's best to just leave the improved log message. We could `{stop, unexpected_message, State}` but if I'm not mistaken, that would take the whole broker down, and since these listeners feel of lower precedence, I would avoid that.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
pjk25: yes

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: I favor the state as the second arg, as otherwise the invocations are for me harder to read, but otherwise, yes.

apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
pjk25: wait_for/1 went away in favor of only wait_for/2.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
dumbbell: Do you agree to read `$NOTIFY_SOCKET` directly in this function, instead of relying on the context?

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
dumbbell: Ok

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
dumbbell: For now, you can change the message to something logged (`rabbit_log:debug(...)` or `rabbit_log:warning(...)`, as you wish).

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
```diff
@@ -0,0 +1,157 @@
...
+%% limitations under the License.
+%%
+%% The Original Code is RabbitMQ.
+%%
+%% Copyright (c) 2015-2020 Pivotal Software, Inc.  All rights reserved.
+%%
+
+-module(rabbit_boot_state_systemd).
+
+-behavior(gen_server).
```
dumbbell: We tend to use the British spelling, `behaviour`.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: Yes, I agree. I'll make the change.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: Works for me. Will change it.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: since handle_info is optional, and it wasn't clear to me if this should be a `rabbit_log_prelaunch:debug` or `rabbit_log:debug`, I decided to just remove the function.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: I yield to consistency.

pjk25:
It turns out calling `rabbit_prelaunch:get_context()` during `rabbit_boot_state_systemd:init/1` was too early during my testing in Ubuntu, so I have gone ahead and moved to reading the env var directly. 

I tested by installing 3.9.0-alpha.212 on Ubuntu, and swapping in the `*.ez` files from my dev build. Prior to the aforementioned fix, `sudo service rabbitmq-server restart` would hang, but now it exits cleanly, and I see
```
2020-02-07 10:17:31.290 [debug] <0.511.0> Marking RabbitMQ as running
2020-02-07 10:17:31.290 [debug] <0.511.0> Change boot state to `ready`
2020-02-07 10:17:31.290 [debug] <0.154.0> notifying systemd of readiness via socat
```
in the logs, so I believe things are working correctly.

apps/rabbitmq_prelaunch/src/rabbit_boot_state_systemd.erl
pjk25: Turns out the context was not yet properly initialized, so the change has been made.

dumbbell:
@pjk25: I will test your patch as well. Do you mind if we squash all your commits into a single one?

pjk25:
I don't typically both with the squash, but I don't mind doing so. There are a lot of commits here. Shall I do so and force push?

dumbbell:
Yes, you can do that. The idea is that the work in progress is not really relevant. You can explain in the commit message the various things which were tried and why some of them were abandonned.

While at it, you can rebase on top of the latest `master`.

dumbbell:
I did a couple style fixes (mostly to stay into the 80 columns), added a prefix to log messages ("Boot state/systemd:") and disabled the systemd backend on Microsoft Windows.

dumbbell:
It should be merged with all other pull requests involved in #2180.

PR #2247, opened 2020-02-17
Add ability to customize product name, version & banner

dumbbell:
To override the product name (defaulting to "RabbitMQ"):
* set the `$RABBITMQ_PRODUCT_NAME` environment variable, or
* set the `rabbit` application `product_name` variable.

To override the product version:
* set the `$RABBITMQ_PRODUCT_VERSION` environment variable, or
* set the `rabbit` application `product_version` variable.

To add content to the banner (both the copy logged and the one printed to stdout), indicate the filename which contains it, à la `/etc/motd` using:
* the `$RABBITMQ_MOTD_FILE` environment variable, or
* the `rabbit` application `motd_file` variable.

The default motd file is `/etc/rabbitmq/motd` on Unix and `%APPDATA%\RabbitMQ\motd.txt` on Windows.

Here is an example of the printed banner with name, version & motd configured:

```
  ##  ##      WeatherMQ 1.2.3
  ##  ##
  ##########  Copyright (c) 2007-2020 Pivotal Software, Inc.
  ######  ##
  ##########  Licensed under the MPL 1.1. Website: https://rabbitmq.com

  This is an example of a RabbitMQ message of the day.

  The message is written in Paris, France.    \  /
  It is partly cloudy outside, with a       _ /"".-.
  temperature of 12°C. Wind is around         \_(   ).
  30-40 km/h, from south-west.                /(___(__)

  Doc guides: https://rabbitmq.com/documentation.html
  Support:    https://rabbitmq.com/contact.html
  Tutorials:  https://rabbitmq.com/getstarted.html
  Monitoring: https://rabbitmq.com/monitoring.html

  Logs: /tmp/rabbitmq-test-instances/rabbit/log/rabbit@cassini.log
        /tmp/rabbitmq-test-instances/rabbit/log/rabbit@cassini_upgrade.log

  Config file(s): /tmp/rabbitmq-test-instances/test.config

  Starting broker... completed with 0 plugins.
```

New APIS are available to query those product informations and use them in e.g. plugins such as the management API/UI:
* `rabbit:product_info/0`
* `rabbit:product_name/0`
* `rabbit:product_version/0`
* `rabbit:motd_file/0`
* `rabbit:motd/0`

Depends on rabbitmq/rabbitmq-common#356.

### DIFF
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -30,6 +30,11 @@
 
 -export([start/2, stop/1, prep_stop/1]).
 -export([start_apps/1, start_apps/2, stop_apps/1]).
+-export([product_info/0,
+         product_name/0,
+         product_version/0,
+         motd_file/0,
+         motd/0]).
 -export([log_locations/0, config_files/0]). %% for testing and mgmt-agent
 -export([is_booted/1, is_booted/0, is_booting/1, is_booting/0]).
 
@@ -435,10 +440,12 @@ stop() ->
         ok ->
             case rabbit_boot_state:get() of
                 ready ->
-                    rabbit_log:info("RabbitMQ is asked to stop..."),
+                    Product = product_name(),
+                    rabbit_log:info("~s is asked to stop...", [Product]),
                     do_stop(),
                     rabbit_log:info(
-                      "Successfully stopped RabbitMQ and its dependencies"),
+                      "Successfully stopped ~s and its dependencies",
+                      [Product]),
                     ok;
                 stopped ->
                     ok
@@ -462,7 +469,9 @@ stop_and_halt() ->
     try
         stop()
     catch Type:Reason ->
-        rabbit_log:error("Error trying to stop RabbitMQ: ~p:~p", [Type, Reason]),
+        rabbit_log:error(
+          "Error trying to stop ~s: ~p:~p",
+          [product_name(), Type, Reason]),
         error({Type, Reason})
     after
         %% Enclose all the logging in the try block.
@@ -518,9 +527,9 @@ stop_apps([]) ->
     ok;
 stop_apps(Apps) ->
     rabbit_log:info(
-        lists:flatten(["Stopping RabbitMQ applications and their dependencies in the following order:~n",
+        lists:flatten(["Stopping ~s applications and their dependencies in the following order:~n",
                        ["    ~p~n" || _ <- Apps]]),
-        lists:reverse(Apps)),
+        [product_name() | lists:reverse(Apps)]),
     ok = app_utils:stop_applications(
            Apps, handle_app_error(error_during_shutdown)),
     case lists:member(rabbit, Apps) of
@@ -663,7 +672,7 @@ maybe_print_boot_progress(true, IterationsLeft) ->
                {memory, any()}].
 
 status() ->
-    {ok, Version} = application:get_key(rabbit, vsn),
+    Version = product_version(),
     S1 = [{pid,                  list_to_integer(os:getpid())},
           %% The timeout value used is twice that of gen_server:call/2.
           {running_applications, rabbit_misc:which_applications()},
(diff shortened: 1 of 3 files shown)


src/rabbit.erl
```diff
@@ -1236,6 +1271,112 @@ validate_msg_store_io_batch_size_and_credit_disc_bound(CreditDiscBound,
...
+    #{name := ProductName} = product_infos(),
+    ProductName.
+
+-spec product_version() -> string().
+
+product_version() ->
+    #{version := ProductVersion} = product_infos(),
+    ProductVersion.
+
+-spec product_infos() -> #{name := string(),
```
gerhard: Can we rename this to `product_info`?

💡 [Information is a mass noun and doesn't have a plural](https://www.grammar-monster.com/plurals/plural_of_information.htm) 💡

src/rabbit.erl
```diff
@@ -1236,6 +1271,112 @@ validate_msg_store_io_batch_size_and_credit_disc_bound(CreditDiscBound,
...
+
+product_infos() ->
+    PTKey = {?MODULE, product},
+    try
+        %% The value is cached the first time to avoid calling the
+        %% application master many times just for that.
+        persistent_term:get(PTKey)
+    catch
+        error:badarg ->
+            %% FIXME: Move those variables to rabbit_env.
```
gerhard: I cannot think of a good reason to allow this to be configured via an environment variable. I am leaning towards keeping this as application config only.

What was your thinking behind this?

src/rabbit.erl
```diff
@@ -1094,9 +1114,24 @@ print_banner() ->
               "~n  Config file(s): ~ts" ++ CfgFmt ++ "~n"
               "~n  Starting broker...",
               [Product, Version, ?COPYRIGHT_MESSAGE, ?INFORMATION_MESSAGE] ++
+              MOTDArgs ++
               LogLocations ++
               CfgLocations).
 
+log_motd() ->
```
gerhard: MOTD was a brilliant idea - love it!

src/rabbit.erl
```diff
@@ -1236,6 +1271,112 @@ validate_msg_store_io_batch_size_and_credit_disc_bound(CreditDiscBound,
...
+
+-spec product_infos() -> #{name := string(),
+                           version := string()}.
+
+product_infos() ->
+    PTKey = {?MODULE, product},
+    try
+        %% The value is cached the first time to avoid calling the
+        %% application master many times just for that.
+        persistent_term:get(PTKey)
```
gerhard: Nice optimisation, definitely an Enterprise Feature™️ 👔

src/rabbit.erl
michaelklishin: `info` and `infos` (info items) is a term commonly used in the codebase, although in a different context (metric emission)

gerhard: I would have never guessed that `infos` stands for `info_items`.

`product_info_items` makes sense, even though I find `product_info` more succinct and easier to understand. Either of them work, while `product_infos` not so much.

src/rabbit.erl
dumbbell: I generally like the ability to set things through the environment. This made manual testing way easier for instance.

dumbbell: I'm not saying these variables should be documented and people should rely on them though.

src/rabbit.erl
dumbbell: > bulb [Information is a mass noun and doesn't have a plural](https://www.grammar-monster.com/plurals/plural_of_information.htm) bulb

Thank you, I always forget about this rule...

src/rabbit.erl
gerhard: I see. If we agree to not document this and only use it for testing, happy to keep as is.

gerhard:
Ready to merge from my POV, not sure if you wanted to do anything else before changing it from WIP to RTM (Ready to Merge).

dumbbell:
I just want to move the environment variables handling in `rabbit_env` as said in the TODO, then I will mark it as ready.

PR #2250, opened 2020-02-20
Add ability to customize product name, version & banner [v3.8.x]

dumbbell:
This the equivalent of #2247 for `v3.8.x`.

### DIFF
--- a/docs/rabbitmq.conf.example
+++ b/docs/rabbitmq.conf.example
@@ -520,6 +520,17 @@
 ##
 # proxy_protocol = false
 
+## Overriden product name and version.
+## They are set to "RabbitMQ" and the release version by default.
+# product.name = RabbitMQ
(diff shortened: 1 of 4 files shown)


gerhard:
Deferring this to 3.8.4 so that investigating & fixing [this CI failure](https://ci.rabbitmq.com/teams/main/pipelines/server-release:v3.9.x/jobs/test-package-windows-on-windows-server-2012/builds/198) doesn't block the release of v3.8.3.

dumbbell:
Using `os:set_signal/2` on Windows was indeed the cause of the failure in CI. Thsi has been fixed in `master` and this patch was updated accordingly.

dumbbell:
Ci is happy again!

dumbbell:
I will update the patch with the recently written testsuite.

PR #2260, opened 2020-02-26
Optimise messages_ready function by keeping counts of prefix messages

kjnilsson:
rather than taking length/1. There could be a lot of prefix messages
during during recovery which could make recovery unbearably slow.

### DIFF
--- a/src/rabbit_fifo.erl
+++ b/src/rabbit_fifo.erl
@@ -327,7 +327,7 @@ apply(#{index := RaftIdx}, #purge{},
                                                   messages = #{},
                                                   returns = lqueue:new(),
                                                   msg_bytes_enqueue = 0,
-                                                  prefix_msgs = {[], []},
+                                                  prefix_msgs = {0, [], 0, []},
                                                   low_msg_num = undefined,
                                                   msg_bytes_in_memory = 0,
                                                   msgs_ready_in_memory = 0},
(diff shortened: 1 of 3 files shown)


gerhard:
Waiting for a few million messages before I trigger a restart. Current state:

image
image

gerhard:
Given:

* 50 quorum queues
* with `x-max-length` of 100k
* and `x-max-in-memory-bytes` of 100MB
* and 100k messages of size 26KB each (2.4GiB in total per queue)

When rabbit app got restarted via `rabbitmqctl stop_app ; rabbitmqctl start_app`:

* rabbit app restarted within 2 minutes
* all quorum queues recovered 8 minutes afterwards
* each quorum queue process was sustaining 600k reductions/second
* disk reads were a sustained 280MiB/s
* CPU user was 17%
* CPU system was 6%

Screenshot 2020-02-26 at 19 50 45

Screenshot 2020-02-26 at 19 51 02

Screenshot 2020-02-26 at 19 51 21

Screenshot 2020-02-26 at 19 51 33

Screenshot 2020-02-26 at 19 38 25

This is such a great improvement that it deserves an instant back-port to `v3.8.x` & :shipit:. As I won't be here for this pleasure, can @dumbbell or @michaelklishin take this over?

michaelklishin:
Backported to `v3.8.x`.

PR #2262, opened 2020-02-28
Fix QQ dead letter crash

kjnilsson:
that would occur when dead-lettering a message that was not kept in
memory. At that point the dead lettering function failed as the message
had not be restored from the Ra log.


acogoluegnes:
Backported to `v3.8.x`.

PR #2270, opened 2020-03-06
Run checks & tests in GitHub Actions on every push

gerhard:
This is related to https://github.com/rabbitmq/rabbitmq-common/pull/361

Leaving as WIP since it's not finished.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+.github/workflows/ @gerhard @dumbbell
--- a/.github/workflows/otp-21.3.yml
+++ b/.github/workflows/otp-21.3.yml
@@ -1,0 +1,3898 @@
+# https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow
+name: OTP v21.3
+on: push
+jobs:
+  checks:
+    name: checks
+    runs-on: ubuntu-18.04
+    steps:
+      - name: CHECKOUT REPOSITORY
+        uses: actions/checkout@v2
+      # https://github.com/marketplace/actions/setup-elixir
+      - name: CONFIGURE OTP & ELIXIR
+        uses: actions/setup-elixir@v1
+        with:
+          otp-version: 21.3
+          # https://github.com/elixir-lang/elixir/releases
+          elixir-version: 1.10.2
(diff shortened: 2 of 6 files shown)


gerhard:
These tests sometimes pass and sometimes fail, same as they did in Concourse, so this has nothing to do with GitHub Actions.

When a test fails, does it fail because it's a race condition in the test or in the code? Maybe the test fails for a genuine reason, but only sometimes, so how can I tell since I have no context on the feature or the test? Yes, I could spend time to investigate, but then a change that should take a few weeks at most snowballs into a few months, or even longer - none knows. Some test suites have been fixed and as a result un-deleted, so if this is the desired outcome, let's split the deleted test suites among all those that will be joining this undertaking and fix them.

Alternatively, and this is the option that doesn't impose on anyone's time, let's delete all tests that sometimes fail, and stay focused on what is expected from this PR.

What makes sense to you?

michaelklishin:
I can see how some very intricate multi-node tests can be considered not worth the effort (perhaps we should do more property-based or workload simulation testing elsewhere) but deleting dozens of tests because one of them hangs on CI https://github.com/rabbitmq/rabbitmq-server/commit/3d09f39263a57e5b941f78089c9065f4bffbc989 is complete overkill.

michaelklishin:
I am splitting those kitchen sink test suites dozens of small ones in master. This way it would be a lot easier to investigate failures.

gerhard:
**clustering_management_SUITE > unclustered_2_nodes > cluster_size_2 > erlang_config** continues to fail predictably on both OTP [v21.3](https://github.com/rabbitmq/rabbitmq-server/runs/549369764?check_suite_focus=true#step:6:81) & [v22.3](https://github.com/rabbitmq/rabbitmq-server/runs/549371440?check_suite_focus=true#step:6:81)

Just fixed **ct-cluster_formation_locking** & rebased onto `master`.

gerhard:
Requires the following to be merged as well:

- https://github.com/rabbitmq/rabbitmq-erlang-client/tree/github-actions
- https://github.com/rabbitmq/rabbitmq-common/tree/github-actions
- https://github.com/rabbitmq/rabbitmq-cli/tree/github-actions

gerhard:
No tests were deleted, all checks pass, merging :shipit:

PR #2272, opened 2020-03-09
rabbit_fifo: force gc when queue is empty

kjnilsson:
And it uses more than a fixed limit of ~2Mb of total memory. An empty
queue should only need around ~100-200Kb of memory so this should avoid
any unnecessary full sweeps.

[#171644231]

### DIFF
--- a/src/rabbit_fifo.erl
+++ b/src/rabbit_fifo.erl
@@ -646,26 +646,51 @@ get_checked_out(Cid, From, To, #?MODULE{consumers = Consumers}) ->
             []
     end.
 
+-record(aux_gc, {last_raft_idx = 0 :: ra:index()}).
+-record(aux, {name :: atom(),
+              utilisation :: term(),
+              gc = #aux_gc{} :: #aux_gc{}}).
+
 init_aux(Name) when is_atom(Name) ->
     %% TODO: catch specific exception throw if table already exists
     ok = ra_machine_ets:create_table(rabbit_fifo_usage,
                                      [named_table, set, public,
                                       {write_concurrency, true}]),
     Now = erlang:monotonic_time(micro_seconds),
-    {Name, {inactive, Now, 1, 1.0}}.
+    #aux{name = Name,
+         utilisation = {inactive, Now, 1, 1.0}}.
 
-handle_aux(_, cast, Cmd, {Name, Use0}, Log, _) ->
-    Use = case Cmd of
+handle_aux(_RaState, cast, Cmd, #aux{name = Name,
+                                     utilisation = Use0} = State0,
+           Log, MacState) ->
+    State = case Cmd of
               _ when Cmd == active orelse Cmd == inactive ->
-                  update_use(Use0, Cmd);
+                  State0#aux{utilisation = update_use(Use0, Cmd)};
               tick ->
                   true = ets:insert(rabbit_fifo_usage,
                                     {Name, utilisation(Use0)}),
-                  Use0;
+                  eval_gc(Log, MacState, State0);
               eval ->
-                  Use0
+                  State0
           end,
-    {no_reply, {Name, Use}, Log}.
+    {no_reply, State, Log}.
+
+eval_gc(Log, #?MODULE{cfg = #cfg{resource = QR}} = MacState,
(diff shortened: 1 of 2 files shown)


src/rabbit_fifo.erl
```diff
@@ -646,26 +646,51 @@ get_checked_out(Cid, From, To, #?MODULE{consumers = Consumers}) ->
...
+eval_gc(Log, #?MODULE{cfg = #cfg{resource = QR}} = MacState,
+        #aux{gc = #aux_gc{last_raft_idx = LastGcIdx} = Gc} = AuxState) ->
+    {Idx, _} = ra_log:last_index_term(Log),
+    {memory, Mem} = erlang:process_info(self(), memory),
+    case messages_total(MacState) of
+        0 when Idx > LastGcIdx andalso
+               Mem > ?GC_MEM_LIMIT_B ->
+            garbage_collect(),
+            {memory, MemAfter} = erlang:process_info(self(), memory),
+            rabbit_log:info("~s: forcing full sweep GC "
```
michaelklishin: Do we want to log this as `info`? What if there are thousands of queues?

lukebakken: Seems like a `debug` message to me...

michaelklishin: This question still stands

gerhard: Will fix when we know that everything else works.

gerhard:
@Jrmcneil can you re-run & share your metrics from GKE with [pivotalrabbitmq/rabbitmq:3.9.0-2020.03.12-pr2272](https://hub.docker.com/layers/pivotalrabbitmq/rabbitmq/3.9.0-2020.03.12-pr2272/images/sha256-e1b244aa692b7160c08dae89423cddb6aab33add0f2f03b3f22071c51993a027?context=explore) please? Does this fix the issue that you have observed?

This is what I am observing locally:

image

image

```
rmq2-qq_1           | 2020-03-13 23:54:06.779 [debug] <0.1093.0> queue 'qq1' in vhost '/': full GC sweep complete. Process memory reduced from 3560.74MB to 0.03MB.
rmq0-qq_1           | 2020-03-13 23:54:09.502 [debug] <0.1337.0> queue 'qq1' in vhost '/': full GC sweep complete. Process memory reduced from 2971.39MB to 0.03MB.
rmq1-qq_1           | 2020-03-13 23:54:10.488 [debug] <0.681.0> queue 'qq1' in vhost '/': full GC sweep complete. Process memory reduced from 1355.53MB to 0.05MB.
```

@kjnilsson we may want to do the same for ETS. This is the state of `rmq0-qq_1` after all messages get drained:

image

`rmq1-qq_1`, where the leader is running, has the largest ETS usage which does not get cleaned up after messages are drained:

image

kjnilsson:
The ETS usage are the log entries that are currently in the WAL. It will get reclaimed when the WAL reaches it's max size and the entries a flushed to disk segments so there is no need to do this manually as the max ETS memory usage has a maximum limit (ish). You can test that the ETS memory is reclaimed by forcing the WAL to close it's current file and start a new one by evaluating `ra_log_wal: force_roll_over(ra_log_wal).`

If you want a smaller WAL file (and thus less max ETS memory usage) you can configure `raft.wal_max_size_bytes` accordingly.

gerhard:
Thanks for reminding me re ETS memory usage. Could we add this info to the GC debug message?

That's the last change that I would like to see before I'm happy to merge this.

gerhard:
The conclusion to an offline discussions is that we shouldn't log the ETS memory usage in the context of the queue GC message since WAL (uses most of the ETS in the above cases) is shared across all ra servers (QQs):

> I think the fact that the WAL may use some fixed (ish) amount of memory should be shared through docs not necessarily be mentioned in somewhat unrelated log messages. We make some attempt here: https://www.rabbitmq.com/quorum-queues.html#resource-use

We will merge this as is and follow-up on these ideas at some point in the future
* `reclaim-qq-memory` command that forces GC of all qq members as well as force the WAL to flush
* review the 512MB default WAL size & warnings or node boot failures if available memory is less than this limit (1.5x - 2x maybe)

Issue #2254, opened 2020-02-24
Node can use significant (e.g. 80GB) amounts of RAM during startup


gerhard:
Part of [TGIR S01E02: Help! RabbitMQ ate my RAM!](https://github.com/rabbitmq/tgir/pull/7) I came across the following unexpected behaviour when rabbit app gets restart via `rabbitmqctl stop_app ; rabbitmqctl start_app`

<img width="982" alt="Screenshot 2020-02-24 at 13 57 31" src="https://user-images.githubusercontent.com/3342/75158272-ac972780-570d-11ea-9522-df330e9b28fe.png">

<img width="947" alt="Screenshot 2020-02-24 at 13 45 18" src="https://user-images.githubusercontent.com/3342/75158374-da7c6c00-570d-11ea-9f91-9dd14e60be0c.png">

A worker process is using 56GB of physical memory while reading through all queue segments, across all queues. This operation takes ~30 minutes to complete (13:44:16 - 14:15:46) with the disk reads starting at 410MiB/s and maxing out at 965MiB/s for 15 minutes straight:

![image](https://user-images.githubusercontent.com/3342/75166233-7a8cc200-571b-11ea-8841-5470308c26a6.png)

This is what the system physical memory breakdown looks like:

![image](https://user-images.githubusercontent.com/3342/75166388-b031ab00-571b-11ea-81ec-0b368d4d7bb9.png)

Notice:

1. Slow 1.8GiB -> 7.7GiB memory growth spanning ~13 minutes (13:28:00 - 13:41:00)
1. Large spike to 42GiB the next minute (13:42:00)
1. Another spike to 60GiB for which cached memory is made available (13:43:00)
1. Last spike to 80GiB (13:43:45) 🤯
1. A drop to 60GiB for 30 minutes during which the disk gets thrashed (13:44:15)
1. Drop to 8GiB (14:16:00) followed by a slow ramp-up during which publishers/consumers get reconnected
1. Regular 11.5GiB -> 17.5GiB spikes, as expected for this workload.

I have captured the runtime thread stats to show the 88% idle regular schedulers (most time is spent in `send`) and 94% idle dirty_io schedulers. All other thread types are 98% or above idle.

```
rabbitmq-diagnostics runtime_thread_stats --sample-interval 60 | grep -v async
Will collect runtime thread stats on rabbit@tgir-s01e02-gerhard-rabbitmq-durable for 60 seconds...
Average thread real-time    :  60152930 us
[...]

PR #2274, opened 2020-03-12
Reduce memory consumption during startup when there are a large amount of messages on disk

pjk25:
Should address #2254 when complete


pjk25:
After testing this particular refactor, no detectable effect on memory was had. Closing this one. Will introduce a new PR if I have success with other changes.

PR #2277, opened 2020-03-16
Always handle config files with Cuttlefish

dumbbell:
This has several benefits:

1.  It simplifies the code, all configuration being handled by the same code path (no more condition on Erlang-term-based vs. Cuttlefish). `rabbit_config` shrinks quite a lot in the process.

2.  We can use additional configuration files AND an Erlang-term-based configuration file. In other words, it is possible to use the same existing Erlang-term-based file and introduce Cuttlefish files when needed.

    It allows a user to run RabbitMQ with:

    ```
    RABBITMQ_CONFIG_FILE=/path/to/rabbitmq.config \
    RABBITMQ_CONFIG_FILES=/path/to/conf.d/*.conf \
    ./sbin/rabbitmq-server
    ```

    A developer can do the same with `make run-broker`:

    ```
    make run-broker \
    RABBITMQ_CONFIG_FILES=/path/to/conf.d/*.conf
    ```

    In the example above, the main configuration file generated by rabbitmq-run.mk is an Erlang-term-based one.

This is implemented by calling Cuttlefish with a (possibly empty) list of additional files and the Erlang-term-based file as the advanced configuration file.

References #2180.

### DIFF
--- a/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_conf.erl
+++ b/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_conf.erl
@@ -26,29 +26,13 @@ setup(Context) ->
     AdvancedConfigFile = find_actual_advanced_config_file(Context),
     State = case find_actual_main_config_file(Context) of
                 {MainConfigFile, erlang} ->
-                    case AdditionalConfigFiles of
-                        [] ->
-                            ok;
-                        _ ->
-                            rabbit_log_prelaunch:notice(
-                              "The following additional configuration "
-                              "files are not loaded when the main "
-                              "configuration file uses the Erlang terms "
-                              "based format"),
-                            lists:foreach(
-                              fun(File) ->
-                                      rabbit_log_prelaunch:notice(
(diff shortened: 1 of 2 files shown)


gerhard:
Works as expected, confirmed by running the following command locally:

```
cat > rabbitmq.conf <<EOF
load_definitions = /Users/gerhard/github.com/rabbitmq/3.9.x/deps/rabbit/rabbitmq-definitions.json
EOF

cat > rabbitmq-definitions.json <<EOF
{
  "global_parameters": [
    {"name": "cluster_name", "value": "pr-2277"}
  ]
}
EOF

RABBITMQ_CONFIG_FILES=/Users/gerhard/github.com/rabbitmq/3.9.x/deps/rabbit/rabbitmq.conf make run-broker
...
(rabbit@focker)1> rabbit_mnesia:status().
[{nodes,[{disc,[rabbit@focker]}]},
 {running_nodes,[rabbit@focker]},
 {cluster_name,<<"pr-2277">>},
 {partitions,[]}]
```

PR #2279, opened 2020-03-18
Reduce memory usage during startup

pjk25:
In the case of large backlogs of persistent messages (10s of millions
of messages)

Previously we queued a job for every file with worker_pool:submit_async,
however if there are 50 million messages, this corresponds to ~79,000 files
and the same number of pending tasks in the worker pool. The mailbox for
worker_pool explodes under these circumstances, using massive amounts of
memory.

The following was helpful in zeroing in on the problem:
https://elixirforum.com/t/extremely-high-memory-usage-in-genservers/4035

When testing with the setup given in #2254, the following applies:

RAM usage hovered around 7GB startup, so in that regard this change fixes the memory usage in dramatic fashion.

During my testing, there were several occasions where the broker ultimately failed to start as memory usage went over 100GB. gcloud appeared to be restarting the container for me.

Startup took approximately 1 hour and 15 minutes. This matched the reduced CPU usage I observed. I'll comment further inline with the code on this point.


src/rabbit_msg_store.erl
```diff
@@ -1810,15 +1789,61 @@ build_index_worker(Gatherer, State = #msstate { dir = Dir },
...
+                                  right            = Right,
+                                  file_size        = FileSize1,
+                                  locked           = false,
+                                  readers          = 0 }),
     ok = gatherer:finish(Gatherer).
 
+enqueue_build_index_workers(Gatherer, Left, [], State) ->
+    exit(normal);
+enqueue_build_index_workers(Gatherer, Left, [File|Files], State) ->
+    ok = worker_pool:submit(
```
pjk25: Here I believe the spawned process should queue work into the worker pool only as workers are free, avoiding the massive backlog which is the source of the massive memory usage. In my tests, however, CPU usage hovered around 100% (1 CPU) during the index building. This may be because the worker tasks are quick enough that this single process cannot feed the pool fast enough.

Without theses changes, CPU hovered around 200%, and startup was about 40% faster, so this is not ideal. Ideally we could bound the backlog of the worker pool, such that even a `submit_async` is blocking given that depth (which could default to `infinity` to preserve the existing behavior). Another option might be to look at refactoring the worker pool such that it doesn't degrade for large backlogs (if such a change is possible within acceptable tradeoffs).

src/rabbit_msg_store.erl
```diff
@@ -1810,15 +1789,61 @@ build_index_worker(Gatherer, State = #msstate { dir = Dir },
...
+                         State #msstate {
+                           sum_valid_data = SumValid + ValidTotalSize,
+                           sum_file_size  = SumFileSize + FileSize })
+    end.
+
+rebuild_index(Gatherer, Files, State) ->
+    lists:foreach(fun (File) ->
+                          ok = gatherer:fork(Gatherer)
+                  end, Files),
+    spawn(
```
pjk25: Initially this was a `spawn_link` call, but doing so altered the supervisor tree such that https://github.com/rabbitmq/rabbitmq-server/blob/abc0b522de83e36650e873d2065646dc1f4ae1d4/src/rabbit_vhost_msg_store.erl#L54 would fail because https://github.com/rabbitmq/rabbitmq-server/blob/abc0b522de83e36650e873d2065646dc1f4ae1d4/src/rabbit_vhost_msg_store.erl#L69 would return `[undefined]` rather than a valid pid.

michaelklishin:
The general idea makes sense: the whole point of using a worker pool is to *reduce* concurrency. Scanning and recovering with 79K processes would not be productive and neither is enqueueing 79K pending operations. The heavy lifting part would still be bounded by the number of work pool processes.

This is also proportional to the number of on-disk segment files, which is an acceptable trade-off.

pjk25:
I tested changing the `worker_pool` process to a `gen_server` rather than a `gen_server2`, to determine if the additional prioritization logic in `gen_server2` contributes to the memory usage. Results did not change.

pjk25:
After further investigation, I realized that the behavior of `worker_pool:submit` was not as I expected - it runs the task synchronously, in the process of an available worker node. Therefore, the `enqueue_build_index_workers` loop was only capable of loading one worker in the pool (matching the observation in my tests of ~1 utilized CPU).
https://github.com/rabbitmq/rabbitmq-common/pull/368 adds and additional function to `worker_pool`, `dispatch_sync` which blocks the caller only while all workers are busy. This is now applied with 01270fd3cdd381d14af7a1861e7f8b839a5d3130.
Testing as before with this setup brings CPU usage up to the 200-300% range, memory usage up to the 8-9GB range, and corrects the startup time penalty of bfdb74cbd867d6e882c4e6fa434793b94c8d229f.

src/rabbit_msg_store.erl
pjk25: > the spawned process should queue work into the worker pool only as workers are free

turned out not to be true for worker_pool:submit, but it corrected in a new commit.

michaelklishin:
@pjk25 excellent work, the extension of `worker_pool` sounds like the right thing to do to me.

gerhard:
@pjk25 do we have a Docker image with this change? I would like to re-run the workload that I used when creating #2254.

pjk25:
@gerhard  eu.gcr.io/cf-rabbitmq-core/rabbitmq:3.8.0-pjk-8 is the image that I used, however it would not have included the two new commits I made today (they are not performance related, but part of the PR nonetheless). It was created from rabbitmq-public-umbrella's `make docker-image` from master, after checking out the corresponding branches of `deps/rabbit` and `deps/rabbit_common`. I'm happy to produce a new image in the morning if that is helpful.

pjk25:
@gerhard I have produced a new image for you: `eu.gcr.io/cf-rabbitmq-core/rabbitmq:3.9.0-pjk-2279`

gerhard:
@michaelklishin I need the `startup_memory_fix` branch & the `worker_pool_dispatch_sync` from https://github.com/rabbitmq/rabbitmq-common/pull/368 for https://github.com/rabbitmq/tgir/pull/8. Let's keep them around for the duration of TGIR Season 1 which runs through to 25th December 2020. Thanks!

michaelklishin:
Well, you might want to protect them since it'd be easy to forget :) Otherwise I have no objections.

dumbbell:
Backported to 3.8.x as part of rabbitmq/rabbitmq-server#2343.

PR #2281, opened 2020-03-19
Fix feature flags registry loading + improve feature flags testing

dumbbell:


dumbbell:
There was a follow-up commit after this patch: commit 194bc4c8b665c47a152f6b186a0c3579717c5cdc.

PR #2292, opened 2020-03-31
Increase wait timeouts in rabbit_fifo_int

kjnilsson:
250ms is probably a bit on the low side.


michaelklishin:
The failure in `peer_discovery_classic_config_SUITE` is unrelated.

michaelklishin:
Backported to `v3.8.x`.

PR #2293, opened 2020-03-31
Move rabbit_channel config value to config record

kjnilsson:
writer_gc_threshold is a static value and shoudl be in the static config
record not in the main channel record that should only hold mutable data
fields.


michaelklishin:
Backported to `v3.8.x`.

PR #2294, opened 2020-03-31
clustering_management_SUITE: No need to stop node after start_app failure in `erlang_config`

dumbbell:
Since #2180, a failed `start_app` does not take the node down anymore. Restarting the node just after was failing since (because the node is still there), but this remained unnoticed so far because the return value of `start_node()` is not checked.

However, since rabbitmq/rabbitmq-ct-helpers@c033d9272afaf3575505533c81f1c0c7cfcb6206, the Make recipe which starts the node automatically stops it if the start failed somewhere. This is in order to not leave an unwanted node around.

This means that after the failing `rabbit_ct_broker_helpers:start_node()`, the node was effectively stopped this time, leading to the rest of the testcase to fail.


gerhard:
This looks good to me. Anything else that needs to happen before this is ready to merge?

PR #2295, opened 2020-04-01
rabbit_fifo: set timer when publish rejected

kjnilsson:
and no current leader is known so that we can re-try after a timeout

together with https://github.com/rabbitmq/ra/commit/f788622ce94700b9d4dd655ff0126715be727c72 this should help avoid a channel crash scenario when the leader is not known during publish.


dumbbell:
Backported to 3.8.x as part of rabbitmq/rabbitmq-server#2343.

PR #2300, opened 2020-04-03
Remove Ra segment_max_entries override

kjnilsson:
So that it uses Ra's internal default of 4096 instead which is safer for
larger message sizes.


michaelklishin:
Ironically this would be a net negative for environments where the only use of Raft is in the MQTT plugin. We'll address this there, with docs and maybe default changes.

PR #2302, opened 2020-04-06
feature_flags_SUITE: Work around CLI/rabbitmq-server circular dependency

dumbbell:
We need to copy `rabbit` in `my_plugins` plugins directory, not because `my_plugin` depends on it, but because the CLI erroneously depends on the broker.

This can't be fixed easily because this is a circular dependency (i.e. the broker depends on the CLI). So until a proper solution is implemented, keep this second copy of the broker for the CLI to find it.


dumbbell:
It depends on rabbitmq/rabbitmq-common#370 to fully fix the testsuite in the described scenario.

PR #2308, opened 2020-04-09
Remove flaky tests

dumbbell:
Those testcases fail quite often in various CIs. Most of the time, there is a race condition or timing issue behind the failure.

We can still restore them when someone has time to fix them.


michaelklishin:
Via git history and this/GitHub Actions PRs.

michaelklishin:
Backported to `v3.8.x`.

PR #2320, opened 2020-04-21
Introduce rabbit_upgrade_preparation

michaelklishin:
Part of rabbitmq/rabbitmq-cli#408.


michaelklishin:
Backported to `v3.8.x`.

PR #2330, opened 2020-04-28
QQ: Fix crash bug when reaching delivery limit

kjnilsson:
This only happens when using basic.get but would crash the quorum queue
when the delivery limit was reached due to the transient basic.get
consumer being removed


michaelklishin:
Backported to `v3.8.x`.

PR #2340, opened 2020-05-12
Add some missing license headers

michaelklishin:
Some files were missing license headers.


michaelklishin:
Backported to `v3.8.x`.

PR #2346, opened 2020-05-14
Initial work on supporting HA policies as Operator policies

atomicules:
Opening this PR for discussion - not expecting it to be merged as is 😉. Re-using some text from my commit message.

---

## Proposed Changes

This is initial work on supporting HA policies as Operator policies.

The reason for doing this is because these are the main things I want as operator policies from the point of view of hosting RabbitMQ. We can (and do) set default HA policies, but it is very easy for people to inadvertently overrule those (invariably they don't realise they need to add HA rules into their policies).

## Types of Changes

## Checklist

Regarding the CA:

I am employed by a company that offers RabbitMQ hosting and it's my experiences working there that lead to me making this PR, but I am making this PR from my personal account. I am authorised by my employer to contribute to open-source projects, but am not making this on their behalf. Therefore I've signed the personal CA.

Regarding tests passing:

`gmake ct-fast` runs and completes with some errors related to mirroring - the exact same errors are obtained on the master branch, but because there are errors I've not checked "All tests pass locally with my changes"

## Further Comments

- Built and tested on NetBSD from the point of view of running a single broker via `gmake run-broker` (`gmake` as I'm on NetBSD) and then using `rabbitmqctl` to set the policies

- Then built a rabbitmq-server-generic-unix package on OSX and tested via docker/docker-compose a simple three node cluster:

    Screenshot 2020-05-14 at 15 33 42

    Screenshot 2020-05-14 at 15 33 54

    Screenshot 2020-05-14 at 15 34 17

- I'm not sure how merging should work for HA. `ha-params` makes sense as a `Max`. For everything else I'm simply overriding `Val` with `OpVal` which is probably what I actually want. I need to think over that a bit more and play about with my test cluster. Maybe I only want to override in certain situations. E.g: HA of quorum nodes is ok and I don't need to override
with "all"? Hence why I have "opened for discussion"

- I'm also not 100% sure if the validations here just affect operator policies or whether they are going to interfere with normal policies. I.e. I am fine with operator policies for HA just allowing a `ha-mode` of `all` and `exactly` and not `nodes`, but we still want normal policies to allow that. Testing locally everything is ok. I can set a normal policy of `ha-mode` and `nodes` just fine still; but I don't understand the codebase enough to be 100% sure ;-)

- I haven't made any changes to the rabbitmq-management UI. That doesn't seem necessary just yet - I'd be getting ahead of myself. But would eventually be needed.

### DIFF
--- a/src/rabbit_policies.erl
+++ b/src/rabbit_policies.erl
@@ -53,13 +53,23 @@ register() ->
                           {operator_policy_validator, <<"max-in-memory-length">>},
                           {operator_policy_validator, <<"max-in-memory-bytes">>},
                           {operator_policy_validator, <<"delivery-limit">>},
+                          {operator_policy_validator, <<"ha-mode">>},
+                          {operator_policy_validator, <<"ha-params">>},
+                          {operator_policy_validator, <<"ha-sync-mode">>},
+                          {operator_policy_validator, <<"ha-promote-on-shutdown">>},
+                          {operator_policy_validator, <<"ha-promote-on-failure">>},
                           {policy_merge_strategy, <<"expires">>},
                           {policy_merge_strategy, <<"message-ttl">>},
                           {policy_merge_strategy, <<"max-length">>},
                           {policy_merge_strategy, <<"max-length-bytes">>},
                           {policy_merge_strategy, <<"max-in-memory-length">>},
                           {policy_merge_strategy, <<"max-in-memory-bytes">>},
-                          {policy_merge_strategy, <<"delivery-limit">>}]],
+                          {policy_merge_strategy, <<"delivery-limit">>},
+                          {policy_merge_strategy, <<"ha-mode">>},
+                          {policy_merge_strategy, <<"ha-params">>},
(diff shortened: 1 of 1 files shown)


michaelklishin:

Operator policies were originally introduces as a way to enforce resource usage limits in RabbitMQ-as-a-Service scenarios, not really as a way to enforce policy management. Therefore it does not necessarily make sense to "pick the highest replication factor" as the strategy, in particular because we try hard to avoid [excessive mirroring](https://www.rabbitmq.com/ha.html#replication-factor) which is quite common.

I personally believe such features belong to the platform. For resource limits it made sense to have a policy-level enforcement because a fast or bogus publisher can exceed certain limits in a very short period of time, so it would be difficult for the platform to react.

On top of that this does not support quorum queues. Classic mirrored queues hopefully will be entirely gone one day, perhaps in 4.0 or a similar major release.

atomicules:
Ok, thanks for the prompt feedback. I'll explore other ways of achieving this then outside the server codebase.

PR #2352, opened 2020-05-22
Add honeycomb integration for tests

pjk25:
Report test results from Github Actions to honeycomb for statistical analysis

`honeycomb_cth.erl` has been duplicated from https://github.com/rabbitmq/rabbitmq-management/pull/811 but, I don't think it is worth determining it's proper home until it's clear whether honeycomb will be a long term thing.

### DIFF
--- a/test/honeycomb_cth.erl
+++ b/test/honeycomb_cth.erl
@@ -1,0 +1,105 @@
+-module(honeycomb_cth).
+
+-export([id/1]).
+-export([init/2]).
+
+-export([pre_init_per_testcase/4]).
+-export([post_end_per_testcase/5]).
+
+-record(state, {directory, github_workflow, github_run_id,
+                github_repository, github_sha, github_ref,
+                base_rmq_ref, secondary_umbrella,
+                erlang_version, elixir_version,
+                otp_release, cpu_topology, schedulers,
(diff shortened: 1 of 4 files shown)


test/honeycomb_cth.erl
```diff
@@ -0,0 +1,101 @@
...
+                erlang_version = os:getenv("ERLANG_VERSION", "unknown"),
+                elixir_version = os:getenv("ELIXIR_VERSION", "unknown"),
+                otp_release = erlang:system_info(otp_release),
+                cpu_topology_json = cpu_topology_json(erlang:system_info(cpu_topology)),
+                schedulers = erlang:system_info(schedulers),
+                system_architecture = erlang:system_info(system_architecture),
+                system_memory_data_json = json_string_memory(memsup:get_system_memory_data())}}.
+
+pre_init_per_testcase(_Suite, TC, Config, #state{start_times = StartTimes} = State) ->
+    {Config, State#state{start_times = StartTimes#{TC => erlang:timestamp()}}}.
```
pjk25, line 37: This start times map is created only on init. It should probably be created per-suite to avoid testcase name clashes

pjk25: I have corrected this

PR #2355, opened 2020-05-29
Introduce raft.data_dir to new style configuration schema

### LINKED ISSUE #2354: Expose Ra data directory configuration to rabbitmq.conf
With the `rabbit_env` refactoring backported into `v3.8.x` this now becomes feasible since a couple of chicken-and-egg problems are gone.

The proposal therefore is to expose `raft.data_dir` via configuration, first as an experiment (so, not mentioned in the docs), and then as we  change how Ra is started during boot process, a recommended way over `RABBITMQ_QUORUM_DIR`.

References #2317.

michaelklishin:
As an experimental feature for now. To be sure that this value
has the desired effect we may have to tweak the way Ra dependency
is started.

Some early tests suggest this works as expected already.

Closes #2354 together with https://github.com/rabbitmq/rabbitmq-cli/pull/423.

### DIFF
--- a/priv/schema/rabbit.schema
+++ b/priv/schema/rabbit.schema
@@ -1666,6 +1666,18 @@ end}.
     end
 }.
 
+{mapping, "raft.data_dir", "ra.data_dir", [
+  {datatype, string}
+]}.
+
+{translation, "ra.data_dir",
+    fun(Conf) ->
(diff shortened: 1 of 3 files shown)


michaelklishin:
Travis failures have to do with Bintray host resolution by apt. Testing this locally.

michaelklishin:
Backported to `v3.8.x`.

PR #2358, opened 2020-06-01
Handle undefined exchange

jshen28:
## Proposed Changes

Occasionally, I saw rabbitmq start failed because for some reason
exchange is undefined. Under this case, maybe it is reasonable to
skip this exchange and go on the clean up process.


michaelklishin:
Thank you for taking the time to contribute. We would like to understand under what conditions you'd get an `undefined` instead of exchange record. Please post your findings together with logs and RabbitMQ version information to the mailing list.

I'm afraid we cannot accept this PR without understanding what is going on.

Issue #2317, opened 2020-04-16
RABBITMQ_QUORUM_DIR usage causes feature flags to not get enabled


Vanlightly:
If you set the RABBITMQ_QUORUM_DIR to a value that is not the mnesia directory, then on start-up, none of the feature flags get enabled.

This seems to be related to the is_virgin_node() function in rabbit_mnesia.erl. It relies on rabbit_node_monitor:quorum_filename() to return the correct quorum queue data_dir but it doesn't. It is hardcoded:

```
quorum_filename() ->
    filename:join(rabbit_mnesia:dir(), "quorum").
```

michaelklishin:
@dumbbell I could reproduce this and this (or a related `rabbit_env`) change has affected the tile.

michaelklishin:
So it must stem from [rabbit_env:quorum_queue_dir/1`](https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_env.erl#L833) but I haven't traced it down further yet.

michaelklishin:
`rabbit_feature_flags:is_supported_locally/1` returns `false` for the `quorum_queues` flag in the problematic case. So something causes the flag to not be registered.

michaelklishin:
Curious I no longer can reproduce this but now see the findings by @Vanlightly 👍

michaelklishin:
`rabbit_mnesia:is_virgin_node/0` was last changed in 8600e543e384470f972d81238a94449433ea0645.

michaelklishin:
Waiting for feedback to see if #2359 addressed this.

Vanlightly:
I can confirm that this is fixed.

PR #2359, opened 2020-06-02
rabbit_mnesia:is_virgin_node/0: take overridden Ra data directory into account

michaelklishin:
as well as the default location because it can exist for various reasons
early on node startup.

Part of rabbitmq/rabbitmq-server#2317.

### DIFF
--- a/src/rabbit_mnesia.erl
+++ b/src/rabbit_mnesia.erl
@@ -558,6 +558,7 @@ dir() -> mnesia:system_info(directory).
 %% file.
 init_db(ClusterNodes, NodeType, CheckOtherNodes) ->
     NodeIsVirgin = is_virgin_node(),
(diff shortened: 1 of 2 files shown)


michaelklishin:
@Vanlightly I tried to investigate the hypothesis mentioned in #2317 and cannot immediately see how the hardcoded path can cause `rabbit_mnesia:is_virgin_node/0` return an incorrect value since the Ra data directory is one of the ignored ones.

Nonetheless, this PR takes actual Ra data director into account and adds additional logging. It would be great to see if you can reproduce #2317 with this PR.

dumbbell:
This looks good to me. I retriggered the GitHub Actions tests because there is a failure in `simple_ha_SUITE`. Let's see how it goes.

michaelklishin:
Backported to `v3.8.x`.

PR #2362, opened 2020-06-03
Use the newly defined MSG_STORE_WORKER_WAIT for rabbit_msg_store

pjk25:
rabbit_msg_store in particular may need on the order of minutes to shutdown gracefully when there are millions of persistent messages


michaelklishin:
Backported to `v3.8.x`.

Issue #2363, opened 2020-06-03
Failing operations on users and permissions should be logged as such


jordanfranssen:
Environment
RabbitMQ: 3.8.4
OS: CentOS 7

Additional logging is needed to meet cyber security guidelines in certain organizations. Below are several scenarios that do not produce expected logging.

The log file referenced is the rabbit.log (e.g. /var/log/rabbitmq/rabbit.log).

Logging configuration for RabbitMQ is set as follows in rabbitmq.conf:
```
log.file.level          = debug
log.default.level       = debug
log.connection.level    = debug
log.channel.level       = debug
log.queue.level         = debug
log.mirroring.level     = debug
log.upgrade.level       = debug
```

## Adding an Already-Existing User

In this case, I attempt to add a user that already exists. I expect the log to recognize the failure of the action due to the existence of that user.

Command (user "foobar" already **exists**):
`rabbitmqctl add_user foobar`

Command Line Output:
```
Adding user "foobar" ...
Password:
Error:
User "foobar" already exists
```

Log File:
```
2020-06-03 18:17:49.167 UTC [info] <0.28605.30> Creating user 'foobar'
```

## Deleting a Non-Existent User

In this case, I attempt to delete a user that does not exist. I expect the log to recognize the failure of the action and the reason.

Command (user "foobar" **does not exist**):
`rabbitmqctl delete_user foobar`

Command Line Output:
```
Deleting user "foobar" ...
Error:
{:no_such_user, "foobar"}
```

Log File:
```
2020-06-03 18:23:51.507 UTC [info] <0.29395.30> Deleting user 'foobar'
```

## Setting Permission for a Non-Existent User

In this case, I attempt to set permissions for a user that does not exist. I expect the log to recognize the failure of this action and the reason.

Command (user foobar **does not exist** ):
`rabbitmqctl set_permissions foobar '.*' '.*' '.*'`

Command Line Output:
```
Setting permissions for user "foobar" in vhost "/" ...
Error:
{:no_such_user, "foobar"}
```

Log File:
`2020-06-03 18:55:53.788 UTC [info] <0.32518.30> Setting permissions for 'foobar' in '/' to '.*', '.*', '.*'`

## Clearing Permissions

In this case, I successfully clear permissions for the user "foobar". I expect the log to recognize the success of this action.

Command (user foobar **exists** )
`rabbitmqctl clear_permissions foobar`

Command Line Output:
```
[...]

PR #2364, opened 2020-06-04
Improves logging for user and permission-related management functions

michaelklishin:
Closes rabbitmq/rabbitmq-server#2363.


michaelklishin:
Backported to `v3.8.x`.

PR #2365, opened 2020-06-04
Stricter test setup for simple_ha_SUITE

pjk25:
This change is aimed at reducing spurious test failures in CI.


michaelklishin:
Should this be backported or an effectiveness conclusion is still pending?

pjk25:
I believe yes. It's run 4 or 5 times, and the suite in question hasn't flaked. Even if this fails to fix it, the error should be more informative.

michaelklishin:
Backported to `v3.8.x`.

pjk25:
Backported to 3.8 https://github.com/rabbitmq/rabbitmq-server/commit/4112120bddd0953a2917112228786332a5e59967

pjk25:
Backported to `v3.7.x`.

PR #2366, opened 2020-06-04
Deprecate 'ctl node_health_check'

michaelklishin:
* It requires a fully booted node, so not generally suitable for a Kubernetes readiness probe.
 * It can produce false positives
 * It is too intrusive and CPU-intensive to use at scale
 * Most operators do not understand what it really does and when they learn about it,
   consider it to be too opinionated and intrusive

Time for the One True Health Check™ to retire from duty.

Part of rabbitmq/rabbitmq-cli#426

Per discussion with @lukebakken


michaelklishin:
Backported to `v3.8.x`.

PR #2374, opened 2020-06-09
Make connection_metric_idemp_test example rather than property based

pjk25:
The property test flakes occasionally, and without consistent input values.
Idempotent behavior should be properly demonstrated with 2 or more attempts at the
action, so I think we are safe to reduce this from the 25 random examples to 3.


michaelklishin:
Should this be backported to `v3.8.x`?

pjk25:
Backported to `v3.8.x`.

PR #2376, opened 2020-06-09
Stricter setup in the confirms_survive_* tests

pjk25:
Wait until at least one message has been confirmed before triggering the survivable change


michaelklishin:
Should this be backported to `v3.8.x`?

pjk25:
From recent runs they appear to be effective, so yes. I'll go ahead and do so shortly.

pjk25:
Backported to `v3.8.x`.

PR #2378, opened 2020-06-10
rabbit_prelaunch: Load configuration before Erlang dist. is started

### LINKED ISSUE #2377: Configured net_ticktime is ignored in RabbitMQ 3.8.4
This problem was reported on the mailing-list:
https://groups.google.com/forum/#!topic/rabbitmq-users/PGG-LVpUJSM/discussion

Indeed, `rabbit_prelaunch` starts Erlang distribution before the configuration is checked and loaded:
https://github.com/rabbitmq/rabbitmq-server/blob/02aa73c36b720ae713d182ede1b8fc59966fcb41/apps/rabbitmq_prelaunch/src/rabbit_prelaunch.erl#L107-L111

dumbbell:
The reason is that kernel's `net_ticktime` can be configured in the Cuttlefish or Erlang configuration files. If we start the distribution first, then the configuration parameter is ignored.

Fixes #2377.

### DIFF
--- a/apps/rabbitmq_prelaunch/src/rabbit_prelaunch.erl
+++ b/apps/rabbitmq_prelaunch/src/rabbit_prelaunch.erl
@@ -104,11 +104,11 @@ do_run() ->
     %% 1. Erlang/OTP compatibility check.
     ok = rabbit_prelaunch_erlang_compat:check(Context),
 
-    %% 2. Erlang distribution check + start.
-    ok = rabbit_prelaunch_dist:setup(Context),
-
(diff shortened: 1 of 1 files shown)


michaelklishin:
Package and [upgrade tests](https://ci.rabbitmq.com/teams/main/pipelines/server-release:v3.9.x/jobs/test-upgrade-from-3.6-release-and-erlang-19.3/builds/179) are also green.

michaelklishin:
Backported to `v3.8.x`.

dumbbell:
Thank you @michaelklishin!

PR #2396, opened 2020-06-26
Import definitions in the post-launch phase

### LINKED ISSUE #2384: Perform definition import after plugins are enabled
Currently there is a chicken-and-egg problem triangle between

 * Default data seeding
 * [Definition import](https://www.rabbitmq.com/definitions.html) on boot
 * Plugin activation

which can be problematic when definitions related to or dependent on plugins are involved, e.g. an exchange of a custom plugin-provided type.

We can potentially untangle this with a few low risk changes:

 * Avoid default data seeding when `load_definitions` is configured
 * Load definitions not in a core boot step but explicitly in a "post-run" phase, after plugins have been enabled
 * As a bonus, consider enabling client connection listeners after all plugins have been enabled (but this can be problematic: consider a Shovel that has to connect to the local node upon startup)

While definitions can be imported at any point after node startup, it would be convenient for Shovel and custom exchange type users if this limitation of `load_definitions` was lifted.

See #2381 for the background.

michaelklishin:
To make sure definitions that depend on plugins (federation upstreams, shovels, exchanges of custom types) can be imported.

This preserves an implicit feature of the previous implementation: since definition import was performed in a boot step
that preceded the empty DB check one in the boot step DAG, a node with a definition file to load would not seed its default data (virtual host and user).

Closes #2384.

### DIFF
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -981,9 +981,14 @@ recover() ->
 -spec maybe_insert_default_data() -> 'ok'.
 
 maybe_insert_default_data() ->
-    case rabbit_table:needs_default_data() of
-        true  -> insert_default_data();
-        false -> ok
+    NoDefsToImport = not rabbit_definitions:has_configured_definitions_to_load(),
+    case rabbit_table:needs_default_data() andalso NoDefsToImport of
+        true  ->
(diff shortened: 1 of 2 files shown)


michaelklishin:
Backported to `v3.8.x`.

PR #2397, opened 2020-06-30
Introduce a way to check if a node is the only cluster member

michaelklishin:
Part of rabbitmq/rabbitmq-cli#432


michaelklishin:
Backported to `v3.8.x`.

PR #2400, opened 2020-07-03
Definition import: pre-format more values as maps where possible

michaelklishin:
To make it easier to serialize the returned value and rely less on Erlang- or RabbitMQ-specific collection types such as proplists (meh) and AMQP 0-9-1 argument tables (which are typed proplists, still a meh).

This also adds some export-then-reimport tests that discovered the issue with x-argument formatting not being "JSON-friendly" without post-processing.

@lukebakken  I will investigate if the issue that https://github.com/rabbitmq/rabbitmq-server/commit/460a432ddb2337dad6a7ca350497a8e2aee5063f tries to address is still relevant in one of the commercial plugins. As far as I can tell, it should not be, or at least should be addressed on the import side.


michaelklishin:
Backported to `v3.8.x`.

PR #2402, opened 2020-07-07
Attempt to further de-flake metrics/connection_metric_idemp_test

pjk25:
Wait until the connection_metrics report the new connections, before
triggering the idempotent action under which they should stay stable.

Depends on https://github.com/rabbitmq/rabbitmq-ct-helpers/pull/42


pjk25:
@michaelklishin @dumbbell I believe with this change we just eliminate a race condition, rather than change the meaning of the test (which by name is about being idempotent anyway). Agreed?

michaelklishin:
Yes, the test is timing dependent as is.

michaelklishin:
Backported to `v3.8.x`.

PR #2417, opened 2020-07-24
Mark node as ready after it actually is

dumbbell:


michaelklishin:
Backported to `v3.8.x`. While the maintenance mode is not yet in `v3.8.x` (will be for `3.8.7` or so), this change makes sense there as well.

dumbbell:
Thank you!

PR #2436, opened 2020-08-28
Add auth attemp metrics

dcorbacho:
## Proposed Changes

The number of successful and failed authorization attempts per ip and user can be useful for troubleshooting, however the addition of these metrics has some implications. When the attempt metrics contain the ip and username, they cardinality could grow infinitely. This is not only an issue for Prometheus, which we've configured to aggregate by default since 3.8.2, but internally the ets table containing the metrics doesn't have a criteria for clean up. Thus, these metrics will have special treatment.
Auth attempt metrics:
- Are aggregated by default with its own switch on the broker: {rabbit, [{return_per_user_auth_attempt_metrics, false}]}
-  Are not handled as other rabbit metrics, there is no aggregation or computation of stats
-  Have a HTTP API management endpoint to query and reset
-  Have a CLI command to enable/dissable aggregation, reset and list metrics
-  Should only be used by the operator in the non-aggregated mode for troubleshootting or on environments where the number of ip and users are guaranteed to be bound
- Could cause a memory leak if used on non-aggregated mode
- Are exported by the Prometheus plugin
- Must be incremented by each protocol, the server does it for AMQP 0.9.1 and  AMQP 1.0 and MQTT on their respective plugins. We don't have IP information when the AMQP 1.0 authorization happens, so this one reports an empty ip on non-aggregated mode.

## Types of Changes

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,7 @@ define PROJECT_ENV
 	    %% interval at which connection/channel tracking executes post operations
 	    {tracking_execution_timeout, 15000},
 	    {stream_messages_soft_limit, 256},
-        {return_per_user_auth_attempt_metrics, false}
+        {track_auth_attempt_source, false}
 	  ]
 endef
 
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,8 @@ define PROJECT_ENV
 	    {writer_gc_threshold, 1000000000},
 	    %% interval at which connection/channel tracking executes post operations
 	    {tracking_execution_timeout, 15000},
-	    {stream_messages_soft_limit, 256}
+	    {stream_messages_soft_limit, 256},
+        {return_per_user_auth_attempt_metrics, false}
 	  ]
 endef
 
--- a/src/rabbit_reader.erl
+++ b/src/rabbit_reader.erl
@@ -1413,15 +1413,19 @@ auth_phase(Response,
                                        auth_mechanism = {Name, AuthMechanism},
                                        auth_state     = AuthState},
                        sock = Sock}) ->
+    Ip = list_to_binary(inet:ntoa(Connection#connection.host)),
     case AuthMechanism:handle_response(Response, AuthState) of
         {refused, Username, Msg, Args} ->
+            rabbit_core_metrics:auth_attempt_failed(Ip, Username),
             auth_fail(Username, Msg, Args, Name, State);
         {protocol_error, Msg, Args} ->
+            rabbit_core_metrics:auth_attempt_failed(Ip, ""),
             notify_auth_result(none, user_authentication_failure,
                                [{error, rabbit_misc:format(Msg, Args)}],
                                State),
             rabbit_misc:protocol_error(syntax_error, Msg, Args);
         {challenge, Challenge, AuthState1} ->
+            rabbit_core_metrics:auth_attempt_succeeded(Ip, ""),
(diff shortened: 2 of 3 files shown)


gerhard:
I have noticed that logins via the API are not counted. Should they?

Makefile
```diff
@@ -115,7 +115,8 @@ define PROJECT_ENV
 	    %% Default max message size is 128 MB
 	    {max_message_size, 134217728},
 	    %% Socket writer will run GC every 1 GB of outgoing data
-	    {writer_gc_threshold, 1000000000}
+	    {writer_gc_threshold, 1000000000},
+	    {return_per_user_auth_attempt_metrics, false}
```
gerhard, line 119: What do you think about renaming this config to `track_auth_attempt_source`?

src/rabbit_reader.erl
```diff
@@ -1394,25 +1394,31 @@ auth_phase(Response,
                                        auth_mechanism = {Name, AuthMechanism},
                                        auth_state     = AuthState},
                        sock = Sock}) ->
+    Ip = list_to_binary(inet:ntoa(Connection#connection.host)),
     case AuthMechanism:handle_response(Response, AuthState) of
         {refused, Username, Msg, Args} ->
+            rabbit_core_metrics:auth_attempt_failed(Ip, Username),
```
gerhard, line 1400: Should we track the protocol as well? In this case it would be `amqp091`.

michaelklishin: That's a great idea.

michaelklishin (approved):

michaelklishin:
This is a good starting point and does give operators a way to find out what IP and usernames repeatedly fail authentication but there's definitely more to do:

 * Consider if this is enough for a sensible Grafana chart
 * Support STOMP
 * Documentation, including `man` pages

michaelklishin:
Backported to `v3.8.x`.

michaelklishin:
This introduced a seriou regression (client connections for AMQP 0-9-1 and AMQP 1.0 are no longer possible) for users who run with

``` ini
reverse_dns_lookup = true
```

```
2021-01-20 19:16:44.423 [error] <0.1157.0> closing AMQP connection <0.1157.0> (localhost:51588 -> localhost:5672 - repl 0.992502025475856):
{handshake_error,starting,0,{error,badarg,'connection.start_ok',[{erlang,list_to_binary,[{error,einval}],[]},{rabbit_reader,auth_phase,2,[{file,"src/rabbit_reader.erl"},{line,1416}]},{rabbit_reader,handle_method0,3,[{file,"src/rabbit_reader.erl"},{line,1136}]},{rabbit_reader,handle_input,3,[{file,"src/rabbit_reader.erl"},{line,1050}]},{rabbit_reader,recvloop,4,[{file,"src/rabbit_reader.erl"},{line,480}]},{rabbit_reader,run,1,[{file,"src/rabbit_reader.erl"},{line,462}]},{rabbit_reader,start_connection,4,[{file,"src/rabbit_reader.erl"},{line,361}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,226}]}]}}
2021-01-20 19:16:44.424 [info] <0.1160.0> Closing all channels from connection '<<"localhost:51588 -> localhost:5672">>' because it has been closed
```

PR #2441, opened 2020-09-02
Test with Erlang 22.3 and 23.0

pjk25:
Replace the GitHub Actions workflow for Erlang 21.3 with 22.3

Therefore the set of tested Erlang is #{22.3, 23.0}.

### DIFF
--- a/.github/workflows/test-erlang-otp-22.3.yaml
+++ b/.github/workflows/test-erlang-otp-22.3.yaml
@@ -1,6 +1,6 @@
 # vim:sw=2:et:
 # https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow
-name: "Test - Erlang 21.3"
+name: "Test - Erlang 22.3"
 on:
   push:
   repository_dispatch:
@@ -18,9 +18,9 @@ jobs:
       - name: CONFIGURE OTP & ELIXIR
         uses: actions/setup-elixir@v1
         with:
-          otp-version: 21.3
+          otp-version: 22.3
           # https://github.com/elixir-lang/elixir/releases
-          elixir-version: 1.8.0
+          elixir-version: 1.10.3
       - name: CHECK RABBITMQ COMPONENTS
         # https://github.community/t5/GitHub-Actions/How-can-I-set-an-expression-as-an-environment-variable-at/m-p/41804/highlight/true#M4751
         id: ref
@@ -59,12 +59,12 @@ jobs:
         uses: actions/cache@v1
         with:
           path: umbrellas
-          key: secondary-umbrellas-v3.7.27-rc.1-v3.8.5-erlang-21.3-rev4
+          key: secondary-umbrellas-v3.7.28-v3.8.8-erlang-22.3-rev4
       - name: PREPARE SECONDARY UMBRELLA COPIES
         if: success() && 'oldest' == 'oldest'
         run: |
           set -ex
-          for version in v3.7.27-rc.1 v3.8.5; do
+          for version in v3.7.28 v3.8.8; do
             umbrella="umbrellas/$version"
             if ! test -d "$umbrella"  ||
                ! make -C "$umbrella/deps/rabbit" test-dist; then
@@ -122,9 +122,9 @@ jobs:
         if: success() && 'oldest' == 'latest'
         uses: actions/setup-elixir@v1
         with:
-          otp-version: 21.3
+          otp-version: 22.3
           # https://github.com/elixir-lang/elixir/releases
-          elixir-version: 1.8.0
+          elixir-version: 1.10.3
       - name: DOWNLOAD DEPS ARCHIVE
         if: success() && 'oldest' == 'latest'
         uses: actions/download-artifact@v2
@@ -156,9 +156,9 @@ jobs:
       - name: CONFIGURE OTP & ELIXIR
         uses: actions/setup-elixir@v1
         with:
-          otp-version: 21.3
+          otp-version: 22.3
           # https://github.com/elixir-lang/elixir/releases
-          elixir-version: 1.8.0
+          elixir-version: 1.10.3
       - name: DOWNLOAD DEPS ARCHIVE
         uses: actions/download-artifact@v2
         with:
@@ -190,9 +190,9 @@ jobs:
       - name: CONFIGURE OTP & ELIXIR
         uses: actions/setup-elixir@v1
         with:
-          otp-version: 21.3
+          otp-version: 22.3
           # https://github.com/elixir-lang/elixir/releases
-          elixir-version: 1.8.0
+          elixir-version: 1.10.3
       - name: DOWNLOAD DEPS ARCHIVE
         uses: actions/download-artifact@v2
         with:
@@ -205,8 +205,8 @@ jobs:
           branch_or_tag_name=${GITHUB_REF#refs/*/}
           ! test -d ebin || touch ebin/*
           export BASE_RMQ_REF=master
-          export ERLANG_VERSION=21.3
-          export ELIXIR_VERSION=1.8.0
+          export ERLANG_VERSION=22.3
+          export ELIXIR_VERSION=1.10.3
           make ct-amqqueue_backward_compatibility \
             base_rmq_ref=master \
             current_rmq_ref=$branch_or_tag_name \
@@ -285,9 +285,9 @@ jobs:
       - name: CONFIGURE OTP & ELIXIR
         uses: actions/setup-elixir@v1
         with:
-          otp-version: 21.3
+          otp-version: 22.3
           # https://github.com/elixir-lang/elixir/releases
-          elixir-version: 1.8.0
+          elixir-version: 1.10.3
       - name: DOWNLOAD DEPS ARCHIVE
         uses: actions/download-artifact@v2
         with:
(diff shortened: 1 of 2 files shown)


pjk25:
Waiting for green GitHub Actions before merging.

.github/workflows/test-erlang-otp-22.3.yaml
```diff
@@ -18,9 +18,9 @@ jobs:
       - name: CONFIGURE OTP & ELIXIR
         uses: actions/setup-elixir@v1
         with:
-          otp-version: 21.3
+          otp-version: 22.3
           # https://github.com/elixir-lang/elixir/releases
-          elixir-version: 1.8.0
+          elixir-version: 1.10.3
```
gerhard, line 23: This has changed in multiple place. [This tells me that we are still on 1.8](https://github.com/rabbitmq/rabbitmq-common/blob/d4efb0a4a243a1cad326f7f01e6bb006344916d3/mk/rabbitmq-github-actions.mk#L9). Can you confirm @dumbbell?

gerhard:
Thanks for picking this up 😉

.github/workflows/test-erlang-otp-22.3.yaml
pjk25: @gerhard Looks like that was pollution from my local env. I've regenerated the files. Thanks for pointing it out. I noticed it in the diff but didn't realize it was unintended.

gerhard:
The workflow configs now looks good to me.

Can you confirm @michaelklishin if the [3 per_user_connection failures in are 22.3](https://github.com/rabbitmq/rabbitmq-server/runs/1060947344??check_suite_focus=true) genuine? This is related to https://github.com/rabbitmq/rabbitmq-server/pull/2380. The [per_vhost_connection_limit somewhat related failure seems suspicious](https://github.com/rabbitmq/rabbitmq-server/runs/1060947420?check_suite_focus=true).

As soon as all checks pass this is good to merge from my POV.

pjk25:
While there were some failures in GitHub Actions, I think this branch is safe to merge so that any necessary fixes can happen on master.

dumbbell:
I'm preparing a patch to `per_vhost_connection_limit_SUITE.erl`. The testsuite is calling a function which is new in `master`, thus failing with mixed-versions clusters. However, that new function only calls another function which exists in all release branches. The patch is small; I'm waiting for the last run to finish.

dumbbell:
About the newer lower version to support: we document that we support Erlang 22.x, so to me, the minimum should be set to 22.0, not 22.3.

That said, the documentation does not reflect what we do, because it says RabbitMQ works with Erlang 21.x, but it's not true: the minimum is 21.3.

dumbbell:
Here is the patch for `per_vhost_connection_limit_SUITE.erl`:
https://github.com/rabbitmq/rabbitmq-server/commit/693be55f88e4cb603a002f08daf6710ed6f3fdf5

michaelklishin:
22.0 does not get any more patch releases. The docs say we support [Erlang 22.x](https://www.rabbitmq.com/which-erlang.html) which can be interpreted as "any 22.x" or "latest 22.x" depending on who you ask.

michaelklishin:
I won't object to testing `22.0` but in practice, virtually no one today does or will run on Erlang 22.0 in the future. Nearly all package installations use 22.3 for obvious reasons. So why would we test on 22.0?

Standard [Debian repositories](https://packages.debian.org/search?keywords=erlang-base&searchon=names&suite=all&section=all) either offer 21.2 or 23.x, so this is not really a concern. Our own repos offer 22.3.x and 23.x.

gerhard:
I've just re-triggered the Erlang 22.3 build. Thanks for the quick patch @dumbbell! Any reason why the build passed on Erlang 23.0 but failed on 22.3?

Testing the latest minor for each OTP major makes most sense to me. While in a perfect world we would test against all minors, trusting the OTP team and testing against latest available minor feels pragmatic.

gerhard:
Same suites failed in the last build. Did the patch make it through @dumbbell?

dumbbell:
> I've just re-triggered the Erlang 22.3 build. Thanks for the quick patch @dumbbell! Any reason why the build passed on Erlang 23.0 but failed on 22.3?

We test mixed-versions clusters with the oldest supported version only.

> Testing the latest minor for each OTP major makes most sense to me. While in a perfect world we would test against all minors, trusting the OTP team and testing against latest available minor feels pragmatic.

Except that later patch release may have new functions that the initial .0 release didn't have.

> Same suites failed in the last build. Did the patch make it through @dumbbell?

Was the branch rebased on top of the latest master?

pjk25:
I will rebase it, hang on

michaelklishin (approved):

dumbbell:
I just updated the GitHub Actions workflows in `master` to run the recently added testsuites. They fail with this pull request, so I would like to learn if they fail in plain `master` as well.

pjk25:
https://github.com/rabbitmq/rabbitmq-server/commit/9ff1d6fa91209991f074d8295049979d6e3defdb & https://github.com/rabbitmq/rabbitmq-server/commit/839724befcb4c6da1417ba6e7ca39ec670b85e38 contain additional adjustments for tests, but have not been merged. @dumbbell do either of those change the validity of the suites?

gerhard:
> We test mixed-versions clusters with the oldest supported version only.
> Except that later patch release may have new functions that the initial .0 release didn't have.

Are you saying that we test from first minor in the min supported major, 21.0? If that's the case, then we need to resolve the contradiction with [Supported Erlang Version Policy](https://www.rabbitmq.com/which-erlang.html) (we write OTP 21.3 there, but do 21.0 in our mixed tests) :

image

> I just updated the GitHub Actions workflows in master to run the recently added testsuites. They fail with this pull request, so I would like to learn if they fail in plain master as well.

I support taking a more cautious approach with this.

dumbbell:
> Are you saying that we test from first minor in the min supported major, 21.0? If that's the case, then we need to resolve the contradiction with Supported Erlang Version Policy (we write OTP 21.3 there, but do 21.0 in our mixed tests)

No, we test against 21.3 because that's the minimum supported (not just on the paper, but in the code).

To me this is in contradiction with the following general sentence:
> RabbitMQ supports two most recent Erlang release series: 23.x and 22.x, plus 21.x (...)

21.x means any version in the 21 release branch. We didn't do that for 21.x, but that's ok: I think that documentation came after the minimum was bumped to 21.3.

But that's why I think we should support 22.0: that's what we state and 22.0 means "compatible with later minor releases on that branch". The opposite is not true (being compatible with 22.3 does not mean it will work with older 22.x minor releases).

michaelklishin:
We don't state `22.0` specifically anywhere and no user will run on 22.0 when 22.3 has been around for a while. Most our team uses the latest available minor in practice. If a major Ubuntu LTS release shipped with 22.0 I'd understand but [that's not the case](https://packages.ubuntu.com/search?keywords=erlang-base&searchon=names&suite=all&section=all): Eoan has 22.2 and will likely have 22.3 via backports eventually but never 22.0.

Testing on 22.0 makes no practical sense to me. In fact, I have rewritten the Erlang compatibility guide to say "the most recent minor version" locally earlier today because that's what we have been effectively doing in the last year or so, without changing the language.

michaelklishin:
The new tests fail on Erlang 21.3 in master. This is why I originally suggested that since we can drop 21.x support now, we probably should instead of spending more time on it. Per-user limits are going to make it into `3.8.9` the earliest, and we can drop 21.3 support by then, or at least document it as an unsupported version.

dumbbell:
@pjk25: https://github.com/rabbitmq/rabbitmq-server/commit/839724befcb4c6da1417ba6e7ca39ec670b85e38 is a good idea! If this is an actual wait loop, you can set an even longer timeout (like one minute) because it will return as soon as the condition is met. Feel free to merge it as soon as you want because the problem happens with plain `master`.

For the patch to `feature_flags_SUITE.erl`, are you sure the partition is always detected properly with the shorter timer?

pjk25:
https://github.com/rabbitmq/rabbitmq-server/commit/839724befcb4c6da1417ba6e7ca39ec670b85e38 became https://github.com/rabbitmq/rabbitmq-server/commit/931f7d4fbca334b5da99ec2cc45121c8a2917dde when I cherry picked it over to master. While it is a wait loop underneath, I prefer to keep timeouts low to reduce the chances they mask other issues.

As for https://github.com/rabbitmq/rabbitmq-server/commit/9ff1d6fa91209991f074d8295049979d6e3defdb, it was my impression that the partition was already in place, but we wanted to heal it after invoking the rpc call, so that the call happens while the partition was in place. Even 1 second seems enough in that case, unless the original 5s corresponds to an interval within the broker that I am unaware of.

gerhard:
@dumbbell @michaelklishin I agree that the **Minimum required Erlang/OTP** should become 22.3 for the next RabbitMQ release.

[Per-user limits are still failing on 22.3](https://github.com/rabbitmq/rabbitmq-server/runs/1085212024?check_suite_focus=true). So does [maintenance mode](https://github.com/rabbitmq/rabbitmq-server/runs/1085207508). Everything passes of 23.0

pjk25:
I looked briefly at the per-user limits, but the cause was not readily apparent to me.

michaelklishin:
All GitHub Actions now pass, will merge.

PR #2443, opened 2020-09-10
Support recently introduced Ra WAL settings in rabbitmq.conf

michaelklishin:
This adds the settings from https://github.com/rabbitmq/ra/pull/196 to `rabbitmq.conf`.


michaelklishin:
Backported to `v3.8.x`.

PR #2450, opened 2020-09-18
Increase aten poll interval

kjnilsson:
A longer interval will reduce the likelihood that unnecessary leader
elections are triggered in quorum queues due to transient network
issues. This increases it from the default of 1s to 5s which increases
the time it takes to detect genuine network partitions somewhat. Ra still uses
monitors so connection failures will not have to wait this additional
time.


michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #2468, opened 2020-10-14
Attempt at improving the channel count consistency (+ improvements to testcases)

dumbbell:
The main reason for this patch is the third commit in the branch.

Other commits are there to help testcases flakiness and CI.

### DIFF
--- a/src/rabbit_channel_tracking.erl
+++ b/src/rabbit_channel_tracking.erl
@@ -102,9 +102,9 @@ handle_cast({connection_closed, ConnDetails}) ->
             rabbit_log_connection:info(
                 "Closing all channels from connection '~p' "
                 "because it has been closed", [pget(name, ConnDetails)]),
+            %% Shutting down channels will take care of unregistering the
+            %% corresponding tracking.
             shutdown_tracked_items(TrackedChs, undefined),
-            [unregister_tracked(rabbit_tracking:id(ThisNode, Name)) ||
-                #tracked_channel{name = Name} <- TrackedChs],
             ok;
         _DifferentNode ->
             ok
--- a/test/many_node_ha_SUITE.erl
+++ b/test/many_node_ha_SUITE.erl
@@ -12,6 +12,11 @@
 
 -compile(export_all).
 
+suite() ->
+    [
+     {timetrap, {minutes, 5}}
+    ].
+
 all() ->
     [
       {group, cluster_size_6}
--- a/test/per_user_connection_channel_tracking_SUITE.erl
+++ b/test/per_user_connection_channel_tracking_SUITE.erl
@@ -10,6 +10,7 @@
 -include_lib("common_test/include/ct.hrl").
 -include_lib("amqp_client/include/amqp_client.hrl").
 -include_lib("eunit/include/eunit.hrl").
+-include_lib("rabbitmq_ct_helpers/include/rabbit_assert.hrl").
 
 -compile(export_all).
 
@@ -148,13 +149,13 @@ single_node_user_connection_channel_tracking(Config) ->
     ?assertEqual(true, is_process_alive(Conn1)),
     ?assertEqual(true, is_process_alive(Chan1)),
     close_channels([Chan1]),
-    ?assertEqual(0, count_channels_in(Config, Username)),
-    ?assertEqual(0, tracked_user_channel_count(Config, Username)),
-    ?assertEqual(false, is_process_alive(Chan1)),
+    ?awaitMatch(0, count_channels_in(Config, Username), 20000),
+    ?awaitMatch(0, tracked_user_channel_count(Config, Username), 20000),
+    ?awaitMatch(false, is_process_alive(Chan1), 20000),
     close_connections([Conn1]),
-    ?assertEqual(0, length(connections_in(Config, Username))),
-    ?assertEqual(0, tracked_user_connection_count(Config, Username)),
-    ?assertEqual(false, is_process_alive(Conn1)),
+    ?awaitMatch(0, length(connections_in(Config, Username)), 20000),
+    ?awaitMatch(0, tracked_user_connection_count(Config, Username), 20000),
+    ?awaitMatch(false, is_process_alive(Conn1), 20000),
 
     [Conn2] = open_connections(Config, [{0, Username2}]),
     Chans2  = [_|_] = open_channels(Conn2, 5),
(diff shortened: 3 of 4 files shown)


dumbbell:
@Ayanda-D: Hi! I'm trying to improve concurrent/conflicting updates to channel tracking without introducing transactions. Could you please take a look at the third commit in that branch and tell me what you think? I mean, is tracking still ok with the patch, or am I missing/breaking some cases?

Ayanda-D:
@dumbbell looks good, the `shutdown_tracked_items/2` step guarantees us emission of `channel_closed` events for the associated (active) channels as they shutdown, allowing the single point of un-registration on mnesia as you state. Channels terminating from any abrupt reasons (e.g. a terminating connection with loads of channels concurrently crashing or already crashed) will still run through the channel `terminate/2` callback to emit `channel_closed`.

Just ran relevant channel tracking tests from your branch, all passing OK! Thanks 👍

dumbbell:
Thank you @Ayanda-D for your very quick feedback!

michaelklishin:
@dumbbell we should remember to backport all these updates to `v3.8.x`.

dumbbell:
I will remember. I keep the tab open until it's done. But I would like to see this change (and many others) go through the entire pipeline :-)

michaelklishin:
Backported to `v3.8.x` because these tests are the only ones consistently failing in the `v3.8.x` pipeline. We will get better turnaround time there.

michaelklishin:
References #2380.

Ayanda-D:
@dumbbell @michaelklishin i highlight a certain case in previous response which I thought would spark some thoughts :-)
> a terminating connection with loads of channels concurrently crashing or already crashed

Imagine 1 connection with 2000 channels. Are we 100% fully certain that all terminating channels will behave as expected,
terminate as expected and clean-up their registered entities in Mnesia? In an ideal world, yes. I did some tests to prove
execution of the `terminate/2` callback of channels on different failure/crash scenarios, all operate as expected,
hence https://github.com/rabbitmq/rabbitmq-server/pull/2468#issuecomment-708391611 

However, the reason why channels are/were being un-registered here is to give the **connection** responsibility of ensuring
its own channels are un-registered from the tracking table (regardless of whether a channel un-registers itself or not
on termination, assuming for some reason or not, channels crashing will not terminate in a clean manner).

e.g. with queues, we've seen a lot of stale queue entities in Mnesia which are expected to be cleaned-up, or updated
consistently, but operators eventually have to carry out manual intervention with `rabbit_amqqueue:delete_immediately/1`.

This https://github.com/rabbitmq/rabbitmq-server/pull/2468/commits/1ce9df26951db79a052d75651c84e5ffca397846 will work fine, and does, but from an operators perspective, we don't have the luxury to come back and manually cleanup any dead channels which may still be in the tracking table(s) for some unexpected/unclean terminations, such as the case with "stale"/unresponsive queues.

Then on a 2nd look at [this](https://github.com/rabbitmq/rabbitmq-server/pull/2468/commits/1ce9df26951db79a052d75651c84e5ffca397846), another question I have is why are you calling a **"timing problem"** a **"double channel
count decrease"**? The main issue here is timing the deletion of tracked channels correctly, rather than dealing with
"double deletions". Calling `unregister_tracked/1` as was poses no harm if timed correctly, which is what we can
instead focus on, the timing, rather than removing this step completely.

All that's needed to solve this problem is to add a **delay** which is proportional to the number of channels, when `connection_closed` event is received, just before attempting to un-register tracked channels from the table. For
instance:

```
   shutdown_tracked_items(TrackedChs, undefined),
   Delay = length( TrackedChs ) x ChannelTerminationTimeout,
   timer:sleep(Delay),
   [unregister_tracked(rabbit_tracking:id(ThisNode, Name)) || #tracked_channel{name = Name} <- TrackedChs],
```

All events are handled asynchronously in their isolated processes, so that delay is non-blocking to any entity. If
during the `Delay` channels had terminated as expected, Mnesia lookup in `unregister_tracked/1` will not find the channels
and proceed with execution (so no chance of any conflicting concurrent Mnesia operations). If not, then `unregister_tracked/1` still/always guarantees that the channels are un-registered and cleaned-up.

That's the more appropriate solution to this timing issue, than removing double deletion which guarantees removal of
channels on weird cases (as we've seen with stale queue entries in `rabbit_queue` and `rabbit_durable_queue` tables).

The current implementation of vhost connection tracking already suffers from entries that aren't removed from tracking
tables when connections terminate. This seeks to ensure such problems don't exist in any scenario.

[ **NOTE**: Only looking at this with more scrutiny now. We/ESL are predominately on customer projects Mon-Fri, and have very limited time to look at on open-source work. Along with https://github.com/rabbitmq/rabbitmq-server/pull/2380 - we will respond to questions when time permits, apologies for the delays in advance ]

michaelklishin:
@Ayanda-D thanks for the follow-up. Without #2380, we have never had the tests originally introduced to pass on CI. This is a *major* problem for our team because we cannot ship RabbitMQ releases when the pipeline is blocked.

Adding a `timer:sleep/1` is not going to make this predictable. Making sure that channels are responsible for the removal of their entries mostly does. Double removal of things is harmless for the purpose of active channel tracking but from our experience, "timing things correctly" is impossible to achieve. This can be yet another case where a "metadata GC" could be introduced (much like what we have for metric samples), although it will also have to work concurrently with the regular tracking table updates and can produce inconsistencies in some very specifically timed scenarios.

dumbbell:
> Then on a 2nd look at this, another question I have is why are you calling a "timing problem" a "double channel
count decrease"?

When a connection is closed, the "connection closed" event triggers the termination of channels, plus the unregistration of the tracking. Each channel termination triggers the same unregistration. It happens that, at least in the associated testsuite, the two conflicting unregistrations of the same channel run concurrently. The code being outside of any transaction, the `mnesia:dirty_update_counter(...)` is called twice for the same channel. Thus the wrong metric value and the failing testcase.

I agree with @michaelklishin that a delay won't fix anything.

Ayanda-D:
> When a connection is closed, the "connection closed" event triggers the termination of channels, plus the unregistration of the tracking

Yep, The **Delay** merely postpones this action to **first** give time to individual channel terminations
to trigger their unregistrations. By the time the **Delay** elapses, channels would've terminated and
completed un-registrations and be [non-existent in the tracking tables](https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_channel_tracking.erl#L149-L150), thereby eliminating any
concerns on "double channel count decrease".

All channels are terminating in parallel:
https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_channel_tracking.erl#L59

The "delay" **will definitely work** - just as it is already being done/used for the `user_deleted` event:
https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_channel_tracking.erl#L115

No reason why it wouldn't work in this case, considering how fast channels terminate (no longer than
any few milli- or micro-seconds). Delaying by a few seconds would be more than sufficient for channels to
terminate and be wiped out from the tracking tables to avoid any interference.

> This can be yet another case where a "metadata GC" could be introduced 

I agree with @michaelklishin here - perhaps fixing this in bits & pieces not the most ideal, the need
for a more global "metadata GC" approach remains ( _unsolved_ 😭 ). This would help solve the remaining
problem of crashing processes that don't cleanup as expected - such as seen with 'stale'/'unresponsive'
queues as well which are causing headaches for operators (`amqqueue-process` manifests this, can't
rule out `channel` processes).

I'm even more concerned here with dealing with such cases under high connection & channel churn.
(1 or 2 unresponsive 'stale'/'unresponsive' queues is already hard enough to GC. Imagine numerous
channels 👀 in tracking tables ).

Those are my 2-cents on this, I guess my concerns here are on the 'unforeseen/unexpected' issues which,
as with queues, we tend to see in rare corner cases on live environments. Probably will be valid only
until we see such problems on production environments. Let's see how it goes in prod environments 👍

PR #2476, opened 2020-10-19
Stop local quorum queue followers after transferring leadership

### LINKED ISSUE #2474: Stop Ra nodes of QQ followers when a node is put under maintenance
Currently when a RabbitMQ node hosting quorum queue replicas is put under maintenance, it will [only stop the Ra nodes for quorum queues that had their leader on this RabbitMQ node](https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_maintenance.erl#L226).

Followers will not be stopped and would still be considered for leader elections. They should be shut down as well.

michaelklishin:
This way the shutdown of a node that's been put into maintenance
mode won't have any further effects on the quorum
(or rather, the effects will be "frontloaded" which is the goal
of maintenance mode).

Closes #2474.

### DIFF
--- a/src/rabbit_maintenance.erl
+++ b/src/rabbit_maintenance.erl
@@ -89,6 +89,7 @@ do_drain() ->
                     [length(TransferCandidates), ReadableCandidates]),
     transfer_leadership_of_classic_mirrored_queues(TransferCandidates),
     transfer_leadership_of_quorum_queues(TransferCandidates),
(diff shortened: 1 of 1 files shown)


michaelklishin:
I found a test that checks that there are no more replicas on the node to be relatively useless but we can add it if needed.

michaelklishin:
Backported to `v3.8.x`.

PR #2481, opened 2020-10-23
Add rabbit_non_amqp_connections process group

acogoluegnes:
This commit allows to propagate the force_event_refresh
event to non-AMQP connections. This way other types of connections can
be notified of the management plugin activation and start emitting
statistics.

This will be used for now by connections from the stream plugin.

The existing rabbit_connections process group cannot be re-used in this
case as it is used by the list_connections CLI command, and we do not
want non-AMQP connections to show up in the output.

### DIFF
--- a/src/rabbit.erl
+++ b/src/rabbit.erl
@@ -1073,7 +1073,7 @@ force_event_refresh(Ref) ->
     ok = rabbit_networking:force_connection_event_refresh(Ref),
     % "external" connections, which are not handled by the "AMQP core",
     % e.g. connections to the stream plugin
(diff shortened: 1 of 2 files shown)


src/rabbit.erl
```diff
@@ -1069,6 +1069,7 @@ config_locations() ->
 force_event_refresh(Ref) ->
     ok = rabbit_direct:force_event_refresh(Ref),
     ok = rabbit_networking:force_connection_event_refresh(Ref),
+    ok = rabbit_networking:force_external_connection_event_refresh(Ref),
```
michaelklishin, line 1072: should we explain why we have two of these with a comment?

PR #2482, opened 2020-10-23
Fix concurrent startup of the stream coordinator

dcorbacho:
The coordinator can return an error (unavailable) when we try to start it in parallel, as `ra:start_cluster` detects some nodes up when it thinks it should be fully down and shuts down the cluster. 

This happens because of the lazy initializacion of the coordinator, only on queue declare. We had implemented it initially as a boot step which is now commented out: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit.erl#L124 The implementation is still there and starting the coordinator on node startup would also avoid this issue. My guess is that we decided to not start the coordinator until a declare happens, which means that the feature flag is enabled. But I really can't remember why, and with the single stream queues commit we lost all the commit history. Blame no longer blames! Should we remove the boot step forever? @kjnilsson @dumbbell ?

## Types of Changes

## Checklist

### DIFF
--- a/src/rabbit_stream_coordinator.erl
+++ b/src/rabbit_stream_coordinator.erl
@@ -115,7 +115,9 @@ delete_replica(StreamId, Node) ->
     process_command({delete_replica, #{stream_id => StreamId, node => Node}}).
 
 process_command(Cmd) ->
+    global:set_lock(?STREAM_COORDINATOR_STARTUP),
     Servers = ensure_coordinator_started(),
+    global:del_lock(?STREAM_COORDINATOR_STARTUP),
     process_command(Servers, Cmd).
 
 process_command([], _Cmd) ->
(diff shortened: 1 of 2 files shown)


acogoluegnes (approved):

src/rabbit_stream_coordinator.erl
```diff
@@ -115,7 +115,9 @@ delete_replica(StreamId, Node) ->
     process_command({delete_replica, #{stream_id => StreamId, node => Node}}).
 
 process_command(Cmd) ->
+    global:set_lock(?STREAM_COORDINATOR_STARTUP),
```
dumbbell, line 118: I don't know what the stream coordinator is, but do we need to acquire the lock on all connected nodes (as opposed to the local node only)?

dcorbacho: We do, because we could end up starting the whole cluster here. It will only affect declare/delete operations which if stream queues are used properly won't have a high churn

PR #2487, opened 2020-10-29
Make it possible to extract username from certificate SANs

michaelklishin:
This makes it possible to use a subject alternative name of a certain type and position (since there can be more than one value available) like so:

``` ini
ssl_cert_login_from      = subject_alternative_name
ssl_cert_login_san_type  = email
ssl_cert_login_san_index = 0
```

Supported SAN value types currently are `dns`, `email,` `uri` and `other_name`. This should be a sufficient set
for starters.

Part of rabbitmq/rabbitmq-auth-mechanism-ssl#12.


michaelklishin:
Backported to `v3.8.x`.

PR #2491, opened 2020-11-02
Fix unknown type specs

kjnilsson:


src/rabbit_maintenance.erl
```diff
@@ -38,6 +38,7 @@
 -define(DRAINING_STATUS, draining).
 
 -type maintenance_status() :: ?DEFAULT_STATUS | ?DRAINING_STATUS.
+-type mnesia_table() :: atom().
```
michaelklishin, line 41: Should we rename this to `schema_table_name/0`?

michaelklishin: As discussed on Slack, we will use this name for now.

michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #2504, opened 2020-11-11
Use correct credit mode x-credit

kjnilsson:
When the x-credit consumer arg is defined Quorum Queues should use use
credit mode `credited` and not `simple_prefetch`.


dcorbacho (approved):

pjk25:
@kjnilsson I've rebased the branch to resolve the conflict

PR #2515, opened 2020-11-15
Allow "all" as a feature flag name

mkuratczyk:
`rabbitmqctl enable_feature_flag all` enables all currently disabled feature flags.

While https://github.com/rabbitmq/rabbitmq-cli/issues/455 asked for `--all`, I decided to just use "all" for consistency with `rabbitmq-queues rebalance all`.

Closes rabbitmq/rabbitmq-cli#455


pjk25:
I noticed that this didn't seem to run the full suite of tests, even though I would have expected it to. I will double check the GitHub workflows.

pjk25:
To follow up on the GitHub Actions - they are not enabled for PRs currently, so things are behaving correctly. I don't expect to enable them anytime soon, as we would need to ensure that a PR could not accidentally leak any credentials.

michaelklishin:
Merged manually.

michaelklishin:
Backported to `v3.8.x`.

PR #2638, opened 2020-11-17
Confirm on ack info item to display in list_queues

dcorbacho:
Moved from https://github.com/rabbitmq/rabbitmq-cli/pull/359
Author: @hairyhum 

Requires rabbitmq/rabbitmq-server#2017


michaelklishin:
I could not find any evidence that such info item (metric) exists for queues. If this is suggested again, we can dig deeper.

PR #2650, opened 2020-11-19
Introduce stream management plugin

acogoluegnes:


acogoluegnes:
To accept:

* Run the management plugin on the branch:

```shell
cd /tmp
git clone git@github.com:rabbitmq/rabbitmq-server.git
cd rabbitmq-server
git checkout rabbitmq-stream-management
make run-broker PLUGINS='rabbitmq_stream_management'
```

* Go to http://localhost:15672 and authenticate. The home page should show up, there should be a new `Stream` tab.
* Click on the `Stream` tab. The stream connections page shows up, it should be empty.
* Use the stream perf test tool to create some connections

```shell
cd /tmp
git clone git@github.com:rabbitmq/rabbitmq-stream-java-client.git
cd rabbitmq-stream-java-client
./mvnw clean package -Dmaven.test.skip -P performance-tool
java -jar target/stream-perf-test.jar --rate 10000 --streams stream1,stream2 -x 2 -y 2
```

* Go back to the stream connections page, the connections should show up.
* Click on a `locator` connection. Check the details, there should be no publishers and no consumers.
* Go back to the stream connections page, click on a `publisher` connection. Check the details, there should be one publisher and no consumer.
* Go back to the stream connections page, click on a `consumer` connection. Check the details, there should be one consumer and no publisher.
* Click on the `Connections` tab. The stream connections should show up as well. The `Protocol` column should say `stream` for each of them.
* Go to the details page for different connections. The details page should show no data for AMQP-specific things (e.g. channels).

dcorbacho:
This can be merged as is, but I think we should report the consumers to the core metrics so we have details of them on the queue page. Otherwise, it's really difficult to figure out if there is any consuming happening on the stream queues.

dimonzozo:
Hi. Is https://github.com/rabbitmq/rabbitmq-stream-management correct path to repository? It is missing now and it fails resolving dependencies while building rabbitmq plugins.

acogoluegnes:
After the switch to monorepo, all plugins are included in the `rabbitmq-server` repository, so stream management is there: https://github.com/rabbitmq/rabbitmq-server/tree/master/deps/rabbitmq_stream_management.

dimonzozo:
Makes sense. Thank you for your clarification. `make deps` works when executed inside plugin folder within rabbitmq repository. Will try to find out how to make it work in standalone plugin repository.

pjk25:
@dimonzozo If you are building your plugin with Erlang.mk, you can depend on a specific plugin within the monorepo in the following using the `git-subfolder` fetch method. For example: https://github.com/rabbitmq/rabbitmq-ct-helpers/blob/79fe83f2d4c2635f053ce8952cfadc0d63f0c3cc/Makefile#L7-L8

PR #2651, opened 2020-11-19
QQ: Emit release cursor for empty basic gets

kjnilsson:
Else an application that polled an empty quorum queue frequently using basic.get
would never result in a snapshot being taken and results in unlimited
log growth.

Possible fix for: https://github.com/rabbitmq/discussions/issues/164

@michaelklishin do you think we can get this into 3.8.10?


michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #2686, opened 2020-12-17
Expose `bypass_pem_cache` through rabbitmq.conf

mkuratczyk:
Bypassing PEM cache may speed up TLS handshakes in some cases as described
here:
https://blog.heroku.com/how-we-sped-up-sni-tls-handshakes-by-5x

Please note that Heroku's use cases with certificates provided directly to RabbitMQ is unusual. When certificates are stored on disk, PEM cache should generally improve performance but may delay certificate rotation (if the certificate changes on disk, RabbitMQ will still use the cached certificate).


michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #2695, opened 2020-12-22
Update retention when only stream retention policy has changed

dcorbacho:
In any other case, the worker needs to be restarted

## Types of Changes

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_stream_coordinator.erl
+++ b/deps/rabbit/src/rabbit_stream_coordinator.erl
@@ -185,8 +185,15 @@ apply(#{from := From}, {policy_changed, #{stream_id := StreamId}} = Cmd,
                 Conf ->
                     %% No changes, ensure we only trigger an election if it's a must
                     {State, ok, []};
-                _ ->
-                    {State, ok, [{mod_call, osiris_writer, stop, [Conf]}]}
+                #{retention := Retention,
+                  leader_pid := Pid} = Conf0 ->
+                    case maps:remove(retention, Conf) == maps:remove(retention, Conf0) of
+                        true ->
+                            %% Only retention policy has changed, it doesn't need a full restart
+                            {State, ok, [{mod_call, osiris, update_retention, [Pid, Retention]}]};
+                        false ->
+                            {State, ok, [{mod_call, osiris_writer, stop, [Conf]}]}
+                    end
             end;
         SState0 ->
             Streams = maps:put(StreamId, add_pending_cmd(From, Cmd, SState0), Streams0),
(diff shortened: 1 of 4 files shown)


deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -185,8 +185,15 @@ apply(#{from := From}, {policy_changed, #{stream_id := StreamId}} = Cmd,
...
                     %% No changes, ensure we only trigger an election if it's a must
                     {State, ok, []};
-                _ ->
-                    {State, ok, [{mod_call, osiris_writer, stop, [Conf]}]}
+                #{retention := Retention,
+                  leader_pid := Pid} = Conf0 ->
+                    case maps:remove(retention, Conf) == maps:remove(retention, Conf0) of
+                        true ->
+                            %% Only retention policy has changed, it doesn't need a full restart
+                            {State, ok, [{mod_call, osiris, update_retention, [Pid, Retention]}]};
```
kjnilsson, line 193: This needs to be done for each member, also this is a call so should be done in a separate process with failure handling.

kjnilsson (changes requested):
Each member needs to have their retention changed also we need to handle failures so start a new worflow effectively.

deps/rabbit/src/rabbit_stream_coordinator.erl
dcorbacho: Done

deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -253,6 +271,23 @@ apply(_Meta, {start_replica_failed, StreamId, Node, Retries, Reply},
...
+      #?MODULE{streams = Streams0} = State) ->
+    rabbit_log:debug("rabbit_stream_coordinator: ~p update retention failed", [StreamId]),
+    case maps:get(StreamId, Streams0, undefined) of
+        undefined ->
+            {State, {error, not_found}, []};
+        #{reply_to := From} = SState  ->
+            Streams = Streams0#{StreamId => clear_stream_state(SState)},
+            reply_and_run_pending(
+              From, StreamId, ok, Reply,
+              [{timer, {pipeline,
```
kjnilsson, line 284: I'm not sure why it is wrapped in `{pipeline, []}`? what does this do?

kjnilsson (changes requested):
xref fails but apart from that it looks good to me.

deps/rabbit/src/rabbit_stream_coordinator.erl
dcorbacho: It adds the command to the pipeline command, but delayed instead of immediately. I guess we don't want to retry immediately?

PR #2697, opened 2020-12-23
Check that the queue exists before calling declare

dcorbacho:
Rabbit channels are responsible of this check before calling declare,
skipping it on the manager meant that the queue was partly redeclared
and a new data directory created. The old one was still on disk with
a different timestamp, but from the user point of view the queue data
has been erased.

@acogoluegnes This is a bug found by @Gsantomaggio, where by re-declaring a queue the messages are apparently purged despite receiving an `STREAM_ALREADY_EXISTS` error. Feel free to cherry-pick it to your branch but might need a review first. I wonder why declare is not idempotent when using the stream protocol?

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_manager.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_manager.erl
@@ -119,20 +119,31 @@ handle_call({create, VirtualHost, Reference, Arguments, Username}, _From, State)
                 none, true, false, none, StreamQueueArguments,
                 VirtualHost, #{user => Username}, rabbit_stream_queue
             ),
-            try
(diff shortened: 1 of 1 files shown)


michaelklishin (approved):

acogoluegnes (approved):

acogoluegnes:
I don't have strong opinion WRT idempotence for stream creation. Resource creation is idempotent in AMQP, but not in other stream protocols AFAIK, so it's rather a matter of staying close to AMQP or not. WDYT @dcorbacho @kjnilsson @Gsantomaggio ?

Gsantomaggio:
I think the RabbitMQ users expect for the same AMQP behaviour.

PR #2707, opened 2021-01-07
Remove Ranch dependency from rabbit_common

michaelklishin:
It is not actually used there and may complicate dependency resolution
for RabbitMQ Erlang client users.

Per discussion with @dumbbell

### DIFF
--- a/deps/rabbit_common/Makefile
+++ b/deps/rabbit_common/Makefile
@@ -16,7 +16,7 @@ define PROJECT_APP_EXTRA_KEYS
 endef
 
 LOCAL_DEPS = compiler crypto public_key sasl ssl syntax_tools tools xmerl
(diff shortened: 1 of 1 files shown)


dumbbell (changes requested):
Can we just keep the commented out specs? They were commentted out with at least another one, unrelated to Ranch. If the problem mentionned in the commit commenting them out is fixed one day, we can restore them (or move them to `deps/rabbit`).

Other than that, I agree with the patch and welcome the removal of this dependency.

michaelklishin:
@dumbbell pushed a new commit which keeps the commented out specs.

dumbbell (approved):
Thank you, looks good to me!

michaelklishin:
Backported to `v3.8.x`.

PR #2712, opened 2021-01-11
Add stream prometheus plugin

acogoluegnes:


deps/rabbitmq_stream_prometheus/src/collectors/prometheus_rabbitmq_stream_collector.erl
```diff
@@ -0,0 +1,162 @@
...
+     || {Table, Contents} <- ?METRICS_RAW],
+    ok.
+
+get_data(?TABLE_PUBLISHER = Table, false) ->
+    {Table, A1, A2, A3} =
+        ets:foldl(fun({_, Props}, {T, A1, A2, A3}) ->
+                     {T,
+                      sum(proplists:get_value(published, Props), A1),
+                      sum(proplists:get_value(confirmed, Props), A2),
+                      sum(proplists:get_value(errorred, Props), A3)}
```
kjnilsson, line 85: should it be `errored`?

acogoluegnes: Yes :)

acogoluegnes:
To accept:

 * Run the stream prometheus plugin on the branch:

```
cd /tmp
git clone git@github.com:rabbitmq/rabbitmq-server.git
cd rabbitmq-server
git checkout rabbitmq-stream-prometheus
make run-broker PLUGINS='rabbitmq_stream_prometheus'
```
 * go to http://localhost:15692/metrics/
 * look for the `rabbitmq_stream_` metrics, they should all be set to 0
 * go to http://localhost:15692/metrics/per-object
 * look for the `rabbitmq_stream_` metrics, they should not have any values (no object yet)
 * use the performance tool to create some publishers and consumers:

```
docker run -it --rm --network host pivotalrabbitmq/stream-perf-test --rate 10000 --streams stream1,stream2 -x 2 -y 2
```
 * refresh http://localhost:15692/metrics/
 * the `rabbitmq_stream_` metrics should show 2 publishers and consumers, with their respective message counts (published, confirmed, consumed)
 * refresh http://localhost:15692/metrics/per-object
 * there should be no metrics for publisher and consumer counts, the `rabbitmq_stream_` metrics should show the number of messages by publishers and consumers
 * stop the performance tool
 * the stream metrics on http://localhost:15692/metrics/ and http://localhost:15692/metrics/per-object should go back to their initial state after a refresh

PR #2713, opened 2021-01-11
Stream: Channel resend on leader change

kjnilsson:
Detect when a new stream leader is elected and make stream_queues
re-send any unconfirmed, pending messages to ensure they did not get
lost during the leader change. This is done using the osiris
deduplication feature to ensure the resend does not create duplicates of
messages in the stream.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -462,6 +462,8 @@ filter_per_type(all, _) ->
     true;
 filter_per_type(quorum, Q) ->
     ?amqqueue_is_quorum(Q);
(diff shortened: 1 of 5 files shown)


deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1792,6 +1795,7 @@ on_node_down(Node) ->
     ok.
 
 delete_queues_on_node_down(Node) ->
+    rabbit_log:info("delete_queues_on_node_down GAHHH", []),
```
michaelklishin, line 1798: We should remove this debug message before merging

michaelklishin (changes requested):

PR #2719, opened 2021-01-15
Filter Grafana dashboards by namespace

ansd:
such that RabbitMQ clusters with the same rabbitmq_cluster name in different K8s namespaces don't clash.

Relates to https://github.com/rabbitmq/cluster-operator/pull/534.

I added another variable / dropdown for the namespace.
There doesn't seem to be a way to add this variable only if a `namespace` label is found.
Therefore, if there is no `namespace` label  (e.g. when using the dashboard without running RabbitMQ on K8s), the dropdown is still there, but will show value `None`.

The `rabbitmq_cluster` variable will only show the RabbitMQ clusters for the filtered namespace in the 1st dropdown.

The regex `namespace=~\"^${namespace}$|''\"` contains empty string as second part. This makes sure that time series data is selected even if there is no `namespace` label at all. However if there is a namespace label, it must match the `$namespace` variable.
> Label matchers that match empty label values also select all time series that do not have the specific label set at all.


gerhard (commented):
I pushed a change to RabbitMQ-Oveview, all details are in the comment. Let me know how that works for you.

ansd:
@gerhard thank you, yes that's much better :) it works for me.

(I wrongly assumed that the "None" in the drop down wouldn't translate to an empty namespace value. I think I just didn't wait long enough after running `m metrics overview` for metrics to show up.)

Okay, I'll do the change for the other dashboards.

gerhard (approved):
🚢

PR #2728, opened 2021-01-20
Fix crash bug in QQ state conversion

kjnilsson:
When there are consumers in the service queue.

This needs to be added to v3.8.x asap and we'll need a new release shortly after this.


michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #2741, opened 2021-01-22
Avoid duplicate auth_attempts declarations when per-object metrics enabled

### LINKED ISSUE #2740: Prometheus: when per-object metrics are enabled, repeated TYPE declarations for the auth attempt metrics fail Telegraf Prometheus validations
With RabbitMQ 3.8.10 and 3.8.11 and per-object metrics enabled in rabbitmq.conf:

```
prometheus.tcp.port = 9419
prometheus.return_per_object_metrics = true
```

`curl http://localhost:9419/metrics` shows multiple TYPE declarations for the auth attempt metrics:

```
...
# TYPE rabbitmq_auth_attempts_total counter
# HELP rabbitmq_auth_attempts_total Total number of authorization attempts on a node
rabbitmq_auth_attempts_total{protocol="http"} 92
rabbitmq_auth_attempts_total{protocol="amqp091"} 6
...
# TYPE rabbitmq_auth_attempts_total counter
# HELP rabbitmq_auth_attempts_total Total number of authorization attempts on a node
...
```

This breaks metrics exporting via the Telegraf Prometheus plugin:

```
2021-01-22T02:25:15Z E! [inputs.prometheus] Error in plugin: error reading metrics for http://127.0.0.1:9419/metrics: reading text format failed: text format parsing error in line 461: second TYPE line for metric name "rabbitmq_auth_attempts_total", or TYPE reported after samples
```

Similar issue: https://github.com/deadtrickster/prometheus_rabbitmq_exporter/issues/19

gerhard:
Closes #2740

### DIFF
--- a/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
+++ b/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
@@ -231,7 +231,7 @@ collect(PerObjectMetrics, Callback) ->
     [begin
          Data = get_data(Table, PerObjectMetrics),
          mf(Callback, Contents, Data)
-     end || {Table, Contents} <- ?METRICS_RAW, needs_processing(PerObjectMetrics, Table)],
+     end || {Table, Contents} <- ?METRICS_RAW, include_when_per_object_metrics(PerObjectMetrics, Table)],
     [begin
          Size = ets:info(Table, size),
          mf_totals(Callback, Name, Type, Help, Size)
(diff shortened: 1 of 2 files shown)


deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
```diff
@@ -240,11 +240,11 @@ collect(PerObjectMetrics, Callback) ->
     add_metric_family(identity_info(), Callback),
     ok.
 
-needs_processing(false, auth_attempt_detailed_metrics) ->
-    %% When per object metrics are disabled the detailed authentication attempt metrics
-    %% create duplicates. Totals are carried on `auth_attempt_metrics`
+include_when_per_object_metrics(true, auth_attempt_metrics) ->
```
dcorbacho, line 243: If we exclude the non-detailed totals and the user has not set `track_auth_attempt_source` to `true` it will end up without metrics. This is a misconfiguration issue, but it makes me wonder if we should always report the globals anyway. We only need to use different keys for the values (such as `auth_attempts_detailed_total`), and hide them when not required. @michaelklishin @gerhard thoughts?

gerhard: I like that idea very much! Having a go at it now.

dcorbacho:
> Is that what you had in mind @dcorbacho?

Yes!

dcorbacho (approved):

michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

Issue #2749, opened 2021-01-25
Classic mirrored queue leadership transfer eagerly clears the list of mirrors which can result in data loss


ChunyiLyu:
## Describe the bug

Data loss issue with mirror queues on operator deployed clusters. Reported by internal team. cc @mkuratczyk @Gsantomaggio 

## To Reproduce
1. Deploy 3 nodes rabbitmq cluster
2. Create HA policy with params: "ha-mode: exactly, ha-params: 2, ha-promote-on-failure: when-synced, ha-sync-batch-size: 5000, ha-sync-mode: automatic"
3. Use perf tests to populate the cluster with unconsumed messages `-y0 -queue-pattern 'perf-test-%d' \
--auto-delete false --queue-pattern-from 1 --queue-pattern-to 10  -s 1000 -x 10 -y 0 -r 1000 -R 0  -f persistent`
4. Stop perf-tests (without triggering any alarms and all queues should be synced)
5. Delete one node which has queue masters located on and observe data loss (0 message left)

Include any YAML or manifest necessary to reproduce the problem.

**Expected behavior**

After deleting one rabbitmq node, the mirror queue should be promoted as the master queue and rabbitmq should create a new mirror in node that's not deleted.

## Screenshots

![data loss](https://user-images.githubusercontent.com/6786193/105691342-080cff00-5ef5-11eb-85ac-32e4cec9d0e5.png)

## Version and environment information
- RabbitMQ: 3.8.9
- RabbitMQ Cluster Operator: Latest main: https://github.com/rabbitmq/cluster-operator/commit/c4206617eaac1a895936e9f805ae7586cc3c49f0
- Kubernetes: 1.17.0
- Cloud provider or hardware configuration:
 
## Additional context

This is a reported issue. The user has also seem similar data loss with "ha-params: 3" on 5 nodes cluster.

mkuratczyk:
The issue is in the implementation of the maintenance mode (`rabbitmq-upgrade drain` command). For classic mirrored queues it selects a new leader randomly from the pool of all other cluster nodes and then deletes existing mirrors. Since the synchronization is async, if a node with no replica is picked, existing mirrors can get deleted before synchronization completes. We are discussing potential solutions, such as:
1. explicitly waiting for the synchronization to finish
2. selecting the new leader from the pool of nodes that already have replicas

[...]

PR #2755, opened 2021-01-27
Don't perform CMQ leadership transfer when entering maintenance mode

michaelklishin:
The time this operation can take in clusters with a lot of classic
mirrored queue (say, 10s or 100s of thousands) be prohibitive for
upgrades.

Upgrades that use a health check to ensure that there are in-sync
replicas before entering maintenance mode, in which case
the transfer is not really necessary.

All of the above is more obvious with the recent changes in #2749.


michaelklishin:
Backported to `v3.8.x`.

PR #2765, opened 2021-01-28
Remove a log line related to CMQ leader transfer

mkuratczyk:
We've removed this functionality in https://github.com/rabbitmq/rabbitmq-server/commit/c7b9c393527fa69633383b856700a6d151c27e3d
so we shouldn't log that we would do that.


michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #2777, opened 2021-02-01
Add OTP version to the startup banner

mkuratczyk:
Print Erlang/OTP version after the RabbitMQ version in the startup banner. This is purely for information purposes - often users will copy-paste console output when reporting issues. This way we don't have ask about the Erlang version.

```
RabbitMQ 3.8.10+18.gea1f4a3.dirty, Erlang/OTP 23.2.3
```

### DIFF
--- a/deps/rabbit/src/rabbit.erl
+++ b/deps/rabbit/src/rabbit.erl
@@ -858,8 +858,9 @@ start(normal, []) ->
                    #{domain => ?RMQLOG_DOMAIN_PRELAUNCH});
             _ ->
                 ?LOG_INFO(
-                   "~n Starting ~s ~s on Erlang ~s~n ~s~n ~s",
+                   "~n Starting ~s ~s on Erlang ~s [~s]~n ~s~n ~s",
                    [product_name(), product_version(), rabbit_misc:otp_release(),
+                    emu_flavor(),
                     ?COPYRIGHT_MESSAGE, ?INFORMATION_MESSAGE],
                    #{domain => ?RMQLOG_DOMAIN_PRELAUNCH})
         end,
@@ -1183,18 +1184,33 @@ print_banner() ->
     io:format(Logo ++
               "~n" ++
               MOTDFormat ++
-              "~n  Doc guides: https://rabbitmq.com/documentation.html"
-              "~n  Support:    https://rabbitmq.com/contact.html"
-              "~n  Tutorials:  https://rabbitmq.com/getstarted.html"
-              "~n  Monitoring: https://rabbitmq.com/monitoring.html"
+              "~n  Erlang:      ~ts [~ts]"
+              "~n  SSL Library: ~ts"
+              "~n"
+              "~n  Doc guides:  https://rabbitmq.com/documentation.html"
+              "~n  Support:     https://rabbitmq.com/contact.html"
+              "~n  Tutorials:   https://rabbitmq.com/getstarted.html"
+              "~n  Monitoring:  https://rabbitmq.com/monitoring.html"
               "~n"
               "~n  Logs: ~ts" ++ LogFmt ++ "~n"
               "~n  Config file(s): ~ts" ++ CfgFmt ++ "~n"
               "~n  Starting broker...",
               [Product, Version, ?COPYRIGHT_MESSAGE, ?INFORMATION_MESSAGE] ++
+              [rabbit_misc:otp_release(), emu_flavor(), crypto_version()] ++
               MOTDArgs ++
               LogLocations ++
               CfgLocations).
+
+emu_flavor() ->
+    %% emu_flavor was introduced in Erlang 24 so we need to catch the error on Erlang 23
+    case catch(erlang:system_info(emu_flavor)) of
+        {'EXIT', _} -> "emu";
+        EmuFlavor -> EmuFlavor
+    end.
+
+crypto_version() ->
+    [{CryptoLibName, _, CryptoLibVersion}] = crypto:info_lib(),
+    [CryptoLibName, " - ", CryptoLibVersion].
 
 log_motd() ->
     case motd() of


deps/rabbit/src/rabbit.erl
```diff
@@ -1123,7 +1123,7 @@ log_broker_started(Plugins) ->
 
 print_banner() ->
     Product = product_name(),
-    Version = product_version(),
+    Version = [product_version(), ", ", rabbit_misc:platform_and_version()],
```
gerhard, line 1126: This line is specific to the product version, and adding OTP version here does not make sense.

Maybe that's why we have the Erlang version printed a few lines above, e.g.:

```
 Starting RabbitMQ 3.8.10+8.g5247909.25+stream on Erlang 23.2.3
 Copyright (c) 2007-2021 VMware, Inc. or its affiliates.
 Licensed under the MPL 2.0. Website: https://rabbitmq.com

  ##  ##      RabbitMQ 3.8.10+8.g5247909.25+stream
  ##  ##
  ##########  Copyright (c) 2007-2021 VMware, Inc. or its affiliates.
  ######  ##
  ##########  Licensed under the MPL 2.0. Website: https://rabbitmq.com
```

Is the Erlang version listed a few lines above not enough? Do users start copying from the rabbit ASCII art down?

The problem with this change is that we have the Copyright notice 2 lines below, so adding the Erlang version here, then the VMware copyright doesn't seem right. By that logic, listing the Erlang version a few lines above, then the copyright, isn't right either.

Why don't we list the Erlang version, OS & the OpenSSL versions below the rabbit ASCII, between `Licensed ...` & `Doc guides...`, like so:

```
 Starting RabbitMQ 3.8.10+8.g5247909.25+stream on Erlang 23.2.3
 Copyright (c) 2007-2021 VMware, Inc. or its affiliates.
 Licensed under the MPL 2.0. Website: https://rabbitmq.com

  ##  ##      RabbitMQ 3.8.10+8.g5247909.25+stream
  ##  ##
  ##########  Copyright (c) 2007-2021 VMware, Inc. or its affiliates.
  ######  ##
  ##########  Licensed under the MPL 2.0. Website: https://rabbitmq.com

  Erlang 23.2.3
  Ubuntu 20.04.1 x86_64
  OpenSSL 1.1.1f

  Doc guides: https://rabbitmq.com/documentation.html
  Support:    https://rabbitmq.com/contact.html
  Tutorials:  https://rabbitmq.com/getstarted.html
  Monitoring: https://rabbitmq.com/monitoring.html
```

michaelklishin: We can do that. We should be defensive if OpenSSL version cannot be inferred, e.g. Erlang was built without OpenSSL support.

gerhard: That sounds like a reasonable compromise.

I would even be OK to not include the platform architecture if it's difficult to get it. I think we get some (if not all) of the OS details in `rabbitmqctl status`.

Part of this, I would like to see us to drop the `on Erlang 23.2.3` part from the `Starting` line. We can rename the PR to **Add useful environment details to the Startup banner**.

gerhard:
WDYT @mkuratczyk about this proposal?

```diff
 Starting RabbitMQ 3.8.10+8.g5247909.25+stream on Erlang 23.2.3
 Copyright (c) 2007-2021 VMware, Inc. or its affiliates.
 Licensed under the MPL 2.0. Website: https://rabbitmq.com

  ##  ##      RabbitMQ 3.8.10+8.g5247909.25+stream
  ##  ##
  ##########  Copyright (c) 2007-2021 VMware, Inc. or its affiliates.
  ######  ##
  ##########  Licensed under the MPL 2.0. Website: https://rabbitmq.com

+  Erlang 23.2.3
+  Ubuntu 20.04.1 x86_64
+  OpenSSL 1.1.1f

  Doc guides: https://rabbitmq.com/documentation.html
  Support:    https://rabbitmq.com/contact.html
  Tutorials:  https://rabbitmq.com/getstarted.html
  Monitoring: https://rabbitmq.com/monitoring.html
```

mkuratczyk:
I can't find a way to retrieve the operating system details but I implemented the other two lines, additionally with info whether JIT is used:

```
  Erlang: 24.0.2 [jit]
  OpenSSL: OpenSSL 1.1.1k  25 Mar 2021
```

OpenSSL details look a bit funny perhaps but I just print them verbatim as returned by `crypto:info_lib()`. I'm not sure what this will look like with LibreSSL, on Windows, etc.

gerhard:
That looks good to me! I really like the `[jit]` addition, but it fails on Erlang 23, which we need to support:

```erlang
BOOT FAILED
===========
Exception during startup:

error:badarg

    erlang:system_info/1
        args: [emu_flavor]
    rabbit:print_banner/0, line 1192
    rabbit:start/2, line 880
    application_master:start_it_old/4, line 277
```

If we decide to keep the `[jit]` (which I like!) then we should add it to the first line too:

```log
2021-06-22 14:17:02.872657+00:00 [info] <0.217.0>  Starting RabbitMQ 223ad24351928941e77c38ad5ee879b8b241730a on Erlang 24.0.2 [jit]
```

WDYT @dumbbell, assuming that you also think it would be a good idea to do this, how can we retrieve the OS name, version & platform?

gerhard:
By the way, you may want to rebase this onto `master` to fix the 23 & 24 tests.

mkuratczyk:
It should work on Erlang 23 now and the flavor (jit/emu) has been added to the log line. Test failures are unrelated. Give it a try and if it works - let's just do it. This PR has been open for too long given it's scope. ;)

gerhard:
You are right @mkuratczyk

While we could have continued with the PR 🏓  and made suggestions, we took the quick route. After tests pass, this is good to merge. Thanks for sticking with this!

image

image

gerhard:
Backported to `v3.9.x`

gerhard:
[Backported to v3.8.x](https://github.com/rabbitmq/rabbitmq-server/commit/441bcebd288016b50ac2a13f37cbd12b83723a05)

PR #2791, opened 2021-02-03
Support fetching JWT signing keys from JWKS endpoint

teozkr:
## Proposed Changes

JWKS allows the JWT token issuer to rotate its keys without having to (know about and) notify all token consumers.

With this change, if `rabbitmq_auth_backend_oauth2.key_config.jwks_url` is set, the key set will be updated from JWKS whenever validating a token with a key ID that is not already known about.

Previously proposed at https://groups.google.com/g/rabbitmq-users/c/3r2csfp7LMI

## Types of Changes

## Checklist


pivotal-issuemaster:
@teozkr Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-server&pullRequestId=2791)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-server&pullRequestId=2791) to manually synchronize the status of this Pull Request.

deps/rabbitmq_auth_backend_oauth2/src/uaa_jwt.erl
```diff
@@ -51,6 +51,23 @@ update_uaa_jwt_signing_keys(UaaEnv0, SigningKeys) ->
...
+-spec update_jwks_signing_keys() -> ok | {error, term()}.
+update_jwks_signing_keys() ->
+    UaaEnv = application:get_env(?APP, key_config, []),
+    case proplists:get_value(jwks_url, UaaEnv) of
+        undefined ->
+            {error, no_jwks_url};
+        JwksUrl ->
+            case httpc:request(JwksUrl) of
+                {ok, {_, _, JwksBody}} ->
+                    KeyList = maps:get(<<"keys">>, jose:decode(erlang:iolist_to_binary(JwksBody)), []),
```
michaelklishin, line 63: Is this response format standard? Where is it documented?

deps/rabbitmq_auth_backend_oauth2/src/uaa_jwt.erl
```diff
@@ -51,6 +51,23 @@ update_uaa_jwt_signing_keys(UaaEnv0, SigningKeys) ->
...
+                    Keys = maps:from_list(lists:map(fun(Key) -> {maps:get(<<"kid">>, Key, undefined), {json, Key}} end, KeyList)),
+                    update_uaa_jwt_signing_keys(UaaEnv, Keys);
```
teozkr, line 65: Note: This silently overwrites `signing_keys` if the administrator specified both that and `jwks_url`.. Is there a good place to warn on startup in that case?

We can't just check whether `signing_keys` is already set, since that would break refreshing.

deps/rabbitmq_auth_backend_oauth2/src/uaa_jwt.erl
```diff
@@ -67,21 +84,33 @@ decode_and_verify(Token) ->
...
+    Keys = signing_keys(),
+    case maps:get(KeyId, Keys, undefined) of
+        undefined ->
+            if
+                AllowUpdateJwks ->
+                    case update_jwks_signing_keys() of
+                        ok ->
+                            get_jwk(KeyId, false);
+                        {error, no_jwks_url} ->
+                            {error, key_not_found};
```
teozkr, line 99: `no_jwks_url` indicates that the user isn't interested in JWKS, so `key_not_found` is probably a more helpful error message..

michaelklishin (commented):
Thank you. We would like to test this proposed feature before we agree to merge it. It also must be documented.

I don't see any details as to how this can be verified to work, if this token refresh over HTTP approach is standard, what tools support it, either in this PR or on the mailing list.

Would you mind providing some steps that we can use to verify? How about putting together a small test using a local HTTP server? [`rabbitmq_trust_store` does something similar](https://github.com/rabbitmq/rabbitmq-server/tree/master/deps/rabbitmq_trust_store) for its HTTP-based certificate source.

deps/rabbitmq_auth_backend_oauth2/src/uaa_jwt.erl
```diff
@@ -67,21 +84,33 @@ decode_and_verify(Token) ->
 
 -spec get_jwk(binary()) -> {ok, map()} | {error, term()}.
 get_jwk(KeyId) ->
-    case signing_keys() of
-      undefined -> {error, signing_keys_not_configured};
```
teozkr, line 71: Wasn't really sure about this case.. it would have meant duplicating the JWKS code updating code "just" to preserve the error message for a very similar case.

deps/rabbitmq_auth_backend_oauth2/src/uaa_jwt.erl
teozkr: Yes, it's part of the JWK RFC: https://tools.ietf.org/html/rfc7517#section-5

teozkr: This is also part of the OpenID Connect spec, including the refresh mechanism: https://openid.net/specs/openid-connect-core-1_0.html#RotateSigKeys

teozkr:
@michaelklishin

> Would you mind providing some steps that we can use to verify? How about putting together a small test using a local HTTP server? rabbitmq_trust_store does something similar for its HTTP-based certificate source.

Sure, I'll hopefully get to that in one of the coming days.

pivotal-issuemaster:
@teozkr Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-server&pullRequestId=2791)!

deps/rabbitmq_auth_backend_oauth2/src/uaa_jwt.erl
michaelklishin: Plugins can define boot steps and they can perform configuration validation. Here's one basic boot step (that has nothing to do with validation): https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit/src/rabbit.erl#L251. We can add it on your behalf later.

michaelklishin:
Sounds great, thank you!

deps/rabbitmq_auth_backend_oauth2/src/jwks_http_app.erl
```diff
@@ -0,0 +1,22 @@
+-module(jwks_http_app).
```
teozkr, line 1: Doesn't feel great to pollute the primary namespace with this.. should it be in the tests folder instead?

deps/rabbitmq_auth_backend_oauth2/ebin/jwks_http.app
```diff
@@ -0,0 +1,9 @@
+{application, 'jwks_http', [
```
teozkr, line 1: Seems like these are generally generated by erlang.mk, and `make clean` tramples all over this... I assume there's a better way to do this?

deps/rabbitmq_auth_backend_oauth2/test/jwks_SUITE.erl
```diff
@@ -0,0 +1,390 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
```
teozkr, line 1: Not really sure that it's great to copy/paste all of `system_SUITE`, but it was one way to avoid interfering with the common broker (which could break other tests, or at the very least change their meaning significantly).

deps/rabbitmq_auth_backend_oauth2/src/jwks_http_app.erl
michaelklishin: You should be able to, e.g. [this suite uses a Common Test data directory](https://github.com/rabbitmq/rabbitmq-server/tree/master/deps/rabbit/test/unit_config_value_encryption_SUITE_data) (but it loads a file app descriptor directly).

`test` should be on the source code path. One complication here is that the node that runs the tests — a Common Test runner node — is separate from RabbitMQ nodes that run the plugins.

deps/rabbitmq_auth_backend_oauth2/ebin/jwks_http.app
michaelklishin: I'd try starting a Cowboy server with some routes right on the Common Test runner node. We just need a local HTTP endpoint, right? It doesn't really have to be a separate Erlang node?

deps/rabbitmq_auth_backend_oauth2/test/jwks_SUITE.erl
michaelklishin: Having a separate suite is perfectly fine and what I would do 👍 . Test suites generally start nodes that are not shared by any other suites. The same can be done for individual groups of tests or even individual tests. See the [main quorum queue suite](https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit/test/quorum_queue_SUITE.erl), it has quite a few extensive examples.

This way you do not affect other suites (groups, tests).

deps/rabbitmq_auth_backend_oauth2/src/jwks_http_app.erl
teozkr: That should be fine, as long as they're in the same network namespace.

deps/rabbitmq_auth_backend_oauth2/ebin/jwks_http.app
teozkr: Yep, that's correct. I mostly did that because I was following the `trust_store` environment, which seemed to tie cowboy's lifetime into the `app`'s, I think? Guess we need to shut down Cowboy some other way otherwise.

deps/rabbitmq_auth_backend_oauth2/test/jwks_SUITE.erl
teozkr: Alright, closing this thread then.

deps/rabbitmq_auth_backend_oauth2/src/uaa_jwt.erl
michaelklishin: Sounds reasonable.

michaelklishin (changes requested):
The tests fail for me on this branch:

```
jwks_SUITE
    {'EXIT',{{badmatch,{error,{"no such file or directory","jwks_http.app"}}},
             [{jwks_SUITE,start_jwks_server,1,
                          [{file,"test/jwks_SUITE.erl"},{line,146}]},
                                 [{file,"src/rabbit_ct_helpers.erl"},
                                  {line,113}]},
              {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1754}]},
                           [{file,"test_server.erl"},{line,1351}]},
                           [{file,"test_server.erl"},{line,1195}]}]}}
```

(I run `gmake tests`)

It would be great to move these test modules to the test directory. It's confusing when they are under `src`.

teozkr:
> It would be great to move these test modules to the test directory. It's confusing when they are under src.

Done.

> The tests fail for me on this branch:

Yeah, this is https://github.com/rabbitmq/rabbitmq-server/pull/2791#discussion_r570889684: the `jwks_http.app` gets deleted by `make`. As a workaround, you can run `make && git restore ebin/jwks_http.app && make tests`.

deps/rabbitmq_auth_backend_oauth2/ebin/jwks_http.app
teozkr: 543e8aa does away with `jwks_http.app`.

deps/rabbitmq_auth_backend_oauth2/test/jwks_SUITE.erl
```diff
@@ -0,0 +1,389 @@
...
+                                      [rabbitmq_auth_backend_oauth2, resource_server_id, ?RESOURCE_SERVER_ID]),
+    Config.
+
+start_jwks_server(Config) ->
+    Jwk   = ?UTIL_MOD:fixture_jwk(),
+    %% Assume we don't have more than 100 ports allocated for tests
+    PortBase = rabbit_ct_broker_helpers:get_node_config(Config, 0, tcp_ports_base),
+    JwksServerPort = PortBase + 100,
+    ok = application:set_env(jwks_http, keys, [Jwk]),
+    {ok, _} = application:ensure_all_started(cowboy),
```
teozkr, line 147: This leaks the cowboy application, but that seems like a better option than forcibly shutting down all cowboys at the end, regardless of whether they were already running since before?

michaelklishin: Starting and not shutting down a Cowboy listener for the duration of this suite sounds acceptable.

teozkr: Alright.

michaelklishin (approved):

michaelklishin:
Thank you!

michaelklishin:
Backported to `v3.8.x`.

teozkr:
Thanks!

copiltembel:
Is it possible to specify multiple JWKS endpoints? If yes, how? 

The issue mentions `rabbitmq_auth_backend_oauth2.key_config.jwks_url` which I guess is only one.

copiltembel:
@teozkr Thank you for the very thorough answer! 

I'll implement the merging gateway for my project.

PR #2792, opened 2021-02-03
Bump gen_server:call timeuts for rabbitmq_auth_backend_cache operations

michaelklishin:
as we have seen with channel operation timeouts, 5s is too low
under load spikes.


lukebakken (approved):
Auth cache test suite passes, `rabbit` test suite is mid-run. Everything's looking good.

michaelklishin:
I tested this in combination with `auth_backend_http` with Spring and Django. Found no surprises.

michaelklishin:
Backported to `v3.8.x`.

PR #2797, opened 2021-02-08
Don't consider exclusive classic queues to be mirrored

### LINKED ISSUE #2795: Exclusive classic queues should not be rebalancing candidates
Currently `rabbitmq-queues rebalance` will consider exclusive queues that are matched by a classic queue mirroring policy
because the criteria of "replicated" only relies on policy matching. Exclusive queue should be excluded from rebalancing as they are not mirrored.

michaelklishin:
Because they are not, and should not be considered for operations
or features that assume mirroring, such as queue rebalacing.

Closes #2795.


michaelklishin:
Per verbal review from @dcorbacho

michaelklishin:
Backported to `v3.8.x`.

PR #2802, opened 2021-02-10
Allow dynamic Shovels specify queue x-arguments when topology is declared

### LINKED ISSUE #2798: Provide a way to specify queue type for dynamic Shovel destination
Currently there is no way to specify queue type for dynamic Shovel destinations (`dest-queue`), so they will be declared as classic queues unless already exist.

michaelklishin:
This makes dynamic shovels recognize two new options, `"src-queue-args"` and `"dest-queue-args"`,  respectively.
They are passed for [optional queue arguments](https://www.rabbitmq.com/queues.html#optional-arguments) and should be used for the arguments that cannot be set using a policy, such as the type of the queue.
For everything else, [policies is the right choice](https://www.rabbitmq.com/parameters.html#policies).

This way Shovels can be instructed to declare quorum queues, e.g. on the destination side.

Closes #2798.


michaelklishin:
Backported to `v3.8.x`.

Issue #2756, opened 2021-01-27
Queue decorators are not supported by quorum queues


michaelklishin:
Queue federation uses decorators to register callbacks that react to queue policy changes, so that links can be started/stopped. This is currently not supported by quorum queues, at least not in every case.

For example, `rabbit_amqqueue:notify_decorators/1` will fail with an exception if invoked for a quorum queue:

```
{function_clause,
                           [{delegate,invoke_no_result,
                                [{'%2F_test.federation.orange',
                                     'rabbit@rabbitmq-1.rabbitmq-headless.rabbitmq.svc.cluster.local'},
                                 {gen_server2,cast,[notify_decorators]}],
                                [{file,"src/delegate.erl"},{line,137}]},
                            {rabbit_federation_queue_link,init,1,
                                [{file,"src/rabbit_federation_queue_link.erl"},
                                 {line,72}]},
                            {gen_server2,init_it,6,
                                [{file,"src/gen_server2.erl"},{line,565}]},
                            {proc_lib,init_p_do_apply,3,
                                [{file,"proc_lib.erl"},{line,249}]}]}}},
```

As you can see above, a Ra server ID is passed to `delegate:invoke_no_result/2` which expects an Erlang PID instead.

There are a few more places with basically the same problem.

michaelklishin:
#2772 is a band aid we can ship very soon while we are working on QQ federation support.

michaelklishin:
Backported to `v3.8.x` in https://github.com/rabbitmq/rabbitmq-server/pull/2831.

PR #2804, opened 2021-02-10
Add federation support for quorum queues

dcorbacho:
Adds decorator support for quorum queues, in order to support the federation plugin. There are no changes on the federation plugin, but the queue specific test suite uses now both classic and quorum queues.

This is part of #2756, which will be finished when this is backported to `master`. Small changes are needed to apply this patch in `master`, due to the queue type refactor.

## Types of Changes

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -1530,7 +1530,7 @@ checkout(#{index := Index} = Meta, #?MODULE{cfg = #cfg{resource = QName}} = OldS
                                             Effects0, {#{}, #{}}),
     case evaluate_limit(Index, false, OldState, State1, Effects1) of
         {State, true, Effects} ->
-            case have_active_consumers_changed(State, HandleConsumerChanges) of
+            case maybe_notify_decorators(State, HandleConsumerChanges) of
                 {true, {MaxActivePriority, IsEmpty}} ->
                     NotifyEffect = notify_decorators_effect(QName, MaxActivePriority, IsEmpty),
                     update_smallest_raft_index(Index, State, [NotifyEffect | Effects]);
@@ -1538,7 +1538,7 @@ checkout(#{index := Index} = Meta, #?MODULE{cfg = #cfg{resource = QName}} = OldS
                     update_smallest_raft_index(Index, State, Effects)
             end;
         {State, false, Effects} ->
-            case have_active_consumers_changed(State, HandleConsumerChanges) of
+            case maybe_notify_decorators(State, HandleConsumerChanges) of
                 {true, {MaxActivePriority, IsEmpty}} ->
                     NotifyEffect = notify_decorators_effect(QName, MaxActivePriority, IsEmpty),
                     {State, ok, [NotifyEffect | Effects]};
@@ -2164,9 +2164,9 @@ get_priority_from_args(#{args := Args}) ->
 get_priority_from_args(_) ->
     0.
 
-have_active_consumers_changed(_, false) ->
+maybe_notify_decorators(_, false) ->
     false;
-have_active_consumers_changed(State, _) ->
+maybe_notify_decorators(State, _) ->
     {true, query_notify_decorators_info(State)}.
 
 notify_decorators_effect(#?MODULE{cfg = #cfg{resource = QName}} = State) ->
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -41,6 +41,7 @@
          query_single_active_consumer/1,
          query_in_memory_usage/1,
          query_peek/2,
+         query_notify_decorators_info/1,
          usage/1,
 
          zero/1,
@@ -241,7 +242,7 @@ apply(Meta, #credit{credit = NewCredit, delivery_count = RemoteDelCnt,
             {State1, ok, Effects} =
             checkout(Meta, State0,
                      State0#?MODULE{service_queue = ServiceQueue,
-                                    consumers = Cons}, []),
+                                    consumers = Cons}, [], false),
             Response = {send_credit_reply, messages_ready(State1)},
             %% by this point all checkouts for the updated credit value
             %% should be processed so we can evaluate the drain
@@ -299,7 +300,8 @@ apply(#{index := Index,
     Exists = maps:is_key(ConsumerId, Consumers),
     case messages_ready(State0) of
         0 ->
-            update_smallest_raft_index(Index, {dequeue, empty}, State0, []);
+            update_smallest_raft_index(Index, {dequeue, empty}, State0,
+                                       [notify_decorators_effect(State0)]);
         _ when Exists ->
             %% a dequeue using the same consumer_id isn't possible at this point
             {State0, {dequeue, empty}};
@@ -330,8 +332,8 @@ apply(#{index := Index,
                     {{dequeue, {MsgId, Msg}, Ready-1}, Effects1}
 
             end,
-
-            case evaluate_limit(Index, false, State0, State4, Effects2) of
+            NotifyEffect = notify_decorators_effect(State4),
+            case evaluate_limit(Index, false, State0, State4, [NotifyEffect | Effects2]) of
                 {State, true, Effects} ->
                     update_smallest_raft_index(Index, Reply, State, Effects);
                 {State, false, Effects} ->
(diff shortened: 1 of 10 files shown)


deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -299,7 +300,8 @@ apply(#{index := Index,
     Exists = maps:is_key(ConsumerId, Consumers),
     case messages_ready(State0) of
         0 ->
-            update_smallest_raft_index(Index, {dequeue, empty}, State0, []);
+            update_smallest_raft_index(Index, {dequeue, empty}, State0,
+                                       [notify_decorators_effect(State0)]);
```
kjnilsson, line 304: why are we notifying decorators here?

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -456,11 +460,13 @@ apply(#{system_time := Ts} = Meta, {down, Pid, noconnection},
     % Monitor the node so that we can "unsuspect" these processes when the node
     % comes back, then re-issue all monitors and discover the final fate of
     % these processes
+
+    NotifyEffect = notify_decorators_effect(State),
```
kjnilsson, line 464: what is the scenario we are notifying here?

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -1501,14 +1518,27 @@ return_all(Meta, #?MODULE{consumers = Cons} = State0, Effects0, ConsumerId,
...
 %% checkout new messages to consumers
-checkout(#{index := Index} = Meta, OldState, State0, Effects0) ->
+checkout(#{index := Index} = Meta, #?MODULE{cfg = #cfg{resource = QName}} = OldState, State0,
+         Effects0) ->
     {State1, _Result, Effects1} = checkout0(Meta, checkout_one(Meta, State0),
                                             Effects0, {#{}, #{}}),
     case evaluate_limit(Index, false, OldState, State1, Effects1) of
         {State, true, Effects} ->
-            update_smallest_raft_index(Index, State, Effects);
+            case have_active_consumers_changed(OldState, State) of
```
kjnilsson, line 1527: I am not sure about adding more code this very hot path. is there nowhere else we an do this and achieve the same?

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1524,3 +1529,25 @@ update_type_state(Q, Fun) when ?is_amqqueue(Q) ->
...
+    QPid = amqqueue:get_pid(Q),
+    case ra:local_query(QPid, fun rabbit_fifo:query_notify_decorators_info/1) of
+        {ok, {_, {MaxActivePriority, IsEmpty}}, _} ->
+            notify_decorators(QName, consumer_state_changed, [MaxActivePriority, IsEmpty]);
+        _ -> ok
+    end.
+
+notify_decorators(QName, Event) -> notify_decorators(QName, Event, []).
+
+notify_decorators(QName, F, A) ->
```
kjnilsson, line 1545: This needs to run in a separate processes else it may crash or block the Ra leader.

kjnilsson (changes requested):

deps/rabbit/src/rabbit_fifo.erl
dcorbacho: There is a new consumer, so the federation can set the corresponding consumer on the upstream. Strictly it might only be needed when the first consumer appears and there is no other active consumer. I'll have a think if we can easily verify that

deps/rabbit/src/rabbit_fifo.erl
dcorbacho: Consumers are marked as suspected down, so it could end up with no active consumers at all and that needs to be notified.

deps/rabbit/src/rabbit_fifo.erl
dcorbacho: We have to detect changes in the active consumers, unless we keep track of them on a different way for this exclusive use I'm not sure how we can do it

deps/rabbit/src/rabbit_quorum_queue.erl
dcorbacho: done

deps/rabbit/src/rabbit_fifo.erl
michaelklishin: If we don't do this here, what would be the observable effects? If it's a minor delay in consumer topology changes, we can live with that over more code executed on the hot path.

dcorbacho: I've removed the notifications for enqueue/return/credit/settle/discard operations. It seems enough for federation purposes (our only decorator?) to only notify on checkout/node up/node down/config updates. At least, federation test suite still passes.

michaelklishin (changes requested):
Right now policies that include federation-related keys are not applied to quorum queues.

michaelklishin (approved):

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -1500,14 +1521,30 @@ return_all(Meta, #?MODULE{consumers = Cons} = State0, Effects0, ConsumerId,
...
+    checkout(Meta, OldState, State, Effects, true).
+
+checkout(#{index := Index} = Meta, #?MODULE{cfg = #cfg{resource = QName}} = OldState, State0,
+         Effects0, HandleConsumerChanges) ->
     {State1, _Result, Effects1} = checkout0(Meta, checkout_one(Meta, State0),
                                             Effects0, {#{}, #{}}),
     case evaluate_limit(Index, false, OldState, State1, Effects1) of
         {State, true, Effects} ->
-            update_smallest_raft_index(Index, State, Effects);
+            case have_active_consumers_changed(State, HandleConsumerChanges) of
```
michaelklishin, line 1533: @dcorbacho @kjnilsson didn't we want to get rid of this on the hot delivery path?

dcorbacho: I thought I did, unless it got lost on the cherry-pick. Weird, I'll review it again

dcorbacho: Sorry, that's right. Now we only apply it if the `HandleConsumerChanges` flag is set to `true`, excluding all enqueue/return/credit/settle/discard operations - the hot delivery path. It's only on checkout/node up/node down/config updates that we notify the decorators

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -2126,3 +2163,16 @@ get_priority_from_args(#{args := Args}) ->
     end;
 get_priority_from_args(_) ->
     0.
+
+have_active_consumers_changed(_, false) ->
```
kjnilsson, line 2167: This function name doesn't seem to match what it does. The result doesn't answer the question if active consumer have changed.

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -1500,14 +1521,30 @@ return_all(Meta, #?MODULE{consumers = Cons} = State0, Effects0, ConsumerId,
                 end, {State, Effects0}, Checked).
 
 %% checkout new messages to consumers
-checkout(#{index := Index} = Meta, OldState, State0, Effects0) ->
+checkout(Meta, OldState, State, Effects) ->
+    checkout(Meta, OldState, State, Effects, true).
```
kjnilsson, line 1525: Would it be better to default to false here in terms of number of lines changes?

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1525,3 +1541,26 @@ parse_credit_args(Default, Args) ->
...
+notify_decorators(Q) when ?is_amqqueue(Q) ->
+    QName = amqqueue:get_name(Q),
+    QPid = amqqueue:get_pid(Q),
+    case ra:local_query(QPid, fun rabbit_fifo:query_notify_decorators_info/1) of
+        {ok, {_, {MaxActivePriority, IsEmpty}}, _} ->
+            notify_decorators(QName, consumer_state_changed, [MaxActivePriority, IsEmpty]);
+        _ -> ok
+    end.
+
+notify_decorators(QName, Event) -> notify_decorators(QName, Event, []).
```
kjnilsson, line 1555: newline

kjnilsson (changes requested):
The function naming is the main thing I think we should clarify before merging.

deps/rabbit/src/rabbit_fifo.erl
dcorbacho: Not really, there are 2 more 'true' than 'false'

PR #2806, opened 2021-02-10
New configuration settings for TLS: wildcard-aware hostname verification, SNI

### LINKED ISSUE #2805: LDAP connections that use TLS can fail with "mismatching" SNI/hostname verification settings
Modern Erlang versions automatically [enable SNI/hostname verification](https://erlang.org/doc/man/ssl.html#connect-3). If target servers use wildcard x.509 certificates,
then Erlang's TLS library requires the [hostname check to be customized](https://erlang.org/doc/man/ssl.html#type-sni), which is not currently possible using a static term config (`advanced.config`).

So we need to provide a configuration setting that would support a few cases:

 * Wildcard certificates
 * Exact matching
 * Disabling hostname verification entirely (not to be confused with [disabling peer verification](https://www.rabbitmq.com/ssl.html#peer-verification), which actually affects security meaningfully and can already be done)

michaelklishin:
* [Client] hostname verification matching: wildcard or none
 * SNI hostname or none

Closes #2805


michaelklishin:
Backported to `v3.8.x`.

PR #2810, opened 2021-02-11
Provide a means to disable and enable classic queue mirroring within a vhost

Ayanda-D:
## Proposed Changes

The following changes provide a means of enabling and disabling
HA mirroring for classic queues in a virtual host. The following CLI
commands are implemented:

```
rabbitmqctl -p vhost-1  disable_classic_mirroring
```
```
rabbitmqctl -p vhost-1  enable_classic_mirroring
```

Internally, a generic ACL for runtime parameters is used to manage
the policies and parameters which are allowed/disallowed. Only HA
mirroring is currently implemented.

Reference: https://groups.google.com/g/rabbitmq-users/c/PR5D8GAbWyo

## Types of Changes

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue_process.erl
+++ b/deps/rabbit/src/rabbit_amqqueue_process.erl
@@ -1794,9 +1794,15 @@ log_auto_delete(Reason, #q{ q = Q }) ->
 needs_update_mirroring(Q, Version) ->
     {ok, UpQ} = rabbit_amqqueue:lookup(amqqueue:get_name(Q)),
     DBVersion = amqqueue:get_policy_version(UpQ),
-    case DBVersion > Version of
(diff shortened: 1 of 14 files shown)


michaelklishin:
Interesting. Thank you for investigating this. It will take a team-wide discussion about how and if we'd like to consider deprecating CMQs like this (for now, eventually they will be removed).

Ayanda-D:
We understand decisions on this might take time. In the time being, we provide a means 
to deprecate classic queues at the channel layer through use of the following plugin:
https://github.com/esl/rabbitmq-deprecate-classic-queues

michaelklishin:
@Ayanda-D a plugin is a better option for us because it is entirely opt-in, nothing changes (intentionally or not) for those who do not run it. But yeah, the race towards 3.9 and some commercial extensions means some PRs haven't been reviewed after a few months :(

PR #2823, opened 2021-02-17
Restore packaging dir

pjk25:
https://github.com/rabbitmq/rabbitmq-server/commit/2067a687351e1caf5ddb8b54b8e9118c2a74d60b removed the `packaging` directory and extracted to https://github.com/rabbitmq/rabbitmq-packaging in order to satisfy certain Concourse pipeline issues.

In hindsight that appears to have been overly aggressive, as broke the ability to conveniently run `make package-generic-unix` and similar commands from `rabbitmq-server`.

This brings all files back, including recent changes made in the new repo, except the minimal set required by the pipelines.


packaging/docker-image/Makefile
```diff
@@ -0,0 +1,32 @@
...
+	xzcat $(GENERIC_UNIX_ARCHIVE) | tar xvf -
+	docker build --pull \
+	  --build-arg PGP_KEYSERVER=pgpkeys.uk \
+	  --build-arg OTP_VERSION=$(OTP_VERSION) \
+	  --build-arg OTP_SHA256=$(OTP_SHA256) \
+	  --build-arg RABBITMQ_VERSION=$(VERSION) \
+	  --build-arg RABBITMQ_BUILD=rabbitmq_server-$(VERSION) \
+	  --tag pivotalrabbitmq/rabbitmq:$(VERSION) \
+	  .
+	docker push pivotalrabbitmq/rabbitmq:$(VERSION)
```
Gsantomaggio, line 29: I think we should remove:
```
docker push pivotalrabbitmq/rabbitmq:$(VERSION)
```
Build the docker image based on source code is useful for k8s tests. 
We'd like to avoid to push the tag each time.
We can:
 - just remove the `push` and leave to the user the decision
 - create anther make step like `make push-docker-image`


@gerhard thought?

gerhard: Yes, pushing should be a separate target. We do the same in rabbitmq_prometheus: https://github.com/rabbitmq/rabbitmq-server/blob/0dc46b41de9f5f51dfa12bfb5e4336c7b32e2e6d/deps/rabbitmq_prometheus/Makefile#L130-L133

packaging/Makefile
```diff
@@ -0,0 +1,120 @@
...
+	$(gen_verbose) $(MAKE) -C RPMS/Fedora $(VARS) RPM_OS=sles11 all $(DO_CLEAN)
+
+package-windows: $(SOURCE_DIST_FILE)
+	$(gen_verbose) $(MAKE) -C windows $(VARS) all $(DO_CLEAN)
+	$(verbose) $(MAKE) -C windows-exe $(VARS) all $(DO_CLEAN)
+
+package-generic-unix: $(SOURCE_DIST_FILE)
+	$(gen_verbose) $(MAKE) -C generic-unix $(VARS) all $(DO_CLEAN)
+
+docker-image:
```
Gsantomaggio, line 112: `make docker-image` doesn't work because of the RabbitMQ version format. 

ex:
```bash
$  make docker-image
.....
docker build --pull \
          --build-arg PGP_KEYSERVER=pgpkeys.uk \
          --build-arg OTP_VERSION=23.2.4 \
          --build-arg OTP_SHA256=e72aa084907e0f34f932cf00caa33aba93147b0a7c9c35569d6bd1c402f532de \
          --build-arg RABBITMQ_VERSION=3.8.10+66.gee45535 \
          --build-arg RABBITMQ_BUILD=rabbitmq_server-3.8.10+66.gee45535 \
          --tag pivotalrabbitmq/rabbitmq:3.8.10+66.gee45535 \
          .
invalid argument "pivotalrabbitmq/rabbitmq:3.8.10+66.gee45535" for "-t, --tag" flag: invalid reference format
See 'docker build --help'.
make[2]: *** [dist] Error 125
make[1]: *** [docker-image] Error 2
make: *** [docker-image] Error 2
```

The docker `tag` doesn't support `+` (ex `--tag pivotalrabbitmq/rabbitmq:3.8.10+66.gee45535 \`)  so we should replace `+` with something else when we tag the image

gerhard: This is what I did last time I had to run this:

```
make source-dist PROJECT_VERSION=3.8.12-rc.3-20200212.1331
make package-generic-unix SOURCE_DIST_FILE=/Users/gerhard/github.com/rabbitmq/rabbitmq-server/PACKAGES/rabbitmq-server-3.8.12-rc.3-20200212.1331.tar.xz
make docker-image SOURCE_DIST_FILE=/Users/gerhard/github.com/rabbitmq/PACKAGES/rabbitmq-server-generic-unix-3.8.12-rc.3-20200212.1331.tar.xz
```

packaging/docker-image/Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+dist:
+	xzcat $(GENERIC_UNIX_ARCHIVE) | tar xvf -
+	docker build --pull \
+	  --build-arg PGP_KEYSERVER=pgpkeys.uk \
+	  --build-arg OTP_VERSION=$(OTP_VERSION) \
+	  --build-arg OTP_SHA256=$(OTP_SHA256) \
+	  --build-arg RABBITMQ_VERSION=$(VERSION) \
+	  --build-arg RABBITMQ_BUILD=rabbitmq_server-$(VERSION) \
+	  --tag pivotalrabbitmq/rabbitmq:$(VERSION) \
```
Gsantomaggio, line 27: @gerhard about https://github.com/rabbitmq/rabbitmq-server/pull/2823#discussion_r580329587 it is ok but I would fix `$(VERSION)` with by replacing `+` with something else, so it will easier to build a temp docker image.
At the moment it fails.

gerhard: OK 👍🏻

packaging/docker-image/Makefile
pjk25: I'd vote for the separate target

Gsantomaggio: OK

packaging/Makefile
pjk25: Depending on the circumstances, I have found that using a date as above can be more convenient when iterating on unpublished changes. However, I doubt there is an ideal solution for all cases. It's also worth noting that if you have uncommitted changes, you will currently get a version such as `rabbitmq-server-3.8.0+rc.1.733.gee45535.dirty` which is reasonable, but not reproducible. I think replacing `+` with `-` at the point of tagging would be reasonable, since I think it's important that setting `PROJECT_VERSION` as a make var still has the intended effect.

packaging/docker-image/Makefile
```diff
@@ -1,4 +1,25 @@
...
+if test -f git-revisions.txt; then \
+	head -n1 git-revisions.txt | \
+	awk '{print $$$(words $(PROJECT_DESCRIPTION) version);}'; \
+else \
+	(git describe --dirty --abbrev=7 --tags --always --first-parent \
+	 2>/dev/null || echo rabbitmq_v0_0_0) | \
+	sed -e 's/^rabbitmq_v//' -e 's/^v//' -e 's/_/./g' -e 's/-/+/' \
+	 -e 's/-/./g'; \
+fi)
+endif
```
pjk25, line 20: This is copied from rabbitmq-components.mk. I preferred it to including the whole file, since rabbitmq-components.mk is semi-redundant in the monorepo.

pjk25: After further inspection, I realize that by default after a successful build we do clean, and the original logic is sufficient. Still I think it's worth keeping the enhancement.

pjk25:
Backported to `3.8.x`

24fe2c1e8c88d39cd92bd476004da25152323de5
fcba51e54abd3e76d165c810d123abc674688af9
cc25c052b8e9149244537dc7185d1ff567befdbb
0c5a3210ac6fd42d064cc9a0a1aa8f6be772d2b9
afdc54cd909ed4dfd0f2e6760a994b09f3e6b2f5
c06788494bfed0d012a7f5da59cb7fe35e7985d7
fb40233b06b8dbe90f27f917d32ea74253aa664b
0fdfa6f469a22e9beb4309af6b7a648e7b06d13e

(next time I will remember not to rebase merge a PR)

PR #2829, opened 2021-02-19
Make it possible to use a single channel for message transfers and commands

michaelklishin:
For the rare case where federated exchange experience rapid
binding changes. As of rabbitmq/rabbitmq-federation#97,
such environments have a race condition between binding
and message propagation.


michaelklishin:
Backported to `v3.8.x`.

Issue #2837, opened 2021-02-23
`make docker-image` fails on master due to the packaging of plugins as directories


pjk25:
`make docker-image` builds a `management` image where the management plugin is enabled. The current `Dockerfile` attempts to extract the `.ez` file at https://github.com/rabbitmq/rabbitmq-server/blob/7c63b8e4fb1e90f6e3991958824f9f90f1c1dd50/packaging/docker-image/Dockerfile#L278-L296 and fails.

pjk25:
Fixed with #2838

PR #2838, opened 2021-02-23
Fix the docker image for use with uncompressed plugins

pjk25:
Historically the docker image build extracts the rabbitmqadmin cli from the
management .ez. This step is now conditional.

This should address #2837


pjk25:
@Gsantomaggio please give this a try when you have a chance.

PR #2842, opened 2021-02-24
Encode direct reply-to "client identifiers" in a more compact way

michaelklishin:
This change limits direct reply-to identifier length growth as node names grow.

Prior to this change, direct reply-to consumer channels
were encoded using term_to_binary/1, which means the result
would grow together with node name (since node name
is one of the components of an Erlang pid type).

This means that with long enough hostnames, reply-to
identifiers could overflow the 255 character limit of
message property field type, longstr.

With this change, the encoded value uses a hash of the node name
and then locates the actual node name from a map of
hashes to current cluster members.

In addition, instead of generating non-predictable "secure"
GUIDs the feature now generates "regular" predictable GUIDs
which compensates some of the additional PID pre- and post-processing
outlined above.

### DIFF
--- a/deps/rabbit/src/pid_recomposition.erl
+++ b/deps/rabbit/src/pid_recomposition.erl
@@ -1,0 +1,58 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
+%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%
+%% Copyright (c) 2007-2021 VMware, Inc. or its affiliates.  All rights reserved.
+%%
+
+-module(pid_recomposition).
+
+
+%% API
+-export([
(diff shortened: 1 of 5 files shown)


ansd (approved):
Tested in a single node cluster that direct-reply-to succeeds with a hostname length of 200 characters. Thanks @michaelklishin !

dumbbell:
Are there interoperability issues if some nodes in the cluster run a version which doesn't understand this new encoding?

michaelklishin:
@dumbbell such nodes won't be able to route responses produced for requests issued against newer nodes. Otherwise this is node-local state and it does not require any special consideration during upgrades.

As nodes are upgraded and restarted, all channels on them will be closed, reopened and will generate new keys and encoded pid values for the response consumers.

michaelklishin:
Backported to `v3.8.x`.

PR #2847, opened 2021-02-25
WIP Upgrade Lager to 3.9.0 for OTP 24 compatibility

michaelklishin:
`lager_util:expand_path/1` use changes are due to erlang-lager/lager#540


michaelklishin:
Closing for now, we will revisit after the dust with this Lager release settles.

PR #2850, opened 2021-02-26
OTP 24 compat: switch management and management_agent to use pg

michaelklishin:
since pg2 was removed in OTP 24.

The only decision worth mentioning here is that both plugins share
a pg scope, which is started in rabbitmq_management_agent supervision
tree and idempotently started in rabbitmq_management without
attaching the scope pid to its tree.

Per discussion with @lhoguin.


deps/rabbitmq_management_agent/src/rabbit_mgmt_agent_sup_sup.erl
```diff
@@ -13,16 +13,40 @@
...
     supervisor2:start_child(?MODULE, sup()).
 
 sup() ->
-    {rabbit_mgmt_agent_sup, {rabbit_mgmt_agent_sup, start_link, []},
-     temporary, ?SUPERVISOR_WAIT, supervisor, [rabbit_mgmt_agent_sup]}.
+    #{
+        id      => rabbit_mgmt_agent_sup,
+        start   => {rabbit_mgmt_agent_sup, start_link, []},
+        restart => temporary,
+        wait    => ?SUPERVISOR_WAIT,
```
lhoguin, line 26: The `supervisor2:child_spec()` type says this is called `shutdown`, not `wait`. Same problem in `rabbit_mgmt_sup_sup`.

lhoguin (changes requested):

lhoguin (approved):

michaelklishin:
For as long as we support Erlang 22, we'd have to keep this 3.9-specific. After Erlang 22 support is dropped, we can backport this.

PR #2854, opened 2021-03-01
Replace Async threads warning with Dirty I/O schedulers warning

mkuratczyk:
Async threads are basically not used these days - there is no reason to warn about a low number of them (the default is 1 so the warning is displayed in almost all environments). Dirty I/O schedulers, on the other hand, are used a lot so we can warn when there are only a few (the default in Erlang is 10 so I made it the warning threshold).

### DIFF
--- a/deps/rabbit/src/rabbit.erl
+++ b/deps/rabbit/src/rabbit.erl
@@ -271,7 +271,7 @@
 
 -define(APPS, [os_mon, mnesia, rabbit_common, rabbitmq_prelaunch, ra, sysmon_handler, rabbit, osiris]).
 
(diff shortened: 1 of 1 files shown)


lhoguin (approved):

lhoguin:
@michaelklishin What about cherry-picking PRs such as this one to `v3.8.x`? Should a separate `v3.8.x` be opened?

michaelklishin:
@lhoguin no, simply cherry-pick this using

```
git cherry-pick -x $(git show-ref refs/heads/master -s) -m 1
```

lhoguin:
Done.

PR #2858, opened 2021-03-02
WIP Add metric for quorum queue availability

ansd:
**Why this change?**
Having a metric reporting availability for a quorum queue enables us to define alerts if quorum queues are down and to define SLOs on quorum queue availability.

**How this change has been tested**:

1. Built docker image, deployed on 3-node K8s cluster.
2. Declared quorum queue `A` with leader on node 0.
3. Declared quorum queue `B` with leader on node 1.
4. ` kubectl exec r-server-2 -- rabbitmqctl stop_app`
5. ` kubectl exec r-server-1 -- rabbitmqctl stop_app`
6. The following PromQL alerting rule will result in 2 active alerts (one for `A`, one for `B`):

```yaml
# Alert if some node reports that a quorum queue exists without having the leader reporting that the quorurm queue is up.
# This alert covers the following two use cases:
# 1. Leader explicitly reports that quorum queue is down because ra:consistent_query() failed - i.e. leader is live, but doesn't have a majority.
# 2. No rabbitmq_quorum_queues_up got reported for a quorum queue although some node reported that the quorum queue exists. This happens if there is no leader at all but at least a follower or candidate is live.
# NB: This alert will NOT fire if ALL replicas of a quorum queue are down (since we can't differentiate between deleted quorum queues and all replicas are down). In future, we can solve this by having the messaging topology operator report the desired quorum queues.
# Prerequisite for this alert to fire is `prometheus.return_per_object_metrics = true`.
- alert: QuorumQueueDown
  expr: |
    group by (namespace, rabbitmq_cluster, vhost, queue) (rabbitmq_raft_term_total{queue!=""} * on(instance) group_left(rabbitmq_cluster) rabbitmq_identity_info)
    unless on(namespace, rabbitmq_cluster, vhost, queue)
    max by (namespace, rabbitmq_cluster, vhost, queue) (rabbitmq_quorum_queues_up * on(instance) group_left(rabbitmq_cluster) rabbitmq_identity_info) == 1
  for: 15m
  labels:
    severity: warning
  annotations:
    description: "Quorum queue `{{ $labels.queue }}` is unavailable in vhost `{{ $labels.vhost }}` in RabbitMQ cluster `{{ $labels.rabbitmq_cluster }}` in namespace `{{ $labels.namespace }}`."
    summary: |
      This alert fires if nodes report that a quorum queue exists for which no majority is available.
      For example, in a 3-node RabbitMQ cluster with a 3-replicas quorum queue, this alert fires if 2 out of 3 replicas are down.
      Check the logs of the RabbitMQ nodes: `kubectl -n {{ $labels.namespace }} logs --tail 100 -l app.kubernetes.io/component=rabbitmq,app.kubernetes.io/name={{ $labels.rabbitmq_cluster }} | grep {{ $labels.queue }}`
```

**Implementation Details**

- The per-object metric returns `1` for up, and `0` for down.
- The aggregated metric returns the sum of quorum queues having a leader on this node for which the leader reports that a majority is available in the Raft cluster.
- Reuse ETS table `queue_metrics` since it already gets initialised and destroyed for us. This table already contains different metrics depending on queue type.
- I first implemented a version where `ra:consistent_query()` is only executed if `prometheus.return_per_object_metrics = true` to not cause unnecessary load. However, this results in `/metrics/per-object` endpoint to not contain this new metric. Therefore, I discarded that optimisation again.

**Open questions**
- In certain scenarios `ra:consistent_query()` failed although I could write to and read from the quorum queue and all replicas were up. This happens for example after nodes were restarted. It took a while (10 - 90 seconds) until the query was successful again. It seems that `ra:consistent_query()` is stricter than a "leader is active and has a majority" check. Should we therefore rename the metric from `rabbitmq_quorum_queues_up` to `rabbitmq_quorum_queues_consistent`?
- As of now, a  `ra:consistent_query()` is executed every 5 seconds (configured via [TICK_TIMEOUT](https://github.com/rabbitmq/rabbitmq-server/blob/4465071820e143d8f380baca9dd2b4802da42840/deps/rabbit/src/rabbit_quorum_queue.erl#L90)). It's sufficient to execute the check less often. (I expected increasing `collect_statistics_interval` results in collecting statistics less often for a quorum queue. However, this configuration setting didn't have any effect at all on quorum queue statistics. It seems to apply primarily to the management plugin.) Increasing `TICK_TIMEOUT` seems to have other side effects since this value is not only used for how often [tick/2](https://github.com/rabbitmq/rabbitmq-server/blob/4465071820e143d8f380baca9dd2b4802da42840/deps/rabbit/src/rabbit_fifo.erl#L692-L710) is called, but also used for leader election algorithm in Ra. Do you think it's necessary to execute the check less often? If so, should we either add a new variable to the Raft state which is only used for how often `tick/2` is called? Or, should we instead write a timestamp into the `queue_metrics` ETS table in addition to (or instead of) the `{up, 1}` when having executed a `ra:consistent_query()` and in the next tick only execute `ra:consistent_query()` if the timestamp is old enough (e.g. older than 15 seconds)? Looking at Prometheus best practices, it's cleaner to simply report `1` for "up" instead of reporting a timestamp.

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -900,10 +900,19 @@ policy_changed(Q) ->
 cluster_state(Name) ->
     case whereis(Name) of
         undefined -> down;
-        _ ->
+        Pid ->
             case ets:lookup(ra_state, Name) of
                 [{_, recover}] -> recovering;
-                _ -> running
+                _ ->
+                    case ra:consistent_query(Pid,
+                                             fun (_) -> ok end,
+                                             ?RPC_TIMEOUT) of
+                        {ok, ok, _} ->
+                            running;
+                        _ ->
+                            rabbit_log:debug("ra:consistent_query failed for quorum queue ~w", [Name]),
+                            down
+                    end
             end
     end.
 
--- a/deps/rabbit/test/quorum_queue_SUITE.erl
+++ b/deps/rabbit/test/quorum_queue_SUITE.erl
@@ -66,6 +66,7 @@ groups() ->
                                             metrics_cleanup_on_leadership_takeover,
                                             metrics_cleanup_on_leader_crash,
                                             consume_in_minority,
+                                            state_in_minority,
                                             shrink_all,
                                             rebalance,
                                             file_handle_reservations,
@@ -692,6 +693,31 @@ consume_in_minority(Config) ->
     ok = rabbit_ct_broker_helpers:start_node(Config, Server1),
     ok = rabbit_ct_broker_helpers:start_node(Config, Server2),
     ok.
+
+state_in_minority(Config) ->
+    [Server0, Server1, Server2] =
+        rabbit_ct_broker_helpers:get_node_configs(Config, nodename),
+
+    Ch = rabbit_ct_client_helpers:open_channel(Config, Server0),
+    QQ = ?config(queue_name, Config),
+    ?assertEqual({'queue.declare_ok', QQ, 0, 0},
+                 declare(Ch, QQ, [{<<"x-queue-type">>, longstr, <<"quorum">>}])),
+    RaName = ra_name(QQ),
+    {ok, _, {_, Leader}} = ra:members({RaName, Server0}),
+    QNameRes = rabbit_misc:r(<<"/">>, queue, QQ),
+    AssertState = fun(ExpectedState) ->
+                      timer:sleep(1000),
+                      [{_, PropList, _}] = rpc:call(Leader, ets, lookup, [queue_metrics, QNameRes]),
+                      ?assertEqual(ExpectedState, proplists:get_value(state, PropList))
+                  end,
+
+    AssertState(running),
+    ok = rabbit_ct_broker_helpers:stop_node(Config, Server1),
+    ok = rabbit_ct_broker_helpers:stop_node(Config, Server2),
+    AssertState(down),
+    ok = rabbit_ct_broker_helpers:start_node(Config, Server1),
+    ok = rabbit_ct_broker_helpers:start_node(Config, Server2),
+    AssertState(running).
 
 shrink_all(Config) ->
     [Server0, Server1, Server2] =
@@ -2522,9 +2548,7 @@ consumer_metrics(Config) ->
     timer:sleep(5000),
     QNameRes = rabbit_misc:r(<<"/">>, queue, QQ),
     [{_, PropList, _}] = rpc:call(Leader, ets, lookup, [queue_metrics, QNameRes]),
-    ?assertMatch([{consumers, 1}], lists:filter(fun({Key, _}) ->
-                                                        Key == consumers
-                                                end, PropList)).
+    ?assertEqual(1, proplists:get_value(consumers, PropList)).
 
 delete_if_empty(Config) ->
     Server = rabbit_ct_broker_helpers:get_node_config(Config, 0, nodename),
--- a/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
+++ b/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
@@ -36,7 +36,7 @@
 %% ==How to determine if a metric should be of type GAUGE or COUNTER?==
 %%
 %% * GAUGE if you care about its value rather than rate of change
-%%   - value can decrease as well as decrease
+%%   - value can increase as well as decrease
 %% * COUNTER if you care about the rate of change
 %%   - value can only increase
 %%
@@ -188,7 +188,8 @@
         {2, undefined, queue_messages_paged_out, gauge, "Messages paged out to disk", messages_paged_out},
         {2, undefined, queue_messages_paged_out_bytes, gauge, "Size in bytes of messages paged out to disk", message_bytes_paged_out},
         {2, undefined, queue_disk_reads_total, counter, "Total number of times queue read messages from disk", disk_reads},
-        {2, undefined, queue_disk_writes_total, counter, "Total number of times queue wrote messages to disk", disk_writes}
+        {2, undefined, queue_disk_writes_total, counter, "Total number of times queue wrote messages to disk", disk_writes},
+        {2, {bool, running}, queues_up, gauge, "Number of queues in state running", state}
     ]},
 
     {auth_attempt_metrics, [
@@ -318,6 +319,8 @@ mf(Callback, Contents, Data) ->
         Fun = case Conversion of
                   undefined ->
                       fun(D) -> proplists:get_value(Key, element(Index, D)) end;
+                  {bool, MatchPattern} ->
+                      fun(D) -> bool(proplists:get_value(Key, element(Index, D)), MatchPattern) end;
                   BaseUnitConversionFactor ->
                       fun(D) -> proplists:get_value(Key, element(Index, D)) / BaseUnitConversionFactor end
               end,
@@ -441,9 +444,9 @@ get_data(channel_metrics = Table, false) ->
                {messages_uncommitted, A4}, {acks_uncommitted, A5}, {prefetch_count, A6},
                {global_prefetch_count, A7}]}];
 get_data(queue_metrics = Table, false) ->
-    {Table, A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12, A13, A14, A15, A16} =
+    {Table, A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12, A13, A14, A15, A16, A17} =
         ets:foldl(fun({_, Props, _}, {T, A1, A2, A3, A4, A5, A6, A7, A8, A9, A10,
-                                      A11, A12, A13, A14, A15, A16}) ->
+                                      A11, A12, A13, A14, A15, A16, A17}) ->
                           {T,
                            sum(proplists:get_value(consumers, Props), A1),
                            sum(proplists:get_value(consumer_utilisation, Props), A2),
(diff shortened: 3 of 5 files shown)


deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -406,7 +406,14 @@ handle_tick(QName,
                                {messages_persistent, M}
 
                                | infos(QName, ?STATISTICS_KEYS -- [consumers])],
-                      rabbit_core_metrics:queue_stats(QName, Infos),
+                      {ok, Queue} = rabbit_amqqueue:lookup(QName),
```
michaelklishin, line 409: This does not handle the case where the queue record no longer exists.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -406,7 +406,14 @@ handle_tick(QName,
...
+                      Up = case ra:consistent_query(amqqueue:get_pid(Queue), fun (_) -> ok end) of
```
michaelklishin, line 410: This can be really expensive to do this for every quorum queue every N seconds with 10s of thousands of queues.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -406,7 +406,14 @@ handle_tick(QName,
...
+                          {ok, ok, _} -> 1;
+                          _ ->
+                              rabbit_log:debug("ra:consistent_query failed for queue ~p~n", [QName]),
```
michaelklishin, line 413: Instead of using `~p` on queue name, consider `rabbit_misc:rs/1`

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -406,7 +406,14 @@ handle_tick(QName,
...
+                              0
+                      end,
+                      rabbit_core_metrics:queue_stats(QName, [{up, Up} | Infos]),
```
michaelklishin, line 416: Were other names considered? All this tells us is that a quorum queue does or does not have an online quorum.

deps/rabbit/test/quorum_queue_SUITE.erl
```diff
@@ -2526,6 +2527,23 @@ consumer_metrics(Config) ->
...
+    Ch1 = rabbit_ct_client_helpers:open_channel(Config, Server),
+    QQ = ?config(queue_name, Config),
+    ?assertEqual({'queue.declare_ok', QQ, 0, 0},
+                 declare(Ch1, QQ, [{<<"x-queue-type">>, longstr, <<"quorum">>}])),
+
+    RaName = ra_name(QQ),
+    {ok, _, {_, Leader}} = ra:members({RaName, Server}),
+    timer:sleep(5000),
+    QNameRes = rabbit_misc:r(<<"/">>, queue, QQ),
+    [{_, PropList, _}] = rpc:call(Leader, ets, lookup, [queue_metrics, QNameRes]),
```
michaelklishin, line 2542: Converting this proplist to a map and getting the key would be clearer.

deps/rabbit/src/rabbit_quorum_queue.erl
ansd: If the queue record doesn't exist anymore, will the `handle_tick` function still be called for the queue?
I thought it's safe to pattern match here give the same is done in https://github.com/rabbitmq/rabbitmq-server/blob/4465071820e143d8f380baca9dd2b4802da42840/deps/rabbit/src/rabbit_quorum_queue.erl#L425
? However, I might be wrong.

deps/rabbit/src/rabbit_quorum_queue.erl
ansd: Yes, see my comment in the PR under `Open questions`. We can execute it less often. If it's still too expensive, can you think of another way to report a metric for quorum queue availability?

deps/rabbit/src/rabbit_quorum_queue.erl
ansd: Yes, see my comment in the PR under Open questions. Instead of `rabbitmq_quorum_queues_up`, we can call the metric `rabbitmq_quorum_queues_consistent`?

deps/rabbit/test/quorum_queue_SUITE.erl
ansd: I copied that from the surrounding test cases. I'm happy to change it in this test case and the surrounding test cases as well. Thanks.

deps/rabbit/src/rabbit_quorum_queue.erl
ansd: However, if the quorum queue doesn't have an online quorum, I think it's good to report this queue to be "down" since from the client perspective we can't write to it.

michaelklishin:
This can be very expensive to do for thousands of quorum queues. Please measure CPU usage with, say, 10K queues with and without this PR.

It general it makes sense to run this every net tick interval seconds. Ra does not rely on net ticks for peer unavailability detection but for this metric and alerting, it should be good enough. 

This mechanism is effectively polling. It'd be interesting to investigate if Ra could maintain a metric like this using the data it has, so that we potentially would have a less expensive way of computing this.

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
```diff
@@ -441,9 +442,9 @@ get_data(channel_metrics = Table, false) ->
                {messages_uncommitted, A4}, {acks_uncommitted, A5}, {prefetch_count, A6},
                {global_prefetch_count, A7}]}];
 get_data(queue_metrics = Table, false) ->
-    {Table, A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12, A13, A14, A15, A16} =
+    {Table, A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12, A13, A14, A15, A16, A17} =
```
michaelklishin, line 445: Instead of introducing a new metric, we could use the existing `status` field which all queue types have.

deps/rabbit/src/rabbit_quorum_queue.erl
ansd: Done.

dcorbacho:
What's the reasoning behind introducing a new metric for this? We already have a `state` field for all queue types, if the quorum queue state doesn't reflect the 'real' state then it should be updated and not superseded by yet another state field

https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit/src/rabbit_quorum_queue.erl#L1306

ansd:
@dcorbacho the current implementation of the `state` metric isn't good enough. We'd like to get an alert of a quorum is lost for a quorum queue. Currently, the `state` metric for the queue reports `running` even if the leader lost majority.

I like the idea of re-using the existing `state` field. I'll try that out.

@michaelklishin thanks for reviewing. I will check how CPU stats compare with 10K queues with and without this PR.

For the next couple of weeks, I'll have to work on something else having higher priority than alerting.
I'll get back to this PR thereafter.

deps/rabbit/src/rabbit_quorum_queue.erl
michaelklishin: It can be deleted concurrently. I guess then the process will typically be terminating, too, so this is good enough.

michaelklishin (commented):
@ansd our suggestion is to use the field and its current "down" (or a reasonable equivalent) value instead of introducing a new one. There can be many reasons why a queue can be considered "down".

Per discussion with @kjnilsson, this may be less expensive than I thought it but it is still a [Raft] cluster-wide fanout that returns when a majority of responses is received. We could run this on a different schedule, let's see how expensive this really is with 10K queues.

With tens of thousands of queues metric emission will almost always be performed less frequently than every 5s, which helps in this case.

deps/rabbit/src/rabbit_quorum_queue.erl
kjnilsson: I think your problems may come from the use of get_pid here. As this executes on the (assumed) leader node you should just be able to use 'Name' directly to issue a request to the leader on the local node. (Assuming Name is the locally registered name of the QQ process, else we'd have to pass that along).

kjnilsson: The consistent query timeout should probably be less than the tick interval btw.

deps/rabbit/src/rabbit_quorum_queue.erl
kjnilsson: you shouldn't need to get the queue record at all as all the information you need should be passed to this function from the tick.

ansd:
Thanks for your feedback.
I force pushed a commit (rebasing onto master).
I implemented your suggestion to re-use the existing `state` field.

I compared CPU with 10k quorum queues in a 3 node cluster.
In all the runs, CPU usage was much higher (around 20% higher) with the changes in this PR.

So, running these queries every 5 seconds is way too often.
Should we then try to run these checks less often (e.g. every 30 or 60 seconds)?
(I could re-run the tests with less frequent polling?)
Or, is polling not a viable option?

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
ansd: Done!
In the previous change, there was a new entry `up` in the ETS table which was exposed as a Prometheus metric.
In the [latest commit](https://github.com/rabbitmq/rabbitmq-server/pull/2858/commits/a42ea8a3bbecf00f6e5d79a447963a46db2bf216), the entry `state` is re-used in the ETS table and a `state==running` metric (called `queues_up`)  is exposed as a Prometheus metric.

deps/rabbit/test/quorum_queue_SUITE.erl
ansd: Using `proplists:get_value` to make it clearer.

deps/rabbit/src/rabbit_quorum_queue.erl
ansd: Using `?RPC_TIMEOUT` now since that is used by other RPCs as well. Might be too low however. I was thinking of using something like `round(0.75 * ?TICK_TIMEOUT)`.

gerhard:
Do you want to keep the branch @ansd ?

ansd:
@gerhard, I delete the branch here since I have a backup at https://github.com/ansd/rabbitmq-server/tree/qq-metric. Thanks!

PR #2873, opened 2021-03-07
Use AMQP subject for routing key in AMQP 1.0 source shovels.

Shigawire:
## Proposed Changes

In dynamic shovels, `dst-exchange-key` may be omitted. In such a case, the source routing key, or, in case of AMQP 1.0 source shovels, the AMQP 1.0 `subject` is supposed to be used as routing key. However, as discussed in https://groups.google.com/g/rabbitmq-users/c/3n0lNVmVbNo, this does not work for AMQP 1.0 source shovels.

This PR introduces a fix to actually 
a) use the AMQP 1.0 subject as routing key
b) gracefully handle `undefined` routing keys in shovelled messages.

Unfortunately, Erlang is an entirely new concept to me, I tried my best to follow the conventions I learned by browsing the project.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue (not issue #))

## Checklist

## Further Comments

I wasn't able to run the test suite, due to the following error: 
```Failed to locate Elixir lib dir```.

### DIFF
--- a/deps/amqp10_client/src/amqp10_msg.erl
+++ b/deps/amqp10_client/src/amqp10_msg.erl
@@ -11,6 +11,7 @@
          % "read" api
          delivery_id/1,
          delivery_tag/1,
+         subject/1,
          handle/1,
          settled/1,
          message_format/1,
@@ -124,6 +125,10 @@ to_amqp_records(#amqp10_msg{transfer = T,
 -spec delivery_tag(amqp10_msg()) -> delivery_tag().
 delivery_tag(#amqp10_msg{transfer = #'v1_0.transfer'{delivery_tag = Tag}}) ->
     unpack(Tag).
+
+-spec subject(amqp10_msg()) -> string().
+subject(#amqp10_msg{properties = #'v1_0.properties'{subject = Subject}}) ->
+    unpack(Subject).
 
 -spec delivery_id(amqp10_msg()) -> non_neg_integer().
 delivery_id(#amqp10_msg{transfer = #'v1_0.transfer'{delivery_id = Id}}) ->
--- a/deps/rabbitmq_shovel/src/rabbit_amqp091_shovel.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_amqp091_shovel.erl
@@ -155,7 +155,12 @@ forward(IncomingTag, Props, Payload,
     SrcUri = rabbit_shovel_behaviour:source_uri(State0),
     % do publish
     Exchange = maps:get(exchange, Props, undefined),
-    RoutingKey = maps:get(routing_key, Props, undefined),
+    % If no routing key is set or routing key is undefined, use an empty string.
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by michaelklishin** [comment]
This behavior is too specific and may or may not be welcomed by other AMQP 1.0 Shovel users. I'm afraid we cannot accept it. We would prefer to have a mapping or a set of settings where the routing key is taken from.

**Comment by kjnilsson** [comment]
Yes I think this behaviour is too implicit. Something like an [AMQP 1.0 Source configuration](https://www.rabbitmq.com/shovel-static.html) that allowed the exchange and routing key to be taken from any AMQP 1.0 data would be good. E.g.:

```
{source,
                   [ {protocol, amqp10,
                      uris, [ "amqp://fred:secret@host1.domain/my_vhost",
                            ]},
                     {source_address, <<"my-source">>},
                     {prefetch_count, 10},
                     {exchange, properties, to},
                     {routing_key, properties, subject},


                   ]}

```

Shigawire:
@michaelklishin and @kjnilsson Thanks a lot for the brief review and reply.
My apologies, I may be missing something here. I've oriented myself at the following docs:
- https://github.com/rabbitmq/rabbitmq-amqp1.0#routing-and-addressing
- https://www.rabbitmq.com/shovel-dynamic.html

I'm quoting:
> Routing key when using dest-exchange. If this is not set, the original message's routing key will be used.

So, as per my understanding, both links up there combined seem to indicate that the AMQP 1.0 subject is mapped to the routing key in AMQP 0-9-1.
The original shovel behaviour is untouched: If a `dest-exchange-key` is set in the shovel configuration, this static key takes precedence over the AMQP 1.0 subject, I just double checked that.

Other shovel users seem to have exactly this requirement (https://groups.google.com/g/rabbitmq-users/c/3n0lNVmVbNo).

However, I'll gladly change PR, please advise then further how the prospective configuration might look like. Obviously, it doesn't make too much sense to ask for a `dest-exchange-key` and then further if the user wishes to use the AMQP 1.0 subject instead.

kjnilsson:
Our main issue is that the use of `subject` may not necessarily be a routing key as this is a just a convention that the RabbitMQ AMQP 1.0 plugin uses. AMQP 1.0 doesn't have a concept of routing keys and the subject for any given AMQP 1.0 source could be used for other things which is why the use of this as a routing key should be something that is configured explicitly. I suggest you try something like my configuration suggestion where the routing key on the AMQP 1.0 source can be taking from any accessible bit of AMQP 1.0 message meta data and that you also add the option to map the exchange. Then if these values are configured and set in the source message they are used - else we fall back to the `dest-*` configuration.

Shigawire:
Ahh, I get that now. Thank you so much for the clarification. I will make some arrangements to the code. As I'm unfamiliar with Erlang, this make take a little while and may not adhere to your code standards, but I'll give it a try and open a new PR then. Thank you!

PR #2885, opened 2021-03-11
Avoid network calls

ansd:
by flipping the two list comprehension conditions.
If not `is_local_to_node`, then `is_down` will not be evaluated.
This saves (R-1) * Q network calls every 2 minutes where R is the number
of replicas per quorum queue and Q is the number of quorum queues in the
RabbitMQ cluster.


michaelklishin (approved):

michaelklishin:
Good catch!

michaelklishin:
Backported to `v3.8.x`.

PR #2886, opened 2021-03-12
MQTT: correct a typo in mqtt_machine

michaelklishin:
Introduced in #2861.


michaelklishin:
Did a PR so that we won't forget about this if/when we backport #2861.

PR #2890, opened 2021-03-12
Make 'rabbitmq-queues await_online_quorum_plus_one' a no-op for single node clusters

michaelklishin:
Since it does not make sense in that case. We previously did the same thing for e.g. `rabbitmq-diagnostics check_if_node_is_quorum_critical`.

Per discussion with @ferozjilla @ansd.


michaelklishin:
Backported to `v3.8.x`.

PR #2891, opened 2021-03-13
STOMP: adapt Python test suite to stomp.py 6.x

michaelklishin:
We are two major versions behind, so I decided to invest some time into upgrading.


michaelklishin:
Backported to `v3.8.x`.

PR #2906, opened 2021-03-18
WIP: Limit the number of messages in q3 for lazy queues

lhoguin:
See *Further comments* for open questions and considerations.

This commit introduces min and max options used when converting
messages from deltas (full on-disk) to q3 (index or full in memory).

Before this commit there could be up to 16k messages read at once
and sitting in memory (the full index segment). This has been
causing issues depending on the number of queues and the rate
of consumption.

With this commit the new default depends on the consumption rate.
The formula applied is 1.5*RateOut so we read from disk less than
once per second. If there is capacity then we may still read up
to 16k messages. If the rate is very low, then we have another
"minimum" default value of 100. So we will always read at least
100 (less if there are less than 100 messages in the queue) and
at most 16k depending on the consumption rate.

The values can be configured so that if there is a large number
of queues we can limit the amount of messages in memory to a
fixed value (for example, set min to 50 and max to 500). This
allows greater control over the memory usage of lazy queues.

## Types of Changes

## Checklist

- [x] All tests pass locally with my changes (*at least those that don't hit timetrap timeout on my machine*)

## Further Comments

Are the configuration key names good?

I have benchmarked this in a variety of scenarios to confirm this does not result in a performance degradation when few queues are used. The system seems to behave like before.

The process memory seems a little better but it's hard to tell because of other factors that I will investigate next (there can be a lot of messages in `disk_pending_ack` - thousands).

A handy way of finding how much is in delta or q3 is the following snippet:

``` erlang
{ok, Q} = rabbit_amqqueue:lookup(rabbit_misc:r(<<"/">>, queue, <<"lazy-1">>)).
QPid = rabbit_amqqueue:pid_of(Q).
fun() -> St = element(3,element(6,element(4,sys:get_state(QPid)))), [{delta, element(3,element(4,St))}, {q3, element(1,element(5,St))}, {ram_pending_ack, element(1,element(8,St))}, {disk_pending_ack, element(1,element(9,St))}] end().
```

It's possible that default classic queues would benefit from this as well, but I have not tested it with them, which is why it's restricted to lazy queues for now.

### DIFF
--- a/deps/rabbit/src/rabbit_variable_queue.erl
+++ b/deps/rabbit/src/rabbit_variable_queue.erl
@@ -2605,6 +2605,7 @@ maybe_deltas_to_betas(_DelsAndAcksFun,
     State;
 maybe_deltas_to_betas(DelsAndAcksFun,
                       State = #vqstate {
+                        mode                 = Mode,
                         q2                   = Q2,
                         delta                = Delta,
                         q3                   = Q3,
@@ -2613,14 +2614,33 @@ maybe_deltas_to_betas(DelsAndAcksFun,
                         ram_bytes            = RamBytes,
                         disk_read_count      = DiskReadCount,
                         delta_transient_bytes = DeltaTransientBytes,
-                        transient_threshold  = TransientThreshold }) ->
+                        transient_threshold  = TransientThreshold,
+                        rates                = #rates{out = RatesOut }}) ->
     #delta { start_seq_id = DeltaSeqId,
              count        = DeltaCount,
              transient    = Transient,
              end_seq_id   = DeltaSeqIdEnd } = Delta,
+    %% We limit the number of messages we read from the index. We want a large
+    %% enough value but not one that will use too much memory because when there
+    %% are many queues we may end up exploding the memory. We can use the current
+    %% rate to determine the upper value unless configured otherwise. If the rate
+    %% is too low, the min value will be used instead, and it defaults to 100
+    %% as that's neither too low nor too high.
+    %%
+    %% Note that the max ever fetched will be ?SEGMENT_ENTRY_COUNT which currently
+    %% defaults to 16k. So configuring the max to a higher value than that will
+    %% not have an effect (effectively disabling the limit).
+    %%
+    %% We currently only do this for lazy queues because it has not been tested
+    %% with default queues.
+    DeltaConfigEndList = case Mode of
+        lazy -> [DeltaSeqId + max(application:get_env(rabbit, lazy_queue_min_deltas_to_betas, 100),
+                                  application:get_env(rabbit, lazy_queue_max_deltas_to_betas, floor(RatesOut * 1.5)))];
+        default -> []
+    end,
     DeltaSeqId1 =
         lists:min([rabbit_queue_index:next_segment_boundary(DeltaSeqId),
-                   DeltaSeqIdEnd]),
+                   DeltaSeqIdEnd|DeltaConfigEndList]),
     {List, IndexState1} = rabbit_queue_index:read(DeltaSeqId, DeltaSeqId1,
                                                   IndexState),
     {Q3a, RamCountsInc, RamBytesInc, State1, TransientCount, TransientBytes} =


lhoguin:
This is the memory usage of the node before/after applying this patch:

Screenshot 2021-03-19 at 13 05 53

The patch was applied around 13:08. It then took some time for memory to stabilize. Before the patch it was reaching the memory high watermark at 2.2GB; after the patch that stopped.

The heap memory also looks better:

Screenshot 2021-03-19 at 13 06 13

lhoguin:
By setting `fullsweep_after` to 0 the memory is even better. On the left RabbitMQ has both this PR and queues have `fullsweep_after`to 0. On the right this is RabbitMQ master from a few days ago:

Screenshot 2021-03-19 at 13 05 53

Screenshot 2021-03-19 at 13 06 13

I am working on a PR to Erlang/OTP to allow setting `fullsweep_after` dynamically. We can then enable it when there's memory pressure for example, or depending on configuration.

lhoguin:
Note that the test done above is very nice to current RabbitMQ because queues will have at most 500 messages in memory, it would be much more obvious if it was designed to make current RabbitMQ consume as much memory as possible (up to 16k in memory). I can probably try that later.

deps/rabbit/src/rabbit_variable_queue.erl
```diff
@@ -2613,14 +2614,33 @@ maybe_deltas_to_betas(DelsAndAcksFun,
...
+    %% as that's neither too low nor too high.
+    %%
+    %% Note that the max ever fetched will be ?SEGMENT_ENTRY_COUNT which currently
+    %% defaults to 16k. So configuring the max to a higher value than that will
+    %% not have an effect (effectively disabling the limit).
+    %%
+    %% We currently only do this for lazy queues because it has not been tested
+    %% with default queues.
+    DeltaConfigEndList = case Mode of
+        lazy -> [DeltaSeqId + max(application:get_env(rabbit, lazy_queue_min_deltas_to_betas, 100),
```
michaelklishin, line 2637: All calls to `application:get_env/2` go through a single process in OTP, and it's easy to overwhelm. Should we cache these in process state?

michaelklishin (changes requested):

deps/rabbit/src/rabbit_variable_queue.erl
```diff
@@ -2613,14 +2614,33 @@ maybe_deltas_to_betas(DelsAndAcksFun,
...
+                                  application:get_env(rabbit, lazy_queue_max_deltas_to_betas, floor(RatesOut * 1.5)))];
```
gerhard, line 2638: What was the reasoning behind the multiplication factor, `1.5` ?

michaelklishin: My guess would be that rates change all the time and we would like to be a little ahead of the delivery rate.

deps/rabbit/src/rabbit_variable_queue.erl
```diff
@@ -2613,14 +2614,33 @@ maybe_deltas_to_betas(DelsAndAcksFun,
...
-                        transient_threshold  = TransientThreshold }) ->
+                        transient_threshold  = TransientThreshold,
+                        rates                = #rates{out = RatesOut }}) ->
     #delta { start_seq_id = DeltaSeqId,
              count        = DeltaCount,
              transient    = Transient,
              end_seq_id   = DeltaSeqIdEnd } = Delta,
+    %% We limit the number of messages we read from the index. We want a large
+    %% enough value but not one that will use too much memory because when there
+    %% are many queues we may end up exploding the memory. We can use the current
```
gerhard, line 2625: How does a large enough value help?

When you write `exploding the memory`, do you mean that we may keep too many messages in memory unnecessarily, and not be memory frugal, as lazy queues are expected to be?

deps/rabbit/src/rabbit_variable_queue.erl
```diff
@@ -2613,14 +2614,33 @@ maybe_deltas_to_betas(DelsAndAcksFun,
...
+    %% rate to determine the upper value unless configured otherwise. If the rate
+    %% is too low, the min value will be used instead, and it defaults to 100
+    %% as that's neither too low nor too high.
```
gerhard, line 2628: Does the 100 value result in different behaviour if message payloads vary? For example, how does the lazy queue behave for 1KB messages? What about 1MB?

gerhard:
30% savings in memory are impressive! 👏🏻 

`fullsweep_after = 0` results in a more volatile memory usage - I am looking at the red RSS line which goes up & down in larger steps (as does eheap_alloc). Considering this, I am not sure that I would want to have this setting in my cluster. The 2GB stable & predictable feels healthier than the following 800MB - 2200MB range.

It would be great to see this patch in the context of https://github.com/rabbitmq/rabbitmq-server/discussions/2785

michaelklishin:
I agree that `fullsweep_after` set to zero will likely introduce volatility that won't be well-received. At the very least this begets a lot of experimentation before we adopt it.

deps/rabbit/src/rabbit_variable_queue.erl
lhoguin: No they don't, they haven't for a very long time.

https://github.com/erlang/otp/blob/master/lib/kernel/src/application_controller.erl#L334

deps/rabbit/src/rabbit_variable_queue.erl
lhoguin: Fetch from disk less than once per second. There might be a better default but it would take quite a bit of time to test many different values under different scenarios.

lhoguin: Adding details: because the function is used when fetching/prefetching, it makes sense to base it on outgoing rates, and the 1.5 is arbitrary as explained in previous comment.

deps/rabbit/src/rabbit_variable_queue.erl
lhoguin: It's about not reading every time a message must be accessed. Note that some operations will end up needing multiple messages at once (dropping expired messages, fetching and prefetching, for example). Before, those operations could end up reading 16k messages which were not necessarily used afterwards. Now we read potentially much less. The function is called again if more messages are needed.

deps/rabbit/src/rabbit_variable_queue.erl
lhoguin: Only the index is read into memory.

lhoguin:
There is going to be a lot of additional experiments for `fullsweep_after`.

Note that the problems might not be directly related, in this case I applied `fullsweep_after` **dynamically** in an otherwise already running node with lots of messages. I had no such problem when applying `fullsweep_after` **on spawn**, in that case memory use was completely flat.

The dynamic `fullsweep_after` will not be in OTP before 24. On the other hand, having a configuration value to set it on spawn might be good to have before then.

deps/rabbit/src/rabbit_variable_queue.erl
gerhard: So `floor(RatesOut * 1.5)` stands for "less than once per second"?

I would very much like to see this moved into a function that captures this. Nothing fancy, `less_than_once_per_second(RatesOut)` is sufficient.

deps/rabbit/src/rabbit_variable_queue.erl
gerhard: Got it 👍🏻

gerhard:
> There is going to be a lot of additional experiments for `fullsweep_after`.

💯 

> having a configuration value to set it on spawn might be good to have before then.

Great idea!

lhoguin:
I think the `fullsweep_after=0` problem comes from times when the queues need to do "heavy" operations (such as dropping expired messages) which may suffer from contention (10 dirty I/O schedulers for 500 queues) and result in a lot of GC and therefore a lot of memory copied every time (which produces these peaks because there's 500 queues). With a dynamic `fullsweep_after` approach we could simply set `fullsweep_after` back to a more appropriate value for the times when we do those operations.

I will try that tomorrow. Off-topic anyway. Don't let talks of `fullsweep_after` distract you, it's not necessary for this PR.

deps/rabbit/src/rabbit_variable_queue.erl
lhoguin: Happy to make that clearer if there's no objection to the value being what it is.

lhoguin:
Quick update: I will most likely retract this PR in favor of a PR that allows using smaller index segment files. While this PR is useful to reduce the memory footprint of the *state*, smaller index segment files reduce the memory footprint of the *process* which is what we want. I will keep this PR open only if I can see a benefit when used in combination with smaller index segment files.

lhoguin:
Closing in favor of a much better patch, https://github.com/rabbitmq/rabbitmq-server/pull/2939

The code in this PR might still be useful down the line when we have a better index, but it brings no noticeable improvement in the near future.

PR #2911, opened 2021-03-19
rabbit_prelaunch_logging: $RABBITMQ_LOGS doesn't override log level if it is set in the configuration file

dumbbell:
Here is an example of that use case:
*   The official Docker image sets `RABBITMQ_LOGS=-` in the environment
*   A user of that image adds a configuration file with:
    ```
    log.console.level = debug
    ```

The initial implementation, introduced in rabbitmq/rabbitmq-server#2861, considered that if the output is overriden in the environment (through `$RABBITMQ_LOGS`), any output configuration in the configuration file is ignored.

The problem is that the output-specific configuration could also set the log level which is not changed by `$RABBITMQ_LOGS`. This patch fixes that by keeping the log level from the configuration (if it is set obviously) even if the output is overridden in the environment.

### DIFF
--- a/deps/rabbit/src/rabbit_prelaunch_logging.erl
+++ b/deps/rabbit/src/rabbit_prelaunch_logging.erl
@@ -680,7 +680,10 @@ handle_default_main_output(
     Overridden = rabbit_env:has_var_been_overridden(Context, main_log_file),
     Outputs1 = if
                    NoOutputsConfigured orelse Overridden ->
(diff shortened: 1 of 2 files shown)


ansd (approved):
Tested that with this PR `log.console.level = debug` outputs debug logs on K8s.
Thanks @dumbbell!

dumbbell:
Thank you @ansd! Merged.

PR #2925, opened 2021-03-24
Revisit what drain and revive do when their feature flag is not enabled

michaelklishin:
If maintenance mode feature flag is not enable, drain and revive should return an error.


michaelklishin:
Backported to `v3.8.x`.

PR #2940, opened 2021-03-29
Logging: Add journald support

dumbbell:
The implementation depends on [erlang-systemd](https://github.com/rabbitmq/erlang-systemd) which uses Unix socket support introduced in Erlang 19. Therefore it doesn't rely on a native library. We also don't need special handling if the host doesn't use journald.

To enable the journald handler, add the following configuration variable:
```
log.journald = true
```

The log level can also be set the same way it is with other handlers:
```
log.journald.level = debug
```

The log messages are communicated to journald using structured data. It is possible to configure which fields are transmitted and how they are named:
```
log.journald.fields = SYSLOG_IDENTIFIER="rabbitmq-server" syslog_timestamp syslog_pid priority ERL_PID=pid
```

In this example:
  * the `SYSLOG_IDENTIFIER` is set to a string literal
  * `syslog_timestamp and `syslog_pid` are aliases for SYSLOG_TIMESTAMP=time and SYSLOG_PID=os_pid
  * `priority` is a special field computed from the log level
  * `ERL_PID=pid` indicates `pid` should be sent as the `ERL_PID` field.

The message itself is implicit and always sent. Otherwise, the list of fields must be exhaustive: fields which are unset in a particular log event meta are sent as an empty string and non-mentionned fields are not sent. The order is not important.

Here are some messages printed by `journalctl -f` during RabbitMQ startup:
```
Mar 26 11:58:31 ip-172-31-43-179 rabbitmq-server[19286]: Ready to start client connection listeners
Mar 26 11:58:31 ip-172-31-43-179 rabbitmq-server[19286]: started TCP listener on [::]:5672
Mar 26 11:58:31 ip-172-31-43-179 rabbitmq-server[19286]: Server startup complete; 0 plugins started.
```

### DIFF
--- a/deps/rabbit/Makefile
+++ b/deps/rabbit/Makefile
@@ -133,13 +133,15 @@ APPS_DIR := $(CURDIR)/apps
 LOCAL_DEPS = sasl rabbitmq_prelaunch os_mon inets compiler public_key crypto ssl syntax_tools xmerl
 
 BUILD_DEPS = rabbitmq_cli
-DEPS = ranch rabbit_common ra sysmon_handler stdout_formatter recon observer_cli osiris amqp10_common syslog
+DEPS = ranch rabbit_common ra sysmon_handler stdout_formatter recon observer_cli osiris amqp10_common syslog systemd
 TEST_DEPS = rabbitmq_ct_helpers rabbitmq_ct_client_helpers amqp_client meck proper
 
 PLT_APPS += mnesia
 
 dep_syslog = git https://github.com/schlagert/syslog 4.0.0
 dep_osiris = git https://github.com/rabbitmq/osiris master
+# TODO: Use systemd from Hex.pm, once there is a new post-0.6.0 release.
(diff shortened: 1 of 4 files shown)


dumbbell:
Note that erlang-systemd is only used for its journald support in this pull request. I plan to change how we tell systemd RabbitMQ is ready to use this new dependency as well. This will allow us to get rid of the previous methods (`socat` and `sd_notify`). This will come in a separate patch.

dumbbell:
Here is how I tested the patch:

1. I used a Debian Buster VM because journald requires systemd as far as I know.
 
2. To view the logs, I just used the "tail -f" journald equivalent:

    ```
    journalctl -f
    ```
3. I configured RabbitMQ to use both journald and stdout:

    ```
    cat >/tmp/rabbitmq.conf <<EOF
    log.console = true
    log.journald = true
    EOF
    ```

4. I started RabbitMQ with that configuration file:

    ```
    gmake run-broker RABBITMQ_CONFIG_FILES=/tmp/rabbitmq.conf
    ```

PR #2945, opened 2021-04-01
Build & push container image on every commit

gerhard:
Before this, someone would need to run the package-generic-unix and docker-image make targets in order to produce a container image. This wouldn't always work: more than one archive in the PACKAGES dir, VERSION & RABBITMQ_VERSION had to be specified, + characters in VERSION. It also required me to have Docker running, which I am reluctant to leave running on macOS because it uses CPU even when idle, and starting it every time that I wanted an ad-hoc container image implied waiting a few minutes, and then waiting another 15 minutes for the container image to be produced and published to hub.docker.com.

This commit delegates all the above to GitHub Actions. The best part is that it happens asynchronously - every commit triggers it - and a container image for that commit will appear on hub.docker.com within ~15 minutes. In other words, anyone can test any commit in their container runtime within 20 minutes, without needing to lift a finger.  No Docker, no make targets, just reference the new tag (matches the git sha) and off you go. It's the beginning of something better, for sure.

This is what that looks like on https://hub.docker.com/r/pivotalrabbitmq/rabbitmq/tags?page=1&ordering=last_updated
image

For the first one that reviews this: does this work for you?
```
docker run -it --rm --network host pivotalrabbitmq/rabbitmq:f7a36f9f87853077b49a4bc54e23a24c71f9384c
```

If you are reviewing this on a Mac, try this docker one-liner instead ([this explains why](https://docs.docker.com/docker-for-mac/networking/)):

```
docker run -it --rm -p 5672:5672 -p 15672:15672 pivotalrabbitmq/rabbitmq:f7a36f9f87853077b49a4bc54e23a24c71f9384c
```

Should be added to `v3.8.x` as-is after merge.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/.github/workflows/oci.yaml
+++ b/.github/workflows/oci.yaml
@@ -1,0 +1,53 @@
+# https://github.com/marketplace/actions/build-and-push-docker-images
+name: OCI
+on:
+  push:
+    paths:
+      - 'deps/**'
+      - 'escript/**'
+      - 'packaging/**'
+      - 'scripts/**'
+      - Makefile
+      - plugins.mk
+      - rabbitmq-components.mk
+env:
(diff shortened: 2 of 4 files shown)


kjnilsson:
Backported to [v3.8.x](https://github.com/rabbitmq/rabbitmq-server/commit/cb4c1f8a8c760a787dfdf567373addafb08b29bf)

PR #2952, opened 2021-04-05
Support rabbit_peer_discovery_aws to work with instance metadata serv…

thuandb:
…ice v2 (IMDSv2).

IMDSv2 uses session-oriented requests. With session-oriented requests, a session token is retrieved first
then used in subsequent GET requests for instance metadata values such as instance-id, credentials, etc.

Details could be found here https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

## Proposed Changes

The following change is an implementation to the support aws_peer_discover_aws plugin to work with EC2 instance metadata service version 2 (IMDSv2). IMDSv2 is a new version of the instance metadata service which adds defense in depth against open firewalls, reverse proxies, and SSRF vulnerabilities. With IMDSv2, every request submitted to the instance metadata service is now protected by session authentication. Specifically, querying IMDSv2 is done in 2 steps:

* Start a session with an HTTP PUT request to IMDSv2. IMDSv2 then returns a secret token.
* Use the secret token to make any HTTP GET request to retrieve interested metadata values such as availability zone, instance role, instance ID, credentials, etc.

This change is backward compatible. If for some reason, it fails to obtain a secret token from IMDSv2, rabbit_peer_discovery_aws plugin will fallback to use IMDSv1. 

In addition, a new configuration flag `aws_prefer_imdsv2` is also introduced to allow users explicitly enable or disable using IMDSv1. Per AWS security recommendation, IMDSv2 is preferable. Specifically, if the configuration flag a new configuration flag: `aws_prefer_imdsv2` is not set or set to true, the rabbitmq_peer_discovery_aws plugin will attempt to retrieve a secret token first. If a secret token is received successfully, it will be used in subsequent requests submitted to the instance metadata service. On the other hand, if  `aws_prefer_imdsv2` is set to false, the rabbitmq_peer_discovery_aws plugin will use IMDSv1 for instance metadata queries. 

## Types of Changes

## Checklist

- [x] I have tested this change with AWS and with both IMDSv2 and failure scenario to fallback to IMDSv1 as well as explicitly turn off IMDSv2 preference.

## Further Comments

IMDSv2 secret token retrieval result (both success and failure) is cached and reused. This is to guarantee that there is at most 1 HTTP PUT request is used to request IMDSv2 secret token per session which included multiple HTTP GET requests to query metadata values: instance-id, availability zone, instance role, and credentials.

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -1733,6 +1733,19 @@ end}.
     end
 }.
 
+
+% ===============================
+% AWS section
+% ===============================
+
+%% Whether or not to prefer IMDSv2 when querying instance metadata service (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html).
+%% If not set or set to true, IMDSv2 will be preferred to use first. If fails, IMDSv1 will be used.
+%% {aws_prefer_imdsv2, false}
+
+{mapping, "aws_prefer_imdsv2", "rabbit.aws_prefer_imdsv2",
+    [{datatype, {enum, [true, false]}}]}.
+
+
 % ===============================
 % Validators
 % ===============================
--- a/deps/rabbitmq_aws/README.md
+++ b/deps/rabbitmq_aws/README.md
@@ -54,6 +54,8 @@ configuration or to impact configuration behavior:
  ``rabbitmq_aws:set_region/1``          | Manually specify the AWS region to make requests to.
  ``rabbitmq_aws:set_credentials/2``     | Manually specify the request credentials to use.
  ``rabbitmq_aws:refresh_credentials/0`` | Refresh the credentials from the environment, filesystem, or EC2 Instance Metadata service.
+ ``rabbitmq_aws:ensure_imdsv2_token_valid/0`` | Make sure IMDSv2 token is acctive and valid.
+ ``rabbitmq_aws:api_get_request/2``           | Perform an AWS service API request.
  ``rabbitmq_aws:get/2``                 | Perform a GET request to the API specifying the service and request path.
  ``rabbitmq_aws:get/3``                 | Perform a GET request specifying the service, path, and headers.
  ``rabbitmq_aws:post/4``                | Perform a POST request specifying the service, path, headers, and body.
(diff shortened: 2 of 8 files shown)


thuandb:
Submitted by accident. I will need to do some clean up and resubmit.

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -1733,6 +1733,19 @@ end}.
...
+
+% ===============================
+% AWS section
+% ===============================
+
+%% Whether or not to prefer IMDSv2 when querying instance metadata service (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html).
+%% If not set or set to true, IMDSv2 will be preferred to use first. If fails, IMDSv1 will be used.
+%% {aws_prefer_imdsv2, false}
+
+{mapping, "aws_prefer_imdsv2", "rabbit.aws_prefer_imdsv2",
```
michaelklishin, line 1745: Can we namespace this the same way [other AWS peer discovery keys](https://www.rabbitmq.com/cluster-formation.html#peer-discovery-aws) are namespaced? For example, `cluster_formation.aws.prefer_imdsv2` instead of `aws_prefer_imdsv2`.

thuandb: Hi Micheal, thank you for your quick feedback. 

I was thinking that this new `aws_prefer_imdsv2` flag is used in the generic [`rabbitmq_aws` application](https://github.com/rabbitmq/rabbitmq-server/tree/master/deps/rabbitmq_aws) which the `rabbit_peer_discovery_aws` plugin relies on. It is independent from peer discovery plugins and/or cluster formation. 

For example, in future there might be new plugins need interaction with AWS for other purposes such as publishing metrics/logs directly to CloudWatch, and It could be run from a single instance RabbitMQ fleets and/or clusters may not use the `rabbit_peer_discovery_aws` plugin.  

Please feel free to suggest a better namespace for this new flag. And in case `cluster_formation.aws.prefer_imdsv2` still better, please let me know. 

P/S: I accidentally submitted this PR request while preparing a private one for being reviewed internally first, hence I have closed it. I will submit a new PR request once it is approved internally and will make sure your comments here is addressed.

michaelklishin: I would not object to `aws.prefer_imdsv2` or `amws.preferred_imds = v2` or similar. I have a slight preference towards the latter but recognize that many tools use settings such as `prefer X` which is a boolean (e.g. the IPv6 vs. IPv4 preference on the JVM).

thuandb: Sure. I will change to `aws.prefer_imdsv2 = true | false` then. Thanks again.

PR #2953, opened 2021-04-07
AMQP 1.0 binary parser: treat arrays with extra or missing input as fatal errors

michaelklishin:
## Proposed Changes

With some input it is possible that the terminating clause will never
match.

While at it, consume binary input when parsing short form
primitives: null, true, false, as well as uint/ulong zero values.

## Types of Changes

## Checklist

## Further Comments

Pair: @lhoguin.


lhoguin (approved):
Approved based on re-reading of the changes. I have no tested them on my machine.

deps/amqp10_common/src/amqp10_binary_parser.erl
```diff
@@ -122,6 +122,14 @@ parse_compound1(Count, Bin, Acc) ->
     {Value, Rest} = parse(Bin),
     parse_compound1(Count - 1, Rest, [Value | Acc]).
 
+parse_array_primitive(16#40, <<_:8/unsigned, R/binary>>) -> {null, R};
+parse_array_primitive(16#41, <<_:8/unsigned, R/binary>>) -> {true, R};
+parse_array_primitive(16#42, <<_:8/unsigned, R/binary>>) -> {false, R};
+parse_array_primitive(16#43, <<_:8/unsigned, R/binary>>) -> {{uint, 0}, R};
+parse_array_primitive(16#44, <<_:8/unsigned, R/binary>>) -> {{ulong, 0}, R};
+parse_array_primitive(Marker, Data) ->
```
kjnilsson, line 130: Maybe use 'Type' here instead of 'Marker' for consistency?

kjnilsson (approved):
Just one minor naming comment

deps/amqp10_common/src/amqp10_binary_parser.erl
michaelklishin: Done

lhoguin:
Reproduced the issue before the patch and confirmed that the patch fixes it.

michaelklishin:
Backported to `v3.8.x`.

PR #2956, opened 2021-04-07
Support rabbit_peer_discovery_aws to work with EC2 IMDSv2

thuandb:
Hi, I’m from the Amazon MQ team in AWS. We are proposing the following change to prefer using the EC2 Instance Metadata Service v2 based on [AWS security recommendations](https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/).

## Proposed Changes

The following change is an implementation to support the `rabbit_peer_discovery_aws` plugin to work with EC2 instance metadata service version 2 (IMDSv2). IMDSv2 is a new version of the instance metadata service which adds defense in depth against open firewalls, reverse proxies, and SSRF vulnerabilities. With IMDSv2, every request submitted to the instance metadata service is now protected by session authentication. Specifically, querying IMDSv2 is done in 2 steps:

* Start a session with an HTTP PUT request to IMDSv2. IMDSv2 then returns a secret token.
* Use the secret token to make any HTTP GET request to retrieve interested metadata values such as availability zone, instance role, instance ID, credentials, etc.

Both IMDSv1 and IMDSv2 are available by default on all EC2 instances. This change is backward compatible. If for some reason, it fails to obtain a secret token from IMDSv2, `rabbit_peer_discovery_aws` plugin will automatically fallback to use IMDSv1. 

[Per AWS security recommendation, IMDSv2 is preferable](https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/). A new configuration flag `aws.prefer_imdsv2` is introduced to prefer using IMDSv2. Specifically, if the configuration flag `aws.prefer_imdsv2` is set to true (which is also the default value), the `rabbitmq_peer_discovery_aws` plugin will attempt to use IMDSv2 by retrieving a secret token. If a secret token is received successfully, it will be used in subsequent requests submitted to the Instance Metadata Service. If a secret token cannot be obtained, then subsequent calls will be made to IMDSv1. On the other hand, if  `aws.prefer_imdsv2` is set to false, the `rabbitmq_peer_discovery_aws` plugin will only use IMDSv1 for instance metadata queries. 

### Notes

IMDSv2 secret token retrieval result (both success and failure) is cached and reused. This is to guarantee that there is at most 1 HTTP PUT request is used to request IMDSv2 secret token per session which included multiple HTTP GET requests to query metadata values: instance-id, availability zone, instance role, and credentials. 

## Types of Changes

## Checklist

- [x] I have tested this change on AWS.

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -1739,7 +1739,9 @@ end}.
 % ===============================
 
 %% @doc Whether or not to prefer EC2 IMDSv2 when querying instance metadata service.
-%% If not set or set to true, EC2 IMDSv2 will be preferred to use first. If fails, IEC2 MDSv1 will be used.
+%% The setting defaults to true. When true, instance metadata will first be attempted using EC2 IMDSv2
+%% and will fall back to EC2 IMDSv1 upon failure.
+%% When false, EC2 IMDSv1 will be used first and no attempt will be made to EC2 IMDSv2..
 %% See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html.
 
 {mapping, "aws.prefer_imdsv2", "rabbit.aws_prefer_imdsv2",
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -1742,7 +1742,7 @@ end}.
 %% If not set or set to true, IMDSv2 will be preferred to use first. If fails, IMDSv1 will be used.
 %% {aws_prefer_imdsv2, false}
 
-{mapping, "aws_prefer_imdsv2", "rabbit.aws_prefer_imdsv2",
+{mapping, "aws.prefer_imdsv2", "rabbit.aws_prefer_imdsv2",
     [{datatype, {enum, [true, false]}}]}.
 
 
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -1741,7 +1741,7 @@ end}.
 %% @doc Whether or not to prefer EC2 IMDSv2 when querying instance metadata service.
 %% The setting defaults to true. When true, instance metadata will first be attempted using EC2 IMDSv2
 %% and will fall back to EC2 IMDSv1 upon failure.
-%% When false, EC2 IMDSv1 will be used first and no attempt will be made to EC2 IMDSv2..
+%% When false, EC2 IMDSv1 will be used first and no attempt will be made to EC2 IMDSv2.
 %% See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html.
 
 {mapping, "aws.prefer_imdsv2", "rabbit.aws_prefer_imdsv2",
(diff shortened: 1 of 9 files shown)


michaelklishin (changes requested):
We generally don't reject pull requests over things like this but in this PR, multiple lines
do not use any whitespace when binding or destructuring:

``` erl
{ok, State}=gen_server:call(rabbitmq_aws, get_state)
```

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -1733,6 +1733,21 @@ end}.
...
+% AWS section
+% ===============================
+
+%% @doc Whether or not to prefer EC2 IMDSv2 when querying instance metadata service.
+%% The setting defaults to true. When true, instance metadata will first be attempted using EC2 IMDSv2
+%% and will fall back to EC2 IMDSv1 upon failure.
+%% When false, EC2 IMDSv1 will be used first and no attempt will be made to EC2 IMDSv2..
+%% See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html.
+
+{mapping, "aws.prefer_imdsv2", "rabbit.aws_prefer_imdsv2",
```
michaelklishin, line 1747: This arguably belongs to `rabbitmq_aws` which can have its own Cuttlefish schema file. Nothing in RabbitMQ uses AWS or is aware of it.

michaelklishin: I should perhaps clarify that RabbitMQ will discover all Cuttlefish schema files under `priv/schema` in every plugin on the code path, so I can't think of a reason why this can't be moved to `rabbitmq_aws`.

See [`rabbit_prelaunch_conf`](https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_conf.erl#L274-L292) if you are curious how schema file discovery is performed. TL;DR: we simply scan `priv` directories of all applications on the code path :)

deps/rabbitmq_aws/src/rabbitmq_aws_config.erl
```diff
@@ -628,6 +658,23 @@ perform_http_get(URL) ->
...
+  rabbit_log:debug("Querying instance metadata service: ~p", [URL]),
+  httpc:request(get, {URL, instance_metadata_request_headers()},
+    [{timeout, ?DEFAULT_HTTP_TIMEOUT}], []).
+
+-spec get_instruction_on_instance_metadata_error(string()) -> string().
+%% @doc Return error message on failures related to instance metadata service with a reference to AWS document.
+%% end
+get_instruction_on_instance_metadata_error(ErrorMessage) ->
+  ErrorMessage ++
+  " Please refer to the AWS documentation for details on how to configure the instance metadata service: "
```
michaelklishin, line 674: This is a good idea to log a link to a doc guide 👍

deps/rabbit/priv/schema/rabbit.schema
thuandb: Thank you for suggestion. I was afraid of having to add a new schema parser to `rabbitmq_aws` hence tried to reuse `rabbit.schema`. However, per your feedback, I've moved this AWS specific configuration flag to a new file`rabbitmq_aws.schema` in the `rabbitmq_aws/priv/schema` folder.

michaelklishin:
Follow-up PR: #2959.

michaelklishin:
Backported to `v3.8.x`.

PR #2958, opened 2021-04-08
DO NOT MERGE Only generate guid for classic queues

kjnilsson:
As streams and quorum queues do not need this guid and it creates
unnecessary work and disk use.

### DIFF
--- a/deps/rabbit/src/rabbit_basic.erl
+++ b/deps/rabbit/src/rabbit_basic.erl
@@ -136,7 +136,7 @@ message(XName, RoutingKey, #content{properties = Props} = DecodedContent) ->
         {ok, #basic_message{
            exchange_name = XName,
            content       = strip_header(DecodedContent, ?DELETED_HEADER),
-           id            = rabbit_guid:gen(),
+           % id            = rabbit_guid:gen(),
            is_persistent = is_message_persistent(DecodedContent),
            routing_keys  = [RoutingKey |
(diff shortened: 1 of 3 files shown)


michaelklishin (changes requested):
Some tests in `queue_parallel_SUITE` fail on this branch, possibly some tests bypass the relevant code paths where the guid is generated. We will investigate this later, this is not a priority right now.

lhoguin:
In the work I am doing for #3029 I can see a clear path to removing the MsgId use in almost all cases. This ID will still be necessary for messages that do fan-out, because that's how the per-vhost message store works, and there has to be some way of addressing the messages anyway. But when using the per-queue store it becomes unnecessary.

The ID is also used to keep track of confirms, but I think this can be reworked to use something else.

As far as the new index is concerned, this means that 17 bytes will be necessary to store the message location:

- 1 byte `1` and 16 bytes `MsgId` when the per-vhost store is used
- 1 byte `2`, 8 bytes `StoreOffset` (not sure we need the full 8 bytes but it doesn't hurt) and 4 bytes `StoreSize`

Because the index entries are using a fixed size for quick lookup we lose 4 bytes when the per-queue message store is used, but in total we are only using 32 bytes so that's probably fine.

kjnilsson:
Replaced by https://github.com/rabbitmq/rabbitmq-server/pull/4990

PR #2965, opened 2021-04-10
CLI: handle {:relative, val} cases when formatting effective memory watermark value

michaelklishin:
## Proposed Changes

CLI: handle `{:relative, val}` cases when formatting effective memory watermark value.

It can be exposed to CLI tools when relative watermark value is
configured via advanced.config.

I fixed a couple of tiny related things discovered along the way.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #2964)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)

## Checklist

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [x] All tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in related repositories

## Further Comments

Closes #2964.

Issue #2694, opened 2020-12-22
Invalid access check in adding shovel


inikulshin:
Hi,

Here are the logs we get (RabbitMQ 3.7.24, acting user is **5697bfdf-b663-441e-9078-ae9567c11cb6**)

11:18:05 TRC:   Upserting shovel nasuni.imcoming.shovel on vhost / {
11:18:05 INF:    Invoking PUT: http://localhost:15672/api/parameters/shovel/%2f/nasuni.imcoming.shovel with json: {"value":{"dest-exchange":"vsb.topic=nasuni.incoming.ex","dest-uri":"**amqp:///%2f?heartbeat=900**","src-queue":"nasuni.incoming.q","src-uri":"**amqp://VaronisOwner@/nasuni?heartbeat=900**"}}
11:18:05 ERR:    Bad request: Validation failed

user "**5697bfdf-b663-441e-9078-ae9567c11cb6**" may not connect to vhost "nasuni"

I don't understand why is this access check performed at all.

**5697bfdf-b663-441e-9078-ae9567c11cb6** should have access to vhost '/' (to create shovel)
**guest** should have access to vhost '/' (dest-exchange)
**VaronisOwner** should have access to vhost 'nasuni' (src-queue)

I'm not sure the last 2 should be validated during PUT, but AFAIU, the failed access check shouldn't be performed.

PR #2966, opened 2021-04-11
High VM watermark: support {relative, N} values set via advanced.config

michaelklishin:
for usability. It is not any different from when a float value
is used and only exists as a counterpart to '{absolute, N}'.

Also nothing changes for rabbitmq.conf users as that format performs
validation and correct value translation.

See #2694, #2965 for background.

### DIFF
--- a/deps/rabbit/test/unit_vm_memory_monitor_SUITE.erl
+++ b/deps/rabbit/test/unit_vm_memory_monitor_SUITE.erl
@@ -21,7 +21,8 @@ groups() ->
     [
       {sequential_tests, [], [
           parse_line_linux,
-          set_vm_memory_high_watermark_command
+          set_vm_memory_high_watermark_relative1,
+          set_vm_memory_high_watermark_relative2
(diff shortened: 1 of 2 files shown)


eLvErDe:
Thanks a lot, I think it makes configuration more consistent that way :)

michaelklishin:
Follow-up test suite and Dialyzer changes: 19b104a22fc089a00617d171b432a372aeacd7dc, 7f98bc3d1c1baef45a6919af526b48482d8a4111

eLvErDe:
So, may I consider relative keyword is officially supported for disk and memory starting 3.8.15 in my recipes ? :)

michaelklishin:
@eLvErDe I don't have much to add to my recommendation of using `rabbitmq.conf`, which would have never allowed for an unsupported value to be set to begin with. No one should configure these values in `advanced.config`.

eLvErDe:
Got it, I'll see if it's possible easily.

PR #2968, opened 2021-04-12
Allow quorum queue names to exceed atom max chars

kjnilsson:
If the concatenation of the vhost and the queue name exceeds 255 chars
we instead generate an arbitrary atom name instead of throwing an
exception.

Fixes: #2968


michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #2973, opened 2021-04-14
Classic queue mirrors should also continue pushing messages to disk when syncing

luos:
## Proposed Changes

 Hi,

I have two RabbitMQs, full of 2 kb messages, memory limit 1.2 GB, sync mode automatic:

kép

The synchronisation can get stuck because the queue does not reduce its memory usage on the mirror process:

kép

```
(rabbit@rmq2.monitored.host)4> erlang:process_info(<0.1066.0>, messages).
{messages,[emit_gen_server2_stats,bump_reduce_memory_use]}
```

Because it's not handling `bump_reduce_memory_use`, it is only scheduling the next batch to disk when `set_ram_duration_target` is received, as that call is calling `reduce_memory_use`. 

You can see it on this graph that it takes around 10 minutes to synchronise this queue, because it's constantly hitting the watermark, because the queue is not sending messages to disk:

kép

Even after the sync is finished, there I think unexpectedly many messages in memory:

kép

Total memory usage for the mirror:

kép

After the fix, sync takes around 3 minutes:

kép
kép

There are still some messages in memory but significantly lower as the message store actually is sending these bump_reduce_memory_use messages but they were not handled properly:

kép

Interestingly, same test using lazy queues:

kép

Done in under 30 seconds.

Probably `wait_for_resources` should also handle `bump_reduce_memory_usage`, it's in the critical path, basically it does not do anything instead of pushing messages to disk.

I think there are some issues with `set_ram_duration_target` as well as it looks like flip flopping around, even though the calculated target ram messages should be 0, as the queue has no consumers, but it's setting non zero targets.

I've tried to handle set_ram_duration_target in the sync slave process to force the duration to 0, but curiously that did not really help, possibly because wait_for_resources does not handle the bump message.

These resume 'bump_reduce_memory_use' messages were not handled on the mirror, which I think may lead to high memory usage on the queue mirror process, so I added that as well.

Here is the proof of concept fix, let me know what you think. As I said, with this it takes around 3 minutes to sync the queue, which is still pretty far from the lazy queue version. 

## Types of Changes

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_mirror_queue_slave.erl
+++ b/deps/rabbit/src/rabbit_mirror_queue_slave.erl
@@ -385,8 +385,13 @@ handle_info({bump_credit, Msg}, State) ->
     credit_flow:handle_bump_msg(Msg),
     noreply(State);
 
-handle_info(bump_reduce_memory_use, State) ->
-    noreply(State);
+handle_info(bump_reduce_memory_use, State = #state{backing_queue       = BQ,
+                                                backing_queue_state = BQS}) ->
+    BQS1 = BQ:handle_info(bump_reduce_memory_use, BQS),
+    BQS2 = BQ:resume(BQS1),
+    noreply(State#state{
+        backing_queue_state = BQS2
+    });
 
 %% In the event of a short partition during sync we can detect the
 %% master's 'death', drop out of sync, and then receive sync messages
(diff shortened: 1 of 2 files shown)


michaelklishin (approved):

michaelklishin:
Forward ported to `master`.

PR #2976, opened 2021-04-14
QQ: ensure that messages are delivered in order

### LINKED ISSUE #2974: Quorum Queue does not take snapshots for a give configuration
Repro:

Create a qq "q1" with `x-max-length-bytes=1000000000,x-max-in-memory-bytes=1000000`

Run `perftest -p -u q1 -s 10000 -x 1 -vr 100:30 -vr 500:30 -y 1 -L 100000 -q 100 -c 100 -A 5` until it reaches the length limit and let it continue. No further snapshots can be observed the the queue's data directory.

kjnilsson:
In the case where there are some messages kept in memory mixed with
some that are not it is possible that a messages are delivered to the
consuming channel with gaps/out of order which would in some cases cause
the channel to treat them as re-sends it has already seen and just
discard them. When this happens the messages get stuck in the consumer
state inside the queue and are never seen by the client consumer and
thus never acked. When this happen the release cursors can't be emitted
as the smallest raft index will be one of the stuck messages.

Fixes #2974

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -370,7 +370,7 @@ apply(#{index := Index}, #purge{},
     Effects0 = [garbage_collection],
     Reply = {purge, Total},
     {State, _, Effects} = evaluate_limit(Index, false, State0,
-                                            State1, Effects0),
+                                         State1, Effects0),
     update_smallest_raft_index(Index, Reply, State, Effects);
 apply(_Meta, #garbage_collection{}, State) ->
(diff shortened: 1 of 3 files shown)


ansd:
I successfully tested that this PR fixes the issue:
Once the messages are consumed, the segment files get deleted.

gerhard:
This looks great - super team-work @kjnilsson & @ansd! - and the last comment gives me confidence to say just this:

# 🚀

michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #2981, opened 2021-04-16
Report stream coordinator unavailable as an amqp error

dcorbacho:
Uses amqp code 506: resource_error

## Types of Changes

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_channel.erl
+++ b/deps/rabbit/src/rabbit_channel.erl
@@ -2123,30 +2123,28 @@ deliver_to_queues({Delivery = #delivery{message    = Message = #basic_message{
                                         QRef = amqqueue:get_name(Q),
                                         [QRef | Acc]
                                 end, [], Qs),
-    try
-        {ok, QueueStates, Actions} =
-            rabbit_queue_type:deliver(Qs, Delivery, QueueStates0),
-        %% NB: the order here is important since basic.returns must be
-        %% sent before confirms.
-        ok = process_routing_mandatory(Mandatory, Qs, Message, State0),
-        State1 = process_routing_confirm(Confirm, AllQueueNames,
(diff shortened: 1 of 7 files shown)


deps/rabbit/src/rabbit_queue_type.erl
```diff
@@ -533,9 +533,14 @@ get_ctx_with(Q, #?STATE{ctxs = Contexts}, InitState)
             %% not found and no initial state passed - initialize new state
             Mod = amqqueue:get_type(Q),
             Name = amqqueue:get_name(Q),
-            #ctx{module = Mod,
-                 name = Name,
-                 state = Mod:init(Q)};
+            case Mod:init(Q) of
```
kjnilsson, line 536: should init return {ok, State} in the success clause?

deps/rabbit/src/rabbit_channel.erl
```diff
@@ -2123,27 +2123,35 @@ deliver_to_queues({Delivery = #delivery{message    = Message = #basic_message{
...
-        fine ->
-            ?INCR_STATS(exchange_stats, XName, 1, publish),
-            [?INCR_STATS(queue_exchange_stats,
-                         {amqqueue:get_name(Q), XName}, 1, publish)
-             || Q <- Qs];
-        _ ->
-            ok
-    end,
-    State.
+    try
```
kjnilsson, line 2126: Is this try catch here just to catch init errors? Just wondering if queue_type:deliver can return an error instead to make it a bit more portable when we want to use the queue type outside of the channel.

kjnilsson (changes requested):
Mostly wondering if we should tweak the queue type behaviour to return the errors.

kjnilsson (approved):

PR #2982, opened 2021-04-16
Correctly parse tags provided as a list

michaelklishin:
Discovered while testing a PR for rabbit-hole.


michaelklishin:
Backported to `v3.8.x`.

Issue #2983, opened 2021-04-18
ssl_cert_login_san_type = other_name returns a non-standard attribute value


Thibi2000:
Hello,

I've posted about it before [here](https://groups.google.com/g/rabbitmq-users/c/mQ84WxGVgRo), but no on responded, so I am trying
here again with a fix.

# Context

I want to be able to authenticate users with their `otherName` of Subject Alternative
Name of their certificates. This is possible using the following
`rabbitmq.conf`:

    listeners.ssl.default = 5671
    auth_mechanisms.1 = EXTERNAL
    ssl_cert_login_from = subject_alternative_name
    ssl_cert_login_san_type = other_name
    ssl_cert_login_san_index =0
    
    ssl_options.cacertfile = /path/root-cert.pem
    ssl_options.certfile   = /path/server-cert.pem
    ssl_options.keyfile    = /path/server.key
    ssl_options.verify     = verify_peer
    ssl_options.depth  = 1
    ssl_options.fail_if_no_peer_cert = true

To create a test situation, you can use the following commands. It
will create the necessary certificates:

    # self signed root
    openssl req -x509 -newkey rsa:2048 -keyout root-key.pem -out root-cert.pem -days 365 -nodes -subj "/CN=RABBIT TEST ROOT"
    # server cert
    openssl req -new -out server.csr -newkey rsa:2048 -nodes -sha256 -keyout server.key -subj "/CN=server"
    openssl x509 -req -in server.csr -CA root-cert.pem -CAkey root-key.pem -CAcreateserial -out server-cert.pem
    # client cert with SAN
    openssl req -new -out client.csr -newkey rsa:2048 -nodes -sha256 -keyout client.key -subj "/CN=client"
    openssl x509 -req -in client.csr -out client-cert.pem -CA root-cert.pem -CAkey root-key.pem  -CAcreateserial \
            -extensions SAN \
            -extfile <(cat /etc/ssl/openssl.cnf \
                           <(printf "\n[SAN\nsubjectAltName=otherName:1.3.6.1.4.1.54392.5.436;UTF8:username,DNS:username,URI:username"))
    openssl pkcs12 -export -in <(cat root-cert.pem client-cert.pem client.key) -password pass:"verystrong" \
            -name "client" -out "client.p12"
    keytool -import -alias root -file root-cert.pem \
            -keystore truststore -storepass verystrong \
            -noprompt

Make sure the certificates are stored in the correct paths according
to the configuration.

[...]

PR #2985, opened 2021-04-19
Mk decode other name as directory string

michaelklishin:
Decode other name using 'OTP-PUB-KEY':decode/2

and assume it is a string-like value ("directory string")
because other values would not make much sense in the
username extraction context.

References #2983.


michaelklishin:
Using the following `otherName` value

``` ini
otherName = 1.3.6.1.4.1.54392.5.436;FORMAT:UTF8,UTF8String:unicøde-vàlüe
```

I can tell that it is parsed correctly on this branch. It also authenticates when I create a user and grant it some permissions:

``` shell
rabbitmqctl add_user "unicøde-vàlüe"
rabbitmqctl set_permissions -p / "unicøde-vàlüe" ".*" ".*" ".*"
```

I also tested with some Cyrillic and ASCII values ranging from 4 to 390 bytes in length.

michaelklishin:
Backported to `v3.8.x`.

PR #2992, opened 2021-04-21
Capture and log crashes on the stream coordinator machine

dcorbacho:
## Types of Changes

## Checklist


kjnilsson:
I think this is fine. Ideally state machines should be total functions so should never need a try catch. For the coordinator that does so much work in the "real" world we probably should be more defensive. I think we should allow some further testing before we merge this however to make it easier to spot genuine errors.

PR #2993, opened 2021-04-21
Ignore drained nodes when rebalancing

mkuratczyk:
When running `rabbitmq-queues rebalance`, drained nodes (nodes in maintenance mode) should not be considered for queue leader replica placement.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -504,7 +504,7 @@ maybe_migrate(ByNode, MaxQueuesDesired, [N | Nodes]) ->
         [{_, Q, false} = Queue | Queues] = All when length(All) > MaxQueuesDesired ->
             Name = amqqueue:get_name(Q),
             Module = rebalance_module(Q),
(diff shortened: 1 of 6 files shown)


deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -504,7 +504,7 @@ maybe_migrate(ByNode, MaxQueuesDesired, [N | Nodes]) ->
         [{_, Q, false} = Queue | Queues] = All when length(All) > MaxQueuesDesired ->
             Name = amqqueue:get_name(Q),
             Module = rebalance_module(Q),
-            Candidates = Module:get_replicas(Q) -- [N],
+            Candidates = rabbit_maintenance:filter_out_drained_nodes_consistent_read(Module:get_replicas(Q) -- [N]),
```
michaelklishin, line 507: Do we really need a consistent read here? It involves a cluster-wide query. We will do this for every queue and it is really unlikely that the list of nodes under maintenance will change in the middle of the rebalancing operation (in practice).

mkuratczyk: You're right. Changed to dirty

michaelklishin:
Backported to `v3.8.x`.

Issue #2991, opened 2021-04-21
Timestamp is int32 on a received trace message


bollhals:
Copy from https://groups.google.com/g/rabbitmq-users/c/YipXsgD7XpM

We've setup the vhost tracing to queue all published messages into a separate queue. When consuming messages from said queue in dotnet (probably irrelevant), the timestamp of the original message is delivered/read as int32. Shouldn't this be a long/int64?

It seems to me (I don't really know the language, so I can be wrong) that [this line is responsible for the incorrect forwarding](https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit/src/rabbit_trace.erl#L127).

=> This issue prevents us from sending a timestamp with ms resolution instead of s, as the resulting long value is > 32bit value.

michaelklishin:
The timestamp field in AMQP 0-9-1 has a a second resolution, not ms. Which is why https://github.com/rabbitmq/rabbitmq-event-exchange/commit/97a0ac7d3e023688281ae47cc91885226132a3b7 and IIRC a couple of similar changes were introduced. The same can be done to/in tracing.

michaelklishin:
Ah, I forget that we only use message headers in tracing events. So you may be right, it can be a matter of using an unsigned type since timestamps can't be negative.

michaelklishin:
So the type itself is not going to help much. We have to use a new header, `timestamp_in_ms`, much like we do with Shovel.

bollhals:
I think I confused you a bit by the ms topic. My bad.

In the C# client it passes the timestamp as AMQPTimestamp, which is serialized as a long value. 

My company is investigating the possibility to instead sending the timestamp value as ms instead of s. 
=> the actual value passed is basically just 1000x higher, as seen [here](https://www.textmagic.com/free-tools/timestamp-converter) in "Timestamp in milliseconds" instead of "Epoch timestamp"

This causes it currently to overflow due to the int32 size used.

bollhals:
Also for normal message delivery this works flawlessly, as there the value is transmitted as long(64bits) as it was sent.

michaelklishin:
@bollhals OK, that was more or less my understanding. I got a little confused because I used a different client to observe.

I can produce a one-off build of this plugin that uses `long` for value type for you to try with your actual apps by replacing the `rabbitmq_tracing.ez` in your . Would that work? What Erlang version should I compile for?

bollhals:
That would certainly work👍🏼 
[...]

PR #2995, opened 2021-04-21
WIP Tracing: serialize timestamp value as a long

michaelklishin:
I took the opportunity to simplify the message-to-headers conversion. The code may look silly to some but it is immediately
obvious what's going on now and we can special case certain keys, while previously it was a moderately complex `foldl/3`
which said nothing about what the intent is.

The message properties record hasn't changed in years so the benefit of iterating over its fields is negligible.

References #2991.

### DIFF
--- a/deps/rabbit/src/rabbit_trace.erl
+++ b/deps/rabbit/src/rabbit_trace.erl
@@ -105,24 +105,57 @@ trace(X, Msg = #basic_message{content = #content{payload_fragments_rev = PFR}},
 msg_to_table(#basic_message{exchange_name = #resource{name = XName},
                             routing_keys  = RoutingKeys,
                             content       = Content}) ->
-    #content{properties = Props} =
-        rabbit_binary_parser:ensure_content_decoded(Content),
-    {PropsTable, _Ix} =
-        lists:foldl(fun (K, {L, Ix}) ->
-                            V = element(Ix, Props),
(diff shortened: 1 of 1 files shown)


deps/rabbit/src/rabbit_trace.erl
```diff
@@ -105,24 +105,57 @@ trace(X, Msg = #basic_message{content = #content{payload_fragments_rev = PFR}},
...
+    ].
+
+append_property(_Key, undefined, PropsTable) ->
+    PropsTable;
+append_property(Key = headers, Val, PropsTable) ->
+    [{rabbit_data_coercion:to_binary(Key), table, Val} | PropsTable];
+append_property(timestamp, Val, PropsTable) ->
+    [
+        %% TODO: should timestamp_in_ms added?
+        {<<"timestamp">>, long, Val} | PropsTable
```
bollhals, line 156: should this be long or timestamp? So it hits this path => https://github.com/rabbitmq/rabbitmq-server/blob/b62c5499cefa2f57d8516f3948543383d8387d4b/deps/rabbit_common/src/rabbit_binary_generator.erl#L117 (If I found the correct place it at all^^, but you get the point, so it is transfered as 'T' and not 'l')

michaelklishin:
We may come back to this at some point but right now, it is not a priority.

PR #2996, opened 2021-04-22
Streams: safer replica addition

kjnilsson:
Disallow replica additions if any of the existing replicas are more than
10 seconds out of date.


kjnilsson:
@Gsantomaggio says it works :)

PR #3002, opened 2021-04-26
Support streams in STOMP plugin

acogoluegnes:
This commit introduces the support of an x-stream-offset header
in the SUBSCRIBE frame to start consuming from a specific place
in a stream. The possible values are first, last, next, offset:<offset-value>
(e.g. offset:40000), timestamp:<timestamp-in-seconds> (e.g. timestamp:1619428685).

This commit also propagates the x-stream-offset header in the MESSAGE frame
to know the offset of a the delivered message in the stream.

### DIFF
--- a/deps/rabbitmq_stomp/include/rabbit_stomp_headers.hrl
+++ b/deps/rabbitmq_stomp/include/rabbit_stomp_headers.hrl
@@ -26,6 +26,7 @@
 -define(HEADER_PASSCODE, "passcode").
 -define(HEADER_PERSISTENT, "persistent").
 -define(HEADER_PREFETCH_COUNT, "prefetch-count").
+-define(HEADER_X_STREAM_OFFSET, "x-stream-offset").
 -define(HEADER_PRIORITY, "priority").
 -define(HEADER_RECEIPT, "receipt").
 -define(HEADER_REDELIVERED, "redelivered").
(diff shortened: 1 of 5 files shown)


acogoluegnes:
Acceptance steps

Get the code and run the broker:

```
cd /tmp
git clone git@github.com:rabbitmq/rabbitmq-server.git rabbitmq-server-stomp-for-streams
cd rabbitmq-server-stomp-for-streams
git checkout stomp-for-streams
make run-broker PLUGINS='rabbitmq_stream rabbitmq_stomp'
```

Get stream PerfTest and run it at a slow rate:

```
cd /tmp
wget https://github.com/rabbitmq/rabbitmq-java-tools-binaries-dev/releases/download/v-stream-perf-test-latest/stream-perf-test-latest.jar
java -jar /tmp/stream-perf-test-latest.jar --rate 100
```

Connect to the STOMP plugin with telnet:

```
telnet
telnet> open localhost 61613
```

Start the STOMP session (add a new line after the `CONNECT` command and insert the ^@ character with ctrl+@ and shift):

```
CONNECT

^@
```

You should get something like:

```
CONNECTED
server:RabbitMQ/3.8.10+329.g065b921
session:session-Z7A-n27Ar9iWvg5RVAvTaA
heart-beat:0,0
version:1.0
```

Now subscribe the stream the performance tool is publishing to:

```
SUBSCRIBE
destination:/amq/queue/stream
ack:client
prefetch-count:10
x-stream-offset:next

^@
```

10 messages should be delivered. Their body should not be readable, but they should each contain the `x-stream-offset` header and it should not start at 0 (it should be the end of the stream when the subscription was created):

```
MESSAGE
destination:/queue/stream
message-id:Q_/amq/queue/stream@@session-Z7A-n27Ar9iWvg5RVAvTaA@@1
redelivered:false
x-stream-offset:36652
content-length:10

	�;�M�
```

kjnilsson (approved):
LGMT - tests steps passed just fine

PR #3005, opened 2021-04-27
Adapt stream code to osiris tracking types introduction

acogoluegnes:
References rabbitmq/osiris#25.

### DIFF
--- a/deps/rabbit/src/rabbit_stream_queue.erl
+++ b/deps/rabbit/src/rabbit_stream_queue.erl
@@ -554,10 +554,10 @@ tracking_status(Vhost, QueueName) ->
         {ok, Q} when ?amqqueue_is_stream(Q) ->
             Leader = amqqueue:get_pid(Q),
             Map = osiris:read_tracking(Leader),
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -554,9 +554,9 @@ tracking_status(Vhost, QueueName) ->
         {ok, Q} when ?amqqueue_is_stream(Q) ->
             Leader = amqqueue:get_pid(Q),
             Map = osiris:read_tracking(Leader),
-            maps:fold(fun(K, V, Acc) ->
+            maps:fold(fun(K, {offset, OffsetValue}, Acc) ->
                               [[{reference, K},
-                                {offset, V}] | Acc]
+                                {offset, OffsetValue}] | Acc]
```
kjnilsson, line 559: how about we change `{offset, Value}` to `{type, Type}, {value, Value}`?

then it will work for all future tracking types such as `timestamp`

kjnilsson (approved):

PR #3007, opened 2021-04-28
Convert policy event keys to atoms

michaelklishin:
so that they are easier to match on


michaelklishin:
Backported to `v3.8.x`.

PR #3011, opened 2021-04-29
More accurate ack mode description in the UI

mkuratczyk:
I ran into this when working on metrics. This option sets `no_ack = true` which means an automatic ack. However, the old description, `Ack message requeue false`, suggests to me that there is a manual ack. I spent a lot of time trying to find a bug in the metrics code that counted this as an auto-acked message, when in fact it was an auto-acked message. :)


michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #3015, opened 2021-04-30
Rework AWS peer discovery integration suite

pjk25:
## Proposed Changes

Rework the `rabbitmq_peer_discovery_aws` `integration_SUITE.erl` with a (arguably) simpler approach not based on terraform - and allow the reintroduction of plugin tests to our CI pipelines.

In broad strokes the suite:
1. Starts an Amazon ECS cluster with 3 ec2 nodes.
2. Adjusts the security group to allow inter-node communication.
3. Defines an ECS task for rabbitmq, configured with aws peer discovery and "host" networking.
4. Starts an ECS service with 3 replicas using the task.
5. Contacts each node via the management api and asserts mutual cluster membership between all nodes.

Limitations:
1. The ECS task is based on a rabbitmq docker image, so this is not a "direct from source" test

Regarding Dependencies:
Rather than terraform, the suite requires the `aws` cli and `ecs-cli` cli.

## Types of Changes

- [x] Testing/Continuous Integration change

## Checklist

## Further Comments

The necessity of a docker image for this test is a tradeoff. On the one hand, it makes this suite inconvenient for rapid iteration based on local code changes. However, given that the current CI pipelines produce a docker image with every commit anyway, the image is already at hand in that context. Additionally, it provides the following benefits:
1. There is no longer a need to install Erlang and Elixir on spawned ec2 instances.
2. The tests are insulated from changes in AMIs.
3. I believe we have an overall speed improvement for the suite with a runtime of about 10 minutes (excluding building of the docker image). Having a "hot" ECS cluster would reduce the time to 5 or less.
4. Bringing the suite into Bazel was trivial in comparison to the old suite.
5. Differences between the local elixir version and the ec2 side no longer cause failures.

It's worth noting that while less flexible (we can no longer execute arbitrary erlang code on the remote vms), a hybrid approach may be possible with a different docker image. Given the assertions made by this suite however, this seemed the simplest approach, and reducing the maintenance burden of this suite was a primary goal.

The suite can be invoked with gnu make with:

```shell
make -C deps/rabbitmq_peer_discovery_aws ct-integration \
  AWS_ACCESS_KEY_ID=SOME_KEY_ID \
  AWS_SECRET_ACCESS_KEY=SOME_ACCESS_KEY \
  RABBITMQ_IMAGE=rabbitmq:3.8-management
```


deps/rabbitmq_peer_discovery_aws/test/integration_SUITE.erl
```diff
@@ -38,153 +42,121 @@ groups() ->
...
+                         {ecs_instance_role, "ecs-peer-discovery-aws"},
+                         {ecs_cluster_size, ?CLUSTER_SIZE},
+                         {rabbitmq_default_user, "test"},
+                         {rabbitmq_default_pass, rabbit_ct_helpers:random_term_checksum()},
+                         {rabbitmq_erlang_cookie, rabbit_ct_helpers:random_term_checksum()}
+                        ]),
+    Config2 = rabbit_ct_helpers:register_teardown_step(Config1, fun aws_ecs_util:destroy_ecs_cluster/1),
+    rabbit_ct_helpers:run_steps(
+      Config2, [
+                fun rabbit_ct_helpers:init_skip_as_error_flag/1,
```
michaelklishin, line 61: This function wasn't exported, I pushed a `rabbitmq-ct-helpers` change directly

pjk25:
I forgot to include https://github.com/rabbitmq/rabbitmq-ct-helpers/commit/a74651d18af9559f4137bd8a80cbd774a9d05bf4

Thanks for picking it up @michaelklishin

deps/rabbitmq_peer_discovery_aws/test/integration_SUITE.erl
pjk25: Yes, I made that change locally and forgot to push it. Thanks.

pjk25:
I should also note that this suite makes use of the fact that ec2 nodes created by the `ecs-cli` for an ECS cluster seem to automatically be part of an autoscaling group. Should that change, it would break the test for clustering via autoscaling groups.

pjk25:
Backported to `v3.8.x` with e3d646138902dd65d37dea2de48b50a43e3981de

PR #3024, opened 2021-05-04
Link to the connection to avoid stuck unacked messages when using get messages

luos:
When using the get messages functionality with a high message count the request
may time out if it can not be completed in one minute. This leaves the connection
hanging around.

## Proposed Changes

When running "Get messages" with a high message count the request may time out, causing messages to be stuck in the Unacknowledged state. To reproduce, create a queue with 100k messages and try to get 10k out of it. 

By linking to the connection the connection will be killed when the cowboy request is killed by timeout. 

One drawback of this method may be is that if the connection crashes for any reason this will not provide a proper error message for the user, as it will kill the request. 

Any thoughts?

## Types of Changes

## Checklist


michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #3028, opened 2021-05-06
Escape username before displaying it

michaelklishin:
All other values displayed in pop-ups are already escaped.


michaelklishin:
Backported to `v3.8.x`.

Issue #3032, opened 2021-05-11
A model shutdown occurs if the consumer does not acknowledge the received message in 15 minutes


Kalhorn:
Rabbit MQ Version: 3.8.16
Erlang Version: 23.3.4
Dot Net Client Version: 5.1.2

Since upgrading to 3.8.16 from 3.8.14 we are seeing that a consumer is being shutdown if processing of the received message takes longer than 15 minutes.

The queue is a Durable, Single Access Consumer queue
I am using the 3-management docker image for this test.

Output from my test, the handler should simply wait for 20 minutes then acknowledge the message
```
Received Msg : 0, 1 (0)
Waiting for 20 minutes 11/05/2021 14:52:27
[Shutdown Handler] Model shutdown 11/05/2021 15:08:26
[Shutdown Handler] AMQP close-reason, initiated by Peer, code=406, text='PRECONDITION_FAILED - consumer ack timed out on channel 1', classId=0, methodId=0
Wait complete
````

Same test running against RabbitMQ Version: 3.8.14

```
Received Msg : 0, 1 (0)
Waiting for 20 minutes 11/05/2021 15:13:34
Wait complete
```

PR #3033, opened 2021-05-11
Bump consumer acknowledgement timeout default to 30 minutes

michaelklishin:
Apparently 15 minutes is not enough for some. 1 hour seems to be
unreasonably long to our team, though.

References #2990, #3032


michaelklishin:
Backported to `v3.8.x`.

PR #3034, opened 2021-05-11
Mention delivery ack timeout configurability

michaelklishin:
in the logs and consumer channel error message.

References #2990, #3032.


michaelklishin:
Backported to `v3.8.x`.

PR #3055, opened 2021-05-20
Log heartbets at debug; remove newlines

mkuratczyk:
A log line every minute about a successful heartbeat pollutes the logs.
So do empty log lines.


michaelklishin (approved):

michaelklishin:
These should arguably also use the connection category but fair enough.

PR #3067, opened 2021-05-26
{consistent_hash,recent_history}_exchange plugins: add retries to schema sync

michaelklishin:
So, do what RabbitMQ core does as of 3.6.7 or so.

This makes it possible for nodes with those plugins enabled to be
restarted in arbitrary order within a certain time window, just
like nodes without those plugins.


michaelklishin:
Backported to `v3.8.x`.

PR #3068, opened 2021-05-27
Handle cases where virtual host config file does not yet exist

michaelklishin:
This happens during queue migration from a pre-3.7 version.

References #2954.


michaelklishin:
Backported to `v3.8.x`.

PR #3075, opened 2021-06-02
Remove randomized startup delays

ansd:
On initial cluster formation, only one node in a multi node cluster
should initialize the Mnesia database schema (i.e. form the cluster).
To ensure that for nodes starting up in parallel,
RabbitMQ peer discovery backends have used
either locks or randomized startup delays.

Locks work great: When a node holds the lock, it either starts a new
blank node (if there is no other node in the cluster), or it joins
an existing node. This makes it impossible to have two nodes forming
the cluster at the same time.
Consul and etcd peer discovery backends use locks. The lock is acquired
in the consul and etcd infrastructure, respectively.

For other peer discovery backends (classic, K8s, AWS), randomized
startup delays were used. They work good enough in most cases.
However, in https://github.com/rabbitmq/cluster-operator/issues/662 we
observed that in 1% - 10% of the cases (the more nodes or the
smaller the randomized startup delay range, the higher the chances), two
nodes decide to form the cluster. That's bad since it will end up in a
single Erlang cluster, but in two RabbitMQ clusters. Even worse, no
obvious alert got triggered or error message logged.

To solve this issue, one could increase the randomized startup delay
range from e.g. 0m - 1m to 0m - 3m. However, this makes initial cluster
formation very slow since it will take up to 3 minutes until
every node is ready. In rare cases, we still end up with two nodes
forming the cluster.

Another way to solve the problem is to name a dedicated node to be the
seed node (forming the cluster). This was explored in
https://github.com/rabbitmq/cluster-operator/pull/689 and works well.
Two minor downsides to this approach are: 1. If the seed node never
becomes available, the whole cluster won't be formed (which is okay),
and 2. it doesn't integrate with existing dynamic peer discovery backends
(e.g. K8s, AWS) since nodes are not yet known at deploy time.

In this PR, we take a better approach: We remove randomized startup
delays altogether. We replace them with locks. However, instead of
implementing our own lock implementation in an external system (e.g. in K8s),
we re-use Erlang's locking mechanism global:set_lock/3.

global:set_lock/3 has some convenient properties:
1. It accepts a list of nodes to set the lock on.
2. The nodes in that list connect to each other (i.e. create an Erlang
cluster).
3. The method is synchronous with a timeout (number of retries). It
blocks until the lock becomes available.
4. If a process that holds a lock dies, or the node goes down, the lock
held by the process is deleted.

The list of nodes passed to global:set_lock/3 corresponds to the nodes
the peer discovery backend discovers (lists).

Two special cases worth mentioning:

1. That list can be all desired nodes in the cluster
(e.g. in classic peer discovery where nodes are known at
deploy time) while only a subset of nodes is available.
In that case, global:set_lock/3 still sets the lock not
blocking until all nodes can be connected to. This is good since
nodes might start sequentially (non-parallel).

2. In dynamic peer discovery backends (e.g. K8s, AWS), this
list can be just a subset of desired nodes since nodes might not startup
in parallel. That's also not a problem as long as the following
requirement is met: "The peer discovery backend does not list two disjoint
sets of nodes (on different nodes) at the same time."
For example, in a 2-node cluster, the peer discovery backend must not
list only node 1 on node 1 and only node 2 on node 2.

Existing peer discovery backends fulfil that requirement because the
resource the nodes are discovered from is global.
For example, in K8s, once node 1 is part of the Endpoints object, it
will be returned on both node 1 and node 2.
Likewise, in AWS, once node 1 started, the described list of instances
with a specific tag will include node 1 when the AWS peer discovery backend
runs on node 1 or node 2.

Removing randomized startup delays also makes cluster formation
considerably faster (up to 1 minute faster if that was the
upper bound in the range).

### How these changes were tested

**K8s peer discovery backend:**

Deployed cluster-operator v1.7.0 on a GKE cluster.
<details>
<summary>rabbitmq.yml</summary>
<pre>
apiVersion: rabbitmq.com/v1beta1
kind: RabbitmqCluster
metadata:
  name: cluster-((index))
spec:
  replicas: ((replicas))
  image: "pivotalrabbitmq/rabbitmq:remove-randomized-startup-delays"
  rabbitmq:
    additionalConfig: |
      log.console.level = debug
  resources:
    limits:
      cpu: "1"
      memory: 512Mi
    requests:
      cpu: "0.5"
      memory: 512Mi
</pre>
</details>

Deployed 100 9-node RabbitMQ clusters in sequence checking that all 9 nodes got clustered and checking that there were no container restarts (thanks @mkuratczyk):

<details>
<summary>The following script also stores all the pod logs</summary>
<pre>
#!/usr/bin/env bash
set -euxo pipefail

clusters=100
replicas=9

max_pod_index=$(( replicas - 1 ))
for i in $(seq 1 "$clusters"); do
  sed "s/((index))/$i/ ; s/((replicas))/$replicas/" rabbitmq.yml | kubectl apply -f -
    sleep 5
    printf "\nRun %d:\n" "$i" >> results.txt
    for pod in $(seq 0 $max_pod_index); do
      kubectl wait --for=condition=Ready=true pod/cluster-"$i"-server-"$pod" --timeout=5m
    done
    for pod in $(seq 0 $max_pod_index); do
      kubectl exec cluster-"$i"-server-"$pod" -c rabbitmq -- rabbitmqadmin list nodes | grep -c disc >> results.txt
    done
    kubectl get pods | grep server | awk '{print $4}' | xargs -I {} echo Container restarts: {} >> results.txt
    mkdir -p logs/"$i"
    for pod in $(seq 0 $max_pod_index); do
        # Save logs from all RabbitMQ nodes
        kubectl logs cluster-"$i"-server-"$pod" > logs/"$i"/cluster-"$i"-server-"$pod".log
        if [[ $(kubectl get pod cluster-"$i"-server-"$pod" | grep server | awk '{print $4}') != 0 ]];then 
         # There was a container restart => get logs from previous container
         kubectl logs --previous cluster-"$i"-server-"$pod" > logs/"$i"/cluster-"$i"-server-"$pod"-previous.log
        fi
    done
    kubectl rabbitmq delete cluster-"$i"
    kubectl wait --for=delete pods -l=app.kubernetes.io/name=cluster-"$i" --timeout=5m
    sleep 30
done
</pre>
</details>

**Classic peer discovery backend:**
Same as K8s peer discovery backend but using https://github.com/rabbitmq/cluster-operator/tree/peer-discovery-classic

**AWS peer discovery backend:**
1. Create 3 EC2 instances with Ubuntu 20.04 and the same tag `mykey: myvalue`.
1. For the security group, allow inbound 4369 and 4369 ports between nodes (i.e. from the same security group) and ssh from my IP address.

On all 3 instances, do the following:
1. `ssh ubuntu@<public IP> -i <pem file>`
1. Install Erlang and Elixir via apt-get.
1. rabbit_common depends on python: `sudo ln -s /usr/bin/python3 /usr/bin/python`
1. git clone rabbitmq-server and:
`make run-broker TEST_TMPDIR="/home/ubuntu/rabbitmq-server/tmp/test" RABBITMQ_CONFIG_FILE="/home/ubuntu/rabbitmq-server/tmp/rabbitmq.conf" RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="-setcookie mycookie"`

rabbitmq.conf:
```yaml
cluster_formation.peer_discovery_backend = aws
cluster_formation.aws.region = eu-central-1
cluster_formation.aws.access_key_id = XXX
cluster_formation.aws.secret_key = XXX/aaaa
cluster_formation.aws.use_autoscaling_group = false
cluster_formation.aws.instance_tags.mykey = myvalue
```

### TODOs
- [ ] docs

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -945,11 +945,11 @@ fun(Conf) ->
 end}.
 
 %% Cluster formation: Randomized startup delay
+%%
+%% DEPRECATED: This is a no-op. Old configs are still allowed, but a warning will be printed.
 
-{mapping, "cluster_formation.randomized_startup_delay_range.min", "rabbit.cluster_formation.randomized_startup_delay_range",
-    [{datatype, integer}]}.
-{mapping, "cluster_formation.randomized_startup_delay_range.max", "rabbit.cluster_formation.randomized_startup_delay_range",
-    [{datatype, integer}]}.
+{mapping, "cluster_formation.randomized_startup_delay_range.min", "rabbit.cluster_formation.randomized_startup_delay_range", []}.
+{mapping, "cluster_formation.randomized_startup_delay_range.max", "rabbit.cluster_formation.randomized_startup_delay_range", []}.
 
 {translation, "rabbit.cluster_formation.randomized_startup_delay_range",
 fun(Conf) ->
@@ -957,18 +957,24 @@ fun(Conf) ->
     Max = cuttlefish:conf_get("cluster_formation.randomized_startup_delay_range.max", Conf, undefined),
 
     case {Min, Max} of
-        {undefined, undefined} ->
-            cuttlefish:unset();
-        {undefined, Max}       ->
-            %% fallback default
-            {5, Max};
-        {Min, undefined}       ->
-            %% fallback default
-            {Min, 60};
-        {Min, Max}             ->
-            {Min, Max}
-    end
+      {undefined, undefined} ->
+        ok;
+      _ ->
+        cuttlefish:warn("cluster_formation.randomized_startup_delay_range.min and "
+                               "cluster_formation.randomized_startup_delay_range.max are deprecated")
+    end,
+    cuttlefish:unset()
 end}.
+
+%% Cluster formation: lock acquisition retries as passed to https://erlang.org/doc/man/global.html#set_lock-3
+%%
+%% Currently used in classic, k8s, and aws peer discovery backends.
+
+{mapping, "cluster_formation.internal_lock_retries", "rabbit.cluster_formation.internal_lock_retries",
+    [
+     {datatype, integer},
+     {validators, ["non_zero_positive_integer"]}
+    ]}.
 
 %% Cluster formation: discovery failure retries
 
--- a/deps/rabbit/src/rabbit_mnesia.erl
+++ b/deps/rabbit/src/rabbit_mnesia.erl
@@ -105,21 +105,18 @@ init_with_lock(Retries, Timeout, RunPeerDiscovery) ->
     rabbit_log:debug("rabbit_peer_discovery:lock returned ~p", [LockResult]),
     case LockResult of
         not_supported ->
-            rabbit_log:info("Peer discovery backend does not support locking, falling back to randomized delay"),
-            %% See rabbitmq/rabbitmq-server#1202 for details.
-            rabbit_peer_discovery:maybe_inject_randomized_delay(),
             RunPeerDiscovery(),
             rabbit_peer_discovery:maybe_register();
-        {error, _Reason} ->
-            timer:sleep(Timeout),
-            init_with_lock(Retries - 1, Timeout, RunPeerDiscovery);
         {ok, Data} ->
             try
                 RunPeerDiscovery(),
                 rabbit_peer_discovery:maybe_register()
             after
                 rabbit_peer_discovery:unlock(Data)
-            end
+            end;
+        {error, _Reason} ->
+            timer:sleep(Timeout),
+            init_with_lock(Retries - 1, Timeout, RunPeerDiscovery)
     end.
 
 -spec run_peer_discovery() -> ok | {[node()], node_type()}.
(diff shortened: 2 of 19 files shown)


michaelklishin (approved):

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -945,31 +945,37 @@ fun(Conf) ->
...
-        {Min, undefined}       ->
-            %% fallback default
-            {Min, 60};
-        {Min, Max}             ->
-            {Min, Max}
-    end
+      {undefined, undefined} ->
+        ok;
+      _ ->
+        cuttlefish:warn("cluster_formation.randomized_startup_delay_range.min and "
```
michaelklishin, line 963: FTR, Cuttlefish schema translation happens so early that this won't be visible in the regular or prelaunch log, at least that what I observed. This is a general problem with logging at such *really* early node boot stages, we shouldn't try to address it in this PR (or even for 3.9).

The point is that existing config files with these options will pass validation 👍

michaelklishin (approved):

michaelklishin:
Merging into master (will backport after `3.8.17` ships) because:

 * Code changes look reasonable
 * Kubernetes was extensively tested and will be tested more (as part of the cluster Operator, for example)
 * AWS test suite passes
 * Classic config suite passes and cluster formation works as expected

michaelklishin:
Backported to `v3.8.x`.

PR #3098, opened 2021-06-10
Close stream socket if client doesn't follow authentication protocol

ansd:
Before this commit, sending garbage data to the server stream port caused the RabbitMQ node to eat more and more memory.
In this commit, we fix it by expecting the client to go through the proper authentication sequence. Otherwise, the server closes the socket.

Pair: @mkuratczyk

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
@@ -335,6 +335,21 @@ listen_loop_pre_auth(Transport,
             rabbit_log:info("Transitioned from ~p to ~p",
                             [ConnectionStep0, ConnectionStep]),
             case ConnectionStep of
+                peer_properties_exchanged when ConnectionStep0 =:= tcp_connected ->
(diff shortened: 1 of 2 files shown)


deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -366,19 +381,23 @@ listen_loop_pre_auth(Transport,
...
                                           State1,
                                           Configuration);
-                failure ->
-                    close(Transport, S, State);
                 _ ->
-                    listen_loop_pre_auth(Transport,
-                                         Connection1,
-                                         State1,
-                                         Configuration)
+                    rabbit_log:warning("Closing socket ~w because transition from ~p to ~p is not allowed",
```
michaelklishin, line 385: This should probably use`rabbit_log_connection` so that the message is categorized.

ansd: Thanks. I changed all connection related log messages in this file from `rabbit_log` to `rabbit_log_connection`.

PR #3107, opened 2021-06-15
Avoid an exception when queues of different types are federated

kjnilsson:
Queue federation fails when federating from a classic queue to a quorum queue (and vice versa). This PR supports mixed queue type federation, by avoiding to re-declare the upstream if it already exists (as it might have a different type).


dcorbacho (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #3124, opened 2021-06-21
Use Erlang 24 with JIT in Docker image

ansd:
GitHub action triggered by this PR created a new image:
```
> docker run -ti pivotalrabbitmq/rabbitmq:docker-image-erlang-24 erl
Erlang/OTP 24 [erts-12.0.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Eshell V12.0.2  (abort with ^G)
1>
```
As seen above, Erlang 24 and JIT is used.

Pair: @mkuratczyk

### DIFF
--- a/.github/workflows/oci.yaml
+++ b/.github/workflows/oci.yaml
@@ -21,15 +21,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2.3.4
 
-      # RabbitMQ master requires Erlang 23 (correct at Mar 2021)
+      # RabbitMQ master supports Erlang 24 (correct at June 2021)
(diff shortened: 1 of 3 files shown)


pjk25 (commented):
in line 41 the `OTP_VERSION` and `OTP_SHA256` could be passed through like `SKIP_PGP_VERIFY`, which would ensure that compilation and runtime Erlang are the always in sync

ansd:
Thanks @pjk25. Good point, I did the change (and force pushed to re-trigger and therefore re-test the `OCI / build-publish-dev` action).

michaelklishin (approved):

Issue #2783, opened 2021-02-02
Aggregated `queue_messages_published_total` metric violates Prometheus expectations about counters


michaelklishin:
See #2781 for the background.

Some metrics, e.g. `queue_messages_published_total` are computed as a basic sum aggregation of samples from a local ETS table, e.g. `channel_queue_exchange_metrics`. When a channel is closed, its samples are removed from the table,
decreasing the sum. This violates an expectation for `counter` metrics in Prometheus: they can only increment or stay flat or reset to 0 but not decrease.

The issue is not present when per-object metrics are used: when a channel is closed, all of its **metrics** go away, which is what the user expects to happen.

This is a side-effect of our quick-and-dirty switch to aggregated metrics. We need to retain this historical total or delegate
to the Prometheus client library which will do most aggregation work and handle resets.

This is node local state, so we can address this even with a significant rework of the Prometheus plugin and still ship it in a `3.8.x` release.

Per discussion with @dcorbacho @gerhard @kjnilsson.

gerhard:
Will be picking this one up first thing, and working with @dcorbacho on the review so that she can continue focusing on her in-flight.

dcorbacho:
Luckily we're not tracking totals on the channel but doing increments, see https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit/src/rabbit_channel.erl#L205

So when `queue_exchange_stats` are increased we can use that value to increase a *new* publish counter to track totals per node (or even exchange).  Here: https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit/src/rabbit_channel.erl#L2123

karanlik:
Hi All, 

Please sorry if i'm posting at a wrong place. 

Currently we're seeing some behaviour similar to this issue; our "rabbitmq_channel_messages_published_total" metrics (with aggregation enabled) decreases which causes false alarms & mis-guiding graphs. I guess this is some sort of bug or at least something not good for Prometheus. What is the proposed workaround-solution for such kind of situation? Going to per-object metrics ? 
Once again sorry if I'm at wrong location.
Thanks...

gerhard:
The fix is already available in what will become 3.9.0: https://github.com/rabbitmq/rabbitmq-server/pull/3127

You will notice that the PR has 5 more outstanding tasks, the next one being a back-port to v3.8.x. Once this is done, the issue will be fixed via an update to the dashboard (also one of those 5 tasks).

[...]

PR #3127, opened 2021-06-21
Global counters per protocol + protocol AND queue_type

gerhard:
This way we can show how many messages were received via a certain protocol (stream is the second real protocol besides the default amqp091 one), as well as by queue type, which is something that many asked for a really long time.

The most important aspect is that we can also see them by protocol AND queue_type, which becomes very important for Streams, which have different rules from regular queues (e.g. for example, consuming messages is non-destructive, and deep queue backlogs - think billions of messages - are normal). Alerting and consumer scaling due to deep backlogs will now work correctly, as we can distinguish between regular queues & streams.

This has gone through a few cycles, with @mkuratczyk & @dcorbacho covering most of the ground. @dcorbacho had most of this in https://github.com/rabbitmq/rabbitmq-server/pull/3045, but the main branch went through a few changes in the meantime. Rather than resolving all the conflicts, and then making the necessary changes, we (@gerhard + @kjnilsson) took all learnings and started re-applying a lot of the existing code from #3045. We are confident in this approach and would like to see it through. We continued working on this with @dumbbell, and the most important changes are captured in https://github.com/rabbitmq/seshat/pull/1.

We expose these global counters in rabbitmq_prometheus via a new collector. We don't want to keep modifying the existing collector, which grew really complex in parts, especially since we introduced aggregation, but start with a new namespace, `rabbitmq_global_`, and continue building on top of it. The idea is to build in parallel, and slowly transition to the new metrics, because semantically the changes are too big since streams, and we have been discussing protocol-specific metrics with @kjnilsson, which makes me think that this approach is least disruptive and... simple.

While at this, we removed redundant empty return value handling in the channel. The function called no longer returns this.

Also removed all DONE / TODO & other comments - we'll handle them when the time comes, no need to leave TODO reminders.

Pairs @kjnilsson @dcorbacho @dumbbell 
(this is multiple commits squashed into one)

## Next steps
- [x] Create new PR and let @mkuratczyk and @ansd know so that they become familiar with the changes - fresh 👀 👀
- [x] Close https://github.com/rabbitmq/rabbitmq-server/pull/3045 (superseded by this PR)
- [ ] After both OTP 23 & 24 tests pass, merge this PR as a few of us have already paired on it and we have been reviewing it while building, multiple times, so there is no need for further review
- [ ] Backport to 3.9.x as is
- [ ] Backport parts of this to 3.8.x so that we can finally address https://github.com/rabbitmq/rabbitmq-server/issues/2783
- [ ] Fix & publish new version of the RabbitMQ-Overview Grafana dashboard
- [ ] Fix & finally publish the new RabbitMQ-Streams Grafana dashboard


workspace_helpers.bzl
```diff
@@ -239,8 +239,8 @@ erlang_lib(
     github_bazel_erlang_lib(
         name = "seshat",
         org = "rabbitmq",
-        ref = "main",
-        version = "main",
+        ref = "refs/heads/valid-prometheus-format",
+        version = "refs/heads/valid-prometheus-format",
```
pjk25, line 243: The value will be injected into `ebin/seshat.app` so I would probably use `version = "valid-prometheus-format"`. I have been considering some enhancements to `github_bazel_erlang_lib` where the version would be derived automatically unless overridden. It would eliminate the need for this line in this case, which would be nice.

deps/rabbit/src/rabbit_global_counters.erl
```diff
@@ -0,0 +1,183 @@
...
+
+messages_received(Protocol, Num) ->
+    counters:add(fetch(Protocol), ?MESSAGES_RECEIVED, Num).
+
+messages_received_confirm(Protocol, Num) ->
+    counters:add(fetch(Protocol), ?MESSAGES_RECEIVED_CONFIRM, Num).
+
+% Formerly known as queue_messages_published_total
+% This is a more accurate representation of what happens with messages that
+%   arrive on a channel or connection in the case of streams.
```
ansd, line 142: As commented above, in my test, I didn't see that metric getting incremented when sending a message via `stream` protocol.

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_global_metrics_collector.erl
```diff
@@ -0,0 +1,50 @@
...
+-export([register/0,
+         deregister_cleanup/1,
+         collect_mf/2]).
+
+-import(prometheus_model_helpers, [create_mf/4,
+                                   counter_metric/2]).
+
+%% This exposes the new global metrics which are instrumented directly, and bypass the Rabbit Core metrics entirely.
+%% The short-term plan is to start building the new metrics sub-system alongside RabbitMQ Core metrics
+%% The long-term plan is to replace RabbitMQ Core metrics with this new approach.
```
ansd, line 21: Is my understanding correct that the new approach will only by used for aggregated counters, i.e. not for gauges and  not for per-object metrics?
Per-object metrics need to be handled by RabbitMQ Core metrics and can't be migrated to the new approach since the new approach "is best to only be used by global persistent counters that will never be deleted."?

ansd (commented):
Awesome change!

I left some in-line comments.

I know it's not part of this PR, but I looked into seshat as well.
Should we use the `write_concurrency` option for the counters initialised in https://github.com/rabbitmq/seshat/blob/4e190ebec1707df8cfaba51753f92a7588fe0e2e/src/seshat_counters.erl#L34?
Our use case fits that option very well:
1. The channel processes and stream reader connection processes write concurrently.
2. The writes are very frequent compared to reading via `seshat_counters:prometheus_format/1` and `seshat_counters:overview/1`.
3. We don't require "absolute read consistency".

gerhard:
> Awesome change!
> 
> I left some in-line comments.
> 
> I know it's not part of this PR, but I looked into seshat as well.
> Should we use the `write_concurrency` option for the counters initialised in https://github.com/rabbitmq/seshat/blob/4e190ebec1707df8cfaba51753f92a7588fe0e2e/src/seshat_counters.erl#L34?
> Our use case fits that option very well:
> 
> 1. The channel processes and stream reader connection processes write concurrently.
> 2. The writes are very frequent compared to reading via `seshat_counters:prometheus_format/1` and `seshat_counters:overview/1`.
> 3. We don't require "absolute read consistency".

That makes sense to me. WDYT @kjnilsson & @dcorbacho?

If we reach soft consensus, do you want to contribute that PR to seshat @ansd ?

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -135,6 +135,7 @@
          peer_cert_issuer,
          peer_cert_subject,
          peer_cert_validity]).
+-define(STREAM_QUEUE_TYPE, rabbit_stream_queue).
```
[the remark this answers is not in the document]
gerhard: 1. Great spot - done!
2. Yes, it's the module name, and gets used here: https://github.com/rabbitmq/rabbitmq-server/pull/3127/files#diff-ab2459b0afd286f14f2f08978ddbc9c55ee927f933eb3d687a4f0bd4a8c572c6R1367

deps/rabbit/src/rabbit_global_counters.erl
```diff
@@ -0,0 +1,183 @@
...
+    init([{protocol, amqp091}, {queue_type, rabbit_stream_queue}]).
+
+init(Labels = [{protocol, Protocol}, {queue_type, QueueType}]) ->
+    _ = seshat_counters:new_group(?MODULE),
+    Counters = seshat_counters:new(?MODULE, Labels, ?PROTOCOL_QUEUE_TYPE_COUNTERS),
+    persistent_term:put({?MODULE, Protocol, QueueType}, Counters),
+    ok;
+init(Labels = [{protocol, Protocol}]) ->
+    _ = seshat_counters:new_group(?MODULE),
+    Counters = seshat_counters:new(?MODULE, Labels, ?PROTOCOL_COUNTERS),
```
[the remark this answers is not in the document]
gerhard: You are right, they don't, but the way seshat currently works would require us to duplicate counters. We picked up on this issue, but the simplest alternative makes things worse. For example, this is what we have now:

```erlang
%% PROTOCOL COUNTERS:
-define(MESSAGES_RECEIVED, 1).
-define(MESSAGES_RECEIVED_CONFIRM, 2).
-define(MESSAGES_ROUTED, 3).
-define(MESSAGES_UNROUTABLE_DROPPED, 4).
-define(MESSAGES_UNROUTABLE_RETURNED, 5).
-define(MESSAGES_CONFIRMED, 6).
-define(PROTOCOL_COUNTERS,
            [
                {
                    messages_received_total, ?MESSAGES_RECEIVED, counter,
                    "Total number of messages received from publishers"
                },
                {
                    messages_received_confirm_total, ?MESSAGES_RECEIVED_CONFIRM, counter,
                    "Total number of messages received from publishers expecting confirmations"
                },
                {
                    messages_routed_total, ?MESSAGES_ROUTED, counter,
                    "Total number of messages routed to queues or streams"
                },
                {
                    messages_unroutable_dropped_total, ?MESSAGES_UNROUTABLE_DROPPED, counter,
                    "Total number of messages published as non-mandatory into an exchange and dropped as unroutable"
                },
                {
                    messages_unroutable_returned_total, ?MESSAGES_UNROUTABLE_RETURNED, counter,
                   "Total number of messages published as mandatory into an exchange and returned to the publisher as unroutable"
                },
                {
                    messages_confirmed_total, ?MESSAGES_CONFIRMED, counter,
                    "Total number of messages confirmed to publishers"
                }
            ]).
```

And this is the simplest change that we could make, which doesn't look like a step in the right direction (more code is required, but this is the bulk of it):

```erlang
%% STREAM PROTOCOL COUNTERS:
-define(STREAM_MESSAGES_RECEIVED, 1).
-define(STREAM_MESSAGES_RECEIVED_CONFIRM, 2).
-define(STREAM_MESSAGES_CONFIRMED, 3).
-define(STREAM_PROTOCOL_COUNTERS,
            [
                {
                    messages_received_total, ?STREAM_MESSAGES_RECEIVED, counter,
                    "Total number of messages received from publishers"
                },
                {
                    messages_received_confirm_total, ?STREAM_MESSAGES_RECEIVED_CONFIRM, counter,
                    "Total number of messages received from publishers expecting confirmations"
                },
                {
                    messages_confirmed_total, ?STREAM_MESSAGES_CONFIRMED, counter,
                    "Total number of messages confirmed to publishers"
                }
            ]).

%% AMQP091 PROTOCOL COUNTERS:
-define(AMQP091_MESSAGES_RECEIVED, 1).
-define(AMQP091_MESSAGES_RECEIVED_CONFIRM, 2).
-define(AMQP091_MESSAGES_ROUTED, 3).
-define(AMQP091_MESSAGES_UNROUTABLE_DROPPED, 4).
-define(AMQP091_MESSAGES_UNROUTABLE_RETURNED, 5).
-define(AMQP091_MESSAGES_CONFIRMED, 6).
-define(AMQP091_PROTOCOL_COUNTERS,
            [
                {
                    messages_received_total, ?AMQP091_MESSAGES_RECEIVED, counter,
                    "Total number of messages received from publishers"
                },
                {
                    messages_received_confirm_total, ?AMQP091_MESSAGES_RECEIVED_CONFIRM, counter,
                    "Total number of messages received from publishers expecting confirmations"
                },
                {
                    messages_routed_total, ?AMQP091_MESSAGES_ROUTED, counter,
                    "Total number of messages routed to queues or streams"
                },
                {
                    messages_unroutable_dropped_total, ?AMQP091_MESSAGES_UNROUTABLE_DROPPED, counter,
                    "Total number of messages published as non-mandatory into an exchange and dropped as unroutable"
                },
                {
                    messages_unroutable_returned_total, ?AMQP091_MESSAGES_UNROUTABLE_RETURNED, counter,
                   "Total number of messages published as mandatory into an exchange and returned to the publisher as unroutable"
                },
                {
                    messages_confirmed_total, ?AMQP091_MESSAGES_CONFIRMED, counter,
                    "Total number of messages confirmed to publishers"
                }
            ]).
```

Rather than doing the above, we (+@dumbbell) want to go for a refactor in seshat, which will allow counters to be initialised independently, without hard-coding the context (e.g. protocol type) into the counters themselves.

Another good spot @ansd which confirms that what we picked up on bugs you too 😉

deps/rabbit/src/rabbit_global_counters.erl
gerhard: Right, this is not something that a `stream_reader` does. These are all the counters that it currently tracks: https://github.com/rabbitmq/rabbitmq-server/blob/b2200046c474efddba614254760cce10eaab6273/deps/rabbitmq_stream/src/rabbit_stream_reader.erl#L277-L294 . I will remove the comment because it's obviously confusing.

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_global_metrics_collector.erl
gerhard: Yes, `rabbit_global_counters` is only meant to be used for aggregated counters. We want to track per-object metrics using seshat too, but via a different module (or maybe modules?).

The long-term plan is to extend the seshat library with per-object capabilities, and eventually deprecate the way we currently track metrics, including emitting them periodically, which creates a lot of overhead. @dcorbacho has been talking about this for years, and we finally started addressing it, which makes me super happy, and others too.

@deadtrickster will be joining this party soon 🙌🏻

deps/rabbit/src/rabbit_global_counters.erl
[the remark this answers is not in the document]
gerhard: Please re-open @ansd if you feel this is not resolved 🧐

gerhard:
Back-ported to `v3.9.x`

ggustafsson:
Hello!

Any updates on task "Update RabbitMQ-Overview Grafana dashboard to use the new global counters & update the public versions"?

We got bitten by this one at work last week so I took a look at all the `*_total` metric points in the dashboard and remapped them. It seems to work as intended but I am not 100% sure as not all metric names was directly translatable so I had to guess.

## Remap

### Certain
    rabbitmq_channel_get_empty_total                    -> rabbitmq_global_messages_get_empty_total
    rabbitmq_channel_messages_acked_total               -> rabbitmq_global_messages_acknowledged_total
    rabbitmq_channel_messages_confirmed_total           -> rabbitmq_global_messages_confirmed_total
    rabbitmq_channel_messages_delivered_total           -> rabbitmq_global_messages_delivered_total
    rabbitmq_channel_messages_redelivered_total         -> rabbitmq_global_messages_redelivered_total
    rabbitmq_channel_messages_unroutable_dropped_total  -> rabbitmq_global_messages_unroutable_dropped_total
    rabbitmq_channel_messages_unroutable_returned_total -> rabbitmq_global_messages_unroutable_returned_total

### Almost Certain
    rabbitmq_channel_get_ack_total                -> rabbitmq_global_messages_delivered_get_manual_ack_total
    rabbitmq_channel_get_total                    -> rabbitmq_global_messages_delivered_get_auto_ack_total
    rabbitmq_channel_messages_delivered_ack_total -> rabbitmq_global_messages_delivered_consume_manual_ack_total
    rabbitmq_channel_messages_published_total     -> rabbitmq_global_messages_received_total

### Unknown
    rabbitmq_channels_closed_total          -> ?
    rabbitmq_channels_opened_total          -> ?
    rabbitmq_connections_closed_total       -> ?
    rabbitmq_connections_opened_total       -> ?
    rabbitmq_queue_messages_published_total -> ?
    rabbitmq_queues_created_total           -> ?
    rabbitmq_queues_declared_total          -> ?
    rabbitmq_queues_deleted_total           -> ?

## Solution

This is what I ran to remap the values in the dashboard file:
```
sed -i "" \
  -e 's/rabbitmq_channel_get_ack_total/rabbitmq_global_messages_delivered_get_manual_ack_total/g' \
  -e 's/rabbitmq_channel_get_empty_total/rabbitmq_global_messages_get_empty_total/g' \
  -e 's/rabbitmq_channel_get_total/rabbitmq_global_messages_delivered_get_auto_ack_total/g' \
  -e 's/rabbitmq_channel_messages_acked_total/rabbitmq_global_messages_acknowledged_total/g' \
  -e 's/rabbitmq_channel_messages_confirmed_total/rabbitmq_global_messages_confirmed_total/g' \
  -e 's/rabbitmq_channel_messages_delivered_ack_total/rabbitmq_global_messages_delivered_consume_manual_ack_total/g' \
  -e 's/rabbitmq_channel_messages_delivered_total/rabbitmq_global_messages_delivered_total/g' \
  -e 's/rabbitmq_channel_messages_published_total/rabbitmq_global_messages_received_total/g' \
  -e 's/rabbitmq_channel_messages_redelivered_total/rabbitmq_global_messages_redelivered_total/g' \
  -e 's/rabbitmq_channel_messages_unroutable_dropped_total/rabbitmq_global_messages_unroutable_dropped_total/g' \
  -e 's/rabbitmq_channel_messages_unroutable_returned_total/rabbitmq_global_messages_unroutable_returned_total/g' \
  RabbitMQ-Overview.json
```

mkuratczyk:
Hi. Thanks for sharing this. I'll update the dashboards in the the next few days hopefully.

ggustafsson:
> Hi. Thanks for sharing this. I'll update the dashboards in the the next few days hopefully.

Any updates on this?

michaelklishin:
According to ZenHub, the Grafana dashboard update task is still open.

coro:
@ggustafsson I've just made the above PR to first get the dashboards into a healthier state - they've fallen behind some Grafana changes. My plan for next week is to then make the changes from aggregated to global metrics as you requested.

ggustafsson:
@michaelklishin @coro Thank you for the update! I’ll take a closer look at the PR next week. I can test each revision of the dashboard at work.

coro:
Thanks @ggustafsson. As a heads up, I haven't pushed anything to grafana.com just yet, but as of the changes in that PR you should be able to copy the dashboard JSON straight from the repo and paste it into the import menu in Grafana in order to test them.

coro:
The draft PR containing the global metrics is here: https://github.com/rabbitmq/rabbitmq-server/pull/5463

ggustafsson:
@coro I tried out the first PR and from what I can see it looks good, but it is a bit hard for me to do a good test because our metrics values are all over the place when we don't use the global metric paths. I will keep an eye on the second PR and try it out early. We have been using the existing dashboard with my `sed` fix above for around two months now and we haven't seen anything odd with it (yet).

PR #3138, opened 2021-06-23
Add bazel-test make target so that we can run tests in bazel quickly

gerhard:
How do we solve this one @pjk25?

```shell
    $ make bazel-test DEP=deps/rabbit SUITE=queue_type
    /usr/local/bin/bazel test //deps/rabbit:queue_type_SUITE --config=ci
    ERROR: Config value 'ci' is not defined in any .rc file
    make: *** [mk/bazel.mk:16: bazel-test] Error 2
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ endif
 include rabbitmq-components.mk
 include erlang.mk
 include mk/github-actions.mk
+include mk/bazel.mk
 include mk/topic-branches.mk
 
 # --------------------------------------------------------------------
--- a/mk/bazel.mk
+++ b/mk/bazel.mk
@@ -1,16 +1,41 @@
-BAZEL ?= /usr/local/bin/bazel
+BAZELISK ?= /usr/local/bin/bazelisk
 ifeq (darwin,$(PLATFORM))
-$(BAZEL):
-	brew install bazel
+$(BAZELISK):
+	brew install bazelisk
 else
-	$(error Install bazel for your platform: https://docs.bazel.build/versions/4.1.0/install.html)
+	$(error Install bazelisk for your platform: https://github.com/bazelbuild/bazelisk)
 endif
 
-bazel-test: $(BAZEL)
+define USER_BAZELRC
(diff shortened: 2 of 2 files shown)


pjk25:
@gerhard I started adding bazel to ra, and in doing so defined a make rule that will generate a basic `.bazelrc` file https://github.com/rabbitmq/ra/blob/32eaf3541371eb014a20fdef751a6961ffb2b285/Makefile#L51-L62

Said file doesn't allow for remote execution, however, as that requires credentials. We could use external credentials like [this](https://github.com/rabbitmq/rabbitmq-server/blob/c8c055e158934b13a191916d721aedabd3f3ff71/.github/workflows/test.yaml#L43-L44), but the `.bazelrc` still contains machine specific paths, and thus I have not come up with a good way of sharing a `.bazelrc` such that it could be checked in. There is no need to use `--config=ci` in the command in this case as it would be a local run.

But, at least your above `bazel-test` goal could depend on the generated file and avoid the error you see above.

pjk25:
With a bit of practice I have found invoking tests with bazel no less intuitive than using make. In fact I avoid using make if I can. So, I don't think I would use this make rule myself, but if it's an easy way to get started with bazel, I'm happy for it to be available.

For rabbitmq-server in particular, there are some additonal lines that ought to be in the generated `.bazelrc`, that aren't necessary for ra:
```
# rabbitmqctl wait shells out to 'ps', which is broken in the bazel macOS
# sandbox (https://github.com/bazelbuild/bazel/issues/7448)
# adding "--spawn_strategy=local" to the invocation is a workaround
build --spawn_strategy=local

# run one test at a time on the local machine
build --test_strategy=exclusive

# don't re-run flakes automatically on the local machine
build --flaky_test_attempts=1

build --incompatible_strict_action_env
```

Additionally, ra does not need elixir, so the generated rabbitmq-server `.bazelrc` would also need to interpolate `build --//:elixir_home=` with a correct path.

pjk25:
@gerhard I made another attempt at this by pushing a new commit

gerhard:
This looks great, I can hardly wait to use it! Thank you 👏🏻

PR #3143, opened 2021-06-24
Build OCI image for each supported Erlang major version

ansd:
As requested by @gerhard in https://github.com/rabbitmq/rabbitmq-server/pull/2777#issuecomment-866035696, auto-build OCI images for both Erlang 23 and 24.

Therefore, this job will create two images
1. for minimum required OTP
2. for maximum supported OTP

Each image will have two tags
1. git commit sha (i.e. the tip of the branch as requested by @michaelklishin)
2. branch name (this stable tag gets used for example in https://github.com/rabbitmq/cluster-operator/blob/a74ec732f31303c1e94f3ec30f3eb77961113c96/.github/workflows/pr.yml#L55).

For example, for Git commit SHA `1111aaa` and branch name `main`, the following two docker images and four tags will be pushed to DockerHub:
* `1111aaa-otp-min` (image 1)
* `main-otp-min` (image 1)
* `1111aaa-otp-max` (image 2)
* `main-otp-max` (image 2)

(While we could have used tag `main-otp-24.0.2` instead of `main-otp-max`, I prefer the latter option since we can use that stable tag when specifying the RabbitMQ image in pipelines or LREs.)

Needs to be backported to `v3.9.x` and `v3.8.x`.

### DIFF
--- a/.github/workflows/oci.yaml
+++ b/.github/workflows/oci.yaml
@@ -10,6 +10,7 @@ on:
       - Makefile
       - plugins.mk
       - rabbitmq-components.mk
+      - .github/workflows/oci.yaml
 env:
   GENERIC_UNIX_ARCHIVE: ${{ github.workspace }}/PACKAGES/rabbitmq-server-generic-unix-${{ github.sha }}.tar.xz
   RABBITMQ_VERSION: ${{ github.sha }}
--- a/.github/workflows/oci.yaml
+++ b/.github/workflows/oci.yaml
@@ -15,34 +15,38 @@ env:
   RABBITMQ_VERSION: ${{ github.sha }}
   VERSION: ${{ github.sha }}
 jobs:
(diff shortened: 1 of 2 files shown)


lhoguin:
Hm maybe instead of min/max we could have max, max-1 and max-2. We currently support only 23 and 24 but I think that is unusual and we typically support up to 3 major versions?

ansd:
Good idea @lhoguin. I pushed a commit implementing that.

michaelklishin:
We support two series and when we can, three for a 3 month long transition period. Still, N and N-1 is a good naming pattern to follow 👍

gerhard:
This is awesome! Thank you @ansd 👏

cc @Zerpet for the big-picture OCI view

gerhard:
Can someone backport to `v3.9.x` & `v3.8.x` please?

ansd:

Issue #3149, opened 2021-06-25
Logs on window inconsistent use of \ / in Path.


LindaLawton:
I was just checking the logs and its strange that the` config file(s)` would have Linux slashes / and the rest of them are using windows \ .

```
 Starting RabbitMQ 3.8.17 on Erlang 24.0.2
 Copyright (c) 2007-2021 VMware, Inc. or its affiliates.
 Licensed under the MPL 2.0. Website: https://rabbitmq.com
2021-06-25 10:21:30.207 [info] <0.282.0> 
 node           : rabbit@DESKTOP-R0C0P2L
 home dir       : C:\windows\system32\config\systemprofile
 config file(s) : c:/Users/linda/AppData/Roaming/RabbitMQ/rabbitmq.config
 cookie hash    : aDFhdX6hHRC6KFNpSOR9Xg==
 log(s)         : c:/Users/linda/AppData/Roaming/RabbitMQ/log/rabbit@DESKTOP-R0C0P2L.log
                : c:/Users/linda/AppData/Roaming/RabbitMQ/log/rabbit@DESKTOP-R0C0P2L_upgrade.log
```

Side note

    rabbitmq-diagnostics.bat status

Seems to be consistent with its use of /'s at least

```
Node data directory: c:/Users/linda/AppData/Roaming/RabbitMQ/db/rabbit@DESKTOP-R0C0P2L-mnesia
Raft data directory: c:/Users/linda/AppData/Roaming/RabbitMQ/db/rabbit@DESKTOP-R0C0P2L-mnesia/quorum/rabbit@DESKTOP-R0C0P2L

←[1mConfig files←[0m

 * c:/Users/linda/AppData/Roaming/RabbitMQ/rabbitmq.config

←[1mLog file(s)←[0m

 * c:/Users/linda/AppData/Roaming/RabbitMQ/log/rabbit@DESKTOP-R0C0P2L.log
 * c:/Users/linda/AppData/Roaming/RabbitMQ/log/rabbit@DESKTOP-R0C0P2L_upgrade.log
```

Yes this is probably just me nitpicking.

michaelklishin:
They are retrieved differently. This has been the case for years and this is the first time I recall someone bringing this up.
I'm sorry but our team has much more pressing things to work on.

lhoguin:
[...]

PR #3152, opened 2021-06-28
Canonicalize a few more paths in the startup banner

michaelklishin:
This canonicalises a few more paths (effective user's home directory, log file locations) that are printed
and logged in the startup banner. The goal is to make sure the paths use consistent slashes
on Windows. Note that `filename:absname/1` [use *nix-style slashes and not backslashes](https://erlang.org/doc/man/filename.html).

References #3149.


lhoguin (approved):
LGTM

michaelklishin:
Backported to `v3.8.x`.

michaelklishin:
Backported to `v3.9.x`.

PR #3167, opened 2021-06-30
Acquire locks before starting or restarting dynamic Shovels

michaelklishin:
When N cluster nodes are configured to import the same definition
file that includes Shovel declarations, there is a natural race
condition that can lead to Shovels being started on multiple
nodes under the same name.

This can be hard to spot but is guaranteed to be problematic for
Shovels that use an exchange as a source.

See #3154 for the background.


michaelklishin:

PR #3169, opened 2021-07-01
Do not update virtual host tracing state when it's not necessary

michaelklishin:
Primarily to avoid refreshing state of all channels since it is expensive in virtual hosts with a large number of channels.

Per discussion with @ansd.


michaelklishin:
Backported to `v3.8.x`.

michaelklishin:
Backported to `v3.9.x`.

ansd:
I manually tested with [rabbit-hole](https://github.com/michaelklishin/rabbit-hole/blob/555bcffe2314cd902beac8156b70f97270e672a5/vhosts.go#L168-L185) that the logic works.

PR #3043, opened 2021-05-18
Add consumer offset_lag to rabbitmq-stream CLI command & Management

gerhard:
This is an important metric to keep track of and be aware (maybe even alert on) when consumers fall behind consuming stream messages. While they should be able to catch up, if they fall behind too much and the stream gets truncated, they may miss on messages.

This is something that we want to expose via Prometheus metrics as well, but we've started closer to the core, CLI & Management.

While at it, we've made the commands use `PrettyTable by default`.

This should be merged as soon as it passes CI, we shouldn't wait on the Prometheus changes - they can come later.

Pair: @kjnilsson

image
image

PR #3177, opened 2021-07-08
Use "store" instead of "commit" for offset tracking

acogoluegnes:


gerhard:
This looks good! Should anything introduced in the rabbitmq CLI part of https://github.com/rabbitmq/rabbitmq-server/pull/3043 be revised part of this too?

acogoluegnes:
Good point. The `committed_offset` in the `stream_status` command is the chunk id (= first offset in the chunk) of the last (stored) chunk, so the committed offset for the replication replication. So this should not be changed, because it's not the same concept as the stored offset for consumer tracking.

This is `committed_offset` that triggered a discussion to change "commit" to "store" for consumer tracking.

pjk25:
This should be backported to the `v3.9.x` branch before 3.9 goes GA.

michaelklishin:
Backported to `v3.9.x`.

PR #3178, opened 2021-07-08
Pin prometheus.erl to hex 4.8.1

deadtrickster:


deadtrickster:
@pjk25 please check

pjk25:
@deadtrickster we can fetch from hex.pm in bazel too, so while you change would have worked, I made things consistent.

PR #3187, opened 2021-07-12
Change a log line from INFO to DEBUG

mkuratczyk:
This line is printed on every new MQTT connection which leads to very chatty logs when there is a lot of connections. Given that the way MQTT uses vhosts is generally static (once set up, always the same for all connections), I think this can be a debug message instead.


michaelklishin (approved):

michaelklishin:
Backported to `v3.9.x`.

michaelklishin:
Backported to `v3.8.x`.

PR #3188, opened 2021-07-14
Set info/2 timeout to infinity to list connections

dcorbacho:
Default gen_server timeout is not enough to list busy connections.
Setting it to infinity allows the caller to decide the timeout,
as classic queues do. The `emit_info` function's family sets its
own timeout for all the cli commands.

## Types of Changes

## Checklist


michaelklishin:
@dcorbacho can we bump it to, say, 30s or 60s but still not use `infinity`?

dcorbacho:
> @dcorbacho can we bump it to, say, 30s or 60s but still not use `infinity`?
@michaelklishin That's how we do it for classic queues at the moment, whatever we do they should behave the same

michaelklishin:
Backported to `v3.9.x`.

PR #3189, opened 2021-07-14
Bring the docker image up to date with the docker-library version

pjk25:
This is a companion to https://github.com/docker-library/rabbitmq/pull/501 and also adopts the simplified `docker-entrypoint.sh` recently introduced there.

~~I opted to copy the example conf from `deps/rabbit/docs/rabbitmq.conf.example` instead of modifying it in flight. I also considered adding it the `*-generic-unix package` so that yet another copy need not exist in the `docker-library` repo, however since this is a docker image specific config it did not seem reasonable to include it in the generic-unix archive. Since changes in the docker image currently flow from https://github.com/docker-library/rabbitmq to the `packaging/docker-image` directory of this repo, I think it is reasonable that the source of truth for this particular rabbitmq.conf file will be in practice the version in the `docker-library` repo.~~


michaelklishin (approved):

pjk25:
Backported to `v3.9.x`

PR #3194, opened 2021-07-15
Add stream reader state timeouts

ansd:
Add state timeouts.
If the client takes more than 10s for a single step in the authentication
protocol, make the server close the TCP connection.

Also close the TCP connection if the server times out in state
`close_sent`. That's the case when the client sends an invalid command
(after successful authentication), the server requests the client to
close the connection, but the client doesn't respond anymore.

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
@@ -396,7 +396,9 @@ tuned(info, Msg, StateData) ->
                 end).
 
 state_timeout(State, Transport, Socket) ->
-    rabbit_log_connection:warning("Closing connection because of timeout in state '~s'.", [State]),
+    rabbit_log_connection:warning(
+      "Closing connection because of timeout in state '~s' likely due to lack of client action.",
+      [State]),
     close_immediately(Transport, Socket),
     stop.
 
@@ -1004,7 +1006,9 @@ close_sent(state_timeout, close, #statem_data{
                                     connection = #stream_connection{socket = Socket} = Connection,
                                     connection_state = State
                                    }) ->
-    rabbit_log_connection:warning("Closing connection because of timeout in state '~s'.", [?FUNCTION_NAME]),
+    rabbit_log_connection:warning(
+      "Closing connection because of timeout in state '~s' likely due to lack of client action.",
+      [?FUNCTION_NAME]),
     close(Transport, Socket, State),
     rabbit_networking:unregister_non_amqp_connection(self()),
     notify_connection_closed(Connection, State),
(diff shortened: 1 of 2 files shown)


deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -146,6 +146,12 @@
          peer_cert_subject,
          peer_cert_validity]).
 
+-ifdef(TEST).
+-define(STATE_TIMEOUT, 500).
+-else.
+-define(STATE_TIMEOUT, 10_000).
```
michaelklishin, line 152: Should we be more specific about what states this covers?

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -949,6 +997,18 @@ open(cast, {force_event_refresh, Ref}, #statem_data{
...
     {keep_state, StatemData#statem_data{connection = Connection2}}.
 
+close_sent(enter, _OldState, StateData) ->
+    {next_state, ?FUNCTION_NAME, StateData, {state_timeout, ?STATE_TIMEOUT, close}};
+close_sent(state_timeout, close, #statem_data{
+                                    transport = Transport,
+                                    connection = #stream_connection{socket = Socket} = Connection,
+                                    connection_state = State
+                                   }) ->
+    rabbit_log_connection:warning("Closing connection because of timeout in state '~s'.", [?FUNCTION_NAME]),
```
michaelklishin, line 1007: It would be useful to indicate that this is likely due to a [lack of] client action.

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
ansd: Do you mean renaming the `STATE_TIMEOUT` macro?
We could, yes.
That timeout applies to states `tcp_connected`, `peer_properties_exchanged`, `authenticating`, `tuning`, `tuned`, `close_sent`. It does not apply to state `open`.
I can't come up with a good name, though 🤔
Do you have a suggestion?

michaelklishin: `CONNECTION_NEGOTIATION_STEP_TIMEOUT`?

ansd: That's a very good name, thank you 🙂
Renamed it.

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
ansd: Added that indication, thank you.

michaelklishin (approved):

michaelklishin:
Backported to `v3.9.x`.

PR #3195, opened 2021-07-15
Remove amqp_ssl:add_verify_fun_to_opts/2,3

britto:
## Proposed Changes

These functions have become no-op after d76234f. The `maybe_add_verify_fun/3` function has now become unreachable and was removed too.

## Types of Changes

## Checklist

## Further Comments

Since `add_verify_fun_to_opts/2` was exported, this might be a breaking change, so we could deprecate it instead.

I noticed that `maybe_add_verify/1` has also become mostly a no-op after d76234f, except for logging a warning message. This function is reachable from 2 locations: `amqp_connection.start/2` and `amqp_uri.parse/2`. I believe this warning could be valuable in the former but not much in the latter, especially if we parse before connecting, which would log it twice.

This is my first time contributing to an Erlang project. Please let me know if these changes make any sense at all.

### DIFF
--- a/deps/amqp_client/src/amqp_ssl.erl
+++ b/deps/amqp_client/src/amqp_ssl.erl
@@ -5,7 +5,6 @@
 -include_lib("public_key/include/public_key.hrl").
 
 -export([maybe_enhance_ssl_options/1,
-         add_verify_fun_to_opts/2,
          verify_fun/3]).
 
 maybe_enhance_ssl_options(Params = #amqp_params_network{ssl_options = none}) ->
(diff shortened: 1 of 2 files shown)


michaelklishin:
This looks reasonable but I'm afraid it's too late to get it into `3.9.x`, so we'll have to stick to the main branch for now.

michaelklishin (approved):

michaelklishin:
Thank you!

britto:
Thank you for looking into it so quickly, @michaelklishin! ❤️ 

Any thoughts on skipping the warning from `maybe_add_verify/1` while parsing URLs and restrict it only to connect-time? I could give it a shot.

michaelklishin:
@britto assuming that it would still be emitted for Shovels and Federation links, that sounds like a good idea.

PR #3204, opened 2021-07-20
Use keep_state_and_data

ansd:
Small refactor follow up from https://github.com/rabbitmq/rabbitmq-server/pull/3194


michaelklishin (approved):

michaelklishin:
Backported to `v3.9.x`.

PR #3190, opened 2021-07-14
Use layer caching when building the oci

pjk25:
## Proposed Changes

This speeds up the workflow that produces OCI images on each commit by leveraging the caching options afforded by https://github.com/marketplace/actions/build-and-push-docker-images

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask on the
mailing list. We're here to help! This is simply a reminder of what we are
going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [x] All tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in related repositories

PR #3213, opened 2021-07-21
Fix accept dependency in rabbitmq_prometheus

gerhard:
It's a runtime dependency, not a build dependency.

This is a fix and should be backported to `v3.9.x`, after rc.2 and just before the final release.

Would you disagree @dumbbell?

### DIFF
--- a/deps/rabbitmq_prometheus/Makefile
+++ b/deps/rabbitmq_prometheus/Makefile
@@ -19,10 +19,10 @@ endef
 
 PROJECT := rabbitmq_prometheus
 PROJECT_MOD := rabbit_prometheus_app
-DEPS = rabbit rabbitmq_management_agent prometheus rabbitmq_web_dispatch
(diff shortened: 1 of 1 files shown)


deadtrickster (approved):

gerhard:
Single reviewer is fine, first one wins.

Thanks @deadtrickster, that was a new record! Approved in 16 seconds 😲

Let's wait for all checks to pass before merging & back-porting.

gerhard:
I don't know what magic you have rubbed on that Bazel @pjk25, but I'm seriously digging the new GitHub Actions speed. Everything passed in 12 minutes! That is seriously impressive 👏

I am looking for the ones responsible for bringing the container image builds down to 4 minutes (they used to take at least 12 minutes before). They deserve a big thank you as well.

gerhard:
No way! It's also you @pjk25! https://github.com/rabbitmq/rabbitmq-server/pull/3190

# 👏👏 🚀

gerhard:
Backported to `v3.9.x`

PR #3229, opened 2021-07-23
Fix function_clause error in tracking_status/2

ansd:
Before this commit:
```
> ./sbin/rabbitmq-streams stream_status --tracking s1
Status of stream s1 on node rabbit@localhost ...
Error:
{:function_clause, [{:rabbit_stream_queue, :"-tracking_status/2-fun-0-", [:offsets, %{"s1-1" => 5}, []], [file: 'src/rabbit_stream_queue.erl', line: 608]}, {:maps, :fold_1, 3, [file: 'maps.erl', line: 410]}, {:rabbit_stream_queue, :tracking_status, 2, []}]}
```

After this commit:
```
> ./sbin/rabbitmq-streams stream_status --tracking s1
Status of stream s1 on node rabbit@localhost ...
┌────────┬───────────┬───────┐
│ type   │ reference │ value │
├────────┼───────────┼───────┤
│ offset │ s1-1      │ 51    │
└────────┴───────────┴───────┘
```

`rabbitmq-streams stream_status --tracking <stream>` also fails on `v3.9.0`.
Needs to be backported to `v3.9.x`.

GitHub action failures seem to be unrelated?

### DIFF
--- a/deps/rabbit/src/rabbit_stream_queue.erl
+++ b/deps/rabbit/src/rabbit_stream_queue.erl
@@ -605,10 +605,14 @@ tracking_status(Vhost, QueueName) ->
         {ok, Q} when ?amqqueue_is_stream(Q) ->
             Leader = amqqueue:get_pid(Q),
             Map = osiris:read_tracking(Leader),
-            maps:fold(fun(K, {Type, Value}, Acc) ->
-                              [[{reference, K},
-                                {type, Type},
-                                {value, Value}] | Acc]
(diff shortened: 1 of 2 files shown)


gerhard:
Great find & fix!

I am curious to know when this started failing. Was it before the 3.9.0 RCs or after? Maybe this bug was there even before the betas 🙇‍♂️

ansd:
The bug was there before the betas.

michaelklishin:
Backported to `v3.9.x`.

PR #3230, opened 2021-07-26
Fix channel rate limiter for ack messages

acogoluegnes:


pjk25:
@acogoluegnes I have updated some test timeouts based on the failures in the checks. A rebase should pick them up without conflict.

gerhard:
This looks good to me too. I kicked off the jobs again as there were failures in the last run. As soon as everything goes green, this is good to merge and backport to `v3.9.x`.

Does this need backporting to `v3.8.x` too? Which issue does this fix @acogoluegnes?

michaelklishin:
I think this is specific to `v3.9.x`.

gerhard:
@pjk25 does this fail often? Should we just re-run?

https://github.com/rabbitmq/rabbitmq-server/pull/3230/checks?check_run_id=3172111643#step:4:1487

pjk25:
> does this fail often? Should we just re-run?

It looks like the only hard failure was for `//deps/rabbit:unit_disk_monitor_mocks_SUITE`, in the form of a timeout. We should in this case update the size attribute for that test (many tests have needed this treatment recently as an artificial timeout extension was set by buildbuddy until last week). I will do so in master.

pjk25:
The timeout for //deps/rabbit:unit_disk_monitor_mocks_SUITE has been adjusted in master with 68361d5d79e93c843ea84aacd62745e72e16c055

pjk25:
FYI I have clicked re-run as it seemed the most expedient way to address the failure. Should we hit another timeout, I would advise correcting it in master and rebasing this branch to pick up both corrections.

gerhard:
Backported to `v3.9.x`

PR #3237, opened 2021-07-28
Update auth name

william00179:
In the examples given, the auth plugin is referred to as `rabbitmq_auth_backend_oauth2` which is incorrect. This is the name of the plugin, and what must be set in the `enabled_plugins` file. 

The name is correctly set in the demo files.

## Proposed Changes

Fix minor issue in the documentation that I spent too long to admit debugging.

## Types of Changes

- [x ] Documentation improvements (corrections, new content, etc)

## Checklist

- [x ] I have read the `CONTRIBUTING.md` document

### DIFF
--- a/deps/rabbitmq_auth_backend_oauth2/README.md
+++ b/deps/rabbitmq_auth_backend_oauth2/README.md
@@ -86,7 +86,7 @@ it will translate into the following configuration (in the [advanced RabbitMQ co
 [
   %% ...
   %% backend configuration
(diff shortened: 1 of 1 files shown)


michaelklishin:
The example [is correct](https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbitmq_auth_backend_oauth2/src/rabbit_auth_backend_oauth2.erl#L30) as far as I can tell.

The name of the key is the Erlang application name and it doesn't have to match any module names.

PR #3247, opened 2021-07-29
Add rebar3_hex plugin reference to rebar3.config generated by erlang.mk

deadtrickster:
This allows erlang.mk hex publish target to progress beyond `===> Command hex not found` error

### DIFF
--- a/erlang.mk
+++ b/erlang.mk
@@ -5645,6 +5645,7 @@ $(call comma_list,$(foreach d,$(DEPS),\
 		{$(call dep_name,$d)$(comma)"$(call dep_repo,$d)"},\
 		{$(call dep_name,$d)$(comma)".*"$(comma){git,"$(call dep_repo,$d)"$(comma)"$(call dep_commit,$d)"}})))
 ]}.
+{plugins, [rebar3_hex]}.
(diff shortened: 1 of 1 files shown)


gerhard:
Forgot to add the `backport v3.9.x` label before closing this. Doing it manually now, but will remember to add it next time.

gerhard:
Backported to `v3.9.x` & `v3.8.x`

michaelklishin:
We should perhaps upstream this? @dumbbell @lhoguin

deadtrickster:
Turns out we don't use bare rebar3, but our own wrapper - hexpm-cli, It has hardcoded config with rebar3_hex. I created PR to upgrade it to recent rebar3 with OTP 24 support - https://github.com/rabbitmq/hexpm-cli/pull/2

lhoguin:
For the record: no this kind of change should not be added to Erlang.mk. Erlang.mk produces a rebar.config file that's meant to be used for projects used as dependencies. There is no problem in that case because the plugins can be added to the top-level rebar.config file.

dumbbell:
Also, `erlang.mk` is a generated file. We should revert this change now instead of being caught by surprise with some obscure regressions when we update Erlang.mk.

PR #3250, opened 2021-07-29
Publish RabbitMQ-Stream dashboard to grafana.com

gerhard:
- Removed the Dockerfile
- Slimmed down the Makefile

All of this is now handled by https://github.com/rabbitmq/rabbitmq-server/blob/master/.github/workflows/oci.yaml
cc @Zerpet @pjk25

More details here (including the steps used to publish to grafana.com): https://github.com/rabbitmq/release-engineering/issues/11#issuecomment-887627938

I don't want to hold up this PR, will invest in automating the steps described in the previous link another time.


deps/rabbitmq_prometheus/docker/docker-compose-overview.yml
```diff
@@ -173,6 +170,11 @@ services:
       CONSUMERS: 0
       SERVERS_STARTUP_TIMEOUT: *startup_timeout
       METRICS_PROMETHEUS: "true"
+  stream:
+    image: pivotalrabbitmq/stream-perf-test
+    networks:
+      - "rabbitmq-prometheus"
+    command: "--uris rabbitmq-stream://guest:guest@rmq2:5552/%2f --max-length-bytes 100mb --rate 1000"
```
ansd, line 177: When I run `make overview`, container `docker_stream_1` gets created but exits immediately with
```
> docker logs docker_stream_1
...
Connection refused: rmq2/172.21.0.12:5552
...
```
Therefore, I'm unable to see any metrics in the Grafana Streams dahsboard.

(Running `make overview` a second time after a couple of seconds will succeed and will show up stream metrics.)

Adding
```
    depends_on:
      - "rmq2"
```
does not help since `rmq2` takes too long to open port 5552.

I suggest to either make stream-perf-test retry to connect or to add a wait wrapper script to the stream-perf-test image as described in https://docs.docker.com/compose/startup-order/ (cc @acogoluegnes).

ansd (changes requested):
Thanks a lot @gerhard! Looks good!

I pushed one [commit](https://github.com/rabbitmq/rabbitmq-server/pull/3250/commits/47d572908ded4c921a9a78f4958726f2f74ff3d1) since `make overview` was failing for me.

I also left one in-line comment since the stream metrics don't show up for me in the Grafana dashboards when running `make overview metrics`.

deps/rabbitmq_prometheus/docker/docker-compose-overview.yml
ansd: We can also merge this PR. We should make this change in https://github.com/rabbitmq/rabbitmq-stream-java-client/blob/main/Dockerfile ?

gerhard: Good spot, let me know if https://github.com/rabbitmq/rabbitmq-server/pull/3250/commits/35a6369327a07b1957550b2fb6e57ab34053a054 fixes it for you too.

ansd: I just tried adding a one-liner `        wait-for-it \` to https://github.com/rabbitmq/rabbitmq-stream-java-client/blob/8866cebafda34c15104f3557b3df8d74037abcfb/Dockerfile#L67 and changed the docker compose file to:
```
  stream:
    image: pivotalrabbitmq/stream-perf-test
    networks:
      - "rabbitmq-prometheus"
    depends_on:
      - "rmq2"
    entrypoint: ["wait-for-it"]
    command: ["rmq2:5552", "--timeout=180", "--",
        "java", "-Dio.netty.processId=1", "-jar", "stream-perf-test.jar",
        "--uris", "rabbitmq-stream://guest:guest@rmq2:5552/%2f",
        "--max-length-bytes", "100mb",
        "--rate", "1000"]

```
which works fine. It takes around 20 seconds until the port is open.

Your approach in https://github.com/rabbitmq/rabbitmq-server/commit/35a6369327a07b1957550b2fb6e57ab34053a054 works for me. It restarts the container a couple of times which is fine. It's the far simper option :)

ansd (approved):

gerhard:
I'm trying out https://github.com/tibdex/backport for the first time. Let's see how well it works @pjk25 😉

gerhard:
I'm not sure why the above backport failed. I could do the cherry-pick locally with no issues. The focus is on producing 3.9.1, so I'll stick with that and maybe come back to this later. For now I did a manual backport to `v3.9.x`, as per usual.

pjk25:
It looks like the backport action might just be broken. Rather than spend much time investigating I'm tempted to give https://mergify.io/features/backports a try. I will reconfigure things when I have a chance.

PR #3263, opened 2021-08-04
Partially reintroduce locking to mirrored_supervisor

### LINKED ISSUE #3260: Duplicate dynamic shovels can be started in certain scenarios
Originally reported in #3132 and #3154 that @mkuratczyk and I could reproduce (hopefully reliably) using the Kubernetes cluster Operator.

The steps are:

 * Form cluster S, the source
 * Form cluster D, the destination
 * Configure a Shovel in the D cluster so that it uses an **exchange** source in cluster S
 * Verify that the shovel is running and functional
 * For every node, put it under maintenance then restart
 * Observe duplicate mirrored supervisor children with `rabbitmqctl eval 'mirrored_supervisor:which_children(rabbit_shovel_dyn_worker_sup_sup).'`

My testing suggests that just a restart is not enough but putting a node under maintenance mode first at least makes the issue
significantly more likely to resurface. Something like this:

```
for n in rabbit hare flopsy
do
  rabbitmq-upgrade drain -n $n@hostname
  rabbitmqctl stop_app -n $n@hostname
  rabbitmqctl start_app -n $n@hostname
done
```

michaelklishin:
## Proposed Changes

Unlike the now removed `pg2` module, `pg` in Erlang 24 does not acquire locks when group
membership changes. It is strongly eventually consistent by design. `mirrored_supervisor`
implicitly relied on the `pg2` locking behavior and without it, duplicate children can
be started in some scenarios.

This reintroduces some of the locking performed by `pg2`. For the use cases of `mirrored_supervisor`,
locking across all connected nodes is acceptable.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #3260)

## Checklist

## Further Comments

Closes #3260. References #3132, #3154.

### DIFF
--- a/deps/rabbit_common/src/mirrored_supervisor.erl
+++ b/deps/rabbit_common/src/mirrored_supervisor.erl
@@ -278,6 +278,7 @@ handle_call({init, Overall}, _From,
                            initial_childspecs = ChildSpecs}) ->
     process_flag(trap_exit, true),
     LockId = mirrored_supervisor_locks:lock(Group),
+    maybe_log_lock_acquisition_failure(LockId, Group),
     ok = pg:join(Group, Overall),
     rabbit_log:debug("Mirrored supervisor: initializing, overall supervisor ~p joined group ~p", [Overall, Group]),
     Rest = pg:get_members(Group) -- [Overall],
@@ -309,6 +310,7 @@ handle_call({start_child, ChildSpec}, _From,
                            group    = Group,
                            tx_fun   = TxFun}) ->
     LockId = mirrored_supervisor_locks:lock(Group),
+    maybe_log_lock_acquisition_failure(LockId, Group),
     rabbit_log:debug("Mirrored supervisor: asked to consider starting a child, group: ~p", [Group]),
     Result = case maybe_start(Group, TxFun, Overall, Delegate, ChildSpec) of
                  already_in_mnesia ->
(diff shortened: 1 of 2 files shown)


deps/rabbit_common/src/mirrored_supervisor.erl
```diff
@@ -277,13 +277,16 @@ handle_call({init, Overall}, _From,
                            tx_fun             = TxFun,
                            initial_childspecs = ChildSpecs}) ->
     process_flag(trap_exit, true),
+    LockId = mirrored_supervisor_locks:lock(Group),
```
binarin, line 280: Should the return value be checked here? If the `lock` didn't actually get lock in the expected 300s?

michaelklishin: It's a good point. `mirrored_supervisor:unlock/1` will be a no-op if the lock id is `undefined`.

michaelklishin: This is a convention we use on top of `global:set_lock/3` in peer discovery so migrated to Shovel and `mirrored_supervisor` as well.

binarin: So it's not a mandatory lock, just some relaxed check. Probably the way the code is structured will never cause it fail - until one day it will. Maybe a warning should be logged in such a situation?

lhoguin: Seconded. Since not having a lock can cause issues, we would like to know if users who run into issues have obtained a lock.

lhoguin (approved):
LGTM otherwise.

deps/rabbit_common/src/mirrored_supervisor.erl
michaelklishin: Added a warning

michaelklishin:

michaelklishin:
Note: `v3.8.x` requires conflict resolution and some massaging to compile on Erlang 24 (it relies on Lager) => will backport manually.

michaelklishin:
Backported to `v3.8.x`.

michaelklishin:
Backported to `v3.9.x`.

PR #3281, opened 2021-08-09
Logging: Fix crash when Epoch-based timestamps are used with JSON

dumbbell:
The code was passing a number (the timestamp) to unicode:characters_to_binary/1 which expects an iolist to convert to UTF-8.

We now verify if we have a number before calling that function. If this is a number (integer or float), we keep it as is because JSON supports that type.


dumbbell:
@Mergifyio backport v3.9.x

PR #3282, opened 2021-08-09
Logging: Always represent Epoch-based timestamps as integers

dumbbell:
Before this change, seconds were represented as floats.


dumbbell:
This change could be considered a breaking change. However, given the JSON logging feature is undocumented for now, I believe it can be backported.

What do you thinkn @gerhard and @ansd?

gerhard:
Yes, and https://github.com/rabbitmq/rabbitmq-server/pull/3283#issuecomment-896065132

dumbbell:
@Mergifyio backport v3.9.x

PR #3283, opened 2021-08-09
Logging: Use `:` as the separator in verbosity mapping

dumbbell:
This is the same separator as the field mapping. I don't remember why I picked a different character... Now, it feels awkward and non-intuitive for users.


dumbbell:
This change could be considered a breaking change. However, given the JSON logging feature is undocumented for now, I believe it can be backported.

What do you thinkn @gerhard and @ansd?

gerhard:
Yes, backporting sounds right, the reasoning is sound and accurate.

dumbbell:
Yes, I used to backport other patches :)

dumbbell:
@Mergifyio backport v3.9.x

PR #3284, opened 2021-08-09
Revert "CQ: don't deliver right before acking in the index"

michaelklishin:
This reverts commit 3ef858746ceba1e27167264351d8450f5fc780b4.

This change does not handle expired messages safely.

See https://github.com/rabbitmq/rabbitmq-server/discussions/3272


michaelklishin:

PR #3298, opened 2021-08-11
Logging: Disable burst limit if the log level is set to `debug`

dumbbell:
Otherwise, messages are being dropped, in particular during RabbitMQ startup because of the amount of debug messages logged in that phase.

Burst limit is enabled again if the log level is set to `info` or higher.


michaelklishin:
@mergifyio backport v3.9.x

PR #3299, opened 2021-08-11
Add support to override `default_{user,pass,vhost}` and the Erlang cookie from the environment

dumbbell:
Historically, the Docker image allowed the user to override the default user/password and vhost as well as the Erlang cookie from environment variables among other things. After some discussion in https://github.com/docker-library/rabbitmq/issues/506#issuecomment-887786516, support for environment variables specific to the Docker image was removed in docker-library/rabbitmq#467.

Several users were impacted by this removal (see docker-library/rabbitmq#508). After some more discussion, it was decided to re-introduce some variables directly into RabbitMQ itself. The following variables are added by this patch:
* `$RABBITMQ_DEFAULT_USER` (overrides `default_user`)
* `$RABBITMQ_DEFAULT_PASS` (overrides `default_pass`)
* `$RABBITMQ_DEFAULT_VHOST` (overrides `default_vhost`)
* `$RABBITMQ_ERLANG_COOKIE` (overrides the Erlang cookie file content)

Fixes docker-library/rabbitmq#508

### DIFF
--- a/deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_dist.erl
+++ b/deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_dist.erl
@@ -56,8 +56,8 @@ do_setup(#{nodename := Node,
     %% Override the Erlang cookie if one was set in the environment.
     case maps:get(erlang_cookie, Origins, default) of
         environment ->
-            ?LOG_DEBUG(
-               "Override Erlang cookie using the one set in the environment",
+            ?LOG_WARNING(
+               "Overriding Erlang cookie using the value set in the environment",
                #{domain => ?RMQLOG_DOMAIN_PRELAUNCH}),
             Cookie = maps:get(erlang_cookie, Config),
             ?assert(is_atom(Cookie)),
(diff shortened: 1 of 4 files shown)


deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_dist.erl
```diff
@@ -49,6 +52,19 @@ do_setup(#{nodename := Node, nodename_type := NameType}) ->
             {ok, _} = net_kernel:start([Node, NameType]),
             ok
     end,
+
+    %% Override the Erlang cookie if one was set in the environment.
+    case maps:get(erlang_cookie, Origins, default) of
+        environment ->
+            ?LOG_DEBUG(
```
michaelklishin, line 59: Should this be a warning? It can be a painful thing to troubleshoot, since `rabbitmq-diagnostics` and friends depend on a cookie match.

michaelklishin:
It works as expected:

``` shell
RABBITMQ_DEFAULT_VHOST=pr-3299-vhost RABBITMQ_DEFAULT_USER=pr-3299 RABBITMQ_DEFAULT_PASS=pr-3299 gmake run-broker
```

``` shell
rabbitmqctl list_users
# => Listing users ...
# => user	tags
# => pr-3299	[administrator]
rabbitmqctl list_vhosts
# => Listing vhosts ...
# => name
# => pr-3299-vhost
```

michaelklishin:
``` shell
RABBITMQ_ERLANG_COOKIE=abcd gmake run-broker
```

Makes CLI authentication fail unless `--erlang-cookie abcd` is provided.

michaelklishin:
@mergifyio backport v3.9.x

Issue #3307, opened 2021-08-13
Latest rabbit_common support OTP 22/23/24


vkatsuba:
Hi All,

The [mirrored_supervisor.erl](https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit_common/src/mirrored_supervisor.erl) of theatest version of [rabbit_common](https://github.com/rabbitmq/rabbitmq-server/tree/master/deps/rabbit_common) can working only on OTP versions 23 and 24, this is related to changes [pg2 => pg for OTP 24 compatibility ](https://github.com/rabbitmq/rabbitmq-server/commit/98724eff09adb765d407a9f6787bdd774508879e). The main issue is that the [pg](https://erlang.org/doc/man/pg.html) module was added in OTP 23 and not exist in OTP 22. So, currently latest version of `rabbit_common` is supported OTP 23/24.

Regards,
--V

vkatsuba:
Fix https://github.com/rabbitmq/rabbitmq-server/pull/3308.

michaelklishin:
I'm sorry but we are not interested in adding this kind of complexity. Erlang 22 is no longer supported by RabbitMQ. Those who need to use `rabbit_common` on Erlang 22 can use an older version.

michaelklishin:
By the way, `pg` and `pg2` may seem to be almost a drop-in replacement of one another but they are not. `pg2` relies on distributed locks and `pg` is strongly eventually consistent by design. This [can matter a great deal](https://github.com/rabbitmq/rabbitmq-server/pull/3263). I would not recommend anyone to assume that a macro shim would be safe to use or provide in a library.

vkatsuba:
@michaelklishin thanks for code review and for providing explanation. Nevertheless it sounds sad, a lot of libraries try to support the last three versions and this means that everyone who uses this library will have to make a choice stop support OTP 22 or not etc. But in any case, many thanks for info.

Regards,
--V

PR #3308, opened 2021-08-13
rabbit_common add support for OTP 22

vkatsuba:
PR is related to https://github.com/rabbitmq/rabbitmq-server/issues/3307

### DIFF
--- a/deps/rabbit_common/src/mirrored_supervisor.erl
+++ b/deps/rabbit_common/src/mirrored_supervisor.erl
@@ -124,6 +124,22 @@
          delete_child/2, terminate_child/2,
          which_children/1, count_children/1, check_childspecs/1]).
 
+-ifdef(OTP_RELEASE).
+-if(?OTP_RELEASE >= 23).
+-define(PG_START_LINK(), pg:start_link()).
+-define(PG_JOIN(Group, PID), pg:join(Group, PID)).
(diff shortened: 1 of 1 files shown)


michaelklishin:
I'm sorry but we are not interested in adding this kind of complexity. Erlang 22 is no longer supported by RabbitMQ. Those who need to use `rabbit_common` on Erlang 22 can use an older version.

michaelklishin:
By the way, `pg` and `pg2` may seem to be almost a drop-in replacement of one another but they are not. `pg2` relies on distributed locks and `pg` is strongly eventually consistent by design. This [can matter a great deal](https://github.com/rabbitmq/rabbitmq-server/pull/3263). I would not recommend anyone to assume that a macro shim would be safe to use or provide in a library.

vkatsuba:
@michaelklishin thanks for code review and for providing explanation. Nevertheless it sounds sad, a lot of libraries try to support the last three versions and this means that everyone who uses this library will have to make a choice stop support OTP 22 or not etc. But in any case, many thanks for info.

Regards,
--V

michaelklishin:
If the choice is between dropping support for 22 or not adopting 24, the answer is obvious: OTP 24 has so much to offer that people will be picking it over OTP 22 any day of the week.

PR #3312, opened 2021-08-16
Expose alarms as Prometheus metrics

### LINKED ISSUE #2653: Expose resource alarms as Prometheus metrics
Alarms could make it into a core metrics table as binary entries and be reported by Prometheus. 
CC @MarcialRosales

deadtrickster:
## Proposed Changes

This exposes alarms as Prometheus metrics

## Types of Changes

## Checklist

## Further Comments

Closes #2653


deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_alarm_metrics_collector.erl
```diff
@@ -0,0 +1,80 @@
...
+                                #{},
+                                Alarms),
+
+                Callback(create_mf(?METRIC_NAME(<<"file_descriptor_limit">>),
+                                   <<"is 1 if file descriptor limit alarm is in effect">>,
+                                   untyped,
+                                   [untyped_metric(maps:get(file_descriptor_limit,
+                                                            ActiveAlarms,
+                                                            0))])),
+                Callback(create_mf(?METRIC_NAME(<<"disk_limit">>),
```
michaelklishin, line 54: Can this be renamed to "free_disk_space_watermark"?

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_alarm_metrics_collector.erl
```diff
@@ -0,0 +1,80 @@
...
+                                   <<"is 1 if disk alarm is in effect">>,
+                                   untyped,
+                                   [untyped_metric(maps:get(disk_limit, ActiveAlarms, 0))])),
+                Callback(create_mf(?METRIC_NAME(<<"memory_limit">>),
```
michaelklishin, line 58: Can this be renamed to "memory_used_watermark"?

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_alarm_metrics_collector.erl
```diff
@@ -0,0 +1,80 @@
...
+                                        maps:put(disk_limit, 1, Acc);
+                                    ({{resource_limit, memory, _}, _}, Acc) ->
+                                        maps:put(memory_limit, 1, Acc);
+                                    ({file_descriptor_limit, _}, Acc) ->
+                                        maps:put(file_descriptor_limit, 1, Acc)
+                                end,
+                                #{},
+                                Alarms),
+
+                Callback(create_mf(?METRIC_NAME(<<"file_descriptor_limit">>),
```
michaelklishin, line 48: "file_descriptors_used" would sound slightly more specific to me

michaelklishin (changes requested):
This works as expected but I think we can do better than "X_limit" for alarm names. Except for, perhaps, the file descriptor alarm, they do not represent limits but just watermarks.

Disk space can continue to decreased after the configured (low) watermark is crossed.

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_alarm_metrics_collector.erl
michaelklishin: Or maybe specifically here "file_descriptors_limit" makes more sense. "_used" may suggest to some that it is the number of descriptors used, since booleans are represented as integers.

deadtrickster:
With name I was following the alarms names. Really straightforward mapping: {resource_limit, disk, _} -> rabbitmq_alarms_disk_limit. 
That being said, these pages https://www.rabbitmq.com/memory.html and https://www.rabbitmq.com/disk-alarms.html point towards some inconsistency in naming - namely vm_memory_high_watermark.relative = 0.4 vs disk_free_limit.absolute = 1000000000. 
Now I wonder if we should have these 3:
* `rabbitmq_alarms_file_descriptor_limit`
* `rabbitmq_alarms_vm_memory_high_watermark`
* `rabbitmq_alarms_disk_free_limit`.

I.e. follow rabbitmqctl/config conventions?

michaelklishin:
There is no convention around configuration setting names: they were introduced at different points without much focus on making them consistent.

I kept `file_descriptor_limit` as is because it's the most precise name I could think of and changed the other two to say "watermark" instead of "limit" because they both are watermarks/thresholds.

michaelklishin:

michaelklishin:
Backported manually to `v3.8.x` for `3.8.25`.

PR #3314, opened 2021-08-17
Observer CLI plugin: classic queues

lhoguin:
We can write `observer_cli` plugins. Here's one for classic queues.

I see these as shortcuts for us to retrieve information about a system we need to debug, especially for cases where we would need to deal with customers directly and ask them to retrieve information from the Erlang node. But it could be used for any useful information to retrieve from the system.

## Types of Changes

## Checklist

- [ ] I have added this change to the first version(s) in release-notes that I expect to introduce it

### DIFF
--- a/deps/rabbit/src/rabbit.erl
+++ b/deps/rabbit/src/rabbit.erl
@@ -251,6 +251,12 @@
                     {requires,    [core_initialized, recovery]},
                     {enables,     routing_ready}]}).
 
+-rabbit_boot_step({rabbit_observer_cli,
+                   [{description, "Observer CLI configuration"},
+                    {mfa,         {rabbit_observer_cli, init, []}},
+                    {requires,    [core_initialized, recovery]},
+                    {enables,     routing_ready}]}).
+
 -rabbit_boot_step({pre_flight,
                    [{description, "ready to communicate with peers and clients"},
                     {requires,    [core_initialized, recovery, routing_ready]}]}).
--- a/deps/rabbit/src/rabbit_classic_queue_index_v2.erl
+++ b/deps/rabbit/src/rabbit_classic_queue_index_v2.erl
@@ -9,7 +9,7 @@
 
 -export([erase/1, init/3, reset_state/1, recover/7,
          terminate/3, delete_and_terminate/1,
-         publish/7, publish/8, ack/2, read/3]).
+         info/1, publish/7, publish/8, ack/2, read/3]).
 
 %% Recovery. Unlike other functions in this module, these
 %% apply to all queues all at once.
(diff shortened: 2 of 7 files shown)


lhoguin:
Example screenshot:

Screenshot 2021-08-17 at 16 32 43

lhoguin:
More recent screenshot:

Screenshot 2021-08-19 at 15 01 28

lhoguin:
Next steps:

* Wait/Help with https://github.com/zhongwencool/observer_cli/issues/78
* Wait/Help with https://github.com/zhongwencool/observer_cli/issues/79
* Create more plugins/panels (I can think of one more for classic queues)

Let's see what @gerhard thinks about this as well.

gerhard:
Oh wow, I can't believe that I haven't thought about this! This would be an amazing addition. Love it! 💙

gerhard:
Relevant: https://github.com/rabbitmq/rabbitmq-server/commit/09b92bacedc85c0878480feb559f2c57bb22d168#diff-18813c86948efc57e661623d7ba48ff94325c9b5421ec9177f724922dd553a35R35

gerhard:
> @lhoguin The hard part moving forward is determining what values are the most useful.

This looks like a good starting point to me:
- vhost
- name
- reductions / s
- process memory
- minor GCs
- major GCs
- process mailbox
- gen_server2 buffer

lhoguin:
Following the work on CQv2 and beyond the needs for this type of plugin have changed. I propose the following fields:

- queue pid/vhost/name
- mailbox + GS2 mailbox size
- # messages in memory (q3)
- (v2 only) index buffer size (with/without acks?) (delta, not yet flushed to disk)
- (v2 only) store buffer size (delta, not yet flushed to disk)
- # messages on disk (delta, flushed to disk)
- # messages in transit (pending acks)
- # messages to be confirmed

With that we should have a clear view of where messages are in each CQ.

lhoguin:
To test, create some work using classic queues, then connect via `rabbitmq-diagnostics observer` and press `P` and `Enter` to access the plugin.

lhoguin:
Screenshot from 2022-10-07 11-52-28

lhoguin:
Hold on before merge there's a small visual bug when there are many queues.

michaelklishin (approved):

illotum:
Fantastic addition. Any chance to backport it into 3.9?

michaelklishin:
It cannot even be backported to 3.11 as it targets a CQv2 revision scheduled to ship in 3.12.0.

lhoguin:
I will soon open a PR for 3.11 and 3.10.

PR #3323, opened 2021-08-19
Set terminal titlebar to the node name & RabbitMQ version

dumbbell:
With this change and if the RabbitMQ node is running on Unix and accepts input, the titlebar of an Xterm-compatible terminal emulator will show a few details about the running node. Specifically, it will indicate the name of the node and the version of RabbitMQ.

This is a change I have locally in the Khepri branch for quite some time. Time to extract it from that branch and see if it can help other people!

2021-08-19T13:09:27+02:00

### DIFF
--- a/deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
+++ b/deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_boot_state.erl
@@ -19,7 +19,7 @@
          has_reached/1,
          has_reached_and_is_active/1]).
 
--define(PT_KEY_BOOT_STATE,    {?MODULE, boot_state}).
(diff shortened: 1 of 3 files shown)


michaelklishin:
It works as expected but with iTerm 2, the visibility of this title is questionable to me. This will vary from terminal to terminal, of course.

@dumbbell we can merge if you find this useful. I wonder if there is any risk to e.g. Windows users?

dumbbell:
Thank you for testing!

This is only enabled if `os:type()` returns `{unix, _}`, so it should not impact Windows users. Also, if the code crashes for whatever reason, RabbitMQ is unaffected.

On Mac OSX, I only tested with Alacritty. Is iTerm 2 the most popular terminal on this platform?

michaelklishin:
@dumbbell as far as I can tell iTerm 2 and the standard Termnal are the two most popular options on macOS.

PR #3324, opened 2021-08-19
rabbit_{connection,channel}_tracking: Fix race condition in list()

dumbbell:
The table might not exist yet (or is already gone) between the time `rabbit_nodes:all_running()` runs and returns a specific node, and `mnesia:dirty_match_object()` is called for that node's table.

This seems to happen frequently in CI.


dumbbell:
If this patch is accepted, I believe it's a good candidate for backport to 3.9.x and 3.8.x.

michaelklishin (approved):

michaelklishin:

PR #3340, opened 2021-08-26
Improve stream reader metrics cleanup

kjnilsson:
When stream readers crash, we want to cleanup metrics, otherwise stream consumer and publisher metrics will diverge from reality, and nodes will report a higher number of these than there are in reality.

We cleaned up some logs in the process, and the `replica_recovery` test in rabbit_stream_queue SUITE was made more reliable.


michaelklishin:
@mergifyio backport v3.9.x

PR #3342, opened 2021-08-27
Make run-broker: set default node name to rabbit@localhost

deadtrickster:
This way code_reload via lsp works for everybody, regardless of actual hostnames


michaelklishin:
This makes sense but won't work well in practice. `gmake start-broker` does pick up the new name but CLI tools do not and changing CLI tool node name resolution default is incredibly risky.

I suggest that you override `RABBITMQ_NODENAME` system-wide if that'd be more convenient.

PR #3355, opened 2021-09-01
rabbitmq_cli: Fix `maintenance_mode_status` feature flag name in testsuite

dumbbell:
This didn't really affect the testsuite because all feature flags are enabled earlier if I understand correctly.


dumbbell:
@Mergifyio backport v3.9.x v3.8.x

PR #3360, opened 2021-09-01
Fix store offset parameters

korsmakolnikov:
As you can see https://github.com/rabbitmq/rabbitmq-server/blob/860333a08816e05749ebfbdad32a7da66d5c9b3f/deps/rabbitmq_stream_common/src/rabbit_stream_core.erl#L239

There is the stream name and not the subscription id in this message.

## Types of Changes

- [ x] Documentation improvements (corrections, new content, etc)


acogoluegnes:
Good catch, thanks!

acogoluegnes:
@Mergifyio backport v3.9.x

PR #3363, opened 2021-09-02
Update Github Actions workflow triggering to include bazel build

pjk25:
Github Actions Test and Test Mixed Versions workflows should trigger when bazel files change.

(opened as a PR for convenience of backporting).


pjk25:
@Mergifyio backport v3.9.x

PR #3378, opened 2021-09-07
Add debug utilities to Docker image

gerhard:
We need these to debug various rabbitmq-related issues. Ephemeral debug containers are not enabled on GKE Autopilot (not sure if GKE enables this alpha feature at all), so we are doing this instead.

```
 error: ephemeral containers are disabled for this cluster (error from server: "the server could not find the requested resource").
```

Pair @Gsantomaggio

---

Backport to v3.9.x & v3.8.x if build passes

### DIFF
--- a/.github/workflows/test-mixed-versions.yaml
+++ b/.github/workflows/test-mixed-versions.yaml
@@ -7,7 +7,6 @@ on:
       - v3.8.x
     paths:
       - 'deps/**'
-      - 'packaging/**'
       - 'scripts/**'
       - Makefile
       - plugins.mk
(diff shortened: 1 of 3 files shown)


pjk25 (approved):

pjk25:
@Mergifyio backport v3.9.x

pjk25:
@gerhard I was wondering why mergify didn't pick up on the label, the above explains why. Looks like this one will need a manual backport.

gerhard:
Yup, thanks, I will do it shortly 👍🏻

gerhard:
Yup, thanks, I will do it shortly 👍🏻

PR #3379, opened 2021-09-07
Use master branch of rabbitmq-ct-helpers and rabbitmq-ct-client-helpers

pjk25:
## Proposed Changes

Historically release branches, like `v3.7.x` or `v3.8.x`, existed across rabbit, the plugins, and various helpers. With the monorepo there are fewer repos, but the convention was held for rabbitmq-ct-helpers. Recently we ran into some issues because commits to rabbitmq-ct-helpers were not backported to all necessary branches. I have since update the helpers to make them sufficiently version agnostic that all rabbitmq-server branches can use the master branch of rabbitmq-ct-helpers and rabbitmq-ct-client-helpers

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [x] CI

## Checklist

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

PR #3387, opened 2021-09-08
Fix mixed version testing for 3.7/3.8

pjk25:
The `v3.8.x` branch of https://github.com/rabbitmq/rabbitmq-ct-helpers and https://github.com/rabbitmq/rabbitmq-ct-helpers was out of date such that the mixed version tests never actually used 3.7 code. This adopts master of https://github.com/rabbitmq/rabbitmq-ct-helpers and fixes the resulting breakage, so that we should be able to avoid this scenario again.


pjk25:
This is equivalent of #3379 for the `v3.8.x` branch

PR #3398, opened 2021-09-10
Revisit two rabbitmq.conf validators

### LINKED ISSUE #3390: rabbitmq.conf validation fails when max_message_size is set to 536870912
Hello,

When I set `max_message_size = 536870912` in RabbitMQ configuration file, it fails to start with the following logs:

> 2021-09-09 10:37:54.290911+00:00 [erro] <0.130.0> Error preparing configuration in phase validation:
2021-09-09 10:37:54.290987+00:00 [erro] <0.130.0>   - max_message_size invalid, Max message size should be less than 512MB and gre than 0

If I set it to `536870911`, it starts fine.

I'm surprised because the max value for this parameter is supposed to be `536870912` according to https://www.rabbitmq.com/configure.html.

I've reproduced this issue with two Docker images (didn't check other ones): rabbitmq:3.8.7-management and rabbitmq:3.9.5-management.

Thanks for your attention.

michaelklishin:
* max_message_size had an off-by-one error and unfortunate naming
 * classic mirrored queue batch size was not validating the size in messages.
   The limit of over 2B messages did not make much sense. 1M is a still very
   high but a more reasonable upper bound

Fixes #3390

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -743,7 +743,7 @@ end}.
 
 
 {mapping, "max_message_size", "rabbit.max_message_size",
-    [{datatype, integer}, {validators, ["less_then_512MB"]}]}.
+    [{datatype, integer}, {validators, ["max_message_size"]}]}.
 
 %% Customising Socket Options.
 %%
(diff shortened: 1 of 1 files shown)


michaelklishin:
Backported to `v3.8.x` manually.

PR #3392, opened 2021-09-09
Update Ranch to 2.1

lhoguin:
Update Ranch to 2.1 and refactor newly added code to make use of an equivalent function in Ranch.

Changelog: https://ninenines.eu/docs/en/ranch/2.1/guide/migrating_from_2.0/

**Should Ranch 2.1 update be for 3.9+ or 3.8+?**

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [x] Dependency update

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
If you're unsure about any of them, don't hesitate to ask on the mailing list.
We're here to help!
This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [x] All tests pass locally with my changes
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

PR #3401, opened 2021-09-10
Use Ranch 2.1.0 in bazel build

pjk25:
Bazel side of #3392

### DIFF
--- a/BUILD.ranch
+++ b/BUILD.ranch
@@ -32,7 +32,7 @@ genrule(
 app_file(
     name = "app_file",
     app_name = "ranch",
-    app_src = ["src/ranch.app.src"],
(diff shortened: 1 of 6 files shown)


lhoguin:
Ah thanks! So there *was* more to do about this update.

Would this have made the PR rebuild Ranch as well? Tests were failing because it was picking up 2.0.

pjk25:
I'm not sure I totally understand the question, but I can say that the Makefiles don't actually affect the checks performed in Github for the Test workflows. And yes, a new branch (and associated PR) can pick up arbitrary dep versions based on `WORKSPACE.bazel` and `workspace_helpers.bzl`.

PR #3421, opened 2021-09-14
Make prometheus plugin output customizable

binarin:
## Proposed Changes

Prometheus plugin always exposes all possible metrics (even in aggregated mode, where the output is smaller, but the same amount of data still needs to be processed), and this can be problematic when there is a large number of connections/channels/queues.

For existing endpoints this change allows to disable some metric families altogether or filter out some vhosts (but this is optional, and without configuration change everything is 100% backwards compatible). This can greatly reduce the amount of data exposed, while still providing enough details for a proper monitoring.

And then there is a new endpoint where the same customizations can be applied on the fly, in case more information is needed, or with different level of details.

## Types of Changes

## Checklist

### DIFF
--- a/deps/rabbitmq_prometheus/test/rabbit_prometheus_http_SUITE.erl
+++ b/deps/rabbitmq_prometheus/test/rabbit_prometheus_http_SUITE.erl
@@ -22,7 +22,8 @@ all() ->
         {group, aggregated_metrics},
         {group, per_object_metrics},
         {group, per_object_endpoint_metrics},
-        {group, commercial}
+        {group, commercial},
+        {group, detailed_metrics}
     ].
 
 groups() ->
@@ -48,6 +49,14 @@ groups() ->
         ]},
         {commercial, [], [
             build_info_product_test
+        ]},
+        {detailed_metrics, [], [
+                                     detailed_metrics_no_families_enabled_by_default,
+                                     queue_consumer_count_single_vhost_per_object_test,
+                                     queue_consumer_count_all_vhosts_per_object_test,
+                                     queue_coarse_metrics_per_object_test,
+                                     queue_metrics_per_object_test,
+                                     queue_consumer_count_and_queue_metrics_mutually_exclusive_test
         ]}
     ].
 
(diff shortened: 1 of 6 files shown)


deps/rabbitmq_prometheus/test/rabbit_prometheus_http_SUITE.erl
```diff
@@ -84,6 +96,72 @@ init_per_group(per_object_endpoint_metrics, Config0) ->
...
+
+    VHost1Consumer = sleeping_consumer(),
+    #'basic.consume_ok'{consumer_tag = VHost1CTag} =
+        amqp_channel:subscribe(VHost1Ch, #'basic.consume'{queue = <<"vhost-1-queue-with-consumer">>}, VHost1Consumer),
+
+    VHost2Consumer = sleeping_consumer(),
+    #'basic.consume_ok'{consumer_tag = VHost2CTag} =
+        amqp_channel:subscribe(VHost2Ch, #'basic.consume'{queue = <<"vhost-2-queue-with-consumer">>}, VHost2Consumer),
+
+    timer:sleep(10000),
```
pjk25, line 151: What underlying condition is this sleep meant to account for?

binarin: I did it by analogy with the other test cases in this file. It is about `collect_statistics_interval`, and this sleep makes sure that all stats will be properly accounted in ETS tables (where prometheus plugin reads them from).

vikinghawk:
This looks really promising!

Our usecase is probably similar to many others... we are only interested in queue message & consumer count at the per-object granularity. But scraping the current all or nothing per-object endpoint with 10k+ queues can take a long time (60+ seconds) and causes high memory usage on the node.

Couple questions:
1) Have you guys decided what releases this will be targeted for? Specifically, is it possible to be backported to 3.8.x?
2) Would it be feasible to add "rabbitmq_queue_consumer_utilisation" to the "queue_consumer_count" group? It would be useful to see the current consumer utilization metrics alongside the count

binarin:
>     1. Have you guys decided what releases this will be targeted for? Specifically, is it possible to be backported to 3.8.x?

For now we are planning to do it, unless there will be some unexpected problem.

 
>     2. Would it be feasible to add "rabbitmq_queue_consumer_utilisation" to the "queue_consumer_count" group? It would be useful to see the current consumer utilization metrics alongside the count

No, but I've did a bit of benchmarking on a 10k queues/10k publishers/10k consumers rabbit:
- Current per-object takes around 2 minutes.
- `queue_coarse_metrics` and `queue_consumer_count` - 1.3 second
- `queue_metrics` - 8.5 seconds

If you need utilization, you'll be able to scrape it now in a more reasonable fashion anyway.

deps/rabbitmq_prometheus/test/rabbit_prometheus_http_SUITE.erl
pjk25: Various other tests reduce that interval when setting up the test (e.g. https://github.com/rabbitmq/rabbitmq-server/blob/3f1b228dccd698e6f7bc07420387064cf2853597/deps/rabbitmq_stream_management/test/http_SUITE.erl#L39-L45). I wonder if it is worth doing the same here so that we don't have to sleep 10s per test.

binarin: I've reduced this and sleep time to 1 second, works on my machine.

luos:
Hi @binarin, seems like this did not get backported by Mergify to v3.8.x. Do you think there is a chance this could be  merged into it? Thanks

sherifkayad:
@binarin similar to @luos I am missing this commit in 3.8.x .. any plans to include it there?

PR #3448, opened 2021-09-17
Quorum Queue consumer cancellation fixes

### LINKED ISSUE #3445: Subscribing/unsubscribing from an empty Quorum Queue causes unbounded log growth.
Very similarly to #2651 

It could also cause RAM growth.

kjnilsson:
Fixes #3445 

If the queue is empty when a consumer is cancelled it would leave the
consumer id inside the service queue. If an application subscribes/unsubscibes
in a loop from an empty queue this would cause the service queue to never be
cleared up.

NB: whenever we make a change to how the quorum queue state machien is
calculated we need to consider how this effects determinism as during an
upgrade different members may calculate a different service queue state.
In this case it should be ok as they will eventually converge on the same
state once all "dead" consumer ids have been removed from the queue.

QQ: emit release cursors after consumer cancel, `purge_nodes`, `garbage_collection` and
`update_config` commands to ensure that repeated use of these commands
against an empty queue will not grow the log excessively.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -340,10 +340,13 @@ apply(#{index := Index,
                     {State, Reply, Effects}
             end
     end;
-apply(Meta, #checkout{spec = cancel, consumer_id = ConsumerId}, State0) ->
-    {State, Effects} = cancel_consumer(Meta, ConsumerId, State0, [],
-                                       consumer_cancel),
-    checkout(Meta, State0, State, Effects);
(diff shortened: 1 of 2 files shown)


kjnilsson:
Acceptance steps:
1. configure a system with a very small WAL size (optional but speeds stuff up)
2. subscribe / unsubscribe from an empty QQ in a loop and check that segments in the mnesia/quorum/NODE/queue directory are deleted periodically

acogoluegnes (approved):

acogoluegnes:

PR #3506, opened 2021-09-24
Remove redundant code in internal auth backend and queue consumers modules

Ayanda-D:
## Types of Changes

- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_auth_backend_internal.erl
+++ b/deps/rabbit/src/rabbit_auth_backend_internal.erl
@@ -234,12 +234,9 @@ add_user_sans_validation(Username, Password, ActingUser) ->
         throw:{error, {user_already_exists, _}} = Error ->
             rabbit_log:warning("Failed to add user '~s': the user already exists", [Username]),
             throw(Error);
-        throw:Error ->
+        Class:Error ->
             rabbit_log:warning("Failed to add user '~s': ~p", [Username, Error]),
-            throw(Error);
-        exit:Error ->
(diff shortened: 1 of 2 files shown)


michaelklishin (approved):

deps/rabbit/src/rabbit_auth_backend_internal.erl
```diff
@@ -234,12 +234,9 @@ add_user_sans_validation(Username, Password, ActingUser) ->
...
             rabbit_log:warning("Failed to add user '~s': the user already exists", [Username]),
             throw(Error);
-        throw:Error ->
+        Class:Error ->
             rabbit_log:warning("Failed to add user '~s': ~p", [Username, Error]),
-            throw(Error);
-        exit:Error ->
-            rabbit_log:warning("Failed to add user '~s': ~p", [Username, Error]),
-            exit(Error)
+            erlang:Class(Error)
```
lhoguin, line 239: We might want to consider using `erlang:raise/3` to have the correct stacktrace for the original exception.

Ayanda-D: True. I'll submit an update (later) to include stacktrace in final exception.

It is something which I find still being used quite sparingly since its introduction in otp-21. Do we make it a convention to always, or more often than not, use/include it in future? (questions from a contributor 😅 )

lhoguin: @Ayanda-D I've not received any objections so please go ahead with such a patch.

PR #3516, opened 2021-09-28
Move CT helpers to monorepo

binarin:
## Proposed Changes

These repos are very rabbit-specific, and bazel files were already pointing to an exact commit sha in there. So there is no reason to keep them separately.

## Types of Changes

- [X] Build system and/or CI

## Checklist


binarin:
@pjk25

Both these repos contained github actions that were just testing that the thing builds. Should some similar tests be peformed on monorepo, or the assumption should be that helpers only change hand-in-hand with real code?

And how can I forcefully kick-off running all of the CT suites? Just to be sure that this renaming didn't affect any tests.

pjk25:
> @pjk25
> 
> Both these repos contained github actions that were just testing that the thing builds. Should some similar tests be peformed on monorepo, or the assumption should be that helpers only change hand-in-hand with real code?

The helpers will be built as part of `rabbitmq-server`s workflows, so the ones from the helpers can be deleted. In fact the whole ` deps/rabbitmq_ct_client_helpers/.github` directory can be deleted, if there is anything else in there.

> And how can I forcefully kick-off running all of the CT suites? Just to be sure that this renaming didn't affect any tests.

GitHub Actions doesn't allow running arbitrary workflows on arbitrary branches. The triggering is defined by the workflow itself. I would have expected the Test workflow to be run on this branch however. It's not clear to me why it didn't. I will have to investigate further.

And additionally, I have realized a few other changes that will be needed:

1. We should actually have 1 of these PRs for each the `master`, `v3.9.x` & `v3.8.x` branches, which merges in the helpers for the corresponding branch. This should ensure that switching our Concourse pipelines to use this copy of the helpers will still work correctly.
2. In bazel, every reference to `@rabbitmq_ct_helpers//:bazel_erlang_lib` will need to change to `//deps/rabbitmq_ct_helpers:bazel_erlang_lib`.
3. The workflow at `.github/workflows/update-rabbitmq-ct-helpers.yaml` should be deleted.
4. There is probably an update to `rabbitmq-components.mk` that should adjust `dep_rabbitmq_ct_client_helpers` appropriately.

binarin:
I've temporarily pushed a commit that just added an empty line to every test suite - looks like everything is working.

Here are all the changes to bazel- and make files, as a separate commit - https://github.com/rabbitmq/rabbitmq-server/pull/3516/commits/46df4f1689a6b1cf516b73d6b013b49ca2af1895

I checked and all three branches are currently pointing to exactly the same commits of helper repos, so backporting them should be possible.

pjk25:
> I checked and all three branches are currently pointing to exactly the same commits of helper repos, so backporting them should be possible.

That is true yes, but only for GitHub actions. In Concourse, there are some things that are still tested with Make, and the helpers are fetched dynamically via Erlang.mk, where the branch matching is honored. We could try the backport, but it won't actually show up as a failure in https://ci.rabbitmq.com/teams/main/pipelines/server-release:v3.8.x?group=qa%3Apkg or https://ci.rabbitmq.com/teams/main/pipelines/server-release:v3.9.x?group=qa%3Apkg until it is merged.

pjk25:
And as far as the github "checks", it's still a mystery to me why it needed files changed, but it seems to be intentional on Github's part, as once you removed that commit, the checks disappeared again. The AWS integration suite seemed to flake, but I'm happy to merge this to master as is.

binarin:
So I need to backport, and then update CI jobs to remove all helper repos references?

pjk25:
The CI jobs retrieve it via the definition in `rabbitmq-components.mk`. `git_rmq` &`git_rmq-subfolder` Erlang.mk macros check the current branch, and clone the matching branch of the helpers. So, what I'm saying is that we can backport, but we won't know if it worked until fully merged. If it turns the Concourse pipeline red, we will have to revert the merge commit, then do separate PRs as I mentioned.

pjk25:
This is one of the reasons we have been doing more things with GitHub Actions - it hasn't been reasonable to test PRs with Concourse in the past, and probably still isn't just yet.

PR #3517, opened 2021-09-29
Reuse list of nodes in peer discovery plugins that use Erlang global locks

tvhong-amazon:
## Proposed Changes

AWS, Kubernetes and Classic peer discovery plugins use list_nodes and
Erlang global:set_lock to create a mutex lock. To unlock, these plugins
get the latest list with list_nodes and call global:del_lock.

However, if list_nodes within unlock fails, RabbitMQ will throw an
uncaught exception and the lock will not be released until the node
holding the lock is restarted. This prevents new nodes from joining the
cluster.

This failure can be avoided by passing the list of nodes from lock to
unlock. If a node goes away (and comes back) between the lock and unlock
calls, del_lock could still successfully remove the lock. Similarly, if
a new node starts up between the lock and unlock calls, del_lock
wouldn't need to inform the new node.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbitmq_peer_discovery_k8s/src/rabbit_peer_discovery_k8s.erl
+++ b/deps/rabbitmq_peer_discovery_k8s/src/rabbit_peer_discovery_k8s.erl
@@ -68,7 +68,8 @@ register() ->
 unregister() ->
     ok.
 
--spec lock(Node :: node()) -> {ok, {ResourceId :: string(), LockRequesterId :: node()}} | {error, Reason :: string()}.
+-spec lock(Node :: node()) -> {ok, {{ResourceId :: string(), LockRequesterId :: node()}, Nodes :: [node()]}} |
+                              {error, Reason :: string()}.
 
 lock(Node) ->
   %% call list_nodes/0 externally such that meck can mock the function
@@ -81,7 +82,7 @@ lock(Node) ->
           Retries = rabbit_nodes:lock_retries(),
           case global:set_lock(LockId, Nodes, Retries) of
             true ->
-              {ok, LockId};
+              {ok, {LockId, Nodes}};
             false ->
               {error, io_lib:format("Acquiring lock taking too long, bailing out after ~b retries", [Retries])}
           end;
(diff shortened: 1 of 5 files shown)


michaelklishin (approved):

michaelklishin:
Thank you.

`v3.8.x` will require a manual merge.

deps/rabbitmq_peer_discovery_k8s/src/rabbit_peer_discovery_k8s.erl
```diff
@@ -95,16 +96,12 @@ lock(Node) ->
       Error
   end.
 
--spec unlock({ResourceId :: string(), LockRequesterId :: node()}) -> ok | {error, Reason :: string()}.
+-spec unlock({{ResourceId :: string(), LockRequesterId :: node()}, Nodes :: [node()]}) ->
+    ok | {error, Reason :: string()}.
```
ansd, line 100: Can we now omit the `{error, Reason :: string()}` part in the function spec?

Running `make dialyze` within folder `deps/rabbitmq_peer_discovery_k8s` will return:
```
rabbitmq_peer_discovery_k8s.erl:51:2: Invalid type specification for function rabbitmq_peer_discovery_k8s:unlock/1. The success typing is
          ({{string(), atom()}, [atom()]}) -> 'ok'
```
So, we'd also need to modify the function spec in that other file.

pjk25: @ansd this has already been merged, so if the spec were to be adjusted, it will have to be a new PR.

tvhong-amazon: ah right. I forgot about that. Also, I thought the default make target would also run the spec analyzer, but that isn't the case.

tvhong-amazon:
Just for my own reference, this PR was also merged into 3.8.x branch: https://github.com/rabbitmq/rabbitmq-server/commit/74dae11c8fcd29cd7a728ca00cc9b745adc72927 .

Probably released in 3.8.24 on 2021/11/11.

PR #3552, opened 2021-10-08
Improve consuming from a stream using AMQP 1.0

kjnilsson:
Allow an offset spec to be used to attach at an appropriate point in the
stream. This is done by specifying a source filter with the key `rabbitmq:stream-offset-spec`.

The offset is also included as a message annotation with the key `x-stream-offset`.

When a link is detached we also issue a basic.cancel to the 0.9.1 channel. If this wasn't done
and you detached then re-attached a link for the same queue you'd get a consumer-tag offset
error from the 0.9.1 channel.


lukebakken (approved):
@kjnilsson gave a nice tour of this feature :+1:

PR #3587, opened 2021-10-18
Pre-render prometheus labels

binarin:
## Proposed Changes

Pre-render labels for prometheus exported. Character escaping is not needed, as AMQP object names are quite restricted in what characters they can have.

Doing it this way removes not only CPU overhead, but also a lot of garbage collection overhead (as for every output line labels were converted from a proplist to a list of records before rendering).

This makes per-object metrics twice as fast, time to produce ~630k metrics (10k queues/10k consumer/10k producers) went from ~6 to ~3 seconds from this change.

Depends on https://github.com/deadtrickster/prometheus.erl/pull/137

This also pulls an performance optimization https://github.com/deadtrickster/prometheus.erl/pull/136 - across all those changes on my test machine I saw scraping time dropping 40 times for the aforementioned 630k metrics.

Testing was performed by running 2 instances of perftest with the following arguments (just with a slightly modified queue pattern each time):
```
--queue-pattern 'perf-test-%d' \
  --queue-pattern-from 1 --queue-pattern-to 5000 \
  --producers 1 --consumers 1 --rate 1
```
2 instances because perftest was running out of memory, and running two instances was easier for me than learning how to tune java memory settings.

The timing of following urls are of special interest for testing:
```
http://localhost:15692/metrics/per-object
http://localhost:15692/metrics/detailed?family=queue_coarse_metrics&family=queue_consumer_count
http://localhost:15692/metrics/detailed?family=queue_metrics
```

## Types of Changes

- [ ] Build system and/or CI

## Checklist


pjk25:
@binarin `git_repository` refers to an external repository that has it's own bazel files. `github_bazel_erlang_lib` is a helper we have to import a non-bazel repo into bazel-land.

pjk25:
@binarin I think this is looking good now

michaelklishin (approved):
I can confirm 20-50% reduction in response latency (the number of samples is small but the results are consistent).

michaelklishin:
Backported to `v3.9.x` manually for `3.9.9`.

Issue #3386, opened 2021-09-08
Consistent hash exchange bindings are changed and break when a node restarts


luos:
Hi,

When a single node is restarted out of a cluster the consistent hash exchange will add existing bindings again into the ring. 

Reproduction: 

1. Create 3 node cluster
2. Create a consistent hash exchange
3. Bind a queue to it with any weight (for example 1)
4. Stop a node (with stop_app or anything) 

Output: 

```
root@rmq1:/# rabbitmqctl eval 'ets:tab2list(rabbit_exchange_type_consistent_hash_ring_state).'
[{chx_hash_ring,{resource,<<"/">>,exchange,<<"test-exchange">>},
                #{13 => {resource,<<"/">>,queue,<<"test-qq">>}},
                14}]
root@rmq1:/# rabbitmqctl eval 'ets:tab2list(rabbit_exchange_type_consistent_hash_ring_state).'^C
root@rmq1:/# rabbitmqctl start_app^C
root@rmq1:/# rabbitmqctl stop_app && rabbitmqctl start_app
Stopping rabbit application on node rabbit@rmq1.monitored.host ...
Starting node rabbit@rmq1.monitored.host ...
root@rmq1:/# rabbitmqctl eval 'ets:tab2list(rabbit_exchange_type_consistent_hash_ring_state).'
[{chx_hash_ring,{resource,<<"/">>,exchange,<<"test-exchange">>},
                #{13 => {resource,<<"/">>,queue,<<"test-qq">>},
                  14 => {resource,<<"/">>,queue,<<"test-qq">>},
                  15 => {resource,<<"/">>,queue,<<"test-qq">>}},
                16}]
root@rmq1:/# rabbitmqctl stop_app && rabbitmqctl start_app
Stopping rabbit application on node rabbit@rmq1.monitored.host ...
Starting node rabbit@rmq1.monitored.host ...
root@rmq1:/# rabbitmqctl eval 'ets:tab2list(rabbit_exchange_type_consistent_hash_ring_state).'
[{chx_hash_ring,{resource,<<"/">>,exchange,<<"test-exchange">>},
                #{13 => {resource,<<"/">>,queue,<<"test-qq">>},
                  14 => {resource,<<"/">>,queue,<<"test-qq">>},
                  15 => {resource,<<"/">>,queue,<<"test-qq">>},
                  16 => {resource,<<"/">>,queue,<<"test-qq">>},
[...]

PR #3594, opened 2021-10-19
Correct consistent hashing exchange binding deletion when there are duplicate bindings

FalconerTC:
## Proposed Changes
The `x-consistent-hash` plugin has unexpected behavior when multiple bindings are established to the same queue (see https://github.com/rabbitmq/rabbitmq-server/issues/3386). This causes the hash ring to enter an invalid state, where messages can not be consistently routed anywhere. When entering this state, the only resolution is to delete and recreate the exchange. To replicate:

1. Create an `x-consistent-hash` exchange
2. Create a binding to a queue Q1
3. Create a second binding of different weight to queue Q1
4. Delete either binding, or delete Q1

This PR hopes to resolve the `Bucket 0 not found` errors that occur when the hash ring enters an invalid state. When removing a binding, it adjusts queues based on the buckets being removed rather than the weight specified, which differ only with duplicate bindings.

This is possibly also related to https://github.com/rabbitmq/rabbitmq-server/discussions/3001

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

I'm not familiar with the project, nor with erlang, so open to any an all suggestions. 

This does not fix the greater issue brought up in https://github.com/rabbitmq/rabbitmq-server/issues/3386, where nodes coming up and down result in multiple entries in the hash ring. Longer term, it may be better to disallow adding multiple bindings to a single queue. Given this issue with binding deletion, that feature has probably not been in use much.

### DIFF
--- a/deps/rabbitmq_consistent_hash_exchange/src/rabbit_exchange_type_consistent_hash.erl
+++ b/deps/rabbitmq_consistent_hash_exchange/src/rabbit_exchange_type_consistent_hash.erl
@@ -227,7 +227,7 @@ remove_bindings(none, X, Bindings) ->
     ok.
 
 remove_binding(#binding{source = S, destination = D, key = RK}) ->
-    Weight = rabbit_data_coercion:to_integer(RK),
+    _Weight = rabbit_data_coercion:to_integer(RK),
     rabbit_log:debug("Consistent hashing exchange: removing binding "
                      "from exchange '~p' to destination '~p' with routing key '~s'",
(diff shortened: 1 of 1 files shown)


michaelklishin:
This exchange was never meant to be used with duplicate bindings, this is why it has routing key weights.

Handing duplicates is nice but there is a reason why this has gone largely unreported for years: this is not how this plugin is supposed to be used.

michaelklishin:
Thank you.

michaelklishin:
Backported to `v3.9.x` manually.

PR #3628, opened 2021-11-02
QQ: stability and channel side improvements

kjnilsson:
This should avoid channel crashes when the leader has recently changed
and the amqqueue record hasn't yet been updated.

Also catch errors in the QQ tick temporary process that updates metrics.
During shutdown the metrics ETS table may disappear before this process has finished
which would result in a noisy error in the logs.

Also handle case where a QQ "pid" was passed to a function expecting an actual pid.


mkuratczyk (approved):
I can no longer reproduce the failures we wanted to address.

PR #3631, opened 2021-11-02
Use erlang:system_info(creation) as GUID

mkuratczyk:
The goal of these changes is to prevent a false-positive network partition:
in some cases, in a multi-node cluster, a single node restart could trigger
node monitor to declare a partition and stop the remaining nodes.

Node GUID allows to differentiate between different incarnations of a node.
However, since rabbit may take some time to start (many queues/bindings, etc),
there could be a significant difference between Erlang VM being up and
responding to RPC calls and the new GUID being announced. During that
time, node monitor could incorrectly assume there was a network
partition, while in fact a node was simply restarted. With this change,
as soon as the Erlang VM is up, we can tell whether it was restarted and
avoid false positives.

Additionally, we now log if any queues were deleted on behalf of the
restarted node. This can take quite a long time if there are many transient
queues (eg. auto-delete queues). The longer this takes, the higher were the
odds of a restarted node being up again by the time
check_partial_partition was called. We may need to reconsider this logic
as well but for now - we just log this activity.

Co-authored-by: Loïc Hoguin <lhoguin@vmware.com>

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -1905,7 +1905,11 @@ maybe_clear_recoverable_node(Node, Q) ->
 -spec on_node_down(node()) -> 'ok'.
 
 on_node_down(Node) ->
-    {QueueNames, QueueDeletions} = delete_queues_on_node_down(Node),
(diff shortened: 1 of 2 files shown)


michaelklishin:
@mergifyio rebase

michaelklishin:
Backported to `v3.8.x` for `3.8.24` manually.

lukebakken (commented):
:+1:

PR #3635, opened 2021-11-02
Introduce a target cluster size hint setting

michaelklishin:
This is meant to be used by deployment tools,
core features and plugins that expect a certain minimum
number of cluster nodes to be present.

For example, certain setup steps
in distributed plugins might require
at least three nodes to be available.

This is just a hint, not an enforced
requirement. The default value is 1
so that for single node clusters,
there would be no behavior changes.


dcorbacho (approved):

pjk25:
@mergifyio rebase

pjk25:
Rebasing to pick up CI fixes

PR #3643, opened 2021-11-03
Turn down logging from queue rebalancing

mkuratczyk:
With many queues, rebalancing can log hundreds of lines at warning/info
level, even though nothing exciting happened. I think we can tune that
down - if there is nothing to do - that's a debug level,
if things go well - that's info, not a warning.

I labeled it for backporting to 3.8 and 3.9 but feel free to change that.


michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x` manually for `3.8.24`.

PR #3676, opened 2021-11-09
Avoid slow mnesia transaction on QQ init

kjnilsson:
As we only need to make sure the rabbit_queues table is populated
use a dirty write function that only does this instead. This could potentially
half recovery times for many QQ scenarios.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -7,6 +7,7 @@
 
 -module(rabbit_amqqueue).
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
One interesting side effect of this: the faster the (parallel) QQ recovery process gets, the higher will be peak file handle use on the node.

michaelklishin:
WIth this change, 7K quorum queues recover in 70s vs. over 240s with `3.9.8`.

michaelklishin:
We still need to pin `v3.9.x` to a new Ra release in order for this to be completely backported.

PR #3659, opened 2021-11-05
Adopt latest rabbitmq-server-buildenv:linux-erlang-24.1 for RBE

pjk25:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #3683, opened 2021-11-10
Use externailzed rbe config in bazel

pjk25:
This moves from the embedded `erlang_23_platform` and `erlang_24_platform` used for Bazel remote execution to more complete platform definitions generated with https://github.com/bazelbuild/bazel-toolchains. These configs are based 

This serves 2 purposes:
1. Provide a properly configured C compiler (clang) so that nifs can be compiled (if present)
2. Relieve some of our other projects from an artificial dependency on this repo

Note: The `erlang_git_platform` platform definition remains embedded for now, as it's updated just in time for use in the GitHub Actions workflow that tests against the Erlang master branch. For now externalizing that definition is more trouble that it's worth.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -14,38 +14,25 @@ build:rbe --config=buildbuddy
 
 build:rbe --remote_executor=grpcs://remote.buildbuddy.io
 
-build:rbe --crosstool_top=@buildbuddy_toolchain//:toolchain
-build:rbe --extra_toolchains=@buildbuddy_toolchain//:cc_toolchain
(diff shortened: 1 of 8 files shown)


pjk25:
So this hits the same SSL issue as #3659, as it inadvertently picks up a newer erlang 24 than is currently used in the passing build. I will leave it as a draft until the fix in otp is available.

pjk25:
The issue on the otp side is https://github.com/erlang/otp/issues/5255

pjk25:
@mergifyio rebase

pjk25:
@mergifyio rebase

PR #3700, opened 2021-11-12
Add --remote_download_toplevel to the bazel config

pjk25:
Based on the testing and advice from BuildBuddy:

Network limitations/flakiness in GitHub actions can be mitigated by
reducing the number of artifacts that get downloaded. Since we only
use bazel for testing currently, we don't actually care about the
artifacts, so there is no downside to the flag.

If this isn't enough, we can try the --remote_download_minimal flag,
which limits the amount of downloads even further.


pjk25:
Thanks @siggisim for testing this

PR #3740, opened 2021-11-16
Distribution listener settings support in rabbitmq.conf

### LINKED ISSUE #3739: Consider a setting that controls inter-node communication interface
See https://github.com/rabbitmq/rabbitmq-website/issues/1304 for the background.

There are [settings that control what port range](https://rabbitmq.com/networking.html#epmd-inet-dist-port-range) will be used
by inter-node connections but the only way to limit the interface is to use `advanced.config` and `inet:parse/1` from an Erlang shell.

Note that we already support interface names/IP addresses in `rabbitmq.conf` (thanks to Cuttlefish).

michaelklishin:
* `distribution.listener.interface`
 * `distribution.listener.port_range.min`
 * `distribution.listener.port_range.max`

A `rabbitmq.conf` that can be used for manual testing:

``` ini
distribution.listener.port_range.min = 25679
distribution.listener.port_range.max = 25679
distribution.listener.interface = 127.0.0.1
```

Then:

```
rabbitmq-diagnostics listeners
```

```
rabbitmqctl eval 'application:get_all_env(kernel).'
```

Closes #3739.

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -1803,6 +1803,37 @@ end}.
     {validators, ["non_zero_positive_integer"]}
 ]}.
 
+{mapping, "distribution.listener.port_range.min", "kernel.inet_dist_listen_min", [
+    {datatype, [integer]},
+    {validators, ["non_zero_positive_integer"]}
+]}.
+
+{mapping, "distribution.listener.port_range.max", "kernel.inet_dist_listen_max", [
+    {datatype, [integer]},
+    {validators, ["non_zero_positive_integer"]}
+]}.
+
+{mapping, "distribution.listener.interface", "kernel.inet_dist_use_interface", [
+    {datatype, [string]},
+    {validators, ["is_ip"]}
+]}.
+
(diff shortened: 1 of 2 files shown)


polaris-alioth:
It there any conflict between `RABBITMQ_CTL_DIST_PORT_MIN` and `distribution.listener.port_range.min`?

michaelklishin:
@polaris-alioth the former is for CLI tools, the latter is for RabbitMQ nodes.

lukebakken (approved):
Tests pass. Manually started broker with this config:

```
distribution.listener.port_range.min = 25679
distribution.listener.port_range.max = 25679
distribution.listener.interface = 127.0.0.1
```

Verification:

```
$ epmd -names
epmd: up and running on port 4369 with data:
name rabbit at port 25679
```

`ss -4nl` output, truncated:

```
Netid State  Recv-Q Send-Q Local Address:Port  Peer Address:PortProcess
tcp   LISTEN 0      128        127.0.0.1:25679      0.0.0.0:*          
tcp   LISTEN 0      4096         0.0.0.0:4369       0.0.0.0:*          

```

PR #3754, opened 2021-11-17
Single Active Consumer for streams

acogoluegnes:
References #3753


deps/rabbit/BUILD.bazel
```diff
@@ -749,6 +749,12 @@ suites = [
             "//deps/rabbit_common:erlang_app",
         ],
     ),
+    rabbitmq_suite(
+        name = "rabbit_stream_sac_coordinator_SUITE",
+        deps = [
+            "//deps/rabbit_common:bazel_erlang_lib",
```
pjk25, line 755: Should be `"//deps/rabbit_common:erlang_app"`

acogoluegnes:
A/C (WIP):

```shell
docker rmi pivotalrabbitmq/rabbitmq-stream:sac
```

```shell
docker run -it --rm --name rabbitmq -p 5552:5552 -p 5672:5672 -p 15672:15672 \                      
    -e RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS='-rabbitmq_stream advertised_host localhost' \
    pivotalrabbitmq/rabbitmq-stream:sac
```

```shell
cd /tmp
git clone git@github.com:rabbitmq/rabbitmq-stream-java-client.git
cd rabbitmq-stream-java-client
./mvnw clean install -DskipTests
./mvnw clean package -Dmaven.test.skip -P performance-tool
cd /tmp
git clone https://github.com/acogoluegnes/rabbitmq-stream-single-active-consumer.git
cd rabbitmq-stream-single-active-consumer
```

pjk25:
> This pull request modifies the bazel build only. Should the makefiles be updated as well @acogoluegnes?

This is fine. New test suites will touch the Bazel build only.

deps/rabbit/src/rabbit_stream_sac_coordinator.erl
```diff
@@ -0,0 +1,629 @@
...
+                      [], Consumers0),
+    G#group{consumers = CS1}.
+
+mod_call_effect(Pid, Msg) ->
+    {mod_call, rabbit_stream_sac_coordinator, send_message, [Pid, Msg]}.
+
+-spec send_message(pid(), term()) -> ok.
+send_message(ConnectionPid, Msg) ->
+    ConnectionPid ! Msg,
+    ok.
```
ansd, line 626: Why `mod_call` effect instead of directly `send_msg` effect?

deps/rabbit/src/rabbit_stream_sac_coordinator.erl
```diff
@@ -0,0 +1,629 @@
...
+              ConsumerName,
+              Group,
+              StreamGroups) ->
+    maps:put({VirtualHost, Stream, ConsumerName}, Group, StreamGroups).
+
+update_consumer_state_in_group(#group{consumers = Consumers0} = G,
+                               Pid,
+                               SubId,
+                               NewState) ->
+    CS1 = lists:foldr(fun(C0, Acc) ->
```
ansd, line 609: Can we simplify by using `lists:map` instead of `lists:foldr`?

deps/rabbitmq_stream/src/rabbit_stream_manager.erl
```diff
@@ -427,6 +434,60 @@ handle_call({route, RoutingKey, VirtualHost, SuperStream}, _From,
...
+                                      end;
+                                  _ ->
+                                      {ok, -1}
+                              end
+                      end
+              end
+          catch
+              exit:Error ->
+                  rabbit_log:error("Error while looking up exchange ~p, ~p",
+                                   [ExchangeName, Error]),
```
ansd, line 487: Let's use `rabbit_misc:rs/1` to print the exchange?

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -2376,6 +2487,137 @@ handle_frame_post_auth(Transport,
...
+                                          messages_consumed(ConsumerCounters)]),
+
+                        Consumers#{SubscriptionId => Consumer2};
+                    #{SubscriptionId :=
+                          #consumer{configuration =
+                                        #consumer_configuration{active = false,
+                                                                stream = Stream,
+                                                                properties =
+                                                                    Properties}}} ->
+                        rabbit_log:debug("Not an active consumer"),
```
ansd, line 2588: There are quite a lot of debug messages that won't make it into the final version I guess? These were just meant for development, right?

deps/rabbit/src/rabbit_stream_coordinator.hrl
```diff
@@ -59,10 +59,11 @@
 -record(rabbit_stream_coordinator, {streams = #{} :: #{stream_id() => #stream{}},
                                     monitors = #{} :: #{pid() => {stream_id() | %% v0 & v1
                                                                   #{stream_id() => ok}, %% v2
-                                                                  monitor_role()}},
+                                                                  monitor_role()} |
+                                                                 {pid(), sac}},
```
ansd, line 63: In the case of SAC, map key and value have the same PID it seems. Why do we need to store the PID in the value?

ansd (commented):
Thank you, looks very good overall. I did the acceptance. Worked for me.

I left some small inline comments.

More general questions are:

(1)
I'm missing declaring a SAC stream via AMQP / Management API. Will that be added in a future PR or should it be part of this PR or is that not planned at all?

Also when listing the queues in the Management UI, it would be nice to see the `SAC` feature icon (as displayed for other queue types).

(2)
In the example above:

> Stop instance-1 (the first you started) with Ctrl-C. This should trigger a significant rebalancing:
> 
> instance-2: formerly invoices-1, now invoices-0 and invoices-2
> instance-3: formerly invoices-2, now invoices-1 

It would be nice if there would not be a "significant" rebalancing. So, in this case I would have expected `instance-2` to continue with `invoices-1` and `instance-3` to continue with `invoices-2` while instance 2 or 3 gets additionally `invoices-0`. I'm thinking about stream processing applications with more than only 3 instances and only 3 partitions. In this case if a single instance crashes, **all** other instances will need to stop their stream processing (for example aggregate calculations) and start with consuming from new stream partitions even though in above example they could have just continued if there wasn't a "significant" rebalancing happening.

acogoluegnes:
@ansd Thanks for the review. I'll have a look at the code comments. Some answers below.

> (1) I'm missing declaring a SAC stream via AMQP / Management API. Will that be added in a future PR or should it be part of this PR or is that not planned at all?

There's no such thing as a SAC stream. An application sets a property when registering a consumer to declare it's a single active consumer. It must also sets the "name" of the consumer, that is some ID that the several consumer instances should share.

> Also when listing the queues in the Management UI, it would be nice to see the `SAC` feature icon (as displayed for other queue types).

The SAC semantics between queues and streams are different. For a queue, all consumers are the same (they don't have a "name"), so the queue is declared as SAC-compliant and the SAC behavior is enforced for consumers registering to it.

It's different with streams. You declare a regular stream and consumers registers to it, and they can SAC or regular consumers.

> (2) In the example above:
> 
> > Stop instance-1 (the first you started) with Ctrl-C. This should trigger a significant rebalancing:
> > instance-2: formerly invoices-1, now invoices-0 and invoices-2
> > instance-3: formerly invoices-2, now invoices-1
> 
> It would be nice if there would not be a "significant" rebalancing. So, in this case I would have expected `instance-2` to continue with `invoices-1` and `instance-3` to continue with `invoices-2` while instance 2 or 3 gets additionally `invoices-0`. I'm thinking about stream processing applications with more than only 3 instances and only 3 partitions. In this case if a single instance crashes, **all** other instances will need to stop their stream processing (for example aggregate calculations) and start with consuming from new stream partitions even though in above example they could have just continued if there wasn't a "significant" rebalancing happening.

We use this simple and predictable rebalancing algorithm with a modulo on the list of consumers registered to a stream. It's stream-specific, which means there's no coordination between the streams composing the super stream. It gives good results without too much hassle, but it can get a big dumb for rebalancing. We're still to open to find another mechanism that good give better results, especially for the rebalancing.

deps/rabbit/src/rabbit_stream_sac_coordinator.erl
acogoluegnes: `mod_call` provides more guarantee of execution than `send_msg` IIRC. Is that correct @kjnilsson?

deps/rabbit/src/rabbit_stream_coordinator.hrl
acogoluegnes: Yes, this can be simplified.

deps/rabbit/src/rabbit_stream_sac_coordinator.erl
```diff
@@ -0,0 +1,756 @@
...
+                                    StreamGroups0),
+                {State0#?MODULE{groups = SGS}, Effects}
+        end,
+    {State1, ok, Effects1};
+apply({activate_consumer, VirtualHost, Stream, ConsumerName},
+      #?MODULE{groups = StreamGroups0} = State0) ->
+    {G, Eff} =
+        case lookup_group(VirtualHost, Stream, ConsumerName, StreamGroups0) of
+            undefined ->
+                rabbit_log:warning("trying to activate consumer in group ~p, but "
```
kjnilsson, line 267: I think this will return whatever `rabbit_log:warning` returns (probably `ok`) and crash as it cannot match agains `{G, Eff}`

deps/rabbit/src/rabbit_stream_sac_coordinator.hrl
```diff
@@ -0,0 +1,35 @@
...
+-type owner() :: binary().
+
+-record(consumer,
+        {pid :: pid(),
+         subscription_id :: subscription_id(),
+         owner :: owner(), %% just a label
+         active :: boolean()}).
+-record(group,
+        {consumers :: [#consumer{}], partition_index :: integer()}).
+-record(rabbit_stream_sac_coordinator,
```
kjnilsson, line 33: Might be worth putting one or two reserved fields in here for the future

deps/rabbit/src/rabbit_stream_sac_coordinator.erl
```diff
@@ -0,0 +1,756 @@
...
+                                                [], InfoKeys),
+                                [Record | Acc]
+                             end,
+                             [], Consumers),
+            {ok, Cs};
+        _ ->
+            {error, not_found}
+    end.
+
+-spec ensure_monitors(command(), state(), map(), ra:effects()) ->
```
kjnilsson, line 362: ra:effects isn't an exported type. use `ra_machine:effects()`

deps/rabbit/src/rabbit_stream_sac_coordinator.erl
```diff
@@ -0,0 +1,756 @@
...
+                 _PartitionIndex,
+                 ConsumerName,
+                 Pid,
+                 _Owner,
+                 _SubscriptionId},
+                #?MODULE{pids_groups = PidsGroups0} = State0,
+                Monitors0,
+                Effects) ->
+    GroupId = {VirtualHost, Stream, ConsumerName},
+    Groups0 = maps:get(Pid, PidsGroups0, sets:new()),
```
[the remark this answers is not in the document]
ansd: I was about to write the same when reviewing the code. But version 2 and therefore `sets:new/1` is not available in OTP 23. Will the minimum OTP version be 24 for this PR?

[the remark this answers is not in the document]
kjnilsson: not sure, I forgot to edit my comment, discussed with @acogoluegnes about just using a map instead

deps/rabbit/src/rabbit_stream_sac_coordinator.erl
```diff
@@ -0,0 +1,756 @@
...
+%% The Initial Developer of the Original Code is Pivotal Software, Inc.
+%% Copyright (c) 2021-2022 VMware, Inc. or its affiliates.  All rights reserved.
+%%
+
+-module(rabbit_stream_sac_coordinator).
+
+-include("rabbit_stream_sac_coordinator.hrl").
+
+-opaque command() ::
+    {register_consumer,
```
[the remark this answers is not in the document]
acogoluegnes: The commands are tuples as well in the stream coordinator, that's why I did the same in the SAC coordinator, but I can change them for records, there's just a couple of them.

MarcialRosales:
I did the acceptance and it worked for me too. I only have the following  comments below.

Testing SAC =>
1) when I launch the single active consumer app, I noticed these 7 connections: 1 locator, 3 producer(s), and 3 consumer(s). I did not expect the 3 producer(s) connections though. 

2) When I tested SAC with two separate applications instances on the same consumer group, I got the following warning message on the first instance. However, it did not affect the consumer and it continued consuming fine afterwards.
```
12:33:23.622 [pool-10-thread-1] WARN  c.r.stream.impl.StreamConsumer - Previous and new status are the same (true), there should be no consumer update in this case.
```

Testing Super Streams =>
1) The "consumers" column on the queues tab did not show the number of consumers in the partitioned streams. It did show `?`. I am not sure what that value should be, whether the total count of consumers (active and waiting) or 1 if it has an active consumer and 0 if it has no consumers at all.

 2) Have you considered having `rabbitmqctrl` commands for super-streams such as this one that given the name of the super-stream and the consumer group, it gives me who is consuming from the super-stream without having to run this command once per partition.
```
rabbitmqctl list_stream_group_consumers --super-stream invoices --reference my-app
```

3) The documentation looks very good. The `RoutingStrategy` in the super stream is very powerful. Maybe it is worth mentioning that it allows us to route the message to multiple partitions or none rather than the standard routing function that routes the message always to one partition. Although from the code sample you have provided it is clear that we are returning a list rather than a single value.

acogoluegnes:
My answers below.

 
> Testing SAC =>
> 
> 1. when I launched the single active consumer app, I noticed these 7 connections: 1 locator, 3 producer(s), and 3 consumer(s). I did not expect the 3 producer(s) connections though.

This is expected, the producer connections handle the offset storage. There's 1 connection per consumer and tracking consumer to simulate independent application (in different OS processes). This is set up in the program:

```java
try (Environment environment =
        Environment.builder()
            .maxConsumersByConnection(1)
            .maxTrackingConsumersByConnection(1)
            .build()) {
...
}
```

> 2. When I tested SAC with two separate applications instances on the same consumer group, I got the following warning message on the first instance. However, it did not affect the consumer and it continued consuming fine afterwards.
> 
> ```
> 12:33:23.622 [pool-10-thread-1] WARN  c.r.stream.impl.StreamConsumer - Previous and new status are the same (true), there should be no consumer update in this case.
> ```
> 

This should not happen, can you tell me how to reproduce?

> Testing Super Streams =>
> 
> 1. The "consumers" column on the queues tab did not show the number of consumers in the partitioned streams. It did show `?`. I am not sure what that value should be, whether the total count of consumers (active and waiting) or 1 if it has an active consumer and 0 if it has no consumers at all.

This is a bug that impacts streams in general, so nothing specific to SAC or super streams.

[Follow-up issue](https://github.com/rabbitmq/rabbitmq-server/issues/4622).

> 2. Have you considered having `rabbitmqctl` commands for super-streams such as this one that given the name of the super-stream and the consumer group, it gives me who is consuming from the super-stream without having to run this command once per partition?
> 
> ```
> rabbitmqctl list_stream_group_consumers --super-stream invoices --reference my-app
> ```

This is a good point, there's no "global" way to monitor a super stream, some CLI commands would be a good start. This may be out of the scope of this PR for now.

> 
> 3. The documentation looks very good. The `RoutingStrategy` in the super stream is very powerful. Maybe it is worth mentioning that it allows us to route the message to multiple partitions or none rather than the standard routing function that routes the message always to one partition. Although from the code sample you have provided it is clear that we are returning a list rather than a single value.

Sounds good, will do.

MarcialRosales:
@acogoluegnes With regards point 1) on Testing SAC, do you think it is too low level detail to mention it, even in the form of a "foot note" or similar, that a consumer stream should expect to see at least one producer connection required to handle the offset storage? Or maybe this low level detail should be on a blog post like you did when you explained [how to connect to streams](https://blog.rabbitmq.com/posts/2021/07/connecting-to-streams/). I do not know where to put but I think soon or later someone will be wondering why they have those producer connections. 

With regards point 2) on Testing SAC, what i did was:
1) follow your acceptance testing steps 
2) and once I saw that SAC was working. I launched a second instance. And when I did it, the first instance immediately showed that warning message. I killed the 2nd instance and started again, and once again the first instance showed that warning message. I will try again and If I reproduce it I ping you on slack.

acogoluegnes:
> @acogoluegnes With regards point 1) on Testing SAC, do you think it is too low level detail to mention it, even in the form of a "foot note" or similar, that a consumer stream should expect to see at least one producer connection required to handle the offset storage? Or maybe this low level detail should be on a blog post like when you explained [how to connect to streams](https://blog.rabbitmq.com/posts/2021/07/connecting-to-streams/). I do not know where to put it but I think soon or later someone will be wondering why they have those producer connections.

The documentation can contain a note about this.

PR #3759, opened 2021-11-18
Fix rabbitmq.conf translation rule for OAuth2 signing keys

mkuratczyk:
The structure of the signing_keys map should be `<<"id">> => {pem, <<"key">>}`. Previously it was mapped directly as `<<"id">> => <<"key">>` which caused a `case_clause` error.


michaelklishin (approved):

michaelklishin:
Backported to `v3.9.x` for `3.9.10`

PR #3763, opened 2021-11-18
Add the bazel target :package-generic-unix

pjk25:
Which can be used to build the artifact with bazel

With the current configuration, it does not include the _example "plugins", as I was not convinced they are intended to be part of the archive.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,9 +1,12 @@
+load("@rules_pkg//:pkg.bzl", "pkg_tar")
 load("@bazel-erlang//:dialyze.bzl", "plt")
 load("@bazel-erlang//:shell.bzl", "shell")
 load("elixir_home.bzl", "elixir_home")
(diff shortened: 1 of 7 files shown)


pjk25:
Backports will likely fail until their respective branches adopt a version of `bazel-erlang` that includes https://github.com/rabbitmq/bazel-erlang/commit/5bf79d7fb9a2c8bf7bfb6a90f5104a10b5c9107b

BUILD.bazel
```diff
@@ -132,3 +99,70 @@ shell(
...
+    ),
+    deps = ALL_PLUGINS,
+)
+
+versioned_rabbitmq_home(
+    name = "dist-home",
+    plugins = ALL_PLUGINS,
+)
+
+# need to add the INSTALL file
```
pjk25, line 120: Actually, the file has been include in the lines below

pjk25: da53d4018a50fe8f887336b689b679271477c297

PR #3798, opened 2021-11-24
Link to most recent description of Osiris chunk format

wrobell:
The most recent description of Osiris chunk format does not reference
the timestamp field to be "posix-ish" anymore. This was bit misleading
as it is Erlang's system time.

Add link to Erlang system time documentation to the subscription command
description to avoid confusion about the timestamp field.


wrobell:
The pull request is because of the discussion: https://github.com/rabbitmq/rabbitmq-server/discussions/3775

PR #3800, opened 2021-11-25
To stop collect_statistics when rabbitmq_management_agent has been disabled

tomyouyou:
After stopping the 'rabbitmq_management', these statistics tables are still being refreshed.

### DIFF
--- a/deps/rabbitmq_management_agent/src/rabbit_mgmt_db_handler.erl
+++ b/deps/rabbitmq_management_agent/src/rabbit_mgmt_db_handler.erl
@@ -93,7 +93,15 @@ handle_info(_Info, State) ->
     {ok, State}.
 
 terminate(_Arg, _State) ->
+    ensure_statistics_disabled(),
     ok.
 
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
+
+ensure_statistics_disabled() ->
+    %% Reset the default values, see Makefile
+    _ = rabbit_log:info("Management plugin: to stop collect_statistics."),
+    application:set_env(rabbit, collect_statistics, none),
+    application:set_env(rabbit, collect_statistics_interval, 5000),
+    ok = rabbit:force_event_refresh(erlang:make_ref()).


michaelklishin:
My testing suggests that when the plugin is re-enabled, the settings are restored, and when it is disabled outright, so will be the Prometheus plugin as `rabbitmq_management_agent` is one of its dependencies.

We can probably safely accept this but I'm curious what problem would it solve @tomyouyou?

michaelklishin (approved):

tomyouyou:
> My testing suggests that when the plugin is re-enabled, the settings are restored, and when it is disabled outright, so will be the Prometheus plugin as `rabbitmq_management_agent` is one of its dependencies.
> 
> We can probably safely accept this but I'm curious what problem would it solve @tomyouyou?

It will lead to performance degradation. Take my test as an example.

The command used in the test is:
bin/runjava com.rabbitmq.perf.PerfTest -h amqp://guest:guest@10.228.103.137:5672  --queue q1 --flag mandatory --qos 300 --confirm 100 --producers 1 --consumers 1


Test results under normal conditions:

id: test-085929-497, time: 400.012s, sent: 18214 msg/s, returned: 0 msg/s, confirmed: 18215 msg/s, nacked: 0 msg/s, received: 18206 msg/s, min/median/75th/95th/99th consumer latency: 2062/5673/6012/6772/7520 µs, confirm latency: 2343/5510/5814/6494/7154 µs
test stopped (Reached time limit)
id: test-085929-497, sending rate avg: 17480 msg/s
id: test-085929-497, receiving rate avg: 17480 msg/s

After the plugin has been enabled and disabled, the test results are:

id: test-090730-902, time: 400.025s, sent: 16261 msg/s, returned: 0 msg/s, confirmed: 16260 msg/s, nacked: 0 msg/s, received: 16264 msg/s, min/median/75th/95th/99th consumer latency: 2957/6407/6875/7645/8454 µs, confirm latency: 2164/6162/6596/7352/8068 µs
test stopped (Reached time limit)
id: test-090730-902, sending rate avg: 16095 msg/s
id: test-090730-902, receiving rate avg: 16095 msg/s

we’ve seen a 7.9% decrease in the number of messages per second.

michaelklishin:
I'm not sure what those "normal" conditions are. If stats are disabled by this PR, why would there be any throughput decrease? Nodes supposedly do (potentially quite a bit) less work in that case.

So what are we actually comparing in these runs?

tomyouyou:
> I'm not sure what those "normal" conditions are. If stats are disabled by this PR, why would there be any throughput decrease? Nodes supposedly do (potentially quite a bit) less work in that case.
> 
> So what are we actually comparing in these runs?

1. The normal condition mentioned above refers to the condition that no plug-in has been enabled or disabled.
2. The test result of this performance degradation is obtained under the version not included the PR.
   In other words, the PR will solve the problem.

PR #3804, opened 2021-11-26
QQ: remove decoded properties before storing messages

kjnilsson:
We're also typically storing the encoded properties as well.
We only really need one. e.g. an enqueue command with a 2 byte payload
serialises to 290 bytes compared to 463. A nice saving.

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -839,7 +839,8 @@ deliver(true, Delivery, QState0) ->
     rabbit_fifo_client:enqueue(Delivery#delivery.msg_seq_no,
                                Delivery#delivery.message, QState0).
 
-deliver(QSs, #delivery{confirm = Confirm} = Delivery) ->
+deliver(QSs, #delivery{confirm = Confirm} = Delivery0) ->
+    Delivery = clean_delivery(Delivery0),
     lists:foldl(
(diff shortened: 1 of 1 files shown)


lhoguin:
We have a function that does this already (or similar): https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit_common/src/rabbit_binary_parser.erl#L149-L157

kjnilsson:
> We have a function that does this already (or similar): https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit_common/src/rabbit_binary_parser.erl#L149-L157

It doesn’t reencode the properties so any headers that got removed (BCC) would re-appear.

lukebakken (commented):
Seems reasonable but don't trust just me :grin:

kjnilsson:
I've done a few dev box perftest runs with `perftest -p -u q1 -x 3 -y 2 -c 2000 --qos 1000 -z 30` and this branch seems to perform a touch better than master (which would make sense). On master I get around 11.5k msg/sec with this I get around 11.8 msg/sec. At the very least it isn't worse!

kjnilsson:
I looked into whether we could avoid re-encoding if, say, there were no headers in the original properties but as we allow near arbitrary modification to properties in plugins such as the [timestamp plugin](https://github.com/rabbitmq/rabbitmq-message-timestamp) it would be hard to validate if modification had happened or not without a complete decode so probably not worth it.

ansd (approved):

PR #3805, opened 2021-11-27
To deactivate limit of all QPids when the limiter has been changed from 'limit' to 'unlimit'.

tomyouyou:
Otherwise, the QPids will still to ask limiter whether it can be sent before delivering.
This will degrade performance, especially when the limiter (channel) and `QPid` are on different nodes.

When 'can_send' is disabled, the test results are as follows:
```
id: test-100147-150, time: 400.016s, sent: 17654 msg/s, returned: 0 msg/s, confirmed: 17658 msg/s, nacked: 0 msg/s, received: 17663 msg/s, min/median/75th/95th/99th consumer latency: 1775/5899/6486/7369/8440 μs, confirm latency: 2171/5581/6127/7026/7911 μs
test stopped (Reached time limit)
id: test-100147-150, sending rate avg: 17630 msg/s
id: test-100147-150, receiving rate avg: 17630 msg/s
```

When limiter and `QPid` are on the same node and 'can_send' is enabled, the test results are as follows:

```
id: test-095229-474, time: 400.015s, sent: 13246 msg/s, returned: 0 msg/s, confirmed: 13247 msg/s, nacked: 0 msg/s, received: 13245 msg/s, min/median/75th/95th/99th consumer latency: 3777/7316/8345/10447/11392 μs, confirm latency: 4074/7308/8257/10336/11341 μs
test stopped (Reached time limit)
id: test-095229-474, sending rate avg: 13317 msg/s
id: test-095229-474, receiving rate avg: 13317 msg/s
```

we have seen, for the message rate, the test showed a 24% drop.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -64,6 +64,8 @@
 -export([check_max_age/1]).
 -export([get_queue_type/1, get_resource_vhost_name/1, get_resource_name/1]).
 
+-export([deactivate_limit_all/2]).
+
 %% internal
 -export([internal_declare/2, internal_delete/2, run_backing_queue/3,
          set_ram_duration_target/2, set_maximum_since_use/2,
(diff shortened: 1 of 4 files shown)


michaelklishin:
@tomyouyou it's still not clear to us what your use case is and what `PerfTest` flags are used. This seemingly affects environments where global QoS is often changed on the fly?

tomyouyou:
> @tomyouyou it's still not clear to us what your use case is and what `PerfTest` flags are used. This seemingly affects environments where global QoS is often changed on the fly?

It can be tested as follows:
1. rabbitmqctl eval 'application:set_env(rabbit, default_consumer_prefetch,{true,300}).' 
   This environment variable can also be configured in the configuration file.
2. Attach a consumer to a classic queue. The QPid will ask the limiter whether the message can be sent before delivering.
3. Run 'basic.qos' with 'global=false,prefetch_count=300'. 
   The Global Qos will be diabled. However, since the QPid is not notified, it will still ask the limiter.

PR #3807, opened 2021-11-29
Try to start after cluster recovery

tomyouyou:
If it fails, the broker will hang here forever.


michaelklishin:
We would appreciate a bit more context on the PRs you submit @tomyouyou. Single sentence descriptions do not give us anything to work with. We won't merge a PR if the intent is not clear to us.

PR #3815, opened 2021-11-30
Introduce a new CLI tool (scope), rabbitmq-tanzu

michaelklishin:
For Tanzu (commercial) plugins to attach their commands to instead of
polluting rabbitmqctl.

Pair: @pjk25


michaelklishin:
@mergifyio backport master

PR #3824, opened 2021-12-01
Return "no offset" (19) code when stored offset is undefined

### LINKED ISSUE #3783: Streams: return appropriate response code when stored offset is undefined
Currently the stream plugin returns OK and offset = 0 if there is no stored offset for a reference in a stream. It should return a more appropriate response to make the different between nothing stored and stored offset = 0.

acogoluegnes:
Fixes #3783


kjnilsson (approved):
LGMT

PR #3825, opened 2021-12-01
Match help box to correct item

johanrhodin:
before it was right after Auto expire
Screen Shot 2021-12-01 at 2 01 40 PM


michaelklishin:
Will be backported to `v3.9.x` for `3.9.11`.

lukebakken (approved):
Tested! Thanks @johanrhodin

PR #3870, opened 2021-12-08
Fix case where confirms may not be sent

lhoguin:
A channel that first sends a mandatory publish before enabling
confirms mode may not receive confirms for messages published
after that. This is because the publish_seqno was increased
also for mandatory publishes even if confirms were disabled.
But the mandatory feature has nothing to do with publish_seqno.

The issue exists since at least
https://github.com/rabbitmq/rabbitmq-server/commit/38e5b687de76739b5419c1f0f6ddf0d8262ea16e

The test case introduced focuses for multiple=false. The issue
also exists for multiple=true but it has a different impact:
sending multiple=true,delivery_tag=2 results in both messages
1 and 2 being acked, even if message 2 doesn't exist as far
as the client is concerned. If the message does exist
it might get confirmed earlier than it should have been. The
issue is a bigger problem the more mandatory messages were
sent before enabling confirms mode.

## Proposed Changes

If English isn't your first language, don't worry about it and try to communicate the problem you are trying to solve to the best of your abilities.
As long as we can understand the intent, it's all good.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_channel.erl
+++ b/deps/rabbit/src/rabbit_channel.erl
@@ -1306,7 +1306,7 @@ handle_method(#'basic.publish'{exchange    = ExchangeNameBin,
     check_expiration_header(Props),
     DoConfirm = Tx =/= none orelse ConfirmEnabled,
     {MsgSeqNo, State1} =
-        case DoConfirm orelse Mandatory of
+        case DoConfirm of
             false -> {undefined, State0};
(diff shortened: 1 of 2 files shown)


lhoguin:
You can remove the fix in rabbit_channel and run the test to trigger the problem.

michaelklishin (approved):

PR #3878, opened 2021-12-10
Get 'writer_gc_threshold' from the environment variables

tomyouyou:
Get 'writer_gc_threshold' from the environment variables to support configurable.


michaelklishin:
Who would need to configure this and why? We always ask for reasoning behind the change, not just what has changed. Please proactively provide one.

tomyouyou:
At present, I don't find any scenarios where this parameter needs to be configured.

PR #3879, opened 2021-12-10
Return appropriate response when stream leader not available

### LINKED ISSUE #3874: Return appropriate response when stream leader not available
`rabbit_stream_manager:lookup_leader/2` function does not make the difference between a non-existing stream and a non-available leader. This leaks up to the client that may think the stream they want to interact with does not exist at all, whereas the leader may not be available yet, especially when the system is starting up.

Original discussion: #3873.

acogoluegnes:
Fixes #3874.


michaelklishin (approved):

michaelklishin:
@acogoluegnes this needs manual backporting due to conflicts

michaelklishin:
Backported to `v3.9.x` manually.

PR #3888, opened 2021-12-13
Remove unnecessary bazel config in template

pjk25:
I've opened this PR to test the running of test suites on external PRs


pjk25:
Though in fact it is a valid change

PR #3908, opened 2021-12-17
Stream coordinator: avoid mnesia update process crashing after delete

kjnilsson:
If a delete happens shortly after a declare or other stream change
there is a chance the mnesia update process that is spawned will crash
when the amqqueue record cannot be recovered from durable storage.

This isn't harmful but does pollute the logs.


Gsantomaggio:
LGTM.

Gsantomaggio (approved):

PR #3911, opened 2021-12-18
Support configure max sync throughput in CMQs

thuandb:
## Proposed Changes

Proposed Changes

We would like to propose a backward compatible change in order to reduce memory utilization during classic mirrored queue sync.

Specifically, in addition to the existing `mirroring_sync_batch_size` configuration key, which controls maximum number of messages per sync batch, we propose to add a new configuration key `mirroring_sync_max_throughput` in bytes per second to control maximum synchronization throughput. The `syncer` at the primary node will make sure the throughput of messages being synchronized does not exceed the value configured. This will give mirroring nodes sufficient time to page messages being synchronized to disk and quickly free up memory.

The motivation for this is that, in our test with a 3-node cluster on AWS with data volumes using [EBS gp2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html) (maximum disk write throughput is 250MiB/s), we observe that the memory utilization during queue sync is high leading to a memory alarm. We suspect that when disk write throughput is low, mirroring nodes take time to page messages to disk while the syncer keeps broadcasting messages at a higher rate. As a result, most messages are kept in memory on the mirror(s) and are waiting to be paged to disk leading to high memory usage.

With this change, if the max sync throughput value is set to less than the write throughput of data volumes, (i.e., the primary node never broadcast data at a speed higher than disk write speed), it will help reduce memory utilization at mirroring node significantly. The new configuration will be disabled by default (`mirroring_sync_max_throughput = 0`) to ensure backward compatibility. I.e., nothing changes to existing/running RabbitMQ clusters. Users will have to explicitly set sync throughput to enable the feature.

Here are samples of memory utilization at the mirroring node and the primary node with different message sizes in tests using RabbitMQ 3.8.22 in a cluster with [EBS gp2 volume](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html) compared with the limit sync throughput feature:

#### RabbitMQ 3.8.22

A test with RabbitMQ 3.8.22 using batch size of 128 shows memory consumption at mirroring node is very high (around memory alarm level); (memory utilization at the primary node is good however).

* Memory consumption during queue sync at the mirroring node is high

image(11)

* Memory consumption during queue sync at the primary node looks good.

image(12)

#### Limit sync throughput

We tested this change with mirroring_sync_max_throughput = 150MiB with similar setting, memory consumption is low at both mirroring node and primary node.

* Memory consumption during queue sync at the mirroring node:

image(13)

* Memory consumption during queue sync at the primary node is low too

image(14)

### Message size of 10MiB

A test with RabbitMQ 3.8.22 using batch size of 128 shows memory utilization is very high at both mirroring node and primary node.

#### RabbitMQ 3.8.22

* Memory consumption during queue sync at the mirroring node is high

image(15)

* Memory consumption during queue sync at the primary node is high too.

image(16)

#### Limit sync throughput

We tested with mirroring_sync_max_throughput = 150MiB, memory consumption drops significantly at both mirroring node and primary node.

* Memory consumption during queue sync at the mirroring node dropped significantly.

image(17)

* Memory consumption during queue sync at the primary node is low as well

image(18)

We also conducted tests with various settings such as using default batch size of 4096 as well as with smaller message sizes; they all show that when sync throughput is limited to a lower level of disk write throughput at mirroring nodes, memory consumed by queue sync processes significantly improves.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

    * We will submit a separate PR for updating doc once this change is approved/merged.

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -40,7 +40,7 @@ git_repository(
 
 git_repository(
     name = "rbe_24",
-    commit = "a087892ef4202dc3245b64d36d5921491848315f",
+    commit = "c8cbf65e2facbe464ebbcee7b6cf6f7a2d422ded",
     remote = "https://github.com/rabbitmq/rbe-erlang-platform.git",
 )
 
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -34,7 +34,7 @@ buildbuddy(
 
 git_repository(
     name = "rbe_23",
-    commit = "d2b454dc5138a2a92de45a0a672241a4fbb5a1e5",
+    commit = "b21c066e426de48e526cc0f8c5158b7024d04e85",
     remote = "https://github.com/rabbitmq/rbe-erlang-platform.git",
 )
 
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -1700,6 +1700,10 @@ end}.
 {mapping, "log.file.rotation.date", "rabbit.log.file.date", [
     {datatype, string}
 ]}.
+{mapping, "log.file.rotation.compress", "rabbit.log.file.compress", [
+    {default, false},
+    {datatype, {enum, [true, false]}}
+]}.
 {mapping, "log.file.rotation.size", "rabbit.log.file.size", [
     {datatype, integer}
 ]}.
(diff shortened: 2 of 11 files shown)


michaelklishin:
@thuandb please rebase this, as it is now 45 commits behind `master`.

michaelklishin:
Moved to #3925.

PR #3934, opened 2021-12-31
Reduce CPU usage of rabbit_framing_amqp_0_9_1:decode_method_fields/2

ansd:
This commit will generate module rabbit_framing_amqp_0_9_1 with
re-ordered clauses of function decode_method_fields/2.

After this change, basic class will be at the top of the function
clauses. That's better because for example basic.publish and basic.ack
are called far more often than for example methods of class connection,
channel or exchange.

Note that "the compiler does not rearrange clauses that match binaries."
See https://www.erlang.org/doc/efficiency_guide/functions.html#pattern-matching

Measurement taken on an Ubuntu 20.04 VM before and after this change:
1. Install latest Erlang from master (i.e. Erlang 25)
2. `RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+JPperf true +S 1" make run-broker`
   (Single scheduler makes the flame graph easier to read.)
3. Run rabbitmq-perf-test without any parameters
4. `sudo perf record -g -F 9999 -p <pid> -- sleep 5`
   where `<pid>` is the output of `os:getpid().` (in the Erlang shell).
   This samples CPU stack traces via frame pointers of
   rabbitmq-server at 9999 Hertz for 5 seconds.
5. Generate a differential flame graph as described in
   https://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html

Before this change, stack frame rabbit_framing_amqp_0_9_1:decode_method_fields/2
was 1.57% present in all stack traces, most of the time (> 1%) running
on the CPU, i.e. directly consuming CPU cycles.
This does not sound like a lot, but is actually quite a lot for a single
function!

The differential flame graph depicts a single dark blue frame: function decode_method_fields
with a reduction of 0.86%:

decode_method_fields


michaelklishin (approved):
Great find!

deps/rabbitmq_codegen/amqp-rabbitmq-0.9.1.json
```diff
@@ -98,6 +98,118 @@
     ],
 
     "classes": [
+        {
+            "id": 60,
+            "methods": [{"id": 10,
```
bollhals, line 103: Just found out about this, nice find! 

@michaelklishin I was wondering would reordering the methods inside the basic also help improve it further?

The first one is qos, which is hardly the most often called one.
I assume it's something like deliver, publish, ack, consume(-ok), ...

bollhals: @ansd @michaelklishin My question probably got lost in the inbox, friendly ping =)

michaelklishin: @bollhals the same idea can be taken further, sure.

PR #3940, opened 2022-01-03
Add a feature flags controller

### LINKED ISSUE #3939: Race condition in the feature flags subsystem
While testing the [integration of Khepri into RabbitMQ](https://github.com/rabbitmq/rabbitmq-server/tree/use-khepri-metadata-store), @mkuratczyk discovered a race condition in the feature flags subsystem.

The scenario is the following one:
1. A cluster is formed.
2. A feature flag is enabled; this feature flag has a migration function which takes a long time.
3. A new node joins the cluster **while** the migration function is still running.
4. The feature flag is marked as enabled on all nodes, including the one which joined during the process.

Depending on what the migration function does or how it determines it should run or not on the new node, it's possible it does not perform its task on the new node, or it corrupts some data.

This is what happens with the Khepri integration branch: the migration functions copies records from Mnesia to Khepri, then delete records in Mnesia. When a node joins during that task, the new node thinks it should run the migration function locally which corrupts the data in Mnesia and Khepri (the migration function is currently not designed to run many times in parallel).

dumbbell:
This gen_statem-based process is responsible for handling concurrency when feature flags are enabled and synchronized when a cluster is expanded.

This clarifies and stabilizes the behavior of the feature flag subsystem w.r.t. situations where e.g. a feature flag migration function takes time to update data and a new node joins a cluster and synchronizes its feature flag states with the cluster. There was a chance that the feature flag was marked as enabled on the joining node, even though the migration function didn't take care of that node.

With this new feature flags controller, enabling or synchronizing feature flags blocks and delays any concurrent operations which try to modify feature flags states too.

This change also clarifies where and when the migration function is called: it is called at least once on each node who knows the feature flag and when the state goes from "disabled" to "enabled" on that node.

Note that even if the feature flag is being enabled on a subset of the nodes (because other nodes already have it enabled), it is marked as "state_changing" everywhere during the migration. This is to prevent that a node where it is enabled assumes it is enabled on all nodes who know the feature flag.

There is a new feature as well: just after a feature flag is enabled, the migration function is called a second time for any post-enable actions. The feature flag is marked as enabled between these "enable" and "post-enable" steps. The success or failure of this "post-enable" run does not affect the state of the feature flag (i.e. it is ignored).

A new migration function API is introduced to allow more advanced things. The new API is:

```erlang
my_migration_function(
  #ffcommand{name = ...,
             props = ...,
             command = enable | post_enable,
             extra = #{...}})
```

The record is defined in `include/feature_flags.hrl`. Here is the meaning of each field:

* `name` and `props` are the equivalent of the `FeatureName` and `FeatureProps` arguments of the previous migration function API.

* `command` is basically the same as the previous `Arg` arguments.

* `extra` is map containing context-specific information. For instance, it contains the list of nodes where the feature flag state changes.

This whole new behavior is behind a new feature flag called `feature_flags_v2`. If a feature flag uses the new migration function API, `feature_flags_v2` will be automatically enabled.

If many feature flags are enabled at once (like when a fresh RabbitMQ node is started for the first time), `feature_flags_v2` will be enabled first if it is in the list.

Fixes #3939.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,8 @@
 !/deps/rabbitmq_cli/
 !/deps/rabbitmq_codegen/
 !/deps/rabbitmq_consistent_hash_exchange/
+!/deps/rabbitmq_ct_helpers/
+!/deps/rabbitmq_ct_client_helpers/
 !/deps/rabbitmq_event_exchange/
 !/deps/rabbitmq_federation/
 !/deps/rabbitmq_federation_management/
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -432,10 +432,15 @@ suites = [
         name = "feature_flags_SUITE",
         size = "large",
         flaky = True,
-        shard_count = 5,
+        shard_count = 9,
         runtime_deps = [
             "//deps/rabbit/test/feature_flags_SUITE_data/my_plugin:erlang_app",
         ],
+    ),
+    rabbitmq_integration_suite(
+        PACKAGE,
+        name = "feature_flags_v2_SUITE",
+        size = "large",
     ),
     rabbitmq_integration_suite(
         PACKAGE,
--- a/deps/rabbit/include/feature_flags.hrl
+++ b/deps/rabbit/include/feature_flags.hrl
@@ -1,0 +1,13 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
+%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%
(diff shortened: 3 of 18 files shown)


michaelklishin:
@dumbbell is this for `v3.10.x` or `3.11`? The `bug` label confuses me somewhat, sorry.

dumbbell:
> @dumbbell is this for `v3.10.x` or `3.11`? The `bug` label confuses me somewhat, sorry.

To repeat our discussion in Slack, the patch fixes a bug in the feature flags synchronization. However, this is a significant rewrite of the synchronization to harden it. Therefore, even though it is a bug fix on the surface, it is not meant to be backported to existing release branches, given the impact on the feature flags subsystem.

This patch should only go into the next minor release branch, e.g. 3.11.

PR #3956, opened 2022-01-05
Distribution listener IP address is hardcoded in listener metadata

tomyouyou:
For security reasons, it is recommended not to listen on the all zero ip address.

Add an item to the advanced configuration file:

``` erl
{kernel, [{inet_dist_use_interface, {8193,291,0,0,0,0,0,1}}]}
```

Use the netstat command to check the IP address of the distribution port(25672):

```
netstat -anp | grep 25672
tcp6       0      0 2001:123::1:25672       :::*                    LISTEN      2075/beam.smp
```

However, `rabbitmqctl status` shows:

```
Interface: [::], port: 25672, protocol: clustering, purpose: inter-node and CLI tool communication
```

After modification, it shows:

```
Interface: [2001:123::1], port: 25672, protocol: clustering, purpose: inter-node and CLI tool communication
```


michaelklishin:
I could reproduce and verify 👍

PR #3963, opened 2022-01-06
Use rules_erlang 2.0.0

pjk25:
rules_erlang is the upcoming bazel-erlang v2, which notably should bring windows support


pjk25:
https://github.com/rabbitmq/rules_erlang/pull/19
https://github.com/rabbitmq/ra/pull/247
https://github.com/rabbitmq/osiris/pull/63

pjk25:
This will likely require a manual backport, only because release branches of rabbitmq-server use a stable rules_erlang release instead of a git commit.

PR #3966, opened 2022-01-07
'rabbit_stream_coordinator:select_leader' runs with wrong comparison

tomyouyou:
The input parameter 'Offsets' consists of candidates which is a tuple {node, tail}, and the tail is made of {epoch, offset}.
While the 'select_leader' think the tail is made of {offset, epoch}.

Suppose there are two candidates:
[{node1,{1,100}},{node2,{2,99}}]

It selects node1 as the leader instead of node2 with larger epoch.


kjnilsson:
Hi I have taken your original commit and created a new PR with the additional changes needed to make this change upgrade safe. See #3967

PR #4043, opened 2022-01-25
Use separate colours for different queue types (and streams) in memory breakdown chart

ansd:
Before:
before

After:
after


michaelklishin:
Should we change the label to say "Queues and Streams" then?

ansd:
Sounds good:
queues-and-streams

Alternatively, we could also have separate colours for classic queue, quorum queue, and streams.

michaelklishin:
@ansd that would be even better, assuming it is reasonably easy to do.

ansd:
Color per queue type:
color-per-queue-type

michaelklishin (approved):

PR #4053, opened 2022-01-28
Content hashing support for definition import

michaelklishin:
## Proposed Changes

This Introduces definition hashing during import 
as an opt-in feature. The goal is to avoid re-importing the definition
from the definition file/directory/source if we know the content
has not changed. Since this feature won't be appropriate for
every environment (sometimes unconditional reimporting is expected),
the feature is opt-in.

This is still a WIP.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

Per discussion with @mkuratczyk.

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -154,10 +154,10 @@ fun(Conf) ->
     end
 end}.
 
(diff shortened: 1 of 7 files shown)


michaelklishin:
Will re-submit.

PR #4073, opened 2022-02-01
Fix performance regression when stream_queue feature is not enabled

kjnilsson:
If the quorum_queue and stream_queue features are not enabled the
classic queues will use the old message format for confirms.

This causes the channel to scan its entire queue type state
map to find the queue name corresponding to the pid. Particularly
when using large ratio fan-outs (1:600) this could cause very poor
performance.

Luckily we can use the queue type's `monitor_registery` to get exactly the
information we need quickly which is what this change implements.

See https://groups.google.com/g/rabbitmq-users/c/1dwZTdhvUWk

perftest command used for testing:

```
perftest \
    --id regres8_3kb_1P_FANOUT_600excl_5m \
    --producers 1 \
    --consumers 600 \
    --size 3000 \
    --time 30 \
    --queue-pattern regres8_3kb_1P_FANOUT_600excl_5m-%d \
    --queue-pattern-from 1 \
    --queue-pattern-to 600 \
    --type fanout \
    --exclusive true \
    --heartbeat-sender-threads 30 \
    --producer-random-start-delay 10 \
    --confirm-timeout 120 \
    --qos 500 \
    --confirm 200 \
    --producer-channel-count 1 \
    --consumer-channel-count 1 \
    --flag false \
    --producer-schedule-threads 100 \
    --consumers-thread-pools  100 \
    --shutdown-timeout 60 \
    --auto-delete true
```

AC: 

1. start broker without any feature flags enabled. e.g. `gmake run-broker PLUGINS="rabbitmq_management" RABBITMQ_FEATURE_FLAGS=`

2. test throughput of above command on master and compare to this branch.

### DIFF
--- a/deps/rabbit/src/rabbit_channel.erl
+++ b/deps/rabbit/src/rabbit_channel.erl
@@ -742,7 +742,7 @@ handle_cast({mandatory_received, _MsgSeqNo}, State) ->
 
 handle_cast({reject_publish, _MsgSeqNo, QPid} = Evt, State) ->
     %% For backwards compatibility
-    QRef = find_queue_name_from_pid(QPid, State#ch.queue_states),
+    QRef = rabbit_queue_type:find_name_from_pid(QPid, State#ch.queue_states),
     case QRef of
         undefined ->
             %% ignore if no queue could be found for the given pid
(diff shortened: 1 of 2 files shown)


lukebakken (approved):
Seems reasonable to me. Hopefully the issue reporter can test this.

adamncasey:
I've tested this patch on top of 3.9.13 in more or less the same setup as the issue reporter & can confirm it brings 3.9.x (with feature flags off) back into the realms of what we are expecting performance wise. I haven't tested this with any topology other than 1:400 fan out yet.

Thank you for fixing this so quickly

motmot80:
This issue got us very hard, too. In our case the "STREAMS" feature wasn't enabled after a rolling upgrade from 3.8.14 to 3.9.13.

**Hint for workaround:** You can enable the features at runtime. This immediately fixes the performance degration problem without restarting the nodes.

PR #4098, opened 2022-02-03
Shovel crashes with delete-after=0, causes RabbitMQ to not boot after

luos:
Hi,

We ran into an issue when a shovel is created with delete-after=0. It is accepted by RabbitMQ as a valid shovel, however it fails to start and it also does not clean up.

Example command:
```
rabbitmqctl set_parameter shovel my-shovel '{"dest-uri": "amqp://guest:guest@localhost", "src-uri": "amqp://guest:guest@localhost", "ack-mode": "on-confirm", "delete-after": 0, "dest-queue": "target", "src-queue": "source", "reconnect-delay": 30 }'
```

It crashes with the following:

```
{error,
         {{exception,
           {case_clause,
            {error,
             {{{case_clause,0},
               [{rabbit_shovel_dyn_worker_sup,init,1,
                 [{file,"rabbit_shovel_dyn_worker_sup.erl"},{line,45}]},
                {supervisor2,init,1,[{file,"supervisor2.erl"},{line,348}]},
                {gen_server,init_it,2,[{file,"gen_server.erl"},{line,417}]},
                {gen_server,init_it,6,[{file,"gen_server.erl"},{line,385}]},
                {proc_lib,init_p_do_apply,3,
                 [{file,"proc_lib.erl"},{line,226}]}]},
              {child,undefined,
               {<<"/">>,<<"test">>},
               {rabbit_shovel_dyn_worker_sup,start_link,
```

I think if the shovel addition is scripted, it is quite possible to have this situation. Therefore I fixed this in a way that a shovel with 0 is accepted, then cleaned up, to make sure script writers don't have to worry about this case. 

If RabbitMQ is restarted with this shovel it will not boot up. The shovel plugin needs to be disabled and the parameter cleared manually. 

Alternatively, the validation could be changed to not support this case. 

Let me know what you think. 

## Proposed Changes

- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbitmq_shovel/src/rabbit_shovel_dyn_worker_sup.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_shovel_dyn_worker_sup.erl
@@ -46,7 +46,7 @@ init([Name, Config0]) ->
             %% always try to reconnect
             <<"never">>                        -> {permanent, N};
             %% this Shovel is an autodelete one
-              M when is_integer(M) andalso M > 0 -> {transient, N};
+              M when is_integer(M) andalso M >= 0 -> {transient, N};
               <<"queue-length">> -> {transient, N}
           end;
         %% reconnect-delay = 0 means "do not reconnect"
--- a/deps/rabbitmq_shovel/src/rabbit_shovel_parameters.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_shovel_parameters.erl
@@ -215,9 +215,9 @@ validate_params_user(#amqp_params_network{}, _User) ->
 
 validate_delete_after(_Name, <<"never">>)          -> ok;
 validate_delete_after(_Name, <<"queue-length">>)   -> ok;
-validate_delete_after(_Name, N) when is_integer(N) -> ok;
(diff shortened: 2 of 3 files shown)


pjk25:
It appears that the tests are currently passing, however since this is a PR, build credentials are not injected and causing the checks to fail. I have made some slight changes to the workflows on master that I hope will alleviate that particular issue going forward.

michaelklishin:
I can't think of a reason to use `0` for the auto-delete timeout. We can both accept this change and a follow-up one that would make such definitions invalid for `3.10`.

luos:
Hi,

It is not a timeout but number of messages in the queue. 
In any case, currently, RabbitMQ accepts it but then will not boot, so effectively it is not supported now either. :-)

michaelklishin:
@luos OK, let's merge this for 3.9 and 3.8 and please submit a PR that makes this combination of parameters invalid for `3.10`, just in case.

michaelklishin:
Will backport all the way to `3.8.x` and assign appropriate milestones to the backport PRs.

PR #4110, opened 2022-02-04
Allocate less memory when invoking lqueue functions

ansd:
Allocate 2 bytes less per `lqueue` operation by changing the `lqueue` state.

Old `lqueue` state:
```
6 = erts_debug:size({0, {[], []}}).
```

New `lqueue` state:
```
4 = erts_debug:size({0, [], []}).
```

This results in less garbage collection when `lqueue` functions are called many thousand times per second (as it is the case with quorum queues).

Before the change:
page-faults-master

Searching for `rabbit_fifo:take_next_msg` matches 1.5% of all page faults.

After the change:
page-faults-lqueue

Searching for `rabbit_fifo:take_next_msg` matches 1.0% of all page faults.

The flame graph was created by running the server:
```
RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+JPperf true +S 1" make run-broker TEST_TMPDIR="$HOME/scratch/rabbit" RABBITMQ_CONFIG_FILE="$HOME/scratch/advanced.config"
```
with following `advanced.config`:
```
[
  {rabbit, [{credit_flow_default_credit, {3200, 1600}},
            {quorum_commands_soft_limit, 256}]}
].
```
starting perf-test:
```
~/workspace/rabbitmq-perf-test$ ./target/rabbitmq-perf-test-2.17.0-SNAPSHOT/bin/runjava com.rabbitmq.perf.PerfTest -x 1 -y 1 -ad false -f persistent -qa "x-queue-type=quorum" -u q1
```
and recording the page faults:
```
sudo perf record -e page-faults -p $(cat $HOME/scratch/rabbit/rabbit@ubuntu/rabbit@ubuntu.pid) -g -- sleep 90
```

Function `lqueue:get/2` is added mainly for https://github.com/rabbitmq/rabbitmq-server/blob/aee6563d619f555a0b469e6a088a4b58e4c89d36/deps/rabbit/src/rabbit_fifo.erl#L1891-L1898 on branch https://github.com/rabbitmq/rabbitmq-server/tree/quorum-queues-v2 where we need to peek the queue for expired messages.

Disadvantage of this PR is that some functions are copied from Erlang OTP's `queue` module.

### DIFF
--- a/deps/rabbit/src/lqueue.erl
+++ b/deps/rabbit/src/lqueue.erl
@@ -21,6 +21,8 @@
 -export([in/2, in_r/2, out/1, out_r/1]).
 %% Less garbage style API
 -export([get/1, get_r/1, peek/1, peek_r/1, drop/1]).
+%% Not present in queue module
+-export([get/2]).
 %% Higher level API
 -export([join/2, fold/3]).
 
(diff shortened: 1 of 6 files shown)


deps/rabbit/src/lqueue.erl
```diff
@@ -7,96 +7,402 @@
...
+    erlang:error(badarg, [Q]).
+
+%% O(1)
+-spec len(Q :: ?MODULE()) -> non_neg_integer().
+len({L, R, F})
+  when is_integer(L), L >= 0, is_list(R), is_list(F) ->
+    L;
+%% accept deprecated state
+len({L, _Q})
+  when is_integer(L), L >= 0 ->
```
pjk25, line 88: Strictly speaking this could guard that `_Q` is actually a tuple of two lists, as elsewhere, like `to_list({L, {R, F}})
   when is_integer(L), L >= 0, is_list(R), is_list(F)`

ansd: Thank you. I pushed a commit which adds the guard.

pjk25 (approved):

PR #4143, opened 2022-02-13
Allow headers exchange to match headers starting with `x-`

ansd:
Before this commit, headers beginning with the string `x-`
were not used to evaluate matches.

This made it impossible to for example route messages
to different target queues based on their `x-first-death-reason header`.
See https://groups.google.com/g/rabbitmq-users/c/XIox73HvInQ

This commit adds a value `any-with-x` to the existing allowed values
`all` and `any` of binding argument `x-match`.
`any-with-x` has the same semantics as `any` but does not filter out
bindings starting with `x-`.

(While we could have added a further value `all-with-x` for completeness
reason, there is currenlty no known use case for such a value.)

### DIFF
--- a/deps/rabbit/src/rabbit_exchange_type_headers.erl
+++ b/deps/rabbit/src/rabbit_exchange_type_headers.erl
@@ -45,10 +45,11 @@ validate_binding(_X, #binding{args = Args}) ->
     case rabbit_misc:table_lookup(Args, <<"x-match">>) of
         {longstr, <<"all">>} -> ok;
         {longstr, <<"any">>} -> ok;
+        {longstr, <<"any-with-x">>} -> ok;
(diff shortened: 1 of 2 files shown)


Avivsalem:
I'd recommend adding "all-with-x" anyway:
1. it's not breaking anything.
2. completeness is important
3. no known use case *currently*, but I can think of some, just around the corner.... (bind x-first-death-reason AND x-first-death-queue...)
4. is the cost of adding it too high?

michaelklishin (approved):

ansd:
Thanks for your feedback @Avivsalem. `all-with-x` is added in https://github.com/rabbitmq/rabbitmq-server/pull/4203.

Avivsalem:
thank you very much!
it's very useful for us!

PR #4261, opened 2022-03-11
Pin unpinned deps

pjk25:
```
ra                   git main -> 2.0.4
gen_batch_server     0.8.6    -> 0.8.7
seshat               git main -> 0.1.0
```


pjk25:
@kjnilsson do we want to backport these changes to v3.9.x as well?

michaelklishin:
@pjk25 we should definitely backport patch release bumps to `v3.9.x`

PR #4275, opened 2022-03-15
File Handle Cache optimizations that benefit classic mirrored queues

felixhuettner:
## Proposed Changes

When trying to use a large amount of durable mirrored queues (non quorum queues) we noticed that the rabbitmq processes spent a lot of time in `ETS` (as reported by `rabbitmq-diagnostics runtime_thread_stats`). When analyzing the details we came to the conclusion that most of this time is spent in lock contention on the `file_handle_cache_stats` ETS table (as reported by lcnt).

This PR tries to improove the situation in two ways:
1. enable `write_concurrency` for the table 
2. add an option to disable the statistics stored in `file_handle_cache_stats` completly

The first point already improves the situation significantly (from 60% collections to 25% as reported by lcnt). 

The second point is not really ideal from my perspective, but it allows us to get our performance requirements fulfilled.
If there is a better solution i would be happy to adapt the change.

Used perftest parameters:
`--auto-delete "false" -f persistent --producer-random-start-delay "30" --queue-pattern perf-test-%d --queue-pattern-from "1" --queue-pattern-to "400" --producers "400" --consumers "500" --heartbeat-sender-threads "10" --rate "10"`

Changes in the amount of consumed messages after each change:
* Without changes we get to 500-1000msg/sec
* With the first change we get to around 1500msg/sec
* The last change gives us a stable 4000msg/sec

We also noticed that rabbitmqctl commands improved significantly in speed after each change.

lcnt statistics after the first point is implemented:
```
 lock                                                                             id  #tries  #collisions  collisions [%]  time [us]  duration [%]
 -----                                                                            --- ------- ------------ --------------- ---------- -------------
db_hash_slot                               file_handle_cache_stats  518404       123849         23.8904  161167277   157.1665
db_hash_slot                               file_handle_cache_stats  262166        13130          5.0083   22819879       22.2534
<snip>
```

## Types of Changes

- [ ] Build system and/or CI

Honestly not sure about this one.

## Checklist

- [in progress] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)

I am unsure how ad where to add tests for this solution as this is also my first time writing erlang.
If you could advise me there i would be happy to introduce apply these changes.

## Further Comments

I came to the conclusion that `file_handle_cache_stats` is only used for statistics provided to the administrators. Therefor i implemented the changes by disabeling the statistics completely.
If i missunderstood this and the statistics are used inside rabbitmq then this change might need to be adapted.

Thank you

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -1243,6 +1243,14 @@ end}.
 {mapping, "collect_statistics_interval", "rabbit.collect_statistics_interval",
     [{datatype, integer}]}.
 
+%% Enable/disable file_handle_cache statistics. Disabeling this
+%% can help in cases of lock contention on the backing ets table.
+%%
+%% {classic_queue.collect_fd_stats, true},
+
+{mapping, "classic_queue.collect_fd_stats", "rabbit.classic_queue_collect_fd_stats",
+    [{datatype, {enum, [true, false]}}]}.
+
 %%
 %% Misc/Advanced Options
 %% =====================
--- a/deps/rabbit_common/src/file_handle_cache_stats.erl
+++ b/deps/rabbit_common/src/file_handle_cache_stats.erl
@@ -31,20 +31,32 @@ init() ->
 
 update(Op, Bytes, Thunk) ->
     {Time, Res} = timer_tc(Thunk),
-    _ = ets:update_counter(?TABLE, {Op, count}, 1),
-    _ = ets:update_counter(?TABLE, {Op, bytes}, Bytes),
-    _ = ets:update_counter(?TABLE, {Op, time}, Time),
-    Res.
+    case application:get_env(rabbit, classic_queue_collect_fd_stats) of
+        {ok, false} -> Res;
+        {ok, true}  ->
+            _ = ets:update_counter(?TABLE, {Op, count}, 1),
+            _ = ets:update_counter(?TABLE, {Op, bytes}, Bytes),
+            _ = ets:update_counter(?TABLE, {Op, time}, Time),
+            Res
+    end.
 
 update(Op, Thunk) ->
     {Time, Res} = timer_tc(Thunk),
-    _ = ets:update_counter(?TABLE, {Op, count}, 1),
-    _ = ets:update_counter(?TABLE, {Op, time}, Time),
-    Res.
+    case application:get_env(rabbit, classic_queue_collect_fd_stats) of
+        {ok, false} -> Res;
+        {ok, true}  ->
+            _ = ets:update_counter(?TABLE, {Op, count}, 1),
+            _ = ets:update_counter(?TABLE, {Op, time}, Time),
+            Res
+    end.
 
 update(Op) ->
-    ets:update_counter(?TABLE, {Op, count}, 1),
-    ok.
+    case application:get_env(rabbit, classic_queue_collect_fd_stats) of
+        {ok, false} -> ok;
+        {ok, true}  ->
+            ets:update_counter(?TABLE, {Op, count}, 1),
+            ok
+    end.
 
 get() ->
     lists:sort(ets:tab2list(?TABLE)).
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -69,6 +69,7 @@ _APP_ENV = """[
 	    {server_properties, []},
 	    {collect_statistics, none},
 	    {collect_statistics_interval, 5000},
+	    {classic_queue_collect_fd_stats, true},
 	    {mnesia_table_loading_retry_timeout, 30000},
 	    {mnesia_table_loading_retry_limit, 10},
 	    {auth_mechanisms, ['PLAIN', 'AMQPLAIN']},
--- a/deps/rabbit/Makefile
+++ b/deps/rabbit/Makefile
@@ -44,6 +44,7 @@ define PROJECT_ENV
 	    {server_properties, []},
 	    {collect_statistics, none},
 	    {collect_statistics_interval, 5000},
+	    {classic_queue_collect_fd_stats, true},
 	    {mnesia_table_loading_retry_timeout, 30000},
 	    {mnesia_table_loading_retry_limit, 10},
 	    {auth_mechanisms, ['PLAIN', 'AMQPLAIN']},
--- a/deps/rabbit/test/unit_gm_SUITE.erl
+++ b/deps/rabbit/test/unit_gm_SUITE.erl
@@ -35,6 +35,7 @@ all() ->
     ].
 
 init_per_suite(Config) ->
+    ok = application:set_env(rabbit, classic_queue_collect_fd_stats, true),
     ok = application:set_env(mnesia, dir, ?config(priv_dir, Config)),
     ok = application:start(mnesia),
     {ok, FHC} = file_handle_cache:start_link(),


deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -1243,6 +1243,14 @@ end}.
 {mapping, "collect_statistics_interval", "rabbit.collect_statistics_interval",
     [{datatype, integer}]}.
 
+%% Enable/disable file_handle_cache statistics. Disabeling this
+%% can help in cases of lock contention on the backing ets table.
+%%
+%% {collect_file_handle_cache_statistics, true},
+
+{mapping, "collect_file_handle_cache_statistics", "rabbit.collect_file_handle_cache_statistics",
```
michaelklishin, line 1251: I'd much prefer `file_handles.collect_statistics` or something like that. `collect_statistics_interval` was a name chosen in 2010 or so.

deps/rabbit_common/src/file_handle_cache_stats.erl
```diff
@@ -31,20 +31,32 @@ init() ->
 
 update(Op, Bytes, Thunk) ->
     {Time, Res} = timer_tc(Thunk),
-    _ = ets:update_counter(?TABLE, {Op, count}, 1),
-    _ = ets:update_counter(?TABLE, {Op, bytes}, Bytes),
-    _ = ets:update_counter(?TABLE, {Op, time}, Time),
-    Res.
+    case application:get_env(rabbit, collect_file_handle_cache_statistics) of
```
michaelklishin, line 34: Using `application:get_env/2` on any hot code path is an absolute no-no because all of those requests go to a single process.

deps/rabbit_common/src/file_handle_cache_stats.erl
```diff
@@ -31,20 +31,32 @@ init() ->
...
+            _ = ets:update_counter(?TABLE, {Op, time}, Time),
+            Res
+    end.
 
 update(Op, Thunk) ->
     {Time, Res} = timer_tc(Thunk),
-    _ = ets:update_counter(?TABLE, {Op, count}, 1),
-    _ = ets:update_counter(?TABLE, {Op, time}, Time),
-    Res.
+    case application:get_env(rabbit, collect_file_handle_cache_statistics) of
```
michaelklishin, line 45: Same as above, no `application:get_env/2` on any hot code paths.

michaelklishin (changes requested):
I am highly skeptical on the idea. Using `application:get_env/2` on any hot code path
is a no-no, and speeding something up by disabling ETS updates but then querying a single process (the application controller) sounds like a change not worth making.

In what kind of scenario does the FHC become the bottleneck?

michaelklishin:
Classic mirrored queues is a design we won't spend much time on. In fact, 3.10 will ship a new storage implementation for classic queues, so this may or may not be relevant at all. There are throughput gains with the new index and message store,
maybe @lhoguin still has some comparison numbers to share.

michaelklishin:
We can enable `write_concurrency` for the table, by the way, if there's evidence that it helps.

The only "statless" alternative I can think of is allowing you to use a custom FHC module, which would not have statistics collection. A trivial plugin would do it. It can be quite a set of changes all across the codebase to make that a reality,
so would be pretty hard to justify for CMQs specifically and in the light of CQv2 storage shipping quite soon in 3.10.

michaelklishin:
@lhoguin suggests that my understanding of `application:get_env/2` internals is outdated and it is basically an ETS read.

So the "second change" gains then likely come from replacing an ETS write with an ETS read when statistics collection is disabled.

lhoguin:
Hello!

I think both changes are good, I only have a few questions as well as a few comments on what I think we want to accept. Let's start with the questions!

- How many queues do you have to produce this much lock contention? I will be looking to reproduce the problem to determine whether the concurrency option or the disabling helps as part of QA-ing this PR.
- Have you tried with non-mirrored queues? Considering the code I would expect them to also benefit in a similar amount. We do not do CMQs anymore but we do want better (and simpler) CQs.
- I do wonder about the effect of adding the `get_env` in environments that enable it. I would not want to make things worse by default. Storing the value in the process dictionary may help in that case.

As for the changes I seek, for the time being:

- Please rename the configuration option to be `classic_queue.collect_stats` and the app env option to be `classic_queue_collect_stats`.
- Some whitespace is wrong.

michaelklishin:
`class_queue.collect_stats` might suggest that all CQ stats will be disabled. If this is focused on FHC metrics only, then it should be `file_handle_cache.collect_stats`.

lhoguin:
Then `classic_queue.collect_fd_stats`. I do not want to refer to `file_handle_cache` because it will get removed.

felixhuettner:
Thank you for all the feedback. I'll include your feedback in the changes within the next days.

felixhuettner:
I can now share the results of our measurements in different cases.

# Tests

We have now tested the change with
1. mirrored queues and durable messages
2. non-mirrored queues and durable messages

The tests where done using the official 3.9 docker image in 3 scenarios:
1. using the default image with no changes
2. including the changes in this PR in the image, but leaving `classic_queue.collect_fd_stats` at its default value
3. including the changes in this PR and setting `classic_queue.collect_fd_stats` to false

In tests we measured the throughput in terms of msg/sec.
We also took a snapshot of `rabbitmq-diagnostics runtime_thread_stats` one each node during the testrun.

All of these tests where run on a 3 node cluster with inter-node tls enabled.
We let each test run for 3 minutes.

# Results for mirrored + durable

This test was run sending 4000 msg/sec over a total of 400 queues

| Case | Median recived msg/sec | Avg received msg/sec | Avg ets usage |
| --- | --- | --- | --- | 
| Official image | 609 | 697 | 92% |
| Patched image without setting | 2651 | 2893 | 28% |
| Patched image with fd_stats disabled | 3153 | 3222 | 1% |

The patch already provides a significant benefit and setting `classic_queue.collect_fd_stats` has a small additional gain.

# Results for non-mirrored + durable

This test was run sending 10000 msg/sec over a total of 1000 queues

| Case | Median recived msg/sec | Avg received msg/sec | Avg ets usage |
| --- | --- | --- | --- |
| Official image | 2541 | 2548 | 78% |
| Patched image without setting | 3274 | 4095 | 79% | 
| Patched image with fd_stats disabled | 9979 | 9990 | 1% |

The patch provides a small gain in throughput (but no change for the ets usage). Setting `classic_queue.collect_fd_stats` sigificantly changes the throughput and the ets usage.

# Reproduction Information

the following policy was applied for the mirrored queues: `{"expires":3600000,"ha-mode":"all","ha-promote-on-failure":"always","ha-promote-on-shutdown":"always","ha-sync-mode":"manual","message-ttl":600000,"queue-master-locator":"client-local"}`

the following settings where used for the mirrored/durable perftest: `--auto-delete "false" -f persistent --producer-random-start-delay "30" --queue-pattern perf-test-%d --queue-pattern-from "1" --queue-pattern-to "400" --producers "400" --consumers "500" --heartbeat-sender-threads "10" --rate "10"`

the following settings where used for the non-mirrored/durable perftest: `--auto-delete "false" -f persistent --producer-random-start-delay "30" --queue-pattern reply_perf-test-%d --queue-pattern-from "1" --queue-pattern-to "1000" --producers "1000" --consumers "1000" --heartbeat-sender-threads "10" --rate "10"`

If there is interest i can also share the raw data from the tests (e.g. the output of perftest)

felixhuettner:
I have updated the PR witht the requested changes.

@michaelklishin i have for now not split out the `write_concurrency` change to a separate PR as i did not want to spread out all the measurements over two PRs.

It also looks like using `get_env` has no measureable sideeffect (at least in the above tests)

michaelklishin:
@felixhuettner I do not see any new PRs from you open. Please submit one?

felixhuettner:
@michaelklishin i have now moved the write_concurrency change to this PR https://github.com/rabbitmq/rabbitmq-server/pull/4324

michaelklishin:
@lhoguin we have partially merged this, are you happy with the state of this PR?

lhoguin:
I noted that the commit message has the wrong option name. Hold on before updating though as I am testing this and might have further comments.

lhoguin:
Hello! I think the problem is that the v1 index simply increments counters far too often, including for writes that never end up on the disk! To compare, the v2 index only increments those same counters when doing disk operations and as a result disabling stats for v2 does not improve anything. So I think the right way forward is to modify the v1 stats to not hit the ets table that often, especially considering how little utility some of these have.

I suspect removing the `queue_index_journal_write` stats (which are incremented on publishes and acks, so not super useful as far as I'm concerned) to give the largest boost. Could you try commenting these lines and see if that's good enough?

* https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit/src/rabbit_queue_index.erl#L408
* https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit/src/rabbit_queue_index.erl#L422
* https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit/src/rabbit_queue_index.erl#L1095

For the record: I could not reproduce on my macOS laptop, only on Linux thanks to @mkuratczyk's help.

ihtiuh:
Thanks @lhoguin. @felixhuettner will be back with the updates next week, as he is away.

lhoguin:
I have opened https://github.com/rabbitmq/rabbitmq-server/pull/4431 with a proposal. It disables the expensive (and IMO not terribly useful) stats. This should provide the same performance as disabling everything, but we can keep the useful stats working. Please try it and report back!

felixhuettner:
I'm finally back and managed to test your changes in #4431 and compare them against the change in here. Please find the results below

# Results for mirrored + durable

This test was run sending 4000 msg/sec over a total of 400 queues

| Case | Median recived msg/sec | Avg received msg/sec | Avg ets usage |
| --- | --- | --- | --- | 
| Patched image (PR 4275) with fd_stats disabled | 3153 | 3222 | 1% |
| Patched image (PR 4431) | 3989 | 3999 | 9% |

# Results for non-mirrored + durable

This test was run sending 10000 msg/sec over a total of 1000 queues

| Case | Median recived msg/sec | Avg received msg/sec | Avg ets usage |
| --- | --- | --- | --- |
| Patched image (PR 4275) with fd_stats disabled | 9979 | 9990 | 1% |
| Patched image (PR 4431) | 9997 | 9999 | 1% |

# Results

from my perspective we can go with #4431 and skip this change here

lhoguin:
Thanks for the feedback!

PR #4292, opened 2022-03-18
The wrong garbage_collection effect was used when purging a quorum queue

tomyouyou:


michaelklishin:
Please resolve the conflicts. Your master branch is likely out of date.

michaelklishin:
This is not wrong. [`garbage_collection`](https://github.com/rabbitmq/ra/blob/main/src/ra_server_proc.erl#L1213) is a valid effect. @kjnilsson confirmed this.

PR #4004, opened 2022-01-18
Delete obsolete 'unacked_message_q' entries when queue leader replica restarts

tomyouyou:
If the client acks very slowly and the qpid restarts abnormally, the following may happen:

2022-01-17 10:01:43.828503+08:00 [info] <0.1693.0> [DBG]["rabbit_channel.erl":2192 record_sent]unacked_message_q num:10
2022-01-17 10:01:43.828591+08:00 [info] <0.1693.0> [DBG]["rabbit_channel.erl":2193 record_sent]{pending_ack,1, <<"msg_test_ctag_0">>, 1642384899892, {resource,<<"/">>,queue, <<"q12">>}, 0}
2022-01-17 10:01:43.828696+08:00 [info] <0.1693.0> [DBG]["rabbit_channel.erl":2193 record_sent]{pending_ack,2, <<"msg_test_ctag_0">>, 1642384899894, {resource,<<"/">>,queue, <<"q12">>}, 1}
2022-01-17 10:01:43.828777+08:00 [info] <0.1693.0> [DBG]["rabbit_channel.erl":2193 record_sent]{pending_ack,3, <<"msg_test_ctag_0">>, 1642384899896, {resource,<<"/">>,queue, <<"q12">>}, 2}
2022-01-17 10:01:43.828850+08:00 [info] <0.1693.0> [DBG]["rabbit_channel.erl":2193 record_sent]{pending_ack,4, <<"msg_test_ctag_0">>, 1642384899897, {resource,<<"/">>,queue, <<"q12">>}, 3}
2022-01-17 10:01:43.828922+08:00 [info] <0.1693.0> [DBG]["rabbit_channel.erl":2193 record_sent]{pending_ack,5, <<"msg_test_ctag_0">>, 1642384899899, {resource,<<"/">>,queue, <<"q12">>}, 4}
2022-01-17 10:01:43.828997+08:00 [info] <0.1693.0> [DBG]["rabbit_channel.erl":2193 record_sent]{pending_ack,6, <<"msg_test_ctag_0">>, 1642384903822, {resource,<<"/">>,queue, <<"q12">>}, 0}
2022-01-17 10:01:43.829066+08:00 [info] <0.1693.0> [DBG]["rabbit_channel.erl":2193 record_sent]{pending_ack,7, <<"msg_test_ctag_0">>, 1642384903824, {resource,<<"/">>,queue, <<"q12">>}, 1}
[...]

PR #4305, opened 2022-03-18
Clear unacknowledged messages in channel state when queue leader restarts

michaelklishin:
This is #4004 by @tomyouyou, rebased against `master`.

## Problem Definition

In a system with slow-acknowledging consumers that use manual acknowledgements, a queue leader restart
with acknowledgements flowing after consumer reconnection can result in extra messages being acknowledged.

## Solution

With this change, deliveries pending acknowledgements ("UAMGs" or "unacknowledged messages" in channel state)
are tracked with some extra bells and whistles:

 * When channel detects queue leader failure, all pending UAMGs for that queue are moved to
    a separate list called "outdated pending acknowledgements". This makes overly eager acking
    reported in #4004 impossible when a "stale" consumer ack comes in
 * When an unknown delivery tags comes in (there are no matches in UAMGs on this channel), instead
    of immediately returning a `406 precondition failed` error, we consult the list of outdated pending acks.
    If a pending ack with the same delivery tag is found there, it is returned for settlement with the queue
    (acknowledgement at the queue level)

Compared to #4004, this PR tracks actual `#pending_ack` records and not just delivery tags as outdated.
This makes it possible to actually settle stale acknowledgements (we still have all the context of the delivery,
namely queue reference and message ID).

## Steps to Reproduce the Issue

Reproduction of this issue is a pain because it's timing dependent.

From #4004: 

1. Make preparations, such as creating queues, etc

``` shell
date;python msg_test.py create_queue localhost 'queue=q12;durable=false' 
date;python msg_test.py create_exchange localhost 'exchange=ex11;exchange_type=topic'
date;python msg_test.py bind localhost 'exchange=ex11;queue=q12;routing_key=exq11'
```

2. Create a consumer, it will send an acknowledgement after 10s when receiving message.

``` shell
date;python msg_test.py consume localhost 'queue=q12;auto_ack=false;interval=10'
```

3. Publish five messages.
  
``` shell
date;python msg_test.py publish localhost 'exchange=ex11;routing_key=exq11;msg_hdr="_unique_id:123";count=5'
```

4. Before the consumer answers, kill the leader pid of the target queue and it will restart automatically.
  
5. Publish another five messages, so ten messages were published in total.  

6. After waiting for the consumer to answer the first five messages, restart the consumer immediately.
 
Note: The new queue leader process will mistakenly think that these first five acknowledgements are the replies of the last five messages. Therefore, the last five messages will be deleted from the queue.

```  shell
date;python msg_test.py consume localhost 'queue=q12;auto_ack=false;interval=10'
# => Wed Jan 19 09:38:11 CST 2022
# => 2022-01-19 09:38:11.699 start the 0 th thread
# => [0]connect to 10.228.103.136 ok
# => 2022-01-19 09:38:11.732  [*] Waiting for messages from q12. To exit press CTRL+C
# => 2022-01-19 09:39:32.670  [0] Received '_unique_id:"_unique_' ack:True delivery_tag:1 red:False sleep:10, average_time:0 tag:None count:1
# => 2022-01-19 09:39:42.674  [0] Received '_unique_id:"_unique_' ack:True delivery_tag:2 red:False sleep:10, average_time:0 tag:None count:2
# => 2022-01-19 09:39:52.677  [0] Received '_unique_id:"_unique_' ack:True delivery_tag:3 red:False sleep:10, average_time:0 tag:None count:3
# => 2022-01-19 09:40:02.680  [0] Received '_unique_id:"_unique_' ack:True delivery_tag:4 red:False sleep:10, average_time:0 tag:None count:4
# => 2022-01-19 09:40:12.690  [0] Received '_unique_id:"_unique_' ack:True delivery_tag:5 red:False sleep:10, average_time:0 tag:None count:5

date;python msg_test.py consume localhost 'queue=q12;auto_ack=false;interval=10'
# => Wed Jan 19 09:40:15 CST 2022
# => 2022-01-19 09:40:15.909 start the 0 th thread
# => [0]connect to 10.228.103.136 ok
# => 2022-01-19 09:40:15.932  [*] Waiting for messages from q12. To exit press CTRL+C


rabbitmqctl list_queues
# => Timeout: 60.0 seconds ...
# => Listing queues for vhost / ...
# => name    messages        messages_ready  messages_unacknowledged consumers
# => q12     0       0       0       1
```

This command shows that there are no messages in the queue, so the new consumer will never receive the last five messages.
  
In other words, there are ten messages in total, while consumers can only process the first five messages. The last five messages were lost.

### DIFF
--- a/deps/rabbit/src/rabbit_channel.erl
+++ b/deps/rabbit/src/rabbit_channel.erl
@@ -120,11 +120,18 @@
           writer_gc_threshold
          }).
 
--record(pending_ack, {delivery_tag,
+-record(pending_ack, {
+                      %% delivery identifier used by clients
+                      %% to acknowledge and reject deliveries
+                      delivery_tag,
+                      %% consumer tag
                       tag,
                       delivered_at,
-                      queue, %% queue name
-                      msg_id}).
+                      %% queue name
+                      queue,
+                      %% message ID used by queue and message store implementations
+                      msg_id
+                    }).
 
 -record(ch, {cfg :: #conf{},
              %% limiter state, see rabbit_limiter
@@ -169,7 +176,7 @@
              publishing_mode = false :: boolean(),
              %% delivery tags that were pending consumer acknowledgement
              %% but became outdated because queue leader process failed
-             outdated_delivery_tags = []
+             outdated_pending_acks = []
             }).
 
 -define(QUEUE, lqueue).
@@ -1343,11 +1350,11 @@ handle_method(#'basic.ack'{delivery_tag = DeliveryTag,
                            multiple     = Multiple},
               _, State = #ch{unacked_message_q      = UAMQ,
                              tx                     = Tx,
-                             outdated_delivery_tags = ODTags}) ->
-    {Acked, Remaining, ODTags1} = collect_acks(UAMQ, DeliveryTag, Multiple, ODTags),
-    State1 = State#ch{unacked_message_q = Remaining, outdated_delivery_tags = ODTags1},
+                             outdated_pending_acks  = OPAs}) ->
+    {Acked, Remaining, OPAs1} = collect_acks(UAMQ, DeliveryTag, Multiple, OPAs),
+    State1 = State#ch{unacked_message_q = Remaining, outdated_pending_acks = OPAs1},
     {noreply, case Tx of
-                  none         -> {State2, Actions} = ack(Acked, State1),
+                  none         -> {State2, Actions} = settle_acks(Acked, State1),
                                   handle_queue_actions(Actions, State2);
                   {Msgs, Acks} -> Acks1 = ack_cons(ack, Acked, Acks),
                                   State1#ch{tx = {Msgs, Acks1}}
@@ -1727,7 +1734,7 @@ handle_method(#'tx.commit'{}, _, State = #ch{tx      = {Deliveries, Acks},
     Rev = fun (X) -> lists:reverse(lists:sort(X)) end,
     {State2, Actions2} =
         lists:foldl(fun ({ack,     A}, {Acc, Actions}) ->
-                            {Acc0, Actions0} = ack(Rev(A), Acc),
+                            {Acc0, Actions0} = settle_acks(Rev(A), Acc),
                             {Acc0, Actions ++ Actions0};
                         ({Requeue, A}, {Acc, Actions}) ->
                             {Acc0, Actions0} = internal_reject(Requeue, Rev(A), Limiter, Acc),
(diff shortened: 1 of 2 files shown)


michaelklishin:
@tomyouyou do you agree with the change in https://github.com/rabbitmq/rabbitmq-server/pull/4305/commits/f39d79bd511e6250dfd84d49129093251139c4bf?

After logging settled message IDs I realize it might have been an intentional decision to not settle those "stale acks".
I am not convinced this is a better solution than unintentionally acknowledging some extra messages
but with the above commit, I could not observe any issues. My guess is that you can only do that by inspecting
message store state, which I will do with some help from someone who's much more familiar with the current state of that subsystem.

So, to simplify my question: was the decision to not propagate outdated acks (settle) to the message store intentional in #4004? Did you realize that it would leak data on disk?

michaelklishin:
To be extra specific, when I say "I observe no problems" I mean both various test suites and the above steps. At least
when I let the consumer recover without restarting it, all 10 messages are consumed and settled as expected. All metrics
seem fine. The only thing left to inspect is the message store state.

michaelklishin:
Alternatively, this can be addressed by making message IDs carry an "epoch" of some kind (like in Raft or some other protocols) that would change message IDs. I suspect this would be a much more involved change and channels would have
to do even more work tracking queue leader replica restarts. Not sure if that would be worth the effort :/

michaelklishin:
I've decided to not proceed with this PR or #4004 as it does as much harm as it does good.

The fundamental issue is that
even if we keep track of `#pending_ack`s in a way that allows us to tell which are "outdated" and which are not, we cannot
use their message IDs for settlement with the queue leader/message store/Raft log. This is because message IDs used
by queue processes will be identical between queue leader terms ("epochs").

Without somehow changing that
we either would potentially delete the wrong messages (not what the client intended to acknowledge but with the same message ID as used by the queue leader) or leak data on disk. For quorum queues, not settling deliveries will be
particularly problematic and may drive a node out of disk. We have seen this in the wild for sure.

To be fair, this problem requires a very specific set of events to happen, namely a queue leader replica failure
that does not involve a channel failure (so, consumers must be connected to a different node or queue replica
has to fail with the rest of the node staying up), a reasonably wide
window of unacknowledged deliveries, plus concurrent publishing to the same queue. All of these, of course,
are not impossible but fairly rare. Solving it with an on disk data leak is too much of a price to pay, **in particular for quorum queues**.

Thank you for bringing this up @tomyouyou. I'll ask someone familiar with message store if we can realistically do something
about message IDs without breaking existing installations.

michaelklishin:
I will also submit all the naming improvements from this PR as a separate one.

PR #4327, opened 2022-03-22
CQv2 index: Increase threshold before flushing to disk

lhoguin:
Before it was flushing when there was 100 entries which was way
too low, especially since entries do not include the message
content. Change it to flushing when there's a segment file
worth of entries. Also change updates to flush them when
there's half a segment file worth of updates (which in
practice doesn't change much from before).

Huge shootout to Michal Kuratczyk for all the testing and
bugging me long enough to make me understand something was
wrong.


michaelklishin:
@lhoguin @mkuratczyk does this delivery any measurable improvements? ;)

mkuratczyk (approved):
pr-4327

Without this change (yellow lines) - could not reach 10k msg/s and and the publisher confirm latency in this test was ~850ms.
With this change (blue line) - 27k msg/s at ~250ms latency.
The other lines are different values we tried before settling on this PR.

Approved! :)

michaelklishin:
Backported manually to `v3.10.x`. mergifyio bot is having a day off I guess ;)

lhoguin:
Thanks!

PR #4343, opened 2022-03-23
Make Quorum queue snapshot interval configurable

kjnilsson:
And increase the default from 4096 to 8192. This should help reducing the
number of snapshots taken by high throughput near-empty queues.


michaelklishin:
We could also expose this via `rabbitmq.conf` but I guess it really is an advanced configuration setting, so `advanced.config` would suffice for now.

michaelklishin (approved):

PR #4354, opened 2022-03-24
Use ra 2.0.6

pjk25:
Use `ra 2.0.6` from hex.pm


michaelklishin:
Marking this PR and its backport so that it's easier for us to track dependency updates in release notes.

PR #4368, opened 2022-03-29
Make Osiris listener registration idempotent

acogoluegnes:
Several listeners for the same offset can be registered,
which causes problems for low rate, where consumers hit
the end of the stream often. Then Osiris messages accumulate
in the process mailbox, which slows it down considerably after
some time.

This commit makes the listener registration idempotent for a given
offset.

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
@@ -42,7 +42,7 @@
          properties :: map()}).
 -record(consumer,
         {configuration :: #consumer_configuration{},
-         credit :: integer(),
+         credit :: non_neg_integer(),
          log :: osiris_log:state(),
          last_listener_offset = undefined :: undefined | osiris:offset()}).
 -record(stream_connection_state,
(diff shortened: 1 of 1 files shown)


deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -41,8 +41,10 @@
          counters :: atomics:atomics_ref(),
          properties :: map()}).
 -record(consumer,
-        {configuration :: #consumer_configuration{}, credit :: integer(),
-         log :: osiris_log:state()}).
+        {configuration :: #consumer_configuration{},
+         credit :: integer(),
```
kjnilsson, line 45: can credit ever be negative or should this be `non_neg_integer()`?

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -1923,19 +1932,25 @@ handle_frame_post_auth(Transport,
...
                              SendFileOct)
             of
                 {error, closed} ->
                     rabbit_log_connection:info("Stream protocol connection has been closed by "
                                                "peer",
                                                []),
                     throw({stop, normal});
-                {{segment, Log1}, {credit, Credit1}} ->
-                    Consumer1 = Consumer#consumer{log = Log1, credit = Credit1},
+                {{segment, Log1}, {credit, Credit1},
```
kjnilsson, line 1948: `send_chunks` takes a Consumer as an argument, why not just modify it there and return the updated consumer?

kjnilsson (approved):
Some suggestions for tweaks / refactoring.

PR #4400, opened 2022-03-31
CQv2: Consolidate index writes when possible

lhoguin:
This is done as an extra pass before giving the data
to file:pwrite. This should boost performance in most
cases, especially with large multi-acks.

While it provides a boost in the tested scenarios, I want a more thorough comparison before removing Draft status.

### DIFF
--- a/deps/rabbit/src/rabbit_classic_queue_index_v2.erl
+++ b/deps/rabbit/src/rabbit_classic_queue_index_v2.erl
@@ -715,9 +715,10 @@ flush_buffer(State0 = #qi { write_buffer = WriteBuffer0,
              AcksAcc}
     end, {#{}, #{}, #{}}, WriteBuffer0),
     %% Then we do the writes for each segment.
-    State = maps:fold(fun(Segment, LocBytes, FoldState0) ->
+    State = maps:fold(fun(Segment, LocBytes0, FoldState0) ->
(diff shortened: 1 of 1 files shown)


mkuratczyk:
Great stuff! This PR improves the performance in all the tests I've run. 20-30% throughput improvement in most cases, and basically twice as good after turning off the consumers - on the `Messages Ready & Unacked` graph, you can see how much faster it reached the max length and how much faster the queues were emptied after turning consumers back on but publishers off.

Not directly related to this change but worth nothing how much more predictable lazy queues are - lower memory usage AND in most cases - the same (or even higher!) and more stable throughput.

pr4400

michaelklishin:
I assume you mean lazy CQv2s are more predictable in terms of throughput stability compared to non-lazy CQv2s, not other queue types :)

mkuratczyk:
@michaelklishin I'd say it applies more broadly. For example this is CQv1 lazy (yellow) and non-lazy (green). There are moment when non-lazy has a bit higher throughput but I'd take lazy any day (for this workload at least).

cqv1

michaelklishin (approved):

michaelklishin:
@mkuratczyk yes, we have seen this with CQv1. I thought you were comparing CQv2 with quorum queues.

PR #4404, opened 2022-04-01
Remove unused code for queue and consumer args

ansd:
Before this commit, rabbit_stream_queue supported nonsensical
queue arguments such as
* x-dead-letter-exchange
* x-dead-letter-routing-key
* x-single-active-consumer

It seems that the listed queue and consumer arguments of classic, quorum, and
stream queues are nowhere validated.

While we could add validation, it will break all client applications
that have accidentally used wrong arguments.


michaelklishin:
They can use as a reference. I'm not sure we have an alternative to these lists.

michaelklishin:
@ansd `unsupported_policy_keys` makes sense to me. Note that such keys don't have to result in errors, they can produce warnings. Policy changes is a relatively rare event, on the other hand, they can match tens of thousands of queues. So maybe an error would be a good thing.

PR #4442, opened 2022-04-05
Add quorum queue-leader-locator

ansd:
### Changes

**Prior** to this PR:
1. When a new quorum queue was created, the local node + **random** nodes were selected as replicas. Nodes that are `down` were chosen with same probability than nodes that are `running`.
2. Always **local** node became leader.

For example, when a client connects to a single RabbitMQ node and creates N quorum queues, all N leaders will be on that node. Also, if N is small and the RabbitMQ cluster has many nodes (e.g. 7), some nodes might not host any quorum queue replicas at all.

**After** this PR:
1. When a new quorum queue is created, the local node + RabbitMQ nodes with the **least** quorum queue replicas are selected (if total number of queues is less than 1k). This will nicely distribute the quorum queue replicas across the RabbitMQ cluster. Furthermore, nodes that are `running` are preferred being selected over nodes that are `down`. The same will apply to streams.
2. Support `(x-)queue-leader-locator` argument, policy, and rabbitmq.conf option with values `client-local` (stays the default), and `balanced`.

From now on, streams and quorum queues share the same replica selection + leader selection code and therefore behave the same.

Prior to this PR, streams used to support `queue-leader-locator` values `client-local`, `random`, and `least-leaders`.
After this PR,`random` and `least-leaders` are deprecated and get mapped to `balanced`.

From my point of view (+ @mkuratczyk came up with the idea), exposing value `random` to users via configuration does not make sense. When would a user want to set `random` over `least-leaders`? I cannot think of any use case.
Instead, what users want to set is either `client-local` or some form of balancing the leaders. How the leaders are balanced is an implementation detail that shouldn't be exposed to users. In this PR's `balanced` implementation, the leader gets placed on the nodes with the least leaders for the same queue type if the number of overall queues is low (currently set to 1k). If the number of overall queues is high, the leader will be placed randomly across the selected replicas. This ensures that queues get created fast. We see with `min-masters` that queue leader calculations becomes slow with many thousand queues. Randomness will do the job of balancing leaders across the cluster good enough when there are already many queues. There are some edge cases in this implementation: For example when there are more than 1k classic queues defined on a single node, we still use randomness to select replicas and leaders for quorum queues and streams. In the long term, the `balanced` implementation should count the leaders across **all** queue types, not just its own queue type. (In the future we could even use current resource usage or other metrics as decision how leaders are balanced across the cluster. Again, it is an implementation detail that the user shouldn't care about.)

I didn't touch the `queue-master-locator` code for classic queues. So, classic queues continue to support `client-local`, `random`, and `min-masters`. Once classic mirrored queues are gone in 4.0, we can consolidate classic queues with this PR's implementation, i.e. using only values `client-local` and `balanced`.

### Performance

Declaring 5k 3-replica quorum queues sequentially via AMQP on a 5-node K8s cluster:
| `queue-leader-locator`  | Time |
| ------------- | ------------- |
| prior to PR  |  4m13s |
| `client-local`  |  3m0s |
| `balanced`  |   3m30s |

### TODOs:
- [ ] Add docs for `queue-leader-locator` for both quorum queues and streams

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -1212,7 +1212,7 @@ end}.
 {mapping, "cluster_keepalive_interval", "rabbit.cluster_keepalive_interval",
     [{datatype, integer}]}.
 
-%% Queue master locator
+%% Queue master locator (classic queues)
 %%
 
 {mapping, "queue_master_locator", "rabbit.queue_master_locator",
@@ -1221,6 +1221,17 @@ end}.
 {translation, "rabbit.queue_master_locator",
 fun(Conf) ->
     list_to_binary(cuttlefish:conf_get("queue_master_locator", Conf))
+end}.
+
+%% Queue leader locator (quorum queues and streams)
+%%
+
+{mapping, "queue_leader_locator", "rabbit.queue_leader_locator",
+    [{datatype, string}]}.
+
+{translation, "rabbit.queue_leader_locator",
+fun(Conf) ->
+    list_to_binary(cuttlefish:conf_get("queue_leader_locator", Conf))
 end}.
 
 %%
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -979,16 +979,15 @@ check_overflow(Val, _Args) when is_binary(Val) ->
 check_overflow(_Val, _Args) ->
     {error, invalid_overflow}.
 
--define(KNOWN_LEADER_LOCATORS, [<<"client-local">>, <<"random">>, <<"least-leaders">>]).
 check_queue_leader_locator_arg({longstr, Val}, _Args) ->
-    case lists:member(Val, ?KNOWN_LEADER_LOCATORS) of
+    case lists:member(Val, rabbit_queue_location:queue_leader_locators()) of
         true  -> ok;
         false -> {error, invalid_queue_locator_arg}
     end;
 check_queue_leader_locator_arg({Type, _}, _Args) ->
     {error, {unacceptable_type, Type}};
 check_queue_leader_locator_arg(Val, _Args) when is_binary(Val) ->
-    case lists:member(Val, ?KNOWN_LEADER_LOCATORS) of
+    case lists:member(Val, rabbit_queue_location:queue_leader_locators()) of
         true  -> ok;
         false -> {error, invalid_queue_locator_arg}
     end;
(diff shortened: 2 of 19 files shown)


mkuratczyk:
Great stuff. I'm still running some tests but found something that definitely requires a fix: nodes under maintenance are considered for leadership (test case: run a 3-node cluster, drain one of the nodes, declare a bunch of queues with least-leaders or random - all nodes, including the drained one, have some queue leaders)

mkuratczyk:
What should the default be? I think `random` or `least-leaders` is a better default than the current, implicit, `client-local`. Also, a setting similar to `queue_master_locator` is probably necessary.

mkuratczyk:
This change degrades the performance of re-declaring queues. While it doesn't sound too important, users that rely on import-on-startup and don't take advantage of the fairly recent skip-if-unchanged setting, do re-import the same definitions on every node's startup. Without this change, I get 90s to import 10k QQs and 4s to re-import the same queues. With this change, I get 95s to import (totally fine I think, given the value this change brings) but 40s to re-import.

mkuratczyk:
No action required, just leaving here for posterity: in a 7-node cluster test, I had the following results (still with 10K QQs):
* default/implicit client-local: 97s
* `random`: 102s
* `least-leaders`: 120s

re-import took 52s in all cases but is expected to be improved

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -1223,6 +1223,17 @@ fun(Conf) ->
     list_to_binary(cuttlefish:conf_get("queue_master_locator", Conf))
 end}.
 
+%% Queue leader locator (quorum queues and streams)
+%%
+
+{mapping, "queue_leader_locator", "rabbit.queue_leader_locator",
```
michaelklishin, line 1229: I wish we could group this under `quorum_queues.` much like we have some new settings under `classic_queues.*` now but I see why you named it this way.

ansd:
Thank you for testing @mkuratczyk.

> I'm still running some tests but found something that definitely requires a fix: nodes under maintenance are considered for leadership (test case: run a 3-node cluster, drain one of the nodes, declare a bunch of queues with least-leaders or random - all nodes, including the drained one, have some queue leaders)

Very good spot! This is now fixed. Furthermore, nodes that are `running` are from now on preferred as replicas over nodes that are `down`.

> What should the default be? I think random or least-leaders is a better default than the current, implicit, client-local.

I think `client-local` should be the default since it means writes will be fast for the client that declares the queue and publishes subsequently (saving the additional hop in Figure 2 of https://blog.rabbitmq.com/posts/2020/06/quorum-queues-local-delivery/).

> Also, a setting similar to queue_master_locator is probably necessary.

rabbitmq.conf / Cuttlefish option is now added.

> This change degrades the performance of re-declaring queues. 

This is now made much faster by https://github.com/rabbitmq/rabbitmq-server/pull/4524 which is also cherry-picked into this PR.

PR #4468, opened 2022-04-08
Fix migrate leadership timeout

felixhuettner:
## Proposed Changes

The code of `migrate_leadership_to_existing_replica` previously assumed
that it can control the target node of a failover of mirrored queues.

While its sibling function `transfer_leadership` can do that (as it drops all
mirrors besides the one on the target node)
`migrate_leadership_to_existing_replica` tries to be less intrusive and
only tries to migrate the queue away for the current primary.

However it then used `wait_for_new_master` to wait for the queue to actually
transfer the primary to the specified target node. As the specified target
node might not ever become the primary of the queue (as this decission is
only made between the remaining mirrors) this can cause the migrate operation
to run into a timeout (10 sec per default).

As `migrate_leadership_to_existing_replica` is only used during
`transfer_leadership_of_classic_mirrored_queues` its only goal is to get the
primary away from the current node. Therefor we can just wait for the
queue to become active on some other node instead of expecting a specific
node to become the primary.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbit/src/rabbit_maintenance.erl
+++ b/deps/rabbit/src/rabbit_maintenance.erl
@@ -270,15 +270,15 @@ transfer_leadership_of_classic_mirrored_queues(TransferCandidates) ->
                           [rabbit_misc:rs(Name), readable_candidate_list(ExistingReplicaNodes)]),
          case random_primary_replica_transfer_candidate_node(TransferCandidates, ExistingReplicaNodes) of
              {ok, Pick} ->
-                 rabbit_log:debug("Will transfer leadership of local ~s to node ~s",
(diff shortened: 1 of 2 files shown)


michaelklishin:
@felixhuettner thank you but this does not really explain what problem this addresses or how to reproduce it. There are no new tests either.

michaelklishin:
Alright, reading the code makes it more obvious:

 * The logging incorrectly indicated that the newly elected leader was on the picked node, while it is not guaranteed
 * One function should await for a leader replica change, not assuming it happens to be the picked one

I don't have many ideas for how to reasonably reliably introduce this but it's a good catch.

michaelklishin (approved):

PR #4474, opened 2022-04-08
Attempt to make the OCI workflow behave reasonably for PRs

pjk25:
When repo secrets are not available to the actions run, the docker image is built, but not published


michaelklishin:
Makes sense, thank you. Let's try it.

Issue #4133, opened 2022-02-11
Monitor stream local member in stream queue


acogoluegnes:
The stream queue module does not monitor the local member of its consumers, so if it goes away for some reasons, the consumer will hang.

This can happen if the node a stream is on is restarted: all the stream member process are restarted, so any local state is wiped out. This means basic consumers wait for (Erlang) messages from dead processes, and they won't dispatch any messages to client, even if the stream see new messages after the election of a new leader.

acogoluegnes:
Note consumer cancel is not sent if the queue is deleted. This should be fixed as well.

acogoluegnes:
Follow-up issue for basic.cancel notification: #4192.

devkits:
Hi, 

Is there a way to determine release milestone(s) where this fix may end up in, e.g. 3.9.14, 3.10.X, etc? 

Thanks.

acogoluegnes:
The fix will go into 3.9.x and 3.10.x. We're currently working on it.

PR #4511, opened 2022-04-11
Handle machine upgrades incrementally in stream coordinator

### LINKED ISSUE #4510: Handle all version-to-version machine upgrades in stream coordinator
See https://github.com/rabbitmq/rabbitmq-server/discussions/4471 for background.

acogoluegnes:
RA applies machine upgrades from any version to any version,
e.g. 0 to 2. This commit "fills in the gaps" in the stream coordinator,
to make sure all 1-to-1 upgrades are applied, e.g. 0 to 1 and 1 to 2
in the previous example.

References #4133.

Fixes #4510

### DIFF
--- a/deps/rabbit/src/rabbit_stream_coordinator.erl
+++ b/deps/rabbit/src/rabbit_stream_coordinator.erl
@@ -515,44 +515,13 @@ apply(Meta, {nodeup, Node} = Cmd,
                   end, {Streams0, Effects0}, Streams0),
     return(Meta, State#?MODULE{monitors = Monitors,
                                streams = Streams}, ok, Effects);
-apply(Meta, {machine_version, From = 1, To = 2}, State = #?MODULE{streams = Streams0,
-                                                                  monitors = Monitors0}) ->
-    rabbit_log:info("Stream coordinator machine version changes from ~p to ~p, updating state.",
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -515,44 +515,13 @@ apply(Meta, {nodeup, Node} = Cmd,
...
-    return(Meta, State#?MODULE{streams = Streams1,
-                               monitors = Monitors2,
-                               listeners = undefined}, ok, Effects);
-apply(Meta, {machine_version, From, To}, State) ->
-    rabbit_log:info("Stream coordinator machine version changes from ~p to ~p, no state changes required.",
-                    [From, To]),
-    return(Meta, State, ok, []);
+apply(Meta, {machine_version, From, To}, State0) ->
+    rabbit_log:info("Stream coordinator machine version changes from ~p to ~p, applying incremental upgrade."),
+    {State1, Effects} = lists:foldl(fun(Version, {S0, Eff0}) ->
```
michaelklishin, line 520: Should we add a comment that would explain why we have to do it this way?

kjnilsson (approved):

PR #4524, opened 2022-04-11
Speed up reimport queue definitions

ansd:
The channel does not declare a queue that already exists.
Let's do the same when importing definitions.

On a 5-node cluster re-importing 40,000 classic non-mirrored queues with
`queue-master-locator` set to `min-masters` takes 11min32sec before this
change and 3sec after this change. So it's 230 times faster after this
change.

With even more queues, the speed-up factor will be much greater.

This solves the pain points of https://github.com/rabbitmq/rabbitmq-server/pull/3402 (column `reimport (sec)` and rows `old min-masters`).

### DIFF
--- a/deps/rabbit/src/rabbit_definitions.erl
+++ b/deps/rabbit/src/rabbit_definitions.erl
@@ -683,12 +683,17 @@ add_queue_int(_Queue, R = #resource{kind = queue,
     rabbit_log:warning("Skipping import of a queue whose name begins with 'amq.', "
                        "name: ~s, acting user: ~s", [Name, ActingUser]);
 add_queue_int(Queue, Name, ActingUser) ->
-    rabbit_amqqueue:declare(Name,
-                            maps:get(durable,                         Queue, undefined),
-                            maps:get(auto_delete,                     Queue, undefined),
-                            args(maps:get(arguments, Queue, undefined)),
-                            none,
-                            ActingUser).
+    case rabbit_amqqueue:lookup(Name) of
+        {ok, _} ->
+            ok;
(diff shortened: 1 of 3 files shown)


mkuratczyk:
Great stuff. I've checked the impact of this change on a 5-node cluster with 100k classic queues:
```+─────────────────+────────+───────────────+─────────────+───────────────+───────────────+────────────────+
| image           | nodes  | placement     | import (s)  | reimport (s)  | stop_app (s)  | start_app (s)  |
+─────────────────+────────+───────────────+─────────────+───────────────+───────────────+────────────────+
| master-otp-max  | 5      | client-local  | 99.635      | 26.193        | 6.933         | 6.859          |
| master-otp-max  | 5      | min-masters   | 1905.737    | 3746.395      | 18.110        | 115.367        |
| pr-4524-otp-max | 5      | client-local  | 96.322      | 6.926         | 6.834         | 8.271          |
| pr-4524-otp-max | 5      | min-masters   | 2038.232    | 13.972        | 17.497        | 112.992        |
+─────────────────+────────+───────────────+─────────────+───────────────+───────────────+────────────────+
```

Massive reduction in re-import time, pretty significant even for `client-local`.

Side note - it's nice to see that even without this PR, all the numbers are significantly lower [than they were in September](https://github.com/rabbitmq/rabbitmq-server/pull/3402). I guess it's a combination of many factors, likely including better HW underneath our Kubernetes cluster, but either way - it's much better.

michaelklishin:
Backported for `3.9.15`, now testing a backport for `3.8.29`.

PR #4558, opened 2022-04-13
CQ: Fix potential crash when collecting stats

essen:
The CQ is looking up itself when collecting CMQ stats.
There seems to be a potential race condition on init
where the initial stats update happens before the
queue can be looked up. To prevent the crash, we
ensure that when the queue is not found we don't
return stats for these items.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue_process.erl
+++ b/deps/rabbit/src/rabbit_amqqueue_process.erl
@@ -1163,25 +1163,37 @@ i(memory, _) ->
     M;
 i(slave_pids, #q{q = Q0}) ->
     Name = amqqueue:get_name(Q0),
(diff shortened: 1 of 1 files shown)


essen:
Ugh opened the PR with the wrong browser. Oh well!

mkuratczyk:
I'm afraid I still have queues down after upgrade. I can only see rabbit_reader and rabbit_channel crashes but there are also `not_found` errors from `queue.declare` and `queue.consume`.
[upgrade.log.gz](https://github.com/rabbitmq/rabbitmq-server/files/8480148/upgrade.log.gz)

lhoguin:
This does not fix the related issue. However we should still merge this because if for whatever reason a crash occurs here, it will lead to v1 CQs being unrecoverable as we could see in the original logs.

PR #4563, opened 2022-04-13
Don't delete durable queues from rabbit_queue table on node down

lhoguin:
This fixes an issue where durable queues are "lost" because the entry in rabbit_queue table gets removed as a result of a race condition when restarting nodes / upgrading. High load seems required to reproduce the issue.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -1058,7 +1058,13 @@ list() ->
     list_with_possible_retry(fun do_list/0).
 
 do_list() ->
-    mnesia:dirty_match_object(rabbit_queue, amqqueue:pattern_match_all()).
+    All = mnesia:dirty_match_object(rabbit_queue, amqqueue:pattern_match_all()),
+    NodesRunning = rabbit_nodes:all_running(),
+    lists:filter(fun (Q) ->
+                         Pid = amqqueue:get_pid(Q),
+                         St = amqqueue:get_state(Q),
+                         St =/= stopped orelse lists:member(node(Pid), NodesRunning)
+                 end, All).
 
 -spec count() -> non_neg_integer().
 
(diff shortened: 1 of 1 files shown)


mkuratczyk:
* I've restarted the cluster multiple times - all queues recovered nicely every time
* I've restarted the nodes a few times without perf-test running - the queues re-apper in the list_queues as soon as the node is back (so it wasn't just perf-test redeclaring them - RabbitMQ doesn't forget they should exist)
* if I have some messages in these queues, and restart the nodes (with perf-test off), the queues re-appear and the message count is as expected (no obvious data loss)

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1994,6 +2007,7 @@ queues_to_delete_when_node_down(NodeDown) ->
             Q <- mnesia:table(rabbit_queue),
                 amqqueue:qnode(Q) == NodeDown andalso
                 not rabbit_mnesia:is_process_alive(amqqueue:get_pid(Q)) andalso
+                not amqqueue:is_durable(Q) andalso
```
lhoguin, line 2010: We should probably limit this check ONLY to classic queues.

lhoguin: Pushed a commit trying to do just that. We'll see if that goes well.

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1292,13 +1302,16 @@ list_down(VHostPath) ->
...
-            PresentS = sets:from_list([amqqueue:get_name(Q) || Q <- Present]),
-            sets:to_list(sets:filter(fun (Q) ->
-                                             N = amqqueue:get_name(Q),
-                                             not sets:is_element(N, PresentS)
-                                     end, sets:from_list(Durable)))
+            lists:filter(fun (Q) ->
+                                 N = amqqueue:get_name(Q),
+                                 Pid = amqqueue:get_pid(Q),
+                                 St = amqqueue:get_state(Q),
+                                 (St =:= stopped andalso not lists:member(node(Pid), rabbit_nodes:all_running()))
```
lhoguin, line 1311: Should not call `all_running` per queue, but instead once before the filter. Same in the other three places.

lhoguin:
I'm marking this as ready to review since the results seem positive. I still have reservations about it but they pertain more to the way the queue state is handled (`rabbit_queue` and `rabbit_durable_queue`, along with different queue types with entirely different behaviors put together in the same tables). So I'll leave it up to reviewers to decide if this should be in 3.10 or not.

lhoguin:
@dcorbacho said it looks good, I think we can merge this for 3.11.

PR #4572, opened 2022-04-14
Validate queue and consumer arguments

ansd:
Follow up of https://github.com/rabbitmq/rabbitmq-server/pull/4404.

Throw a `PRECONDITION_FAILED` error when a queue is created with an
invalid argument or when a consumer subscribes with an invalid argument.

There can be RabbitMQ plugins and extensions out there defining arbitrary
queue arguments. Therefore we only check that a queue is not declared
with arguments that we know are supported only by other queue types.

The benefit of this change is that queues are not mistakenly declared with the wrong arguments.

This change will reduce a lot of confusion when it comes to what arguments are supported by what queue type.

Example 1:
Before this PR, a stream could be created with `x-dead-letter-exchange` queue argument set. Users wonder why the stream does not dead letter any messages. After this PR, stream creation fails early with a meaningful error message because that argument is only supported for classic and quorum queues.

Example 2:
Before this PR, a stream could be created with `x-max-length` queue argument set. Users wonder why the stream contains more messages than configured via `x-max-length`. After this PR, stream creation fails early with a meaningful error message because that argument is only supported for classic and quorum queues. Streams support `x-max-length-bytes` and `x-max-age`.

Example 3:
Before this PR, a quorum queue could be created with `x-queue-master-locator` queue argument set. Users wonder why it doesn't have any effect. After this PR, queue creation will fail with a meaningful error message.

All changes are backwards compatible, except invalid **consumer** arguments which I expect to be rare.
For example, when a consumer had consumer argument `x-priority` set to connect to a stream via AMQP, this will start failing after this PR.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -779,7 +779,7 @@ check_arguments_type_and_value(QueueName, Args, Validators) ->
                           ok             -> ok;
                           {error, Error} -> rabbit_misc:protocol_error(
                                               precondition_failed,
-                                              "invalid arg '~s' for queue ~s: ~255p",
+                                              "invalid arg '~s' for ~s: ~255p",
                                               [Key, rabbit_misc:rs(QueueName),
                                                Error])
                       end
@@ -795,7 +795,7 @@ check_arguments_key(QueueName, QueueType, Args, InvalidArgs) ->
                               true ->
                                   rabbit_misc:protocol_error(
                                     precondition_failed,
-                                    "invalid arg '~s' for queue ~s of type ~s",
+                                    "invalid arg '~s' for ~s of queue type ~s",
                                     [ArgKey, rabbit_misc:rs(QueueName), QueueType])
                           end
                   end, Args).
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -749,19 +749,26 @@ with_exclusive_access_or_die(Name, ReaderPid, F) ->
     with_or_die(Name,
                 fun (Q) -> check_exclusive_access(Q, ReaderPid), F(Q) end).
 
-assert_args_equivalence(Q, RequiredArgs) ->
+assert_args_equivalence(Q, NewArgs) ->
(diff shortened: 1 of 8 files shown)


michaelklishin (commented):
My first reading of this is that such a validation would make it impossible for plugins to
rely on custom x-args, or provide custom queue types.

I doubt those are common but we can't verify that.

ansd:
Plugins using custom x-args will continue to work after this PR because custom x-args are never validated or invalidated (assuming the custom x-arg is not part of the list of supported x-args of any existing queue type).

This is because this PR uses some form of "weak" validation:
Let's say we have queue type `A` with supported x-arg `1` and queue type `B` with supported x-arg `2`. Whenever a user includes x-arg `1` during declaration of queue type `B`, an error will be thrown. However, when a user includes any other x-args (e.g. x-arg `2` and a custom x-arg `3`) when declaring queue type `B`, validation will succeed.

I'm not sure whether it's possible to extend RabbitMQ with custom queue types at the moment. I think there are some hard-coded assumptions that only `classic`, `quorum` and `stream` queues exist, but maybe once we move these types to the registry (as indicated in [this](https://github.com/rabbitmq/rabbitmq-server/blob/37a34486728565e3945fe1b77f29269d2f20bee3/deps/rabbit/src/rabbit_queue_type.erl#L213-L222) TODO), this will become possible, yes. In this case, the implementation of this PR will continue to work, and do the right thing.

We could also enable the new validation of this PR by default, but could add a config option allowing to disable this new validation (just in case I overlooked a breaking change scenario).

Alternatively, if we think this validation will cause more trouble than people find it helpful (my opinion is it will overall cause less confusion) we can also close this PR and just delete the list of supported x-args per queue type as originally done in https://github.com/rabbitmq/rabbitmq-server/pull/4404 since they are currently not used anywhere.

michaelklishin:
Thanks for the explanation, @ansd. "Validate known and accept the rest" sounds like a good balance. No objections from me.

michaelklishin:
This needs conflicts to be resolved, otherwise LGTM.

michaelklishin (approved):

michaelklishin (approved):

michaelklishin:
@mergifyio backport v3.10.x

PR #4599, opened 2022-04-19
Pin Osiris to 1.2.5

michaelklishin:


michaelklishin:
Will backport manually as there will be conflicts.

michaelklishin:
Something about this PR trips up a few seemingly unrelated suites in Bazel. Will experiment more and investigate.

PR #4604, opened 2022-04-20
OAuth 2: support for scope aliases

### LINKED ISSUE #4588: OAuth 2: support for scope aliases (mappings)
In some environments, producing scopes that follow the convention used by our OAuth 2 authN/authZ backend is effectively impossible for non-technical reasons.

@MarcialRosales has suggested a workaround that would allow for a mapping of scopes to a set of
RabbitMQ permission scopes.

Here is an example:

``` erl
{rabbitmq_auth_backend_oauth2, [
    {resource_server_id, <<"rabbitmq-122">>},
    {extra_scopes_source, <<"claims">>},
    {scope_aliases, #{
      <<"megacorp.roles.122">>      => [
        <<"rabbitmq.read:vhost-122/*">>,
        <<"rabbitmq.write:vhost-122/*">>,
        <<"rabbitmq.configure:vhost-122/*">>
      ],
      <<"megacorp.roles.188">>     => [
        <<"rabbitmq.read:vhost-188/*">>,
        <<"rabbitmq.configure:vhost-188/*">>
       ],
      <<"megacorp.roles.199">>      => [<<"rabbitmq.read:*/*">>]
    }}
    {key_config, [
```

``` json
{

  "claims": "megacorp.roles.122",
  "aud": [
    "rabbitmq-122"
  ]
}
```

The `rabbitmq_auth_backend_oauth2.scope_mappings` key, if defined, can be used to look up
a set of RabbitMQ permissions that would then be parsed just like "regular" JWT token scopes
are.

michaelklishin:
## Proposed Changes

This introduces scope alias support to the OAuth 2 plugin.
The goal is to allow users who have no control over what
scopes their JWT tokens include, map "role names" retrieved
from their identity provider to a set of scopes
that follow the convention used by the plugin.
See #4588 for background.

Per discussion with @MarcialRosales, we try to fetch
aliases from two sources, based on feedback from two different
users who seemingly rely on the same family of identity
provider products:

 * Use the JWT scope field value first
 * Use extra_scopes_source app env setting second

Just like with the existing extra scopes/complex claim
support originally contributed for Keycloak/identityProvider,
we merge all these scopes obtained from "alternative sources"
with the value of the JWT scopes field. This implicitly
assumes that the result makes sense semantically and
there will not be conflicting scopes. That's on the user to
make sure of.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

Closes #4588

### DIFF
--- a/deps/rabbitmq_auth_backend_oauth2/src/rabbit_auth_backend_oauth2.erl
+++ b/deps/rabbitmq_auth_backend_oauth2/src/rabbit_auth_backend_oauth2.erl
@@ -37,7 +37,7 @@
 %% a term used by the IdentityServer community
 -define(COMPLEX_CLAIM_APP_ENV_KEY, extra_scopes_source).
 %% scope aliases map "role names" to a set of scopes
--define(SCOPE_ALIASES_APP_ENV_KEY, scope_aliases).
+-define(SCOPE_MAPPINGS_APP_ENV_KEY, scope_aliases).
 
 %%
 %% Key JWT fields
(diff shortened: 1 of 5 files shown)


michaelklishin:
@MarcialRosales I am fairly happy with the test coverage of the new feature, so let's merge this and address any feedback that comes up from end-to-end testing against identity providers.

PR #4623, opened 2022-04-25
Attempt to fix compilation errors from mix when using erlang 25 rc

pjk25:
In the OCI workflow with otp 25.0-rc3, there is the following error:

```
Error: 5.778 [error] beam/beam_load.c(86): Error loading module 'Elixir.Code.Formatter':
[73](https://github.com/rabbitmq/rabbitmq-server/runs/6158043678?check_suite_focus=true#step:7:73)
Warning:   lambda already defined for label 749. To fix this, please recompile this module with an OTP 25 compiler.
[74](https://github.com/rabbitmq/rabbitmq-server/runs/6158043678?check_suite_focus=true#step:7:74)
    (elixir 1.13.4) lib/code/normalizer.ex:22: Code.Normalizer.normalize/2
```

The changes are an attempt to avoid using a precompiled elixir


pjk25:
After digging around a bit, this seems to be a consequence of the `erlef/setup-beam` action which provides elixir compiled with Erlang 22. I'll think a bit on what a good alternative would be.

pjk25:
This is addressed with #4884

PR #4647, opened 2022-04-26
v3.10.x: pin Osiris to 1.2.6

michaelklishin:


pjk25:
The `BUILD.osiris` file will need to be copied over from the `v3.9.x` branch as well.

PR #4671, opened 2022-04-28
Add set_user_limits to rabbitmqctl(8)

michaelklishin:
Closes #1398.


johanrhodin:
I think the reference is wrong #1398 has been closed for years.

johanrhodin:
Ah, it should be https://github.com/rabbitmq/rabbitmq-website/issues/1398

PR #4703, opened 2022-05-03
Use Khepri to store vhosts, users and runtime parameters

dumbbell:
TBW; this pull request is mainly to trigger testing in GitHub Actions. A complete description will be written when the work is closer to completion.


dumbbell:
This phase-1-only pull request was superseded by the full integration in #7206.

PR #4735, opened 2022-05-05
No longer file:allocate CQv2 index files

lhoguin:
As discussed on Slack following https://github.com/rabbitmq/rabbitmq-server/pull/4721.

This is both an alternative fix for https://github.com/rabbitmq/rabbitmq-server/pull/4721 and an improvement to the way v2 index segment files are handled: we no longer pre-allocate the files, so there is no platform differences anymore (there was for at least Windows and FreeBSD+ZFS), and the files do not unnecessarily take space on disk.


lhoguin:
I have fixed the property suite, it was missing a case clause during recovery.

lhoguin:
This is both an alternative fix for https://github.com/rabbitmq/rabbitmq-server/pull/4721 and an improvement to the way v2 index segment files are handled: we no longer pre-allocate the files, so there is no platform differences anymore (there was for at least Windows and FreeBSD+ZFS), and the files do not unnecessarily take space on disk.

Issue #755, opened 2016-04-18
Invalid transaction return value when a vhost is removed while a policy is updated


dumbbell:
In the `update_polices()` function in `rabbit_policy.erl`:

```erlang
    {Xs, Qs} = rabbit_misc:execute_mnesia_transaction(
                 fun() ->
                         [mnesia:lock({table, T}, write) || T <- Tabs], %% [1]
                         case catch list(VHost) of
                             {error, {no_such_vhost, _}} ->
                                 ok; %% [2]
                               % ^^
```

The transaction function returns `ok` if it couldn't query the list of policies for the given vhost. However, the calling code expects a tuple.

I discovered it while working on #744, though I didn't try to reproduce it yet.

dumbbell:
This will be fixed in pull request #745.

PR #4759, opened 2022-05-07
Ensure maintenance mode state table exists after node [re]boot

### LINKED ISSUE #4755: Maintenance mode: restore maintenance state table upon node boot
Per discussion in https://github.com/rabbitmq/discussions/issues/183#issuecomment-1119849840.

This would make maintenance mode tracking slightly more predictable and would not require any extra
upgrade machinery.

The table should only be restored if the maintenance mode feature flag is enabled.

michaelklishin:
## Proposed Changes

Some users were surprised by the current approach to
state table setup, so let's set make sure we ensure the table exists
on every boot (if the feature flag is enabled).

Per discussion in https://github.com/rabbitmq/discussions/issues/183#issuecomment-1119849840.
Closes #4755.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #755)
- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_maintenance.erl
+++ b/deps/rabbit/src/rabbit_maintenance.erl
@@ -29,7 +29,8 @@
     transfer_leadership_of_quorum_queues/1,
     transfer_leadership_of_classic_mirrored_queues/1,
     status_table_name/0,
-    status_table_definition/0
+    status_table_definition/0,
+    boot/0
 ]).
 
 -define(TABLE, rabbit_node_maintenance_states).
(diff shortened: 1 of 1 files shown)


michaelklishin:
@TheGOro does this look like an improvement to you?

TheGOro:
For the first glance it seems like that, but would it be okay to have different type of copies (ram copy on the first node that starts the cluster formation, then disc copies on the rest) of that table among the cluster members?
I'll give a spin to it on Monday.

michaelklishin:
Changing table type would be too risky for mixed version clusters and upgrades. I'd leave the type as is, the table is lightly used that
it can be anything.

Also, in future versions where this might use Khepri, RAM tables won't be an option at all.

michaelklishin:
I will look into the failures.

michaelklishin:
So the failures are legitimate, the current approach is not always safe. I'll look into that.

michaelklishin:
Not forcing a local copy helped. Writes to a created table without a forced local copy succeed, so do reads, so this accomplishes what the previous design did with more certainly that the table will be there after node boot.

TheGOro:
> Not forcing a local copy helped.

My understanding is that this was the very issue that started the original discussion. However, I have no idea how the reporter managed to reach and observe such scenario (having different type of copies of the table among the cluster members), as it doesn't seem to be possible with any version of the upstream code...

I did some trials with your change, by simply just wrapping the compiled beam module into an .ez archive, to be loaded as a plugin. With that in place, I was not able to reproduce the issue where various kind of mnesia resets could eradicate the reminiscence of the maintenance status table from the whole cluster. Therefore, I think this will be fine.
Thanks for your work!

michaelklishin:
@TheGOro thank you for reporting back. For tables that store semantically persistent data, forcing a local copy would be necessary. However here, we only need a couple of things:

 * The table must exist after the node's finished booting
 * The table must allow for status writes

and the current state accomplishes that with very little effort.

PR #4762, opened 2022-05-09
Add test-node-logs and remote-test-node-logs rules for bazel

pjk25:
For convenience when opening up the directory containing the rabbitmq node logs produced by integration tests

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -203,3 +203,56 @@ EOF
 """,
     executable = True,
 )
+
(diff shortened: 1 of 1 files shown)


michaelklishin:
Tried it with

``` shell
bazel test //deps/rabbit:maintenance_mode_SUITE
bazel run test-node-data //deps/rabbit:maintenance_mode_SUITE
```

works like a charm! 🙏

PR #4774, opened 2022-05-10
Check queue and exchange existence with ets:member/2

ansd:
This reduces memory usage since the object does not get copied from ETS and improves code readability.


michaelklishin (approved):

michaelklishin:
@ansd good idea. Just out of curiosity, do you have any metrics that demonstrate the effect of this change? Also, there must be some cases where a "consistent read" may be necessary, so Mnesia-based predicates may still be necessary beyond keeping the API stable.

ansd:
I haven't gathered metrics since the changes in this PR do not touch the hot path. So, the memory gains are marginal.
The main benefit is more the slightly improved code readability.

PR #4844, opened 2022-05-19
CQv2: file:delete/1 -> prim_file:delete/1

essen:
This causes issues with many queues that happen to delete their files at the exact same time.

My bad. And unfortunately the raw option is 24+.

### DIFF
--- a/deps/rabbit/src/rabbit_classic_queue_index_v2.erl
+++ b/deps/rabbit/src/rabbit_classic_queue_index_v2.erl
@@ -331,7 +331,7 @@ recover_segments(State0, ContainsCheckFun, StoreState0, CountersRef, [Segment|Ta
                 keep ->
                     StoreState1;
                 delete ->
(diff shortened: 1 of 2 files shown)


lhoguin:
I opened the PR again with the wrong browser. Discipline where are thou.

michaelklishin (approved):

michaelklishin:
We decided to hold off on backporting this as more data is collected on how it affects different CQ configurations.

mkuratczyk:
Alright, the above is user error. There was 500 queues, each with 1 publisher and 1 consumer. The publisher published 10 msg/s and *the consumer ACKed every 1000 messages*. The much more reasonable value here should be 10. With 1000, with 500 queues, in total there was 400-500k unacked messages and then all of them were acked around the same time. Removing so many messages from the message store at the same time is what caused these spikes.

mkuratczyk:
@Mergifyio backport v3.10.x

PR #4846, opened 2022-05-19
Bump Aten to 0.5.8

michaelklishin:


michaelklishin:
There is no explicit dependency on `aten` in our Makefiles. It is brought in as a dependency of `ra`. Bazel is a different story, though.

PR #4854, opened 2022-05-19
Elie and Seth bug fix

ebenzaqu:
## Proposed Changes

Sometimes the queue is purposely paused so why would we want the whole program to stop working  
What the method at line 468 is doing is checking if the process pid is alive. If that false, which means it's dead, it will wait for a new leader to be elected, and then rerun the method to check if the new leader was in fact elected. If that's not the case it will rerun again and again until the retries run out.
If the queue was stopped purposefully by the supervisor, the pid is still alive, it's just that the queue is stopped, so when it enters into that last method, the assertion will return true and it will move onto the next method which just continues running the code assuming that a new leader was elected, which it wasn't. Since there is no leader there will most likely be another error thrown in some subsequent code.
We saw in some other code that the queue has many different states, and one of them includes “stopped”. If we are able to check that the queue has been stopped, we can try to do a rerun of the leader election by waiting until another slave that is mirrored picks up leadership. As well as implementing reruns which will check if the queue was stopped on purpose or not.

- [ ] Build system and/or CI

## Checklist

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -588,7 +588,7 @@ with(#resource{} = Name, F, E, RetriesLeft) ->
             %% Something bad happened to that queue, we are bailing out
             %% on processing current request.
             E({absent, Q, timeout});
-        {ok, Q} when ?amqqueue_state_is(Q, stopped) andalso RetriesLeft =:= 0 ->
+        {ok, Q = #amqqueue{state = stopped}} andalso RetriesLeft =:= 0 ->
             %% The queue was stopped and not migrated
             E({absent, Q, stopped});
         %% The queue process has crashed with unknown error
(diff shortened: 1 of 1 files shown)


deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -588,7 +588,7 @@ with(#resource{} = Name, F, E, RetriesLeft) ->
             %% Something bad happened to that queue, we are bailing out
             %% on processing current request.
             E({absent, Q, timeout});
-        {ok, Q} when ?amqqueue_state_is(Q, stopped) andalso RetriesLeft =:= 0 ->
+        {ok, Q = #amqqueue{state = stopped}} andalso RetriesLeft =:= 0 ->
```
michaelklishin, line 591: The helper macros are used for a reason: as records evolve, we cannot be certain that fields won't be renamed or even removed. Please avoid using record matching directly, it has significant negative repercussions for upgrades.

michaelklishin (changes requested):

michaelklishin:
I have to close this. Thank you for taking the time to contribute!

Unfortunately we cannot accept this PR as it deviates from our approach for "upgrade-friendly" record access.

PR #4904, opened 2022-05-27
Skip default definitions when definitions are imported using management.load_definitions

mkuratczyk:
There was a missing condition reported on the mailing list.
If the old-school `management.load_definitions` was used in the config file,
the guest user was still created.

Test case:
1. Start RabbitMQ with any definitions loaded using `management.load_definitions`
2. `rabbitmqctl list_users`
3. Without this PR, `guest` user exists, but it shouldn't


michaelklishin:
Good catch. I'd argue that setting should have been replaced now that a core alternative has been around for well over a year but no harm in merging this

PR #4833, opened 2022-05-17
improve credit handling and header conversion for AMQP 1.0 plugin

luos:
Hi, 

Sorry this became a so big but I was trying to debug something and somehow everything fall into one change. 
I can cut it up if needed.

Let me know what you think about these changes.

The AMQP 1.0 plugin grants publishers 65000 credit to the senders by default.
This PR makes the credit given configurable.

The AMQP 1.0 plugin grants credit when the message is received, not when it is
placed into the queue. This means that an AMQP 1.0 sender who does not care
about settlement can overload the receiver process leading to high memory
usage.

The AMQP 1.0 plugin creates a classic queue by default. This is annoying
because the client can not ensure that the target exists (it needs to use
/amq/queue/Q target instead of the name of the queue.
Now the plugin can be configured to use Quorum Queues.

Sending an unsettled message on a link set as 'settled' is now a protocol
error. Before this change it was leading to crashes of the AMQP process with a
crash of the link.

Implements optional conversion of AMQP 1.0 headers to 0.9.1 headers for some
headers types (utf8, uint, bool). Implements #2591

Implements optional conversion of AMQP 0.9.1 headers to AMQP 1.0 App Properties.

Also fixes a bug in the amqp10_client where the client can run out of credit
and get stuck because of bad accounting.

The tests are prepared to run with changes when https://github.com/rabbitmq/rabbitmq-server/issues/4832 is fixed, however they pass without it as well. 

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [x] Bug fix (non-breaking change which fixes issue #NNNN)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
If you're unsure about any of them, don't hesitate to ask on the mailing list.
We're here to help!
This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
[...]

PR #4967, opened 2022-06-03
Implement header conversion between AMQP 1.0 and 0.9.1

luos:
## Proposed Changes

Implements optional conversion of AMQP 1.0 headers to 0.9.1 headers for some
headers types (utf8, uint, bool). Implements rabbitmq#2591

Implements optional conversion of AMQP 0.9.1 headers to AMQP 1.0 App Properties.

I changed the conversion to use `rabbit_msg_record` though moving all this messaging conversion I am not sure it's possible. 

This is picked from #4833. 

Let me know what you think.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbitmq_amqp1_0/src/rabbit_amqp1_0_message.erl
+++ b/deps/rabbitmq_amqp1_0/src/rabbit_amqp1_0_message.erl
@@ -62,11 +62,11 @@ assemble(app_properties, {R, P = #'P_basic'{headers = Headers}, C},
          {#'v1_0.application_properties'{}, Rest}, Uneaten) ->
     AppPropsBin = chunk(Rest, Uneaten),
     Amqp091Headers = case ?CONVERT_APP_PROPS_TO_AMQP091_HEADERS of 
-        true -> 
-            amqp10_app_props_to_amqp091_headers(Headers, AppPropsBin);
-        _ -> 
-            Headers
-    end,
+                        true -> 
+                            amqp10_app_props_to_amqp091_headers(Headers, AppPropsBin);
+                        _ -> 
+                            Headers
+                     end,
     AppPropsAdded = set_header(
         ?APP_PROPERTIES_HEADER,
         AppPropsBin, Amqp091Headers),
@@ -311,15 +311,15 @@ amqp10_app_props_to_amqp091_headers(CurrentHeaders, AppPropsBin) ->
                 Headers -> Headers
             end,
             lists:foldl(fun(Prop, Acc) -> 
-                case Prop of 
-                    {{utf8, Key}, {ValueType, Value}} -> 
-                        case type10_to_type091(Key, ValueType, Value) of 
-                            undefined -> Acc;
-                            Typed -> [Typed |Acc]
-                        end;
-                    _ -> Acc
-                end
-            end, Hs, AppProps);
+                            case Prop of 
+                                {{utf8, Key}, {ValueType, Value}} -> 
+                                    case type10_to_type091(Key, ValueType, Value) of 
+                                        undefined -> Acc;
+                                        Typed -> [Typed |Acc]
+                                    end;
+                                _ -> Acc
+                            end
+                         end, Hs, AppProps);
         _ -> CurrentHeaders
     end.
 type10_to_type091(Key, Type, Value) -> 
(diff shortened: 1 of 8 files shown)


deps/rabbitmq_amqp1_0/src/rabbit_amqp1_0_message.erl
```diff
@@ -282,3 +302,52 @@ map_add(_T, _Key, _Type, undefined, Acc) ->
...
+                            Typed -> [Typed |Acc]
+                        end;
+                    _ -> Acc
+                end
+            end, Hs, AppProps);
+        _ -> CurrentHeaders
+    end.
+type10_to_type091(Key, Type, Value) -> 
+    try
+        rabbit_msg_record:to_091(Key, {Type, Value})
```
kjnilsson, line 327: Do we know which types will cause a function_clause?  perhaps we should just have a fall back clause in the rabbit_msg_record:to_091 function?

kjnilsson (changes requested):
A few formatting issues and a couple of questions.

I think we don't need this to be configurable necessarily, we feel it is better to always perform this conversion. Perhaps we delay releasing this until 3.11

deps/rabbitmq_amqp1_0/src/rabbit_amqp1_0_message.erl
luos: I am not sure which type causes an error, I just defensively added the try/catch as I did not verify that all types are handled according to the specs. I think it would be better to return `undefined`.

luos:
I think conversion can be quite resource intensive and unnecessary, ie when only AMQP 1.0 clients are communicating. Not sure if it's worth adding it always.

deps/rabbitmq_amqp1_0/src/rabbit_amqp1_0_message.erl
luos: In this case though the function would lose it's nice property that the return value is a valid header value which can be immediately encoded as binary. Now we would need to filter the results of all these calls. 

If you think all conversions cases are handled  I can just remove the try catch, but I like to be on the safe side. 😄

michaelklishin:
We are weighing the cost (per message) of such conversion vs. the cost of an ETS read to test if the conversion is enabled. I won't be able to tell what carries greater overhead.

We can [collect some data](https://blog.rabbitmq.com/posts/2022/05/flame-graphs/) and decide.
Targeting the case with only AMQP 1.0 clients communicating makes sense to me.

luos:
Hi,

Yes, so I was meaning that the conversion is wasteful for AMQP 1.0 only clients as the result will be thrown away, as it always uses the AppPropsBin for those from the AMQP 0.9.1 header. 

I tried to run perf, but sadly all functions came out as `unknown`, probably because I did not compile Erlang on the machine but it was from a package. I will try again as soon as I can. 

Are you OK with running perf yourself?

Let me know if there are other changes needed for this PR.

michaelklishin:
Understanding the effects of this on throughput is the primary thing left.

luos:
Hi,

I've made some performance tests. Seems like the cost of converting the headers overshadows everything. For this reason, it may be a good idea to leave this as a flag if someone does not need the conversion it can drastically slow down the incoming messages.

The test is for converting AMQP 1.0 Application properties to AMQP 0.9.1 headers. The test code and headers you can see here. The only changed between runs were the RabbitMQ config / code part.

https://github.com/luos/amqp10client/blob/b973c349ba75c4a55bc027bac44e95126db338d7/src/amclient_client.erl#L206-L215


  | Message Count | Msgs / sec | Total time
-- | -- | -- | --
Disabled (no config) | 300000 | 39751 | 7546.866
Disabled config | 300000 | 42571 | 7.047
Disabled config | 300000 | 37397 | 8.022
Enabled 1 | 300000 | 17140 | 17.503015
Enabled 2 | 300000 | 17227 | 17.41474
Compiled out (disabled) | 300000 | 34672 | 8.652493
Compiled out (disabled) | 300000 | 35971 | 8.340113
Compiled out (disabled) | 300000 | 25420 | 11.8019
Compiled in (enabed)  (whithout ets lookup) | 300000 | 18868 | 15.900334
Compiled in (enabled) (whithout ets lookup) | 300000 | 16291 | 18.415505
Nothing to convert enabled | 300000 | 47693 | 6.290289
Nothing to convert enabled | 300000 | 42762 | 7.015538
Nothing to convert (compiled out) | 300000 | 48823 | 6.144593

michaelklishin:
There are two "Nothing to convert enabled" rows, was one of them mean to say "disabled"?

luos:
For all cases, it's the same test running multiple times. Sorry for the confusion.

luos:
Hi! Is there anything I could add to this to be mergeable? Thanks

kjnilsson:
I think this is ok. The formatting is still a bit off but we are likely to completely revisit the entire 1.0 plugin at some point so it is fine for now.

kjnilsson (approved):

PR #4973, opened 2022-06-05
Add authz, authn log entries AMQP 1.0

per-ivan:
## Problem

We observed that server logs does not contain entries about authentication/authorization when AMQP 1.0 plugin is used.


michaelklishin:
Thank you. This does not match the format used by other connection lifecycle events. I will take care of it.

PR #4982, opened 2022-06-06
Log two more connection lifecycle events for AMQP 1.0 clients

michaelklishin:
Same idea as #4973 by @per-ivan but uses a slightly more standard log message format (standard for other protocols RabbitMQ supports, that is).


michaelklishin:
followed-up to with https://github.com/rabbitmq/rabbitmq-server/commit/0b3946e5af5277890aabb16a5ea534684a7e8347.

Those logger exceptions did not affect any of the tests so took a manual inspection to notice.

PR #4990, opened 2022-06-07
Delay generating a guid for basic_messages until it is needed

kjnilsson:
`rabbit_guid:gen()` has a noticeable cost and only classic queues use/require it.


michaelklishin:
@lhoguin it a merge/rebase artifact.

kjnilsson:
Honestly not sure where that commit came from
I'll see if I can fix

kjnilsson:
OK I think it's ok now.

michaelklishin:
@mergifyio backport v3.10.x v3.9.x

PR #5013, opened 2022-06-09
Use xref2 from rules_erlang 3

pjk25:
This applies the same analysis as the recently updated erlang.mk approach


pjk25:
@lhoguin FYI

michaelklishin (approved):
This makes xref pass on Erlang 25 🥳

PR #5020, opened 2022-06-10
Limit stream max segment size to 3 GB

acogoluegnes:
Values too large can overflow the stream position field
in the index (32 bit int).


ansd (approved):
I tested via latest stream-perf-test and via policy and queue arg in the management UI. Works well. Thanks!

PR #5028, opened 2022-06-13
Use a more unique exchange names in federation exchange suite

kjnilsson:
This is to avoid a case where multiple policies are matches for a given
exchange and the federation exchage suite causing indeterminacy in the
policy being applied to certain exchanges. This caused failures due to
internal changes in ETS set ordering between OTP 24 and 25.


pjk25:
@michaelklishin was there any reason this wasn't merged previously?

michaelklishin:
It does not fully help IIRC. I was going to finish it. Still hope to do :)
________________________________

kjnilsson:
@michaelklishin can we close this?

michaelklishin:
Superseded by a more drastic exchange federation test suite refactoring

PR #5037, opened 2022-06-14
Change normal connection close log level from warning to debug

ansd:
This line gets logged when the client closes the connection to the
stream port before it authenticates successfully.

Some external load balancers for example connect to the stream port to
do health checks without sending any stream protocol frame.

This commits prevents the RabbitMQ log from being polluted.


Gsantomaggio (approved):
We have already received some questions about this `warning`.
So It makes sense to put it in `debug`

PR #5054, opened 2022-06-15
Use socket option {packet, 4} for stream protocol

ansd:
The stream protocol frame (packet) starts with a 4 byte
unsigned integer in big-endian byte order.

Therefore we can make use of socket option `{packet, 4}`.
The runtime will generate the header for send operations
and strip off the header for receive operations.
See `inet:setopts/2` docs for more info.

Before this commit, rabbitmq-server received partial
info messages and had to assemble a stream packet on its own.

After this commit, rabbitmq-server receives one info message
per stream frame.

Advantages of this commit:
1. Less copying of binary data for large frames.
Large frames are for example created with large `--size` or
`--batch-size` flags in stream-perf-test.
Hence, this fixes the following TODO:
https://github.com/rabbitmq/rabbitmq-server/blob/744100ed2fcc0b4ce137a9fd9df4ba3ba5839997/deps/rabbitmq_stream_common/src/rabbit_stream_core.erl#L558-L561
2. A good amount of code gets deleted.

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
@@ -47,8 +47,7 @@
          log :: undefined | osiris_log:state(),
          last_listener_offset = undefined :: undefined | osiris:offset()}).
 -record(stream_connection_state,
-        {data :: rabbit_stream_core:state(), blocked :: boolean(),
-         consumers :: #{subscription_id() => #consumer{}}}).
+        {blocked :: boolean(), consumers :: #{subscription_id() => #consumer{}}}).
 -record(stream_connection,
         {name :: binary(),
          %% server host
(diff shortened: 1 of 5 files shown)


ansd:
Performance tests using `stream-perf-test` with different number for `--size` show that message throughput drops slightly in some cases with this PR.

Running rabbitmq-server with `+S 2` and stream-perf-test for 30 seconds.

Examples:

1.
```
java -jar target/stream-perf-test.jar  -x 1 -y 0 -s 50000 -bs 10 -smssb 3gb -mlb 20gb
```

Before:
```
Summary: published 14825 msg/s, confirmed 14561 msg/s, consumed 0 msg/s, latency 95th 0 ms, chunk size 0
```
After:
```
Summary: published 14391 msg/s, confirmed 14178 msg/s, consumed 0 msg/s, latency 95th 0 ms, chunk size 0
```

2.
```
java -jar target/stream-perf-test.jar -x 1 -y 0 -s 1000 -smssb 3gb -mlb 20gb
```

Before:
```
Summary: published 377109 msg/s, confirmed 376781 msg/s, consumed 0 msg/s, latency 95th 0 ms, chunk size 0
```
After:
```
Summary: published 374069 msg/s, confirmed 373742 msg/s, consumed 0 msg/s, latency 95th 0 ms, chunk size 0
```

I created a minimal example comparing `{packet, raw}` vs `{packet, 4}` without using RabbitMQ at all in https://github.com/ansd/packet-parser which also shows that `{packet, raw}` is slower in some cases.

ansd:
Closing this.

As shown in https://github.com/ansd/packet-parser and discussed in https://erlangforums.com/t/performance-of-socket-option-packet-4/1562 socket option `{packet, 4}` is too slow for our purposes. We achieve higher throughput by parsing the packets on our own in the application level (instead of in the Erlang runtime).

Instead of using `{packet, 4}`, we should handle the `iodata()` in subsequent calls rather than turning it into a `binary()`.

PR #5063, opened 2022-06-17
Use v2 sets when recovering bindings if available

gomoripeti:
## Proposed Changes

The overhead of calling function_exported should be balanced out by
faster set creation in case of from_list, especially for long lists.

timer:tc sets:from_list microbenchmark shows 
few 10K queues: 40ms -> 4ms
100K exchanges: 340ms -> 40ms

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

Loading the binding is actually dominated by inserting into the routes tables, not sets:from_list. In case of these huge definitions the gain might be considered negligible. But this change is applicable even after replacing mnesia with khepri.

Alternatively the compatibility code in rabbit_misc can be avoided (not to add some micor-tech-dept), and just using v2 sets in future RabbitMQ versions which only will support OTP 24+ (ie merge to master but dont backport to 3.9, 3.10, 3.11 etc)

### DIFF
--- a/deps/rabbit/src/rabbit_binding.erl
+++ b/deps/rabbit/src/rabbit_binding.erl
@@ -94,8 +94,8 @@ recover() ->
 -spec recover([rabbit_exchange:name()], [rabbit_amqqueue:name()]) ->
                         'ok'.
 recover(XNames, QNames) ->
-    XNameSet = sets:from_list(XNames),
-    QNameSet = sets:from_list(QNames),
+    XNameSet = sets:from_list(XNames, [{version, 2}]),
+    QNameSet = sets:from_list(QNames, [{version, 2}]),
     SelectSet = fun (#resource{kind = exchange}) -> XNameSet;
                     (#resource{kind = queue})    -> QNameSet
                 end,


michaelklishin:
We will drop Erlang 23 support soon enough, so the check wouldn't be necessary.

gomoripeti:
thanks I will update accordingly

michaelklishin:
@gomoripeti sorry, I did not mean we can do this right away, only that this solution won't last for too long. We have to support Erlang 23 for another month or so.

lhoguin:
The sets code in 24.0 is better but it was improved again in subsequent 24.x releases, so the benchmark results might differ depending on OTP version.

gomoripeti:
true that with these remaining 2 lines of change this PR does not make much sense. Its better to consider all sets in RabbitMQ code base together and consider/measure where it makes sense to use v2 and when.
(This was just a side product of my tinkering which I thought I'd ask for some feedback)

michaelklishin:
Once we drop Erlang 23 support, we can revisit what set types we use.

PR #5077, opened 2022-06-20
Message Containers

kjnilsson:
This PR implements an approach for a "protocol (data format) agnostic core" where the format of the message isn't converted at point of reception. 

Currently all non AMQP 0.9.1 originating messages are converted into a AMQP 0.9.1 flavoured `basic_message` record before sent to a queue. If the messages are then consumed by the originating protocol they are converted back from AMQP 0.9.1. For some protocols such as MQTT 3.1 this isn't too expensive as MQTT is mostly a fairly easily mapped subset of AMQP 0.9.1 but for others such as AMQP 1.0 the conversions are awkward and in some cases lossy even if consuming from the originating protocol.

This PR instead wraps all incoming messages in their originating form into a generic, extensible message container type (`mc`). The container module exposes an API to get common message details such as size and various properties (`ttl`, `priority` etc) directly from the source data type. Each protocol needs to implement the `mc` behaviour such that when a message originating form one protocol is consumed by another protocol we convert it to the target protocol at that point. 

The message container also contains annotations, dead letter records and other meta data we need to record during the lifetime of a message. The original protocol message is never modified unless it is consumed.

- [x] Initial implementation
- [x] Review of initial impl
- [x] rabbit_trace - decide what to do, currently it is AMQP legacy only
- [x] rabbitmq_federation - currently uses AMPQ 0.9.1 so would need conversions, uses header table values
- [x] Backwards compatibility
  - [x] Messages themselves
  - [x] Feature flag
  - [x] Exchange types
  - [x] Add test to explicitly test enabling ff whilst messages are being processed
- [ ] Testing
  - [x] performance
  - [ ] upgrade
  - [ ] explorative 
- [x] ~~Find a way to make the changes made to the exchange type behaviour somehow backwards compatible~~
- [x] MQTT native callback module (this would validate the approach)
- [x] AMQP (1.0) Native prototype
- [ ] Formalize and document protocol conversion rules.
  - [x] Document
  - [x] Unit tests AMQP
  - [x] Unit tests MQTT
- [x] Save a few bytes per mc being serialised by shortening annotation keys, 
  - [x] flipping the default of the `durable` annotation, 
  - [ ] ~combining `exchange` and `routing_keys` into a single `route` annotation~


deps/rabbit/src/rabbit_mirror_queue_slave.erl
```diff
@@ -563,21 +563,24 @@ send_mandatory(#delivery{mandatory  = true,
...
+    case IsPersistent of
+        true when ?amqqueue_is_durable(Q) andalso
+                  Status == published ->
+            maps:put(MsgId, {published, ChPid, MsgSeqNo}, MS);
+        _ ->
+            ok = rabbit_classic_queue:confirm_to_sender(ChPid,
+                                                        amqqueue:get_name(Q),
+                                                        [MsgSeqNo]),
+            MS
+    end.
```
lhoguin, line 583: I think the change here is correct but considering mirrored queues are going away soon(tm) I would prefer a simpler change that leaves the two clauses there was before, so as to minimize the risk.

lhoguin (commented):
I have done a quick look at the CQ related code changes to get started since I'm away tomorrow.

deps/rabbit/src/rabbit_variable_queue.erl
```diff
@@ -1,4 +1,3 @@
-%% This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
kjnilsson: oops will fix

deps/rabbit/src/rabbit_variable_queue.erl
```diff
@@ -1197,10 +1199,10 @@ head_message_timestamp(Q3, RPA) ->
                    HeadMsgStatus#msg_status.msg /= undefined ],
 
     Timestamps =
-        [Timestamp || HeadMsg <- HeadMsgs,
-                      Timestamp <- [rabbit_basic:extract_timestamp(
-                                      HeadMsg#basic_message.content)],
-                      Timestamp /= undefined
+        [Timestamp div 1000
```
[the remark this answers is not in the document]
kjnilsson: `mc:timestamp/1` returns the timestamp in milliseconds but (I belive) the head message timestamp assumes seconds (as that is what amqp legacy records it as).

deps/rabbit/src/rabbit_mirror_queue_slave.erl
kjnilsson: I agree but the previous code relies on pattern matching in the function head which with `mc` we can no longer do so we'd need the ispersistent + is durable check in both clauses in that case.

deps/rabbit/src/rabbit_variable_queue.erl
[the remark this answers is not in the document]
lhoguin: OK.

deps/rabbit/src/rabbit_mirror_queue_slave.erl
lhoguin: OK.

deps/rabbit/test/mc_SUITE.erl
```diff
@@ -0,0 +1,639 @@
...
+    MA = #'v1_0.message_annotations'{content = [{{symbol, <<"k">>}, {utf8, <<"v">>}}]},
+    % Desc = {described,
+    %         {utf8, <<"URL">>},
+    %         {utf8, <<"https://rabbitmq.com">>}},
+    MAEnc = amqp10_framing:encode(MA),
+    ct:pal("~p", [MAEnc]),
+    ct:pal("~p", [amqp10_framing:decode(MAEnc)]),
+
+    % amqp10_framing:decode(Desc),
+    ok.
```
the-mikedavis, line 69: Should this be switched to an `?assertEqual(MA, amqp110_framing:decode(MAEnc))`? I assume this is for checking that encoding/decoding leads to the same terms

deps/rabbitmq_mqtt/src/mc_mqtt.erl
```diff
@@ -0,0 +1,480 @@
...
+                         true ->
+                             {utf8, Corr};
+                         false ->
+                             {binary, Corr}
+                     end;
+                 _ ->
+                     undefined
+             end,
+    %% TODO Translate MQTT Response-Topic to AMQP topic.
+    %% If operator did not mofidy mqtt.exchange, set reply-to address to "/topic/" RK.
```
the-mikedavis, line 232:
```suggestion
    %% If operator did not modify mqtt.exchange, set reply-to address to "/topic/" RK.
```

tools/erlang_ls.bzl
```diff
@@ -59,6 +59,10 @@ def _deps_symlinks(ctx):
         target = path_join("..", "..", "..", "bazel-bin", "deps", "rabbit_common", "src", "rabbit_framing_amqp_0_9_1.erl"),
         source = path_join("deps", "rabbit_common", "src", "rabbit_framing_amqp_0_9_1.erl"),
     ))
+    commands.append(_ln_command(
+        target = path_join("..", "..", "..", "bazel-bin", "deps", "amqp10_common", "include", "amqp10_framing.hrl"),
+        source = path_join("deps", "amqp10_common", "include", "amqp10_framing.hrl"),
+    ))
```
the-mikedavis, line 65: This step fails for me locally since the `deps/amqp10_common/include` dir doesn't exist. We might just need to check in a `deps/amqp10_common/include/.gitkeep` file to fix it?

the-mikedavis (commented):
I am not very familiar with these codepaths but this all looks good to me. The changes to `rabbit_exchange_type:route/3` and the dead lettering code moving mostly into `mc` seem like very nice cleanups to me. I have a few minor/typo/nit comments

ansd (approved):

deps/rabbitmq_consistent_hash_exchange/src/rabbit_exchange_type_consistent_hash.erl
```diff
@@ -259,26 +259,22 @@ jump_consistent_hash_value(_B0, J0, NumberOfBuckets, SeedState0) ->
...
-    end;
-value_to_hash({property, Property}, #basic_message { content = Content }) ->
-    #content{properties = #'P_basic'{ correlation_id = CorrId,
-                                      message_id     = MsgId,
-                                      timestamp      = Timestamp }} =
-        rabbit_binary_parser:ensure_content_decoded(Content),
+value_to_hash(undefined, Msg) ->
+    mc:get_annotation(routing_keys, Msg);
+value_to_hash({header, Header}, Msg0) ->
+    maps:get(Header, mc:routing_headers(Msg0, [x_headers]));
```
lazedo, line 265: @kjnilsson hi, this breaks `consistent_hash_exchange` and possibly others because `Header` may not exist in the message in which case `undefined` should be returned as it was previously.
maybe change to `unwrap(mc:x_header(Header, Msg0));` is more appropriate

kjnilsson: submit a PR with a test case?

ansd: fixed in https://github.com/rabbitmq/rabbitmq-server/pull/11808

PR #5103, opened 2022-06-24
Virtual queue type

kjnilsson:


kjnilsson:
Needs https://github.com/rabbitmq/rabbitmq-server/pull/5262

michaelklishin:
This type was implemented in #14474 for Direct Reply-to support (in AMQP 1.0 and beyond).

PR #5105, opened 2022-06-24
Remove dependency on rabbit_channel for deliveries

kjnilsson:
Some queue -> channel messages from classic queues
were missed when the queue type API was introduced. This commit fixes that
which should make use of classic queues portable outside of the
channel.

This includes some refactoring to make more explicit that
the stream_queue feature flag also enables queue types.

### DIFF
--- a/deps/rabbit/src/rabbit_classic_queue.erl
+++ b/deps/rabbit/src/rabbit_classic_queue.erl
@@ -50,7 +50,9 @@
 
 -export([confirm_to_sender/3,
          send_rejection/3,
-         send_queue_event/3]).
+         deliver_to_consumer/5,
+         send_drained/3,
+         send_credit_reply/3]).
 
 is_enabled() -> true.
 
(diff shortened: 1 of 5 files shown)


kjnilsson:
@dumbbell I am unsure how to best utilise the feature flag system to handle refactoring / fixes like this one. For now I have introduced a bogus feature flag but this isn't a feature. It is just a fix for something that was missed in 3.9.x

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -519,22 +523,49 @@ update_msg_status(down, Pid, #msg_status{pending = P} = S) ->
...
     end.
 
-send_queue_event(Pid, QName, Evt) ->
-    gen_server2:cast(Pid, {queue_event, QName, Evt}).
+deliver_to_consumer(Pid, QName, CTag, AckRequired, Message) ->
+    case rabbit_queue_type:is_supported() of
+        true ->
+            Deliver = {deliver, CTag, AckRequired, [Message]},
+            Evt = {queue_event, QName, Deliver},
+            gen_server:cast(Pid, Evt);
```
dumbbell, line 549: Are this message and the messages in the functions below specific to the interaction between a classic queue and a channel?

In other words, it this the right place to encode/format the message?

dumbbell (commented):
Just one question inlined in the patch.

I like the overall approach and the patch looks good to me.

deps/rabbit/src/rabbit_classic_queue.erl
kjnilsson: yes the compatibility code is specific to classic queues, the queue_event format is generic to any queue type. Arguably we could forward to another function in, say, `rabbit_queue_type` but it would still only be used by classic queues as quorum queues do it differently and streams don't send these messages at all.

dumbbell: Ok, thanks!

dumbbell (approved):

PR #5157, opened 2022-07-05
Lower log level on stream creation validation failure

acogoluegnes:
No need to log an error when the arguments are not the same,
info is enough.

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_manager.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_manager.erl
@@ -563,10 +563,18 @@ create_stream(VirtualHost, Reference, Arguments, Username) ->
                 end
             catch
                 exit:ExitError ->
-                    % likely to be a problem of inequivalent args on an existing stream
-                    rabbit_log:error("Error while creating ~p stream: ~p",
-                                     [Reference, ExitError]),
-                    {error, validation_failed}
+                    case ExitError of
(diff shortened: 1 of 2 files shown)


ansd (approved):

deps/rabbitmq_stream/src/rabbit_stream_manager.erl
```diff
@@ -563,10 +563,18 @@ create_stream(VirtualHost, Reference, Arguments, Username) ->
...
             catch
                 exit:ExitError ->
-                    % likely to be a problem of inequivalent args on an existing stream
-                    rabbit_log:error("Error while creating ~p stream: ~p",
-                                     [Reference, ExitError]),
-                    {error, validation_failed}
+                    case ExitError of
+                        % likely a problem of inequivalent args on an existing stream
+                        {amqp_error, precondition_failed, M, _} ->
+                            rabbit_log:info("Error while creating ~p stream, "
```
michaelklishin, line 569: Was this meant to be an info message or a warning as well?

acogoluegnes: It's meant to be `info`. This should be the most common error we catch here - the one the comments point out - so `info` is enough, we don't want to scare people off. We can figure out the problem thanks to the response code the plugin will return. The other case is really unexpected, so `warn` is more adapted, I don't even know what error we can catch there TBH, so some more logging will come in handy.

Issue #5181, opened 2022-07-08
rabbit_stream_coordinator can get overloaded by many connections


mkuratczyk:
`rabbit_stream_coordinator` needs to keep track of AMQP connections and seems to struggle to keep up when there are many new connections or there is high churn (which is not always due to poorly written apps - can be triggered by a node restart, connection issues, etc). To reproduce:

1. start rabbitmq
2. run `rabbitmq-diagnostics observer`
3. run a few instances of `perf-test -x 3000 -y 0 -P 1 -c 1 -u stream -ms --queue-args x-queue-type=stream -ad false -f persistent` (if needed, you may need to stop/start some of them to simulate churn)

For me it looks like this:
![rabbit_stream_coordinator](https://user-images.githubusercontent.com/9566114/178028887-c2e6440c-3c85-4f8f-b0af-2cd823ac67d3.png)

According to the flamegraph, not much overall time is spent in the coordinator but it clearly cannot keep up (we can also see the well known connection tracking problem to be addressed in https://github.com/rabbitmq/rabbitmq-server/pull/5084):
![stream-many-publishers](https://user-images.githubusercontent.com/9566114/178029101-588b1f2c-bf75-480a-90c7-bceb21cf43fc.svg)

Karl said in our internal chat:
`I can see why eval_listeners might be slow; arguably we can probably move the listener functionality to and aux machine and keep it local within the node`

PR #5192, opened 2022-07-11
Optimise rabbit_stream_coordinator:eval_listeners

kjnilsson:
This function can get very slow when a stream has many listeners.

Now we only evaluate when the members have actually changed after
a normal command and we only evaluate a single listener after a new
listener registration.

Fixes / part fixes  / possibly improves https://github.com/rabbitmq/rabbitmq-server/issues/5181


kjnilsson:
I have tested this with 15k publishers (1 listener per publisher) and killed and restarted groups of 3k and I haven't managed to build a backlog or even capture the coordinator doing listener related work in the observer stacktrace view. 

Prior to this I had no problem at least capturing a small backlog.

acogoluegnes (approved):

PR #5202, opened 2022-07-13
Make RabbitMQ 3.8.x feature flags mandatory

dumbbell:
This patch makes feature flags introduced during the life of RabbitMQ 3.8.x mandatory.

The compatibility and migration code behind feature flags marked mandatory is removed. This means two things:
1. Mandatory feature flags are always enabled on the very first boot of a new RabbitMQ node.
2. Affected feature flags MUST be enabled BEFORE upgrading to the version of RabbitMQ which consider them mandatory. A node will refuse to start if a mandatory feature flag is disabled, because all the compatibility code was removed.

In this patch, the following feature flags are now mandatory:
* `quorum_queue`
* `implicit_default_bindings`
* `virtual_host_metadata`
* `maintenance_mode_status`
* `user_limits`

Therefore, if this patch lands in RabbitMQ 3.11.0, a user must enable `quorum_queue` (and all other feature flags listed below) before he/she upgrades from RabbitMQ 3.10.x (or before) to 3.11.x.

Here is an example of an upgraded RabbitMQ node trying to start without the mandatory feature flags enabled:
```
2022-07-13 11:29:28.366877+02:00 [error] <0.232.0> Feature flags: `implicit_default_bindings`: mandatory feature flag not enabled! It must be enabled before upgrading RabbitMQ.
2022-07-13 11:29:28.366905+02:00 [error] <0.232.0> Failed to initialize feature flags registry: {disabled_mandatory_feature_flag,
2022-07-13 11:29:28.366905+02:00 [error] <0.232.0>                                               implicit_default_bindings}
2022-07-13 11:29:28.372830+02:00 [error] <0.232.0>
2022-07-13 11:29:28.372830+02:00 [error] <0.232.0> BOOT FAILED
2022-07-13 11:29:28.372830+02:00 [error] <0.232.0> ===========
2022-07-13 11:29:28.372830+02:00 [error] <0.232.0> Error during startup: {error,failed_to_initialize_feature_flags_registry}
2022-07-13 11:29:28.372830+02:00 [error] <0.232.0>
```

This patch only covers the support for mandatory feature flags and marks the feature flags listed above. As part of that, their migration code is removed because the code now makes an assertion that a mandatory feature flag has no migration function (which will never be executed anyway). However, the removal of the compatibility code will come in a separate pull request.

### DIFF
--- a/deps/rabbit/src/rabbit_core_ff.erl
+++ b/deps/rabbit/src/rabbit_core_ff.erl
@@ -7,12 +7,7 @@
 
 -module(rabbit_core_ff).
 
--export([quorum_queue_migration/3,
-         implicit_default_bindings_migration/3,
-         virtual_host_metadata_migration/3,
-         maintenance_mode_status_migration/3,
-         user_limits_migration/3,
-         direct_exchange_routing_v2_migration/3]).
+-export([direct_exchange_routing_v2_migration/3]).
 
 -rabbit_feature_flag(
    {classic_mirrored_queue_version,
@@ -24,8 +19,7 @@
    {quorum_queue,
     #{desc          => "Support queues of type `quorum`",
       doc_url       => "https://www.rabbitmq.com/quorum-queues.html",
-      stability     => stable,
-      migration_fun => {?MODULE, quorum_queue_migration}
+      stability     => required
      }}).
 
 -rabbit_feature_flag(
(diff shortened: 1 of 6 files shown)


michaelklishin (approved):

michaelklishin:
BuildBuddy is having a bad day.

michaelklishin:
We are observing broad BuildBuddy failures today, so will use local test runs to decide when to merge things.

Issue #5213, opened 2022-07-13
Keep a local record of local listeners, instead of a replicated table with all nodes' listeners


dumbbell:
@dcorbacho and I asked the rest of the team if we could change how listeners are recorded in Mnesia. Currently, they all are written to a replicated Mnesia table. This causes many issues, especially during a network partition. Indeed, all nodes need to monitor when a node goes away and possibly clean its listeners up (except if it's already back online).

The team agrees that local listeners could be recorded locally only. There is no need to replicate this information everywhere. The management API and UI could instead query each node to gather the complete list of listeners. There would even be no changes to the end user.

If we do that, we will get rid of a source of conflict in Mnesia, and this will the integration of Khepri easier.

dumbbell:
This entire change must be behind a feature flag to allow upgrades. The reasons are:
* the `rabbit_listener` table already exists and it's a Mnesia database, not a named ETS table.
* the table is replicated  and existing nodes are relying on it.

This will be the same for tracking tables.

dumbbell:
I believe this can be closed now that the new ETS table is behind a feature flag.

PR #5222, opened 2022-07-14
Keep listeners on a local ram table

dcorbacho:
## Proposed Changes

Keep listeners on a local ram table instead of a replicated table. This avoids an unnecessary cleanup on node down, as listeners are linked to a node's life.

## Types of Changes
- [x] Bug fix (non-breaking change which fixes issue #5213)
- [ ] Build system and/or CI

## Checklist


michaelklishin (changes requested):
The test failures are repeatable and can be reproduced locally

gomoripeti:
wondering what is the benefit of using ram-only mnesia tables with node-specific name and content instead of just an ets table. It is harder to change the schema eg adding a field to the listener record. Also what will happen with these tables after migration to khepri?
asking because this is also relevant for connection and channel tracking tables.

michaelklishin:
@gomoripeti we would have to give up or change all queries. Here the goal is to simply Khepri integration, not much else.

michaelklishin:
The failures are due to an ongoing BuildBuddy issue.

gomoripeti:
> @gomoripeti we would have to give up or change all queries. Here the goal is to simply Khepri integration, not much else.

aha, so the goal in this particular PR is to keep the code change and the chance of any regression minimal, makes sense, thanks!

PR #5234, opened 2022-07-15
Support Rich Authorization Request OAuth2 specification 

MarcialRosales:
## Proposed Changes

[Rich Authorization Request](https://oauth.net/2/rich-authorization-requests/) specification is an extension to the [OAuth2](https://datatracker.ietf.org/doc/html/rfc6749) specification. It is still a draft but it has been stable for a good while. This specification moves away from the concept of **scopes** and instead proposes a new JWT token field called `authorization_details` which consists of a list of permissions where each permission is meant for one `type` of **resource server** and it defines a list of `locations` and a list of `actions` allowed on those `locations`. Here is an example of a JWT token using this specification:

```
{
    ... ,
  "authorization_details": [
     {
      "type":"rabbitmq-endpoint", 
      "locations": [
          "cluster:rabbitmq/vhost:finance"
      ],
      "actions": [ "read", "write", "configure", "administrator" ]
    }
  ]

}
```
A RabbitMQ server/cluster is configured with a `resource_server_id` which uniquely identifies it. And in this new PR we are introducing a new setting called `resource_server_type` which is used as a permissions' discriminator. In other words, RabbitMq will only look into those permissions which are tagged with a type which matches the value configured in `resource_server_type` . In the example above, `resource_server_id` would be `rabbitmq` and `resource_server_type` would be `rabbitmq-endpoint`. 

The specification does not make any recommendations on the exact format of the values in the `locations` field. The format will be defined by the **resource server** the location is targeting. More on that, this PR suggests the following location's structure : 
`cluster:<resource_server_id_pattern>[/vhost:<vhostpattern>][/queue:<queue_name_pattern>|/exchange:<exchange_name_pattern][/routing-key:<routing_key_pattern>]`. 
Any string separated by / which does not conform to `<key>:<value>` is ignored. For instance, if your locations start with a prefix, e.g. vrn/cluster:rabbitmq, the vrn pattern part is ignored. 

Although it is not strictly speaking part of the Rich Authorization Request specification, we are including into this PR the ability to disable the validation of the `aud` field. Thru a new setting called `verify_aud` we can disable it. By default, it is enabled.

## Types of Changes

- [x ] New feature (non-breaking change which adds functionality)
- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

This PR will be validated by stakeholders who requested it before being reviewed and merged/released.

### DIFF
--- a/deps/rabbitmq_auth_backend_oauth2/README.md
+++ b/deps/rabbitmq_auth_backend_oauth2/README.md
@@ -332,11 +332,13 @@ RabbitMQ supports JWT tokens compliant with the extension. Below is a sample exa
 ```
 {
   "authorization_details": [
-    { "type" : "rabbitmq",  
-      "locations": ["cluster:finance/vhost:primary-*"],
+    { 
+      "type" : "rabbitmq",  
+      "locations": ["cluster:finance/vhost:production-*"],
       "actions": [ "read", "write", "configure"  ]
     },
-    { "type" : "rabbitmq",
+    { 
+      "type" : "rabbitmq",
       "locations": ["cluster:finance", "cluster:inventory" ],
       "actions": ["administrator" ]
     }
@@ -349,9 +351,9 @@ Both permissions are meant for a RabbitMQ server whose `resource_server_type` is
 This field is essentially a permission discriminator.
 
 The first permission grants `read`, `write` and `configure` permissions to any vhost which matches
-the pattern `primary-*` that belongs to a cluster whose `resource_server_id` contains the string `finance`.
-The `cluster` attribute is a regular expression like the `vhost`. If we wanted to match exactly the `finance` cluster
-we would use instead `^finance$`.
+the pattern `production-*` that belongs to a cluster whose `resource_server_id` contains the string `finance`.
+The `cluster` attribute is a regular expression like the `vhost`. To match exactly the string `finance`,
+use `^finance$`.
 
 The second permission grants the `administrator` user-tag to both clusters, `finance` and `inventory`. The other
 supported user-tags as `management`, `policymaker` and `monitoring`.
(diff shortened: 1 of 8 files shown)


srinivasbanoth1988:
@MarcialRosales Why do we need tag in actions (example: tag:administrator)?  there is a possibility that identity providers can reserve ":" for some specific use cases. Why can't we consider administrator as admin without "tag"?

{
  "authorization_details": [
    { "type" : "rabbitmq",  
      "locations": ["cluster:finance/vhost:primary-*"],
      "actions": [ "read", "write", "configure"  ]
    },
    { "type" : "rabbitmq",
      "locations": ["cluster:finance", "cluster:inventory" ],
      "actions": ["tag:administrator" ]
    }
  ]
}

michaelklishin:
Because that's how management plug-in roles have been since 2010.

We cannot avoid potential tag name conflicts, the idea is that RabbitMQ tokens/scopes are not reused by arbitrary apps.

MarcialRosales:
It appears this issue with the colon on the actions for user-tags is a blocker. CSP -the Vmware's identity provider- does not tolerate colons on the action's values. To overcome this issue there are a few alternatives. 

1) **very simple** - Drop the prefix `tag:` from the action mane, e,g. `administrator`  or `management`. There should not be  any clash with other resource servers because every authorization entry within `authorization_details` is for just one resource type. 
2) **rather complex** - Configure rabbitmq with `alias_mapping` similar to `scope_mappings`. This rather complex alternative not only allows us to map a user-defined action name to a rabbitmq action name, e.g. `admin` -> `tag:administrator` but also map one user-defined action to many rabbitmq¡'s actions: 
```
{action_aliases, #{
               <<"FullAccess">>      => [<<"tag:administrator">>, <<"configure">>, <<"write">>, <<"read">> ],
```
Although this latter alternative is more flexible I would rather stick to the simpler one.

michaelklishin:
@MarcialRosales both seem OK to me but we can try dropping the prefix since the set of tags used by the management plugin is known ahead of time.

We can also support a different character for separator, e.g. `tag=management`

michaelklishin (approved):

Issue #5215, opened 2022-07-13
Remove compatibility code after RabbitMQ 3.8.x feature flags are made required


dumbbell:
Once we mark feature flags introduced in RabbitMQ 3.8.x required (see #5202), we can get rid of the corresponding compatibility code.

Consider this a "meta-issue" to track all patches related to this.

dumbbell:
Everything was backported to 3.11.x.

PR #5235, opened 2022-07-15
Remove pre-quorum-queue compatibility code

dumbbell:
Quorum queues were introduced in RabbitMQ 3.8.0. This was first time we added a breaking change protected behind a feature flag. This allowed a RabbitMQ cluster to be upgraded one node at a time, without having to stop the entire cluster.

The breaking change was a new field in the `#amqqueue{}` record. This broke the API and the ABI because records are a compile-time thing in Erlang.

The compatibility code is in the wild for long enough that we want to support the new `#amqqueue{}` record only from now on. The `quorum_queue` feature flag was marked as required in a previous commit (see #5202). This allows us to remove code in this patch.

References #5215.


dumbbell:
@Mergifyio backport v3.11.x

PR #5236, opened 2022-07-15
Remove pre-virtual_host_metadata feature flag compatibility code

dumbbell:
This feature flag was introduced because a new field was added to the `#vhost{}` record. This was a breaking change the feature flag allowed a RabbitMQ cluster to be upgraded one node at a time, without having to stop the entire cluster.

The compatibility code is in the wild for long enough that we want to support the new `#vhost{}` record only from now on. The
`virtual_host_metadata` feature flag was marked as required in a previous commit (see #5202). This allows us to remove code in this patch.

References #5215.


michaelklishin:
@mergifyio backport v3.11.x

dumbbell:
@Mergifyio backport v3.11.x

dumbbell:
Mergify can't recreate a new backport, so I will do it manually.

dumbbell:
The backport was repeated in #5435.

PR #5237, opened 2022-07-15
Remove pre-maintenance_mode_status compatibility code

dumbbell:
Maintenance mode, introduced in RabbitMQ 3.8.x, was a breaking change protected behind a feature flag. This allowed a RabbitMQ cluster to be upgraded one node at a time, without having to stop the entire cluster.

The compatibility code is in the wild for long enough. The `maintenance_mode_status` feature flag was marked as required in a
previous commit (see #5202). This allows us to remove code in this patch.

References #5215.


dumbbell:
@Mergifyio backport v3.11.x

PR #5239, opened 2022-07-15
Remove pre-user_limits compatibility code

dumbbell:
User limits required a breaking change protected behind a feature flag. This allowed a RabbitMQ cluster to be upgraded one node at a time, without having to stop the entire cluster.

The breaking change was a new field in the `#internal_user{}` record. This broke the API and the ABI because records are a compile-time thing in Erlang.

The compatibility code is in the wild for long enough that we want to support the new `#internal_user{}` record only from now on. The `user_limits` feature flag was marked as required in a previous commit (see #5202). This allows us to remove code in this patch.

References #5215.


dumbbell:
@Mergifyio backport v3.11.x

PR #5254, opened 2022-07-18
Use new feature flags API for direct_exchange_routing_v2

ansd:
Changes:

1. Use feature_flags_v2 API for feature flag direct_exchange_routing_v2.
Both feature flags feature_flags_v2 and direct_exchange_routing_v2
will be introduced in 3.11.
Therefore, direct_exchange_routing_v2 can depend on feature_flags_v2.

2. rabbit_binding:populate_index_route_table/0 should be run only during the
feature flag migration. Thereafter, upon node start-ups, binding recovery takes
care of populating the table.

3. Merge direct_exchange_routing_v2_post_3_11_SUITE into
direct_exchange_routing_v2_SUITE. We don't need two separate test
suites with almost identical setup and teardown.

Manually tested on K8s a few different scenarios ensuring that the `rabbit_index_route` table is replicated across all nodes:
* podManagementPolicy Parallel
* podManagementPolicy OrderedReady
* rolling restart
* kill single pod
* delete all pods at once
* scale out
 * single node cluster, kill node

The remaining failing test seems to be unrelated to `direct_exchange_routing_v2`:
```
== clustering_management_SUITE ==

  * [unclustered_3_nodes]

    * [cluster_size_3]
      forget_cluster_node (00:21.652)


clustering_management_SUITE > unclustered_3_nodes > cluster_size_3 > forget_cluster_node
    #1. {error,
            {{badmatch,
                 {error,70,
                     "Removing node rmq-ct-forget_cluster_node-3-23100@localhost from the cluster. Warning: quorum queues cannot be shrunk in offline mode
  * Impersonating node: rmq-ct-forget_cluster_node-2-22950@localhost...
 done
  * Mnesia directory: rabbitmq-server/deps/rabbit/logs/ct_run.ct_rabbit@myhost.2022-07-19_12.50.24/deps.rabbit.clustering_management_SUITE.groups.forget_cluster_node.logs/run.2022-07-19_12.50.24/log_private/rmq-ct-forget_cluster_node-2-22950@localhost/mnesia/rmq-ct-forget_cluster_node-2-22950@localhost...
Stack trace: 

** (UndefinedFunctionError) function :rabbit_misc.rabbitmq_related_apps/0 is undefined or private
    :rabbit_misc.rabbitmq_related_apps()
    (rabbit 3.8.10+1487.gb2fbf15) rabbit_feature_flags.erl:935: :rabbit_feature_flags.query_supported_feature_flags/0
    (rabbit 3.8.10+1487.gb2fbf15) rabbit_ff_registry_factory.erl:164: :rabbit_ff_registry_factory.maybe_initialize_registry/3
    (rabbit 3.8.10+1487.gb2fbf15) rabbit_ff_registry_factory.erl:123: :rabbit_ff_registry_factory.initialize_registry/3
    (rabbit 3.8.10+1487.gb2fbf15) rabbit_ff_registry.erl:124: :rabbit_ff_registry.is_enabled/1
    (rabbit 3.8.10+1487.gb2fbf15) rabbit_feature_flags.erl:709: :rabbit_feature_flags.is_enabled/2
    (rabbit 3.8.10+1487.gb2fbf15) internal_user.erl:116: :internal_user.record_version_to_use/0
    (rabbit 3.8.10+1487.gb2fbf15) internal_user.erl:123: :internal_user.fields/0

:undef"}},
             [{clustering_management_SUITE,forget_cluster_node,1,
                  [{file,"clustering_management_SUITE.erl"},{line,315}]},
              {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1782}]},
                  [{file,"test_server.erl"},{line,1291}]},
                  [{file,"test_server.erl"},{line,1223}]}]}}
```


deps/rabbit/src/rabbit_core_ff.erl
```diff
@@ -100,22 +104,5 @@ direct_exchange_routing_v2_migration(FeatureName, _FeatureProps, enable) ->
...
-                    ok;
-                {error, Reason} ->
-                    rabbit_log_feature_flags:warning("Failed to add copy of table ~s to node ~p: ~p",
-                                                     [TableName, node(), Reason])
-            end;
-        _ ->
-            ok
-    end,
-    Enabled.
+direct_exchange_routing_v2_migration(#ffcommand{command = post_enable}) ->
```
dumbbell, line 107: You should rather use a "catch all" clause and ignore the argument. This allows us to add commands without having to update all migration functions.

deps/rabbit/src/rabbit_core_ff.erl
```diff
@@ -63,25 +65,27 @@
...
-       stability     => stable
-     }}).
 
 %% -------------------------------------------------------------------
 %% Direct exchange routing v2.
 %% -------------------------------------------------------------------
 
-direct_exchange_routing_v2_migration(FeatureName, _FeatureProps, enable) ->
+direct_exchange_routing_v2_migration(#ffcommand{name = FeatureName,
+                                                command = enable}) ->
```
dumbbell, line 88: The function should not return the result of `rabbit_binding:populate_index_route_table()` directly because if this one changes, it will break the migration function. A better way would be to handle the return value of that call and return `ok` if it was successful, or `{error, Reason}` if it was not.

deps/rabbit/src/rabbit_table.erl
```diff
@@ -392,13 +392,16 @@ definitions() ->
 maybe_ensure_index_route_table() ->
```
dumbbell, line 392: Unrelated to this patch, but I have a hard time making sens of the name "maybe ensure". Am I sure in the end that the table exists? Maybe, I don't know :-)

dumbbell (changes requested):

michaelklishin (approved):

michaelklishin (approved):

deps/rabbit/src/rabbit_table.erl
```diff
@@ -389,17 +391,22 @@ definitions() ->
...
-            case rabbit_table:ensure_table_copy(rabbit_index_route, node(), ram_copies) of
-                ok ->
-                    ok;
-                {error, Reason} ->
-                    rabbit_log_feature_flags:warning(
-                      "Failed to add copy of table ~s to node ~p: ~p",
-                      [rabbit_index_route, node(), Reason])
-            end;
+            ok = create(rabbit_index_route, rabbit_binding:index_route_table_definition()),
+            ok = ensure_table_copy(rabbit_index_route, node(), ram_copies);
```
dumbbell, line 409: To me, the reasoning is incorrect or at least confusing. A feature flag sits on top of a feature, not the opposite.

When implementing a new feature, the feature-specific code is responsible for creating any resources it relies on: tables, files, processes, whatever. Then, and only then, we can talk about a feature flag to allow a RabbitMQ node with that feature-specific code to work with a RabbitMQ node which doesn't. A feature should **never** rely on the feature flag to set anything up for it.

Therefore the notion of "ensure_feature_flag_tables()" is incorrect because feature flags are not responsible for any tables.

In the case of this `direct_exchange_routing_v2` feature, we should have a function taking care of creating whatever table it needs (I don't know which module should own that function because currently there is no function, just code duplicated in at least two places).

That function is called when it is relevant during startup. If an upgrade only consists of creating those tables, then the feature flag migration function simply calls that same function.

This whole feature currently feels like it is designed around its feature flag and it's wrong. When the feature flag becomes required, the migration function will go away and at best the comment becomes wrong, at worst, RabbitMQ breaks.

michaelklishin: For an alternative implementation of routing that is currently enabled by a feature flag, we cannot realistically enabled it when it is used: that's on the hot message path.

At some point the feature flag will go away as this routing implementation will be the primary one. But we have to see how it behaves in the wild first.

ansd: Thank you, that's very valuable feedback.
I attempt to provide a better way in https://github.com/rabbitmq/rabbitmq-server/pull/5279 where there is "no special code" anymore: In there, the new table `rabbit_index_route` is treated as any other Mnesia table if feature flag `direct_exchange_routing_v2` is enabled.

PR #5262, opened 2022-07-19
Feature flags: Support "auto-enable" feature flags

dumbbell:
A feature flag can be marked as "auto-enable" by setting `auto_enable` to true in its properties.

An auto-enable feature flag is automatically enabled as soon as all nodes in the cluster support it. This is achieved by trying to enable it when RabbitMQ starts, when a plugin is enabled/disabled or when a node joins/re-joins a cluster. If the feature flag can't be enabled, the error is ignored.

However, it should be used cautiously, especially if the feature flag has a migration function, because it might be enabled at an inappropriate time w.r.t. the user's workload.

### DIFF
--- a/deps/rabbit/src/rabbit.erl
+++ b/deps/rabbit/src/rabbit.erl
@@ -871,6 +871,7 @@ start(normal, []) ->
         %% will be used. We start it now because we can't wait for boot steps
         %% to do this (feature flags are refreshed before boot steps run).
         ok = rabbit_sup:start_child(rabbit_ff_controller),
+        rabbit_feature_flags:enable_auto(),
 
         %% Compatibility with older RabbitMQ versions + required by
         %% rabbit_node_monitor:notify_node_up/0:
--- a/deps/rabbit/src/rabbit_feature_flags.erl
+++ b/deps/rabbit/src/rabbit_feature_flags.erl
@@ -88,6 +88,7 @@
          enable/1,
          enable_all/0,
          enable_all/1,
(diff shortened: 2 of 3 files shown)


kjnilsson (approved):
I can't claim to understand the ff code very well but in terms of the description and this feature looks good to me and will be great for simple cross-broker protocol changes that users should not have to interact with directly.

dumbbell:
@kjnilsson: Do you mind if I wait a real use case before merging this? I mean, when you have a feature flag which would benefit from this, then we can really test that the behavior is the one expected.

kjnilsson:
> @kjnilsson: Do you mind if I wait a real use case before merging this? I mean, when you have a feature flag which would benefit from this, then we can really test that the behavior is the one expected.

The two linked PRs need this feature. How should we proceed?

dumbbell:
> > @kjnilsson: Do you mind if I wait a real use case before merging this? I mean, when you have a feature flag which would benefit from this, then we can really test that the behavior is the one expected.
> 
> The two linked PRs need this feature. How should we proceed?

You could rebase your branches on top of this one and see how it goes when doing manal upgrade tests and/or running testsuites in a mixed-version mode.

You won't have any migration function associated with your feature flags, right?

kjnilsson:
I don't have any feature flags apart from a dummy one so no.

Ok I will take a look

dumbbell:
Closing this pull request for now. We don't have any use case for it currently.

PR #5263, opened 2022-07-19
Restore the feature flags suite under mixed version testing

pjk25:
Previously when CI switched to bazel, these tests were skipped because they relied on functions only exported for test, and bazel mixed version testing uses non-dev builds of rabbit for the secondary umbrella. Recent changes to the suite mean that this is no longer the case.

### DIFF
--- a/deps/rabbit/test/feature_flags_SUITE.erl
+++ b/deps/rabbit/test/feature_flags_SUITE.erl
@@ -847,9 +847,7 @@ clustering_ok_with_new_ff_disabled(Config) ->
                         #{desc => "Time travel with RabbitMQ",
                           provided_by => rabbit,
                           stability => stable}},
-    rabbit_ct_broker_helpers:rpc(
-      Config, 0,
-      rabbit_feature_flags, inject_test_feature_flags, [NewFeatureFlags]),
+    inject_ff_on_nodes(Config, [0], NewFeatureFlags),
 
     FFSubsysOk = is_feature_flag_subsystem_available(Config),
 
@@ -883,9 +881,7 @@ clustering_denied_with_new_ff_enabled(Config) ->
                         #{desc => "Time travel with RabbitMQ",
                           provided_by => rabbit,
                           stability => stable}},
-    rabbit_ct_broker_helpers:rpc(
-      Config, 0,
-      rabbit_feature_flags, inject_test_feature_flags, [NewFeatureFlags]),
+    inject_ff_on_nodes(Config, [0], NewFeatureFlags),
     enable_feature_flag_on(Config, 0, time_travel),
 
     FFSubsysOk = is_feature_flag_subsystem_available(Config),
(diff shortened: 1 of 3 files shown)


pjk25:
@mergifyio rebase

pjk25:
Will need a rebase again after #5302 is merged

pjk25:
@mergifyio rebase

pjk25:
#5302 has now been merged, but I think it would be sensible to get master back to green before this can be merged (provided it does not need further corrections itself).

dumbbell (approved):

pjk25:
Now that this has been rebased against the other changes, it looks like there might be an issue loading https://github.com/rabbitmq/inet_tcp_proxy on the mixed version nodes. I'm investigating.

pjk25:
inet_tcp_proxy was a red herring. One of the failures is due to `rabbit_feature_flags:inject_test_feature_flags/1` being only exported for test on all current releases. I will think a little bit about how to best address/workaround the limitation.

pjk25:
This will require #5336

dumbbell:
I just force-pushed the branch with two more fixes and rebased on the latest `master` (which includes #5336). It passes successfully locally for me, both in regular and mixed-version testing.

deps/rabbit/test/feature_flags_SUITE.erl
```diff
@@ -112,59 +112,53 @@ end_per_suite(Config) ->
...
+    %% `feature_flags_v1' group already covers testing in that case.
+    case rabbit_ct_helpers:get_config(Config, secondary_umbrella) of
+        undefined ->
+            rabbit_ct_helpers:set_config(Config, {enable_feature_flags_v2, true});
+        _ ->
+            %% To determine that `feature_flags_v2' are supported, we can't
+            %% query RabbitMQ which is not started. Therefore, we check if the
+            %% source of `rabbit_ff_controller' is present.
+            Dir1 = ?config(rabbit_srcdir, Config),
+            Dir2 = ?config(secondary_rabbit_srcdir, Config),
```
pjk25, line 126: I'm going to validate that this method of checking will also work with bazel, as there are not secondary sources present

deps/rabbit/test/feature_flags_SUITE.erl
```diff
@@ -112,59 +112,90 @@ end_per_suite(Config) ->
...
+    %% query RabbitMQ which is not started. Therefore, we check if the
+    %% source or bytecode of `rabbit_ff_controller' is present.
+
+    IsSupported = case rabbit_ct_helpers:get_config(Config, rabbitmq_run_cmd) of
+                      undefined ->
+                          %% make
+                          Dir1 = ?config(rabbit_srcdir, Config),
+                          File1 = filename:join([Dir1, "src", "rabbit_ff_controller.erl"]),
+                          SupportedPrimary = filelib:is_file(File1),
+                          case rabbit_ct_helpers:get_config(Config, secondary_umbrella) of
```
dumbbell, line 129: The order of checks was reversed in this patch.

The use of a secondary umbrella should be checked first. If there is none configured, the `feature_flags_v2` test group should be executed regardless of the support. Because the test group was committed with the code, it's an error if the test group runs without the corresponding code in a non-mixed-version scenario.

deps/rabbit/test/feature_flags_SUITE.erl
```diff
@@ -112,59 +112,90 @@ end_per_suite(Config) ->
...
+                                  Dir2 = ?config(secondary_rabbit_srcdir, Config),
+                                  File2 = filename:join([Dir2, "src", "rabbit_ff_controller.erl"]),
+                                  SupportedSecondary = filelib:is_file(File2),
+                                  [SupportedPrimary, SupportedSecondary]
+                          end;
+                      _ ->
+                          %% bazel
+                          Dir1 = ?config(rabbit_srcdir, Config),
+                          File1 = filename:join([Dir1, "ebin", "rabbit_ff_controller.beam"]),
+                          SupportedPrimary = filelib:is_file(File1),
```
dumbbell, line 142: This will work with Make as well.

Beside the paths to `File1` and `File2`, the rest of the code should be common to Make and Bazel. This should simplify it.

deps/rabbit/test/feature_flags_SUITE.erl
```diff
@@ -112,59 +112,90 @@ end_per_suite(Config) ->
...
+            {skip, "This version does not support "
+             "feature_flags_v2"};
+        [false, _] ->
+            {skip,
+             "Current version umbrella does not support "
+             "feature_flags_v2"};
+        [true, false] ->
+            {skip,
+             "Previous version umbrella does not support "
+             "feature_flags_v2"}
```
dumbbell, line 175: Using a secondary umbrella works with swapped versions too (primary == old, secondary == new), so the wording is incorrect.

deps/rabbit/test/feature_flags_SUITE.erl
```diff
@@ -112,59 +112,90 @@ end_per_suite(Config) ->
...
+    case IsSupported of
+        [true] ->
+            rabbit_ct_helpers:set_config(
+              Config, {enable_feature_flags_v2, true});
+        [true, true] ->
+            rabbit_ct_helpers:set_config(
+              Config, {enable_feature_flags_v2, true});
+        [false] ->
+            {skip, "This version does not support "
+             "feature_flags_v2"};
```
dumbbell, line 167: Once the checks are re-ordered, this clause should be removed as it will never happen.

dumbbell (changes requested):

dumbbell (changes requested):
Could you please split this commit into two? Both chunks share nothing.

dumbbell (approved):

PR #5266, opened 2022-07-20
Log when acquiring or releasing a global lock

ansd:
to easily detect dead locks.

Prior to this commit, one caller of function
`rabbit_ff_registry_factory:acquire_state_change_lock/0`
was logging that it acquires the lock.
However, there are other callers which did not log.
Therefore, it was impossible to detect any dead locks by just
looking at the the logs.


michaelklishin (approved):

ansd:
This is specific to the FF subsystem v2 in 3.11.

PR #5301, opened 2022-07-21
Move connection and channel tracking tables to ETS

dcorbacho:
The original cluster-wide tracking tables were transformed into local tables, which are ram only. These can be moved into ets tables making its use much simpler

Same implementation changes as those in https://github.com/rabbitmq/rabbitmq-server/issues/5213

### DIFF
--- a/deps/rabbit/src/rabbit_channel_tracking.erl
+++ b/deps/rabbit/src/rabbit_channel_tracking.erl
@@ -227,7 +227,7 @@ shutdown_tracked_items(TrackedItems, _Args) ->
 list() ->
     lists:foldl(
       fun (Node, Acc) ->
-              list_on_node(Node) ++ Acc
+              Acc ++ list_on_node(Node)
       end, [], rabbit_nodes:all_running()).
 
 -spec list_of_user(rabbit_types:username()) -> [rabbit_types:tracked_channel()].
--- a/deps/rabbit/src/rabbit_channel_tracking.erl
+++ b/deps/rabbit/src/rabbit_channel_tracking.erl
@@ -30,11 +30,17 @@
          tracked_channel_table_name_for/1,
          tracked_channel_per_user_table_name_for/1,
          get_all_tracked_channel_table_names_for_node/1,
+         ensure_tracked_tables_for_this_node/0,
          delete_tracked_channel_user_entry/1]).
+
+-export([migrate_tracking_records/0]).
 
 -include_lib("rabbit_common/include/rabbit.hrl").
 
 -import(rabbit_misc, [pget/2]).
+
+-define(TRACKED_CHANNEL_TABLE, tracked_channel).
+-define(TRACKED_CHANNEL_TABLE_PER_USER, tracked_channel_per_user).
 
 %%
 %% API
@@ -45,11 +51,7 @@
 
 boot() ->
     ensure_tracked_channels_table_for_this_node(),
-    rabbit_log:info("Setting up a table for channel tracking on this node: ~p",
-                    [tracked_channel_table_name_for(node())]),
     ensure_per_user_tracked_channels_table_for_node(),
-    rabbit_log:info("Setting up a table for channel tracking on this node: ~p",
-                    [tracked_channel_per_user_table_name_for(node())]),
     clear_tracking_tables(),
     ok.
 
@@ -116,22 +118,43 @@ handle_cast({user_deleted, Details}) ->
             delete_tracked_channel_user_entry, [Username]),
     ok;
 handle_cast({node_deleted, Details}) ->
-    Node = pget(node, Details),
-    rabbit_log_channel:info(
-        "Node '~s' was removed from the cluster, deleting"
-        " its channel tracking tables...", [Node]),
-    delete_tracked_channels_table_for_node(Node),
-    delete_per_user_tracked_channels_table_for_node(Node).
+    case rabbit_feature_flags:is_enabled(tracking_records_in_ets) of
+        true ->
+            ok;
+        false ->
+            Node = pget(node, Details),
+            rabbit_log_channel:info(
+              "Node '~s' was removed from the cluster, deleting"
+              " its channel tracking tables...", [Node]),
+            delete_tracked_channels_table_for_node(Node),
+            delete_per_user_tracked_channels_table_for_node(Node)
+    end.
 
 -spec register_tracked(rabbit_types:tracked_channel()) -> ok.
 -dialyzer([{nowarn_function, [register_tracked/1]}]).
 
-register_tracked(TrackedCh =
-  #tracked_channel{node = Node, name = Name, username = Username}) ->
+register_tracked(TrackedCh = #tracked_channel{node = Node}) when Node == node() ->
+    case rabbit_feature_flags:is_enabled(tracking_records_in_ets) of
+        true -> register_tracked_ets(TrackedCh);
+        false -> register_tracked_mnesia(TrackedCh)
+    end.
+
+register_tracked_ets(TrackedCh = #tracked_channel{node = Node, name = Name, username = Username}) ->
+    ChId = rabbit_tracking:id(Node, Name),
+    case ets:lookup(?TRACKED_CHANNEL_TABLE, ChId) of
+        []    ->
+            ets:insert(?TRACKED_CHANNEL_TABLE, TrackedCh),
+            ets:update_counter(?TRACKED_CHANNEL_TABLE_PER_USER, Username, 1, {Username, 0});
+        [#tracked_channel{}] ->
+            ok
+    end,
+    ok.
+
+register_tracked_mnesia(TrackedCh =
+                            #tracked_channel{node = Node, name = Name, username = Username}) ->
     ChId = rabbit_tracking:id(Node, Name),
     TableName = tracked_channel_table_name_for(Node),
     PerUserChTableName = tracked_channel_per_user_table_name_for(Node),
-    %% upsert
     case mnesia:dirty_read(TableName, ChId) of
       []    ->
           mnesia:dirty_write(TableName, TrackedCh),
@@ -142,8 +165,21 @@ register_tracked(TrackedCh =
     ok.
 
 -spec unregister_tracked(rabbit_types:tracked_channel_id()) -> ok.
+unregister_tracked(ChId = {Node, _Name}) when Node == node() ->
+    case rabbit_feature_flags:is_enabled(tracking_records_in_ets) of
+        true -> unregister_tracked_ets(ChId);
+        false -> unregister_tracked_mnesia(ChId)
+    end.
 
-unregister_tracked(ChId = {Node, _Name}) when Node =:= node() ->
+unregister_tracked_ets(ChId) ->
+    case ets:lookup(?TRACKED_CHANNEL_TABLE, ChId) of
+        []     -> ok;
+        [#tracked_channel{username = Username}] ->
+            ets:update_counter(?TRACKED_CHANNEL_TABLE_PER_USER, Username, -1),
+            ets:delete(?TRACKED_CHANNEL_TABLE, ChId)
+    end.
+
+unregister_tracked_mnesia(ChId = {Node, _Name}) when Node =:= node() ->
     TableName = tracked_channel_table_name_for(Node),
     PerUserChannelTableName = tracked_channel_per_user_table_name_for(Node),
     case mnesia:dirty_read(TableName, ChId) of
@@ -155,17 +191,30 @@ unregister_tracked(ChId = {Node, _Name}) when Node =:= node() ->
 
 -spec count_tracked_items_in({atom(), rabbit_types:username()}) -> non_neg_integer().
 
-count_tracked_items_in({user, Username}) ->
-    rabbit_tracking:count_tracked_items(
+count_tracked_items_in(Type) ->
+    case rabbit_feature_flags:is_enabled(tracking_records_in_ets) of
+        true -> count_tracked_items_in_ets(Type);
+        false -> count_tracked_items_in_mnesia(Type)
+    end.
+
+count_tracked_items_in_ets({user, Username}) ->
+    rabbit_tracking:count_tracked_items_ets(
+      ?TRACKED_CHANNEL_TABLE_PER_USER, Username,
+      "channels of user").
+
+count_tracked_items_in_mnesia({user, Username}) ->
+    rabbit_tracking:count_tracked_items_mnesia(
         fun tracked_channel_per_user_table_name_for/1,
         #tracked_channel_per_user.channel_count, Username,
-        "channels in vhost").
+        "channels of user").
 
 -spec clear_tracking_tables() -> ok.
 
 clear_tracking_tables() ->
-    clear_tracked_channel_tables_for_this_node(),
-    ok.
+    case rabbit_feature_flags:is_enabled(tracking_records_in_ets) of
+        true -> ok;
+        false -> clear_tracked_channel_tables_for_this_node()
+    end.
 
 -spec shutdown_tracked_items(list(), term()) -> ok.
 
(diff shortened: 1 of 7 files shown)


michaelklishin:
@mergifyio rebase

dumbbell (commented):
We may need to introduce a feature flag, just like we do for `rabbit_listener`. I didn't study the code of this patch yet, just noting this possible issue.

dcorbacho:
Yes, we need to introduce a feature flag too

mkuratczyk:
This is a great change! It significantly improves performance under connection churn. I've tried my reproduction steps for https://github.com/rabbitmq/rabbitmq-server/discussions/4872 and it's much harder to trigger this problem. I believe this PR also makes https://github.com/rabbitmq/rabbitmq-server/pull/5084 completely redundant.

I used https://github.com/everskies/rabbitmq-bench/ (repo provided by another user to show connection churn handling degradation compared to older RabbitMQ versions; you can read more about this issue report here: https://rabbitmq.slack.com/archives/C1EDN83PA/p1654447560355859). The results with `go run main.go amqp://guest:guest@localhost:5672/ 500` (500 concurrent threads open a connection, send an unroutable messages and close the connection). On my test box:

* `master`: initially ~3500/s, but thousands of messages accumulate in `rabbit_event` mailbox and gradually fewer and fewer connections can be opened; if the churn continues, it's almost guaranteed RabbitMQ will run out of memory
* `pr5301`: 4200/s, `rabbit_event`'s mailbox is empty; however, `rabbit_logger`'s mailbox has some 200 msgs all the time

Given these results, I wondered what would happen if we disabled logging and the results are somewhat funny: when logging doesn't slow things down, `rabbit_event`'s mailbox accumulates messages even faster than on `master` (in other words, we can run out of memory more efficiently ;) ). It's one of those situations where adding debug logs solves a problem.

I went down the rabbit hole to check if `rabbit_event` could be improved to cope with this increased influx of messages and it seems like it can by adding `{message_queue_data, off_heap}` in https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit_common/src/rabbit_event.erl#L68. With this PR and this additional change, I get 5500 msg/s and an empty `rabbit_event` mailbox. We can handle roughly twice as many churning connections AND withstand that for a long time without running out of memory.

Conclusion: this PR, combined with a minor `rabbit_event` change, may resolve many situations where users run out of memory or run into other performance issues due to connection churn. Obviously churn should still be avoided but it seems like we can push the limits of where churn becomes a problem quite a bit further.

deps/rabbit/src/rabbit_channel_tracking.erl
```diff
@@ -178,109 +166,52 @@ shutdown_tracked_items(TrackedItems, _Args) ->
...
 -spec list_on_node(node()) -> [rabbit_types:tracked_channel()].
 
+list_on_node(Node) when Node == node() ->
+    ets:tab2list(?TRACKED_CHANNEL_TABLE);
 list_on_node(Node) ->
-    try mnesia:dirty_match_object(
-          tracked_channel_table_name_for(Node),
-          #tracked_channel{_ = '_'})
-    catch exit:{aborted, {no_exists, _}} -> []
+    case rabbit_misc:rpc_call(Node, ets, tab2list, [?TRACKED_CHANNEL_TABLE]) of
```
gomoripeti, line 184: I wonder if it would make future changes easier if the rpc call would call the API of rabbit_channel_tracking this was not exposing the implementation detail of ets? This way in a mixed-version cluster each node can have slight differences how they implement listing/counting. A pattern similar to `rabbit_channel:list/list_local` https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit/src/rabbit_channel.erl#L373-L380
(just as a dummy example in that case renaming the ets table would not require upgrade code or feature flag).

This suggestion is valid for all rpcs made in conn/channel tracking but is beyond the scope of the performance optimisation though.

gomoripeti (commented):
great work

(I think the following cleanup is still valid in this PR as well https://github.com/rabbitmq/rabbitmq-server/pull/5084/commits/83091cee3fa9e7bbb61d3d9eeedc216226866547 )

deps/rabbit/src/rabbit_channel_tracking.erl
michaelklishin: I agree

deps/rabbit/src/rabbit_channel_tracking.erl
```diff
@@ -222,29 +294,64 @@ tracked_channel_per_user_table_name_for(Node) ->
...
+            ensure_per_user_tracked_channels_table_for_this_node_mnesia()
+    end.
 
 %% Create tables
-ensure_tracked_channels_table_for_node(Node) ->
+ensure_tracked_channels_table_for_this_node_ets() ->
+    rabbit_log:info("Setting up a table for channel tracking on this node: ~p",
+                    [?TRACKED_CHANNEL_TABLE]),
+    ets:new(?TRACKED_CHANNEL_TABLE, [named_table, public, {write_concurrency, true},
+                                     {keypos, #tracked_connection.id}]).
```
gomoripeti, line 317: typo `#tracked_connection.id` -> `#tracked_channel.id`

deps/rabbit/src/rabbit_channel_tracking.erl
```diff
@@ -155,17 +193,30 @@ unregister_tracked(ChId = {Node, _Name}) when Node =:= node() ->
...
+count_tracked_items_in(Type) ->
+    case rabbit_feature_flags:is_enabled(tracking_records_in_ets) of
+        true -> count_tracked_items_in_ets(Type);
+        false -> count_tracked_items_in_mnesia(Type)
+    end.
+
+count_tracked_items_in_ets({user, Username}) ->
+    rabbit_tracking:count_tracked_items_ets(
+      ?TRACKED_CHANNEL_TABLE_PER_USER, Username,
+      "channels in vhost").
```
gomoripeti, line 205: "channels for user"

gomoripeti (commented):
this is great work with the feature flags making it compatible, however the current state of the PR reverts the optimisation in a previous state where the key of the channel-tracking table should be pid and not {node, name}. Is this accidental or the optimisation is intentionally left separate for a follow up PR?

deps/rabbit/src/rabbit_tracking.erl
```diff
@@ -70,15 +104,12 @@ match_tracked_items(TableNameFun, MatchSpec) ->
...
 -spec clear_tracking_table(atom()) -> ok.
-
 clear_tracking_table(TableName) ->
     case mnesia:clear_table(TableName) of
         {atomic, ok} -> ok;
         {aborted, _} -> ok
     end.
 
--spec delete_tracking_table(atom(), node(), string()) -> ok.
-
```
dumbbell, line 81: Why is this spec removed?

dumbbell (changes requested):

deps/rabbit/src/rabbit_channel_tracking.erl
```diff
@@ -116,22 +115,46 @@ handle_cast({user_deleted, Details}) ->
...
 
-register_tracked(TrackedCh =
-  #tracked_channel{node = Node, name = Name, username = Username}) ->
+register_tracked(TrackedCh = #tracked_channel{node = Node}) when Node == node() ->
+    case rabbit_feature_flags:is_enabled(tracking_records_in_ets) of
+        true -> register_tracked_ets(TrackedCh);
+        false -> register_tracked_mnesia(TrackedCh)
+    end;
+register_tracked(#tracked_channel{node = Node} = TrackedCh) ->
+    rabbit_misc:rpc_call(Node, ?MODULE, register_tracked, [TrackedCh]).
```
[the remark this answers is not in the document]
dcorbacho: You're right. Even though it's exported, as it's a callback implementation, it's only used internally. I just did a search on the whole server repo and isn't used anywhere else. The function call is guarded both in `channel_created` and `channel_closed` as well as it's connection equivalents. I'll remove the rpc call as it will never be used

dcorbacho:
> this is great work with the feature flags making it compatible, however the current state of the PR reverts the optimisation in a previous state where the key of the channel-tracking table should be pid and not {node, name}. Is this accidental or the optimisation is intentionally left separate for a follow up PR?

@gomoripeti where is that optimisation? In `master` we still use the id

gomoripeti:
> @gomoripeti where is that optimisation? In `master` we still use the id

hi @dcorbacho , sorry if I got confused, I cannot find it in any previous force-pushes of this PR any more, but I was sure I saw it and that your PR supersedes mine (https://github.com/rabbitmq/rabbitmq-server/pull/5084) and included the essence of the same change (`rabbit_channel_tracking:get_tracked_channel_by_pid` does a full table scan because ChannelPid is not the key making unregister expensive in case of lot of channels in the system)
@mkuratczyk did you get good results without the channel keypos optimisation?

anyway, sorry for the noise, I will do some testing and if it still makes sense, submit my changes separately, rebased on yours.

dcorbacho:
Thanks @gomoripeti. My changes were always based on `master` and I'm sure I didn't change how data was stored, only where. I pushed that commit here so we can have a look: https://github.com/rabbitmq/rabbitmq-server/tree/tracking-to-ets-backup

mkuratczyk:
Indeed this version of the change performs worse for me. However, adding `{message_queue_data, off_heap}` to https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit_common/src/rabbit_event.erl#L68 as I mentioned before, seems to still do the magic. I think we should just add it to this PR

dumbbell:
I think any optimizations should go in another pull request, not this one. They are different topics and mixing the two makes it unclear:  optimizations would be lost in the middle of the code refactoring.

dumbbell:
@dcorbacho: In #5345 (not merged as of this writing), I change again the migration function API for `feature_flags_v2`-compliant feature flags like the one you add in this pull request. If #5345 is ready before your pull request, I will coordinate with you and can make the change to adapt the migration function.

mkuratczyk:
Alright, the latest version brings back the performance improvement. The problem was trivial - now that this PR is backwards compatible, a feature flag needs to be enabled to get the new behaviour, but the feature flag wasn't present in the PR, so we were basically testing the old version. :)

Performance improvement is not the goal of this PR and optimizations can be added later. However, the very fact that we are moving functionality from Mnesia to ETS, has significant performance benefits. The fact that a version of the PR wasn't providing these benefits, was a sign that something was off (in this case - we weren't using the new code path at all).

deps/rabbit/src/rabbit_connection_tracking.erl
```diff
@@ -338,91 +427,185 @@ lookup(Name) ->
...
-              catch
-                  exit:{aborted, {no_exists, [Tab, _]}} ->
-                      %% The table might not exist yet (or is already gone)
-                      %% between the time rabbit_nodes:all_running() runs and
-                      %% returns a specific node, and
-                      %% mnesia:dirty_match_object() is called for that node's
-                      %% table.
-                      Acc
-              end
+              list_on_node(Node) ++ Acc
```
dumbbell, line 459: Same comment as `rabbit_channel_tracking:list()` about the reversed order.

deps/rabbit/src/rabbit_tracking.erl
```diff
@@ -37,10 +38,29 @@
...
 
--spec count_tracked_items(function(), integer(), term(), string()) ->
+-spec count_tracked_items_ets(atom(), term(), string()) ->
     non_neg_integer().
+count_tracked_items_ets(Tab, Key, ContextMsg) ->
+    lists:foldl(fun (Node, Acc) when Node == node() ->
+                        N = case ets:lookup(Tab, Key) of
+                                []         -> 0;
+                                [{_, Val}] -> Val
+                            end,
```
dumbbell, line 48: Could you please do the same as you did for `list_on_node()` and use a function to count the local table, then use this function in the RPC call below?

deps/rabbit/src/rabbit_tracking.erl
```diff
@@ -58,9 +78,23 @@ count_tracked_items(TableNameFun, CountRecPosition, Key, ContextMsg) ->
...
--spec match_tracked_items(function(), tuple()) -> term().
-
-match_tracked_items(TableNameFun, MatchSpec) ->
+-spec match_tracked_items_ets(atom(), tuple()) -> term().
+match_tracked_items_ets(Tab, MatchSpec) ->
+    lists:foldl(
+      fun (Node, Acc) when Node == node() ->
+              Acc ++ ets:match_object(
+                       Tab,
+                       MatchSpec);
```
dumbbell, line 87: Likewise, can this be a dedicated function?

deps/rabbit/src/rabbit_tracking_store.erl
```diff
@@ -0,0 +1,47 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
+%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%
+%% Copyright (c) 2007-2022 VMware, Inc. or its affiliates.  All rights reserved.
```
dumbbell, line 5: The copyright year should be 2022 alone. This code didn't exist prior to this year.

deps/rabbit/test/per_user_connection_channel_tracking_SUITE.erl
```diff
@@ -770,30 +771,40 @@ exists_in_tracked_connection_per_vhost_table(Config, VHost) ->
...
+        AllKeys = rabbit_ct_broker_helpers:rpc(Config, NodeIndex,
+                                               mnesia,
+                                               dirty_all_keys, [Tab]),
+        lists:member(Key, AllKeys)
+    catch
+        _:_ ->
+            All = rabbit_ct_broker_helpers:rpc(Config, NodeIndex,
+                                               ets, lookup, [Table, Key]),
+            lists:keymember(Key, 1, All)
+    end.
```
dumbbell, line 807: Could you please check the state of the feature flag instead of trying one then the other?

This will make sure that it's consistent.

dumbbell (changes requested):

dumbbell:
> However, the very fact that we are moving functionality from Mnesia to ETS, has significant performance benefits.

I see. Yes, if the performance improvement is a byproduct of this change, then that's more than fine :-) I was concerned that we changed e.g. the configuration of the tables in the process. The optimization would have been impossible to spot in the diff because of the refactoring at the same time.

dumbbell:
@dcorbacho: I merged #5345. Tell me when you want to rebase your branch, I can assist you in adapting the migration function if you want.

deps/rabbit/src/rabbit_channel_tracking.erl
```diff
@@ -116,22 +118,43 @@ handle_cast({user_deleted, Details}) ->
...
+            delete_tracked_channels_table_for_node(Node),
+            delete_per_user_tracked_channels_table_for_node(Node)
+    end.
 
 -spec register_tracked(rabbit_types:tracked_channel()) -> ok.
 -dialyzer([{nowarn_function, [register_tracked/1]}]).
 
-register_tracked(TrackedCh =
-  #tracked_channel{node = Node, name = Name, username = Username}) ->
+register_tracked(TrackedCh = #tracked_channel{node = Node}) when Node == node() ->
```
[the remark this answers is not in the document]
dcorbacho: Yes, it's exactly the same. It was guarded before on the connection, and nowhere else on the whole `rabbitmq-server` repo is used. I think it's highly unlikely someone uses it on a 3rd party plugin. If we do an `rpc` to the desired node, is the change of behaviour you pointed out on a previous review ;)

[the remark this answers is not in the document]
dumbbell: I like this assertion because it will catch mistakes early. If this is only used locally, then that's an improvement. Thank you for your answer :-)

dumbbell (approved):
Thank you for all the fixes, @dcorbacho!

michaelklishin:
@mergifyio backport v3.11.x

PR #3137, opened 2021-06-23
Faster implementation of rabbit_binding:recover/0

mkuratczyk:
For a large number of bindings, this implementation reduces `rabbit_binding:recover/0`
execution from minutes to seconds. rabbit_binding:recover/2 can still take minutes but it can be improved separately.

Before:
```
[debug] <0.9.0> Time to start RabbitMQ: 136651247 µs
```
After:
```
[debug] <0.222.0> rabbit_binding:recover/0 completed in 1.451836s
[debug] <0.436.0> rabbit_binding:recover/2 for vhost / completed in 51.545797s
[debug] <0.9.0> Time to start RabbitMQ: 83644782 µs
```
That's 136s to 83s for a test with 16k queues and 275k bindings.

PR #5304, opened 2022-07-21
Avoid nesting dirty operations inside transaction

ansd:
From the Mnesia docs:

> Notice that calling (nesting) mnesia:[a]sync_dirty inside a
> transaction-context inherits the transaction semantics.

Therefore, let's not use dirty operations inside a transaction context
because it's confusing to read.

### DIFF
--- a/deps/rabbit/src/rabbit_binding.erl
+++ b/deps/rabbit/src/rabbit_binding.erl
@@ -79,15 +79,17 @@ new(Src, RoutingKey, Dst, Arguments) ->
 
 recover() ->
     rabbit_misc:execute_mnesia_transaction(
-        fun () ->
-            mnesia:lock({table, rabbit_durable_route}, read),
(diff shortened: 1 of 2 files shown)


ansd:
I just did some tests with 50,000 bindings in a 3 node cluster and `rabbit_binding:recover/0` takes always < 0.5 second before and after this PR.
```
14813:2022-07-21 12:27:21.805025+00:00 [debug] <0.677.0> rabbit_binding:recover/0 completed in 0.391371s
```
Although it seems to be consistently a little bit slower after this PR (only max 0.1 second slower).
@mkuratczyk @lhoguin is there a (performance) reason that we use dirty nested operations inside transactions in https://github.com/rabbitmq/rabbitmq-server/pull/3137?

ansd:
As I learnt after some more testing, there is a small performance benefit by using dirty operations inside Mnesia sync transactions. Therefore, closing this PR since it's not worth just for code readability reason.

PR #5305, opened 2022-07-21
Configure default queue type by vhost

kjnilsson:
This allows operators to override the default queue type on a per-vhost
basis by setting the default_queue_type metadata key on the vhost record.

When a queue is declared without specifiying a queue type (x-queue-type)
and there is a default set for the vhost we will augment the declare arguments
with the default. This allows future declares with and without the x-queue-type
argument to succeed.

Also adds a new field to the management create vhost form as well as a new flag for the `rabbitmqctl add_vhost` command.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -17,6 +17,7 @@
          not_found_or_absent/1, not_found_or_absent_dirty/1,
          with/2, with/3, with_or_die/2,
          assert_equivalence/5,
+         augment_declare_args/4,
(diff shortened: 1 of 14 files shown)


michaelklishin (approved):

adamncasey:
I was able to use this when creating a new vhost. Applications connecting to it worked great even though they hadn't been taught about `x-queue-type` 🎉. I appreciate the specific handling of `exclusive` queues (and some other properties) - keeping those classic should help a lot when trying to migrate all distributed queues to quorum. I haven't quite managed to test a [seamless migration](https://youtu.be/7tTfLBqV_zQ?t=767) from classic to quorum with this yet but that's something on my end.

PR #5315, opened 2022-07-22
rabbit_feature_flags: Don't call a feature_flags_v2-compatible migration function with `is_enabled`

dumbbell:
This is a deprecated call which is not present in the `feature_flags_v2` code.

This fixes a crash in mixed-version clusters where `feature_flags_v2` can't be enabled, but the `is_enabled` callback is still called for all feature flags, not just those exposing the old function with an arity of 3.


pjk25:
@dumbbell #5316

dumbbell:
> @dumbbell #5316

Ah, I submitted a concurrent patch for the same change, #5317 :-)

PR #5316, opened 2022-07-22
Update feature flags for mixed version tests

pjk25:
These feature flags are now required and we want to start nodes with them enabled instead of no feature flags at all.

Should not be backported to v3.10.x or other branches.


pjk25:
Closing in favor of #5317

PR #5325, opened 2022-07-25
Fix rabbit_index_route inconsistencies

ansd:
Prior to this commit, when bindings were deleted while enabling feature
flag `direct_exchange_routing_v2` - specifically **after**
`rabbit_binding:populate_index_route_table/0` ran and **before** the feature
flags controller changed the feature state from `state_changing` to `enabled` -
some bindings were incorrectly present in table `rabbit_index_route`.

This commit fixes this issue.

If the state is `state_changing`, bindings must be deleted when the
table already got created.

(Somewhat unexpectedly) checking for table existence within a Mnesia
transaction can return `true` although the subsequent Mnesia table operation
will fail with `{no_exists, rabbit_index_route}`.

Therefore, a lock on the schema table must be set when checking for
table existence.
(Mnesia itself creates a write lock on the schema table when creating a
table, see https://github.com/erlang/otp/blob/09c601fa2183d4c545791ebcd68f869a5ab912a4/lib/mnesia/src/mnesia_schema.erl#L1525 )

An alternative fix would have been to catch `{no_exists,
rabbit_index_route}` outside the Mnesia transaction, i.e. in all the callers of
the `rabbit_binding:remove...` functions and then retry the `rabbit_binding:remove...` function.

This PR should go only into master (and 3.11).

### DIFF
--- a/deps/rabbit/src/rabbit_table.erl
+++ b/deps/rabbit/src/rabbit_table.erl
@@ -25,7 +25,6 @@
 %%----------------------------------------------------------------------------
 
 -type retry() :: boolean().
--type mnesia_table() :: atom().
 
 %%----------------------------------------------------------------------------
 %% Main interface
@@ -40,7 +39,7 @@ create() ->
     ensure_secondary_indexes(),
     ok.
 
--spec create(mnesia_table(), list()) -> rabbit_types:ok_or_error(any()).
+-spec create(mnesia:table(), list()) -> rabbit_types:ok_or_error(any()).
 
 create(TableName, TableDefinition) ->
     TableDefinition1 = proplists:delete(match, TableDefinition),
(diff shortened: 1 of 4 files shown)


deps/rabbit/src/rabbit_table.erl
```diff
@@ -53,8 +52,14 @@ create(TableName, TableDefinition) ->
             throw({error, {table_creation_failed, TableName, TableDefinition1, Reason}})
     end.
 
--spec exists(mnesia_table()) -> boolean().
+-spec exists(mnesia:table()) -> boolean().
 exists(Table) ->
+    case mnesia:is_transaction() of
+        true ->
+            mnesia_schema:get_tid_ts_and_lock(schema, read);
```
michaelklishin, line 59: FTR, this function hasn't really changed in 13 years, so we can assume this "semi-private" API element is here to stay for a few more years before we move entirely to Khepri.

michaelklishin (approved):

mkuratczyk (approved):

ansd:
Manually backported this PR into `v3.11.x` branch.

PR #5332, opened 2022-07-26
Stream queue suite flake fixes

kjnilsson:
Set the policy _before_ creating the stream as there is a current
limitation which means streams won't be updated immediately when
only changing the segment size.


kjnilsson:
Once https://github.com/rabbitmq/osiris/pull/94 is merged this PR can be changed back to use osiris main.

PR #5350, opened 2022-07-28
Bump the rabbitmq version on this branch to 3.12.0 (bazel)

pjk25:
To reduce cache misses, the bazel build uses a hard coded version. A correct value is injected in releases.

However, it's still helpful to have something reasonable when one is looking at logs from tests, especially mixed version tests.

Should not be backported, as the release branches already have reasonable values.

### DIFF
--- a/deps/rabbitmq_cli/test/fixtures/plugins/plugins_with_version_requirements/mock_rabbitmq_plugin_for_3_8-0.1.0/ebin/mock_rabbitmq_plugin_for_3_8.app
+++ b/deps/rabbitmq_cli/test/fixtures/plugins/plugins_with_version_requirements/mock_rabbitmq_plugin_for_3_8-0.1.0/ebin/mock_rabbitmq_plugin_for_3_8.app
@@ -6,5 +6,5 @@
 	{applications, [kernel,stdlib,rabbit]},
 	{mod, {mock_rabbitmq_plugins_01_app, []}},
 	{env, []},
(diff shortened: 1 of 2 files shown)


deps/rabbitmq_cli/test/fixtures/plugins/plugins_with_version_requirements/mock_rabbitmq_plugin_for_3_8-0.1.0/ebin/mock_rabbitmq_plugin_for_3_8.app
```diff
@@ -6,5 +6,5 @@
 	{applications, [kernel,stdlib,rabbit]},
 	{mod, {mock_rabbitmq_plugins_01_app, []}},
 	{env, []},
-	{broker_version_requirements, ["3.8.0", "3.9.0", "3.10.0", "3.11.0"]}
+	{broker_version_requirements, ["3.8.0", "3.9.0", "3.10.0", "3.11.0", "3.12.0"]}
```
michaelklishin, line 9: Should we drop `3.8.0` while at it?

pjk25: Sure, I think that makes sense

PR #5356, opened 2022-07-28
Replace JSON library: move from JSX to Thoas

michaelklishin:
## Proposed Changes

[Thoas](https://github.com/lpil/thoas) is a pure Erlang JSON library that's more efficient
than JSX. Thoas' benchmark suite suggests it is 50% to six times more efficient
on various metrics, from encoding and decoding efficiency to peak memory footprint.

Our interest is two-fold:

 * Reduce peak memory usage for responses that render large collections of objects (e.g. unpaginated listing of queues or connections)
 * Lower the CPU effects of broken monitoring systems that query *all* queues to fetch a single metric value of a single queue (e.g. DataDog)

Pair: @pjk25.

Kudos to @the-mikedavis for contributing to Thoas and fixing an issue
that had been a blocker for the switch; @lpil for producing a new release so quickly (and developing Thoas in the first place) 👏 

## Types of Changes

- [ ] Build system and/or CI

## Checklist


pjk25:
@michaelklishin https://github.com/rabbitmq/trust-store-http still uses jsx, which what causes the build failures: https://github.com/rabbitmq/trust-store-http/blob/a3927b22f1bc72339d05ef9fdb368d5edf4bc0ab/Makefile#L10

The options are:
1. update `trust-store-http` as well
2. have both `jsx` and `thoas` in the `workspace_helpers.bzl` and `MODULE.bazel`

michaelklishin:
Sure, I will update the trust store plugin. Good catch!

pjk25:
I also created https://github.com/rabbitmq/rabbitmq-server/pull/5357 which removes the explicit reference

pjk25:
Actually, #5357 probably would just make the error more obscure. I will adjust it.

deps/rabbit/test/logging_SUITE.erl
```diff
@@ -1383,7 +1383,9 @@ log_and_return_json_object(Context, Metadata, DecodeOpts) ->
                         Content,
                         "^.+\"" ++ RandomMsg ++ "\".+$",
                         ReOpts),
-    Term = rabbit_json:decode(Line, [{labels, attempt_atom} | DecodeOpts]),
+
+    Term0 = rabbit_json:decode(Line, DecodeOpts),
+    Term = rabbit_data_coercion:atomize_keys(Term0),
```
the-mikedavis, line 1388: There was a question on the forum about decoding to existing atom keys: https://erlangforums.com/t/thoas-elixirs-jason-json-library-converted-to-erlang/879/38

Should we await some of that discussion? `Jason` has `Jason.decode(binary, keys: :atoms)` or `Jason.decode(binary, keys: :atoms!)` (existing atoms) but it looks like these may have been intentionally removed from Thoas https://github.com/lpil/thoas/commit/d672dce4aeb64a5e1c606e136238fecf453e7456

I think the separate call to `atomize_keys/1` is a good alternative though

deps/rabbit/test/logging_SUITE.erl
```diff
@@ -1383,7 +1383,9 @@ log_and_return_json_object(Context, Metadata, DecodeOpts) ->
                         Content,
                         "^.+\"" ++ RandomMsg ++ "\".+$",
                         ReOpts),
-    Term = rabbit_json:decode(Line, [{labels, attempt_atom} | DecodeOpts]),
+
+    Term0 = rabbit_json:decode(Line, DecodeOpts),
```
the-mikedavis, line 1387: It looks like these `DecodeOpts` may need to be adjusted to what thoas expects. I see `[return_maps]` being passed to `log_and_return_json_object/3` elsewhere in this suite

Along these lines I see some options being passed in `deps/rabbitmq_ct_helpers/src/rabbit_mgmt_test_util.erl:50` and `deps/rabbitmq_aws/src/rabbit_aws_json.erl:17` that are in jsx format (specifically `[{return_maps, false}]`)

deps/rabbit/test/logging_SUITE.erl
michaelklishin: Yes, it is good enough for our needs

deps/rabbit/test/logging_SUITE.erl
michaelklishin: Hm, yeah, there is no alternative in `thoas` so functions such as `http_get_as_proplist/2` would have to convert the returned value themselves. Let me see where that function is used.

pjk25:
> @michaelklishin https://github.com/rabbitmq/trust-store-http still uses `jsx`, https://github.com/rabbitmq/trust-store-http/blob/a3927b22f1bc72339d05ef9fdb368d5edf4bc0ab/Makefile#L10, which what causes the build failures
> 
> The options are:
> 
> 1. update `trust-store-http` as well
> 2. have both `jsx` and `thoas` in the `workspace_helpers.bzl` and `MODULE.bazel`
> 
> The reference in this repo is in `BUILD.trust_store_http`

For clarity, I'll note that we chose option 3, #5359

mkuratczyk:
The details depend heavily on what we query but for pure JSON encoding efficiency, I think querying the `/api/users` is a good choice since it doesn't involve heavy metric calculations. So I imported [100k users](https://github.com/rabbitmq/sample-configs/blob/main/users/users-100k.json) and started querying `/api/users` over and over. The results are very nice:

* on master, it takes 7.6s on average and the cowboy_stream_h:request_process/3 grows to 331MB
* on this branch, it takes 3.1s and I haven't seen the process grow above 118MB

mkuratczyk (approved):
I took this opportunity to play with https://k6.io/ (a load testing tool, particularly useful for HTTP testing). I run `perf-test -x 0 -y 5000 -Y 4` to create 5000 idle connections, each with 4 channels and each channel declared a queue. For the `/api/queues` tests I additionally imported 100k queues. Each endpoint was hit 10 times and the results are:

`/api/connections/` with 5000 connections:
* `master`: avg=1.47s   min=1.13s   med=1.2s    max=2.18s    p(90)=2.15s   p(95)=2.17s
* `pr5356`: avg=709.44ms min=552.12ms med=619.49ms max=930.43ms p(90)=885.16ms p(95)=907.8ms

`/api/channels/` with 20000 channels:
* `master`: avg=3.24s   min=2.52s   med=2.58s   max=5s       p(90)=4.85s   p(95)=4.93s
* `pr5356`: avg=1.65s    min=1.28s   med=1.36s   max=2.46s    p(90)=2.39s    p(95)=2.42s

`/api/queues/` with 20000 queues:
* `master`: avg=7.43s   min=5.66s   med=5.99s   max=10.99s   p(90)=10.97s  p(95)=10.98s
* `pr5356`:  avg=4.16s   min=3.3s    med=3.44s   max=6.04s    p(90)=5.93s   p(95)=5.98s

`/api/users/` with 120000 users:
* `master`: avg=7.24s   min=7.09s   med=7.19s   max=7.5s     p(90)=7.48s   p(95)=7.49s
* `pr5356`:  avg=2.72s   min=2.64s   med=2.71s   max=2.86s    p(90)=2.75s   p(95)=2.8s

Notes:
1. `k6` splits the results between waiting and receiving. The results above are total duration but in all cases "receiving" was negligible (milliseconds).
2. Memory is hard to monitor as it's only briefly needed for the process that handles the request. Monitoring overall RabbitMQ memory usage shows no difference (but eyeballing `observer` while the requests are handled suggests lower memory usage as mentioned above). I'm not sure if there's a simple way to capture the max memory used by the process before it terminates.

the-mikedavis (approved):

lpil:
Very exciting to see, thank you all!

If this change is detailed in a release post, change log, or similar, and if it is appropriate, I would appreciate it if you could link to my [GitHub sponsors page](https://github.com/sponsors/lpil/). It is thanks to my sponsors that I am able to make open source Erlang projects like Thoas.

michaelklishin:
@lpil we usually link to dependency's repositories. I don't see a lot of harm in linking to your sponsors page if you ask so.

We produce blog posts about new developments from time to time, e.g. with benchmarks.

PR #5357, opened 2022-07-28
Eliminate the need for BUILD.trust_store_http

pjk25:
Instead declare it directly in workspace_helpers.bzl, for simplicity


pjk25:
A note on the asymmetry of the reference in `MODULE.bazel` vs `workspace_helpers.bzl` - which of the files applies depends on whether or not the build is run with `--experimental_enable_bzlmod`. The handling of deps is more robust with bzlmod turned on, and in the case of trust-store-http, it's reading the Makefile and automatically determining the deps. The classic pre-bzlmod behavior is less fancy, and requires us to declare the deps explicitly.

pjk25:
Closing as #5359 addresses it.

PR #5390, opened 2022-07-29
Remove pre-implicit_default_bindings compatibility code

ansd:
Implicit default bindings, introduced in RabbitMQ 3.8.x, was a breaking change protected behind a feature flag. This allowed a RabbitMQ cluster to be upgraded one node at a time, without having to stop the entire cluster.

The compatibility code is in the wild for long enough. The `implicit_default_bindings` feature flag was marked as required in a
previous commit (see https://github.com/rabbitmq/rabbitmq-server/pull/5202). This allows us to remove code in this patch.

References https://github.com/rabbitmq/rabbitmq-server/issues/5215.

Note that we rely here on dirty Mnesia operations of the removed migration function to have done its job (of removing default exchange binding rows).


michaelklishin (approved):

dumbbell:
The backport was repeated in #5439.

PR #5408, opened 2022-08-01
Import default queue type when virtual host is imported

### LINKED ISSUE #5399: 3.11: definition import ignores virtual hosts' default queue type setting
This is in follow-up to #5305. As reported by @adamcasey in https://github.com/rabbitmq/rabbitmq-server/pull/5330#discussion_r932081302:

---
I do think `rabbitmqctl import_definitions` isn't picking this setting up

1. Create vhost-def.json containing `{"vhosts":[{"limits":[],"metadata":{"default_queue_type":"quorum","description":"","tags":[]},"name":"manual-dqt-test"}]}`
    1. Worth noting `export_definitions` does give this output
3. `rabbitmqctl import_definitions vhost-def.json`
4. `rabbitmqctl ctl list_vhosts name,default_queue_type` shows all vhosts as default queue type undefined and client-declared queues default to classic

The export / import behaviour here is a little tricky besides this though - imo the perfect outcome would be all queues were declared on import as though the client had re-sent the queue.declare method. I suspect this would require storing what the `x-queue-type` declaration property was alongside the default queue type though.

ikvmw:
## Proposed Changes

Set VHost's default queue type via import too.
Close #5399

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbit/src/rabbit_definitions.erl
+++ b/deps/rabbit/src/rabbit_definitions.erl
@@ -660,8 +660,9 @@ add_vhost(VHost, ActingUser) ->
     Metadata         = rabbit_data_coercion:atomize_keys(maps:get(metadata, VHost, #{})),
     Description      = maps:get(description, VHost, maps:get(description, Metadata, <<"">>)),
     Tags             = maps:get(tags, VHost, maps:get(tags, Metadata, [])),
+    DefaultQueueType = maps:get(default_queue_type, Metadata, undefined),
 
-    rabbit_vhost:put_vhost(Name, Description, Tags, IsTracingEnabled, ActingUser).
+    rabbit_vhost:put_vhost(Name, Description, Tags, DefaultQueueType, IsTracingEnabled, ActingUser).
 
 add_permission(Permission, ActingUser) ->
     rabbit_auth_backend_internal:set_permissions(maps:get(user,      Permission, undefined),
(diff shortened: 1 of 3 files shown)


ikvmw:
Also what I noticed is that we disassemble and reassamble VHost metadata in two places:
- https://github.com/rabbitmq/rabbitmq-server/blob/ik-import-vhost-default-queue-type-5399/deps/rabbit/src/rabbit_definitions.erl#L657
- https://github.com/rabbitmq/rabbitmq-server/blob/ik-import-vhost-default-queue-type-5399/deps/rabbit/src/rabbit_vhost.erl#L300

Is it really necessary? Can put_vhost deal with metadata on its own?

deps/rabbit_common/src/rabbit_data_coercion.erl
```diff
@@ -32,7 +32,7 @@ to_atom(Val) when is_binary(Val) -> binary_to_atom(Val, utf8).
 -spec to_atom(Val :: atom() | list() | binary(), Encoding :: atom()) -> atom().
 to_atom(Val, _Encoding) when is_atom(Val)   -> Val;
 to_atom(Val, _Encoding) when is_list(Val)   -> list_to_atom(Val);
-to_atom(Val, Encoding)  when is_binary(Val) -> binary_to_atom(Val, Encoding).
+to_atom(Val, Encoding)  when is_binary(Val) -> binary_to_existing_atom(Val, Encoding).
```
michaelklishin, line 35: This function is used all over the place, we cannot make it throw a `badarg` for unknown atoms.

michaelklishin (changes requested):

deps/rabbit_common/src/rabbit_data_coercion.erl
lhoguin: The change doesn't make sense since it leaves another `binary_to_atom` and also the two `list_to_atom` anyway.

If necessary we could have a `to_existing_atom` to be explicit when we know the atom is already defined.

ikvmw: ok, so too brave, I'll remove this commit.

michaelklishin (approved):

PR #5410, opened 2022-08-01
ctl add_vhost: check if relevant feature flags are enabled

michaelklishin:
This adds extra CLI command-level validations for #5305: if the respective queue type feature flag (e.g. `quorum_queue`) is not enabled, specifying `--default-queue-type` would fail with an error.

### DIFF
--- a/deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/add_vhost_command.ex
+++ b/deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/add_vhost_command.ex
@@ -5,7 +5,7 @@
 ## Copyright (c) 2007-2022 VMware, Inc. or its affiliates.  All rights reserved.
 
 defmodule RabbitMQ.CLI.Ctl.Commands.AddVhostCommand do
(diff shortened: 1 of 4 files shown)


deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/add_vhost_command.ex
```diff
@@ -24,7 +24,19 @@ defmodule RabbitMQ.CLI.Ctl.Commands.AddVhostCommand do
     meta = %{description: desc,
              tags: parse_tags(tags),
              default_queue_type: default_qt}
-    :rabbit_misc.rpc_call(node_name, :rabbit_vhost, :add, [vhost, meta, Helpers.cli_acting_user()])
+    # check if the respective feature flag is enabled
+    case default_qt do
+      "quorum" ->
```
ikvmw, line 29: does it make sense to delegate feature flag assertion and rpc call to the relevant queue_type module and have a registry of availbe queue types and do lookup/dispatch instead of hard-coded case

michaelklishin: `rabbit_vhost:do_add/4` can do that, we don't need a registry for that

PR #5415, opened 2022-08-01
Remove the old upgrade subsystem

dumbbell:
Until RabbitMQ 3.8.0, the Upgrade subsystem was the solution to provide upgrade functions to e.g. change a Mnesia table schema or convert an on-disk file format. It did its magic during node startup. Unfortunately, it meant that the whole cluster would have to be shut down first, then a single node would start, run the upgrade steps and finally the rest of the cluster could start again.

RabbitMQ 3.8.0 introduced the new Feature flags subsystem. It allows to provide compatibility code which knows the old and new thing, whatever that thing is (a new file format, a modified Erlang record, a different protocol between two Erlang processes, etc.). A feature flag is enabled at runtime, which means that the cluster can continue to run meanwhile. There is no need to stop a cluster anymore to upgrade.

The Feature flags subsystem is now heavily used and it's time to get rid of the old Upgrade subsystem. That's ok to do it now because starting with RabbitMQ 3.11.0, a cluster can only be upgraded from 3.9.x. Cluster running older versions will have to upgrade to RabbitMQ 3.9.x or 3.10.x before jumping to 3.11.x+.

This patch removes the following things:
* the Upgrade subsystem itself
* the upgrade functions
* the upgrade-specific log file


deps/rabbit/src/rabbit_mnesia_rename.erl
```diff
@@ -159,6 +165,9 @@ finish(FromNode, ToNode, AllNodes) ->
               [FromNode, ToNode, node()])
     end.
 
+nodes_running(Nodes) ->
+    [N || N <- Nodes, rabbit:is_running(N)].
```
michaelklishin, line 169: The same function exists in `rabbit_nodes` as `rabbit_nodes:all_running/0`

dumbbell: If I understand correctle, they don't have the same definition of "running". `rabbit_nodes:all_running/0` returns a list of nodes where Mnesia is running, whereas `rabbit_mnesia_rename:nodes_running/1` filters nodes where RabbitMQ is running.

michaelklishin: Fair enough, we should port this to `rabbit_nodes` under a slightly different name, e.g. `rabbit_nodes:all_with_rabbitmq_running/0`.

michaelklishin: And perhaps alias `rabbit_nodes:all_running/0` to `rabbit_nodes:all_with_mnesia_running/0`. Eventually it can become  `rabbit_nodes:all_with_khepri_running/0` if we need one.

dumbbell: Good idea. I did what you suggest but used a slightly shorted name, `all_running_{mnesia,rabbitmq}()`.

PR #5433, opened 2022-08-04
CQv1: Fix failure to recover messages in rare cases

essen:
When a full recovery was done it was possible to lose messages
for v1 queues when the queues only had a journal file and no
segment files.

In practice it should be a rare event because it requires the
queue (or maybe the node) to crash first and then the vhost or
the node to be restarted gracefully.

## Types of Changes

- [ ] Build system and/or CI

## Other notes

This introduces the first regression test for the property suite.
I plan to do more of that for future fixes when I have an easy
to reproduce test case.

Edit: .... and I opened on the wrong account, again. Sorry about that.

### DIFF
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -320,7 +320,7 @@ suites = [
         PACKAGE,
         name = "classic_queue_prop_SUITE",
         size = "large",
-        shard_count = 4,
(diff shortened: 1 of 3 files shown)


lhoguin:
Failure is just sharding complaining. I updated BUILD.bazel and force pushed.

lhoguin (approved):

PR #5434, opened 2022-08-04
Remove pre-virtual_host_metadata feature flag compatibility code (backport #5236)

dumbbell:
Same as #5409 but after reverting this initial backport to merges all #5215-related patches in the right order.


dumbbell:
Oops, wrong target branch...

PR #5331, opened 2022-07-26
Add retry for erl_epmd:port_please

Gsantomaggio:
Fixes #5322.

The docker image is: `pivotalrabbitmq/rabbitmq:empd_not_found_5322-otp-max`
You should see:
```
my-rabbit-server-0 rabbitmq 2022-07-26 07:51:17.266838+00:00 [debug] <0.689.0> Getting epmd port node 'rabbit', 10 retries left
my-rabbit-server-3 rabbitmq 2022-07-26 07:51:18.596978+00:00 [debug] <0.668.0> Getting epmd port node 'rabbit', 10 retries left
my-rabbit-server-1 rabbitmq 2022-07-26 07:51:20.156612+00:00 [debug] <0.678.0> Getting epmd port node 'rabbit', 10 retries left
my-rabbit-server-2 rabbitmq 2022-07-26 07:51:27.894543+00:00 [debug] <0.701.0> Getting epmd port node 'rabbit', 10 retries left
my-rabbit-server-1 rabbitmq 2022-07-26 07:51:50.167880+00:00 [debug] <0.678.0> Getting epmd port node 'rabbit', 9 retries left
```

last line means that it waited for a bit then re-tried again:
```
my-rabbit-server-1 rabbitmq 2022-07-26 07:51:50.167880+00:00 [debug] <0.678.0> Getting epmd port node 'rabbit', 9 retries left
```

## Types of Changes

What types of changes does your code introduce to this project?

- [x] Bug fix (non-breaking change which fixes issue #NNNN)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

Issue #5322, opened 2022-07-25
Sporadic error during startup: no_epmd_port


ansd:
Observed in Kubernetes. To reproduce:

```
kind create cluster --image kindest/node:v1.21.12@sha256:f316b33dd88f8196379f38feb80545ef3ed44d9197dca1bfd48bcb1583210207
kubectl rabbitmq install-cluster-operator
kubectl rabbitmq create my-rabbit --image rabbitmq:3.10.6-management --replicas 5
kubectl get pods -lapp.kubernetes.io/component=rabbitmq,app.kubernetes.io/name=my-rabbit --watch

# Wait until all pods are READY.
# If you don't see a RESTART, delete that Rabbit cluster.
kubectl rabbitmq delete my-rabbit

# Wait a few seconds until all pods are deleted.
# In case, the pods are stuck in STATUS 'Terminating':
kubectl delete pods -lapp.kubernetes.io/name=my-rabbit --force=true --grace-period=0

# For me a RESTART occurs after the 2nd cluster creation already.
# In below output my-rabbit-server-0 was restarted (but it can be any other pod)
kubectl rabbitmq create my-rabbit --image rabbitmq:3.10.6-management --replicas 5
kubectl get pods -lapp.kubernetes.io/component=rabbitmq,app.kubernetes.io/name=my-rabbit --watch

NAME                 READY   STATUS    RESTARTS   AGE
my-rabbit-server-0   0/1     Running   0          35s
my-rabbit-server-1   1/1     Running   0          35s
my-rabbit-server-2   0/1     Running   0          35s
my-rabbit-server-3   1/1     Running   0          35s
my-rabbit-server-4   1/1     Running   0          35s
my-rabbit-server-2   1/1     Running   0          36s
my-rabbit-server-0   0/1     Completed   0          42s
my-rabbit-server-0   0/1     Running     1          43s
my-rabbit-server-0   1/1     Running     1          54s

# To see the logs of the failed container:
kubectl logs my-rabbit-server-0 -c rabbitmq --previous

...
BOOT FAILED
===========
Error during startup: {error,no_epmd_port}

2022-07-22 16:28:12.538492+00:00 [error] <0.627.0>
2022-07-22 16:28:12.538492+00:00 [error] <0.627.0> BOOT FAILED
2022-07-22 16:28:12.538492+00:00 [error] <0.627.0> ===========
[...]

PR #5438, opened 2022-08-04
Prevent global:sync/0 from being stuck when hostname resolution is not available early on boot

ansd:
Prior to this PR, `global:sync/0` gets sometimes stuck when either performing a rolling update on Kubernetes or when creating a new RabbitMQ cluster on Kubernetes.

When performing a rolling update, the node being booted will be stuck in:
```
2022-07-26 10:49:58.891896+00:00 [debug] <0.226.0> == Plugins (prelaunch phase) ==
2022-07-26 10:49:58.891908+00:00 [debug] <0.226.0> Setting plugins up
2022-07-26 10:49:58.920915+00:00 [debug] <0.226.0> Loading the following plugins: [cowlib,cowboy,rabbitmq_web_dispatch,
2022-07-26 10:49:58.920915+00:00 [debug] <0.226.0>                                 rabbitmq_management_agent,amqp_client,
2022-07-26 10:49:58.920915+00:00 [debug] <0.226.0>                                 rabbitmq_management,quantile_estimator,
2022-07-26 10:49:58.920915+00:00 [debug] <0.226.0>                                 prometheus,rabbitmq_peer_discovery_common,
2022-07-26 10:49:58.920915+00:00 [debug] <0.226.0>                                 accept,rabbitmq_peer_discovery_k8s,
2022-07-26 10:49:58.920915+00:00 [debug] <0.226.0>                                 rabbitmq_prometheus]
2022-07-26 10:49:58.926373+00:00 [debug] <0.226.0> Feature flags: REFRESHING after applications load...
2022-07-26 10:49:58.926416+00:00 [debug] <0.372.0> Feature flags: registering controller globally before proceeding with task: refresh_after_app_load
2022-07-26 10:49:58.926450+00:00 [debug] <0.372.0> Feature flags: [global sync] @ rabbit@r1-server-3.r1-nodes.default
```

During cluster creation, an example log of global:sync/0 being stuck can be found in bullet point 2 of
https://github.com/rabbitmq/rabbitmq-server/pull/5331#pullrequestreview-1050715029.

When global:sync/0 is stuck, it never receives a message in line
https://github.com/erlang/otp/blob/bd05b07f973f11d73c4fc77d59b69f212f121c2d/lib/kernel/src/global.erl#L2942.

This issue can be observed in both `kind` and GKE.

`kind` uses CoreDNS, GKE uses kubedns.

CoreDNS does not resolve the hostname of RabbitMQ and its peers
correctly for up to 30 seconds after node startup.
This is because the default cache value of CoreDNS is 30 seconds and
CoreDNS has a bug described in
https://github.com/kubernetes/kubernetes/issues/92559

global:sync/0 is known to be buggy "in the presence of network failures"
unless the kernel parameter `prevent_overlapping_partitions` is set.

When either:
1. setting CoreDNS cache value to 1 second (see
https://github.com/rabbitmq/rabbitmq-server/issues/5322#issuecomment-1195826135
on how to set this value), or
2. setting the kernel parameter `prevent_overlapping_partitions` to `true`

rolling updates do NOT get stuck anymore.

This means we are hitting here a combination of:
1. Kubernetes DNS bug not updating DNS caches promptly for headless
   services with `publishNotReadyAddresses: true`, and
2. Erlang bug which causes global:sync/0 to hang forever in the presence
   of network failures.

The Erlang bug is fixed by setting `prevent_overlapping_partitions` to `true` (default in Erlang/OTP 25).

In RabbitMQ however, we explicitly set `prevent_overlapping_partitions`
to `false` because other issues may arise if we set this paramter to
`true`.

To resolve this issue of global:sync/0 being stuck, we can just
call function rabbit_node_monitor:global_sync/0 which provides a
workaround.

With this commit applied, rolling updates are not stuck anymore and we
see in the debug log the workaround sometimes being applied.

### DIFF
--- a/deps/rabbit/src/rabbit_ff_controller.erl
+++ b/deps/rabbit/src/rabbit_ff_controller.erl
@@ -268,7 +268,7 @@ register_globally() ->
        "Feature flags: [global sync] @ ~s",
        [node()],
        #{domain => ?RMQLOG_DOMAIN_FEAT_FLAGS}),
-    ok = global:sync(),
+    ok = rabbit_node_monitor:global_sync(),
     ?LOG_DEBUG(
        "Feature flags: [global register] @ ~s",
        [node()],
--- a/deps/rabbit/src/rabbit_node_monitor.erl
+++ b/deps/rabbit/src/rabbit_node_monitor.erl
@@ -276,13 +276,13 @@ pause_if_all_down_guard(PreferredNodes, LastNodes, LastState) ->
 %% state, causing global:sync/0 to never return.
 %%
 %%     1. A process is spawned.
-%%     2. If after 15", global:sync() didn't return, the "global"
+%%     2. If after 10", global:sync() didn't return, the "global"
 %%        state is parsed.
 %%     3. If it detects that a sync is blocked for more than 10",
 %%        the process sends fake nodedown/nodeup events to the two
 %%        nodes involved (one local, one remote).
 %%     4. Both "global" instances restart their synchronisation.
-%%     5. globao:sync() finally returns.
+%%     5. global:sync() finally returns.
 %%
 %% FIXME: Remove this workaround, once we got rid of the change to
 %% "dist_auto_connect" and fixed the bugs uncovered.
@@ -297,13 +297,13 @@ workaround_global_hang() ->
     receive
         global_sync_done ->
             ok
-    after 10000 ->
+    after 10_000 ->
             find_blocked_global_peers()
     end.
 
 find_blocked_global_peers() ->
     Snapshot1 = snapshot_global_dict(),
-    timer:sleep(10000),
+    timer:sleep(10_000),
     Snapshot2 = snapshot_global_dict(),
     find_blocked_global_peers1(Snapshot2, Snapshot1).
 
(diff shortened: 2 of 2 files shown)


michaelklishin:
Should this be backported to 3.10?

ansd:
This should only be backported to 3.11 because rolling updates are only broken on unreleased 3.11 with the new feature flags controller.

dumbbell:
Thank you @ansd! I will review it once I'm done with the backports of the compatibility code removal.

dumbbell (changes requested):
In fact, I have comments on the commit messages, not the code :-)

In the first commit:
> global:sync/0 is known to be buggy "in the presence of network failures" unless the kernel parameter `prevent_overlapping_partitions` is set.

It should be more explicit if it says "is set to true".

In:
> In RabbitMQ however, we explicitly set `prevent_overlapping_partitions` to `false` because other issues may arise if we set this paramter to `true`.

We should be clear the possibility of issues is a feeling, not a fact. I would say something like "because we fear other issues could arise if we set this parameter to `true`". Perhaps we could also mention the commit adding the workaround in the first place (it gives some details).

In the second commit, you should also mention that you fix errors in the comment in addition to changing the log level of the message.

ansd:
Thank you @dumbbell. I amended the commit messages.

dumbbell (approved):
Thanks!

PR #5442, opened 2022-08-04
Set kernel param prevent_overlapping_partitions to true

ansd:
This kernel parameter got introduced in Erlang 24.3.
It is set to `false` by default in Erlang 24.
It is set to `true` by default in Erlang 25.

**This commit requires Erlang >= 24.3.**

As described in commit message https://github.com/rabbitmq/rabbitmq-server/commit/4bf78d822d7496e03061119f4cb07c0b306e4c03 setting this flag to `true` will prevent `global:sync/0` from hanging in the presence of network failures.

Instead of relying on our own workaround of `global:sync/0` being stuck introduced in https://github.com/rabbitmq/rabbitmq-server/commit/9fcb31f348590a74fd526333cf881cfbe27241e6 let us instead rely on the official Erlang fix that comes by setting `prevent_overloapping_partitions` to true.

**We want this commit only in `master`.**

### DIFF
--- a/deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_conf.erl
+++ b/deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_conf.erl
@@ -24,7 +24,7 @@ setup(Context) ->
     %% TODO: Check if directories/files are inside Mnesia dir.
 
     ok = set_default_config(),
-    ok = disable_kernel_overlapping_partitions(),
+    ok = enable_kernel_overlapping_partitions(),
 
     AdditionalConfigFiles = find_additional_config_files(Context),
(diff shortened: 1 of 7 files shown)


ansd:
As expected, the Erlang 23 tests are failing.
We will have to remove Erlang 23 from CI before merging this PR.

Also, before merging this PR (or at the very latest before releasing 3.12 or 4.0), we have to do further network partition testing (for example using [Chaos Mesh](https://chaos-mesh.org/docs/simulate-network-chaos-on-kubernetes/#partition-example) or Jepsen)  to ensure RabbitMQ's partition handling strategies continue to work as expected.

michaelklishin:
We have dropped Erlang 23 tests from Actions.

PR #5483, opened 2022-08-10
Revert "Set kernel param prevent_overlapping_partitions to true"

ansd:
This reverts commit 8070344a38b5d3efb2e6687c73e0a163c12bd5aa (https://github.com/rabbitmq/rabbitmq-server/pull/5442).

We learnt during the last 6 days on master branch that RabbitMQ - as
of today - is not compatible with kernel parameter
`prevent_overlapping_partitions` set to `true`.

RabbitMQ explicitly disconnects node in at least two places:
1. `rabbit_node_monitor` to "promote" a partial network partition
to a full partition, and
2. `rabbit_mnesia` after a node reset to disconnect it from the
rest of the cluster.
There is no atomicity in the way we disconnect several nodes,
because it's a simple loop. Therefore, remote nodes may/will detect
disconnection at different times obviously. In global's new
behavior behind `prevent_overlapping_partitions`, our attempt to
disconnect all nodes in `rabbit_mnesia` creates a partial network
partition from global's point of view, leading to a complete
disconnection of the cluster.

For example, test
```
make ct-clustering_management t=cluster_size_3:join_and_part_cluster
```
was flaky and demonstrates the 2nd bullet point above where RabbitMQ
interfering with Erlang distribution conflicts with global's
`prevent_overlapping_partitions`.
When RabbitMQ resets a node, its last step is to loop over
clustered nodes and disconnect from them one at a time.
In this test with a 3-node cluster where we reset node A:
1. Node A instructs node B and C to remove node A from their view
   of the cluster
2. Node A disconnects from node B
3. global on node B get a nodedow event for node A, but node C is
   still connected to node A
4. global on node B concludes there is a network partition and
   disconnect from node A and node C
At this point, each node is on its own.
Nothing in RabbitMQ tries to restore the connection between
nodes B and C.

The correct path forward is:
1. Get rid of Mnesia replacing it with Khepri.
2. Once mirrored classic queues are removed, get rid of `rabbit_node_monitor`.
3. Have a clear and consistent view of the nodes comprising a RabbitMQ Cluster:
   In other words, do not use different sources of truths like `nodes()`,
   Mnesia, Ra clusters, global monitor at different places in the code.

For the time being we live with `prevent_overlapping_partitions` set to `false`
and with the workaround for `global:sync/0` being stuck introduced in
https://github.com/rabbitmq/rabbitmq-server/commit/9fcb31f348590a74fd526333cf881cfbe27241e6

### DIFF
--- a/deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_conf.erl
+++ b/deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_conf.erl
@@ -24,7 +24,7 @@ setup(Context) ->
     %% TODO: Check if directories/files are inside Mnesia dir.
 
     ok = set_default_config(),
-    ok = enable_kernel_overlapping_partitions(),
+    ok = disable_kernel_overlapping_partitions(),
 
     AdditionalConfigFiles = find_additional_config_files(Context),
     AdvancedConfigFile = find_actual_advanced_config_file(Context),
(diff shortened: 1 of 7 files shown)


dumbbell (approved):
Than you!

PR #5609, opened 2022-08-22
Fix duplicate credit grant in quorum queue

ansd:
Prior to this commit, when a consumer NACKed a message with requeue=true
(resulting in a Ra #return{} command) and that message got
dropped or dead-lettered (for example because the delivery-limit
was exceeded), that consumer got too many credits granted,
which could lead to exceeding (and therefore violating) the
configured Prefetch value.
The consumer got credit granted twice: 1. for completing the message
and 2. for returning the message.

This bug has existed since 3.8. It got reported in
* https://groups.google.com/g/rabbitmq-users/c/iMcX0oXzURQ
* https://rabbitmq.slack.com/archives/C1EDN83PA/p1661257750373939

From now on, credit for a consumer is increased in only 3 places:

* a message is completed, or
* a message is returned, or
* a message is requeued using the new #requeue{} Ra command (when
delivery-limit is not configured)

Furthermore, this commit also fixes a somewhat related bug:
When a consumer with checked out messages gets cancelled, and
a new consumer subscribes with the same consumer ID and on the same channel
but with a lower Prefetch value, that new consumer's Prefetch value was
not always respected. This commit fixes this issue by ensuring that
the credit in simple_prefetch mode does not exceed the consumer
Prefetch value.

We cannot backport to v3.9.x because that branch uses rabbit_fifo Ra machine version 1.
Let's only backport to 3.11.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -626,7 +626,7 @@ convert_msg({Header, empty}) ->
 convert_msg(Header) when ?IS_HEADER(Header) ->
     ?MSG(undefined, Header).
 
-convert_consumer({ConsumerTag, Pid}, CV1) ->
+convert_consumer_v1_to_v2({ConsumerTag, Pid}, CV1) ->
     Meta = element(2, CV1),
     CheckedOut = element(3, CV1),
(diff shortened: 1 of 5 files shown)


kjnilsson (approved):
One tweak possibly

PR #5737, opened 2022-09-09
Stop sending stats to rabbit_event

ansd:
Stop sending `connection_stats` from protocol readers to `rabbit_event`.
Stop sending `queue_stats` from queues to `rabbit_event`.
Sending these stats every 5 seconds to the event manager process is superfluous because noone handles these events.

In theory, a 3rd party plugin could handle `connection_stats` or `queue_stats` events. However, I'm not aware of such a plugin.

They seem to be a relict from before `rabbit_core_metrics` ETS tables got introduced in 2016.


michaelklishin:
I see that the line has previously changed 6 years ago in https://github.com/rabbitmq/rabbitmq-server/commit/b9bdedf0574783be316a9d451076064e2a30b910, and in https://github.com/rabbitmq/rabbitmq-server/commit/47aedfb2b6c1587f0320da4e24cf414d16493fbb before that (9 years ago).

So I conclude that the event indeed is a leftover from before the core stats redesign.

michaelklishin (approved):

michaelklishin:
@mergifyio backport v3.10.x

PR #5769, opened 2022-09-13
rabbit_definitions: Apply default-queue-type logic to imported queues

adamncasey:
## Background
The default-queue-type property on a vhost was added recently in #5305 to aid a workflow for converting to quorum queues. The core logic of that change was applying some special logic which injects `x-queue-type` into `arguments` on a queue when it is declared.

## The issue
I've noticed that this logic isn't applied when queues are imported into a vhost. This is particularly important when importing queues into vhosts which have a default queue type assigned and the imported queue information does not specify a queue type. The imported definition can be missing a queue type for two reasons:

1. The source broker is older than `x-queue-type` has been a concept
2. The metadata has been modified to strip this queue type to facilitate a Quorum Queue migration - possible/desirable if we know that the queue was declared without any `x-queue-type` argument.

The steps to reproduce this issue are:

1. Start a rabbitmq broker without this patch
2. Create a vhost with a default queue type of `quorum`: `rabbitmqctl add_vhost vhost-dqt --default-queue-type quorum`
3. Import a queue to this vhost: `echo '{"queues":[{"name": "test-queue", "durable": true, "auto_delete": false, "vhost": "vhost-dqt"}]}' > defs.json && rabbitmqctl import_definitions defs.json`
     1. Notice that the queue is actually created as a classic queue
4. Start an amqp application which declares this queue with the same arguments
     1. Notice that the amqp application's declaration fails with incompatible arguments: `"PRECONDITION_FAILED - inequivalent arg 'x-queue-type' for queue 'test-queue' in vhost 'vhost-dqt': received the value 'quorum' of type 'longstr' but current is none")`

## The change
This PR extends the import logic to apply this same `x-queue-type` adjustment.

There is quite a bit of code in rabbit_channel.erl for `queue.declare` which isn't applied to queues being imported. But I think this logic is the only part which would make sense to share between import and AMQP (or otherwise) `queue.declare`.

When running through the exact same steps as above, the queue is declared with the correct `default-queue-type` and the AMQP client is able to declare the queue.

Although I think this change fixes the issue - I'm opening this PR as a conversation starter to validate I'm on the right lines for where to fix this. Any & all feedback appreciated.

### DIFF
--- a/deps/rabbit/src/rabbit_definitions.erl
+++ b/deps/rabbit/src/rabbit_definitions.erl
@@ -731,10 +731,22 @@ add_queue_int(Queue, Name, ActingUser) ->
         true ->
             ok;
         false ->
+            VHostName = maps:get(vhost, Queue, undefined),
+            AutoDelete = maps:get(auto_delete, Queue, false),
+            DurableDeclare = maps:get(durable, Queue, true),
+            ExclusiveDeclare = maps:get(exclusive, Queue, false),
+
+            Args0 = args(maps:get(arguments, Queue, undefined)),
+            Args1 = rabbit_amqqueue:augment_declare_args(VHostName,
+                                                         DurableDeclare,
+                                                         ExclusiveDeclare,
+                                                         AutoDelete,
+                                                         Args0),
+
             rabbit_amqqueue:declare(Name,
-                                    maps:get(durable, Queue, undefined),
-                                    maps:get(auto_delete, Queue, undefined),
-                                    args(maps:get(arguments, Queue, undefined)),
+                                    DurableDeclare,
+                                    AutoDelete,
+                                    Args1,
                                     none,
                                     ActingUser)
     end.


michaelklishin:
Yes, this is a good approach. I am curious if the function should remain in the `rabbit_amqqueue` module, perhaps it should as I cannot think of a better place and it has to do with a "queue-y" behavior :)

michaelklishin (approved):

michaelklishin (changes requested):
The test failure in

```
bazel test --config=local //deps/rabbitmq_management:rabbit_mgmt_http_SUITE --test_env FOCUS="-group all_tests_without_prefix -case definitions_vhost_test"
```

seem legit:

```
=== Ended at 2022-09-13 19:27:02
=== Location: [{rabbit_mgmt_test_util,assert_code,194},
              {rabbit_mgmt_test_util,http_upload_raw,151},
              {rabbit_mgmt_http_SUITE,defs_vhost,[1707](file:///private/var/tmp/_bazel_antares/5f8c95396018822f6839c201f00962b5/execroot/__main__/bazel-out/darwin_arm64-fastbuild/testlogs/deps/rabbitmq_management/rabbit_mgmt_http_SUITE/shard_2_of_2/test.outputs/outputs/ct_run.ct-rabbitmq_management-rabbit_mgmt_http_SUITE@sunnyside.2022-09-13_19.25.43/deps.rabbitmq_management.rabbit_mgmt_http_SUITE.groups.cases.logs/run.2022-09-13_19.25.43/rabbit_mgmt_http_suite.src.html#1707)},
              {rabbit_mgmt_http_SUITE,defs_vhost,[1681](file:///private/var/tmp/_bazel_antares/5f8c95396018822f6839c201f00962b5/execroot/__main__/bazel-out/darwin_arm64-fastbuild/testlogs/deps/rabbitmq_management/rabbit_mgmt_http_SUITE/shard_2_of_2/test.outputs/outputs/ct_run.ct-rabbitmq_management-rabbit_mgmt_http_SUITE@sunnyside.2022-09-13_19.25.43/deps.rabbitmq_management.rabbit_mgmt_http_SUITE.groups.cases.logs/run.2022-09-13_19.25.43/rabbit_mgmt_http_suite.src.html#1681)},
              {rabbit_mgmt_http_SUITE,definitions_vhost_test,[1725](file:///private/var/tmp/_bazel_antares/5f8c95396018822f6839c201f00962b5/execroot/__main__/bazel-out/darwin_arm64-fastbuild/testlogs/deps/rabbitmq_management/rabbit_mgmt_http_SUITE/shard_2_of_2/test.outputs/outputs/ct_run.ct-rabbitmq_management-rabbit_mgmt_http_SUITE@sunnyside.2022-09-13_19.25.43/deps.rabbitmq_management.rabbit_mgmt_http_SUITE.groups.cases.logs/run.2022-09-13_19.25.43/rabbit_mgmt_http_suite.src.html#1725)},
              {test_server,ts_tc,1782},
              {test_server,run_test_case_eval1,1291},
              {test_server,run_test_case_eval,1223}]
=== === Reason: {expected,{group,'2xx'},
                           got,400,type,post,path,"/definitions/test",body,
                           "{\"error\":\"bad_request\",\"reason\":\"function_clause\"}"}
  in function  rabbit_mgmt_test_util:assert_code/5 (rabbit_mgmt_test_util.erl, line 194)
  in call from rabbit_mgmt_test_util:http_upload_raw/8 (rabbit_mgmt_test_util.erl, line 151)
  in call from rabbit_mgmt_http_SUITE:defs_vhost/10 (rabbit_mgmt_http_SUITE.erl, line 1707)
  in call from rabbit_mgmt_http_SUITE:defs_vhost/5 (rabbit_mgmt_http_SUITE.erl, line 1681)
  in call from rabbit_mgmt_http_SUITE:definitions_vhost_test/1 (rabbit_mgmt_http_SUITE.erl, line 1725)
```

michaelklishin:
Unfortunately the node only logs

```
2022-09-13 19:37:46.655957+00:00 [error] <0.951.0> Encountered an error when importing definitions: <<"function_clause">>
```

I am investigating

michaelklishin:
800f3586ec9cf5f25bc976af633ccbf1191ff646 should address the failure

PR #5775, opened 2022-09-13
rabbit_definitions: Apply default-queue-type logic to imported queues

michaelklishin:
This is #5769 by @adamncasey with one extra commit from me.


adamncasey:
Thank you 🙂

PR #5788, opened 2022-09-14
Handle error during shovel topology setup gracefully

gomoripeti:
## Proposed Changes

Eg. when the source exchange does not exist the source channel is closed during `queue.bind`, this used to cause the shovel to crash.
With this change the shovel is restarted gracefully, the same way as when eg the source queue is deleted during operation.

This is just a cosmetic change to avoid one crash log which might look scary at first.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbitmq_shovel/src/rabbit_shovel_behaviour.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_shovel_behaviour.erl
@@ -152,6 +152,8 @@ nack(Tag, Multi, #{source := #{module := Mod}} = State) ->
     Mod:nack(Tag, Multi, State).
 
 %% Common functions
+
+%% Count down until we stop publishing in on-confirm mode
 decr_remaining_unacked(State = #{source := #{remaining_unacked := unlimited}}) ->
     State;
 decr_remaining_unacked(State = #{source := #{remaining_unacked := 0}}) ->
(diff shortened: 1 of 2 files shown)


michaelklishin:
I can reproduce the timeouts locally with

``` sh
bazel run test-logs --config=local //deps/rabbitmq_shovel:dynamic_SUITE
```

gomoripeti:
indeed there is a bug when an autodelete condition is immediately satisfied, the parameters are not cleaned up. did not think this small change can have such side effect, good test coverage!

I used to ignore the github actions failures because I had the false impression they always failed, but I will make sure to always check them going forward. sorry about that

michaelklishin:
When Actions fail, the rest of the pipeline (that produces packages, tests them, tests the result against several client libraries) cannot start.

michaelklishin (changes requested):
There are legitimate test failures.

gomoripeti:
closing this as it results in `delete-after=never` (aka permanent) dynamic shovels and autodelete (aka transient) shovels behaving differently in case of an error at topology setup. The autodelete shovels would not be restarted while other shovels would be, even though a topology error has nothing to do with the autodelete property.
(Initially I only considered and tested with permanent shovels)
Will reopen if I have a clearer idea on what should be the behaviour in all error cases.

PR #5811, opened 2022-09-18
Require Erlang/OTP 25.0 (3.11 and 3.12/main only)

michaelklishin:
So that we can be sure that the JIT is available for both amd64
and arm64 users, and we benefit from all the latest compiler and runtime
optimization work.


michaelklishin:
The failures do not seem to be related to this change. There are some ongoing OCI changes, so let's proceed with merging this.

PR #5845, opened 2022-09-22
Use rules_cc 0.0.2

pjk25:
Adopt rules_cc 0.0.2

May help with Xcode 14 compatibility


michaelklishin:
Yes, this does work correctly in my environment with XCode 14.

Issue #2554, opened 2019-05-16
MQTT v5.0 support


wrsrap:
Are there plans to support MQTT v5.0 in this plugin？Thank you.

petersilva:
in case it is helpful, the feature that most interests me is shared queues.   with AMQP 0.9 you just have multiple processes connect to the same queue to do sharing.  with MQTTv3, only one consumer at a time.  there is a concept of shared queues that promises to address that.  Very interested in that feature in particular.

michaelklishin:
That will be provided by streams in 3.9.

petersilva:
No need for new features on the rabbit side.  The existing sharing of queues (has been there last ten years or so) is fine... I just have multiple subscribers connect to the same queue with a prefetch, and I can have 40 subscribers share a high speed queue (that does not build up.)  I don't see a way to do that with MQTTv3.  I think v5 might help.

adiroiban:
For the reference, a document with differences between v3.1.1 and v5 - https://github.com/mqtt/mqtt.org/wiki/Differences-between-3.1.1-and-5.0

briggySmalls:
For me a big appeal of v5.0 is the "Reason code on all ACKs" (in same link as @adiroiban).

If a client attempts to publish/subscribe to a topic that an ACL denies them, the MQTT 3.x specs silently accept the operation. MQTT 5.x introduces the return code for "Not authorised".

petersilva:
What would be really cool for v5 support:
   -- map each topic exchange to the root of mqtt topic tree.
   -- prepend with $share/queue_name/exchange/ .... rest of the topics.

If the exchange is not a topic exchange, it does not participate in the mapping.
That way, people can connect to the same broker with either protocol and access the entire broker.

example:

I want five processes to use the same queue "worker" bind to  "canada.public" topic on the "xwarnings" exchange  in AMQP sharing the messages among the five.  That should be equivalent to subscribing to "$share/worker/xwarnings/canada.public" in MQTT plugin.

petersilva:
fwiw... I use this mapping here:
https://github.com/MetPX/sarracenia/tree/v03_wip/sarracenia/moth

There are MQTT and AMQP sub-classes of the "Messages Organized by Topic Header" I use mosquitto for the MQTT stuff, because the rabbitmq isn't rich enough currently.

BastianMattes:
Any updates on when MQTT 5 will be supported? Do you have a roadmap somewhere on planned features?

[...]

PR #5895, opened 2022-09-28
Native MQTT

ansd:
## Concept

Until today MQTT (like STOMP and AMQP 1.0) is proxied via AMQP 0.9.1:
Screenshot 2023-01-17 at 09 21 27

Pros:
* neat mechanism to easily extend RabbitMQ’s traditional AMQP 0.9.1 core model with support for more protocols

Cons:
* in case of MQTT: >20 Erlang processes per incoming MQTT connection
* poor scaling with many MQTT connections

This PR implements **Native MQTT**. By "native" we mean that MQTT becomes a first class protocol in RabbitMQ that does not get proxied via AMQP 0.9.1:
Screenshot 2023-01-17 at 09 21 39

Pros:
* Only 1 Erlang process per incoming MQTT connection
* Less memory usage => better scalability => can handle more MQTT connections

Native MQTT is possible thanks to the [queue_type interface](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_queue_type.erl) introduced by @kjnilsson 2 years ago. The queue_type interface decouples the [AMQP 0.9.1 channel](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_channel.erl) from queues (classic queues, quorum queues, or streams) making it possible to have the MQTT connection process publish to (and receive from) queue processes directly (without the need to proxy via the AMQP 0.9.1 channel).

## Result

The result of this PR (Native MQTT) is that resource usage drops drastically and that RabbitMQ is able to accept millions of MQTT connections.

Scaling tests can be found in https://github.com/rabbitmq/mqtt-testing (private repo).

Connecting in total 1 million "background" MQTT connections (i.e. neither publishing nor consuming, but only sending PING packets) to a 3 node RabbitMQ cluster with small buffer sizes and disabling management metrics:
```
mqtt.tcp_listen_options.sndbuf  = 2048
mqtt.tcp_listen_options.recbuf  = 2048
```
requires in RabbitMQ 3.11 97 GB of memory per node and with this PR 7 GB of memory per node. This is a permanent memory saving of (90GB * 3) 270 GB for the cluster and a memory reduction of factor 13. (For only QoS 0 subscribers using the new queue type - more details below - the memory reduction factor will even be much higher.)

93% of the memory required on `main` (or 3.11) is process memory, meaning memory required for Erlang processes.

Further scaling tests have shown that with this PR,
* we can connect more than 9.3 million MQTT connections across a 3 node RabbitMQ cluster requiring ~50 GB of memory per node
* 3 million MQTT connections in total across a 3 node cluster out of which 1.5 million are publishing and 1.5 million are receiving via 1:1 topology (i.e. each publisher sends to exactly one consumer) 1 message of size 64 bytes every 3 minutes with QoS 1 works well. This requires ~49 GB and ~19 CPUs per node. In this scenario there are 1.5 million classic queues v2 created where https://github.com/rabbitmq/rabbitmq-server/pull/6684 causes multiple GBs of memory savings.

## Implementation details
* Message routing and authentication / authorisation checks are now happening in the MQTT connection process instead of the channel process
* The queue_type interface is used directly by the MQTT connection process
* There is a single Erlang process per incoming MQTT connection. Most Erlang processes got automatically removed once the proxying via AMQP 0.9.1 was omitted. However, more work was needed to get rid of the `rabbit_hearbeat`, heartbeat supervisor and rabbit_mqtt_connection_sup procsses. The same is done for Web MQTT: There is now a single Erlang process per incoming Web MQTT connection.
* Heartbeats / MQTT Keepalives are now implemented in-place in the single MQTT connection process using Erlang timers.
* The MQTT parser got optimised preventing binaries converted to lists back and forth
* This PR contains many bug fixes for bugs that still exist on 3.11.x (see commit messages for details)
* This PR contains MQTT 3.1.1 features that were previously not implemented (see commit messages for details)
* A new queue type has been introduced. In this PR it's called `rabbit_mqtt_qos0_queue`. (We can rename it if we think it we can re-use it somewhere else.) The idea is that this queue type is a "pseudo queue" or "virtual queue" as for example experimented in https://github.com/rabbitmq/rabbitmq-server/tree/virtual-queue-type for direct-reply-to. The queue process is the receiving MQTT connection process. The queue is basically the Erlang mailbox of that process. It will be used for MQTT connections that connect with clean session and subscribe with QoS 0. There is no point in forwarding a message to a real queue first. Instead, MQTT publishing connections send a message directly to the new queue type - that is to the receiving MQTT connection process.
* The new queue type is also interoperable with other protocols, e.g. you can send via AMQP 0.9.1 to the new queue type. (You can even send a message from the Management UI directly to the receiving MQTT connection process.)
* The new queue type lives in `rabbitmq_mqtt`. Note that the `rabbit` app is not aware that this queue type exists.
* The benefit of the new queue type is (1) 1 message hop less, (2) huge reduction in memory usage because we can omit 2 Erlang processes (1 for the classic queue, 1 for its supervisor), (3) its client does not maintain any state when sending to the destination meaning it simplifies large fanouts. Sending also uses `delegate` as done in classic queues. The new queue type comes behind the feature flag `rabbit_mqtt_qos0_queue`.
* The new queue type does not have any form of flow control because no state can be maintained between sending and receiving process. Therefore it drops messages if it becomes overloaded: specifically if its Erlang mailbox has more than 1k messages configurable via `mqtt_mailbox_soft_limit` AND the network to the MQTT client is congested meaning the MQTT client app cannot consume messages fast enough.
* The new queue type can be listed via CLI or Management UI, but client apps are not aware that it exists.
* The Ra cluster `mqtt_node` is deleted (hidden behind feature flag `delete_ra_cluster_mqt_node`). Instead, MQTT client ID to Pid tracking is done via a local-only process group using module  [pg](https://www.erlang.org/doc/man/pg.html) (thanks to @lhoguin for coming up with this idea). The Ra cluster's procsses grew huge with many MQTT connections and became a bottleneck when mass disconnecting clients. Local-only `pg` scales better. For the purpose of MQTT client ID tracking (i.e. disconnect the old client if a new client with the same ID connects), we don't need the strong Raft consistency guarantees. This does mean however, that - in the presence of lost messages or network partitions - it can happen that 2 clients connect with the same ID (which shouldn't be huge problem - but does violate the protocol spec temporarily in the presence of such misconfigured clients and network issues within the broker).
* The code has much higher test coverage compared to v3.11.x. There is a new test suite called `shared_SUITE` that is shared between MQTT and Web MQTT. This introduces only a **test** dependency from `rabbitmq_mqtt` to `rabbitmq_web_mqtt`.
* The code is much clearer now. The code on 3.11.x is hard to understand.
* Hundreds of lines of old copied `emqttc` library code are deleted from the Web MQTT tests. Instead, the `emqtt` web socket option is used.
* Apart from reducing Erlang processes, there are numerous memory optimisations done within the single MQTT connection process itself (see commit messages)
* (Only) existing global counters are implemented. (Note that we plan to introduce MQTT specific Prometheus metrics when we implement support for MQTT 5.0.)
* `rabbit_mqtt_reader` got converted from `gen_server2` to `gen_server` (thanks @lhoguin)
* Feature flags `stream_queue` and `classic_queue_type_delivery_support` (the latter introduced in v3.10.9 and v3.11.1) are required. (That's not a problem because https://github.com/rabbitmq/rabbitmq-server/pull/6810 present in 3.12 will already require to upgrade to 3.11.x first.)

## Limitations
Following limitations are already present today in v3.11, but are especially important to mention now that RabbitMQ can handle many more MQTT connections:
* Mass client disconnection and node draining still needs to be optimised. For example there are some bottlenecks such as [rabbit_alarm](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_alarm.erl) and our old custom [pg_local](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/pg_local.erl) (which could be replaced by `pg`?) Nodes might even crash.
* Turning on management metrics collection `management_agent.disable_metrics_collector = false` (the default) will double or triple memory usage. The management plugins is not designed to handle such a huge number of stats emitting objects (connections / queues). Management metric collection should therefore be turned off and Prometheus should be used. We don't plan to improve management agent metrics in the future. Prometheus is the way to go.
* Huge fanouts are still not optimised and require lots of memory (although the new queue type helps somewhat for QoS 0 subscribers). Large fanouts should be done rarely. If sending to a target classic or quorum queue, they should additionally originate from only few publishers (because every publisher will maintain state to every destination queue).

Newly introduced limitations with this PR:
* As observed by @mkuratczyk: [channel specific fine grained metrics](https://github.com/rabbitmq/rabbitmq-server/blob/c6f38702b505f7466c2f8c60aa13a873d975b4d6/deps/rabbit/src/rabbit_channel.erl#L209-L224) are not emitted anymore for MQTT connections. This means for example users won't see in the Management UI how many messages were sent to the topic exchange. However, Prometheus global counters are supported.

## Future Work
(Prioritisation not yet decided.)
* [Support for MQTT 5.0](https://github.com/rabbitmq/rabbitmq-server/issues/2554) This PR is a prerequisite.
* Support huge fanouts: for example by sending to a local stream that asynchronously dispatches to local queues; also routing needs to be optimised, ideally streamed meaning not all million destination queue names being fetched in memory at once and sent to at once.
* more performant topic routing (prototype already implemented by @dcorbacho and @the-mikedavis for Khepri)
* handle mass client disconnection and node drain with millions of MQTT connections

Thanks to @ChunyiLyu who equally contributed to this PR!


lhoguin:
I do not see any obvious problem with the Web MQTT changes.

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -521,59 +542,43 @@ update_msg_status(confirm, Pid, #msg_status{pending = P,
...
 
 send_rejection(Pid, QName, MsgSeqNo) ->
-    case rabbit_queue_type:is_supported() of
-        true ->
-            gen_server:cast(Pid, {queue_event, QName,
-                                  {reject_publish, MsgSeqNo, self()}});
-        false ->
-            gen_server2:cast(Pid, {reject_publish, MsgSeqNo, self()})
-    end.
+    Msg = case rabbit_feature_flags:is_enabled(no_queue_name_in_classic_queue_client) of
```
kjnilsson, line 565: As discussed with @ansd rather than changing the message format we could amend the queue type callbacks involved with the stateful operation to also take the queue name record as an argument. This way we don't need to maintain the extra queue name (which uses memory for known but obscurely technical reasons with how maps work) in the queue type state (as it is used in the queue type state map as the key)

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -5,651 +5,855 @@
...
+                    ConnectionName :: binary(),
+                    SendFun :: fun((mqtt_packet(), state()) -> any()),
+                    PeerAddr :: inet:ip_address()) ->
+    state().
+initial_state(Socket, ConnectionName, SendFun, PeerAddr) ->
+    Flow = case rabbit_misc:get_env(rabbit, mirroring_flow_control, true) of
+               true   -> flow;
+               false  -> noflow
+           end,
+    #state{socket         = Socket,
```
kjnilsson, line 104: I think it would be lovely when making this many changes to a module to try to move away from the hard to read redundant whitespace formatting style.... :)

kjnilsson (changes requested):

deps/rabbitmq_mqtt/src/rabbit_mqtt_reader.erl
```diff
@@ -24,19 +24,21 @@
...
 -record(state,
-        {socket,
-         proxy_socket :: undefined | {rabbit_proxy_soket, any(), any()},
+        {socket :: rabbit_net:socket(),
+         proxy_socket :: option({rabbit_proxy_socket, any(), any()}),
          await_recv :: boolean(),
-         deferred_recv :: undefined | binary(),
-         parse_state,
+         deferred_recv :: option(binary()),
+         parse_state :: atom(),
```
ansd, line 37: The `parse_state` can also be a tuple in https://github.com/rabbitmq/rabbitmq-server/blob/1fd4a6d353bd57793d38f7640bdfce1c6e6ada64/deps/rabbitmq_mqtt/src/rabbit_mqtt_frame.erl#L32 . Ideally, we should also declare the parser state to be opaque.

deps/rabbitmq_mqtt/src/rabbit_mqtt_reader.erl
```diff
@@ -24,19 +24,21 @@
...
          proc_state :: rabbit_mqtt_processor:state(),
          connection_state :: running | blocked,
          conserve :: boolean(),
-         stats_timer,
+         stats_timer :: option(reference()),
```
ansd, line 41: The `stats_timer` is a record returned in https://github.com/rabbitmq/rabbitmq-server/blob/1fd4a6d353bd57793d38f7640bdfce1c6e6ada64/deps/rabbit_common/src/rabbit_event.erl#L95-L96 . Ideally, we should also declare the stats_timer to be opaque.

deps/rabbit/src/rabbit_queue_type.erl
```diff
@@ -148,20 +137,33 @@
...
 -callback purge(amqqueue:amqqueue()) ->
     {ok, non_neg_integer()} | {error, term()}.
 
 -callback policy_changed(amqqueue:amqqueue()) -> ok.
 
-%% stateful
+-callback is_stateful() -> boolean().
+
+%% stateful callbacks are optional
+-optional_callbacks([init/1,
```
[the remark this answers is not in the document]
ansd: Done

deps/rabbit/src/rabbit_classic_queue.erl
ansd: I removed the unnecessary feature flag and all queue names in the queue clients.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -5,651 +5,855 @@
...
+                    }).
+
+-record(info, {prefetch :: non_neg_integer(),
+               host :: inet:ip_address(),
+               port :: inet:port_number(),
+               peer_host :: inet:ip_address(),
+               peer_port :: inet:port_number(),
+               connected_at :: pos_integer()}).
+
+-record(state,
```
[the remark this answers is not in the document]
ansd: Thanks, I added a nested `#cfg{}` record.
The number of fields in the `rabbit_mqtt_processor` `state` decreased from 25 to 10.

deps/rabbitmq_mqtt/src/rabbit_mqtt_reader.erl
```diff
@@ -7,146 +7,165 @@
...
 
--include_lib("amqp_client/include/amqp_client.hrl").
 -include("rabbit_mqtt.hrl").
 
--define(SIMPLE_METRICS, [pid, recv_oct, send_oct, reductions]).
--define(OTHER_METRICS, [recv_cnt, send_cnt, send_pend, garbage_collection, state]).
+-define(HIBERNATE_AFTER, 1000).
+-define(PROTO_FAMILY, 'MQTT').
+
+-record(state,
```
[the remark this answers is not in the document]
ansd: I just did it for `rabbit_mqtt_processor` where the number of fields dropped from 26 to 11.
I think I would leave it as-is for `rabbit_mqtt_reader` because it has "only" 13 fields, if that's okay. Splitting also causes access to become more expensive.

deps/rabbitmq_mqtt/src/rabbit_mqtt_packet.erl
```diff
@@ -7,19 +7,23 @@
...
+
 -define(RESERVED, 0).
 -define(MAX_LEN, 16#fffffff).
 -define(HIGHBIT, 2#10000000).
 -define(LOWBITS, 2#01111111).
 
+-spec initial_state() -> state().
 initial_state() -> none.
 
+-spec parse(binary(), state()) -> {more, state()} | {ok, any(), any()} | {error, any()}.
```
ansd: Instead of `{ok, any(), any()}` we should be more specific here.

deps/rabbitmq_mqtt/src/rabbit_mqtt_packet.erl
```diff
@@ -7,19 +7,23 @@
...
--export([parse/2, initial_state/0]).
--export([serialise/2]).
-
 -include("rabbit_mqtt_packet.hrl").
 -include("rabbit_mqtt.hrl").
 
+-export([parse/2, initial_state/0, serialise/2]).
+-export_type([state/0]).
+
+-opaque state() :: 'none' | {more, any()}.
```
ansd: Instead of `any()` we could be more specific here.
Also, `more` is not part of the state. The state is either `none` or a `fun`.

deps/rabbitmq_mqtt/src/rabbit_mqtt_reader.erl
```diff
@@ -7,146 +7,165 @@
...
+        {socket,
```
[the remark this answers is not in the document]
ansd: missing type defs in `rabbit_mqtt_reader` are now added

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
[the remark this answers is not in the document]
ansd: missing type defs in `rabbit_mqtt_processor` are now added

deps/rabbitmq_mqtt/src/rabbit_mqtt_reader.erl
```diff
@@ -176,240 +195,206 @@ handle_info({bump_credit, Msg}, State) ->
...
-    case MqttConn of
-        undefined  -> rabbit_log_connection:debug(Fmt, Args);
-        _          -> rabbit_log_connection:info(Fmt, Args)
+              State = #state{conn_name  = ConnName,
+                             received_connect_packet = Connected}) ->
+    Fmt = "MQTT connection ~p will terminate because peer closed TCP connection",
+    Args = [ConnName],
+    case Connected of
+        true -> ?LOG_INFO(Fmt, Args);
+        false  -> ?LOG_DEBUG(Fmt, Args)
```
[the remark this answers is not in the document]
ansd: fixed

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
ansd: Agreed :) I will run a formatter on the entire `rabbitmq_mqtt` and `rabbitmq_web_mqtt` plugins because I don't want to manually move whitespaces around :)
I will do so in a separate PR because this PR becomes hard to diff and hard to rebase otherwise.

deps/rabbitmq_mqtt/src/rabbit_mqtt_reader.erl
ansd: done

deps/rabbitmq_mqtt/src/rabbit_mqtt_reader.erl
ansd: done

deps/rabbitmq_mqtt/include/mqtt_machine.hrl
```diff
@@ -5,9 +5,12 @@
 %% Copyright (c) 2020-2023 VMware, Inc. or its affiliates.  All rights reserved.
 %%
 
--record(machine_state, {client_ids = #{},
-                        pids = #{},
-                        %% add acouple of fields for future extensibility
-                        reserved_1,
-                        reserved_2}).
+-record(machine_state, {
```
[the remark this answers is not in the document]
ansd: I will run a formatter in a separate PR.

deps/rabbitmq_mqtt/src/rabbit_mqtt_reader.erl
```diff
@@ -7,146 +7,165 @@
...
 -include("rabbit_mqtt.hrl").
 
--define(SIMPLE_METRICS, [pid, recv_oct, send_oct, reductions]).
--define(OTHER_METRICS, [recv_cnt, send_cnt, send_pend, garbage_collection, state]).
+-define(HIBERNATE_AFTER, 1000).
+-define(PROTO_FAMILY, 'MQTT').
+
+-record(state,
+        {socket,
+         proxy_socket :: undefined | {rabbit_proxy_soket, any(), any()},
```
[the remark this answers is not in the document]
ansd: fixed

deps/rabbitmq_mqtt/src/rabbit_mqtt_packet.erl
ansd: fixed

deps/rabbitmq_mqtt/src/rabbit_mqtt_packet.erl
ansd: fixed

deps/rabbitmq_mqtt/src/rabbit_mqtt.erl
```diff
@@ -57,6 +58,7 @@ emit_connection_info_all(Nodes, Items, Ref, AggregatorPid) ->
             rabbit_control_misc:await_emitters_termination(Pids)
     end.
 
+-spec emit_connection_info_local([atom()], reference(), pid()) -> ok.
```
ansd, line 61: Could you the alias `rabbit_types:info_keys()` instead of `[atom()]`

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -1329,9 +1337,13 @@ serialise_and_send_to_client(Packet, #state{cfg = #cfg{proto_ver = ProtoVer,
                          [Sock, Error, Packet#mqtt_packet.fixed, Packet#mqtt_packet.variable])
     end.
 
+-spec serialise(#mqtt_packet{}, state()) ->
+    iodata().
 serialise(Packet, #state{cfg = #cfg{proto_ver = ProtoVer}}) ->
     rabbit_mqtt_packet:serialise(Packet, ProtoVer).
 
+-spec terminate(any(), binary(), atom(),rabbit_mqtt_processor:state()) ->
```
ansd, line 1345: `boolean()` instead of `any()` to be more specific

ansd: can just use `state()` instead of `rabbit_mqtt_processor:state()`?

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -1796,6 +1812,7 @@ throttle(Conserve, Connected, #state{queues_soft_limit_exceeded = QSLE,
     not sets:is_empty(QSLE) orelse
     credit_flow:blocked().
 
+-spec info(info_commands(), state()) -> any.
```
ansd, line 1815: `any()` instead of `any`. The latter is a specific atom.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -87,6 +87,14 @@
...
 -opaque state() :: #state{}.
 
+-type info_commands() :: host | port | peer_host | peer_port
+    | connected_at | ssl_login_name | vhost | user_who_performed_action | user
+    | clean_sess | will_msg | retainer_pid | exchange | prefetch
+    | messages_unconfirmed | messages_unacknowledged
+    | node
+    | client_id | client_properties
+    | channel_max | frame_max | auth_mechanism.
+
```
ansd, line 97: I would delete this whole block and instead simply use `rabbit_types:info_key()` for the function spec in `rabbit_mqtt_processor:info/1` because this is a long block and doesn't add much benefit from my point of view.

deps/rabbitmq_web_mqtt/src/rabbit_web_mqtt_handler.erl
```diff
@@ -28,17 +28,19 @@
          upgrade/5,
          takeover/7]).
 
+-type option(T) :: undefined | T.
+
 -record(state, {
-          socket,
+          socket :: rabbit_net:socket(),
```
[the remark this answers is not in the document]
ChunyiLyu: Thanks. Will update to `{rabbit_proxy_socket, any(), any()} | rabbit_net:socket()`

deps/rabbitmq_mqtt/src/rabbit_mqtt_util.erl
```diff
@@ -145,6 +146,8 @@ coerce_env_value(exchange, Val)     -> rabbit_data_coercion:to_binary(Val);
 coerce_env_value(vhost, Val)        -> rabbit_data_coercion:to_binary(Val);
 coerce_env_value(_, Val)            -> Val.
 
+-spec table_lookup([tuple()],  binary()) ->
```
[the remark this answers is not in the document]
ChunyiLyu: Thanks. Will update to `-spec table_lookup(rabbit_framing:amqp_table() | undefined,  binary()) ->
    tuple() | undefined.`

ChunyiLyu (approved):
🚀

kjnilsson (approved):

janiu-001:
@ChunyiLyu would you help to provide a doc on how to get the mqtt connect/disconnect notification? And when will you push the offical docker image to docker hub?

ansd:
Hi @janiu-001 

> would you help to provide a doc on how to get the mqtt connect/disconnect notification?

There is a doc in https://www.rabbitmq.com/event-exchange.html . For each MQTT CONNECT / DISCONNECT, an event `connection.created` and `connection.closed` gets emitted.

> And when will you push the offical docker image to docker hub?

Yesterday, [3.12.0-beta.1](https://github.com/rabbitmq/rabbitmq-server/releases/tag/v3.12.0-beta.1) was released. You can try out Native MQTT by using Docker image [pivotalrabbitmq/rabbitmq:v3.12.0-beta.1-otp-max-bazel](https://hub.docker.com/layers/pivotalrabbitmq/rabbitmq/v3.12.0-beta.1-otp-max-bazel/images/sha256-4ea93ea86ffc5a882b79d95aca0fe96ad0ea6f9f04d2ec44f822a404851e75e7?context=explore). The official Docker image `rabbitmq/3.12.0-rc.1-management` will be available once the 3.12 RC1 is released (happening probably end of March / beginning of April). 

We are looking forward to your feedback. Thanks!

janiu-001:
@ansd Thanks very much for you response, i tried the image pivotalrabbitmq/rabbitmq:v3.12.0-beta.1-otp-max-bazel, it failed to load the definitions.json which create the queue and bindings! But it could work with rabbitmq:3.11-management.

janiu-001:
> Thanks @janiu-001.
> 
> > But it failed to load the definitions.json which create the queue and bindings!
> 
> What is the error message? Are you trying to export queues created by the MQTT plugin in 3.11 and re-import those in 3.12? If you could provide step by step instructions that reproduce this issue and open a separate GitHub issue if you think something doesn't work as expected, that would be great!

1. the definition.json
image

2. load the definition.json file 
image

3. mount the definition.json file

image

4. check the queue on the dashboard

### with the 3.12 image

image

### with the rabbitmq:3.11-management image, we could see the queues defined in the definition.json
image

lukebakken:
@janiu-001 the best and fastest way to assist us is to **attach** your **complete** `definitions.json` file to your next response. Feel free to edit the file to change sensitive data, but keep it as original as possible.

janiu-001:
{
  "rabbit_version": "3.11.2",
  "rabbitmq_version": "3.11.2",
  "product_name": "RabbitMQ",
  "product_version": "3.11.2",
  "users": [
    {
      "name": "device",
      "password_hash": "kUpnMHYRMANoCXHfIw7NRQoHDJr20lEx+zx60TJKDmGEkJ9d",
      "hashing_algorithm": "rabbit_password_hashing_sha256",
      "tags": ["device"],
      "limits": {}
    },
    {
      "name": "guest",
      "password_hash": "STmaB2Mo0rSImlgTW4Ejj7MejhAzWGb6laMysaUyUGt5vdk3",
      "hashing_algorithm": "rabbit_password_hashing_sha256",
      "tags": ["administrator"],
      "limits": {}
    }
  ],
  "vhosts": [{ "name": "/" }],
  "permissions": [
    {
      "user": "guest",
      "vhost": "/",
      "configure": ".*",
      "write": ".*",
      "read": ".*"
    },
    {
      "user": "device",
      "vhost": "/",
      "configure": "^([amq.topic].*)$",
      "write": "^([amq.topic].*)$",
      "read": "^([amq.topic].*)$"
    }
  ],
  "topic_permissions": [
    {
      "user": "device",
      "vhost": "/",
      "exchange": "amq.topic",
      "write": "^user\\.(.*)\\.(state|(telemetry\\.(.*)))$",
      "read": "^user\\.(.*)\\.(config|command)$"
    }
  ],
  "parameters": [],
  "global_parameters": [
    {
      "name": "internal_cluster_id",
      "value": "rabbitmq-cluster-id-MpMhYRbJSd_6LInkcy8Tcg"
    }
  ],
  "policies": [],
  "queues": [
    {
      "name": "*.device.online.status",
      "vhost": "/",
      "durable": true,
      "auto_delete": false,
      "arguments": {}
    },
    {
      "name": "*.user.events",
      "vhost": "/",
      "durable": true,
      "auto_delete": false,
      "arguments": {}
    },
    {
      "name": "*.user.events.state",
      "vhost": "/",
      "durable": true,
      "auto_delete": false,
      "arguments": {}
    },
    {
      "name": "*.user.events.telemetry",
      "vhost": "/",
      "durable": true,
      "auto_delete": false,
      "arguments": {}
    },
    {
      "name": "*.user.events.analytics",
      "vhost": "/",
      "durable": true,
      "auto_delete": false,
      "arguments": {}
    },
    {
      "name": "*.user.events.eventsproto",
      "vhost": "/",
      "durable": true,
      "auto_delete": false,
      "arguments": {}
    },
    {
      "name": "*.user.events.records",
      "vhost": "/",
      "durable": true,
      "auto_delete": false,
      "arguments": {}
    },
    {
      "name": "*.user.command",
      "vhost": "/",
      "durable": true,
      "auto_delete": false,
      "arguments": {}
    }
  ],
  "exchanges": [],
  "bindings": [
    {
      "source": "amq.rabbitmq.event",
      "vhost": "/",
      "destination": "*.device.online.status",
      "destination_type": "queue",
      "routing_key": "*",
      "arguments": {}
    },
    {
      "source": "amq.topic",
      "vhost": "/",
      "destination": "*.user.events",
      "destination_type": "queue",
      "routing_key": ".*.*.user.events",
      "arguments": {}
    },
    {
      "source": "amq.topic",
      "vhost": "/",
      "destination": "*.user.events.state",
      "destination_type": "queue",
      "routing_key": ".*.*.user.state",
      "arguments": {}
    },
    {
      "source": "amq.topic",
      "vhost": "/",
      "destination": "*.user.events.telemetry",
      "destination_type": "queue",
      "routing_key": ".*.*.user.events.telemetry",
      "arguments": {}
    },
    {
      "source": "amq.topic",
      "vhost": "/",
      "destination": "*.user.events.analytics",
      "destination_type": "queue",
      "routing_key": ".*.*.user.events.analytics",
      "arguments": {}
    },
    {
      "source": "amq.topic",
      "vhost": "/",
      "destination": "*.user.events.eventsproto",
      "destination_type": "queue",
      "routing_key": ".*.*.user.events.eventsproto",
      "arguments": {}
    },
    {
      "source": "amq.topic",
      "vhost": "/",
      "destination": "*.user.events.records",
      "destination_type": "queue",
      "routing_key": ".*.*.user.events.records",
      "arguments": {}
    },
    {
      "source": "amq.topic",
      "vhost": "/",
      "destination": "*.user.command",
      "destination_type": "queue",
      "routing_key": ".*.*.user.command",
      "arguments": {}
    }
  ]
}

janiu-001:
Sorry for the troubles. Since github do not support to upload json file, i change the suffix from json to text

[definitions.txt](https://github.com/rabbitmq/rabbitmq-server/files/10910520/definitions.txt)

michaelklishin:
This belongs to a discussion.

lukebakken:
@michaelklishin that would be ideal but @janiu-001 responded to a closed PR 🤷‍♂️ I'm going to check this out today.

lukebakken:
Yep, something is up with definitions import - https://github.com/rabbitmq/rabbitmq-server/issues/7532

lukebakken:
@janiu-001 thanks for reporting the issue. In the future, please either start a discussion or file an issue rather than replying to a closed PR. That's not just Team RabbitMQ's preference but is pretty standard practice. Thanks!

PR #5945, opened 2022-09-30
Grafana: make sure dashboards do not break when detailed metrics are used

luos:
If detailed metrics are pulled into the same prometheus, then we get an error in Grafana:

```
execution: many-to-many matching not allowed:
matching labels must be unique on one side
```

This is because both endpoints provide the same `rabbit_identity_info` which causes the join to return multiple rows.

I think this is an non-breaking change. I did a test and all metrics seems to be working as expected.

Alternative solution would be to add an endpoint name to the identity info: 

```
rabbitmq_identity_info{endpoint="summary|per-object|detailed"}
```

Then add some filtering into the metrics.

I see the same solution was used here: 
https://github.com/rabbitmq/cluster-operator/pull/865/files#diff-0375b6730fa13ef8ae8767e09d2ff6138d29e8104fa73b9355379155db014a28R168

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments


baldarn:
Hi! this was a problem we were addressing too!

I have a feedback of the current dashboard with a cluster with only one node
Screenshot from 2022-10-03 11-11-32

as you can see the `nodes` counter shows 2, instead of 1, and the list of the `nodes` section shows the same node 2 times.

our approach to the problem was filtering the doubled metrics on the prometheus job, leaving the dashboard untouched.

an example of the prometheus.yml
```
  - job_name: "rabbitmq"
    static_configs:
      - targets: ["rabbitmq:15692"]

  - job_name: "rabbitmq-detailed"
    metrics_path: "/metrics/detailed"
    params:
      family: ["queue_coarse_metrics"]
    static_configs:
      - targets: ["rabbitmq:15692"]
    metric_relabel_configs:
      - source_labels: [ __name__ ]
        regex: 'rabbitmq_build_info'
        action: drop
      - source_labels: [ __name__ ]
        regex: 'rabbitmq_identity_info'
        action: drop

```

what do you think the best approach is?

coro:
There are several different approaches we could take here.

Option 1: we could modify the dashboards as this PR does to account for the duplicated metrics that are reported on the same `instance` from the `/metrics` and `/metrics/detailed` endpoints, making sure to modify the table that is currently reporting two rows for each node as @baldarn has found.

Option 2: we could instead document that if you are scraping from both endpoints, you will need to use relabel config to drop the superfluous information.

Option 3: reassessing whether the `/metrics/detailed` endpoint should return identity info at all. Theoretically, if you are scraping only the detailed endpoint, you need the identity info in order to filter by the name of the RabbitMQ cluster, since you aren't getting it from the basic metrics endpoint. Everything else on the detailed metrics endpoint is opt-in, so perhaps identity info should also be opt-in on that endpoint rather than by default.

In my opinion, the approach that leads to the best 'out of the box' experience is option 1 - a user doesn't need to know the specifics of what metrics are emitted by what endpoint and how they collide in order to scrape both endpoints.

In terms of this PR, I also tested it out and saw the same as @baldarn in terms of the duplicated nodes in the Overview table. It would also be great to add the detailed endpoint as a scrape target in [the example prometheus config](https://github.com/rabbitmq/rabbitmq-server/blob/cb380dfd6a3350f9ae98917fadd17e42f8372b6c/deps/rabbitmq_prometheus/docker/prometheus.yml#L37-L51) as part of this PR to make sure we don't miss things like this in the future, if that's OK. Otherwise, looks great, thank you for your contribution.

luos:
I just pushed the `prometheus.yml` with the detailed config, and I can confirm the node is visible twice. Not sure how to go about that?

ejf-coders51:
IMHO the Option 1 involve an hardcode job name value in the dashboard or a dropdown with all the job name  like namespace and cluster and all the future query will have to take care of this new variable. Another option could be have only one endpoint (only /detailed) and use the family filter with a value that return all the metrics that /metrics return or move the filter in /metrics.

stocksr:
Just to add an additional point in favor of removing or renaming the metrics in the detailed endpoint - I was getting 
"Error on ingesting out-of-order samples"   errors from prometheus that I tracked down to the rabbitmq_build_info and rabbitmq_identity_info metrics being duplicated

 thanks to this comment https://github.com/rabbitmq/rabbitmq-server/pull/5945#issuecomment-1265170855 I was able to update my service monitor config to drop the metrics from the detailed endpoint and the errors stopped
``` yaml
  endpoints:
    - port: prometheus
      scheme: http
      path: /metrics
    - port: prometheus
      scheme: http
      path: /metrics/detailed
      metricRelabelings:
        - sourceLabels: [__name__]
          regex: "rabbitmq_build_info" # Drop this metric from the detailed endpoint it is a duplicate from the basic one
          action: drop
        - sourceLabels: [__name__]
          regex: "rabbitmq_identity_info" # Drop this metric from the detailed endpoint it is a duplicate from the basic one
          action: drop
      params:
        family:
          - queue_coarse_metrics
          - queue_metrics
 
 
```

mkuratczyk:
This PR fell through the cracks but has now been merged and new revisions have been uploaded to https://grafana.com/orgs/rabbitmq/dashboards

michaelklishin:
Let's mark it as a "3.12.11" issue just because it's shipped around the same time.

PR #5968, opened 2022-10-03
Refactor how selenium suites are run

MarcialRosales:
## Proposed Changes

Before this PR, we only had 2 suites of selenium test cases. One which run against UAA and another which run when UAA was down. It was simple enough to run them one by one.

However, to fix an issue on the oauth2 part of the management plugin we need to add another suite. We needed to have a easier way to add more suites and to run them either locally or via github actions. 

This PR is about refactoring how Selenium suites are created and how they are run. 
The github workflow .github/workflows/test-selenium.yaml has been simplified a lot and we do not need to update it when we add more suites. 

## Types of Changes

- [X] Build system and/or CI


.github/workflows/test-selenium.yaml
```diff
@@ -6,7 +6,7 @@ on:
       - v3.11.x
       - bump-otp-for-oci
       - bump-rbe-*
-      - bump-rules_erlang
+      - bump-rules_erlang
```
pjk25, line 9: Looks like this introduces unnecessary whitespace

pjk25 (commented):
Overall this looks good to me.

PR #5984, opened 2022-10-04
Support management path prefix when using OAuth2

MarcialRosales:
## Proposed Changes

Support management plugin's [path prefix](https://www.rabbitmq.com/management.html#path-prefix) when OAuth 2.0 authentication is enabled.  

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/oidc-oauth/helper.js
+++ b/deps/rabbitmq_management/priv/www/js/oidc-oauth/helper.js
@@ -4,8 +4,9 @@ var _management_logger;
 
 function oauth_initialize_if_required() {
     rabbit_port = window.location.port ? ":" +  window.location.port : ""
+    rabbit_path_prefix = window.location.pathname.replace(/(\/js\/oidc-oauth\/.*$|\/+$)/, "")
     rabbit_base_uri = window.location.protocol + "//" + window.location.hostname
-      + rabbit_port + rabbit_path_prefix()
+      + rabbit_port + rabbit_path_prefix
 
     var request = new XMLHttpRequest();
     request.open("GET", rabbit_base_uri + "/api/auth", false);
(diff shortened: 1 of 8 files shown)


deps/rabbitmq_management/priv/www/js/oidc-oauth/helper.js
```diff
@@ -4,18 +4,24 @@ var _management_logger;
...
         return oauth_initialize(JSON.parse(request.responseText));
-    }else {
+    } else {
         return { "enabled" : false };
     }
 
 }
+function rabbit_path_prefix() {
+  paths = window.location.pathname.split("/")
+  return paths[1] === "js" || paths[1] === "" ? "" : "/" + paths[1]
```
CAPCHIK, line 22: Why only first part is used? According to [documentation](https://www.rabbitmq.com/management.html#path-prefix) `anagement.path_prefix setting allows an arbitrary prefix to be set`, but that code for that pull request page returns only `/rabbitmq`, but should get all path
image

michaelklishin: The prefix is not limited to a single segment, that's correct.

michaelklishin:
@CAPCHIK this is ready for another round, let us know if this now LGTY.

deps/rabbitmq_management/priv/www/js/oidc-oauth/helper.js
```diff
@@ -4,18 +4,22 @@ var _management_logger;
 
 function oauth_initialize_if_required() {
     rabbit_port = window.location.port ? ":" +  window.location.port : ""
-    rabbit_base_uri = window.location.protocol + "//" + window.location.hostname + rabbit_port
+    rabbit_path_prefix = window.location.pathname.replace(/(\/js\/oidc-oauth\/.*$|\/+$)/, "")
```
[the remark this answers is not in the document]
MarcialRosales: @CAPCHIK For now the path I have chosen is to document it. I wanted to keep the oidc callback onto separate html pages rather than having to handle them as sammy routes, i.e. with `#/js/...`. 

Once again thanks for reviewing and excellent feedback !

CAPCHIK:
@michaelklishin as a result, it was decided that path_prefix should not **contain** `/js/oidc-oauth/`, not **end** in `/js/oidc-oauth/`?

michaelklishin:
I doubt it would make any practical difference. If you really care about the difference, please submit a PR that would only check the final segments.

PR #6015, opened 2022-10-04
OAuth 2.0: support IDP-initiated login in the management UI

MarcialRosales:
## Proposed Changes

Support "Identity Provider initiated login" workflow whereby a user comes to RabbitMQ with an access_token rather than coming without any and being redirected to the authorization server to get one. The http endpoint to login directly with an access token is  *POST* `/login` with the token as the value of the form field `access_token` ([here](https://github.com/rabbitmq/rabbitmq-server/blob/oidc_idp_initiated_login/deps/rabbitmq_management/selenium/fakeportal/views/rabbitmq.html) is the sample form used by our automated testing).

This PR must take into account the following flows : 
- login, 
- logout, 
- session expiry and 
- token expiry

It requires the following oauth2 related configuration settings in the management plugin : 
- oauth_enabled 
- oauth_provider_url : This is the URL of the idp where to redirect the user when the user clicks on the button "Click here to login" which is the landing page when Oauth2 is enabled. Although the idp may not necessarily be an oauth 2.0 provider but in practice, it is ultimately the entity responsible of granting token to users and redirecting them to RabbitMQ.
- oauth_initiated_logon_type : This new setting configures the type of logon which can be one of these two: 
      - service-provider initiated logon (`sp_initiated` term) which is the default value and corresponds to **Oauth 2.0 authorization code flow**. 
      - identity-provider initiated logon (`idp_initiated` term)  

It has been considered to use this PR to merge another 2 PRs very much related to the same area of code as this PR. 
- https://github.com/rabbitmq/rabbitmq-server/pull/6201 - client_secret should be optional
- https://github.com/rabbitmq/rabbitmq-server/pull/6247 - configure preferred username claims 

There is a documentation PR linked to this PR: https://github.com/rabbitmq/rabbitmq-website/pull/1504

This PR introduces the following potential breaking change: With this PR the management UI no longer stores credentials (basic or token) on cookies. Instead it uses `window.localStorage`. Furthermore, and here it is the breaking change: if the window.localStorage is disabled, the management ui will not proceed with the login process. It will not fallback to cookies. This is because bearer tokens may be larger than 4Kb in length and browsers will not accept cookies whose value is greater than 4Kbytes.

## Types of Changes

- [X] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Build system and/or CI


deps/rabbitmq_auth_backend_oauth2/src/rabbit_auth_backend_oauth2.erl
```diff
@@ -542,24 +547,32 @@ token_from_context(AuthProps) ->
...
-        undefined ->
-            case ClientProvidedUsername of
-                undefined -> undefined;
-                <<>>      -> undefined;
-                _Other    -> ClientProvidedUsername
-            end;
-        Value     ->
-            Value
-    end.
+find_claim_in_token(Claim, Token) ->
```
michaelklishin, line 570: It is much more common to see `{ok, Value} | undefined` to be returned instead of `{true, Value} | false`.

MarcialRosales: The reason for returning `(boolean) | {true, Value}` is because this function is used as a filter function in [filtermap](https://www.erlang.org/doc/man/lists.html#filtermap-2) and it expects those return values....

PR #6039, opened 2022-10-06
Try out mirroring hex packages into our BCR fork

pjk25:
`rules_erlang` `3.8` includes a new `untar` rule, which allows adding packages from hex.pm directly into our fork of the bazel-central-repository. For now this branch is a test of that approach with rabbitmq-server.


pjk25:
I'm beginning to think that rather than relying on hosting erlang hex.pm packages in BCR or another bazel registry, it's better to invest in the gazelle extension and let it correctly generate the build for such packages when gazelle is run.

pjk25:
The gazelle extension has been released with rules_erlang 3.9.0, so I will close this and give that approach a try.

PR #6126, opened 2022-10-17
Partially revert delegate optimisation

### LINKED ISSUE #6115: Broker 3.10.0 and newer stops sending publisher confirms
### Discussed in https://github.com/rabbitmq/rabbitmq-server/discussions/6083

<div type='discussions-op-text'>

<sup>Originally posted by **bording** October 10, 2022</sup>
It appears that a bug has been introduced in 3.10.0 that causes the broker to not send publisher confirms when publishing messages to exchanges bound to mirrored classic queues and the names of the exchanges involved exceed some (unknown) length.

This problem was reported to us by a customer who ran into it while testing upgrading to 3.10.7 in their environment. I've been able to create a repro that just uses the .NET RabbitMQ client, without NServiceBus being involved:

https://github.com/bording/exchange-repro

I have included a docker compose file that creates a 3-node cluster and sets up a classic queue mirroring policy.

When the repro is run, it creates the necessary exchanges, queues, and bindings, Fills the `receiver` queue with 10,000 messages, and then consumes those messages.

For each message consumed, it attempts to publish some messages that get routed through the exchanges, and then waits for the publisher confirms.

Very quickly, you'll observe that the consumer stops processing messages, and you'll see that the channel used to publish the messages is showing that it has unconfirmed messages on it, which the broker never sends.

It's not clear to me what specifically is required to trigger the problem, but here is what I do know are factors:

**1. Length of the names of the exchanges**
If I shorten the names of the exchanges being used, then the problem is less likely to occur.

**2. Mirrored classic queues**
If the queues aren't mirrored, then this also doesn't seem to be a problem, regardless of exchange name length. It also doesn't trigger if quorum queues are used instead of classic queues. You can uncomment the line in the repro to observe this.

**3. 3.10.0 or newer**
When the docker compose file is changed to use 3.9.22 or older, the problem does not occur.

</div>

kjnilsson:
The changes introduced in https://github.com/rabbitmq/rabbitmq-server/pull/3865 can cause message arrival ordering guarantees
between two logical erlang process (sending messages via delegate) to
be violated as a message sent to a single destination can overtake a prior
message sent as part of a fan-out. This is due to the fact that the fan-out
take a different route via the delegate process than the direct delivery that
bypasses it.

This commit only reverses it for the `invoke_no_result/2|3` API and leaves the
optimisation in for the synchronous `invoke/` API. This means that the message
send ordering you expect between erlang processes still can be violated when
mixing invoke and invoke_no_result invocations. As far as I can see there are
no places where the code relies on this and there are uses of invoke (mgmt db)
that very well could benefit from avoiding the additional copying.

Fixes #6115

### DIFF
--- a/deps/rabbit_common/src/delegate.erl
+++ b/deps/rabbit_common/src/delegate.erl
@@ -169,9 +169,6 @@ demonitor(Ref) when is_reference(Ref) ->
 demonitor({Name, Pid}) ->
     gen_server2:cast(Name, {demonitor, self(), Pid}).
 
-invoke_no_result(Pid, FunOrMFA = {gen_server2, _F, _A}) when is_pid(Pid) ->
-    _ = safe_invoke(Pid, FunOrMFA), %% we don't care about any error
-    ok;
 invoke_no_result(Pid, FunOrMFA) when is_pid(Pid) andalso node(Pid) =:= node() ->
     %% Optimization, avoids calling invoke_no_result/3.
     %%
@@ -192,9 +189,6 @@ invoke_no_result(Pid, FunOrMFA) when is_pid(Pid) ->
     ok;
 invoke_no_result([], _FunOrMFA) -> %% optimisation
     ok;
-invoke_no_result([Pid], FunOrMFA = {gen_server2, _F, _A}) when is_pid(Pid) -> %% optimisation
-    _ = safe_invoke(Pid, FunOrMFA), %% must not die
-    ok;
 invoke_no_result([Pid], FunOrMFA) when node(Pid) =:= node() -> %% optimisation
     _ = safe_invoke(Pid, FunOrMFA), %% must not die
     ok;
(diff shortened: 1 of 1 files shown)


kjnilsson:
@bording do you think you could test this PR as well? It seems to work against your repro project. 

https://hub.docker.com/layers/pivotalrabbitmq/rabbitmq/partially-revert-delegate-opt-3865-otp-max-bazel/images/sha256-2d71e431855b1c28617e69c1405598e29d74cc030795bcb33772fb7d84157966?context=explore


Cheers

oskar:
@kjnilsson When this PR is accepted, roughly how long do you think it will be until there is a new patch version for 3.10?

I'm not familiar with the release cycles of rabbitmq-server but we've hit this bug (with NServiceBus) and it is causing us quite some issues, so we're really eager to patch our RabbitMQ cluster.

kjnilsson:
> @kjnilsson When this PR is accepted, roughly how long do you think it will be until there is a new patch version for 3.10?
> 
> I'm not familiar with the release cycles of rabbitmq-server but we've hit this bug (with NServiceBus) and it is causing us quite some issues, so we're really eager to patch our RabbitMQ cluster.

Once the PR is accepted (and backported) it will ship in the next patch. Patches are shipped when we feel it is necessary (you can look at prior intervals for example). I would imagine this fix would qualify for triggering a new patch sooner rather than later.

michaelklishin:
@oskar you can get an idea of how quickly RabbitMQ patch releases ship from [the changelog](https://www.rabbitmq.com/changelog.html). Sometimes we ship patch releases sooner, it really depends on what's in the release and how many users are affected, or how severely.

Issue #4999, opened 2022-06-08
A way to "pre-configure" limits for virtual hosts


michaelklishin:
At the moment virtual host limits can only be applied to existing virtual hosts. It would be
useful to allow for pre-defined limits. With a feature like this, it would be realistic
for RabbitMQ-as-a-Service environments to allow the user to create more virtual hosts while still
enforcing some limits. 

Suggested by @illotum.

What I have in mind is something like this:

``` ini
default_limits.vhosts.1.pattern = ^device
default_limits.vhosts.1.max_connections = 10
default_limits.vhosts.1.max_queues = 10

default_limits.vhosts.2.pattern = ^system
default_limits.vhosts.2.max_connections = 100
default_limits.vhosts.2.max_queues = -1

default_limits.vhosts.3.pattern = .*
default_limits.vhosts.3.max_connections = 20
default_limits.vhosts.3.max_queues = 20
```

defines three groups of limits that match virtual hosts by name. The first group to match will
be used to apply the limits right after the matching virtual host is declared.

michaelklishin:
Done in #6172 and #6191.

PR #6172, opened 2022-10-18
Add ability to set default vhost limits by pattern

illotum:
## Proposed Changes

A fairly straightforward implementation of #4999. Per ticket, vhost limit defaults can be set up with regex patterns in config, and the first matching is applied during vhost creation. Defaults come from a file and are not synced between cluster members.

Unlike the ticket, I do not allow negative limits in favor of omission.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

This change _is not_ documented, nor added to release-notes. LMK if this needs to be addressed.

Before I'm happy with it, several concerns and questions:

- I'll admit I'm confused about the two places to store VHost limits: the record and the runtime params. Is one deprecated?
- In my manual tests validators didn't error nor log anything on bad config inputs. Any advice?

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -632,6 +632,7 @@ end}.
 %% {default_user,        <<"guest">>},
 %% {default_pass,        <<"guest">>},
 %% {default_permissions, [<<".*">>, <<".*">>, <<".*">>]},
+%% {default_limits,      [{vhost, []}]
 
 {mapping, "default_vhost", "rabbit.default_vhost", [
     {datatype, string}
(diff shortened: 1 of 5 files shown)


illotum:
Right, I see I failed to follow the breadcrumbs deep enough. Let me fix the snippets.

illotum:
This time it feels like flaky tests.

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -725,18 +725,25 @@ fun(Conf) ->
                    [ ID || {ID, _} <- GetAll(["$id", "max_connections"]) ] ++
                    [ ID || {ID, _} <- GetAll(["$id", "max_queues"]) ]
                   ),
-    [rabbit_log:warning("No matching pattern for default_limits.vhost.~s", [ID])
-     || ID <- SettingIDs, Get([ID, "pattern"]) =:= undefined],
+    [ cuttlefish:invalid(
+        io_lib:format(
+           "No matching pattern for record '~ts'",
```
michaelklishin, line 730: "For record" seems to be a copy-and-paste artifact

deps/rabbit/src/rabbit_vhost.erl
```diff
@@ -224,6 +221,8 @@ do_add(Name, Metadata, ActingUser) ->
           fun (VHost1, true) ->
                   VHost1;
               (VHost1, false) ->
+                  rabbit_log:debug("Applying default limits to vhost '~tp': ~tp", [Name, DefaultLimits]),
```
michaelklishin, line 224: This arguably can be an `info` message.

deps/rabbit/test/config_schema_SUITE_data/rabbit.snippets
```diff
@@ -100,6 +100,38 @@ ssl_options.fail_if_no_peer_cert = true",
...
+                                       ]}]}]}],
+  []},
+
+ {default_limits_vhost_no_undefined,
+ "
+  default_limits.vhost.a.pattern = .*
+  default_limits.vhost.a.max_queues = 10
+ ",
+  [{rabbit, [{default_limits, [{vhost, [
+      {{re_pattern,0,0,0,
```
michaelklishin, line 126: Can we avoid matching against a record or at least compile it in the test using `re`? The record will change over time, it would be nice to not have to update this test every time it does.

michaelklishin (changes requested):

michaelklishin:
@rin and I have noticed a genuine failure on this branch:

``` shell

```

fails with

``` 
=== Ended at 2022-10-19 08:47:05
=== Location: [{rabbit_mgmt_rabbitmqadmin_SUITE,parameters,[319](file:///private/var/tmp/_bazel_antares/14f90f8a31f5e7ccfc4ded95d3ba96d8/execroot/__main__/bazel-out/darwin_arm64-fastbuild/testlogs/deps/rabbitmq_management/rabbit_mgmt_rabbitmqadmin_SUITE/test.outputs/outputs/ct_run.ct-rabbitmq_management-rabbit_mgmt_rabbitmqadmin_SUITE@sunnyside.2022-10-19_08.43.51/deps.rabbitmq_management.rabbit_mgmt_rabbitmqadmin_SUITE.logs/run.2022-10-19_08.43.51/rabbit_mgmt_rabbitmqadmin_suite.src.html#319)},
              {test_server,ts_tc,1782},
              {test_server,run_test_case_eval1,1291},
              {test_server,run_test_case_eval,1223}]
=== === Reason: no match of right hand side value 
                 {ok,[["test","good","/","123"],
                      ["vhost-limits","limits","/"]]}
  in function  rabbit_mgmt_rabbitmqadmin_SUITE:parameters/1 (rabbit_mgmt_rabbitmqadmin_SUITE.erl, line 319)
```

so the HTTP API response must have changed but the test hasn't.

michaelklishin:
There is also a CLI suite test failure:

``` shell
bazel test --config=local //deps/rabbitmq_cli:all
```

```
SetParameterCommandTest [test/ctl/set_parameter_command_test.exs]
  * test banner [L#133]
  * test banner (0.4ms) [L#133]
  * test run: an invalid component_name returns a validation failed error [L#88]
  * test run: an invalid component_name returns a validation failed error (1.8ms) [L#88]

  1) test run: an invalid component_name returns a validation failed error (SetParameterCommandTest)
     test/ctl/set_parameter_command_test.exs:88
     Assertion with == failed
     code:  assert list_parameters(context[:vhost]) == []
     left:  [[component: "vhost-limits", name: "limits", value: "[]"]]
     right: []
     stacktrace:
       test/ctl/set_parameter_command_test.exs:96: (test)

--max-failures reached, aborting test suite
Finished in 228.4 seconds (14.6s async, 213.7s sync)
642 tests, 1 failure
```

deps/rabbit/test/config_schema_SUITE_data/rabbit.snippets
illotum: I see no easy way to inject code into s test snippet (rabbit_ct_config_schema Plugins seems to be a stub?). But upon reflection re:compile is unnecessary optimization in this context. I'll revert to using binaries.

illotum:
Addressed comments, squashed & rebased.

michaelklishin:
The example in the commit uses `default_limits.vhosts.` but the actual schema uses `default_limits.vhost.` which has resulted some head scratching on my end.

I will rename "vhosts" to be plural because this is a collection.

michaelklishin:
I have also noticed that if there are multiple patterns that match, the effective one is picked undeterministically. I suspect that it depends on collection iteration ordering. There are
other places in RabbitMQ that works this way, namely policies with equal priorities.

illotum:
I could sort them, e.g. by the pattern hash, if you think it's worth it.

I feel the config file mismatch (between instances) makes this a minor issue by comparison.

illotum:
I had two extra questions in the description: most importantly, I was able to boot with config values that violate cuttlefish validators. Should I file a separate issue?

michaelklishin:
If you can reproduce this outside of RabbitMQ, then sure. Cuttlefish is at [Kyorai/cuttlefish](https://github.com/Kyorai/cuttlefish).

PR #6177, opened 2022-10-19
Add 3.11.2 to mixed version tests

ansd:
such that main branch will test against 3.11.2


ansd:
splitting into 2 PRs go keep the tests green.
1st one in https://github.com/rabbitmq/rabbitmq-server/pull/6178

PR #6183, opened 2022-10-19
QQ: don't try to contact non-connected nodes for stats

kjnilsson:
Some systems may incur a substantial latency penalty when attempting reconnections to down nodes so to avoid this some stat related functions that gather info from all QQ member nodes now only try those nodes that are connected. This should help keeping things like the mgmt API functions and ctl commands a bit more responsive when RabbitMQ nodes are down.

Context: https://github.com/rabbitmq/rabbitmq-server/discussions/6048

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -971,9 +971,8 @@ cleanup_data_dir() ->
              end
              || Q <- rabbit_amqqueue:list_by_type(?MODULE),
                 lists:member(node(), get_nodes(Q))],
-    NoQQClusters = rabbit_ra_registry:list_not_quorum_clusters(),
     Registered = ra_directory:list_registered(?RA_SYSTEM),
-    Running = Names ++ NoQQClusters,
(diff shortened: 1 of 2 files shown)


ansd (changes requested):

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1691,18 +1705,37 @@ prepare_content(Content) ->
...
     catch
         _:badarg ->
             Default
     end.
+
+erpc_call(Node, M, F, A, Timeout)
+  when is_integer(Timeout) andalso Node == node()  ->
+    %% Only timeout 'infinity' optimises the local call in OTP 23-25 avoiding a new process being spawned:
+    %% https://github.com/erlang/otp/blob/47f121af8ee55a0dbe2a8c9ab85031ba052bad6b/lib/kernel/src/erpc.erl#L121
+    try erpc:call(Node, M, F, A, Timeout) of
```
ansd, line 1720: We should pass in `infinity` here to optimise the local call.

PR #6191, opened 2022-10-19
Add ability to set default vhost limits by pattern (with a naming change)

michaelklishin:
This is #6172 by @illotum plus one naming change from me.


michaelklishin:
@illotum do you see anything I might have missed?

illotum:
LGTM. Snippet test case names perhaps, `default_limits_vhost` and `default_limits_vhost_no_undefined`.

PR #6225, opened 2022-10-24
Persist static configuration

ansd:
Put configuration that does not change during runtime, but is read often into persistent_term. This requires less CPU and memory as the flame graphs show.

Tested on https://github.com/rabbitmq/rabbitmq-server/tree/mqtt of 24 October 2022.

Start broker
```
 make run-broker TEST_TMPDIR="$HOME/scratch/rabbit/test" PLUGINS="rabbitmq_mqtt" RABBITMQ_CONFIG_FILE="$HOME/scratch/rabbit/rabbitmq.conf" RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+JPperf true +S 2 -mnesia dump_log_write_threshold=50000" RABBITMQ_MAX_NUMBER_OF_PROCESSES=134217727
```
where `rabbitmq.conf` defines
```
classic_queue.default_version = 2
mqtt.tcp_listen_options.sndbuf  = 2048
mqtt.tcp_listen_options.recbuf  = 2048
```
Run, then stop
```
./emqtt_bench sub -V 4 -c 25000 -t a/b -i 2 --prefix my_client_prefix --clean false --qos 1
```
This created 25k classic queues v2 bound to the topic exchange.

Next, while sending with large fan out as follows
```
./emqtt_bench pub -V 4 -c 1 -t a/b -I 10 --qos 0 --size 12
```
capture the different kind of flame graphs as follows:
```
sudo perf record --pid $(cat "test/rabbit@$(hostname --short)/rabbit@$(hostname --short).pid") --freq 9999 -g -- sleep 10
sudo perf record --pid $(cat "test/rabbit@$(hostname --short)/rabbit@$(hostname --short).pid") --event page-faults -g -- sleep 20
sudo perf record --pid $(cat "test/rabbit@$(hostname --short)/rabbit@$(hostname --short).pid") --event syscalls:sys_enter_mmap -g -- sleep 40
```

Before this PR, function `application:get_env/3` (called by rabbit_classic_queue_store_v2 and rabbit_classic_queue_index_v2)  causes:
* 7% of all CPU time
* 5.6% of all page faults
* 13% of all mmap system calls

CPU (zoomed):
Screenshot 2022-10-24 at 11 29 18

Page faults (zoomed):
Screenshot 2022-10-24 at 11 29 33

Memory maps:
Screenshot 2022-10-24 at 11 29 45

After this PR:
Page faults and memory maps for `application:get_env` and `persistent_term` is obviously 0.
CPU time for `persistent_term` is 0.6%.

I suggest we write static configuration that is never changed directly early in `rabbit.erl`. We can add there more stuff in the future if we discover that `application:get_env` is slow elsewhere.
Also we can't write that config in the queue supervisors because they in turn have vhost supervisors and we don't want to `persistent_term:put` when a new vhost is added.

### DIFF
--- a/deps/rabbit/src/rabbit.erl
+++ b/deps/rabbit/src/rabbit.erl
@@ -910,6 +910,8 @@ start(normal, []) ->
         ok = rabbit_feature_flags:refresh_feature_flags_after_app_load(
                Plugins),
 
+        persist_static_configuration(),
+
         ?LOG_DEBUG(""),
         ?LOG_DEBUG("== Boot steps =="),
 
@@ -1592,3 +1594,25 @@ ensure_working_fhc() ->
     after Timeout ->
             throw({ensure_working_fhc, {timeout, TestPid}})
     end.
+
+%% Any configuration that
+%% 1. is not allowed to change while RabbitMQ is running, and
+%% 2. is read often
+%% should be placed into persistent_term for efficiency.
+persist_static_configuration() ->
+    persist_static_configuration(
+      [{rabbit, classic_queue_index_v2_segment_entry_count},
+       {rabbit, classic_queue_store_v2_max_cache_size},
+       {rabbit, classic_queue_store_v2_check_crc32}
+      ]).
+
+persist_static_configuration(AppParams) ->
+    lists:foreach(
+      fun(Key = {App, Param}) ->
+              case application:get_env(App, Param) of
+                  {ok, Value} ->
+                      ok = persistent_term:put(Key, Value);
+                  undefined ->
+                      ok
+              end
+      end, AppParams).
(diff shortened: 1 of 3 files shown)


lhoguin (approved):

lhoguin:
I'm not sure the PR really does much about page faults or mmap, it's not the copying of the integers that do this it's the GC that may occur more often in this function because it's doing an ets call. I'm also not sure about the CPU improvement because again it's just that it triggers GC more often here instead of elsewhere. But I'm OK with the change regardless because persistent_term is slightly more efficient than ets.

ansd:
I did some more measurements. As Loïc correctly wrote (thanks for pointing this out), the flame graphs here are indeed misleading and the PR does not bring any significant performance improvements in terms of messages delivered to the queues.

PR #6228, opened 2022-10-24
AMQP 1.0: Support the modified outcome

### LINKED ISSUE #6121: Support for AMQPv1 modified outcome
We are using QPID with the AMQPv1 plugin which is the default JAVA JMS + RMQ combination.
When QPID uses the AMQPv1 "modified" outcome RMQ cuts the connection.
While there's no need for full modified state support the RabbitMQ should not cut off a consumer which use a standardized AMQPv1 frame combination.

# Description
Using amqpv1 with classic queues and a high message througput combined with TTL causes protocol errors with QPID. 
The default implementation of QPID uses the modified state in the disposition frame to tell RMQ that the TTL has expired after TRANSFER. 

OASIS AMQPv1 this state:

http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-modified

# Reproduction
1. Send disposition frame in modified state 
2. Use QPID consumer on high message througput with a short message TTL

# Error
```
{'v1_0.disposition',true,
                                     {uint,0},
                                     {uint,0},
                                     true,
                                     {'v1_0.modified',true,true,undefined},
                                     undefined} [condition = amqp:invalid-field]
```

# Workaround
Disable QPID expiry feature

`jms.localMessageExpiry=false`

# Details
AMQPv1 trace:
```

RECV: Transfer{handle=0, deliveryId=0, deliveryTag=\x00\x00\x00\x00\x00\x00\x00\x01, messageFormat=0, settled=false, more=false, rcvSettleMode=null, state=null, resume=false, aborted=false, batchable=false} (429) "..."

SENT: Flow{nextIncomingId=1, incomingWindow=2047, nextOutgoingId=1, outgoingWindow=2147483647, handle=0, deliveryCount=1, linkCredit=1, available=null, drain=false, echo=false, properties=null}

SENT: Disposition{role=RECEIVER, first=0, last=0, settled=true, state=Modified{deliveryFailed=true, undeliverableHere=true, messageAnnotations=null}, batchable=false}
```

kjnilsson:
Some client libraries (QPid) will automatically send a disposition with the 'modified' outcome in response to a client local message TTL expiry.

To support this case and others we treat 'modified' the same as 'accepted' and simply ack the message back to the queue.

This change also contains some API extensions to the amqp10_client to better support sending the various delivery states (outcomes).

Fixes #6121


lukebakken (approved):
Makes as much sense as it can given my AMQP 1.0 knowledge 😉 I reverted the removal of the symlinks and Makefile change.

michaelklishin (approved):

PR #6244, opened 2022-10-25
Correctly decrease global counters in rabbit_channel:terminate/2

### LINKED ISSUE #5462: rabbitmq_global_publishers & rabbitmq_global_consumers fail to handle unclean client termination
## Overview

When a client terminates uncleanly, the global counters for publishers and consumers do not handle this gracefully.

For the consumer metric, the value of the gauge does not decrease after the termination, and so the value ever-increases.
For the publisher metric, the value of the gauge decreases below zero and does not recover, leading to a ever-decreasing negative value on the gauge.

Note that the non-global equivalent of these metrics are **unaffected** by this behaviour, and the metrics reported match my expectation (oscillating between 0 and 1).

I discovered this by using a script which ran `while true; do bin/runjava com.rabbitmq.perf.PerfTest --time 60; sleep 10; done`.

## Repro (with observability stack to spot this)

On k8s, assuming you have the `kubectl rabbitmq` plugin [installed](https://www.rabbitmq.com/kubernetes/operator/kubectl-plugin.html#install):

```bash
# Install Cluster Operator & observability stack
kubectl rabbitmq install-cluster-operator
curl -L https://raw.githubusercontent.com/rabbitmq/cluster-operator/main/observability/quickstart.sh | bash

# Create RabbitmqCluster and perf-test Pod
kubectl rabbitmq create sample --image rabbitmq:3.10-management --replicas 3
kubectl wait --for=condition=AllReplicasReady=true rmq/sample
kubectl rabbitmq perf-test sample --time 20
kubectl -n kube-prometheus port-forward svc/prom-kube-prometheus-stack-prometheus 9090
```
Then navigate to the [prometheus service](http://localhost:9090/graph?g0.expr=sum(rabbitmq_global_consumers)&g0.tab=0&g0.stacked=0&g0.show_exemplars=0&g0.range_input=15m&g1.expr=sum(rabbitmq_global_publishers)&g1.tab=0&g1.stacked=0&g1.show_exemplars=0&g1.range_input=15m), and you will see that the values either increase over time with every restart of the perf-test Pod (for `rabbitmq_global_consumers`) or decrease below zero over time (for `rabbitmq_global_publishers`).
Screenshot 2022-08-08 at 17 34 09
Screenshot 2022-08-08 at 17 34 22
You will see that the metrics continue to deviate from 0. Note that the exponential back off due to the PerfTest Pod timing out after 20s will cause the time between 'steps' to increase over time.

See also that the non-global equivalents of these metrics are unaffected:
Screenshot 2022-08-08 at 17 43 48
Screenshot 2022-08-08 at 17 44 18

## Additional info

Reproduced on both k8s and docker-compose environments, all with perf-test.
Affects RabbitMQ 3.9.x and 3.10.x.

kjnilsson:
Previously publisher counts would be decremented only if "publishing_mode" was set to false resulting in ever decreasing global counters.

No consumer counts were decremented in terminate previously resulting in ever growing consumer counts.

Fixes #5462

### DIFF
--- a/deps/rabbit/src/rabbit_channel.erl
+++ b/deps/rabbit/src/rabbit_channel.erl
@@ -877,6 +877,7 @@ handle_post_hibernate(State0) ->
 
 terminate(_Reason,
           State = #ch{cfg = #conf{user = #user{username = Username}},
+                      consumer_mapping = CM,
                       queue_states = QueueCtxs}) ->
     _ = rabbit_queue_type:close(QueueCtxs),
     {_Res, _State1} = notify_queues(State),
@@ -885,6 +886,10 @@ terminate(_Reason,
                             fun() -> emit_stats(State) end),
     [delete_stats(Tag) || {Tag, _} <- get()],
     maybe_decrease_global_publishers(State),
+    _ = maps:map(
+          fun (_, _) ->
+                  rabbit_global_counters:consumer_deleted(amqp091)
+          end, CM),
     rabbit_core_metrics:channel_closed(self()),
     rabbit_event:notify(channel_closed, [{pid, self()},
                                          {user_who_performed_action, Username}]).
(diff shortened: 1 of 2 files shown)


coro (approved):
I've tested with the original repro and this fixes the issue. LGTM!

PR #6247, opened 2022-10-25
Configure how user identity is determined

MarcialRosales:
There are two situations where user's identity is use. One is when we display the username in the management ui. And the second one when we have to capture it in the logs. 
RabbitMQ first looks up if a username was provided along with the token. This is possible when we authenticate via one of the messaging protocols, e.g. amqp. The username field should be left blank and the password carries the JWT. 
If the username is not present, RabbitMQ uses the `sub` claim and if it does not have it either, it uses the `client_id` claim. 

Problems with the current implementation:
- I think the username should be determined exclusively from the token and not from any other source. 
- The majority of Identity Providers use the  `sub` field for some internal primary key associated to the user. When the user logs in to the management ui, it is not a good experience to show an identity the user does not relate to. 
- `client_id` claim is not really representative of the user's identity. Although we can use as the last resort.

In this PR we are adding a new setting in the auth2 backend plugin that allows us to specify one or multiple claims that RabbitMQ can use to determine the user's identity. For instance, UAA typically sets the user's identity in the `user_name` claim. In this case, we would configure RabbitMQ to look for `user_name` first and then for `email`. 
````
[
  {rabbitmq_auth_backend_oauth2, [
   ...
    {preferred_username_claims, [ <<"user_name">>, <<"email">> ]}
    ....
  ]},
].
````

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments


MarcialRosales:
This PR has been merged onto https://github.com/rabbitmq/rabbitmq-server/pull/6015

PR #6263, opened 2022-10-26
Report flow/blocked shovel status in Mgmt UI

gomoripeti:
## Proposed Changes

Some aspects of the implementation considered:
- should work in mixed version clusters:
  The return format of `rabbit_shovel_status:status/0`, which is called via RPC from different nodes is preserved. Instead new elements were added to the info proplist. New nodes can process these new elements while old nodes would ignore them.
- should address the "state-change-interval delay" of credit flow:
  Instead of only reporting the `running -> flow` transition after a 1sec delay, the transitions are immediately reported and stored in ETS together with a `blocked_at` timestamp. The shovel mgmt plugin reimplements the logic found in `credit_flow` to decide on the delay. Advantage is that an idle shovel won't stay stuck in "in flow" state. Disadvantage is that it causes an ETS update for each state transition (and a process can flip between running and flow states dozens of times per second) and it currently goes through a singleton gen_server (which could be a bottleneck)
- should not send shovel_worker_status rabbit events for every flow/blocked/running status change ✅ 

Because the "delay" logic calls the `credit_flow` module which is not available directly for CLI escripts I couldn't use the same solution for the `shovel_status` CLI command. (Maybe the CLI does not even need a state-change delay?)

Alternative: introduce a feature flag so that `rabbit_shovel_status:status/0` can include the delay logic and directly return `flow`/`blocked` statuses in place of `running`.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit_common/src/credit_flow.erl
+++ b/deps/rabbit_common/src/credit_flow.erl
@@ -51,7 +51,7 @@
             Val       -> Val
         end).
 
--export([send/1, send/2, ack/1, ack/2, handle_bump_msg/1, blocked/0, state/0]).
+-export([send/1, send/2, ack/1, ack/2, handle_bump_msg/1, blocked/0, state/0, state_delayed/1]).
 -export([peer_down/1]).
 -export([block/1, unblock/1]).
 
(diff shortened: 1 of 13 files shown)


michaelklishin:
If this expected to be reviewed by the core team, please mark this PR accordingly (it is currently a draft).

gomoripeti:
thanks for the heads up, I will remove the draft marker soon, and will use draft marker in the future accordingly

PraveenNanda124:
Good work

PR #6273, opened 2022-10-27
AMQP 1.0 - improve credit handling for the AMQP 1.0 plugin

luos:
The AMQP 1.0 plugin grants publishers 65000 credit to the senders by default. This PR makes the credit given configurable.

The AMQP 1.0 plugin grants credit when the message is received, not when it is placed into the queue. This means that an AMQP 1.0 sender who does not care about settlement can overload the receiver process leading to high memory usage. Now it is configurable to grant credit on confirm.

Also fixes a bug in the amqp10_client where the client can run out of credit and get stuck because of bad accounting.

## Types of Changes

- [ ] Build system and/or CI


deps/amqp10_client/src/amqp10_client_session.erl
```diff
@@ -839,9 +839,9 @@ translate_delivery_state(received) -> #'v1_0.received'{}.
 translate_role(sender) -> false;
 translate_role(receiver) -> true.
 
-maybe_notify_link_credit(#link{link_credit = 0, role = sender},
+maybe_notify_link_credit(#link{link_credit = OldCredit, role = sender},
                          #link{link_credit = Credit} = Link)
-  when Credit > 0 ->
+  when Credit > 0, OldCredit =< 0 ->
```
kjnilsson, line 844: Favour `andalso` over `,` as it reads better

deps/rabbitmq_amqp1_0/README.md
```diff
@@ -51,6 +51,14 @@ Configuration example using [sysctl config format](https://next.rabbitmq.com/con
     amqp1_0.protocol_strict_mode = false
 
 
+Configuration of incoming links: 
+
+```
+amqp1_0.maximum_incoming_credit = 65536
+# Should be set at on_confirm for stability of RabbitMQ
+amqp1_0.grant_credit = on_publish | on_confirm (default on_publish for backwards compatibility)
```
kjnilsson, line 59: These names of these settings assume the use of AMQP legacy (amqp_client) internally which is something we will move away from, hopefully sooner rather than later. That said it is probably ok for now as I very much doubt we would ever grant credit on_publish in the future and we can just deprecate this config setting.

deps/rabbitmq_amqp1_0/src/rabbit_amqp1_0_incoming_link.erl
```diff
@@ -226,3 +225,47 @@ incoming_flow(#incoming_link{ delivery_count = Count }, Handle) ->
     #'v1_0.flow'{handle         = Handle,
                  delivery_count = {uint, Count},
                  link_credit    = {uint, ?INCOMING_CREDIT}}.
+
+max_incoming_credit() -> 
+    ?INCOMING_CREDIT.
+
+-spec adjust_credits(Method :: ack | transfer,
```
kjnilsson, line 232: instead of `ack` I'd suggest something like `queue_settled` or similar - the ack terminology is just an incident of AMQP legacy extensions.

kjnilsson (changes requested):
Nice catch to notice that credits aren't issued on confirm but on publish.

Mostly nit picking but some naming suggestions.

deps/rabbitmq_amqp1_0/src/rabbit_amqp1_0_incoming_link.erl
luos: I used `settled` as it is in `rabbit_channel`.

deps/rabbitmq_amqp1_0/src/rabbit_amqp1_0_incoming_link.erl
```diff
@@ -7,22 +7,23 @@
...
+-define(INCOMING_CREDIT, application:get_env(rabbitmq_amqp1_0, maximum_incoming_credit, 65536)).
+
 
 -record(incoming_link, {name, exchange, routing_key,
                         delivery_id = undefined,
                         delivery_count = 0,
                         send_settle_mode = undefined,
                         recv_settle_mode = undefined,
-                        credit_used = ?INCOMING_CREDIT div 2,
+                        credit_remaining = 0,
```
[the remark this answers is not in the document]
luos: Yes, my intention was to say `credit_remaining` before next flow. Actually the usage of the field did not change, I just renamed it cause it was confusing. :-)
I renamed it to `credits_before_flow`, maybe that's better.

luos:
Thank you for the review. I tried to incorporate your suggestions, hopefully successfully. :)

luos:
I just realised a thing is still missing - the count of how many messages are outstanding, as now it will grant credit on the first confirm after running out of credit - but it would be better to use how many outstanding messages we have.

PraveenNanda124:
Good work

luos:
If we have a fast publisher it will just push more messages, and CreditRemaning will go below zero (which is expected). In it's current form this change gives full credit on every settled message if `CreditRemaining` is below zero. But, we can not just say `when CreditRemaining = 0` because that's not enough info if we can grant credit. We have to say something like `if (CreditRemaining = 0 && MaxUntilFlow/4 = UncomfirmedCount) orelse (CreditRemaining < 0 && UnconfirmedCount = 0)` otherwise we just give credits on every `Confirm` or we never give credit.

I made the change but it required a bigger refactor, because if we have multiple links in `low credit` then we have to send `flow` for each. I will try to find the time soon to update.

luos:
Hi,

I made the required changes, now for handling credit adjustments, the number of unconfirmed messages is also taken into account.

I refactored to not use the process dictionary because then iterating over the incoming links is easier this way. This can be reverted if you feel this is needed for performance or other reasons. 

Iterating over the links is needed because we may deny credit for some links at some point - and we need to refresh them when the unconfirmed messages goes below a level. 

I tested shovel and realised it does not handle credits, so I added a crude handling of them.

Thanks for your patience.

luos:
Hi,

I've ran some performance tests to compare performance before and after. I do not think the performance is significantly changed. Please find the results here. For settled messages the time it takes to publish them, for unsettled, it is the time it takes to confirm all messages.

Due to #7939 I needed to test with Quorum Queues.  


<div dir="ltr" style="margin-left:0pt;" align="left" id="docs-internal-guid-237b77cd-7fff-9bc0-0c38-f03441ad58c6">

  | Unsettled 0kb | Settled 0kb | Unsettled 1kb | Settled 1kb
-- | -- | -- | -- | --
Main QQ | 12020 | 16549 | 11622 | 21324
Main QQ 2 | 12417 | 18584 | 11611 | 19497
Credit Changes "on_publish" | 12172 | 17155 | 11873 | 17481
Credit Changes "on_publish" | 12107 | 16629 | 11776 | 17587
Credit Changes "on_publish" | 11843 | 17738 | 12089 | 20660
Credit Changes "on_confirm" | 13316 | 16639 | 12280 | 16761
Credit Changes "on_confirm" | 12337 | 23473 | 14142 | 19040
Credit Changes "on_confirm" | 12621 | 17740 | 12240 | 19107
Credit Changes "on_publish".Default credit: 1000 | 12190 | 37212 | 11793 | 36470
Credit Changes "on_confirm"Default credit: 1000 | 12065 | 37587 | 12068 | 38309

</div>

Let me know what you think and if there are any changes needed.

michaelklishin:
Given the ongoing work on a native version of the AMQP 1.0 plugin, this PR is unlikely to make it. We'd appreciate a similar set of improvements on the native AMQP 1.0 branch.

ansd:
Yes https://github.com/rabbitmq/rabbitmq-server/pull/9022 will solve it.

PR #6311, opened 2022-11-02
Copy rather than symlink files in rabbitmqctl_test.bzl

pjk25:
Bazel seems to be cleaning up files after following directory symlinks after the test. Not what I would have expected, and maybe a consequence of lighter sandboxing used on macos with rabbitmq. In any case copying seems to be a reasonable workaround.


pjk25:
This should fix the problem where a failed rabbitmqctl test (and maybe successful ones too), breaks the subsequent running of said tests.

PR #6289, opened 2022-10-31
See Issue 6020. Take plugin dependency into account

SimonUnge:
## Proposed Changes

Addresses #6020.

Comments to this issue suggested querying the node of its active plugins, which would require an RPC call. I took an approach where I include a plugins dependencies instead, and thus avoiding the need to do an RPC call. 

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [x] Bug fix (non-breaking change which fixes issue #6020)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
If you're unsure about any of them, don't hesitate to ask on the mailing list.
We're here to help!
This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

PR #6316, opened 2022-11-02
CLI tools: discover commands in plugins that are enabled indirectly as dependencies; rework plugin state assertions

michaelklishin:
This is #6289 by @SimonUnge (AWS) with some test suite improvements from me.

### DIFF
--- a/deps/rabbitmq_cli/lib/rabbitmq/cli/core/command_modules.ex
+++ b/deps/rabbitmq_cli/lib/rabbitmq/cli/core/command_modules.ex
@@ -74,8 +74,31 @@ defmodule RabbitMQ.CLI.Core.CommandModules do
           []
       end
 
+    all_plugins = PluginsHelpers.list(opts)
+    enabled_with_dep = :rabbit_plugins.dependencies(false, enabled_plugins, all_plugins)
+
+    all_enabled_plugins =
+      Enum.flat_map(
+        all_plugins,
+        fn plugin ->
+          name = PluginsHelpers.plugin_name(plugin)
+
+          strictly_member =
+            :rabbit_plugins.is_strictly_plugin(plugin) and
+              Enum.member?(enabled_with_dep, name)
+
+          case strictly_member do
+            true ->
+              [name]
+
+            false ->
+              []
+          end
+        end
+      )
+
     partitioned =
-      Enum.group_by(enabled_plugins, fn app ->
+      Enum.group_by(all_enabled_plugins, fn app ->
         case Application.load(app) do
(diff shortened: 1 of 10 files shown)


michaelklishin:
I can reproduce the CI failure locally with both Bazel and Make:

``` shell
bazel test deps/rabbit/feature_flags_with_unpriveleged_user_SUITE
```

Investigating.

michaelklishin:
From @pjk25:

So, I believe the issue is this. The test helpers that start nodes always invoke `rabbitmqctl` with no args as part of the setup assertions. We have always expected this to just print the help.
The PR adds a call to `PluginsHelpers.list(opts)` in `deps/rabbitmq_cli/lib/rabbitmq/cli/core/command_modules.ex` at line 77. This ends checking if the plugins dir exists synchronously. Since bazel always has an empty temp dir for the node when things start, the plugins dir does not yet exist. Since Make reuses directories, we probably don’t hit it.
Probably the lack of the directory should just be treated as empty, with no extra plugins, at this point in time when re run rabbitmqctl. The other option would be to adjust either the `rabbitmq_ct_*helpers` to create the directory, or `rabbitmq-run.sh`, if we know that all the rabbitmq installers always create the directory.

deps/rabbitmq_cli/lib/rabbitmq/cli/core/command_modules.ex
```diff
@@ -74,8 +74,31 @@ defmodule RabbitMQ.CLI.Core.CommandModules do
           []
       end
 
+    all_plugins = PluginsHelpers.list(opts)
```
michaelklishin, line 77: This change means we either have to always have the plugins directory set, including in all kinds of test environments, or avoid returning an error when the directory does not exist.

michaelklishin: c2f87ad2d7cec6db2718cc8bcd99177a885ff85d and 0c821ded1a1b4a45d1f98387a4abd15db2757e39 are two different solutions. We are discussing with @pjk25 if we should apply both or just one of them.

pjk25:
@michaelklishin let's revert https://github.com/rabbitmq/rabbitmq-server/pull/6316/commits/c2f87ad2d7cec6db2718cc8bcd99177a885ff85d and see it that passes. It seems to interfere with a few of the feature_flags suites which expect to load some additional test plugins from other paths.
If it passes, we can consider if we want to be more judicious injecting the env var, or take a look at the predence of vars in the failing tests.

Issue #6388, opened 2022-11-09
`rabbitmq-diagnostics memory_breakdown` is slow with many quorum queues


mkuratczyk:
There is something special about quorum queues that causes `rabbitmq-diagnostics memory_breakdown` to take much longer when there is a lot of QQs in the system. The slow-down is non-linear. With 10k QQs it takes ~12 seconds on my machine (single node), with 50k it takes a ~3.5 minutes, with 100k it takes over 20 minutes.

```
> eprof:profile(fun() -> rabbit_vm:memory() end).
> eprof:analyze().
FUNCTION                                                   CALLS        %   TIME  [uS / CALLS]
--------                                                   -----  -------   ----  [----------]
...
rabbit_vm:ra_type/1                                        10000     0.01      3135  [      0.31]
recon_alloc:format_blocks/3                                11604     0.01      3141  [      0.27]
rabbit_vm:'-accumulate/5-fun-0-'/3                         10161     0.01      3168  [      0.31]
lists:sum/2                                                30450     0.01      3413  [      0.11]
rabbit_vm:'-sum_processes/4-lc$^2/1-1-'/1                  10035     0.01      3473  [      0.35]
rabbit_vm:'-sum_processes/3-fun-1-'/3                      20680     0.01      3740  [      0.18]
orddict:from_list/1                                        20680     0.01      3764  [      0.18]
erlang:system_info/1                                       10175     0.01      3801  [      0.37]
lists:splitwith/2                                          20680     0.02      4382  [      0.21]
lists:reverse/1                                            20727     0.02      6049  [      0.29]
erlang:'++'/2                                              20869     0.02      6157  [      0.30]
rabbit_vm:extract/4                                        20142     0.02      6733  [      0.33]
rabbit_vm:'-with/2-lc$^0/1-0-'/2                           30149     0.03      7454  [      0.25]
rabbit_vm:'-aggregate/4-lc$^0/1-0-'/4                      20156     0.03      9080  [      0.45]
rabbit_vm:'-quorum_sups/0-lc$^0/1-0-'/1                    40004     0.04     11175  [      0.28]
lists:foldl_1/3                                            21100     0.04     11926  [      0.57]
rabbit_vm:accumulate/5                                     20680     0.05     14454  [      0.70]
erts_internal:await_result/1                               20680     0.06     15894  [      0.77]
rabbit_vm:find_ancestor/3                                  20680     0.06     17500  [      0.85]
[...]

PR #6390, opened 2022-11-09
Aggregate memory usage on ra_server_sup_sup level

mkuratczyk:
With this change, `rabbitmq-diagnostics memory_breakdown` executes much faster with many QQs. Even with 100k QQs, it's under 5 seconds (previously over 20 minutes).

Fixes https://github.com/rabbitmq/rabbitmq-server/issues/6388


michaelklishin:
One of my favorite programming quotes is "no code is faster than no code" (by Ezra Zygmuntowicz) 👏

PR #6394, opened 2022-11-09
Support code coverage

ansd:
Previously it was not possible to see code coverage for the majority of test cases: integration tests that create RabbitMQ nodes. It was only possible to see code coverage for unit tests. This commit allows to see code coverage for tests that create RabbitMQ nodes.

The only thing you need to do is setting the `COVER` variable, for example
```
make -C deps/rabbitmq_mqtt ct COVER=1
```
will show you coverage across all tests in the MQTT plugin.

This commit fixes the erlang.mk file that did not append the `-cover <sepc_path>` flag to the `ct_run` command.

Whenever a RabbitMQ node is started `ct_cover:add_nodes/1` is called. Contrary to the documentation which states

> To have effect, this function is to be called from init_per_suite/1 (see common_test) before any tests are performed.

I found that it also works in `init_per_group/1` or even within the test cases themselves.

Whenever a RabbitMQ node is stopped or killed `ct_cover:remove_nodes/1` is called to transfer results from the RabbitMQ node to the CT node.

Since the erlang.mk file writes a file called `test/ct.cover.spec` including the line:
```
{export,".../rabbitmq-server/deps/rabbitmq_mqtt/cover/ct.coverdata"}.
```
results across all test suites will be accumulated in that file.

The accumulated result can be seen through the link `Coverage log` on the test suite result pages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,7 @@
 /plugins.lock
 /sbin/
 /sbin.lock
+ct.cover.spec
 erl_crash.dump
 rebar3.crashdump
 .envrc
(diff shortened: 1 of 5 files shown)


lhoguin:
Are you sure about the erlang.mk changes? Because there's tests confirming the file does what it should, at least in vanilla Erlang.mk. In any case they should be sent upstream at https://github.com/ninenines/erlang.mk

ansd:
Thank you @lhoguin for fixing the erlang.mk part upstream via https://github.com/ninenines/erlang.mk/commit/6423c1c0570336c85540d8db1126831d319251bd
I just force pushed.

lhoguin (approved):

PR #6399, opened 2022-11-10
Add inclusive aliases to ctl info keys

illotum:
## Proposed Changes

Add inclusive aliases to ctl info keys per #2635. Our internal checks constantly trip on the language of some scripts.

Includes generic ability to provide aliases in other commands, fixed links in readme and expanded man.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

Pending approval I will also submit a rabbitmq-website fix.

- This does _not_ change the web API, since IIUC it goes directly to the internal keys.
- This changes info_keys return type. Still back-compatible if no aliases used, but requires extra step to get out the keys if aliases _were_ used.

### DIFF
--- a/deps/rabbit/docs/rabbitmqctl.8
+++ b/deps/rabbit/docs/rabbitmqctl.8
@@ -1394,11 +1394,11 @@ or prefetch count.
 .It Cm memory
 Bytes of memory allocated by the runtime for the
 queue, including stack, heap and internal structures.
-.It Cm slave_pids
+.It Cm mirror_pids
 If the queue is mirrored, this lists the IDs of the mirrors (follower replicas).
 To learn more, see the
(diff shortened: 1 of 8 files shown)


illotum:
Ah, so indeed `rabbitmq_management` fails, possibly because of the keys mismatch. Converting to WIP, and will investigate tomorrow.

michaelklishin:
Thank you. This is a great change. The only concern here is how to keep this compatible in mixed version clusters.

Adding aliasing to CLI tools and update the docs is what we had in mind. Internal fields
are probably not worth the effort because CMQs will go away in 4.0 anyway. That should significantly cut both scope and risk.

illotum:
Squashed, rebased, made fully backwards compatible. (Both failed suites pass locally)

PR #6440, opened 2022-11-21
Streams rebalancing

kjnilsson:
Implement "tie-break" selection when more than one replica has the potential to become the next writer. 
If the `preferred_leader_node` option is set it will favour this node for selection assuming its log is at least as up to date as any other member in the first quorum of members to respond. If not set or if not in the first quorum a simple modulo based selection is made to better distribute leader over the cluster.

This may improve writer distribution in cases where a rabbit node goes down and there are many streams.

Adding `restart_stream` command to `rabbitmq-streams` cli tool.

This leaders selection code as been included in the existing drain / rebalance features so that streams will

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -236,7 +236,7 @@ erlang_package.hex_package(
 
 erlang_package.git_package(
     repository = "rabbitmq/osiris",
-    tag = "v1.4.1",
+    tag = "v1.4.2",
 )
 
 erlang_package.hex_package(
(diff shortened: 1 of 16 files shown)


PraveenNanda124:
Good work

michaelklishin:
@PraveenNanda124 please stop with the same comment under every PR or we will have to restrict your interactions with the repo.

michaelklishin:
Note that the option mentioned in the description is a part of the API but also a new flag of a few CLI commands.

acogoluegnes (approved):

PR #6448, opened 2022-11-23
Consider streams when quorum critical check

mkuratczyk:
Extended the quorum critical check to also consider streams and their potential unavailability if the node is stopped.

Affects the following commands:
`rabbitmq-diagnostics check_if_node_is_quorum_critical`
`rabbitmq-upgrade await_online_quorum_plus_one`

### DIFF
--- a/deps/rabbit/src/rabbit_stream_queue.erl
+++ b/deps/rabbit/src/rabbit_stream_queue.erl
@@ -33,6 +33,8 @@
          capabilities/0,
          notify_decorators/1]).
 
+-export([list_with_minimum_quorum/0]).
+
 -export([set_retention_policy/3]).
 -export([add_replica/3,
          delete_replica/3]).
@@ -1039,3 +1041,12 @@ set_leader_pid(Pid, QName) ->
 close_log(undefined) -> ok;
 close_log(Log) ->
     osiris_log:close(Log).
+
+-spec list_with_minimum_quorum() -> [amqqueue:amqqueue()].
+list_with_minimum_quorum() ->
+    lists:filter(fun (Q) ->
+                         StreamId = maps:get(name, amqqueue:get_type_state(Q)),
+                         {ok, Members} = rabbit_stream_coordinator:members(StreamId),
+                         RunningMembers = maps:filter(fun(_, {State, _}) -> State =/= undefined end, Members),
+                         length(maps:values(RunningMembers)) =< length(maps:values(Members)) div 2 + 1
+                 end, rabbit_amqqueue:list_local_stream_queues()).
--- a/deps/rabbit/src/rabbit_stream_queue.erl
+++ b/deps/rabbit/src/rabbit_stream_queue.erl
@@ -1042,11 +1042,13 @@ close_log(undefined) -> ok;
 close_log(Log) ->
     osiris_log:close(Log).
 
+%% this is best effort only; the state of replicas is slightly stale and things can happen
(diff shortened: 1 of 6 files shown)


deps/rabbit/src/rabbit_upgrade_preparation.erl
```diff
@@ -49,3 +55,18 @@ do_await_online_synchronised_mirrors(IterationsLeft) ->
...
+
+-spec list_with_minimum_quorum_for_cli() -> [#{binary() => term()}].
+list_with_minimum_quorum_for_cli() ->
+    EndangeredQueues = lists:append(
+                         rabbit_quorum_queue:list_with_minimum_quorum(),
+                         rabbit_stream_queue:list_with_minimum_quorum()),
+    [begin
+         #resource{name = Name} = amqqueue:get_name(Q),
+         #{
+           <<"readable_name">> => rabbit_data_coercion:to_binary(rabbit_misc:rs(amqqueue:get_name(Q))),
```
kjnilsson, line 67: you already got the name on line 65 so perhaps just bind and use that

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1039,3 +1041,12 @@ set_leader_pid(Pid, QName) ->
 close_log(undefined) -> ok;
 close_log(Log) ->
     osiris_log:close(Log).
+
+-spec list_with_minimum_quorum() -> [amqqueue:amqqueue()].
+list_with_minimum_quorum() ->
+    lists:filter(fun (Q) ->
+                         StreamId = maps:get(name, amqqueue:get_type_state(Q)),
+                         {ok, Members} = rabbit_stream_coordinator:members(StreamId),
```
kjnilsson, line 1049: may be worth commenting that this function returns the coordinator state which may well be a little bit stale. That said this calculation always risks being stale so in practice it doesn't really matter :)

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1039,3 +1041,12 @@ set_leader_pid(Pid, QName) ->
...
+                         RunningMembers = maps:filter(fun(_, {State, _}) -> State =/= undefined end, Members),
+                         length(maps:values(RunningMembers)) =< length(maps:values(Members)) div 2 + 1
```
kjnilsson, line 1051: Despite 10 years or so working with quorum systems I never feel comfortable reading quorum calculations when written in code. :)

In this case would splitting the Members by the predicate above (using `lists:partition/2`) so you have `{Running, NotRunning}` and simply checking that `length(Running) > length(NotRunning)` be sufficient and perhaps easier to read?

kjnilsson: Either way `map_size` could replace `length(maps:values(` on line 1051

kjnilsson (changes requested):
Just a suggestion for code clarity, optional. Otherwise LGTM

deps/rabbit/src/rabbit_stream_queue.erl
mkuratczyk: I decided to keep the calculation as is, since it's basically the same as in the proven QQ implementation (https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_quorum_queue.erl#L407-L408).

PR #6467, opened 2022-11-24
Classic Queue Mirroring: reduce memory usage during dead-lettering of many messages

### LINKED ISSUE #5312: Dead lettering en masse can overload the DLQ
_We observe a potential bug in our brokers_

Classic mirrored DLQ _replica_, at a certain load, flips into CPU and memory churn and cannot recover. [First reported here](https://groups.google.com/g/rabbitmq-users/c/XdETmP7me_4) and later confirmed on multiple RabbitMQ versions.

With appropriate data set the instance is OOM-killed, but even on smaller sets I had the broker crossing into red and back for a long time after the test.

Screen Shot 2022-07-20 at 7 58 47 PM

## Reproduction

Environment
```
> cat ../rmq1.conf
listeners.tcp.default = 127.0.0.1:5671
management.tcp.port = 15671
cluster_formation.peer_discovery_backend = classic_config
cluster_formation.classic_config.nodes.1 = rmq2@3c22fb6527ee

> cat ../rmq2.conf
listeners.tcp.default = 127.0.0.1:5672
management.tcp.port = 15672
cluster_formation.peer_discovery_backend = classic_config
cluster_formation.classic_config.nodes.1 = rmq1@3c22fb6527ee

> echo $RABBITMQ_ENABLED_PLUGINS
rabbitmq_management

> alias brun "bazel run --config=local"
```

Two brokers
```sh
brun broker RABBITMQ_CONFIG_FILE=(realpath ../rmq1.conf) RABBITMQ_NODENAME=rmq1 RABBITMQ_DIST_PORT=25671
```

```sh 
brun broker RABBITMQ_CONFIG_FILE=(realpath ../rmq2.conf) RABBITMQ_NODENAME=rmq2 RABBITMQ_DIST_PORT=25672
```

```sh
brun -- rabbitmqctl -n rmq1@3c22fb6527ee set_policy ha "." '{"ha-mode":"all","ha-sync-mode":"automatic","queue-mode":"lazy"}' --apply-to queues
```

Long message backlog nacked between queues

```
./bin/runjava com.rabbitmq.perf.PerfTest -u "input" -h amqp://localhost:5671 --auto-delete false -f persistent --queue-args x-dead-letter-exchange=,x-dead-letter-routing-key=dlq -x 4 -y 0 -s 1000 -pmessages 250000

# "dlq" created beforehand, say in web console
./bin/runjava com.rabbitmq.perf.PerfTest -u "input" -h amqp://localhost:5671 --auto-delete false -f persistent --queue-args x-dead-letter-exchange=,x-dead-letter-routing-key=dlq -x 0 -y 10 -na -re false
```

## Observations

Slave replica memory footprint stays abnormally high, oscillating +/- 2GB, for many hours (so far up to 4) after the test. There is a significant uptick in CPU as well.

```
queue_slave_procs: 13.3358 gb (62.38%)
binary: 6.2803 gb (29.37%)
allocated_unused: 1.6457 gb (7.7%)
other_proc: 0.0529 gb (0.25%)
code: 0.0348 gb (0.16%)
other_system: 0.0195 gb (0.09%)
other_ets: 0.0037 gb (0.02%)
plugins: 0.0031 gb (0.01%)
atom: 0.0014 gb (0.01%)
metrics: 0.0014 gb (0.01%)
mgmt_db: 0.001 gb (0.0%)
connection_other: 0.0001 gb (0.0%)
mnesia: 0.0001 gb (0.0%)
msg_index: 0.0 gb (0.0%)
quorum_ets: 0.0 gb (0.0%)
queue_procs: 0.0 gb (0.0%)
quorum_queue_dlx_procs: 0.0 gb (0.0%)
stream_queue_procs: 0.0 gb (0.0%)
stream_queue_replica_reader_procs: 0.0 gb (0.0%)
connection_readers: 0.0 gb (0.0%)
connection_writers: 0.0 gb (0.0%)
connection_channels: 0.0 gb (0.0%)
quorum_queue_procs: 0.0 gb (0.0%)
stream_queue_coordinator_procs: 0.0 gb (0.0%)
reserved_unallocated: 0.0 gb (0.0%)
```

From the couple hours I spent on this:

1. Confirmed as low as 1M messages. On my laptop 3M is enough to trigger memory alarm.
1. Confirmed for 1-20 consumers.
1. Confirmed for messages as small as 1 byte.
1. The broker recovers after 1-2 minutes of churn
   - at 250k messages
   - non-lazy queues
   - after deleting "dlq"

mkuratczyk:
These changes fix issue #5312 (memory spike during dead-lettering of many messages). The nature of the changes is that they impact all mirroring operations and could therefore lead to some regressions in completely different scenarios. We've run extensive tests and didn't see any obvious regressions, but we need to be prepared that something might come up.

The reduced memory usage is clear for the scenario described in #5312 :
Screenshot 2022-11-24 at 15 32 40

It's worth pointing out that the goal of the change was to reduce memory usage. However, as a side effect, it also led to some performance improvements, when consuming a long queue, so that it actually uses more memory (but consumes much faster). Here's the biggest difference (CQv2, 100B messages, consumption of a queue with 5 million messages; the sudden cliff was simply when the queue became empty - there were no publishers during this test):
Screenshot 2022-11-24 at 18 30 31

Here's a link to additional results with different queue types (CQv1, CQv1 lazy, CQv2) and different message sizes (hover over the bottom of an annotation on the graph to see full perf-test invocation):
https://snapshots.raintank.io/dashboard/snapshot/ct7cS0pp3wyp5r1ne3UUWfv2zhdY5eIT

Fixes #5312

### DIFF
--- a/deps/rabbit/src/rabbit_dead_letter.erl
+++ b/deps/rabbit/src/rabbit_dead_letter.erl
@@ -104,7 +104,7 @@ group_by_queue_and_reason(Tables) ->
                            ensure_xdeath_event_count(Augmented, N),
                            Key, SeenKeys, Acc),
                   {sets:add_element(Key, SeenKeys), Acc1}
-          end, {sets:new(), []}, Tables),
+          end, {sets:new([{version, 2}]), []}, Tables),
     Grouped.
 
 update_x_death_header(Info, undefined) ->
--- a/deps/rabbit/src/rabbit_mirror_queue_master.erl
+++ b/deps/rabbit/src/rabbit_mirror_queue_master.erl
@@ -122,7 +122,7 @@ init_with_existing_bq(Q0, BQ, BQS) when ?is_amqqueue(Q0) ->
                backing_queue_state = BQS,
                seen_status         = #{},
                confirmed           = [],
-               known_senders       = sets:new(),
+               known_senders       = sets:new([{version, 2}]),
                wait_timeout        = rabbit_misc:get_env(rabbit, slave_wait_timeout, 15000)};
     {error, Reason} ->
         %% The GM can shutdown before the coordinator has started up
--- a/deps/rabbit/src/rabbit_mirror_queue_slave.erl
+++ b/deps/rabbit/src/rabbit_mirror_queue_slave.erl
@@ -878,7 +878,7 @@ maybe_enqueue_message(
 
 get_sender_queue(ChPid, SQ) ->
     case maps:find(ChPid, SQ) of
-        error     -> {queue:new(), sets:new(), running};
+        error     -> {queue:new(), sets:new([{version, 2}]), running};
         {ok, Val} -> Val
     end.
 
--- a/deps/rabbit/src/rabbit_mirror_queue_slave.erl
+++ b/deps/rabbit/src/rabbit_mirror_queue_slave.erl
@@ -18,8 +18,7 @@
 -export([set_maximum_since_use/2, info/1, go/2]).
 
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2,
-         code_change/3, handle_pre_hibernate/1, prioritise_call/4,
-         prioritise_cast/3, prioritise_info/3, format_message_queue/2]).
+         code_change/3, handle_pre_hibernate/1, format_message_queue/2]).
 
 -export([joined/2, members_changed/3, handle_msg/3, handle_terminate/2]).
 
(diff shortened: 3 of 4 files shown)


mkuratczyk:
I realized I forgot to set `x-queue-master-locator=client-local` in this test, so it wasn't apples to apples (in some cases the connections were where the leader was, in some cases on a different node). I've re-run the tests with all connections and all leaders on server-0, which has its own problems as a setup, but allows easier comparison. The performance differences are smaller now, but the spikes when consuming a long queue are still there and not only in case of CQv2.

Here's the memory usage difference when consuming a long queue, in this case CQv1 lazy:
Screenshot 2022-11-25 at 08 51 26

Full results:
https://snapshots.raintank.io/dashboard/snapshot/Ruj66h8qyxj9oAMPmzik957BQs06rCUM?orgId=2

PR #6599, opened 2022-12-08
Fix default queue type handling for single vhost imports

binarin:
## Proposed Changes

Default queue type logic didn't apply when import was done for a single vhost. Exported definitions for a single vhost don't contain `vhost` key, and anyway it's better to use the actual vhost name that will be used for queue declaration.

And just in case, `vhost` is now stripped from exported policies for a single vhost - on import it was ignored, but it could have been a source of the bug similar to the one with the default queue type.

## Types of Changes

- [ ] Build system and/or CI

## Checklist


michaelklishin:
References #5305 #5769

michaelklishin:
Dammit, I merged the wrong PR without paying attention to what tab I had active :) I will correct a couple of things in a follow-up.

michaelklishin:
@mergifyio backport v3.11.x v3.10.x

PR #6682, opened 2022-12-15
Feature flags: Add virgin node init support to v2 code path

dumbbell:
So far, when a new node is started, the feature flags v1 code is used to enable all stable feature flags (the default behavior). In the list, there would be `feature_flags_v2`, so this one would be enabled first and all others would be using the v2 code path.

If we mark `feature_flags_v2` as required, the v2 code path didn't handle this so far. This patch adds this missing code. Now, a virgin node's feature flags can be enabled fully using the v2 code path.

As part of that, the `get_forced_feature_flag_names()` function is moved from `rabbit_feature_flags` to `rabbit_ff_controller`.

It depends on #6791.

**Update**: The patch was split into two parts, this one (which was the initial scope) and #6810.

### DIFF
--- a/deps/rabbit/src/rabbit_feature_flags.erl
+++ b/deps/rabbit/src/rabbit_feature_flags.erl
@@ -1826,15 +1826,19 @@ sync_feature_flags_with_cluster(Nodes, NodeIsVirgin) ->
 %% @private
 
 sync_feature_flags_with_cluster(Nodes, NodeIsVirgin, Timeout) ->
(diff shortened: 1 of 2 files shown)


dumbbell:
Note that I widened the scope of this pull request to make the v2 code path required. The description was updated accordingly.

dumbbell:
I marked this pull request as draft again because I would like to sort out the problem I describe in https://github.com/rabbitmq/rabbitmq-server/pull/6701#issuecomment-1363161326 first.

PR #6738, opened 2022-12-21
Check if stream coordinator is quorum critial

### LINKED ISSUE #4890: Add rabbitmq-upgrade command for streams
Currently there is no command equivalent to `await_online_quorum_plus_one` that takes streams into account.

Also there would need to be some way of ensuring that the stream coordinator would not lose quorum.

mkuratczyk:
Since https://github.com/rabbitmq/rabbitmq-server/commit/d8ff9918, streams are considered when checking for a potential quorum loss. Now we also consider `rabbit_stream_coordinator`.

Fixes #4890

### DIFF
--- a/deps/rabbit/src/rabbit_upgrade_preparation.erl
+++ b/deps/rabbit/src/rabbit_upgrade_preparation.erl
@@ -32,15 +32,39 @@ await_online_synchronised_mirrors(Timeout) ->
 %% Implementation
 %%
 
+online_members(Component) ->
+    lists:filter(fun erlang:is_pid/1,
+                 rabbit_misc:append_rpc_all_nodes(rabbit_nodes:all_running(),
+                                                  erlang, whereis, [Component])).
+
+endangered_critical_components() ->
+    CriticalComponents = [rabbit_stream_coordinator],
+    Nodes = rabbit_nodes:all(),
+    lists:filter(fun (Component) ->
+                         NumAlive = length(online_members(Component)),
+                         ServerIds = lists:zip(lists:duplicate(length(Nodes), Component), Nodes),
+                         case ra:members(ServerIds) of
+                             {error, _E} ->
+                                 %% we've asked all nodes about it; if we didn't get an answer,
+                                 %% the component is probably not running at all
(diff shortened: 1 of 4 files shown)


deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -2092,3 +2095,16 @@ transfer_leadership([Destination | _] = _TransferCandidates) ->
...
     end.
+
+online_coordinator_members() ->
+    lists:filter(fun erlang:is_pid/1,
+                 rabbit_misc:append_rpc_all_nodes(rabbit_nodes:all_running(),
+                                                  erlang, whereis, [rabbit_stream_coordinator])).
+
+is_quorum_critical() ->
+    NumAlive = length(online_coordinator_members()),
+    case ra:members(rabbit_stream_coordinator) of
```
kjnilsson, line 2106: Does this mean that if the local stream coordinator member isn't running we say it is ok to go ahead?

deps/rabbit/src/rabbit_upgrade_preparation.erl
```diff
@@ -35,11 +35,12 @@ await_online_synchronised_mirrors(Timeout) ->
 do_await_safe_online_quorum(0) ->
     false;
 do_await_safe_online_quorum(IterationsLeft) ->
+    CriticalComponents = [rabbit_stream_coordinator],
     EndangeredQueues = lists:append(
                          rabbit_quorum_queue:list_with_minimum_quorum(),
                          rabbit_stream_queue:list_with_minimum_quorum()),
     case EndangeredQueues of
-        []  -> true;
+        []  -> not lists:member(true, [C:is_quorum_critical() || C <- CriticalComponents]);
```
kjnilsson, line 43: I like this API? should it have its own module with a behaviour definition?

kjnilsson (changes requested):
A couple of questions / suggestions

michaelklishin:
@mergifyio rebase

michaelklishin:
@mergifyio backport v3.11.x

PR #6777, opened 2022-12-28
Fix error when free space can not be determined

drencrom:
When disk free space can not be determined for any reason the value is set to NaN and that was not considered when rabbitmqctl status command tried to show the free space returning this error instead:

`** (FunctionClauseError) no function clause matching in RabbitMQ.CLI.InformationUnit.convert/2`

The code has been changed to check for the :NaN value explicitly


drencrom:
Now I'm thinking that maybe I should have changed :unknown for :Nan instead of adding a new case.

michaelklishin:
How is this different from #6294?

drencrom:
> How is this different from #6294?

Somehow I missed that. Yes it is the same.

PR #6806, opened 2023-01-05
rabbit_db: Move missed vhost and user Mnesia-specific code

dcorbacho:
This is a follow-up to #6430 (and should have been part of it).

It is a preparation step to be able to change the database layer.


dumbbell (approved):
Thank you!

PR #6807, opened 2023-01-05
erlang_ls workaround for bazel

pjk25:
Adds a new bazel target, `bazel run //tools:symlink_deps_for_erlang_ls`, which will create an `extra_deps` folder with symlinks (to other symlinks) to 3rd party deps. It also created a few special case symlinks for generated headers and sources in `rabbit_common`. Unfortunately it might be necessary to clean the symlinks when switching back to make with `rm -df extra_deps; git clean -xffd deps/rabbit_common`.

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -3,3 +3,4 @@
 .erlang.mk
 deps/osiris
 deps/ra
+extra_deps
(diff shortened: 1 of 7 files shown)


pjk25:
Note: May not work with erlang_ls less that `0.46`

pjk25:
@ikvmw is this good to merge in your opinion?

PR #6821, opened 2023-01-09
Move missing Mnesia-specific code to rabbit_db_* modules

dcorbacho:
Continuation of the work started in https://github.com/rabbitmq/rabbitmq-server/pull/6430

The idea is to isolate in those modules the code to handle the records in the database. For now, we only support Mnesia, but there is work in progress to move to another database engine. Doing this reorganization of the code will also isolate the changes of this upcoming database engine switch, making them easier to rebase and review.


deps/rabbit/src/rabbit_core_ff.erl
```diff
@@ -132,13 +132,10 @@
       Ret :: rabbit_feature_flags:enable_callback_ret().
 direct_exchange_routing_v2_enable(#{feature_name := FeatureName}) ->
     TableName = rabbit_index_route,
-    ok = rabbit_table:wait([rabbit_route, rabbit_exchange], _Retry = true),
```
dumbbell, line 135: The wait for `rabbit_exchange` table seems to have been lost in a rebase. It was added very recently by @ansd in #6847.

deps/rabbit/src/rabbit_binding.erl
```diff
@@ -710,76 +375,56 @@ merge_entry({X1, Deleted1, Bindings1}, {X2, Deleted2, Bindings2}) ->
...
-
-x_callback(Serial, X, F, Bs) ->
-    ok = rabbit_exchange:callback(X, F, Serial, [X, Bs]).
-
--spec populate_index_route_table() -> ok.
-populate_index_route_table() ->
-    rabbit_misc:execute_mnesia_transaction(
-      fun () ->
-              mnesia:read_lock_table(rabbit_route),
-              mnesia:read_lock_table(rabbit_exchange),
```
dumbbell, line 754: This lock of `rabbit_exchange` is missing in the moved function in `rabbit_db_binding`. Is it on purpose?

deps/rabbit/src/rabbit_binding.erl
```diff
@@ -100,204 +89,106 @@ recover(XNames, QNames) ->
...
 -spec add(rabbit_types:binding(), inner_fun(), rabbit_types:username()) -> bind_res().
 
-add(Binding, InnerFun, ActingUser) ->
-    binding_action(
-      Binding,
-      fun (Src, Dst, B) ->
-              case rabbit_exchange:validate_binding(Src, B) of
-                  ok ->
-                      lock_resource(Src, read),
-                      lock_resource(Dst, read),
```
dumbbell, line 163: Those two locks seem to be missing in the converted code. Is this on purpose?

dumbbell (changes requested):
This review is incomplete, but I want to save my progress :-) I stopped before reading `rabbit_db_exchange.erl`.

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1896,7 +1726,7 @@ forget_node_for_queue(DeadNode, [H|T], Q) when ?is_amqqueue(Q) ->
         {false, _} -> forget_node_for_queue(DeadNode, T, Q);
         {true, rabbit_classic_queue} ->
             Q1 = amqqueue:set_pid(Q, rabbit_misc:node_to_fake_pid(H)),
-            ok = mnesia:write(rabbit_durable_queue, Q1, write);
+            ok = rabbit_db_queue:set_many([Q1]);
```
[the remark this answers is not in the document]
dcorbacho: That's because `set_many` it's the only set operation we have that just updates the durable queue. As `forget_node_for_queue` is to be removed when mnesia/transient entities are deleted, I thought that better stick to the API that we already have instead of add a `set_durable` that is going to be removed too. Sadly the queue API will need some changes once mnesia is gone, because we have to deal with durable and transient queues in a non-transparent way.

deps/rabbit/src/rabbit_binding.erl
dcorbacho: They're there: https://github.com/rabbitmq/rabbitmq-server/pull/6821/files/f93b5dccc9909f96933eff11cf7bc81f0fc67dc3#diff-ba4aec2563f03086e68cfd65539cabb6008cb8ad20904c9802830ee67e2eb6dfR482

deps/rabbit/src/rabbit_binding.erl
dcorbacho: Accidental loss

the-mikedavis (commented):
I haven't finished reading but I'm liking the way it's looking so far 👍

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -164,31 +163,23 @@ start(Qs) ->
...
-        fun() ->
-            [ store_queue(amqqueue:set_state(Q, stopped))
-              || Q <- Qs, amqqueue:get_type(Q) =:= rabbit_classic_queue,
-                 amqqueue:get_state(Q) =/= stopped ]
-        end).
+    Qs0 = find_local_durable_queues(VHost),
+    Qs = [amqqueue:set_state(Q, stopped)
+          || Q <- Qs0, amqqueue:get_type(Q) =:= rabbit_classic_queue,
+             amqqueue:get_state(Q) =/= stopped ],
+    rabbit_db_queue:set_many(Qs).
```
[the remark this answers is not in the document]
dcorbacho: We can, we now have `foreach_durable` that is used by on_node_down/node_up that does that. It filters durable queues and does something with them. I'll update it.

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -122,7 +122,7 @@ is_recoverable(Q) when ?is_amqqueue(Q) ->
     %% record if it is a mirrored queue (such info is now obtained from
     %% the policy). Thus, we must check if the local pid is alive
     %% - if the record is present - in order to restart.
-    (mnesia:read(rabbit_queue, amqqueue:get_name(Q), read) =:= []
+    (rabbit_amqqueue:exists(amqqueue:get_name(Q))
```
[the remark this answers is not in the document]
dcorbacho: Right! Well spotted :)

deps/rabbit/src/rabbit_binding.erl
```diff
@@ -100,196 +90,107 @@ recover(XNames, QNames) ->
...
+                   info(Binding) ++ [{user_who_performed_action, ActingUser}]);
+        Err ->
+            Err
     end.
 
--spec remove(rabbit_types:binding()) -> bind_res().
-remove(Binding) -> remove(Binding, fun (_Src, _Dst) -> ok end, ?INTERNAL_USER).
-
--spec remove(rabbit_types:binding(), rabbit_types:username()) -> bind_res().
-remove(Binding, ActingUser) -> remove(Binding, fun (_Src, _Dst) -> ok end, ActingUser).
```
the-mikedavis, line 197: Looks like this is used in `rabbit_mqtt_processor:unbind/3` (might be new from the merge/rebase on #5895 I think?)

deps/rabbit/src/rabbit_db_policy.erl
```diff
@@ -0,0 +1,69 @@
...
+      Ret :: {[{Exchange, Exchange}], [{Queue, Queue}]}.
+
+update(VHost, GetUpdatedExchangeFun, GetUpdatedQueueFun) ->
+    rabbit_db:run(
+      #{mnesia => fun() -> update_in_mnesia(VHost, GetUpdatedExchangeFun, GetUpdatedQueueFun) end
+       }).
+
+%% [1] We need to prevent this from becoming O(n^2) in a similar
+%% manner to rabbit_binding:remove_for_{source,destination}. So see
+%% the comment in rabbit_binding:lock_route_tables/0 for more rationale.
```
the-mikedavis, line 36: This is not really related to this PR since this comment only moved but this `rabbit_binding:lock_route_tables/0` function looks like it hasn't existed for a while. I'm not quite sure which comment in `rabbit_binding` this is referring to now. Maybe this can be removed?

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1306,100 +1237,35 @@ is_in_virtual_host(Q, VHostName) ->
...
 list_down(VHostPath) ->
     case rabbit_vhost:exists(VHostPath) of
         false -> [];
         true  ->
             Alive = sets:from_list([amqqueue:get_name(Q) || Q <- list(VHostPath)]),
-            Durable = list(VHostPath, rabbit_durable_queue),
             NodesRunning = rabbit_nodes:all_running(),
-            lists:filter(fun (Q) ->
+            rabbit_db_queue:filter_all_durable(
+              VHostPath, fun (Q) ->
```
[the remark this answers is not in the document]
dcorbacho: Ouch, thanks! It's now part of the FilterFun. I had initially implemented `filter_all_durable/1` and `filter_all_durable/2` to pass the VHost, but it's really not needed as it can be part of the FilterFun and we have a single function to maintain. I forgot to change this one before pushing it.

deps/rabbit/src/rabbit_db_policy.erl
dcorbacho: I have no clue either, I left it just in case. Legacy code can be full of surprises. It will eventually be removed with mnesia.

deps/rabbit/src/rabbit_binding.erl
dcorbacho: Yes, it looks like is used by the processor now. Restored.

deps/rabbit/src/rabbit_db_exchange.erl
```diff
@@ -0,0 +1,569 @@
...
+    rabbit_mnesia:table_filter(
+      fun (#exchange{name = XName}) ->
+              XName#resource.virtual_host =:= VHost andalso
+                  mnesia:read({?MNESIA_TABLE, XName}) =:= []
+      end,
+      fun (X, true) ->
+              X;
+          (X, false) ->
+              X1 = rabbit_mnesia:execute_mnesia_transaction(
+                     fun() -> set_in_mnesia_tx(X) end),
```
dumbbell, line 474: I can't find in the original `rabbit_exchange:recover()` where the record is written to the `rabbit_durable_exchange` table. The original exchange code is really difficult to follow.

Do you remember where it is written in `recover()`?

deps/rabbit/src/rabbit_exchange.erl
```diff
@@ -286,49 +247,33 @@ lookup_scratch(Name, App) ->
...
-
-update(Name, Fun) ->
-    case mnesia:wread({rabbit_exchange, Name}) of
-        [X] -> X1 = Fun(X),
-               store(X1);
-        []  -> not_found
+    Decorators = case rabbit_db_exchange:get(Name) of
+                     {ok, X} -> rabbit_exchange_decorator:active(X);
+                     {error, not_found} -> []
+                 end,
```
dumbbell, line 253: I can't find where this comes from either in the original code. Do you remember where?

dumbbell (changes requested):
A few more comments and questions. I stopped before `rabbit_exchange_type_topic.erl`.

deps/rabbit/src/rabbit_maintenance.erl
```diff
@@ -180,22 +130,11 @@ is_being_drained_consistent_read(Node) ->
 
 -spec status_local_read(node()) -> maintenance_status().
 status_local_read(Node) ->
-    case catch mnesia:dirty_read(?TABLE, Node) of
-        []  -> ?DEFAULT_STATUS;
-        [#node_maintenance_state{node = Node, status = Status}] ->
-            Status;
-        _   -> ?DEFAULT_STATUS
-    end.
+    rabbit_db_maintenance:get(Node).
```
dumbbell, line 133: `rabbit_db_maintenance:get/1` can return `undefined` which is not the case of the original `status_local_read/1`. The latter would default to `?DEFAULT_STATUS`.

deps/rabbit/src/rabbit_maintenance.erl
```diff
@@ -180,22 +130,11 @@ is_being_drained_consistent_read(Node) ->
...
 -spec status_consistent_read(node()) -> maintenance_status().
 status_consistent_read(Node) ->
-    case mnesia:transaction(fun() -> mnesia:read(?TABLE, Node) end) of
-        {atomic, []} -> ?DEFAULT_STATUS;
-        {atomic, [#node_maintenance_state{node = Node, status = Status}]} ->
-            Status;
-        {atomic, _}  -> ?DEFAULT_STATUS;
-        {aborted, _Reason} -> ?DEFAULT_STATUS
-    end.
+    rabbit_db_maintenance:get_consistent(Node).
```
dumbbell, line 137: Same comment as for `status_local_read/1`.

deps/rabbit/src/rabbit_policy.erl
```diff
@@ -441,72 +430,69 @@ notify_clear(VHost, <<"operator_policy">>, Name, ActingUser) ->
...
+        case catch {list(VHost), list_op(VHost)} of
+            {'EXIT', {throw, {error, {no_such_vhost, _}}}} ->
+                {[], []}; %% [2]
+            {'EXIT', Exit} ->
+                exit(Exit);
+            {Policies, OpPolicies} ->
+                rabbit_db_policy:update(VHost,
+                                        get_updated_exchange(Policies, OpPolicies),
+                                        get_updated_queue(Policies, OpPolicies))
+        end,
```
dumbbell, line 443: Listing the policies for `VHost` is outside the transaction. I have the feeling we already talked about this in Slack, but I can't find the discussion.

Is it ok to list policies outside of the transaction?

dumbbell (changes requested):
Done with the `rabbit` application sources. Onto the testsuites!

deps/rabbitmq_federation/src/rabbit_federation_exchange_link.erl
```diff
@@ -438,12 +438,8 @@ go(S0 = {not_started, {Upstream, UParams, DownXName}}) ->
...
                           _            -> unknown
                       end,
-              {Serial, Bindings} =
-                  rabbit_misc:execute_mnesia_transaction(
-                    fun () ->
-                            {rabbit_exchange:peek_serial(DownXName),
-                             rabbit_binding:list_for_source(DownXName)}
-                    end),
+              {Serial, Bindings} = {rabbit_exchange:peek_serial(DownXName),
+                                    rabbit_binding:list_for_source(DownXName)},
```
dumbbell, line 442: Is it ok to get rid of the transaction?

dumbbell (changes requested):
I'm done with the review.

Overall, I think this pull request could be merged soon. It's long and will be difficult to maintain with other patches merged in parallel. We can fix and improve it later.

deps/rabbit/src/rabbit_binding.erl
dcorbacho: This one: https://github.com/rabbitmq/rabbitmq-server/pull/6821/files#diff-ba4aec2563f03086e68cfd65539cabb6008cb8ad20904c9802830ee67e2eb6dfR499

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -127,7 +127,7 @@ is_recoverable(Q) when ?is_amqqueue(Q) ->
     %% record if it is a mirrored queue (such info is now obtained from
     %% the policy). Thus, we must check if the local pid is alive
     %% - if the record is present - in order to restart.
-    (mnesia:read(rabbit_queue, amqqueue:get_name(Q), read) =:= []
+    (not rabbit_amqqueue:exists(amqqueue:get_name(Q))
```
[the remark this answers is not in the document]
dcorbacho: I don't have it clear either, maybe none but reasoning about transient mirrored records is hard. I'm restoring the original behaviour by introducing `rabbit_db_queue:consistent_exists/1` just for this. It can be removed with transient entities.

deps/rabbit/src/rabbit_db_exchange.erl
dcorbacho: Well spotted, it's not. It needs to call just `set_ram_in_mnesia_tx`

deps/rabbit/src/rabbit_exchange.erl
dcorbacho: It doesn't come exactly as that. The previous `rabbit_exchange:store` would set the decorators for the ram record here: https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_exchange.erl#L157 In prep work for Khepri, we collect the decorators outside of the transaction and pass them to the update/store functions. Decorators can't be collected inside a Khepri function because they use apply to call modules not known in compile time. If you consider this is too ahead of time, I can revert it to the old behaviour of calculating decorators inside of the transaction.

deps/rabbit/src/rabbit_policy.erl
dcorbacho: I think it is. Each policy we set generates one of this `notify` callbacks, so let's say that we update one policy and one operator policy more or less simultaneously. If the first update doesn't pick up both policies, the second one will. Also 
 the same for clear notifications. We don't have an atomic way to update policies, so if there is a temporary mismatch I don't think it can be an issue.

deps/rabbitmq_federation/src/rabbit_federation_exchange_link.erl
dcorbacho: I think so. The original implementation of `rabbit_binding:list_for_source` already uses `mnesia:async_dirty`

deps/rabbitmq_consistent_hash_exchange/src/rabbit_exchange_type_consistent_hash.erl
```diff
@@ -118,171 +111,126 @@ validate_binding(_X, #binding { key = K }) ->
...
-    ok = mnesia:delete({?HASH_RING_STATE_TABLE, Name});
-delete(_Tx, _X, _Bs) ->
-    ok.
+    rabbit_log:debug("Consistent hashing exchange: will recover exchange ~ts", [rabbit_misc:rs(XName)]),
+    create(none, X),
+    rabbit_log:debug("Consistent hashing exchange: recovered exchange ~ts", [rabbit_misc:rs(XName)]),
+    Bindings = rabbit_binding:list_for_source(XName),
+    rabbit_log:debug("Consistent hashing exchange: have ~b bindings to recover for exchange ~ts",
+                     [length(Bindings), rabbit_misc:rs(XName)]),
+    [add_binding(none, X, B) || B <- lists:usort(Bindings)],
```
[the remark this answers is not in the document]
dcorbacho: Once again it uses `mnesia:aysnc_dirty`to list the bindings. We probably don't need the transaction

deps/rabbit/src/rabbit_binding.erl
dcorbacho: Done

deps/rabbit/src/rabbit_classic_queue.erl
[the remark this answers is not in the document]
michaelklishin: In this part of the code, especially given the comment above, I'd keep the original (more consistent) read.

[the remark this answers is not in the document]
dcorbacho: Yes, that is done

the-mikedavis (dismissed):
I gave this another read through and it looks good to me. I'm not super familiar with the server codebase though so take my review with a grain of salt.

I like seeing the direct `mnesia` calls and match-specs moved into db modules, it feels more clean (even independent of the upcoming khepri changes) 👍

deps/rabbit/src/rabbit_exchange.erl
dumbbell: I see. No need to revert, it's probably ok this way and as you said, we can't keep the current behavior in a Khepri world.

deps/rabbit/src/rabbit_policy.erl
dumbbell: Perfect, thank you!

deps/rabbitmq_federation/src/rabbit_federation_exchange_link.erl
dumbbell: Ok, thank you.

dumbbell (approved):
Awesome! Thank you :-)

PR #6867, opened 2023-01-12
v3.10.x: backport #6467

michaelklishin:
As a test to see if there would be any conflicts or CI failures.


michaelklishin:
FTR, #6467 relies on Sets v2 that were introduced in Erlang 24. RabbitMQ 3.10.x already requires 24.x, and later this year will require 25.

PR #6871, opened 2023-01-13
Conform vhost tags to a list when set with the cli in all cases

pjk25:
`rabbitmqctl add_vhost myvhost --tags "my_tag"` would not previously conform "my_tag" to a list before setting vhost metadata, which could cause crashes when the list was read.


pjk25:
This won't correct tags which are already stored incorrectly on [vhost metadata](https://github.com/rabbitmq/rabbitmq-server/blob/43b78aa7adb7319c3c1be076bb83ce5d6bb3a002/deps/rabbit/src/vhost.erl#L43-L58), but re-running the cli command should correct them. Adding coercion to any place that reads vhost metadata seems like overkill, since reports of this bug are few.

PR #6882, opened 2023-01-13
rabbit_db: Add clustering-related functions

dumbbell:
The new `rabbit_db_cluster` module sits on top of `rabbit_mnesia` and provide an API with all cluster-related functions.

`rabbit_mnesia` should be called directly inside Mnesia-specific code only, `rabbit_mnesia_rename` or classic mirrored queues for instance. Otherwise, `rabbit_db_cluster` must be used.

Several modules, in particular in `rabbitmq_cli`, continue to call `rabbit_mnesia` as a fallback option if the `rabbit_db_cluster` module unavailable. This will be the case when the CLI will interact with an older RabbitMQ version.

This is a follow-up to #6430. It depends on #7058. It is a preparation step to be able to change the database layer.

### DIFF
--- a/deps/rabbit/src/rabbit_db.erl
+++ b/deps/rabbit/src/rabbit_db.erl
@@ -13,6 +13,9 @@
 -include_lib("rabbit_common/include/logging.hrl").
 
 -export([init/0,
+         reset/0,
+         force_reset/0,
+         force_load_on_next_boot/0,
          is_virgin_node/0, is_virgin_node/1,
          dir/0,
          ensure_dir_exists/0]).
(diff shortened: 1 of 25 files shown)


deps/rabbit/src/rabbit_db.erl
```diff
@@ -53,14 +56,56 @@ init() ->
...
+      "DB: resetting node forcefully",
+      #{domain => ?RMQLOG_DOMAIN_DB}),
+    rabbit_mnesia:force_reset().
+
+-spec force_load_on_next_boot() -> Ret when
+      Ret :: ok.
+%% @doc Requests that the database to be forcefully loaded during next boot.
+%%
+%% This is necessary when a node refuses to boot when the cluster is in a bad
+%% state, like a critical members are MIA.
```
the-mikedavis, line 98:
```suggestion
%% state, like if critical members are MIA.
```

dumbbell: Thanks! Fixed

the-mikedavis (approved):

dcorbacho (dismissed):

PR #6919, opened 2023-01-17
Rework elixir dialyze

pjk25:
Fix the duplication of modules between elixir and rabbitmqctl, that caused problems with dialyzer

### DIFF
--- a/deps/rabbitmq_cli/rabbitmqctl.bzl
+++ b/deps/rabbitmq_cli/rabbitmqctl.bzl
@@ -18,22 +18,38 @@ load(
     "maybe_install_erlang",
 )
 
+ElixirAppInfo = provider(
+    doc = "Compiled Erlang Application",
+    fields = {
+        "app_name": "Name of the erlang application",
+        "extra_apps": "Extra applications in the applications key of the .app file",
+        "include": "Public header files",
+        "beam": "ebin directory produced by mix",
+        "consolidated": "consolidated directory produced by mix",
+        "priv": "Additional files",
+        "license_files": "License files",
+        "srcs": "Source files",
+        "deps": "Runtime dependencies of the compiled sources",
+    },
+)
+
 def deps_dir_contents(ctx, deps, dir):
     files = []
     for dep in deps:
         lib_info = dep[ErlangAppInfo]
         for src in lib_info.include + lib_info.beam + lib_info.srcs:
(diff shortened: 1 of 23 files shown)


pjk25:
@acogoluegnes do the stream plugin fixes look alright to you?

pjk25:
I will resolve the conflicts for this

deps/rabbitmq_cli/rabbitmqctl.bzl
```diff
@@ -180,15 +207,144 @@ rabbitmqctl_private = rule(
...
+set -euo pipefail
+
+cp {app_ebin}/* {ebin}
+cp -f {app_consolidated}/* {ebin}
+
+for beam in {elixir_ebin}/*; do
+    find {ebin} -name "$(basename $beam)" -delete
+done
+""".format(
+                elixir_ebin = elixir_ebin,
```
binarin, line 266: I think every substituted bit should be shell-quoted, so it won't break apart with all the strange names.

binarin (changes requested):
A round of cleanup will still be needed, e.g. replacing repeated `erts, kernel, stdlib` with global `plt = "//:base_plt"`. But maybe as a separate PR to cleanup everything at once, when all plugins dialyze successfully.

pjk25:
> A round of cleanup will still be needed, e.g. replacing repeated `erts, kernel, stdlib` with global `plt = "//:base_plt"`. But maybe as a separate PR to cleanup everything at once, when all plugins dialyze successfully.

Some cleanup/normalization will be a natural consequence of using gazelle to maintain the BUILD.bazel files

deps/rabbitmq_cli/rabbitmqctl.bzl
binarin: I was thinking about something like https://github.com/bazelbuild/bazel-skylib/blob/main/docs/shell_doc.md#shellquote
In case if someone decides to have a stupidly named file or folder like (literally) `$HOME`

pjk25: In practice, we probably don't have to worry that much here. The files all are either elixir modules, or need to have passed through mix, and bazel would have probably have already failed if the project itself exists in strange dir.
However, I didn't know skylib contained such a helper, so I'll make another pass at it, mostly due to my own curiosity.

binarin (approved):

PR #6928, opened 2023-01-18
Simplify BUILD files by using new macros from rules_erlang 3.9.0

pjk25:
This is also a necessary transformation prior to using gazelle


michaelklishin (approved):
Works as expected for me locally.

PR #6931, opened 2023-01-18
  Support OAuth 2.0 authentication in AMQP 1.0 plugin

MarcialRosales:
Protocol plugins such as mqtt, or stomp, they handle an incoming connection by opening an internal amqp connection and the credentials received over the initial connection are passed to the internal amqp connection. In other words, the authentication/authorization occurs at the internal amqp process.

With AMQP 1.0 plugin things are slightly different. When the plugin receives an amqp 1.0 connection, it reads the user credentials (carried on the sasl frame) and authenticates the user against the `rabbit_access_control` module. In other words, authentication happens on the plugin itself. If the authentication is successful, the username is kept on some state for future usage.
Later on another amqp 1.0 frame arrives which triggers opening an internal amqp connection with just the username as credentials, i..e no password.  However, that has not been a problem because the internal and some other auth-backend like LDAP, does support authenticating a user by its username only. 

However, OAuth 2 (and http backend too) requires the full credentials received during the initial connection phase, ie. the username and password. And this is exactly the [issue](https://github.com/rabbitmq/rabbitmq-server/issues/6909) because the internal amqp process tries to authenticate the user by passing a username and no password with the oauth2 backend which tries to parse a blank token and fails with the error "Invalid token". 

## Proposed Changes

Use the outcome from the first authentication as a valid credential for subsequent authentication attempts which occur when a session is opened.

## Types of Changes

- [ ] Build system and/or CI

## Additional notes

Only the internal and ldap authentication backends support user authentication without password. This is the type of authentication used to create the internal amqp connection. 
The http authentication backend does not support authentication without password. In other words, it requires both,  username and password.

### DIFF
--- a/deps/rabbit/src/rabbit_auth_backend_internal.erl
+++ b/deps/rabbit/src/rabbit_auth_backend_internal.erl
@@ -92,7 +92,11 @@ user_login_authentication(Username, AuthProps) ->
                           false
                   end
               end);
-        false -> exit({unknown_auth_props, Username, AuthProps})
+        false ->
+            case proplists:get_value(rabbit_auth_backend_internal, AuthProps, undefined) of
+                undefined -> exit({unknown_auth_props, Username, AuthProps});
+                _ -> internal_check_user_login(Username, fun(_) -> true end)
+            end
     end.
 
 state_can_expire() -> false.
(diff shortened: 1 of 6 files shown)


kjnilsson (approved):
This looks good to me as an approach until we can get native AMQP 1.0 support and most likely don't need to hang on to the token.

kjnilsson:
> This looks good to me as an approach until we can get native AMQP 1.0 support and most likely don't need to hang on to the token.

that said there are some failed checks that would need looking into

michaelklishin:
```
rabbit_auth_backend_internal.erl:98:17: variable 'Authz_Backends' is unused
```

is the latest Dialyze/Xref failure.

kjnilsson (approved):
This looks good to me but I would love someone with a bit more context around the auth system to take a look also.

michaelklishin (approved):

PR #6935, opened 2023-01-19
Use gazelle for external erlang deps

pjk25:
This adds explicit bazel BUILD info for external deps. This allows the windows build to use bzlmod, since the bzlmod extension can only auto-import deps on macos and linux. This in turn allows deletion of workspace_helpers.bzl, and elimination of double bookkeeping.

The BUILD.lib_name files are generated with gazelle, e.g. `bazel run gazelle -- update-repos --verbose --build_files_dir=bazel hex.pm/redbug@2.0.7`
Running gazelle will update `WORKSPACE`, but the changes should be ignored (because gazelle doesn't yet support MODULE.bazel files). However, MODULE.bazel should be manually updated accordingly.


pjk25:
This can be merged once rules_erlang 3.9.1 is released and this branch is updated to use it.

PR #6960, opened 2023-01-19
Leave `rabbit` dependencies running after `rabbit:stop()`

dumbbell:
WIP


dumbbell:
Not stopping Mnesia and Ra as part of `rabbit:stop/0` led to issues with Ra clusters and it took me some time to figure out what was wrong.

On the surface, it looked like Ra clusters lost their data and were reinitialized from scratch in some testsuites, like `dynamic_qq_SUITE` where messages are published, the cluster is destroyed, a former follower is restarted alone and messages are verified. After that full node restart (i.e. a new Erlang VM), the messages were all lost.

The root cause was not the testcase actions themselves, but the initialization of the RabbitMQ cluster before that. The cluster is created by doing the following steps for each node except the first one:
1. `rabbitmqctl stop_app`; with this pull request, Mnesi and Ra continue to run
2. `rabbitmqctl join_cluster ...`
3. `rabbitmqctl start_app`

At the beginning of `join_cluster`, Mnesia is stopped, the database is reset and the content of the Mnesia directory is deleted. This gives a clean space to join the cluster.

Unfortunately, Ra files are stored under the Mnesia directory and they are deleted at the same time... The reset code doesn't check if the Ra systems are running or not. The solution is quite simple: in addition to make sure Mnesia is stopped before a reset, we also need to make sure Ra systems are stopped. At that point only files can be removed and `join_cluster` can proceed.

Ra systems are restarted as usual with `rabbitmqctl start_app`.

I submitted a pull request to Ra to [provide a `ra_system:stop/1` API](rabbitmq/ra#344). This will allow RabbitMQ to stop them as part of a database reset.

dumbbell:
Another issue discovered as part of this work is that `rabbit` doesn't stop the `rabbit_stream_coordinator` Ra server and the Osiris writer processes associated with stream queues. We rely on the fact that `ra` and `osiris` are stopped too to take their supervision trees down.

If `ra` and `osiris` are left running, the `rabbit_stream_coordinator` Ra server and the Osiris writer processes are still running, but the stream queues entry in the `rabbit_queue` Mnesia table are lost (the table lives only when `rabbit` runs). Therefore the processes are there but RabbitMQ doesn't know about these queues and the `recover` function for stream queues is a no-op (it relies on the fact that `rabbit_stream_coordinator` is restarted and takes care of starting all needed processes and add `rabbit_queue` table entries).

dumbbell:
I'm aborting this work for now. Too many issues which are difficult and risky to resolve.

PR #6977, opened 2023-01-20
Fix all dialyzer warnings in AWS-related plugins

binarin:


michaelklishin:
This has code changes I'd rather not get into `3.9.x` before its final patch release, so let's limit this to `3.11.x` and `3.10.x`.

Issue #6955, opened 2023-01-19
STOMP subscription to an exchange creates a non-exclusive queue


csicar:
# To reproduce

1. Create an exchange `doesexist`
2. Subscribe to it using STOMP: `SUBSCRIBE destination:/exchange/doesexist`

MWE, logs and dumps can be found here: https://gist.github.com/csicar/61512ea22ac2bb5ec81a11e0dc594fb5

# Expected behaviour

As per [stomp plugin documentation](https://www.rabbitmq.com/stomp.html#d.ed)

> For SUBSCRIBE frames, a destination of the form /exchange/<name>[/<pattern>] can be used. This destination creates an exclusive, auto-delete queue on <name> exchange

# Actual behaviour

RabbitMQ creates an `auto-delete` queue, but it is **not `exclusive`**.

Confirm by running `curl -u guest:guest http://localhost:15672/api/queues/`, which returns:
```json
[
  {
    "auto_delete": true,
    "exclusive": false,
    "name": "stomp-subscription-8CfOZeiDPGVGudK6jh_MJg",
    // -snip-
  }
]
```

## Why this matters

Authorization systems that rely on the resource name to decide whether to `allow` and `deny` have no good way of identifying whether a given client should be allowed to access a specific `stomp-subscription-*` queue. 
If the queues were exclusive, then `allow`ing every `stomp-subscription-*` would be safe (as only the creator can access the queue).

csicar:
I think the problem lies here:
https://github.com/rabbitmq/rabbitmq-server/blob/8bbe87ec9eb3261520f58790be0a2f9a427c25fd/deps/rabbitmq_stomp/src/rabbit_stomp_util.erl#L322-L332
- which is used in `build_params`: https://github.com/rabbitmq/rabbitmq-server/blob/8bbe87ec9eb3261520f58790be0a2f9a427c25fd/deps/rabbitmq_stomp/src/rabbit_stomp_util.erl#L308
- which itself is used in `ensure_endpoint` to ensure the queue exists: https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_stomp/src/rabbit_stomp_processor.erl#L1094-L1102

PR #6978, opened 2023-01-20
Fix incorrect default value for exchange subscription

csicar:
## Proposed Changes

Fix https://github.com/rabbitmq/rabbitmq-server/issues/6955 by choosing the correct default values for exchange subscriptions

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #6955)

## Checklist


michaelklishin:
This is a potentially breaking change and has been like this since before https://github.com/rabbitmq/rabbitmq-server/commit/856b0121440ab59ba08d556ce85b393850f5d03e (so, for at least more than 7 years).


We will limit it to `3.11.x`.

PR #7000, opened 2023-01-23
Fix all dialyzer warnings in peer discovery plugins

binarin:
Thoas update contains some fixes in type specs


michaelklishin:
@mergifyio rebase

PR #7001, opened 2023-01-23
Mark AMQP 1.0 properties chunk as binary

acogoluegnes:
This PR improves the support for AMQP 1.0 message format, especially in the context of streams:

* reuse AMQP 1.0 binary chunks if present before storing the message in the stream (when the original message is an AMQP 1.0 message)
* support the different types of bodies a message AMQP 1.0 can have (one AMQP value, several data or AMQP sequence sections)
* keep original order of sections


kjnilsson (changes requested):
A couple of nitpicks and a question otherwise LGTM!

kjnilsson (approved):

PR #7058, opened 2023-01-26
rabbit_nodes: Add list functions to clarify which nodes we are interested in

dumbbell:
So far, we had the following functions to list nodes in a RabbitMQ cluster:
* `rabbit_mnesia:cluster_nodes/1` to get members of the Mnesia cluster; the argument was used to select members (all members or only those running Mnesia and participating in the cluster)
* `rabbit_nodes:all/0` to get all members of the Mnesia cluster
* `rabbit_nodes:all_running/0` to get all members who currently run Mnesia

Basically:
* `rabbit_nodes:all/0` calls `rabbit_mnesia:cluster_nodes(all)`
* `rabbit_nodes:all_running/0` calls `rabbit_mnesia:cluster_nodes(running)`

We also have:
* `rabbit_node_monitor:alive_nodes/1` which filters the given list of nodes to only select those currently running Mnesia
* `rabbit_node_monitor:alive_rabbit_nodes/1` which filters the given list of nodes to only select those currently running RabbitMQ

Most of the code uses `rabbit_mnesia:cluster_nodes/1` or the `rabbit_nodes:all*/0` functions. `rabbit_mnesia:cluster_nodes(running)` or `rabbit_nodes:all_running/0` is often used as a close approximation of "all cluster members running RabbitMQ". This list might be incorrect in times where a node is joining the clustered or is being worked on (i.e. Mnesia is running but not RabbitMQ).

With Khepri, there won't be the same possible approximation because we will try to keep Khepri/Ra running even if RabbitMQ is stopped to expand/shrink the cluster.

So in order to clarify what we want when we query a list of nodes, this patch introduces the following functions:
* `rabbit_nodes:list_members/0` to get all cluster members, regardless of their state
* `rabbit_nodes:list_reachable/0` to get all cluster members we can reach using Erlang distribution, regardless of the state of RabbitMQ
* `rabbit_nodes:list_running/0` to get all cluster members who run RabbitMQ, regardless of the maintenance state
* `rabbit_nodes:list_serving/0` to get all cluster members who run RabbitMQ and are accepting clients

In addition to the list functions, there are the corresponding `rabbit_nodes:is_*(Node)` checks and `rabbit_nodes:filter_*(Nodes)` filtering functions.

The code is modified to use these new functions. One possible significant change is that the new list functions will perform RPC calls to query the nodes' state, unlike `rabbit_mnesia:cluster_nodes(running)`.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -379,7 +379,7 @@ rebalance(Type, VhostSpec, QueueSpec) ->
 maybe_rebalance({true, Id}, Type, VhostSpec, QueueSpec) ->
     rabbit_log:info("Starting queue rebalance operation: '~ts' for vhosts matching '~ts' and queues matching '~ts'",
                     [Type, VhostSpec, QueueSpec]),
-    Running = rabbit_maintenance:filter_out_drained_nodes_consistent_read(rabbit_nodes:all_running()),
+    Running = rabbit_maintenance:filter_out_drained_nodes_consistent_read(rabbit_nodes:list_running()),
     NumRunning = length(Running),
     ToRebalance = [Q || Q <- rabbit_amqqueue:list(),
                         filter_per_type(Type, Q),
@@ -1056,7 +1056,7 @@ check_queue_type(_Val, _Args) ->
 
 list() ->
     All = rabbit_db_queue:get_all(),
-    NodesRunning = rabbit_nodes:all_running(),
+    NodesRunning = rabbit_nodes:list_running(),
     lists:filter(fun (Q) ->
                          Pid = amqqueue:get_pid(Q),
                          St = amqqueue:get_state(Q),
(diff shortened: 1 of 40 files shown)


michaelklishin (approved):

dumbbell:
I will wait for #6821 to be merged before in order to not interfere with the review and rebase of that branch.

dumbbell:
This patch is ready for review again.

I hope it doesn't introduce any regression, it's difficult to test.

If you think of a better naming, please share!

michaelklishin:
It would greatly simplify backports if key functions of this API that can be aliases to the ones they replace in `main`, were added to `v3.11.x` and `v3.10.x`.

michaelklishin (approved):

michaelklishin:
#7282 is an example of a small change that failed to backport because of `rabbit_db_*`

dumbbell:
Perhaps it should not be backported in the first place? It doesn't look like a bug fix to me.

PR #7066, opened 2023-01-27
Add rabbitmq_cli dialyze to bazel

pjk25:
and fix a number of warnings

~~Because we stop mix from recompiling rabbit_common in bazel, many unknown functions are reported, so this dialyzer analysis is somewhat incomplete~~


pjk25:
Requires rules_erlang 3.9.2, which will soon be released

pjk25:
When this is mergable, I would advise the "Squash and merge" option.

PR #7108, opened 2023-01-30
Backport a CT broker helper fix from #5895 

michaelklishin:
Previously the helper was not setting the virtual host limit
correctly.


michaelklishin:
Something is up with BuildBuddy which cannot download a Gazelle release from GH. It's a tiny change that we really need in `v3.11.x` for the commercial edition => will merge.

PR #7144, opened 2023-02-01
Fix issue #7142

### LINKED ISSUE #7142: Issue with Limit Tab in 3.11.8
im using the latest RabbitMQ 3.11.8 in k8s.

im getting this error in the limit tab, tried just about everything which google have of solutions

rabbitmq-admin-limit

had the same issue for the user tab to start with, but a refresh of the browser cache fixed that, but cant seem to get it to work on limit tab this is my cm

```
enabled_plugins: |
[rabbitmq_federation,rabbitmq_management,rabbitmq_peer_discovery_k8s,rabbitmq_prometheus].
rabbitmq.conf: |
loopback_users.guest = false
listeners.tcp.default = 5672
cluster_formation.peer_discovery_backend = k8s
cluster_formation.k8s.host = kubernetes.default.svc.cluster.local
cluster_formation.k8s.address_type = hostname
```

ref to issue created in rabbitmq:docker

https://github.com/docker-library/rabbitmq/issues/603

issue replicated by another user, so it shouldt be related to running containerized

MarcialRosales:
Fixes #7142 

The issue is that users retrieved with the intention to list in the limits view are not paged hence they are not wrapped
around a paging struct where users would be under items attribute. The limits template thinks that users are paged and when it executes `users.items.length` , it cannot find `items`.

Pending selenium tests

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/tmpl/limits.ejs
+++ b/deps/rabbitmq_management/priv/www/js/tmpl/limits.ejs
@@ -1,6 +1,6 @@
 <h1>Limits</h1>
 
-<div class="section">
+<div class="section" id="virtual-host-limits">
     <h2>Virtual host Limits</h2>
     <div class="hider">
         <div class="updatable">
@@ -49,7 +49,7 @@
     </div>
 </div>
 
-<div class="section administrator-only">
+<div class="section administrator-only" id="set-virtual-host-limits">
     <h2>Set / update a virtual host limit</h2>
     <div class="hider">
         <form action="#/limits" method="put">
@@ -88,7 +88,7 @@
     </div>
 </div>
 
-<div class="section">
+<div class="section" id="user-limits">
   <h2>User Limits</h2>
   <div class="hider">
       <div class="updatable">
@@ -137,7 +137,7 @@
   </div>
 </div>
 
-<div class="section administrator-only">
+<div class="section administrator-only" id="set-user-limits">
   <h2>Set / update a user limit</h2>
   <div class="hider">
       <form action="#/user-limits" method="put">
@@ -146,9 +146,9 @@
                 <th><label>User:</label></th>
                 <td>
                   <select name="user">
-                      <% for (var i = 0; i < users.items.length; i++) { %>
-                      <option value="<%= fmt_string(users.items[i].name) %>">
-                          <%= fmt_string(users.items[i].name) %>
+                      <% for (var i = 0; i < users.length; i++) { %>
+                      <option value="<%= fmt_string(users[i].name) %>">
+                          <%= fmt_string(users[i].name) %>
                       </option>
                       <% } %>
                   </select>
--- a/deps/rabbitmq_management/selenium/suites/limits.sh
+++ b/deps/rabbitmq_management/selenium/suites/limits.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+SCRIPT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Name of the suite used to generate log and screen folders
+SUITE=$( basename "${BASH_SOURCE[0]}" .sh)
+
+# Path to the test cases this suite should run. It is relative to the selenium/test folder
+TEST_CASES_PATH=/limits
+# Path to the folder where all configuration file reside. It is relative to the selenim/test folder
+TEST_CONFIG_PATH=/basic-auth
+
+source $SCRIPT/suite_template
+
+_setup () {
+  start_rabbitmq
+}
+_save_logs() {
+  save_container_logs rabbitmq
+}
+_teardown() {
+  kill_container_if_exist rabbitmq
+}
+run
--- a/deps/rabbitmq_management/selenium/test/definitions/Makefile
+++ b/deps/rabbitmq_management/selenium/test/definitions/Makefile
@@ -10,5 +10,11 @@ RABBITMQ_SERVER_ROOT = ../../../../../
 help:
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
+start-rabbitmq: ## Start RabbitMQ
+	@(docker kill rabbitmq >/dev/null 2>&1  && docker rm rabbitmq)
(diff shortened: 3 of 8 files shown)


lukebakken (approved):
Great, I just started via `make run-broker` with this patch and the issue is resolved. I also set and cleared the `max-connections` limit for the `guest` user and `/` vhost :shipit:

michaelklishin:
@MarcialRosales @lukebakken do we know if `v3.10.x` is affected by #7142?

kuPyxa:
> @MarcialRosales @lukebakken do we know if `v3.10.x` is affected by #7142?

3.9.28 has this bug too.

MarcialRosales:
Yes, v3.10.x is also affected.  cc @michaelklishin @kuPyxa

mirobertod:
The `v3.10.18` has this bug too. Do we an ETA for when the bug will be fixed for 3.10.x?
I see that the `v3.11.9`, which include the fix, has been released the same date of `v3.10.18`

michaelklishin:
New releases are shipped every few weeks, there is no fixed schedule.

GBreply:
Hi all, are you aware that version 3.11.10 has the same proble for Admin/Policies:
image
? If so, do you have a milestone (e.g. 3.11.12) for the fix?

kuPyxa:
Maybe something got cached. Try Ctrl + F5.

illotum:
I can confirm this issue affecting 3.10, and considering other reports, perhaps this need to be tagged for backporting to 3.10 and 3.9?

michaelklishin:
There will be no more 3.9 releases.

michaelklishin:
This was backported to `v3.10.x` in https://github.com/rabbitmq/rabbitmq-server/pull/7706

eduardomb08:
I'm having the same issue for Users and Policies after upgrading a cluster to 3.11.13. It's funny that I updated a different cluster to the same version and that one is working just fine. PS: I'm using the Kubernetes RabbitMQ Operator.

image

michaelklishin:
@eduardomb08 clear all local storage and cache for the management UI.

eduardomb08:
It seems the discussion was locked.. so I thought registering my comment here would be the best way to contribute.

All of the systems and frameworks I worked with have feature flags working as switches, and not irreversible changes. That being said, I don't remember working with feature flags in any other distributed system.

I don't really care about the name. But the behavior should be clear in the UI. If enabling a feature flag is designed to be irreversible in RabbiqMQ, it would be nice to have not only a callout to that fact in the UI, but also a confirmation dialog or something when clicking the Enable button.

Lastly, irreversible flags will make it impossible for K8S clusters to rollback upgrades. Given that the upgrade worked without any issues in one of my clusters, but failed in another, no amount of testing in the first environment would guarantee the second would work as well. It that is the case, I believe anyone would feel extremely insecure (to say the least), to turn a feature flag ON in a cluster. Your idea of a CLI command to override the value seems a step in the direction to allowing rollbacks, but still now clear how that would happen, since all of that is abstracted by K8S.

@michaelklishin , given that changes may work in a cluster but not in another, how can a change be safely tested in a separate cluster and then safely to production?

PS: No emotions here besides the frustration of some unexpected rework. And I'm trying to push for any change here, but I would love to know how to make it work since I love and recommend RabbitMQ.

PS2: I was finally able to revert to v3.9, but lost stuff. I'm just thankful that this happened between two of my test clusters and not between real environments.

michaelklishin:
@eduardomb08 you keep assuming that RabbitMQ supports downgrades. It does not, in multiple places. Besides full disk snapshots on all cluster nodes, there is no way to roll back an upgrade, on Kubernetes or not.

Some feature flags can be semantically disabled but their purpose is to enable rolling upgrades, during which a cluster becomes mixed version (e.g. RabbitMQ 3.11 nodes run  alongside a group of 3.10 nodes, all the way until all nodes are upgraded). Some
feature releases [require older feature flags to be enabled](https://blog.rabbitmq.com/posts/2022/07/required-feature-flags-in-rabbitmq-3.11) before
the upgrade. In RabbitMQ, feature flags are not a mechanism that supports reversion
in the vast majority of cases. From [our guide on Feature flags](https://rabbitmq.com/feature-flags.html#how-to-disable-feature-flags):

```
It is impossible to disable a feature flag once it is enabled.
```

[Blue/Green upgrades](https://rabbitmq.com/blue-green-upgrade.html) is the safest upgrade option because the original cluster is not going away.

On Kubernetes specifically, trying out a feature flag in a new cluster should not be difficult. In fact, in new clusters all feature flags common across all nodes are usually automatically enabled, so you don't have to do that manually.

`RABBITMQ_FEATURE_FLAGS` can be used to [start a (new) node with only a subset of the flags](https://rabbitmq.com/feature-flags.html#how-to-start-new-node-disabled-feature-flags).

eduardomb08:
> @eduardomb08 you keep assuming that RabbitMQ supports downgrades. It does not, in multiple places. Besides full disk snapshots on all cluster nodes, there is no way to roll back an upgrade, on Kubernetes or not.
> 
> Some feature flags can be semantically disabled but their purpose is to enable rolling upgrades, during which a cluster becomes mixed version (e.g. RabbitMQ 3.11 nodes run alongside a group of 3.10 nodes, all the way until all nodes are upgraded). Some feature releases [require older feature flags to be enabled](https://blog.rabbitmq.com/posts/2022/07/required-feature-flags-in-rabbitmq-3.11) before the upgrade. In RabbitMQ, feature flags are not a mechanism that supports reversion in the vast majority of cases. From [our guide on Feature flags](https://rabbitmq.com/feature-flags.html#how-to-disable-feature-flags):
> 
> ```
> It is impossible to disable a feature flag once it is enabled.
> ```
> 
> [Blue/Green upgrades](https://rabbitmq.com/blue-green-upgrade.html) is the safest upgrade option because the original cluster is not going away.
> 
> On Kubernetes specifically, trying out a feature flag in a new cluster should not be difficult. In fact, in new clusters all feature flags common across all nodes are usually automatically enabled, so you don't have to do that manually.
> 
> `RABBITMQ_FEATURE_FLAGS` can be used to [start a (new) node with only a subset of the flags](https://rabbitmq.com/feature-flags.html#how-to-start-new-node-disabled-feature-flags).

Thanks for the clarification, @michaelklishin! That's very helpful.

Given you last comment, I believe the UI changes I suggested become even more important. An accidental click could cause great disruption.

Again, thank you very much!

goldenbull:
same bug in  `3.12.2`

image

goldenbull:
same bug in `3.12.2`
image

PR #7148, opened 2023-02-01
Use maybe expression in MQTT plugin

ansd:
This commit is pure refactoring making the code base more maintainable.

Replace rabbit_misc:pipeline/3 with the new OTP 25 experimental maybe expression because
"Frequent ways in which people work with sequences of failable operations include folds over lists of functions, and abusing list comprehensions. Both patterns have heavy weaknesses that makes them less than ideal."
https://www.erlang.org/eeps/eep-0049#obsoleting-messy-patterns

Additionally, this commit is more restrictive in the type spec of rabbit_mqtt_processor state fields.
Specifically, many fields were defined to be `undefined | T` where `undefined` was only temporarily until the first CONNECT packet was processed by the processor.
It's better to initialise the MQTT processor upon first CONNECT packet because there is no point in having a processor without having received any packet.
This allows many type specs in the processor to change from `undefined | T` to just `T`.
Additionally, memory is saved by removing the `received_connect_packet` field from the `rabbit_mqtt_reader` and `rabbit_web_mqtt_handler`.

### DIFF
--- a/deps/rabbitmq_mqtt/src/rabbit_mqtt_packet.erl
+++ b/deps/rabbitmq_mqtt/src/rabbit_mqtt_packet.erl
@@ -37,9 +37,9 @@ parse(<<>>, authenticated) ->
     {more, fun(Bin) -> parse(Bin, authenticated) end};
 parse(<<MessageType:4, Dup:1, QoS:2, Retain:1, Rest/binary>>, authenticated) ->
     parse_remaining_len(Rest, #mqtt_packet_fixed{ type   = MessageType,
-                                                  dup    = bool(Dup),
+                                                  dup    = int_to_bool(Dup),
                                                   qos    = QoS,
-                                                  retain = bool(Retain) });
+                                                  retain = int_to_bool(Retain) });
 parse(<<?CONNECT:4, 0:4, Rest/binary>>, unauthenticated) ->
     parse_remaining_len(Rest, #mqtt_packet_fixed{type = ?CONNECT});
 parse(Bin, Cont)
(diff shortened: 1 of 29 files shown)


ansd:
The failures in `processor_SUITE` are expected for now and will be fixed when adding an `erl_flags` attribute to `ct_run` Erlang rules.
A workaround (not done in this PR) would be: `bazel test //deps/rabbitmq_mqtt:processor_SUITE --test_env FOCUS="-enable-feature maybe_expr"`

deps/rabbitmq_mqtt/src/rabbit_mqtt_packet.erl
```diff
@@ -131,10 +131,10 @@ parse_connect(Bin, Fixed, Length) ->
                             wrap(Fixed,
                                  #mqtt_packet_connect{
                                     proto_ver   = ProtoVersion,
-                                    will_retain = bool(WillRetain),
+                                    will_retain = int_to_bool(WillRetain),
                                     will_qos    = WillQos,
-                                    will_flag   = bool(WillFlag),
-                                    clean_sess  = bool(CleanSession),
+                                    will_flag   = int_to_bool(WillFlag),
+                                    clean_sess  = int_to_bool(CleanSession),
```
ChunyiLyu, line 137: Is there any benefit of having `CleanSession`, `WillFlag `, and `WillRetain` as boolean? If you use the integer value directly, maybe you could avoid converting int <=> bool.

ChunyiLyu (commented):
Awesome! Lots of improvements in this one. Just one small comment.

deps/rabbitmq_mqtt/src/rabbit_mqtt_packet.erl
ansd: Technically, one could use an integer instead of boolean (atom) throughout the code base. However, this conversion is extremely cheap. The benefit of having booleans (atoms) is much more readable code as for example in:
https://github.com/rabbitmq/rabbitmq-server/blob/41f98ccbefddaddb6374d69a0c8eb8af746d69ee/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl#L430
https://github.com/rabbitmq/rabbitmq-server/blob/41f98ccbefddaddb6374d69a0c8eb8af746d69ee/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl#L1241-L1246

and clearer log output as for example in:
https://github.com/rabbitmq/rabbitmq-server/blob/41f98ccbefddaddb6374d69a0c8eb8af746d69ee/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl#L123-L125
printing `false` instead of `0`.

ChunyiLyu (approved):

michaelklishin (dismissed):

Issue #7110, opened 2023-01-31
Target QQ replica count policy


SimonUnge:
Currently rabbitmq only stores the `x-quorum-initial-group-size` - or defaults to
3 if missing. But this value can not be changed.

Suggestion: Add an policy `target-qq-replica-count <size>`

This value can be used for future features such as automatically add node to QQ group when a node is added to a
cluster, if the current group size is less than target-qq-replica count.

Register as operator policy and add to default_policy.operator in Cuttlefish

michaelklishin:
Sounds reasonable. We have a target node count hint in `rabbitmq.conf` for similar reasons but only use it lightly in some tooling at the moment.

michaelklishin:
Team RabbitMQ decided that this cannot be done in a "one step at a time" manner and needs a single issue that will serve as a design document. I will put together something that would capture our current understanding of the goal of dynamic QQ replica management with the goal of maintaining an operator-set minimum.

PR #7150, opened 2023-02-01
See #7110. Target QQ replica count policy

SimonUnge:
## Proposed Changes

Add policy and operator policy for target qq replica count.
See #7110 

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -206,7 +206,7 @@ start_cluster(Q) ->
     Arguments = amqqueue:get_arguments(Q),
     Opts = amqqueue:get_options(Q),
     ActingUser = maps:get(user, Opts, ?UNKNOWN_USER),
-    QuorumSize = get_default_quorum_initial_group_size(Arguments),
+    QuorumSize = get_default_quorum_initial_group_size(Arguments, Q),
     RaName = case qname_to_internal_name(QName) of
                  {ok, A} ->
                      A;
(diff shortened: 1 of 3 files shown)


michaelklishin:
This adds a policy key and a configuration options but they are not used. Is the plan to use these values in later PRs? We can only merge such PRs into `main` then.

SimonUnge:
Right, meant to be used for automatic qq group growth one that is figured out.

Would it make sense to add a flag to the grow command too, all|even| and something like "target"?

michaelklishin:
There is no consensus on our team at the moment about this set of features, so it will have to wait.

SimonUnge:
I was thinking of adding a little logic to the target-qq-replica-count, that if it is present on queue creation, use the max value of target-qq-replica-count, and x-quorum-initial-group-size. That is, update `get_default_quorum_initial_group_size` to check if a policy is present.

SimonUnge:
Updated PR with logic taking policy into account on QQ creation

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1581,12 +1609,18 @@ quorum_ctag(Other) ->
...
+        {undefined, undefined} ->
             application:get_env(rabbit, quorum_cluster_size, 3);
-        {_Type, Val} ->
-            Val
+        {undefined, V} ->
+            V;
+        {{_Type, V}, undefined} ->
+            V;
+        {{_Type, ArgV}, PolV} ->
+            max(ArgV, PolV)
```
SimonUnge, line 1623: @michaelklishin @illotum 

Question: Should Policy have precedence over x-quorum-initial-group-size?

michaelklishin: Optional arguments historically have taken precedence. I now recall that we did not expose `quorum-initial-group-size` via policies intentionally at first. @kjnilsson do you remember why?

michaelklishin:
Team RabbitMQ decided that this cannot be done in a "one step at a time" manner and needs a single issue that will serve as a design document. I will put together something that would capture our current understanding of the goal of dynamic QQ replica management with the goal of maintaining an operator-set minimum.

PR #7155, opened 2023-02-02
CQ: Don't confirm on remove from old message store

lhoguin:
<!> PENDING NUMEROUS BENCHMARKS TO CONFIRM ASSUMPTIONS <!>

The confirms will instead be sent when the message store syncs to disk. The problem with doing confirms on remove is that it happens far too often and it brings down the overall performance despite sending the confirms earlier.


lhoguin:
This is generally good but not good enough to put into main yet. I will do a more thorough rework of store confirms in the lh-msg-store branch.

Issue #7131, opened 2023-01-31
Lock operator policies from HTTP API


illotum:
Since in certain environments the infrastructure operators do not get accounts _in_ RabbitMQ, I'd like to add a config switch that prohibits operator policies from being changed via HTTP API. Tentatively `prohibit_remote_operator_policies=true`

Broker will return a well-formed error to `PUT /api/operator-policies`, and the only way to set them becomes by accessing the host: either via rabbitmqctl, or #6687.

Perhaps extra work may be done to hide operator policies from 'policymaker's in the web console as well.

---

I briefly considered adding instead a new type of policies, say "global", to enforce cross-vhost limits. This feels a bit redundant, but I'm open to all arguments.

michaelklishin:
Disabling this kind of feature is fine.

`operator_policies.enabled = (boolean)` is what I'd use.

illotum:
How about then migrating `default_policies.operator` into `operator_policies.default`? With a bit of data transformation cuttlefish allows both mappings to co-exist.

michaelklishin:
Addressed in #7165.

PR #7165, opened 2023-02-03
HTTP API: make it possible to restrict operator policy changes via API

illotum:
## Proposed Changes

This change adds a `rabbitmq_management` setting to disable operator policy edits, per #7131.

```
management.restrictions.operator_policy_changes.disabled = true
```

Additionally, when the setting is enabled, it hides relevant web console elements and returns a proper error in rabbitmqadmin. I wasn't sure how to proceed with the testing for these two targets, so these changes are not covered.

And I'm open to suggestions about the setting name of course.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbitmq_management/priv/schema/rabbitmq_management.schema
+++ b/deps/rabbitmq_management/priv/schema/rabbitmq_management.schema
@@ -528,3 +528,10 @@ fun(File) ->
     ReadFile = file:list_dir(File),
     element(1, ReadFile) == ok
 end}.
+
+%% Disables setting operator policies over API.
+
+{mapping, "management.restrictions.operator_policy_changes.disabled", "rabbitmq_management.restrictions.operator_policy_changes.disabled", [
+    {datatype, {enum, [true, false]}},
+    {include_default, false}
+]}.
--- a/deps/rabbitmq_management/src/rabbit_mgmt_wm_operator_policy.erl
+++ b/deps/rabbitmq_management/src/rabbit_mgmt_wm_operator_policy.erl
@@ -72,7 +72,12 @@ delete_resource(ReqData, Context = #context{user = #user{username = Username}}) 
     {true, ReqData, Context}.
 
 is_authorized(ReqData, Context) ->
-    rabbit_mgmt_util:is_authorized_admin(ReqData, Context).
+    case rabbit_mgmt_features:is_edit_op_policy_disabled() of
+        true ->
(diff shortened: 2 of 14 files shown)


michaelklishin (changes requested):
Please use [an enum of (`true`, `false`)](https://github.com/rabbitmq/rabbitmq-server/blob/ae8a1bd1ab3a7ac4ca89a505d58bebebd2aca77b/deps/rabbitmq_management/priv/schema/rabbitmq_management.schema#L36-L37) for such boolean types.

illotum:
Ah, I considered `{flag, true, false}`. That would work identical if I understand it right, is `enum` preferred for a particular reason?

michaelklishin:
We do not use `flag` anywhere and I had no idea Cuttlefish supports it.

deps/rabbitmq_management/src/rabbit_mgmt_wm_operator_policy.erl
```diff
@@ -72,7 +72,12 @@ delete_resource(ReqData, Context = #context{user = #user{username = Username}})
     {true, ReqData, Context}.
 
 is_authorized(ReqData, Context) ->
-    rabbit_mgmt_util:is_authorized_admin(ReqData, Context).
+    case rabbit_mgmt_util:is_edit_op_policy_disabled() of
```
michaelklishin, line 75: I think such restrictions should live in a separate module, r.g. `rabbit_mgmt_features`.

deps/rabbitmq_management/priv/schema/rabbitmq_management.schema
```diff
@@ -528,3 +528,10 @@ fun(File) ->
     ReadFile = file:list_dir(File),
     element(1, ReadFile) == ok
 end}.
+
+%% Disables setting operator policies over API.
+
+{mapping, "management.disable_operator_policy_edit", "rabbitmq_management.disable_operator_policy_edit", [
```
michaelklishin, line 534: I'd use `management.restrictions.operator_policy_changes.disabled` and `rabbitmq_management.restrictions.operator_policy_changes.disabled`. It'd be a bit more annoying to access this nested proplist but you'd have to do it only once in one predicate function.

illotum:
Addressed feedback & rebased, ready for review.

illotum:
Addressed feedback & rebased, ready for review.

michaelklishin (approved):

PR #7175, opened 2023-02-03
Quorum queues: chunk consumer log effects

kjnilsson:
Chunk quorum queue deliveries

This puts a limit to the amount of message data that is added
to the process heap at the same time to around 128KB.

Large prefetch values combined with large messages could cause
excessive garbage collection work.

Also simplify the intermediate delivery message format to avoid
allocations that aren't necessary.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -72,7 +72,8 @@
         ]).
 
 -ifdef(TEST).
--export([update_header/4]).
+-export([update_header/4,
+         chunk_disk_msgs/3]).
 -endif.
 
 %% command records representing all the protocol actions that are supported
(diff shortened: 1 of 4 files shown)


deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -1978,18 +2000,20 @@ get_next_msg(#?MODULE{returns = Returns0,
             Msg
     end.
 
-delivery_effect({CTag, CPid}, [{Idx, {MsgId, Header}}],
+delivery_effect({CTag, CPid}, [{MsgId, [Idx | Header]}],
```
ansd, line 2003: Use `?MSG(Idx, Header)` instead of `[Idx | Header]` ?

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -1942,12 +1942,34 @@ evaluate_limit(Index, Result, BeforeState,
...
+                [CurChunk | Chunks])
+  when Bytes >= ?DELIVERY_CHUNK_LIMIT_B ->
+    Size = get_header(size, Header),
+    chunk_disk_msgs(Rem, Size, [[], [Msg | CurChunk] | Chunks]);
+chunk_disk_msgs([{_MsgId, ?MSG(_RaftIdx, Header)} = Msg | Rem], Bytes,
+                [CurChunk | Chunks]) ->
+    Size = get_header(size, Header),
+    chunk_disk_msgs(Rem, Bytes + Size, [[Msg | CurChunk] | Chunks]).
+
+append_delivery_effects(Effects0, AccMap, _State)
```
ansd, line 1962: Not part of this PR, but I guess, we should rename `append_delivery_effects` to `prepend_delivery_effects` since the delivery effects are prepended, right?

deps/rabbit/src/rabbit_fifo.erl
kjnilsson: cheers done

ansd (approved):

michaelklishin:
It was decided to ship this only in `3.12.0` and not to backport.

PR #7177, opened 2023-02-03
Add OAuth 2.0 MQTT system test

ansd:
Add a test that rabbitmq_auth_backend_oauth2 works with MQTT.
See https://github.com/rabbitmq/rabbitmq-oauth2-tutorial#mqtt-protocol
We should add a similar test for AMQP 1.0 given https://github.com/rabbitmq/rabbitmq-server/pull/6931 is merged.


michaelklishin:
We cannot yet backport this to `v3.10.x` as it doesn't have all the modern OAuth 2 bits.

PR #7196, opened 2023-02-06
Fix all dependencies for the dialyzer

binarin:
This is the latest commit in the series, it fixes (almost) all the problems with missing and circular dependencies for typing.

The only 3 unsolved problems are:

- `lg` dependency for `rabbit` - the problem is that it's the only dependency that contains NIF. And there is no way to make dialyzer ignore it - looks like unknown check is not suppressable by dialyzer directives. In the future making `lg` a proper dependency can be a good thing anyway.

- some missing elixir function in `rabbitmq_cli` (CSV, JSON and logging related).

- `eetcd` dependency for `rabbitmq_peer_discovery_etcd` - this one uses sub-directories in `src/`, which confuses dialyzer (or our bazel machinery is not able to properly handle it). I've tried the latest rules_erlang which flattens directory for .beam files, but it wasn't enough for dialyzer - it wasn't able to find core erlang files. This is a niche plugin and an unusual dependency, so probably not worth investigating further.


pjk25:
The `eetcd` issue should be fixed by re-importing the dep with rules_erlang `3.9.5`, but I will have to verify that.

PR #7212, opened 2023-02-08
WIP: Target quorum queue replica count

SimonUnge:
## Proposed Changes
See https://github.com/rabbitmq/rabbitmq-server/issues/7209

Request for feedback PR!

This PR contains a policy target-qq-replica-count, and some logic to automatically
grow a QQ if its size is below target-qq-replica-count. 

The check is periodical, and piggybacks on the already present handle_tick function for quorum queues.
To avoid the check running every 5 seconds, I just added a count that increases per tick, to increase the
likelihood the check is performed (On the 12th tick the probability is 1.0).

The count value is stored as part of the Queue state, and checked and updated per tick, per QQ.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

I have not run full suite of tests etc, I just want some input in this PR for now!

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -743,7 +743,7 @@ end}.
     {datatype, integer}
 ]}.
 
-{mapping, "default_policies.operator.$id.target_qq_replica_count", "rabbit.default_policies.operator", [
+{mapping, "default_policies.operator.$id.target_group_size", "rabbit.default_policies.operator", [
     {include_default, 1},
     {commented, 1},
     {validators, ["non_zero_positive_integer"]},
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -743,6 +743,13 @@ end}.
     {datatype, integer}
 ]}.
 
+{mapping, "default_policies.operator.$id.target_qq_replica_count", "rabbit.default_policies.operator", [
+    {include_default, 1},
+    {commented, 1},
+    {validators, ["non_zero_positive_integer"]},
+    {datatype, integer}
+]}.
+
 {translation, "rabbit.default_policies.operator", fun(Conf) ->
     Props = rabbit_cuttlefish:aggregate_props(Conf, ["default_policies", "operator"]),
     Props1 = lists:map(
--- a/deps/rabbit/src/amqqueue.erl
+++ b/deps/rabbit/src/amqqueue.erl
@@ -117,6 +117,7 @@
           options = #{} :: map() | '_',
           type = ?amqqueue_v1_type :: module() | '_',
           type_state = #{} :: map() | '_',
+          %% New state value to store num of ticks
           tick_count = 0
          }).
 
--- a/deps/rabbit/src/amqqueue.erl
+++ b/deps/rabbit/src/amqqueue.erl
@@ -70,6 +70,8 @@
          set_sync_slave_pids/2,
          get_type/1,
          get_vhost/1,
+         get_tick_count/1,
+         set_tick_count/2,
          is_amqqueue/1,
          is_auto_delete/1,
          is_durable/1,
@@ -114,7 +116,8 @@
           vhost :: rabbit_types:vhost() | undefined | '_', %% secondary index
           options = #{} :: map() | '_',
           type = ?amqqueue_v1_type :: module() | '_',
-          type_state = #{} :: map() | '_'
+          type_state = #{} :: map() | '_',
+          tick_count = 0
          }).
 
 -type amqqueue() :: amqqueue_v2().
@@ -472,6 +475,12 @@ get_type_state(#amqqueue{type_state = TState}) ->
     TState;
 get_type_state(_) ->
     #{}.
+
+get_tick_count(#amqqueue{tick_count = Count}) ->
+    Count.
+
+set_tick_count(State, Count) ->
+    State#amqqueue{tick_count = Count}.
 
 -spec set_type_state(amqqueue(), map()) -> amqqueue().
 set_type_state(#amqqueue{} = Queue, TState) ->
--- a/deps/rabbit/src/amqqueue.erl
+++ b/deps/rabbit/src/amqqueue.erl
@@ -70,8 +70,6 @@
          set_sync_slave_pids/2,
          get_type/1,
          get_vhost/1,
-         get_tick_count/1,
-         set_tick_count/2,
          is_amqqueue/1,
          is_auto_delete/1,
          is_durable/1,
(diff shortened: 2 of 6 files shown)


deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -743,6 +743,13 @@ end}.
     {datatype, integer}
 ]}.
 
+{mapping, "default_policies.operator.$id.target_qq_replica_count", "rabbit.default_policies.operator", [
```
michaelklishin, line 746: We can use just "target_replica_count".

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -111,6 +115,30 @@
 -define(ADD_MEMBER_TIMEOUT, 5000).
 -define(SNAPSHOT_INTERVAL, 8192). %% the ra default is 4096
 
+%%----------- QQ policies ---------------------------------------------------
+
+-rabbit_boot_step(
+   {?MODULE,
+    [{description, "QQ policy validation"},
+     {mfa, {rabbit_registry, register,
+            [policy_validator, <<"target-qq-replica-count">>, ?MODULE]}},
```
michaelklishin, line 124: We can use just "target_replica_count". It will be applicable to streams, too.

kjnilsson: Given we have used groups-size in other config keys should it be `target_group_size`?

kjnilsson (changes requested):
I think we're on the right track assuming we agree async growth is the approach we should take. There are a few immediate things to address and I need to ponder options around one particular aspect.

deps/rabbit/src/amqqueue.erl
```diff
@@ -114,7 +116,9 @@
           vhost :: rabbit_types:vhost() | undefined | '_', %% secondary index
           options = #{} :: map() | '_',
           type = ?amqqueue_v1_type :: module() | '_',
-          type_state = #{} :: map() | '_'
+          type_state = #{} :: map() | '_',
+          %% New state value to store num of ticks
+          tick_count = 0
```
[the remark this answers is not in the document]
SimonUnge: I figured there might be restrictions to update the record, used to for some proof of concept.

What would be an appropriate way to limit the check occurrences? I could store it in an ets table, or process dict(hm). Or have some other indicator other than a tick.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -488,14 +517,57 @@ handle_tick(QName,
...
 
+maybe_grow_qq_members(QName) ->
+    {ok, Q} = rabbit_amqqueue:lookup(QName),
+    Count = amqqueue:get_tick_count(Q),
+    NewCount = case should_check_if_grow(Count) of
+                   false ->
+                       Count + 1;
+                   true ->
+                       Running = rabbit_nodes:all_running(),
+                       #{nodes := MemberNodes} = amqqueue:get_type_state(Q),
```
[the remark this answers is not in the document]
SimonUnge: OK! Good to know. I'll try to understand the callback functionality used a bit more on my side.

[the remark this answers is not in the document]
SimonUnge: I noticed that rabbit_stream_coordinator uses ra:members(...), which I assume is more reliable?

deps/rabbit/src/rabbit_quorum_queue.erl
SimonUnge: Going with target_group_size for now then!

deps/rabbit/src/amqqueue.erl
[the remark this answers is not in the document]
SimonUnge: I noticed there is a 'last_active' time stamp, which at least is something I can use for comparison, so now I rely on last_activity and current system time to calculate interval.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -488,14 +518,58 @@ handle_tick(QName,
...
+                    %% Take node load or availability zones into account?
+                    add_members(Q, NewNodes, 5000, Size - CurrentSize);
+                {_,_} ->
+                    %%Target size already achieved
+                    ok
+            end
+    end.
+
+%% Return true every ~ 30 sec
+should_check_if_grow(LastActive) ->
```
SimonUnge, line 564: @kjnilsson How reliable is this LastActive data, in terms of 'always having a value'?
If we can use this value reliably, it can be used in a way similar to this to control interval!

SimonUnge: I assume last_active will update when for example a queue gets a new policy. But that is perhaps OK, it will 'reset' the interval check is all.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -488,14 +519,73 @@ handle_tick(QName,
...
+    {ok, Q} = rabbit_amqqueue:lookup(QName),
+    % Rely on LastActive and compare it with system_time, to figure out intervall
+    case check_if_grow(LastActive) of
+        false ->
+            rabbit_log:info(">>>> TICK do nothing",[]),
+            ok;
+        true ->
+            rabbit_log:info(">>>> TICK check members",[]),
+            Running = rabbit_nodes:all_running(),
+            {ok, Members, _} = ra:members(amqqueue:get_pid(Q)),
```
SimonUnge, line 543: @kjnilsson Using ra:members() instead of type_state. But if we could get this info in the tick, all for that!

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -488,14 +519,73 @@ handle_tick(QName,
...
+                length([Q || Q <- rabbit_amqqueue:list_by_type(quorum),
+                             amqqueue:get_state(Q) =/= crashed,
+                             lists:member(Node, rabbit_amqqueue:get_quorum_nodes(Q))])
+        end,
+    lists:keysort(
+          2,
+          [{Node, QueueLenFun(Node)} || Node <- Nodes]).
+
+%% Return true every aprox GROW_CHECK_TIMEOUT sec
+check_if_grow(LastActive) ->
```
SimonUnge, line 578: @michaelklishin @kjnilsson 
OK, so my idea here is to instead use 'some' timestamp as a reference, and since a queue has a last_active timestamp, I use that. With it I can calculate an approximate interval check. Approximate as in, if you have say a TICK_TIMEOUT of 4 seconds, and you set GROW_CHECK_TIMEOUT to 30 second, the check will be every 32 second, as I round 'up' to nearest multiple of TICK_TIMEOUT to GROW_CHECK_TIMEOUT

SimonUnge: The last_active time stamp does get reset when something happens for the queue, such as giving it a policy. So adding a policy to existing queues might trigger the interval check a bit early. But for adding nodes etc it should be true to the interval config

PR #7219, opened 2023-02-08
Require all feature flags introduced before 3.11.1

ansd:
RabbitMQ 3.12 requires feature flag `feature_flags_v2` which got introduced in 3.11.0 (see https://github.com/rabbitmq/rabbitmq-server/pull/6810).

Therefore, we can mark all feature flags that got introduced in 3.11.0
or before 3.11.0 as required because users will have to upgrade to
3.11.x first, before upgrading to 3.12.x

The advantage of marking these feature flags as required is that we can
start deleting any compatibliy code for these feature flags, similarly
as done in https://github.com/rabbitmq/rabbitmq-server/issues/5215

This list shows when a given feature flag was first introduced:

```
classic_mirrored_queue_version 3.11.0
stream_single_active_consumer 3.11.0
direct_exchange_routing_v2 3.11.0
listener_records_in_ets 3.11.0
tracking_records_in_ets 3.11.0

empty_basic_get_metric 3.8.10
drop_unroutable_metric 3.8.10
```

In this commit, we also force all required feature flags in Erlang
application `rabbit` to be enabled in mixed version cluster testing
and delete any tests that were about a feature flag starting as disabled.

Furthermore, this commit already deletes the callback (migration) functions
given they do not run anymore in 3.12.x.

All other clean up (i.e. branching depending on whether a feature flag
is enabled) will be done in separate commits.

### DIFF
--- a/deps/rabbit/src/rabbit_core_ff.erl
+++ b/deps/rabbit/src/rabbit_core_ff.erl
@@ -7,20 +7,11 @@
 
 -module(rabbit_core_ff).
 
--include_lib("kernel/include/logger.hrl").
-
--include_lib("rabbit_common/include/logging.hrl").
-
--export([direct_exchange_routing_v2_enable/1,
-         listener_records_in_ets_enable/1,
-         listener_records_in_ets_post_enable/1,
-         tracking_records_in_ets_enable/1,
-         tracking_records_in_ets_post_enable/1]).
-
 -rabbit_feature_flag(
    {classic_mirrored_queue_version,
     #{desc          => "Support setting version for classic mirrored queues",
-      stability     => stable
+      %%TODO remove compatibility code
+      stability     => required
      }}).
 
 -rabbit_feature_flag(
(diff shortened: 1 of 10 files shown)


deps/rabbit/src/rabbit_core_ff.erl
```diff
@@ -81,15 +83,17 @@
 -rabbit_feature_flag(
    {direct_exchange_routing_v2,
     #{desc       => "v2 direct exchange routing implementation",
-      stability  => stable,
+      %%TODO remove compatibility code
+      stability  => required,
       depends_on => [feature_flags_v2, implicit_default_bindings],
       callbacks  => #{enable => {?MODULE, direct_exchange_routing_v2_enable}}
```
dumbbell, line 89: The callback functions should be removed at the same time.

deps/rabbit/src/rabbit_core_ff.erl
```diff
@@ -100,7 +104,8 @@
 -rabbit_feature_flag(
    {tracking_records_in_ets,
     #{desc          => "Store tracking records in ETS instead of Mnesia",
-      stability     => stable,
+      %%TODO remove compatibility code
+      stability     => required,
       depends_on    => [feature_flags_v2],
       callbacks     => #{enable =>
                              {?MODULE, tracking_records_in_ets_enable},
```
dumbbell, line 111: The callback functions should be removed at the same time.

michaelklishin (approved):

dumbbell (approved):

PR #7228, opened 2023-02-09
Remove mixed version check in MQTT tests

ansd:
In the MQTT test assertions, instead of checking whether the test runs in mixed version mode where all non-required feature flags are disabled by default, check whether the given feature flag is enabled.

Prior to this commit, once feature flag `rabbit_mqtt_qos0_queue` becomes required, the test cases would have failed.

### DIFF
--- a/deps/rabbitmq_mqtt/test/reader_SUITE.erl
+++ b/deps/rabbitmq_mqtt/test/reader_SUITE.erl
@@ -299,14 +299,13 @@ rabbit_mqtt_qos0_queue_overflow(Config) ->
     %% We expect that
     %% 1. all sent messages were either received or dropped
     ?assertEqual(NumMsgs, NumReceived + NumDropped),
-    case rabbit_ct_helpers:is_mixed_versions(Config) of
-        false ->
+    case rabbit_ct_broker_helpers:is_feature_flag_enabled(Config, rabbit_mqtt_qos0_queue) of
+        true ->
             %% 2. at least one message was dropped (otherwise our whole test case did not
             %%    test what it was supposed to test: that messages are dropped due to the
             %%    server being overflowed with messages while the client receives too slowly)
             ?assert(NumDropped >= 1);
-        true ->
-            %% Feature flag rabbit_mqtt_qos0_queue is disabled.
+        false ->
             ?assertEqual(0, NumDropped)
     end,
     %% 3. we received at least 1000 messages because everything below the default
--- a/deps/rabbitmq_mqtt/test/reader_SUITE.erl
+++ b/deps/rabbitmq_mqtt/test/reader_SUITE.erl
@@ -261,6 +261,8 @@ event_authentication_failure(Config) ->
 %% Test that queue type rabbit_mqtt_qos0_queue drops QoS 0 messages when its
 %% max length is reached.
 rabbit_mqtt_qos0_queue_overflow(Config) ->
+    ok = rabbit_ct_broker_helpers:enable_feature_flag(Config, rabbit_mqtt_qos0_queue),
+
     Topic = atom_to_binary(?FUNCTION_NAME),
     Msg = binary:copy(<<"x">>, 1000),
     NumMsgs = 10_000,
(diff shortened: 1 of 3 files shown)


deps/rabbitmq_mqtt/test/reader_SUITE.erl
```diff
@@ -299,14 +299,13 @@ rabbit_mqtt_qos0_queue_overflow(Config) ->
     %% We expect that
     %% 1. all sent messages were either received or dropped
     ?assertEqual(NumMsgs, NumReceived + NumDropped),
-    case rabbit_ct_helpers:is_mixed_versions(Config) of
-        false ->
+    case rabbit_ct_broker_helpers:is_feature_flag_enabled(Config, rabbit_mqtt_qos0_queue) of
```
dumbbell, line 302: You should try to enable the feature flag, then handle the error if any and fallback to an alternative code path.

It's possible that both the old and new versions support the feature flag. It's just that in mixed-version mode, all non-required feature flags are disabled by default when nodes start.

dumbbell (changes requested):

deps/rabbitmq_mqtt/test/reader_SUITE.erl
ansd: Thanks. For this specific test case, I pushed a commit where this feature flag gets always enabled because this test case doesn't make sense to be executed without `mqtt_qos0` queue type being supported.

deps/rabbitmq_mqtt/test/reader_SUITE.erl
```diff
@@ -261,6 +261,8 @@ event_authentication_failure(Config) ->
 %% Test that queue type rabbit_mqtt_qos0_queue drops QoS 0 messages when its
 %% max length is reached.
 rabbit_mqtt_qos0_queue_overflow(Config) ->
+    ok = rabbit_ct_broker_helpers:enable_feature_flag(Config, rabbit_mqtt_qos0_queue),
```
dumbbell, line 264: This will crash if the feature flag can't be enabled.

You should enable the feature flag(s) you need from an `init_per_*()` function and decide what do next.

If the feature flag can't be enabled, then either skip the testcase if it doesn't make sense without it, or do something different if the main test scenario can't work without it.

dumbbell (changes requested):

deps/rabbitmq_mqtt/test/reader_SUITE.erl
dumbbell: Ah ok, I missed the commit message, sorry.

deps/rabbitmq_mqtt/test/shared_SUITE.erl
```diff
@@ -446,12 +447,11 @@ events(Config) ->
     QueueNameBin = <<"mqtt-subscription-", ClientId/binary, "qos0">>,
     QueueName = {resource, <<"/">>, queue, QueueNameBin},
     [E2, E3 | E4] = get_events(Server),
-    QueueType = case rabbit_ct_helpers:is_mixed_versions(Config) of
-                    false ->
+    QueueType = case is_feature_flag_enabled(Config, rabbit_mqtt_qos0_queue) of
```
dumbbell, line 450: But if you do this, you never test that the feature flag can be enabled in a cluster, even though all of the nodes support it.

ansd: Why not?

When this test runs in non-mixed version mode, all nodes in the cluster support this feature flag and this feature flag gets enabled by default without me explicitly enabling it within the test suite.
This feature flag does not contain any migration code. But even if it contained migration code, and that migration code failed, this whole cluster would not become healthy and the test case will fail (which is what we want).

In mixed version mode, there can be 3 situations:
1. Not all nodes support this feature flag (as it's the case with the currently configured 3.11.7 "lower version" nodes): In this case, I do want this test to assert that a `classic` queue gets created.
2. All nodes support this feature flag (this will be the case in the future when we bump the "lower version" nodes to 3.12). Even in that scenario I do want the feature flag to be disabled to test that a classic queue gets created (feature flags are disabled by default in our mixed version cluster tests).
3. The feature flag becomes `required` (even further in the future), in which case the feature flag is enabled, and the test will pass as well. Once we `require` this feature flag, this case statement will be removed as part of "cleaning up" the feature flag's compatibility code.

dumbbell: There is a 4th situation: all nodes support this feature flag but don't run the same code, in which case we would also want to make sure they can work together with the feature flag enabled.

Perhaps this situation is not possible today; do you confirm that?

ansd: >There is a 4th situation: all nodes support this feature flag but don't run the same code

This corresponds to situation 2. in my comment above, right?
In that case, what you're suggesting is that instead of only testing with the feature flag being disabled, we could (or should) **additionally** test with the feature flag being enabled?

We could do that. However this sounds to me that we are missing this scenario for **all test suites**.
So, instead of testing what we have today:
- all nodes on the new version with all feature flags enabled, and
-  mixed version nodes with all feature flags disabled by default

we need yet another GitHub action workflow with
- mixed version nodes with all supported feature flags enabled

We could introduce such a new workflow if we think it's worth it?

From my point of view, what you are suggesting is good, but it applies to all feature flags in all test suites, and not only to this single feature flag in this single test case we're discussing, right?

dumbbell: > This corresponds to situation 2. in my comment above, right?

Yes, this is close.

> From my point of view, what you are suggesting is good, but it applies to all feature flags in all test suites, and not only to this single feature flag in this single test case we're discussing, right?

I think it would more be efficient to handle that in each testsuite. There is no need to run testcases which are not affected by feature flags again. Also some testcases may only be relevant in a specific situation, not all. We execute all testsuites twice for regular and mixed-version cluster because the conditions are very different even when considering the compilation.

dumbbell (approved):

dumbbell:
I reviewed and merged your branch, we can continue to discuss how to test separately from this patch. Thanks!

PR #7263, opened 2023-02-12
WIP Support MQTT 5.0

ansd:


deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -293,6 +295,20 @@ process_request(?PUBACK,
...
+%% to a Client and it receives a PUBLISH packet with a QoS greater than this
+%% then it uses DISCONNECT with Reason Code 0x9B (QoS not supported).
+process_request(?PUBLISH,
+                #mqtt_packet{fixed = #mqtt_packet_fixed{qos = ?QOS_2}},
+                State = #state{cfg = #cfg{
+                                        proto_ver = ?MQTT_PROTO_V5,
+                                        client_id = ClientID}}) ->
+    ?LOG_DEBUG("Terminating MQTT connection. QoS not supported, client ID: ~s ,"
+               "protocol version: ~p, qos: ~p",
+               [ClientID, ?MQTT_PROTO_V5, ?QOS_2]),
```
ansd, line 309: Some suggestions on the log line:
1. Given this is a protocol error, should we increase the log level? In other places we log protocol errors at `error` level.
2. For logs output to the client, I would capitalise, i.e. writing `QoS` instead of `qos`.
3. I would first write the comma, then the space. So, instead of `client ID: myid ,protocol version`, it should be `client ID: my_id, protocol version`.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -192,7 +192,8 @@ process_connect(
                        'Maximum-Packet-Size' => persistent_term:get(
                                                   ?PERSISTENT_TERM_MAX_PACKET_SIZE_AUTHENTICATED),
                        'Subscription-Identifier-Available' => 0,
-                       'Shared-Subscription-Available' => 0
+                       'Shared-Subscription-Available' => 0,
+                       'Wildcard-Subscription-Available' => 0
```
ansd, line 196: We should revert this commit because RabbitMQ MQTT does support wildcard subscriptions.

ansd: @ChunyiLyu I dropped your commit `Return wildcard sub unsupport in connack` when rebasing. I pushed a backup to branch https://github.com/ansd/rabbitmq-server/tree/mqtt5-backup-2023-03-15 such that your commit is not lost in case you will need it.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -163,6 +163,7 @@ process_connect(
             %% Therefore, we use the maximum allowed session expiry interval.
             MaxSessionExpiry
     end,
+    ReceiveMax = maps:get('Receive-Maximum', ConnectProps, rabbit_mqtt_util:env(prefetch)),
```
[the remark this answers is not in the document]
ChunyiLyu: Good point! I've modified as you suggested. The processor state config field `ReceiveMaximum` is renamed back to `prefetch`.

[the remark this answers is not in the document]
ansd: Awesome, thanks!

deps/rabbitmq_web_mqtt/src/rabbit_web_mqtt_handler.erl
```diff
@@ -156,7 +155,9 @@ websocket_info({'$gen_cast', duplicate_id}, State = #state{ proc_state = ProcSta
                                                             conn_name = ConnName }) ->
     ?LOG_WARNING("Web MQTT disconnecting a client with duplicate ID '~s' (~p)",
                  [rabbit_mqtt_processor:info(client_id, ProcState), ConnName]),
-    stop(State);
+    rabbit_mqtt_processor:send_disconnect(?RC_SESSION_TAKEN_OVER, ProcState),
+    defer_close(?CLOSE_PROTOCOL_ERROR),
```
ansd, line 159: Should we rather use here `?CLOSE_NORMAL`?
A session being taken over is not really a WebSocket protocol error. It rather fits the description of the Status Code 1000 in the sense that the purpose of the WebSocket / MQTT connection has been fulfilled because a new network connection (but same MQTT session) with the same Client ID exists:
> 1000 indicates a normal closure, meaning that the purpose for which the connection was established has been fulfilled

mkuratczyk:
I haven't found any regressions. There are as always minor differences here and there but in both directions, so I attribute this to entropy. Here with 1kb messages:
Screenshot 2023-06-21 at 09 25 51

Full results:
https://snapshots.raintank.io/dashboard/snapshot/4U5Xu5GwCgo1qKHSECQ2ScS5E7k6fqbW

ChunyiLyu (dismissed):

ChunyiLyu (approved):

PR #7269, opened 2023-02-13
Remove compatibility for feature flag stream_queue

ansd:
Remove compatibility code for feature flag `stream_queue` because this feature flag is required in 3.12.

See #7219


michaelklishin (approved):

dumbbell:
The `rabbit_queue_type:is_enabled()` part seems useful for future queues types, even though all queue types are supported today. @ansd: What do you think?

ansd:
I was previously thinking about the same and decided it's better to delete unused code and add it back in case we have a new queue type in the future.
But it seems that you also find it useful. So, I'm going to add it back :)

PR #7303, opened 2023-02-15
CQ: Remove compat code for classic_mirrored_queue_version

lhoguin:
Remove compat code that is no longer necessary in 3.12+.


ansd (commented):
Thanks. Looks good to me. I also deleted this feature flag in https://github.com/rabbitmq/rabbitmq-server/pull/7319.

PR #7326, opened 2023-02-16
Faster node startup with many classic queues

mkuratczyk:
On my machines, in a test with 100k queues, node startup goes down from 4-5 minutes to roughly 1 minute. The difference will be even larger with more queues as `lists:member` gets very expensive with a long list.

### DIFF
--- a/deps/rabbit/src/rabbit_classic_queue.erl
+++ b/deps/rabbit/src/rabbit_classic_queue.erl
@@ -154,15 +154,14 @@ find_missing_queues(Q1s, Q2s) when length(Q1s) == length(Q2s)->
 find_missing_queues(Q1s, Q2s) ->
     find_missing_queues(lists:sort(Q1s), lists:sort(Q2s), []).
 
-find_missing_queues([], Q2s, Acc) ->
-    %% Q2s should always be empty here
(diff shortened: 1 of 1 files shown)


kjnilsson:
Nice catch. The check when there is a failure could probably be further optimised by using sets and using `sets:subtract/2`. Especially with version 2 sets.

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -142,15 +142,29 @@ recover(VHost, Queues) ->
...
+
+find_missing_queues([], Q2s, Acc) ->
+    %% Q2s should always be empty here
+    Acc ++ Q2s;
+find_missing_queues([Q1|Rem1], [Q2|Rem2] = Q2s, Acc) ->
+    case amqqueue:get_name(Q1) == amqqueue:get_name(Q2) of
+        true ->
+            find_missing_queues(Rem1, Rem2, Acc);
+        false ->
+            find_missing_queues(Rem1, Q2s, Acc ++ Q1)
```
lhoguin, line 165: `[Q1|Acc]` and reverse at the end if necessary.

lhoguin (approved):

PR #7365, opened 2023-02-20
Read JWT token from Authorization header to login to management ui

MarcialRosales:
In identity-provider-initiated scenarios, some idps/proxies send the access token in the standard HTTP Authorization header.. This PR is about supporting this scenario in addition to maintaining the current behaviour which expects the token in the form field "access_token" against the endpoint `/login`

## Proposed Changes

Read access token from the standard HTTP header Authorization while loading a dynamically generated javascript file which initializes the client-side oidc-client library and set up the token received over the http header.

## Types of Changes

- [ ] Build system and/or CI

This PR is accompanied by another [PR](https://github.com/rabbitmq/rabbitmq-website/pull/1631) for the rabbitmq-website


deps/rabbitmq_management/src/rabbit_mgmt_login.erl
```diff
@@ -38,3 +39,6 @@ login(<<"POST">>, Req0, State) ->
 login(_, Req0, _) ->
   %% Method not allowed.
 	cowboy_req:reply(405, Req0).
+
+get_authorization_token(undefined, Access_Token_Form_Field) -> Access_Token_Form_Field;
```
michaelklishin, line 43: Erlang uses CamelCase for bindings (variables), so this should be `AccessTokenFormField`.

MarcialRosales: Thanks @michaelklishin !

PR #7367, opened 2023-02-20
DO NOT MERGE: Update Ra to 2.4.8

kjnilsson:
This release
* improves election availability in mixed versions
* optimises segment reference compaction which may becomes expensive in quorum queues with very long backlogs
* Varous log message improvements and level tweaks
* Better cleans up machine monitor records after quorum queue rebalancing


kjnilsson:
This Ra versions has some unintended consequences so closing.

PR #7412, opened 2023-02-24
3.11.x: resolve a Web MQTT Makefile conflict

michaelklishin:


michaelklishin:
Note that while this branch does not use `emqtt`, for the sake of reducing the delta at least in build files I am happy to have it listed.

PR #7441, opened 2023-02-27
Fix rabbitmq-diagnostics remote_shell for 26.0

lhoguin:
I have removed the format commands because I have no idea why they make compilation fail with my changes and I don't know how to fix (no clear error message). If the format commands are important please someone who knows fix the PR or provide instructions.

This PR fixes remote_shell for 26.0 and future versions and simplifies the code by using the new function from 26 (or will once the <26 compat gets removed).

cc @mkuratczyk


mkuratczyk:
@mergifyio backport v3.11.x

PR #7477, opened 2023-03-01
Use integer as will message correlation

ansd:
Instead of using atom `undefined`, use an integer as correlation term when sending the will message to destination queues.

Classic queue clients for example expect a non negative integer. Quorum queues expect any term.


ChunyiLyu (approved):
🚢

PR #7516, opened 2023-03-03
CQ: Confirm related optimisations

lhoguin:
1: The message(s) was/were already optimised out during write when the client acked faster than we could process the write message.

2: The optimisation avoids calling an expensive sets:subtract/2 when there is a single new confirm and instead does a sets:del_element/2 of the first set.


lhoguin:
First commit provides small improvements, second bigger ones. Targeting 3.12 only, no longer 3.11, because the second commit doesn't apply to 3.11 and the first commit is not that useful there. One more reason to upgrade.

mkuratczyk:
Validated to significantly improve some scenarios. Here:
* yellow - current 3.11 branch
* blue - current 3.12 branch (without this change)
* green - 3.12 with this change
Screenshot 2023-03-06 at 14 19 07

PR #7529, opened 2023-03-07
Expose mqtt.mailbox_soft_limit via Cuttlefish

ansd:


michaelklishin (approved):

michaelklishin:
We could add a validator for positive integers, or even enforce a sane minimum that way.

ansd:
Thank you @michaelklishin. I added a validator for non negative integer.

PR #7548, opened 2023-03-08
Fix rabbit_vm:memory returning stream_queue_coordinator_procs

gomoripeti:
## Proposed Changes

The memory of `rabbit_stream_coordinator` process was reported as "other", because there was a mismatch in ancestors. It is not under `ra_server_sup_sup` but: `[_,ra_coordination_server_sup_sup,_,ra_systems_sup,ra_sup,_]`

Observed with versions ra 2.4.5, 2.4.6 RabbitMQ 3.11.5

I'm not sure if this is the correct fix.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_vm.erl
+++ b/deps/rabbit/src/rabbit_vm.erl
@@ -177,7 +177,8 @@ bytes(Words) ->  try
                  end.
 
 interesting_sups() ->
-    [queue_sups(), quorum_sups(), dlx_sups(), stream_server_sups(), stream_reader_sups(),
+    [queue_sups(), quorum_sups(), dlx_sups(),
+     stream_server_sups(), stream_reader_sups(), stream_coordinator(),
      conn_sups() | interesting_sups0()].
 
 queue_sups() ->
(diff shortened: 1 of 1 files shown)


kjnilsson:
This looks like it could work but as we now have a separate supervisor for quorum queues and coordination type ra machine we don't need to use `ra_type/1` to distinguish between them. I think it would be better just to use the supervisors + maybe the registered name (for difference coordination types) to distinguish between quorum queue and other memory use.

michaelklishin (changes requested):
Should this be under "stream server sups" instead?

gomoripeti:
Pushed a second try. I kept the stream coordinator separate from stream servers, with the consideration that the former is a raft process, the laters are osiris processes. And we've seen cases when the stream coordinator alone was consuming a lot of memory.

gomoripeti:
Looks like deps.rabbit.classic_queue_prop_SUITE.groups.classic_queue_v1 failed. I dont see how it is related to this change.

michaelklishin (approved):

PR #7553, opened 2023-03-09
Ra 2.5.0

kjnilsson:
This Ra release includes improvements to Ra server GC behaviour when receiving a lot of low priority commands with large binary payloads (e.g. quorum queue messages).

Practically this allows quorum queues to accept large amounts of messages in a more predicatble and performant manner.

This change also removes ra_file_handle cache that was used as a bridge between ra file operations and RabbitMQ io metrics. Lots of components in RabbitMQ such as streams and CQv2s do not record io metrics in the previous manner due to overhead incurred for every file io operation. These metrics are better inspected at the OS level anyway.


mkuratczyk:
Link to full performance test results:
https://grafana.rabbitmq.com/goto/2zvsFH-4k?orgId=1

The original goal of this change was to address degradation under sustained publisher load without consumers - when the queue grew longer, its throughput quickly degraded. This has been resolved:

Screenshot 2023-03-10 at 12 13 56

Whit this PR, single queue (publisher+consumer) throughput is higher for all tested message sizes:
Screenshot 2023-03-10 at 12 15 46

In tests where an initial backlog is created and then consumers start (in the first test here - 10 consumers, in the second - SAC), this PR allows to recover publishing throughput quickly and provides much higher consumption rate in the SAC scenario:
Screenshot 2023-03-10 at 12 17 58

PR #7612, opened 2023-03-14
Use pkg_files rules to avoid extra tars

pjk25:
in package-generic-unix and source_archive macros


pjk25:
Will roll this change into #7609

PR #7619, opened 2023-03-15
Avoid fetching test deps when building rabbitmqctl with bazel

pjk25:
Also for bazel offline build readiness

Relates to #7609


pjk25:
Waiting for #7609 before merging and backporting

PR #7640, opened 2023-03-16
Give each of the summary jobs in actions different names

pjk25:
Otherwise they do not appear to be selectable in the github branch protection rules UI


pjk25:
follow up to #7631

PR #7601, opened 2023-03-14
Allow applying policies to specific queue types

mkuratczyk:
Rather than relying on queue name conventions, allow applying policies based on the queue type. For example, this allows multiple policies that apply to all queue names (".*") that specify different parameters for different queue types.

New `rabbitmqctl set_policy` options:
```
rabbitmqctl set_policy ttl10 ".*" '{"message-ttl": 10}' --apply-to classic_queues
rabbitmqctl set_policy ttl20 ".*" '{"message-ttl": 20}' --apply-to quorum_queues
rabbitmqctl set_policy age ".*" '{"max-age": "1h"}' --apply-to streams
```

New UI:
Screenshot 2023-03-14 at 08 21 24

Docs PR:
https://github.com/rabbitmq/rabbitmq-website/pull/1624

PR #7654, opened 2023-03-16
Fix rabbit_policy:match_all/2

pjk25:
when called with names instead of records, by falling back to the old behaviour prior to #7601


mkuratczyk (approved):
LGTM

PR #7676, opened 2023-03-20
Faster node start with many classic queues v2

mkuratczyk:
Faster CQv2 startup.

On my test box, with [100k CQs](https://github.com/rabbitmq/sample-configs/blob/main/queues/100k-classic-queues.json) (with `classic_queue.default_version = 2`), on main:
import: ~70 seconds
start_app: ~55-65 seconds

With this PR:
import: ~48 seconds
start_app: ~29 seconds

### DIFF
--- a/deps/rabbit/src/rabbit_classic_queue_index_v2.erl
+++ b/deps/rabbit/src/rabbit_classic_queue_index_v2.erl
@@ -205,8 +205,7 @@ init1(Name, Dir, OnSyncFun, OnSyncMsgFun) ->
 
 ensure_queue_name_stub_file(#resource{virtual_host = VHost, name = QName}, Dir) ->
     QueueNameFile = filename:join(Dir, ?QUEUE_NAME_STUB_FILE),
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_classic_queue_index_v2.erl
```diff
@@ -1292,3 +1291,25 @@ highest_continuous_seq_id([SeqId1, SeqId2|Tail], EndSeqId)
...
+write_file_and_ensure_dir(Name, IOData) ->
+    case write_file(Name, IOData) of
+        ok -> ok;
+        {error, enoent} ->
+            filelib:ensure_dir(Name),
+            write_file(Name, IOData)
+    end.
+
+write_file(Name, IOData) ->
+    case file:open(Name, [binary, write, raw]) of
```
lhoguin, line 1304: Not what I meant, you can do `ok = file:write_file(QueueNameFile, <<"VHOST: ", VHost/binary, "\n",...>>, [raw])` directly.

lhoguin (approved):

Issue #7743, opened 2023-03-27
SuperStream doesn't elect the single active consumer


Gsantomaggio:
### Describe the bug

The super stream doesn't elect the single active consumer when the consumers are restarted.
The consumers stop consuming and the status is always in `waiting` status. see the image:
![Screenshot 2023-03-27 at 10 42 15](https://user-images.githubusercontent.com/386987/227899778-baf33045-7ea9-442b-8842-9dd535f2d195.png)

### Reproduction steps

1. Create super stream with ten partitions `rabbitmq-streams add_super_stream invoices --partitions 10`
2. pump it with a few thousand messages (70k is enough)
3. stop the producer
4. start ten instances of this java client:
```java
    public static void main(String[] args) throws IOException {

        System.out.println("Connecting...");
        Address entryPoint = new Address("127.0.0.1", 5552);


        Environment environment = Environment.builder()
//                .host(entryPoint.host())
//                .port(entryPoint.port())
//                .username("test")
//                .password("test")
                .addressResolver(address -> entryPoint)
                .maxConsumersByConnection(1).
                build();
        String AppName = "reference";
        String stream = "invoices";

        AtomicInteger consumed = new AtomicInteger();
        Date start = new Date();
        for (int i = 0; i < 500; i++) {


            Map<String, Integer> consumedMap = new HashMap<>();
            Consumer consumer = environment.consumerBuilder()
                    .superStream("invoices")
                    .name("reference")
                    .offset(OffsetSpecification.first())
                    .singleActiveConsumer()
                    .messageHandler((context, message) -> {

                        if (consumedMap.containsKey(stream)) {
                            consumedMap.put(stream, consumedMap.get(stream) + 1);
                        } else {
                            consumedMap.put(stream, 1);
                        }

                        if (consumedMap.get(stream) % 10 == 0) {
                            Date end = new Date();
                            System.out.println("Stream: " + context.stream() + " - Consumed " + consumedMap.get(stream) + " - Time " + (end.getTime() - start.getTime()));
                        }


                        try {
                            Thread.sleep(ThreadLocalRandom.current().nextInt(200, 1000));
                        } catch (InterruptedException e) {
                            throw new RuntimeException(e);
                        }

                    }).build();

            try {
                Thread.sleep(60000);
            } catch (InterruptedException e) {
                throw new RuntimeException(e);
            }
            System.out.println("Restarting");

            consumer.close();
        }

    }
```
5. Wait a couple of restarts
[...]

PR #7765, opened 2023-03-30
Unblock group of consumers on super stream partition

acogoluegnes:
A group of consumers on a super stream can end up blocked without an active consumer. This can happen with consumer churn: one consumer gets removed, which makes the active consumer passive, but the former active consumer never gets to know because it has been removed itself.

This commit changes the structure of the messages the SAC coordinator sends to consumer connections, to embed enough information to look up the group and to instruct it to choose a new active consumer when the race condition mentioned above comes up.

Because of the changes in the structure of messages, a feature flag is required to make sure the SAC coordinator starts sending the new messages only when all the nodes have been upgraded.

References #7743

### DIFF
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -800,6 +800,9 @@ rabbitmq_suite(
     deps = [
         "//deps/rabbit_common:erlang_app",
     ],
+    runtime_deps = [
+        "@meck//:erlang_app",
+    ],
 )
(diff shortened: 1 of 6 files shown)


Gsantomaggio:
I can't reproduce the [issue 7743](https://github.com/rabbitmq/rabbitmq-server/issues/7743) anymore.
I tested it with Java Stream Client and Dotnet stream client.

PR #7806, opened 2023-04-03
Update Cowlib to 2.12.1

lhoguin:
This fixes a compilation error with OTP-26.


lhoguin:
Test failure was due to manual parsing of the setcookie header because there's been some (functionally non-breaking) adjustments in recent Cowlib versions. Manual parsing of headers is bad but we can worry about improving that another time.

PR #7821, opened 2023-04-04
Update ct_helper

lhoguin:
Since ct_helper removes erl_make_certs some tests needed to be updated to use public_key:pkix_test_data/1 instead.

Special care must be taken during review around the long chain test since we expect the trusted and untrusted certificates to be using the same chain (they are issued by the same intermediary certificate).

### DIFF
--- a/deps/rabbitmq_trust_store/test/system_SUITE.erl
+++ b/deps/rabbitmq_trust_store/test/system_SUITE.erl
@@ -201,9 +201,9 @@ validation_success_for_AMQP_client1(Config) ->
     %% Note that when this test is executed together with the HTTP provider group
     %% it runs into unexpected interference and fails, even if TLS app PEM cache is force
     %% cleared. That's why originally each group was made to use a separate node.
-    AuthorityInfo = {Root, _AuthorityKey} = erl_make_certs:make_cert([]),
-    {Certificate, Key} = chain(AuthorityInfo),
-    {Certificate2, Key2} = chain(AuthorityInfo),
+    RootCert = #{cert := Root} = public_key:pkix_test_root_cert("RootCA", []),
+    {Certificate, Key} = chain(RootCert),
+    {Certificate2, Key2} = chain(RootCert),
     Port = port(Config),
     Host = rabbit_ct_helpers:get_config(Config, rmq_hostname),
     %% When: Rabbit accepts just this one authority's certificate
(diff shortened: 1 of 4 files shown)


deps/rabbitmq_trust_store/test/system_SUITE.erl
```diff
@@ -326,18 +326,30 @@ validate_longer_chain1(Config) ->
...
+        root => [],
+        intermediaries => [[{key, KeyInter}]],
+        peer => []
+    }),
+    CertTrusted = proplists:get_value(cert, TestDataTrusted),
+    KeyTrusted = proplists:get_value(key, TestDataTrusted),
+    [CertInter, RootCA] = proplists:get_value(cacerts, TestDataTrusted),
+
+    TestDataUntrusted = public_key:pkix_test_data(#{
+        root => [#{cert => CertInter, key => KeyInter}],
```
the-mikedavis, line 343: It looks like this is expecting a proplist or a map: https://github.com/erlang/otp/blob/e599897f91ca5f05b4d868c6776efad7e6c1ad8d/lib/public_key/src/pubkey_cert.erl#L505-L523. (`pubkey_cert:root_cert/2` tries to get the key out of this with `proplists:get/3`.)

It also looks like this one expects the dec key:

```suggestion
        root => [{cert, CertInter}, {key, KeyInterDec}],
```

Or alternatively I think this should work:

```suggestion
        root => #{cert => CertInter, key => KeyInterDec},
```

deps/rabbitmq_trust_store/test/system_SUITE.erl
```diff
@@ -326,18 +326,30 @@ validate_longer_chain1(Config) ->
...
-    KeyUntrusted = {KindUntrusted, KeyDataUntrusted},
-    {CertTrusted, {Kind, KeyData, _}} = erl_make_certs:make_cert([{issuer, Inter}]),
-    KeyTrusted = {Kind, KeyData},
+
+    KeyInterDec = public_key:generate_key({rsa, 2048, 17}),
+    KeyInter = {'RSAPrivateKey', public_key:der_encode('RSAPrivateKey', KeyInterDec)},
+
+    TestDataTrusted = public_key:pkix_test_data(#{
+        root => [],
+        intermediaries => [[{key, KeyInter}]],
```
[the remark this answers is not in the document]
lhoguin: Hah, this explains a few things I didn't understand.

deps/rabbitmq_trust_store/test/system_SUITE.erl
lhoguin: The map is correct, just not fully documented. It's how you can pass the root certificate from pkix_test_root_cert. Interesting I didn't get a crash because of the list though.

the-mikedavis (approved):
Looks great, thanks for updating this!

PR #7825, opened 2023-04-04
If a queue process is hibernated, consider it up

mkuratczyk:
`rabbit_core_metrics_gc` checks the status of all queues (and other things) every two minutes. With many queues, this can get pretty expensive. Systems with lots of queues will often have many of them idle and hibernated. Before this change `rabbit_core_metrics_gc` would wake up all of them, causing a memory spike. All of that just to check if their state isn't `down`.

With this change, a hibernated queue will be considered up. The effect is visible with many queues (100k in this example) -roughly 0.5GB lower memory usage during metrics GC. Here I have 3 metrics GC runs with this change and then 2 with this change reverted:

Screenshot 2023-04-04 at 10 24 31

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -1077,13 +1077,16 @@ list_local_names_down() ->
                               is_down(Q)].
 
 is_down(Q) ->
-    try
-        info(Q, [state]) == [{state, down}]
-    catch
-        _:_ ->
-            true
+    case rabbit_process:is_process_hibernated(amqqueue:get_pid(Q)) of
+        true -> false;
+        false ->
+            try
+                    info(Q, [state]) == [{state, down}]
+            catch
+                _:_ ->
+                    true
+            end
     end.
-
 
 -spec sample_local_queues() -> [amqqueue:amqqueue()].
 sample_local_queues() -> sample_n_by_name(list_local_names(), 300).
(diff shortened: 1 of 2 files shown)


lhoguin:
Like mentioned in Slack this is valid only for normal classic queues. Have you double checked whether the `state` has additional checks for CMQs? Also would plugins that implement backing queues add additional checks?

mkuratczyk:
I tested this change with mirroring enabled and didn't notice any issues. I've browsed the code for anything special related to the `state` and mirroring but couldn't find anything. Ultimately, [we are going through local queues and only](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_core_metrics_gc.erl#L69-L70) and only the master of a mirrored queue is considered a local queue (we check the `amqqueue:get_pid(Q)` which returns the PID of the master), so in a way this function ignores mirrors anyway.

I have not considered any external implementations. I've never seen one or heard of anyone using one.

lhoguin:
The deduplication plugin is an example of plugin that implements a backing queue (https://github.com/noxdafox/rabbitmq-message-deduplication/blob/main/lib/rabbitmq_message_deduplication/rabbit_message_deduplication_queue.ex). I do not know what other plugins exist or how much they are used.

I figure the change is safe but it's a potential incompatibility to consider.

mkuratczyk:
@noxdafox can we ask for your comment? do you think this change could affect message deduplication?

noxdafox:
The deduplication queue is implementing the `backing_queue_behaviour`. The mechanism in this PR is "above" it and does not seem to be visible from the plugin perspective.

The only effect I see is the `backing_queue_behaviour:info` callback will not be called if the queue is hibernated which I guess is what we want also in case of deduplication queues.

PR #7865, opened 2023-04-07
Fix stream import on startup

acogoluegnes:
This commit fixes 2 problems when importing definitions that contain streams on startup:
* use `rabbit_nodes:list_reachable/0` instead of `rabbit_nodes:list_running/0`. The latter returns an empty list at this point, which makes the startup of the stream coordinator Ra cluster fail. The former returns the cluster nodes, regardless of the status of the `rabbit` application (up for `list_running`), which is enough in our case.
* return the Mnesia record if it exists when trying to create a queue and do not expect the record to be exactly equal to the passed-in queue parameter.


acogoluegnes:
Closing, will split the changes into 2 commits.

PR #7866, opened 2023-04-07
Fix stream import on startup

acogoluegnes:
This commit fixes 2 problems when importing definitions that contain streams on startup:
* use `rabbit_nodes:list_reachable/0` instead of `rabbit_nodes:list_running/0`. The latter returns an empty list at this point, which makes the startup of the stream coordinator Ra cluster fail. The former returns the cluster nodes, regardless of the status of the `rabbit` application (up for `list_running`), which is enough in our case.
* return the Mnesia record if it exists when trying to create a queue and do not expect the record to be exactly equal to the passed-in queue parameter.

### DIFF
--- a/deps/rabbit/src/rabbit_stream_coordinator.erl
+++ b/deps/rabbit/src/rabbit_stream_coordinator.erl
@@ -427,6 +427,7 @@ ensure_coordinator_started() ->
 start_coordinator_cluster() ->
     Nodes = rabbit_nodes:list_reachable(),
     rabbit_log:debug("Starting stream coordinator on nodes: ~w", [Nodes]),
+    true = Nodes =/= [],
     case ra:start_cluster(?RA_SYSTEM, [make_ra_conf(Node, Nodes) || Node <-  Nodes]) of
         {ok, Started, _} ->
             rabbit_log:debug("Started stream coordinator on ~w", [Started]),
--- a/deps/rabbit/src/rabbit_stream_coordinator.erl
+++ b/deps/rabbit/src/rabbit_stream_coordinator.erl
@@ -425,7 +425,7 @@ ensure_coordinator_started() ->
     end.
 
 start_coordinator_cluster() ->
-    Nodes = rabbit_nodes:list_running(),
+    Nodes = rabbit_nodes:list_reachable(),
     rabbit_log:debug("Starting stream coordinator on nodes: ~w", [Nodes]),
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -425,7 +425,7 @@ ensure_coordinator_started() ->
     end.
 
 start_coordinator_cluster() ->
-    Nodes = rabbit_nodes:list_running(),
+    Nodes = rabbit_nodes:list_reachable(),
     rabbit_log:debug("Starting stream coordinator on nodes: ~w", [Nodes]),
```
dumbbell, line 429: What about adding an assert like:
```erlang
Nodes = rabbit_nodes:list_reachable(),
rabbit_log:debug("Starting stream coordinator on nodes: ~w", [Nodes]),
?assertNotEqual([], Nodes),
...
```
... to catch an unwanted empty list early?

dumbbell: You will have to include `assert.hrl` in this case:
```erlang
-include_lib("stdlib/include/assert.hrl").
```

lhoguin: Not a fan of such asserts, especially in code that isn't just tests. The rabbit_log:debug will already tell us there are no nodes. So just doing `true = Nodes =/= []` should be enough.

dumbbell: I prefer the assertion macros because you get more details about the context (not only a vague badmatch exception) and the intent is clearer. But that's personal preferences, not a rule :-)

lhoguin: Crash reports have a lot more info these days and the info you lack is in the log just above on line 429. Even if you didn't have that log, looking at the crash line would be enough in this case.

There aren't many cases where the macros provide more info now, and when they do it's easy to have the same without the macro (for example `Expected = Result` could be written `{Expected, Expected} = {Result, Expected}`). But those cases are rare outside of tests.

dcorbacho (approved):

michaelklishin:
Thank you, @mkuratczyk, for carrying this over the finish line.

michaelklishin:
This is `3.12.0`-specific, so won't go into release notes as the issue should not be present in 3.11.

PR #7875, opened 2023-04-11
Fetch all cli deps with bazel

pjk25:
This avoids issues with hex and OTP 26, and is needed for offline bazel builds anyway


pjk25:
This can be "squash" merged using the "Squash and merge" option when ready.

PR #7898, opened 2023-04-14
Add OTP 26.0-rc3 to the test matrix

pjk25:
Allows `--config=rbe-26` in bazel commands


michaelklishin:
I assume a PR that adds `26.0-rc.3` is no longer relevant after the release of `26.0` GA.

PR #7900, opened 2023-04-14
(Initial, incomplete) OTP26 compatibility

mkuratczyk:
NOTE: do not merge yet (PR created to run tests on OTP25)

WIP: changes needed for OTP26 compatibility. This doesn't address all the issues at this point

### DIFF
--- a/deps/rabbit_common/src/rabbit_net.erl
+++ b/deps/rabbit_common/src/rabbit_net.erl
@@ -168,7 +168,11 @@ port_command(Sock, Data) when ?IS_SSL(Sock) ->
         {error, Reason} -> erlang:error(Reason)
     end;
 port_command(Sock, Data) when is_port(Sock) ->
-    erlang:port_command(Sock, Data).
+    case gen_tcp:send(Sock, Data) of
+        ok -> self() ! {inet_reply, Sock, ok},
+              true;
+        {error, Reason} -> erlang:error(Reason)
+    end.
 
 getopts(Sock, Options) when ?IS_SSL(Sock) ->
     ssl:getopts(Sock, Options);
--- a/deps/rabbit_common/src/rabbit_net.erl
+++ b/deps/rabbit_common/src/rabbit_net.erl
@@ -168,11 +168,22 @@ port_command(Sock, Data) when ?IS_SSL(Sock) ->
         {error, Reason} -> erlang:error(Reason)
     end;
 port_command(Sock, Data) when is_port(Sock) ->
+    port_command_tcp(Sock, Data).
+
+%% gen_tcp:send/2 does a selective receive of {inet_reply, Sock, Status}
+-if(?OTP_RELEASE < 26).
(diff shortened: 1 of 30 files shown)


deps/rabbit_common/src/rabbit_net.erl
```diff
@@ -168,7 +168,11 @@ port_command(Sock, Data) when ?IS_SSL(Sock) ->
         {error, Reason} -> erlang:error(Reason)
     end;
 port_command(Sock, Data) when is_port(Sock) ->
-    erlang:port_command(Sock, Data).
+    case gen_tcp:send(Sock, Data) of
+        ok -> self() ! {inet_reply, Sock, ok},
+              true;
+        {error, Reason} -> erlang:error(Reason)
+    end.
```
ansd, line 175: 1. If we support OTP 25 as minimum in 3.12, shouldn't we keep the old behaviour for OTP 25 and apply the new behaviour for OTP 26 (e.g. by checking macro `?OTP_RELEASE`). Otherwise, we may end up with performance regressions running this code in OTP 25 which does not make use of selective receive optimisations.
2. If we decide to keep this change as-is, why do we send `{inet_reply, Sock, ok}` to our own process? Can we remove the handling of all `inet_reply` messages in the protocol readers and writers?
3. If we decide to keep this change as-is, what's the difference between `sync` and `async` in https://github.com/rabbitmq/rabbitmq-server/blob/0f8a5899de55b928f67d31cdd14ea6c025e3678d/deps/rabbitmq_stomp/src/rabbit_stomp_reader.erl#L374-L384 Can we simply this function?
4. Can we delete `rabbit_net:port_command/2` and just use `rabbit_net:send/2` instead?

ansd (changes requested):

deps/rabbit_common/src/rabbit_net.erl
ansd: I pushed a commit which fixes the regressions mentioned in 1.
2.-4. will be fixed once we require OTP 26 as minimum.

michaelklishin:
@mergifyio backport v3.11.x

michaelklishin:
Backporting this to `v3.11.x` is complicated, the code paths changed involve unrelated changes e.g. from https://github.com/rabbitmq/rabbitmq-server/pull/6440. It's not necessarily worth it for a series that will go out of community support in Dec 2023.

michaelklishin:
I have backported all commits from this PR except for [c8b4742](https://github.com/rabbitmq/rabbitmq-server/pull/7900/commits/c8b4742333e233a230c1f5301218ac5ee132e7ab). The delta in stream coordinator code is non-trivial.

However, with these improvements nodes can accept connections and generally work fine (all Bunny tests pass, for example).

So this is a sufficient improvement to ship, right now RabbitMQ 3.11.16 on Erlang 26 cannot accept client connections, which makes us look really bad to the Windows users who download the latest Erlang installer.

PR #7905, opened 2023-04-14
DRAFT: Use new RA callback

SimonUnge:
## Proposed Changes

Draft for @kjnilsson 

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments


SimonUnge:
> I think the general approach is pretty much what we designed. I am concerned that there are functions called in `eval_members` that are potentially blocking such as `rabbit_nodes:list_running()` and we'd have to see if it feasible to run all "grow" logic in a non blocking way to avoid locking the Ra leader process up.
> 
> That said this is very much on the right track and we just need to investigate a bit further if the logic can be made safe to run inside the ra process.

Awesome. Any hints on how to investigate?

I was thinking of somehow call the 'eval_members' in its own process, mostly because it could be slow. Would not help with blocking though.

SimonUnge:
@kjnilsson Perhaps the eval_member callback could provide an optional 'reply fun', and the implementer could use that fun to report the result to the instead of directly responding, i.e spawn a process that does the eval, and add the 'effects' to the fun, and the funs job would be to send a msg back to the ra_server_proc leader process?

Issue #7856, opened 2023-04-05
Limit max number of dynamic shovels


illotum:
### Is your feature request related to a problem? Please describe.

Shovels are not accounted for in the broker backpressure mechanism. While shovel-spawned connections are part of the flow mode rate limiter, one can keep creating shovels themselves until the broker resources are exhausted.

### Describe the solution you'd like

I'd like to set a static node-local limit to the amount of shovels available for creation. Considering the special status of static shovels, the limit will only apply to the dynamic ones.

Since shovels are set by directly writing `rabbit_runtime_parameters`, it may be easy to extend said limit to federations and other arbitrary types. The setting is meant for service operators and advanced users.

Reaching the limit would fail creation and provide feedback similar to other limits.

### Describe alternatives you've considered

Another options is to apply a "concurrent active" limit instead. This would not error on shovel creation, but park the record until an available slot is freed up. This requires more elaborate user feedback to explain shovel status, e.g. simply saying "pending" may be not enough. This also moves solution close to the existing flow mode.

### Additional context

Followup to #7777 and similar.

PR #7917, opened 2023-04-18
Make it possible to configure per-node runtime parameter limits

illotum:
## Proposed Changes

Introduces an arbitrary per-host runtime parameter limit, e.g.

```
runtime_parameters.limits.shovel = 0
runtime_parameters.limits.federation = 100
```

The limit can be set to any value, negative integers treated as "unlimited". There are no sanity checks, and I expect it to be used only by power-users.

Main purpose is resource management per #7856, although this hypothetically can be used to disable certain broker features outright.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_runtime_parameters.erl
+++ b/deps/rabbit/src/rabbit_runtime_parameters.erl
@@ -157,7 +157,7 @@ set_any0(VHost, Component, Name, Term, User) ->
             E
     end.
 
--spec is_within_limit(binary()) -> rabbit_types:ok_or_error(binary()).
+-spec is_within_limit(binary()) -> ok | {errors, list()}.
 
 is_within_limit(Component) ->
     Params = application:get_env(rabbit, runtime_parameters, []),
--- a/deps/rabbit/src/rabbit_runtime_parameters.erl
+++ b/deps/rabbit/src/rabbit_runtime_parameters.erl
@@ -133,25 +133,42 @@ set_any0(VHost, Component, Name, Term, User) ->
             case flatten_errors(
                    Mod:validate(VHost, Component, Name, Term, get_user(User))) of
                 ok ->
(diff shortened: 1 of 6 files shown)


deps/rabbit/test/config_schema_SUITE_data/rabbit.snippets
```diff
@@ -898,6 +898,21 @@ credential_validator.regexp = ^abc\\d+",
      ]}],
    []},
 
+  %%
+  %% Runtime parameters
+  %%
+
+  {runtime_parameters_limits,
+  "
+   runtime_parameters.limits.b = 2
```
michaelklishin, line 907: Should we use more realistic parameter names (e.g. `shovel`) here so that it's easier to understand how to use this feature?

deps/rabbit/src/rabbit_runtime_parameters.erl
```diff
@@ -130,29 +130,46 @@ set_any0(VHost, Component, Name, Term, User) ->
...
 
+-spec is_within_limit(binary()) -> ok | {errors, list()}.
+
+is_within_limit(Component) ->
+    Params = application:get_env(rabbit, runtime_parameters, []),
+    Limits = proplists:get_value(limits, Params, []),
+    Limit = proplists:get_value(Component, Limits, -1),
+    case Limit < 0 orelse count_component(Component) < Limit of
+       true -> ok;
+       false -> {errors, [{"component ~ts is limited to ~tp per host", [Component, Limit]}]}
```
michaelklishin, line 168: Should this say "per node" to avoid confusion with virtual hosts?

michaelklishin (changes requested):
We should use more realistic parameter names in tests, and clarify what exactly "host" names in the message, e.g. is it a virtual host? a node?

michaelklishin (approved):

PR #7927, opened 2023-04-18
Make tcp send OTP 26 compatible

ansd:
Follow up of https://github.com/rabbitmq/rabbitmq-server/pull/7913 and https://github.com/rabbitmq/rabbitmq-server/pull/7921

This commit uses the approach explained in https://github.com/erlang/otp/issues/7130#issuecomment-1512808759

We cannot use the macro `?OTP_RELEASE` since macros are evaluated at compile time. RabbitMQ can be compiled with OTP 25 and executed with OTP 26. Therefore, we use `erlang:system_info(otp_release)` instead. As `erlang:system_info/1` might be costly, we store the send function in persistent_term.

For OTP 25, we use the "old tcp send workaround" (i.e. `erlang:port_command/2`) which avoids expensive selective receives. For OTP 26, we use `gen_tcp:send/2` which uses the optimised selective receive.

Once the minimum required version becomes OTP 26, we can just switch to `gen_tcp:send/2` and delete the `inet_reply` handling code in the various RabbitMQ reader and writer processes.

Note that `rabbit_net:port_command/2` is not only used by RabbitMQ server, but also by the AMQP 0.9.1 client.
Therefore, instead of putting the OTP version (or send function) into persistent_term within the rabbit app, we just do it the first time `rabbit_net:port_command/2` is invoked.
(`rabbit_common` is just a library without supervision hierarchy.)

### DIFF
--- a/deps/rabbit_common/src/rabbit_net.erl
+++ b/deps/rabbit_common/src/rabbit_net.erl
@@ -168,7 +168,32 @@ port_command(Sock, Data) when ?IS_SSL(Sock) ->
         {error, Reason} -> erlang:error(Reason)
     end;
 port_command(Sock, Data) when is_port(Sock) ->
-    erlang:port_command(Sock, Data).
+    Fun = case persistent_term:get(rabbit_net_tcp_send, undefined) of
+              undefined ->
+                  Rel = list_to_integer(erlang:system_info(otp_release)),
+                  %% gen_tcp:send/2 does a selective receive of
+                  %% {inet_reply, Sock, Status[, CallerTag]}
+                  F = if Rel >= 26 ->
+                             %% Selective receive is optimised:
+                             %% https://github.com/erlang/otp/issues/6455
+                             fun gen_tcp_send/2;
+                         Rel < 26 ->
+                             %% Avoid costly selective receive.
+                             fun erlang:port_command/2
+                      end,
+                  ok = persistent_term:put(rabbit_net_tcp_send, F),
+                  F;
+              F ->
+                  F
+          end,
+    Fun(Sock, Data).
+
+gen_tcp_send(Sock, Data) ->
+    case gen_tcp:send(Sock, Data) of
+        ok -> self() ! {inet_reply, Sock, ok},
+              true;
+        {error, Reason} -> erlang:error(Reason)
+    end.
 
 getopts(Sock, Options) when ?IS_SSL(Sock) ->
     ssl:getopts(Sock, Options);


michaelklishin:
Some basic load tests suggest that this introduces a 1-2% regression on OTP 25 but on OTP 26, the same workload performs 10-13% better. Some examples:

```
## main run 0
id: test-105939-786, sending rate avg: 145274 msg/s
id: test-105939-786, receiving rate avg: 145170 msg/s
id: test-105939-786, consumer latency min/median/75th/95th/99th 312850/475025/532713/695496/841841 µs

## server-7927 run 0
id: test-111609-202, sending rate avg: 143476 msg/s
id: test-111609-202, receiving rate avg: 143359 msg/s
id: test-111609-202, consumer latency min/median/75th/95th/99th 283027/590431/659112/772635/954561 µs

## main run 1
id: test-115132-193, sending rate avg: 144961 msg/s
id: test-115132-193, receiving rate avg: 144754 msg/s
id: test-115132-193, consumer latency min/median/75th/95th/99th 309474/462129/511390/599758/849369 µs

## server-7927 run 1
id: test-112843-602, sending rate avg: 141274 msg/s
id: test-112843-602, receiving rate avg: 140872 msg/s
id: test-112843-602, consumer latency min/median/75th/95th/99th 212112/615371/803510/1069744/1223967 µs

## server-7927 run 2 (OTP 26)
id: test-120306-642, sending rate avg: 161964 msg/s
id: test-120306-642, receiving rate avg: 161799 msg/s
id: test-120306-642, consumer latency min/median/75th/95th/99th 288628/461285/531218/698923/805594 µs

## server-7927 run 3 (OTP 26)
id: test-121701-470, sending rate avg: 165165 msg/s
id: test-121701-470, receiving rate avg: 164740 msg/s
id: test-121701-470, consumer latency min/median/75th/95th/99th 181347/423884/477185/677074/834184 µs
```

This is on a 10 core aarch64 machine with the following PerfTest flags:

```
java -jar ./perf-test.jar -x 3 -y 3 --quorum-queue --queue-pattern "qq.%d" --queue-pattern-from 1 --queue-pattern-to 3 -f persistent --aut-ack
```

michaelklishin (approved):
Some basic load tests suggest that this introduces a 1-2% regression on OTP 25 but on OTP 26, the same workload performs 10-13% better.

This seems perfectly acceptable to me. Later this year we likely will recommend 26 to all 3.12 users.

Issue #7939, opened 2023-04-19
3.12: CQ with TTL seems to be spinning in place after a restart


luos:
### Describe the bug

Hi,

I was testing some changes and I was using the AMQP 1.0 plugin to publish into a Classic Queue, however I do not think this issue is related to the AMQP 1.0 plugin. 

Anyway, this is just a heads up if maybe you are experiencing the same. 

It can not be reproduced consistently, but 

### Reproduction steps

1. Checkout main
2. Use Erlang 25.0.4
1. Set up a classic queue
3. Add a TTL policy of 30 seconds
4. Restart RabbitMQ
5. Use publish confirms but do not actually wait for confirms except at the end of the run
6. Publish 400 000 reasonably small messages with some headers
7. Wait for confirms, disconnect
8. Wait a few seconds 
9. Publish 400 000 reasonably small messages with some headers (This will trigger the issue probably)
10. Wait for confirms (it will time out)
11. Repeat from 4 if it did not get stuck.

Observe the following: 
1. Mgmt interface shows 0 messages
2. Channel shows 400 000 unconfirmed messages
3. It takes about 2-3 minutes to churn through the 400k messages on my macbook. 

![kép](https://user-images.githubusercontent.com/923194/233105501-bebb5ac4-56ee-4b1a-ba84-77c4d3428be6.png)

On RabbitMQ Top, it is visible that the Erlang Mailbox / Gen Server 2 is full with 400k messages. 

The queue is spinning in these stacktraces: 

```
[{prim_file,read_info_1,3,[]},
 {file_handle_cache,with_handle,2,[{file,"file_handle_cache.erl"},{line,572}]},
 {rabbit_file,file_size,1,[{file,"rabbit_file.erl"},{line,52}]},
 {rabbit_queue_index,load_segment,2,
                     [{file,"rabbit_queue_index.erl"},{line,1239}]},
 {rabbit_queue_index,segment_entries_foldr,3,
                     [{file,"rabbit_queue_index.erl"},{line,1224}]},
 {rabbit_queue_index,read_bounded_segment,5,
                     [{file,"rabbit_queue_index.erl"},{line,1206}]},
 {rabbit_queue_index,read,3,[{file,"rabbit_queue_index.erl"},{line,500}]}]

```

```
[{gen,do_call,4,[{file,"gen.erl"},{line,237}]},
 {gen_server2,call,3,[{file,"gen_server2.erl"},{line,342}]},
[...]

PR #7980, opened 2023-04-25
CQv1: Don't limit messages in memory based on consume rate

lhoguin:
The v1 index is not optimised for reading messages except when the entire segment is read. So we always do that.

This change was made because when the read is inefficient and TTL is used the queue can get unresponsive while getting the TTL messages dropped. In that case the queue may drop messages slower than they expire and as a result will not process any Erlang messages until it has dropped all messages in the queue.

Fix for #7939


lhoguin:
There's no need to document this in the 3.12 release notes because this is a fix to changes done for 3.12.

michaelklishin (approved):

PR #7981, opened 2023-04-25
Allow setting consumer timeout via queue policy/arg and as consumer argument

### LINKED ISSUE #5437: Spike: per-queue and per-consumer delivery acknowledgement timeout
I won't recite the discussion in #4099 but there's some interest on our team in making [consumer delivery acknowledgement timeout](https://www.rabbitmq.com/consumers.html#acknowledgement-timeout) configurable per virtual host, queue or consumer, and not just globally.

Starting with 3.11, virtual host metadata is a pre-enabled feature flag, so there is interest in using it more.

This may bring us closer to taking the setting from the optional queue arguments, if any, but this is not in scope for the spike.

Suggested duration: no more than two full days.

ikvmw:
Close #5437

## Proposed Changes

Add ability to set Consumer Acknowledgement Timeout via a queue/consumer argument as well as a queue policy

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_channel.erl
+++ b/deps/rabbit/src/rabbit_channel.erl
@@ -2818,8 +2818,7 @@ get_queue_consumer_timeout(_PA = #pending_ack{queue = QName},
     end.
 
 get_consumer_timeout(PA = #pending_ack{tag  = CTag},
-                     State = #ch{consumer_mapping = CMap,
-                                 cfg = #conf{consumer_timeout = GCT}}) ->
+                     State = #ch{consumer_mapping = CMap}) ->
     case maps:find(CTag, CMap) of
         {ok, {_, {_, _, _, Args}}} ->
 	    case rabbit_misc:table_lookup(Args, <<"x-consumer-timeout">>) of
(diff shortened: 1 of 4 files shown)


ikvmw:
In which version it might go? I need to update release notes and mention on the website

deps/rabbit/src/rabbit_channel.erl
```diff
@@ -2804,34 +2804,68 @@ get_operation_timeout_and_deadline() ->
...
+		undefined -> GCT;
+		Val -> Val
+	    end;
+	_ ->
+	    GCT
+    end.
+
+get_consumer_timeout(PA = #pending_ack{tag  = CTag},
+                     State = #ch{consumer_mapping = CMap,
+                                 cfg = #conf{consumer_timeout = GCT}}) ->
```
michaelklishin, line 2822: This value is unused and it breaks `bazel build`.

michaelklishin (changes requested):

michaelklishin:
We should document the fact that this timeout is (re-)evaluated every so often. When using a low (5-10s) timeout in consumer args some would expect that it will take effect instantly but that takes a moment. Still, this lets folks override the timeout for some queues or consumers and since "override" usually means increase, this initial delay seems perfectly acceptable.

michaelklishin (approved):

PR #7982, opened 2023-04-25
Use rules erlang 3.9.14

HoloRin:
rules_erlang 3.9.14 fixes some bugs in gazelle based bazel file generation/maintenance


pjk25:
Not backported to `v3.10.x` because the change is already applied in #7954

Issue #7823, opened 2023-04-04
AMQP 1.0: notify clients when a message could not be routed to any queues


kjnilsson:
### Is your feature request related to a problem? Please describe.

Right now when publishing a message using AMQP 1.0 and no queues got routed a disposition frame with the `accepted` outcome is returned (equivalent tp AMQP legacy publisher confirm). To avoid this a different outcome should be selected.

### Describe the solution you'd like

When a message cannot be routed to any queues the `released` outcome should be returned.

In addition we could consider returning an error when a link with a queue target (e.g. `/amq/queue/QueueName`) is attached and the queue does not exist.

This feature should always be active.

acogoluegnes:
I'm investigating an "internal" extension to the mandatory flag mechanism to provide the delivery ID in the callback. Publisher confirms and the mandatory flag would be enabled both anyway, as the AMQP 1.0 settled mode assumes outbound messages are already settled (so no need for the disposition frame).

PR #8015, opened 2023-04-27
Settle unroutable message with released state

acogoluegnes:
Currently it is not possible for an AMQP 1.0 to make find out a message is unroutable as it is settled with the accepted state. This is because the current implementation relies only on the publish confirms AMQP 091 extension.

This commit changes this by settling the message with the released state. It uses the mandatory flag mechanism from AMQP 091 and "internally" extends it to provide not only the message in the callback, but the publishing sequence as well. This applies only for AMQP 1.0, not for other cases. Publish confirms and the mandatory flag are used in conjonction in this case.

References #7823

### DIFF
--- a/deps/rabbitmq_amqp1_0/test/amqp10_client_SUITE.erl
+++ b/deps/rabbitmq_amqp1_0/test/amqp10_client_SUITE.erl
@@ -24,6 +24,7 @@ groups() ->
     [
      {tests, [], [
                   reliable_send_receive_with_outcomes,
+                  publishing_to_non_existing_queue_should_settle_with_released,
                   roundtrip_classic_queue_with_drain,
                   roundtrip_quorum_queue_with_drain,
                   roundtrip_stream_queue_with_drain,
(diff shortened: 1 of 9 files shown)


kjnilsson:
I think this looks like on OK work around for this until we can get native AMQP 1.0 support: 👍

kjnilsson (approved):

deps/rabbitmq_amqp1_0/test/amqp10_client_SUITE.erl
```diff
@@ -151,6 +153,68 @@ reliable_send_receive(Config, Outcome) ->
...
+    ok = amqp10_client:close_connection(Connection),
+    flush("post sender close"),
+    ok.
+
+open_link_to_non_existing_destination_should_end_session(Config) ->
+    Container = atom_to_list(?FUNCTION_NAME),
+    Name = Container ++ "foo",
+    Addresses = [
+                 "/exchange/" ++ Name ++ "/bar",
+                 "/amq/queue/" ++ Name
```
ansd, line 193: @acogoluegnes should we also test target addresses `"/queue/" ++ Name` and `Name`?

acogoluegnes: The queue is automatically created for these 2 kinds of addresses IIRC. Feel free to give it a try or have a look at the code, but I'm pretty that's the case.

ansd: You're right, thanks.
I got confused by the test name `open_link_to_non_existing_destination_should_end_session`:
Opening a link to a non existing target should actually work for addresses `"/queue/" ++ Name` and `Name` and only fail non existing exchanges and `"/amq/queue/"`  because the latter has a special meaning:
> Queue names starting with "amq." are reserved for pre-declared and standardised queues.

acogoluegnes: Sure, no problem. I kept the original behavior, a module enforces the existence of `/queue/` and `queue-name` addresses.

PR #8080, opened 2023-05-02
Improve validation of commands that accept replica candidate nodes as arguments

### LINKED ISSUE #8007: Add explicit CLI check for commands like `rabbitmq-queues grow`

Taking `rabbitmq-queues grow` as example. This command returns status code 0 and a log message hinting it did its intended job, even when the user targets a node that does not exist. For example:

```shell
rabbitmq-queues grow 'this-node-does-not-exist' all --queue-pattern some-queue
# Growing all quorum queues on this-node-does-not-exist...
```

The CLI checks whether the target node is part of the cluster, and returns an error if it's not. 

_No response_

### Additional context

This was tested in RabbitMQ 3.11.10. This would be applicable to other CLI commands. I don't have enough knowledge of the internals to suggest other specific CLI commands 🙃

michaelklishin:
Specifically `rabbitmq-queues grow` and `rabbitmq-queues add_member`. Note that I intentionally
did not do the same for `shrink` and `delete_member`, as it may be reasonable to support
scenarios where they would be used with permanently lost or decommissioned nodes.
Well, this does not make that much sense for `shrink` but for `delete_member`, it might.

## QA Procedure

1. Start a node with `bazel run broker`
2. Run `bazel run rabbitmq-queues -- grow unknown@{hostname} all` and see what error it reports
3. Run `bazel run rabbitmq-queues -- grow rabbit@{hostname} all`
4. Run `bazel run rabbitmq-queues -- add_member "qq.1" unknown@{hostname}` and see what error it reports
5. Run `bazel run rabbitmq-queues -- add_member "qq.1" rabbit@{hostname}` and observe an improved error
6. Declare a quorum queue "qq.1" in the default virtual host
7. Run `bazel run rabbitmq-queues -- add_member "qq.1" rabbit@{hostname}` one more time

## Other

Closes #8007.

### DIFF
--- a/deps/rabbit/test/rabbitmq_queues_cli_integration_SUITE.erl
+++ b/deps/rabbit/test/rabbitmq_queues_cli_integration_SUITE.erl
@@ -120,8 +120,8 @@ grow_invalid_node_filtered(Config) ->
     #'queue.declare_ok'{} = declare_qq(Ch, QName, Args),
     DummyNode = not_really_a_node@nothing,
     publish_confirm(Ch, QName),
-    {ok, Out1} = rabbitmq_queues(Config, 0, ["grow", DummyNode, "all"]),
-    ?assertNotMatch(#{{"/", "grow-err"} := _}, parse_result(Out1)),
+    %% validated as of rabbitmq-server#8007
+    {error, _ExitCode, _} = rabbitmq_queues(Config, 0, ["grow", DummyNode, "all"]),
     ok.
 
 parse_result(S) ->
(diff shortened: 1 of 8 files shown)


pjk25:
Given

```
INFO: Running command line: bazel-bin/rabbitmq-queues add_member qq.1 rabbit@kuryloskipZMD6T
Adding a replica for queue qq.1 on node rabbit@kuryloskipZMD6T...
Object (queue, stream, exchange, etc) was not found
```

Can we change the error message to `Object qq.1 was not found`?

michaelklishin:
@pjk25 it now says "queue '…' was not found in virtual host '…'"

pjk25:
@michaelklishin something seems off in the virtual host printing: `queue 'qq.12' was not found in virtual host '/}'`

michaelklishin:
There is a typo on line 478 in rabbitmqctl.ex

PR #8098, opened 2023-05-04
Add MQTT client id to connection closed event

ansd:
As requested in https://github.com/rabbitmq/rabbitmq-server/discussions/6331#discussioncomment-5796154 include all infos that were emitted in the MQTT connection created event also in the MQTT connection closed event.
This ensures infos such as MQTT client ID are part of the connection closed event.
Therefore, it's easy for the user to correlate between the two event types.
Note that the MQTT plugin emits connection created and connection closed events only if the CONNECT packet was successfully processed, i.e.authentication was successful.


ChunyiLyu (approved):
🚀

michaelklishin (approved):

PR #8112, opened 2023-05-05
rabbit_feature_flags: Fix possible deadlock when calling the Code server

dumbbell:
### Why

The Feature flags registry is implemented as a module called `rabbit_ff_registry` recompiled and reloaded at runtime.

There is a copy on disk which is a stub responsible for triggering the first initialization of the real registry and please Dialyzer. Once the initialization is done, this stub calls `rabbit_ff_registry` again to get an actual return value. This is kind of recursive: the on-disk `rabbit_ff_registry` copy calls the `rabbit_ff_registry` copy generated at runtime.

Early during RabbitMQ startup, there could be multiple processes indirectly calling `rabbit_ff_registry` and possibly triggering that first initialization concurrently. Unfortunately, there is a slight chance of race condition and deadlock:

0. No `rabbit_ff_registry` is loaded yet.
1. Both process A and B call `rabbit_ff_registry:something()` indirectly which triggers two initializations in parallel.
2. Process A acquires the lock first and finishes the initialization. A new registry is loaded and the old `rabbit_ff_registry` module copy is marked as "old". At this point, process B still references that old copy because `rabbit_ff_registry:something()` is up above in its call stack.
3. Process B acquires the lock, prepares the new registry and tries to soft-purge the old `rabbit_ff_registry` copy before loading the new one.

This is where the deadlock happens: process B requests the Code server to purge the old copy, but the Code server waits for process B to stop using it.

### How

With this commit, process B calls `erlang:check_process_code/2` before asking for a soft purge. If it is using an old copy, it skips the purge because it will deadlock anyway.

### DIFF
--- a/deps/rabbit/src/rabbit_ff_registry_factory.erl
+++ b/deps/rabbit/src/rabbit_ff_registry_factory.erl
@@ -19,7 +19,8 @@
          release_state_change_lock/0]).
 
 -ifdef(TEST).
--export([registry_loading_lock/0]).
+-export([registry_loading_lock/0,
+         purge_old_registry/1]).
 -endif.
 
 -define(FF_STATE_CHANGE_LOCK, {feature_flags_state_change, self()}).
@@ -734,9 +735,14 @@ registry_vsn() ->
     proplists:get_value(vsn, Attrs, undefined).
 
 purge_old_registry(Mod) ->
-    case code:is_loaded(Mod) of
-        {file, _} -> do_purge_old_registry(Mod);
-        false     -> ok
+    case erlang:check_process_code(self(), rabbit_ff_registry) of
+        false ->
+            case code:is_loaded(Mod) of
+                {file, _} -> do_purge_old_registry(Mod);
+                false     -> ok
+            end;
+        true ->
+            ok
     end.
 
 do_purge_old_registry(Mod) ->
--- a/deps/rabbit/test/feature_flags_SUITE.erl
+++ b/deps/rabbit/test/feature_flags_SUITE.erl
@@ -25,6 +25,7 @@
 
          registry_general_usage/1,
          registry_concurrent_reloads/1,
+         try_to_deadlock_in_registry_reload/1,
          enable_feature_flag_in_a_healthy_situation/1,
          enable_unsupported_feature_flag_in_a_healthy_situation/1,
          enable_feature_flag_when_ff_file_is_unwritable/1,
@@ -100,7 +101,8 @@ groups() ->
      {registry, [],
       [
        registry_general_usage,
-       registry_concurrent_reloads
+       registry_concurrent_reloads,
+       try_to_deadlock_in_registry_reload
       ]},
      {feature_flags_v2, [], Groups}
     ].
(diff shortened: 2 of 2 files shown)


dumbbell:
I should have shared more context on how this issue was reported. It was first reported by a user who hit this deadlock while starting RabbitMQ as a Windows Service.

When RabbitMQ is started as a Windows Service, the Erlang node must start with Erlang distribution enabled right from the command line. This is unlike other ways of starting RabbitMQ where distribution is enabled at runtime by RabbitMQ itself after running several checks.

Therefore, the Erlang node started as a Windows Service is reachable by the RabbitMQ CLI very early, even before the `rabbit` application even starts. The deadlock can be triggered regardless of the time the distribution is enabled. But if e.g. `rabbitmqctl status` runs concurrently to the start of the `rabbit` application, this increases the chance of hitting that deadlock.

I mention `rabbitmq status` because, as part of listing listeners, it queries the state of the `listener_records_in_ets` feature flag which was introduced in RabbitMQ 3.11.0. The process stacktraces shared by the original reporter shows a process querying that feature flag.

Here is the stacktrace of a process waiting for the Code server waiting for the calling process, in a node running RabbitMQ 3.11.5 on Erlang/OTP 25.3.1:

```erlang
2023-05-04 11:46:02 [{pid,<11965.29263.0>},
                     {registered_name,[]},
                     {current_stacktrace,
                          %% The process is waiting for a reply from the Code server.

                         [{code_server,call,1,
                              [{file,"code_server.erl"},{line,139}]},
                          {rabbit_ff_registry_factory,do_purge_old_registry,
                              1,
                              [{file,"rabbit_ff_registry_factory.erl"},
                               {line,707}]},
                          {rabbit_ff_registry_factory,load_registry_mod,3,
                              [{file,"rabbit_ff_registry_factory.erl"},
                               {line,657}]},
                          {rabbit_ff_registry_factory,
                              maybe_initialize_registry,3,
                              [{file,"rabbit_ff_registry_factory.erl"},
                               {line,295}]},
                          {rabbit_ff_registry_factory,initialize_registry,3,
                              [{file,"rabbit_ff_registry_factory.erl"},
                               {line,149}]},

                          %% rabbit_ff_registry (below) calling
                          %% rabbit_ff_registry_factory (above) is the very
                          %% first initialization.

                          {rabbit_ff_registry,is_enabled,1,
                              [{file,"rabbit_ff_registry.erl"},{line,134}]},
                          {rabbit_feature_flags,is_enabled,2,
                              [{file,"rabbit_feature_flags.erl"},{line,755}]},
                          {rabbit_networking,node_listeners,1,
                              [{file,"rabbit_networking.erl"},{line,461}]}]},
                     {initial_call,{erpc,execute_call,4}},
                     {message_queue_len,0},
                     {links,[]},
                     {monitors,[{process,{code_server,rabbit@XYZ}}]},
                     {monitored_by,[<12651.10216.2425>,<11965.54.0>]},
                     {heap_size,4185}]
```

In parallel, here is the Code server stuck, waiting for the process above to run so it can handle the pending signal (an internal ERTS mechanism):

```erlang
2023-05-04 11:46:02 [{pid,<11965.50.0>},
                     {registered_name,code_server},
                     {current_stacktrace,
                         [{erts_code_purger,soft_purge,1,
                              [{file,"erts_code_purger.erl"},{line,135}]},
                          {code_server,handle_call,3,
                              [{file,"code_server.erl"},{line,320}]},
                          {code_server,loop,1,
                              [{file,"code_server.erl"},{line,154}]}]},
                     {initial_call,{erlang,apply,2}},
                     {message_queue_len,0},
                     {links,[<11965.49.0>]},
                     {monitors,[]},
                     {monitored_by,[<11965.29263.0>]},
                     {heap_size,28690}]
```

PR #8136, opened 2023-05-11
Ra 2.6.0

kjnilsson:
This Ra release includes fixes for the default snapshotting module to avoid failed CRC checks for snapshots taken on OTP version prior to 26 that are sent to nodes running OTP 26.

This also includes an API for forcing a single Ra node into a single node cluster which can be used to recover some data even after a quorum has been lost (not implemented in RabbitMQ yet).


michaelklishin:
Superseded by #8143

PR #8188, opened 2023-05-15
Move plugin rabbitmq-message-timestamp to the core

ansd:
As reported in https://groups.google.com/g/rabbitmq-users/c/x8ACs4dBlkI/ plugins that implement rabbit_channel_interceptor break with Native MQTT in 3.12 because Native MQTT does not use rabbit_channel anymore. Specifically, these plugins don't work anymore in 3.12 when sending a message from an MQTT publisher to an AMQP 0.9.1 consumer.

Two of these plugins are
https://github.com/rabbitmq/rabbitmq-message-timestamp and
https://github.com/rabbitmq/rabbitmq-routing-node-stamp

This commit moves both plugins into rabbitmq-server. Therefore, these plugins are deprecated starting in 3.12.

Instead of using these plugins, the user gets the same behaviour by configuring rabbitmq.conf as follows:
```
message_interceptors.incoming.set_header_timestamp.overwrite = false
message_interceptors.incoming.set_header_routing_node.overwrite = false
```

While both plugins were incompatible to be used together, this commit allows setting both headers.

We name the top level configuration key `incoming_message_interceptors` because only incoming messages are intercepted.
Currently, only `set_header_timestamp` and `set_header_routing_node` are supported. (We might support more in the future.)
Both can set `overwrite` to `false` or `true`.
The meaning of `overwrite` is the same as documented in https://github.com/rabbitmq/rabbitmq-message-timestamp#always-overwrite-timestamps i.e. whether headers should be overwritten if they are already present in the message.

Both `set_header_timestamp` and `set_header_routing_node` behave exactly to plugins `rabbitmq-message-timestamp` and `rabbitmq-routing-node-stamp`, respectively.

Upon node boot, the configuration is put into persistent_term to not cause any performance penalty in the default case where these settings are disabled.

The channel and MQTT connection process will intercept incoming messages and - if configured - add the desired AMQP 0.9.1 headers.

For now, this allows using Native MQTT in 3.12 with the old plugins behaviour.

In the future, once "message containers" are implemented, we can think about more generic message interceptors where plugins can be written to modify arbitrary headers or message contents for various protocols.

Likewise, in the future, once MQTT 5.0 is implemented, we can think about an MQTT connection interceptor which could function similar to a `rabbit_channel_interceptor` allowing to modify any MQTT packet.

TODO:
* [ ] Document in https://github.com/rabbitmq/rabbitmq-message-timestamp and https://github.com/rabbitmq/rabbitmq-routing-node-stamp that these plugins are deprecated starting with 3.12 and explain the new `rabbitmq.conf` usage.

### DIFF
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -499,6 +499,11 @@ rabbitmq_integration_suite(
 )
 
 rabbitmq_integration_suite(
+    name = "rabbit_message_interceptor_SUITE",
+    size = "medium",
+)
+
+rabbitmq_integration_suite(
     name = "message_size_limit_SUITE",
     size = "medium",
 )
--- a/deps/rabbit/app.bzl
+++ b/deps/rabbit/app.bzl
@@ -139,6 +139,7 @@ def all_beam_files(name = "all_beam_files"):
             "src/rabbit_looking_glass.erl",
             "src/rabbit_maintenance.erl",
             "src/rabbit_memory_monitor.erl",
+            "src/rabbit_message_interceptor.erl",
             "src/rabbit_metrics.erl",
             "src/rabbit_mirror_queue_coordinator.erl",
             "src/rabbit_mirror_queue_master.erl",
@@ -379,6 +380,7 @@ def all_test_beam_files(name = "all_test_beam_files"):
             "src/rabbit_looking_glass.erl",
             "src/rabbit_maintenance.erl",
             "src/rabbit_memory_monitor.erl",
+            "src/rabbit_message_interceptor.erl",
             "src/rabbit_metrics.erl",
             "src/rabbit_mirror_queue_coordinator.erl",
             "src/rabbit_mirror_queue_master.erl",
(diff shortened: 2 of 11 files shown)


michaelklishin:
Some results from my test environment. Every run was for at least five minutes,
and the data above are averages/percentiles over the entire run.

The PerfTest flags used were

```
./scripts/PerfTest -x 1 -y 1 --quorum-queue --queue-pattern "qq.%d" --queue-pattern-from 1 --queue-pattern-to 1 -f persistent --output-file /tmp/perftest-8188-run-….csv --id perftest-8188-run-…
```

## Run 1: 8188 with one interceptor configured
```
id: server-8188-run-1, sending rate avg: 39847 msg/s
id: server-8188-run-1, receiving rate avg: 39788 msg/s
id: server-8188-run-1, consumer latency min/median/75th/95th/99th 324322/548501/600280/761488/880511 µs
```

## Run 2: 8188 with all defaults

```
id: server-8188-run-2-all-defaults, sending rate avg: 41293 msg/s
id: server-8188-run-2-all-defaults, receiving rate avg: 41226 msg/s
id: server-8188-run-2-all-defaults, consumer latency min/median/75th/95th/99th 511655/691623/766717/888635/1062117 µs
```

## Run 3 (main)
```
id: server-8188-run-3-main, sending rate avg: 39934 msg/s
id: server-8188-run-3-main, receiving rate avg: 39908 msg/s
id: server-8188-run-3-main, consumer latency min/median/75th/95th/99th 456683/758354/868626/1111261/1257522 µs
```

In other words, there is no regression (I would not expect any improvements, despite the data above) for the "no interceptors used" case and a ≈ 4-5% throughput drop with one interceptor.

michaelklishin (approved):

PR #8209, opened 2023-05-16
Check additional applications when comparing bazel and make results

HoloRin:
This will confirm that deps like ra and osiris are also equivalent


michaelklishin:
@mergifyio backport v3.12.x v3.11.x

PR #8277, opened 2023-05-23
Omit log if no channel closed

ansd:
Users were confused by the following log line:

> Closing all channels from connection '10.10.0.1:41128 -> 10.10.0.31:5672' because it has been closed

as for example output every 10 seconds by TCP health checks in RabbitMQ pods deployed by the rabbitmq/cluster-operator.

Let's not log anything if there is no channel for the connection being closed.
Let's also include how many channels are closed if a connection gets closed.

New log line loos like:
> Closing 2 channel(s) because connection '127.0.0.1:41374 -> 127.0.0.1:5672' has been closed


ChunyiLyu (approved):
👍

PR #8308, opened 2023-05-23
Default to classic queues v2

mkuratczyk:
Change the classic queue version default to 2 in preparation for 3.13.


michaelklishin:
We've have decided to revert it because in certain specific scenarios, upgrading CQs at node boot time can be potentially unsafe (but doing the same thing after all nodes have been upgraded and rebooted is safe).

Oh well.

michaelklishin:
We've decided to revert this default change :( https://github.com/rabbitmq/rabbitmq-server/pull/8308

michaelklishin:
#10337 reverts the key part of this change. For now, one day CQv1 will be removed completely.

PR #8322, opened 2023-05-24
Disaster recovery: force shrink all quorum queues to a 1-node cluster

dcorbacho:
In case of a catastrophic failure, this function can be used to force all quorum queues to shrink the membership to the current node. It is meant just for disaster recovery scenarios, where the rest of the cluster is irrecoverable and there are quorum queues in minority that hold important data that must be recovered.

Usage:
`rabbitmqctl eval 'rabbit_quorum_queue:force_all_queues_shrink_member_to_current_member().'`

### DIFF
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -334,6 +334,13 @@ rabbitmq_integration_suite(
 )
 
 rabbitmq_integration_suite(
+    name = "clustering_recovery_SUITE",
+    size = "medium",
+    shard_count = 2,
+    sharding_method = "case",
+)
+
+rabbitmq_integration_suite(
     name = "config_schema_SUITE",
     size = "medium",
     data = [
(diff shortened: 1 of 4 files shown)


michaelklishin:
Should we expose this as a `rabbitmq-queues` command?

michaelklishin:
OK, we agreed to not expose it as a CLI command since it is not meant to be used outside of genuine disaster recovery scenarios. We may reconsider in the future, but this should not be used as a yet another `rabbitmqctl force_boot`.

kjnilsson:
I wonder if it is worth having a function to do this for a single queue also?

dcorbacho:
> I wonder if it is worth having a function to do this for a single queue also?

We can do 👍

michaelklishin (approved):

michaelklishin:
@mergifyio backport v3.12.x

PR #8339, opened 2023-05-25
Handle missing guids on in node_monitor

olikasg:
## Proposed Changes

On `main` and 3.12.x a recent change in node_monitor can prevent partial partition detection when a previous partition handling starts multiple nodes at the same time.

When two or more nodes start at the same time, we have to use the list of running mnesia nodes instead of the fully running rabbit nodes. When a node is in the node_monitor's init stage, it isn't counted as a fully running rabbit node. This means that its GUID isn't sent to the other node starting up. This race condition means that the node_monitor will not do anything with a node_down event, because the down node will be missing from the GUID map.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

To reproduce:
1. Set up a 3-node cluster with pause_minority.
2. Create a partial partition and wait for two of the nodes to stop as expected.
3. Remove the partial partition and observe that both stopped nodes start up.
4. Check that the node monitor state on the newly started nodes doesn't contain the correct guids by running `rabbitmqctl eval 'sys:get_state(rabbit_node_monitor).'`
5. Create another partitial partition on the same node and observe that none of the nodes don't stop as expected. This is a split brain situation, so any changes on the partitioned node will result in inconsistent mnesia.
6. Remove the partial partition and observe the mnesia running partition even.

With the fix the second partition also works as expected, i.e. two nodes stop in the cluster.
(We didn't verify it, but this can be a potential issue with regular node start ups too. When more than one node start up at the same time e.g. after a full partition, automated maintenance activities, docker compose starts.)

### DIFF
--- a/deps/rabbit/src/rabbit_node_monitor.erl
+++ b/deps/rabbit/src/rabbit_node_monitor.erl
@@ -413,7 +413,7 @@ handle_call(_Request, _From, State) ->
     {noreply, State}.
 
 handle_cast(notify_node_up, State = #state{guid = GUID}) ->
-    Nodes = rabbit_nodes:list_running() -- [node()],
+    Nodes = rabbit_nodes:list_reachable() -- [node()],
     gen_server:abcast(Nodes, ?SERVER,
                       {node_up, node(), rabbit_db_cluster:node_type(), GUID}),
     %% register other active rabbits with this rabbit


michaelklishin (changes requested):
Thank you for digging in but we will not accept a change that goes back to `rabbit_mnesia`. `rabbit_nodes` is the public API going forward.

If an important function is missing there, it should be added.
Mnesia will be completely removed from RabbitMQ by early 2024 or so.

olikasg:
I changed the implementation to `list_reachable`. It is not equivalent to the old behaviour but it is very close. It can potentially do some unnecessary `abcast`s but that should not cause any issues. Tested this too, and it seems to be working.

michaelklishin (approved):

PR #8341, opened 2023-05-25
Use 3.12.6 for the secondary umbrella

pjk25:
Set the secondary umbrella for mixed version testing to the latest v3.12.x

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -487,7 +487,7 @@ secondary_umbrella = use_extension(
 
 use_repo(
     secondary_umbrella,
-    "rabbitmq-server-generic-unix-3.11",
+    "rabbitmq-server-generic-unix-3.12",
 )
 
 hex = use_extension(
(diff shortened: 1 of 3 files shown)


michaelklishin:
@pjk25 I assume we should update this PR and merge it, or close it for good?

dumbbell:
The Khepri branch will need something like this because it needs the bug fix in #9005. It was merged in 3.11.21 and 3.12.3.

CI fails currently with the following error:
```
ERROR: Error computing the main repository mapping: module not found in registries: com_github_rabbitmq_osiris@1.6.3
```

pjk25:
It fails because the hypothetical merge commit is what is sent to CI. By updating the branch, it should pick up the newer osiris that is on main. I have updated it.

pjk25:
It seems there may be few legitimate failures in `//deps/rabbit:dynamic_ha_SUITE-mixed`? Are these fixed in the khepri branch?

https://app.buildbuddy.io/invocation/df8fcecd-a1a1-4f9a-8cc5-35ea5fafe97f?target=%2F%2Fdeps%2Frabbit%3Adynamic_ha_SUITE-mixed&targetStatus=6

dumbbell:
There is a failure in that same testsuite in the Khepri branch because 3.11.18 doesn't have the maintenance mode Mnesia table replicated. I will look into this one, but it's probably another issue.

PR #8344, opened 2023-05-25
Add looking_glass

pjk25:
Adds looking_glass to the deps/rabbit plt file so that unknowns no longer need be ignored


pjk25:
I would like to wait until looking_glass has been published to BCR before this is merged. Currently this PR uses an override, which when removed I will take this out of draft status.

pjk25:
looking_glass has now been published in BCR

PR #8358, opened 2023-05-26
Observer CLI plugins for QQs

mkuratczyk:
This plugin allows visualising basic quorum queue metrics/counters within `observer`. To access them, press `P` (to go to Plugins) and then `QQ` (by default, you'll see a classic queue plugin).

With some workload, it should look like this:
Screenshot 2023-05-26 at 10 50 38

Note: the PID column is deliberately narrow to have more space for the counters. Ideally we will get rid of it altogether in the future, since QQ processes are named. However, without a PID somewhere in the table, `observer` doesn't know how to display process details (after selecting a row number).


michaelklishin:
I took the liberty of addressing one metric legend TODO based on my [reading of how Ra metrics are stored](https://github.com/rabbitmq/ra/blob/main/src/ra_server.erl#L1411C22-L1412).

michaelklishin (approved):

PR #8404, opened 2023-05-29
Don't divide by zero

mkuratczyk:
Don't divide by zero 🤦 
(if quorum queues are not used / haven't been used yet)


michaelklishin (approved):

michaelklishin:
References #8358.

PR #8432, opened 2023-05-31
Test with OTP 26 in github actions

pjk25:
Drop 25.0 and add 26.0


pjk25:
Separate PR for `v3.12.x` will opened

PR #8353, opened 2023-05-25
When available, use user-provided ip address for listener metadata

ikvmw:
Closes #8242, references #6852.

## Proposed Changes

Currently for web-related listeners user-provided ip-address is simply ignored when listener metadata saved. 
This patch attempts to fix. It employs intentional copypasta style, because refactoring all this is a much bigger effort it seems.

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [x] Bug fix (non-breaking change which fixes issue #NNNN)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
If you're unsure about any of them, don't hesitate to ask on the mailing list.
We're here to help!
This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

## Further Comments

With this config:
```
[{rabbit, [{tcp_listeners, [{"localhost", 5672}]}]},
 {rabbitmq_web_stomp, [{tcp_config, [{ip, "127.0.0.1"}]}]},
 {rabbitmq_web_mqtt, [{tcp_config, [{ip, "127.0.0.1"}]}]},
 {rabbitmq_prometheus, [{tcp_config, [{ip, "127.0.0.1"}]}]},
 {rabbitmq_management,[{tcp_config, [{ip, "127.0.0.1"}]}]}].
```

I was getting that: 

```
Asking node rabbit@arch to report its protocol listeners ...
Interface: [::], port: 15672, protocol: http, purpose: HTTP API
Interface: [::], port: 1883, protocol: mqtt, purpose: MQTT
Interface: [::], port: 15675, protocol: http/web-mqtt, purpose: MQTT over WebSockets
Interface: [::], port: 61613, protocol: stomp, purpose: STOMP
[...]

PR #8440, opened 2023-06-01
rabbit_networking:ranch_ref/1: use the user-provided IP address/interface

michaelklishin:
Otherwise the ref is miscalculated and certain code paths that depend on the ref correctness will fail. For example,
`rabbit_maintenance:{suspend,resume}_all_client_listeners/0`.

The issue was exposed by the changes in #8353, which shipped in `3.11.17`.
Closes #8415.

Pair: @lhoguin


michaelklishin:
@mergifyio backport main v3.12.x

michaelklishin:
Forward ported manually to `v3.12.x` and `main`.

PR #8502, opened 2023-06-08
rabbit_db: Further split Mnesia-specific and non-Mnesia-specific clustering steps

dumbbell:
### Why

To ease the introduction of Khepri and the future removal of Mnesia, we need to extract the non-Mnesia-related code from `rabbit_mnesia`. This was started in #6882 and this patch continues that work.

### How

This time, the peer discovery driving code is moved to `rabbit_peer_discovery` and the init and join steps are moved to `rabbit_db` and `rabbit_db_cluster`.

The `cluster_status` files management is considered to be Mnesia-specific and is kept in `rabbit_mnesia` and `rabbit_node_monitor`. We will see in the future if this is needed for Khepri, but likely not.


deps/rabbit/src/rabbit_db.erl
```diff
@@ -88,7 +111,9 @@ force_reset_using_mnesia() ->
     ?LOG_DEBUG(
       "DB: resetting node forcefully",
       #{domain => ?RMQLOG_DOMAIN_DB}),
-    rabbit_mnesia:force_reset().
+    ok = rabbit_mnesia:force_reset(),
+    ok = rabbit_node_monitor:reset_cluster_status(),
```
dcorbacho, line 115: I think `rabbit_node_monitor:reset_cluster_status()` should be called from `force_reset()` just after Mnesia ops:
```
force_reset() ->
 force_reset_using_mnesia(),
 ok = rabbit_node_monitor:reset_cluster_status().
```

deps/rabbit/test/clustering_management_SUITE.erl
```diff
@@ -752,11 +752,19 @@ equal(Status0, Status1) ->
...
+    try
+        erpc:call(Node, rabbit_db_cluster, current_or_last_status, [])
+    catch
+        error:{exception, undef,
+               [{rabbit_db_cluster, current_or_last_status, _, _} | _]} ->
+            erpc:call(Node, rabbit_mnesia, cluster_status, [status])
+    end,
     {rpc:call(Node, rabbit_nodes, list_members, []),
      rpc:call(Node, rabbit_nodes, list_running, []),
-     rpc:call(Node, rabbit_mnesia, cluster_nodes, [all]),
```
dcorbacho, line 757: This willfully calls `rabbit_mnesia` so that when a new metadata store is used, we can compare `mnesia` (which is always running at least until 4.x) with the status of the currently enabled store (returned by `rabbit_nodes` above, could be `mnesia` or `khepri` in the future). That said, if we decide to disable `mnesia` as soon as we enable another store, the test as it was needs to be updated.

deps/rabbit/src/rabbit_db.erl
dumbbell: Good catch! Thank you!

deps/rabbit/test/clustering_management_SUITE.erl
dumbbell: As discussed outside of this pull request with Diana, Mnesia and Khepri will not be used at the same time. Therefore, we can start to reduce the use of `rabbit_mnesia`.

dcorbacho: Yes, that can now be removed

the-mikedavis (commented):
I see a few typos in the commit messages

* 8dc68fd03d475c3fb289bdd4dfcd3d152d565130: "some or Mnesia-specific" => "some are Mnesia-specific"
* 5f73cb43f5b8d7d3432d560a58a69941de44c5ba: "single node or a cluster" => "single node of a cluster"

Otherwise, just some minor comments

deps/rabbit/test/peer_discovery_classic_config_SUITE.erl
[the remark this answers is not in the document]
dumbbell: It would be possible to reduce the timeout for testsuites for instance (when `TEST` is defined) and I thought about that. It would be a benefit for this specific testcase. However, I fear that it would hide other timing issues elsewhere.

To be honest, I don't like the retry change I made to the feature flags subsystem. It is supposed to fix a problem related to network issues, but we could never reproduce and we have no evidence that it fixed anything for the couple users who reported network problems affecting RabbitMQ startup at the feature flags init phase.

deps/rabbit/src/rabbit_db_cluster.erl
```diff
@@ -144,6 +188,21 @@ node_type() ->
...
+%% function.
+
+check_compatibility(RemoteNode) ->
+    case rabbit_feature_flags:check_node_compatibility(RemoteNode) of
+        ok    -> check_compatibility_using_mnesia(RemoteNode);
+        Error -> Error
+    end.
+
+check_compatibility_using_mnesia(RemoteNode) ->
+    rabbit_mnesia:check_mnesia_consistency(RemoteNode).
```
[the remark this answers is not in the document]
dumbbell: Good idea! I forgot that we already had `rabbit_db:run/1` in `main`.

dumbbell:
Thank you for the comments!

>     * [5f73cb4](https://github.com/rabbitmq/rabbitmq-server/commit/5f73cb43f5b8d7d3432d560a58a69941de44c5ba): "single node or a cluster" => "single node of a cluster"

About this one, "When a single node or a cluster is initialiazed" is what I wanted to say but perhaps I incorrectly phrased or wrote it.

deps/rabbit/src/rabbit_db.erl
dcorbacho, line 1: Do we need a `rabbit_db:run()` choice in `is_virgin_node`? We'll need a Khepri equivalent

deps/rabbit/src/rabbit_db.erl
```diff
@@ -98,7 +122,8 @@ force_reset_using_mnesia() ->
 %% state, like if critical members are MIA.
 
 force_load_on_next_boot() ->
-    force_load_on_next_boot_using_mnesia().
+    rabbit_db:run(
+      #{mnesia => fun force_load_on_next_boot_using_mnesia/0}).
 
 force_load_on_next_boot_using_mnesia() ->
     ?LOG_DEBUG(
```
dcorbacho, line 129: It might not belong to this PR, but the below message is incorrect. Force load doesn't reset the node, it forces the node to start if it was not the last node going down.

deps/rabbit/src/rabbit_mnesia.erl
```diff
@@ -274,6 +162,34 @@ join_cluster(DiscoveryNode, NodeType) ->
...
+    ok = init_db_with_mnesia(ClusterNodes, NodeType,
+                             true, true, _Retry = true),
+    rabbit_node_monitor:notify_joined_cluster(),
+    ok;
+join_cluster(DiscoveryNode, NodeType) when is_atom(DiscoveryNode) ->
+    %% Code to remain compatible `change_cluster_node_type/1' and with older
+    %% CLI.
+    case can_join_cluster(DiscoveryNode) of
+        {ok, ClusterNodes} when is_list(ClusterNodes) ->
+            ok = reset_gracefully(),
```
dcorbacho, line 184: Isn't it already reset in `rabbit_db:join`?

deps/rabbit/src/rabbit_peer_discovery.erl
```diff
@@ -101,6 +114,216 @@ maybe_init() ->
...
+    %% cluster). If we were to wait for that table locally before
+    %% synchronizing feature flags, we would wait forever; indeed the feature
+    %% flag being disabled before sync, `rabbit_table:definitions()' would
+    %% return the old table.
+    %%
+    %% Feature flags need to be synced before any change to Mnesia membership.
+    %% If enabling feature flags fails, Mnesia could remain in an inconsistent
+    %% state that prevents later joining the nodes.
+    IsVirgin = rabbit_db:is_virgin_node(),
+    rabbit_db_cluster:ensure_feature_flags_are_in_sync([RemoteNode], IsVirgin),
```
dcorbacho, line 323: This is called in `rabbit_db:pre_init`. I believe this might be a duplicated call, but not sure as there are too many clustering code paths...

deps/rabbit/src/rabbit_db.erl
dumbbell: Indeed, fixed!

deps/rabbit/src/rabbit_mnesia.erl
dumbbell: The second clause of `rabbit_mnesia:join_cluster()` will only be used by an old CLI (which won't call `rabbit_db` at all). It's unused in the new code path.

deps/rabbit/src/rabbit_peer_discovery.erl
dumbbell: The one in `pre_init` is used when a node boots (regular boot, including during `start_app` after a `join_cluster` for instance). The one in peer discovery is used after `pre_init` if the node joins a cluster because it won't go through `pre_init` again.

deps/rabbit/src/rabbit_db.erl
dumbbell: I believe we need, fixed!

PR #8507, opened 2023-06-09
CQ shared store write optimisations

lhoguin:
Another round of CQ shared message store optimisations and refactors, this time around writes.

The first commit is more a refactor than anything, although no longer calling fsync does help in some scenarios. The second commit should provide a noticeable boost in performance when there are no consumers.

@mkuratczyk I have renamed the branch.

### DIFF
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -998,18 +998,20 @@ stop_sync_timer(State) ->
     rabbit_misc:stop_timer(State, #msstate.sync_timer_ref).
 
 internal_sync(State = #msstate { current_file_handle = CurHdl,
+                                 clients             = Clients,
                                  cref_to_msg_ids     = CTM }) ->
     State1 = stop_sync_timer(State),
-    CGs = maps:fold(fun (CRef, MsgIds, NS) ->
-                            case sets:is_empty(MsgIds) of
-                                true  -> NS;
-                                false -> [{CRef, MsgIds} | NS]
-                            end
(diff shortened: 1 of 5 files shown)


deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -7,15 +7,14 @@
 
 -module(rabbit_msg_store).
 
--behaviour(gen_server2).
+-behaviour(gen_server).
```
lhoguin, line 10: Oops, revert this.

michaelklishin:
The Dialyzer failure is consistent, I can reproduce it locally:

```
rabbit_msg_store.erl:1009:5: Expression produces a value of type 
          'ok' | {'error', atom()}, but this value is unmatched
```

lhoguin:
This is ready for review/merge.

mkuratczyk:
Full test results:
https://snapshots.raintank.io/dashboard/snapshot/oo6eAYwUI2exFVFxdt6mTQIsb7AT2hrt?orgId=2

The biggest difference is in confirm latency, when using `-c 1` (here for 5kb messages, but similar results for all tested):
Screenshot 2023-06-20 at 19 58 06

And when publishing to a long queue (no consumers):
Screenshot 2023-06-20 at 20 02 12

(the sudden drop in throughput is when the queues reach their `max-length` - dropping messages is currently quite expensive)

mkuratczyk (approved):

michaelklishin:
So, a 40% drop in publisher confirm latency on one workload, and a 60% drop on another. Not bad.

PR #8513, opened 2023-06-09
Add a shorthand for the OAuth 2 authN/authZ backend

michaelklishin:
References #8512


michaelklishin:
@mergifyio backport v3.12.x

PR #8553, opened 2023-06-13
Use OTP 26 features

ansd:
See commit messages.
RabbitMQ 3.13 will require OTP 26 as minimum required version.
The OTP 25 GitHub Action tests are expected to fail.

### DIFF
--- a/.github/workflows/test-mixed-versions.yaml
+++ b/.github/workflows/test-mixed-versions.yaml
@@ -127,7 +127,7 @@ jobs:
       fail-fast: false
       matrix:
         otp_version_id:
-        - "25_3"
+        - "26"
     timeout-minutes: 120
     steps:
     - name: CHECKOUT REPOSITORY
@@ -171,7 +171,7 @@ jobs:
     strategy:
       matrix:
         include:
-        - erlang_version: "25.3"
+        - erlang_version: "26.0"
           elixir_version: 1.14.5
     timeout-minutes: 60
     steps:
--- a/deps/rabbit_common/src/rabbit_writer.erl
+++ b/deps/rabbit_common/src/rabbit_writer.erl
@@ -265,10 +265,6 @@ handle_message({send_command_and_notify, QPid, ChPid, MethodRecord, Content},
 handle_message({'DOWN', _MRef, process, QPid, _Reason}, State) ->
     rabbit_amqqueue_common:notify_sent_queue_down(QPid),
     State;
-handle_message({inet_reply, _, ok}, State) ->
-    rabbit_event:ensure_stats_timer(State, #wstate.stats_timer, emit_stats);
-handle_message({inet_reply, _, Status}, _State) ->
-    exit({writer, send_failed, Status});
 handle_message(emit_stats, State = #wstate{reader = ReaderPid}) ->
     ReaderPid ! ensure_stats,
     rabbit_event:reset_stats_timer(State, #wstate.stats_timer);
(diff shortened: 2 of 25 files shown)


michaelklishin:
@ansd should we bump the [minimum required version](https://github.com/rabbitmq/rabbitmq-server/blob/6aa66a938254efd3dd8b51a99479adfa56353607/deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_erlang_compat.erl#L9-L10) then and delete Erlang 25.3 runs from Actions?

michaelklishin:
We'd have to also bump the minimum required version in `rabbitmq/rabbitmq-packaging` (I assume `main` there is only used by the 3.13 pipeline).

deps/rabbitmq_amqp1_0/src/rabbit_amqp1_0_writer.erl
```diff
@@ -142,10 +142,6 @@ handle_message({send_command_and_notify, QPid, ChPid, MethodRecord, Content},
 handle_message({'DOWN', _MRef, process, QPid, _Reason}, State) ->
     rabbit_amqqueue:notify_sent_queue_down(QPid),
     State;
-handle_message({inet_reply, _, ok}, State) ->
-    rabbit_event:ensure_stats_timer(State, #wstate.stats_timer, emit_stats);
```
lhoguin, line 146: Don't we need to call this somewhere?

deps/rabbit_common/src/rabbit_writer.erl
```diff
@@ -265,10 +265,6 @@ handle_message({send_command_and_notify, QPid, ChPid, MethodRecord, Content},
 handle_message({'DOWN', _MRef, process, QPid, _Reason}, State) ->
     rabbit_amqqueue_common:notify_sent_queue_down(QPid),
     State;
-handle_message({inet_reply, _, ok}, State) ->
-    rabbit_event:ensure_stats_timer(State, #wstate.stats_timer, emit_stats);
```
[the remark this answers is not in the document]
ansd: Very good spot, thank you!
Yes, it seems that the AMQP writers deviate from what's described in https://github.com/rabbitmq/rabbitmq-server/blob/b959a92cfa6961e0bc98986e3ca0bee45ae0945a/deps/rabbit_common/src/rabbit_event.erl#L68-L90 in the sense that they don't stop the stats timer before hibernation (and don't start the stats timer upon wakeup). Since the "connection" stats emitting object is shared between two processes (reader & writer), the writer just let's the reader know to emit stats when the writer's timer fires. Furthermore the writer needs to ensure_stats_timer after having written to the socket.
So, I added that behaviour back.

ansd:
@pjk25 does the PR contain everything needed to bump the minimum OTP to 26? I'm not familiar with what exactly needs to be updated in the Bazel files. It seems we will still require OTP 25 for mixed version tests because the lower version nodes will continue to be OTP 25?

pjk25:
> @pjk25 does the PR contain everything needed to bump the minimum OTP to 26? I'm not familiar with what exactly needs to be updated in the Bazel files. It seems we will still require OTP 25 for mixed version tests because the lower version nodes will continue to be OTP 25?

I think you have covered what is necessary. Mixed version tests use the same erlang for all nodes, so the erlang version must be supported by both. Furthermore, we don't currently differentiate between the erlang version used for compilation, vs the version used to run tests. So, for mixed version tests on this branch, as currently configured, compilation of both versions of the broker, and running of tests all happens on 25.3. Supporting something more complicated will require some further enhancements to the build system.

dcorbacho (approved):

ansd:
Thank you. This means mixed version tests must run on OTP 26. I bumped that version in the latest commit.
The currently configured lower version is 3.11.11 which doesn't support 26.
So, we have to merge https://github.com/rabbitmq/rabbitmq-server/pull/8341 first.

michaelklishin:
@ansd we can bump to `3.11.18` which has virtually every Erlang 26-related change that `3.12.0` has.

pjk25:
> @ansd we can bump to `3.11.18` which has virtually every Erlang 26-related change that `3.12.0` has.

#8584

.github/workflows/test-mixed-versions.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
     runs-on: ubuntu-20.04
     env:
       # otp-min
-      otp_version_id: 25_3
+      otp_version_id: 26
```
pjk25, line 37: This should probably stay `25_3` as it's the compile-time version of the secondary umbrella, not the runtime.

pjk25: Furthermore, since we don't name these archives after their erlang version, if the same secondary umbrella version is shared across release branches (as is temporarily the case now) it may break the other workflow.

ansd:
I revert this PR to a draft as I opened it too early and we ended up mixing multiple issues in this single PR. We should first create a separate PR which only bumps the minimum required version to OTP 26 fixing all tests such as the failing `rabbit_stream_queue_SUITE-mixed` group `cluster_size_2` case `recover`. Once all these tests are green and that other PR has been merged, we go the next step and actually use OTP 26 features as done in this PR.

ansd:
Closing since this PR has been delivered by the following individual PRs:
https://github.com/rabbitmq/rabbitmq-server/pull/9787
https://github.com/rabbitmq/rabbitmq-server/pull/9805
https://github.com/rabbitmq/rabbitmq-server/pull/9808
https://github.com/rabbitmq/rabbitmq-server/pull/9809
https://github.com/rabbitmq/rabbitmq-server/pull/9810

PR #8586, opened 2023-06-16
Use a specific erlang version for the secondary umbrella

pjk25:
rather than selecting it dynamically via toolchain resolution

Requires a rules_erlang version with
https://github.com/rabbitmq/rules_erlang/pull/214 merged

In it's current form this will probably break local execution of mixed version tests. Further updates are needed so that with local execution, the old behavior is restored


pjk25:
It will need this to be available too https://github.com/rabbitmq/rules_erlang/pull/189

PR #8616, opened 2023-06-20
Osiris v1.6.0

kjnilsson:
This version contains support for the new chunk filtering feature (not implemented in this PR).

Also some improvements to replica init failures.


michaelklishin:
Osiris 1.6.0 incorrectly reports its name as 1.5.1.

#8209, #9053, #9054 are some of the necessary follow-ups.

PR #8645, opened 2023-06-21
CQ: Drop expired messages before fetching

lhoguin:
Otherwise we may send expired messages when a consumer connects in some cases.


lhoguin:
I will revisit this later.

PR #8653, opened 2023-06-22
Avoid secondary umbrella archive collisions in actions

pjk25:
`secondary-umbrellas/rbe-${otp_version_id}/package-generic-unix-for-mixed-version-testing-v${version }.tar.xz`

Makes the version of otp used for compiling the archive predictable in tests


pjk25:
This should be squash-merged when ready.

PR #8670, opened 2023-06-23
Keep CC header value in x-death header

acogoluegnes:


ansd (approved):
Fixes the following failing test:
https://github.com/rabbitmq/rabbitmq-java-client/blob/b65db946ee4694ef7f45c65d3bd9be1861a8a022/src/test/java/com/rabbitmq/client/test/functional/DeadLetterExchange.java#L374

Thank you!

ansd (approved):

michaelklishin:
This will not be backported because the issue was specific to `main`.

PR #8680, opened 2023-06-26
Use same AMQP 0.9.1 header for correlation ID

ansd:
For correlation IDs greater than 255 bytes, AMQP 1.0 to AMQP 0.9.1 conversion puts the correlation ID into an AMQP 0.9.1 header called `x-correlation-id`, see
https://github.com/rabbitmq/rabbitmq-server/blob/a17b28ec618cdf79a29fa520748dbe340561e197/deps/rabbit/src/rabbit_msg_record.erl#L380
Let's use the same key for the MQTT 5.0 to AMQP 0.9.1 conversion.


michaelklishin:
It's a good question what release this should go into. `3.13.0`?

PR #8694, opened 2023-06-29
Tests: more resilient time-dependent tests

dcorbacho:
Many test suites use fixed short sleeps to wait for events to happen. This can create flaky tests, when slower machines/configurations might fail test cases for not waiting long enough.

This PR aims to change most of these test cases, waiting for conditions to happen instead of a given fixed time. It also replaces ad-hoc waiting loops for the existing macro and wait functions from the `rabbitmq_ct_helpers` application.


michaelklishin:
The failure in `//deps/rabbit:per_user_connection_channel_limit_partitions_SUITE-mixed` is consistent.

dumbbell (approved):
Great work! Thank you!

PR #8743, opened 2023-07-03
CLI tools: introduce commands that enable and disable free disk space monitoring

### LINKED ISSUE #8741: CLI tools: make it possible to disable or re-enable free disk space monitoring

michaelklishin:
Closes #8741.

References #8740.

### DIFF
--- a/deps/rabbit/src/rabbit_disk_monitor.erl
+++ b/deps/rabbit/src/rabbit_disk_monitor.erl
@@ -168,7 +168,7 @@ handle_call({set_enabled, _Enabled = true}, _From, State = #state{enabled = true
 
 handle_call({set_enabled, _Enabled = true}, _From, State = #state{enabled = false}) ->
   _ = start_timer(set_disk_limits(State, State#state.limit)),
-  rabbit_log:info("Free disk space monitor was enabled"),
+  rabbit_log:info("Free disk space monitor was manually enabled"),
   {reply, ok, State#state{enabled = true}};
 
 handle_call({set_enabled, _Enabled = false}, _From, State) ->
--- a/deps/rabbit/src/rabbit_disk_monitor.erl
+++ b/deps/rabbit/src/rabbit_disk_monitor.erl
@@ -171,10 +171,15 @@ handle_call({set_enabled, _Enabled = true}, _From, State = #state{enabled = fals
   rabbit_log:info("Free disk space monitor was manually enabled"),
   {reply, ok, State#state{enabled = true}};
 
-handle_call({set_enabled, _Enabled = false}, _From, State) ->
+handle_call({set_enabled, _Enabled = false}, _From, State = #state{enabled = true}) ->
     _ = erlang:cancel_timer(State#state.timer),
     rabbit_log:info("Free disk space monitor was manually disabled"),
     {reply, ok, State#state{enabled = false}};
+
+handle_call({set_enabled, _Enabled = false}, _From, State = #state{enabled = false}) ->
+  _ = erlang:cancel_timer(State#state.timer),
+  rabbit_log:info("Free disk space monitor is already disabled"),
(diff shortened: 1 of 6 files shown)


deps/rabbitmq_cli/test/ctl/status_command_test.exs
```diff
@@ -1,4 +1,5 @@
 ## This Source Code Form is subject to the terms of the Mozilla Public
+status_command_test.exs
```
the-mikedavis, line 2: Looks like a small typo here

deps/rabbit/src/rabbit_disk_monitor.erl
```diff
@@ -161,11 +161,16 @@ handle_call({set_min_check_interval, MinInterval}, _From, State) ->
...
-    rabbit_log:info("Free disk space monitor was enabled"),
+    rabbit_log:info("Free disk space monitor is already enabled"),
     {reply, ok, State#state{enabled = true}};
 
+handle_call({set_enabled, _Enabled = true}, _From, State = #state{enabled = false}) ->
+  _ = start_timer(set_disk_limits(State, State#state.limit)),
+  rabbit_log:info("Free disk space monitor was manually enabled"),
+  {reply, ok, State#state{enabled = true}};
+
 handle_call({set_enabled, _Enabled = false}, _From, State) ->
```
the-mikedavis, line 174: What do you think about expanding these `false` cases like the above `true` cases so that the messages are slightly better?

```erlang
handle_call({set_enabled, _Enabled = false}, _From, State = #state{enabled = false}) ->
    %% ... says that the monitor was already disabled
handle_call({set_enabled, _Enabled = false}, _From, State = #state{enabled = true}) ->
    %% ... disables the monitor
```

michaelklishin: Good idea!

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/disable_free_disk_space_monitoring_command.ex
```diff
@@ -0,0 +1,61 @@
...
+
+  def run([], %{node: node_name, timeout: timeout}) do
+    :rabbit_misc.rpc_call(node_name, :rabbit_disk_monitor, :set_enabled, [false], timeout)
+  end
+
+  use RabbitMQ.CLI.DefaultOutput
+
+  def formatter(), do: RabbitMQ.CLI.Formatters.String
+
+  def usage, do: "status [--unit <unit>]"
```
the-mikedavis, line 46:
```suggestion
  def usage, do: "disable_free_disk_space_monitoring [--unit <unit>]"
```

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/enable_free_disk_space_monitoring_command.ex
```diff
@@ -0,0 +1,61 @@
...
+
+  def run([], %{node: node_name, timeout: timeout}) do
+    :rabbit_misc.rpc_call(node_name, :rabbit_disk_monitor, :set_enabled, [true], timeout)
+  end
+
+  use RabbitMQ.CLI.DefaultOutput
+
+  def formatter(), do: RabbitMQ.CLI.Formatters.String
+
+  def usage, do: "status [--unit <unit>]"
```
the-mikedavis, line 46:
```suggestion
  def usage, do: "enable_free_disk_space_monitoring [--unit <unit>]"
```

the-mikedavis (approved):

PR #8752, opened 2023-07-04
Fix parsing of MQTT PUBACK packet

ansd:
The MQTT v5 spec is a bit contradictory and imprecise when it comes to the Property Length.

It first mandates that if there are no properties, this MUST be indidated by including a Property Length of 0:
```
2.2.2.1 Property Length
The Property Length is encoded as a Variable Byte Integer. The Property Length
does not include the bytes used to encode itself, but includes the length of
the Properties. If there are no properties, this MUST be indicated by including
a Property Length of zero [MQTT-2.2.2-1].
```

All MQTT packets Property Length sections then only mention:
```
The length of the Properties in the PUBLISH packet Variable Header encoded as a Variable Byte Integer.
```
So, they follow above requirement.

However, MQTT packets PUBACK, PUBREC, PUBREL, PUBCOMP, and DISCONNECT seem to be exceptions to this rule:
```
3.4.2.2 PUBACK Properties
3.4.2.2.1 Property Length
The length of the Properties in the PUBACK packet Variable Header encoded as a Variable Byte Integer.
If the Remaining Length is less than 4 there is no Property Length and the value of 0 is used.
```

```
3.14.2.2 DISCONNECT Properties
3.14.2.2.1 Property Length
The length of Properties in the DISCONNECT packet Variable Header encoded as a Variable Byte Integer.
If the Remaining Length is less than 2, a value of 0 is used.
```

Since this special case has already been implemented for DISCONNECT, and RabbitMQ does not support QoS 2, this commit implements this special case for the PUBACK packet.

Some MQTT clients (e.g. mqttjs) indeed set a Remaining Length of 3 in the PUBACK packet.

See https://github.com/rabbitmq/rabbitmq-server/issues/2554#issuecomment-1620083550

### DIFF
--- a/deps/rabbitmq_mqtt/src/rabbit_mqtt_packet.erl
+++ b/deps/rabbitmq_mqtt/src/rabbit_mqtt_packet.erl
@@ -8,7 +8,6 @@
 -module(rabbit_mqtt_packet).
 
 -include("rabbit_mqtt_packet.hrl").
--include("rabbit_mqtt.hrl").
 -include_lib("kernel/include/logger.hrl").
 
 -export([init_state/0, parse/2, serialise/2]).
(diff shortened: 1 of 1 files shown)


ansd:
@alaendle would you mind testing this fix with your mqtt.js client? That would be awesome!
You could use docker image `pivotalrabbitmq/rabbitmq:mqtt5-puback-otp-max-bazel` (which uses the branch of this PR) to start RabbitMQ server.

alaendle:
@ansd Awesome! Works now as expected!

Issue #8784, opened 2023-07-05
High connection or CQv1 queue churn can cause file_handle_cache to terminate, making the node unavailable


michaelklishin:
Originally discussed and narrowed down based on log evidence in #8776.

Consider the following scenario in a cluster with very short lived connections and thus very high [connection churn](https://rabbitmq.com/connections.html#high-connection-churn):

 * A client connection is opened, and an Erlang process (a "green thread" if you will) is started for it to handle it. This connection has an identifier such as <0.1139.0>
 * Connection is closed a few milliseconds later
`file_handle_cache` handles the connection closure event and clears a row with some metrics

So far so good. Now, concurrently with that, another connection is open and

 * It gets the same local Erlang process identifier, say, <0.1139.0> (the same value as the recently terminated connection)
 * It starts performing operations that result in metrics being updated
 * But the above connection closure event handler will concurrently delete the metric row table
 * Now all metric updates (writes) operation fail due to the missing key
 * And with that, all client operations on a connection

In other words, for this scenario to happen you need one very short lived connection and another very short-lived connection to get "assigned" the same Erlang process ("green thread") ID, then two independent metric table updates can step over one another.

`file_handle_cache` and `file_handle_cache_stats` should be more defensive when updating metrics.

Besides avoiding connection churn, what reduces the probability of running into this
issue is the use of [classic queues v2](https://blog.rabbitmq.com/posts/2023/05/rabbitmq-3.12-performance-improvements/) that do not rely on the file handle cache where CQv1 did.

lhoguin:
The problem seems related to the many processes MQTT was creating before 3.12.

michaelklishin:
#8790 addresses several scenarios to the best of our current understanding.

In addition, in #8776 there is a report of the issue going away when non-mirrored CQs are used for MQTT and Web MQTT connections that are very short lived. This makes sense because QQs are expensive to set up. QQs must not be used for consumers that churn through connections and their queues (you would not get any of the QQ benefits with that kind of workload anyway).

PR #8790, opened 2023-07-06
FHC: use exception-safe versions of ets update functions

michaelklishin:
Otherwise, if the FHC runs into an exception because of an ETS write failure, most code paths begin failing.

This may help with the condition in #8784.

### DIFF
--- a/deps/rabbit_common/src/file_handle_cache.erl
+++ b/deps/rabbit_common/src/file_handle_cache.erl
@@ -1158,7 +1158,7 @@ handle_call({open, Pid, Requested, EldestUnusedSince}, From,
     case needs_reduce(State #fhc_state { open_count = Count + Requested }) of
         true  -> case ets:lookup(Clients, Pid) of
                      [#cstate { opened = 0 }] ->
-                       safe_ets_update_element(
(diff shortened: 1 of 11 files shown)


deps/rabbit_common/src/file_handle_cache.erl
```diff
@@ -1125,13 +1158,13 @@ handle_call({open, Pid, Requested, EldestUnusedSince}, From,
     case needs_reduce(State #fhc_state { open_count = Count + Requested }) of
         true  -> case ets:lookup(Clients, Pid) of
                      [#cstate { opened = 0 }] ->
-                         true = ets:update_element(
+                       safe_ets_update_element(
```
lhoguin, line 1161: Probably not important but based on the other changes I would expect `_ = safe_ets...` here.

lhoguin (approved):
This works as a temporary measure, approved.

michaelklishin:
This addresses several scenarios to the best of our current understanding without introducing any obvious regressions. Let merge it and allow the user to try out a `3.12.2` alpha.

michaelklishin:
Note that the OCI build failures are not related.

PR #8837, opened 2023-07-12
rabbit_disk_monitor: handle gen_server timeouts to reduce log noise

michaelklishin:
References #8836.


lukebakken (approved):
I tested by setting the `gen_server:call` timeout to `0`. No logging was observed because the `timeout` exception was caught.

PR #8881, opened 2023-07-17
Update Erlang.mk + small cleanup

lhoguin:
Let's wait for tests to go green before merge.


lhoguin:
Not super important to backport since the update is there mostly to bring in OTP-26 fixes.

pjk25:
Backporting so that our build comparison CI with match on the other branches

PR #8930, opened 2023-07-20
rabbit_db: Reset feature flags registry after a database reset

dumbbell:
### Why

A database reset removes the enabled feature flags file on disc. A reset of the registry ensures that the next time the registry is reloaded, it is also initialized from scratch.

### How

We call `rabbit_feature_flags:reset_registry/0` after both a regular reset and a forced reset.

The `reset_registry/0` is also exposed by the `rabbit_feature_flags` module now. The actual implementation in `rabbit_ff_registry_factory` should only be called by the Feature flags subsystem itself.

### DIFF
--- a/release-notes/3.13.0.md
+++ b/release-notes/3.13.0.md
@@ -8,7 +8,7 @@ This release includes several new features, optimizations, and graduates (makes 
 
 The user-facing areas that have seen the biggest improvements in this release are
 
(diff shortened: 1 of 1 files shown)


dumbbell:
Crap, as part of debugging an issue with bazel unrelated to the patch, GitHub closed this pull request... Reopening it as #8935.

PR #8950, opened 2023-07-25
Use rpcs instead of ctl commands to avoid CI failures

dcorbacho:
A few CI runs failed with `Failed to create dirty cpu scheduler thread ...` when running rabbitmqctl commands in the parallel suite. To avoid more complex configuration changes, this suite now uses rpcs to query queue metrics.


michaelklishin:
@dcorbacho should we backport this at least to 3.12.x?

michaelklishin:
@mergifyio backport v3.12.x v3.11.x

PR #8984, opened 2023-07-31
Only copy local listeners when FF listener_records_in_ets is enabled

gomoripeti:
## Proposed Changes

The `rabbit_listener_ets` table normally only contains listener entries for the local node. However after upgrading to 3.11.x and the `listener_records_in_ets` feature flag is enabled, all entries were copied to ETS from Mnesia, and the replicated Mnesia table contained entries for all nodes in a multi-node cluster.

In this state `rabbit_networking:listener_of_protocol_ets/1` could crash with a case clause when multiple rows match a protocol. This can happen for example when the node is put in maintenance and the web_stomp plugin is enabled which queries the ranch refs to close all client connections. (`rabbit_web_stomp_listener:close_all_client_connections/1`)

A simple restart of the node or app reinitializes the listeners and the ETS table so this "corrupt" state is cleared.

Only affects 3.11 series so opening the PR against that branch.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_networking.erl
+++ b/deps/rabbit/src/rabbit_networking.erl
@@ -45,6 +45,8 @@
 
 -export([ensure_listener_table_for_this_node/0]).
 
+-export([is_listener_on_this_node/1]).
+
 -deprecated([{force_connection_event_refresh, 1, eventually}]).
 
 -export([
@@ -819,3 +821,7 @@ ipv6_status(TestPort) ->
 ensure_listener_table_for_this_node() ->
     _ = ets:new(?ETS_TABLE, [named_table, public, bag, {keypos, #listener.node}]),
     ok.
+
+%% Needed for migration when feature flag listener_records_in_ets is enabled
+is_listener_on_this_node(#listener{node = Node}) ->
+    Node =:= node().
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_networking.erl
```diff
@@ -819,3 +821,7 @@ ipv6_status(TestPort) ->
 ensure_listener_table_for_this_node() ->
     _ = ets:new(?ETS_TABLE, [named_table, public, bag, {keypos, #listener.node}]),
     ok.
+
+%% Needed for migration when feature flag listener_records_in_ets is enabled
+is_listener_for_this_node(#listener{node = Node}) ->
```
michaelklishin, line 826: How about `local_listener/1` or `listener_on_this_node/1`?

gomoripeti: took the name from the function above `ensure_listener_table_for_this_node()` but is_local_listener sounds much better.

michaelklishin:
The failures seem unrelated. They had popped up after a Java dependency bump, so, something unrelated to any Erlang code.

We refactored these two suites in main and backported those changes to v3.12.x to avoid flakey parallel tests, so perhaps should do the same thing for v3.11.x.

gomoripeti:
thank you!

PR #9006, opened 2023-08-03
rabbit_stream_queue_SUITE: Fix `consume_and_reject` channel close detection

dumbbell:
### Why

The previous detection was based on a reuse of the channel to get the error from an exit exception. The problem is that it is very dependent on the timing: if the channel process exits before it is reused, the test fails for two possible reasons:

1. The channel and connection processes exit before they are reused and the channel manager opens a new pair. The problem is that the declare suceeds but the test expected a failure.

2. The channel and connection processes exit during the reuse and `rabbit_ct_client_helpers:open_channel` in `retry_if_coordinator_unavailable()` waits a response from the channel manager forever (this is probably a weakness of the channel manager in rabbitmq_ct_client_helpers). This indefinite wait causes the testcase to timeout.

### How

A simpler solution is to monitor the exit reason of the channel process that triggers the error on the server side.


dumbbell:
Reviewed by @michaelklishin.

PR #9083, opened 2023-08-15
Use osiris from Bazel Central Repository

pjk25:
osiris 1.6.3 is identical to 1.6.2, but we needed a new version for the sake of the BCR publish

this makes osiris a native bzlmod dependency


michaelklishin:
Hooray!

PR #9149, opened 2023-08-22
Fix issue 9059

MarcialRosales:
This issue, #9059, occurs when the ldap plugin submits a query to retrieve an LDAP attribute and that attribute has a non-ascii character such as ß. When the plugin receives a binary representation of a string with a non-ascii character, it treated as a list of strings when it was a list of characters. It tried to join those characters with `string:join` function. When this function encounter the non-ascii characters, it failed to concatenate it and threw the exception `badarg`.

## Proposed Changes

Use list comprehensions instead of string:join to produce a string representation of the attribute's value which can be a single string -with only ascii characters or mix of ascii and non-ascii- or multiple strings. 

## Types of Changes

- [ ] Build system and/or CI


michaelklishin:
The fix was confirmed in #9059 by the reporter.

PR #9220, opened 2023-08-30
osiris 1.6.4

kjnilsson:
This includes a fix for a race condition when the commit index known by a stream replica is higher than what is fully written locally.


pjk25:
Waiting on https://github.com/bazelbuild/bazel-central-registry/pull/873

pjk25 (approved):
I've added the override that allows us to merge this without waiting for BCR. We can remove the override later.

PR #9241, opened 2023-08-31
Use ra_leaderboard when initialising quorum queue client.

kjnilsson:
If available, prefer the leader in ra_leaderboard over the one in the queue record as with mnesia the queue record can be very stale if a node has a network issue that hasn't yet been detected by the runtime.

Also change some legacy naming in rabbit_fifo_client

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -161,11 +161,17 @@ is_compatible(_, _, _) ->
 -spec init(amqqueue:amqqueue()) -> {ok, rabbit_fifo_client:state()}.
 init(Q) when ?is_amqqueue(Q) ->
     {ok, SoftLimit} = application:get_env(rabbit, quorum_commands_soft_limit),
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -161,11 +161,17 @@ is_compatible(_, _, _) ->
...
+    Leader = case ra_leaderboard:lookup_leader(Name) of
+                 undefined ->
+                     %% leader from queue record will have to suffice
+                     MaybeLeader;
+                 LikelyLeader ->
+                     LikelyLeader
+             end,
     Nodes = get_nodes(Q),
-    %% Ensure the leader is listed first
+    %% Ensure the leader is listed first to increase likelyhood of first
```
the-mikedavis, line 173:
```suggestion
    %% Ensure the leader is listed first to increase likelihood of first
```

kjnilsson: done\1

the-mikedavis (approved):

PR #9290, opened 2023-09-04
rabbit_ct_helpers: Detect crashes in RabbitMQ logs

dumbbell:
### Why

Some crashes may go under the radar if a testcase didn't check everything (which is hard to do anyway). It could be useful to be aware of those crashes.

### How

This new check will search for any logged crashes in the RabbitMQ log files and throw an exception as part of stopping a node.


deps/rabbitmq_ct_helpers/src/rabbit_ct_broker_helpers.erl
```diff
@@ -1008,6 +1008,31 @@ stop_rabbitmq_nodes(Config) ->
...
+    %% they are stopped. If we find some, we log them again in the common_test
+    %% logs and throw an exception to make the test fail.
+    FindCrashes = case rabbit_ct_helpers:get_config(Config, find_crashes) of
+                      true ->
+                          true;
+                      false ->
+                          false;
+                      undefined ->
+                          case os:getenv("FIND_CRASHES") of
+                              undefined -> true;
```
ansd, line 1021: `undefined` can never be returned causing this new crash detection facility to be disabled by default.

dumbbell: Good catch. The overall patch was incorrectly ported from the Khepri branch anyway...

I opened #9373 to fix this problem and another one.

PR #9319, opened 2023-09-06
Ra 2.7.0 + rabbitmq-queues quorum_status improvements

kjnilsson:
This Ra version includes a new `ra:key_metrics/1` API that is more available
than parsing the output of `sys:get_status/1` especially when a quorum queue member is recovering after a restart.

the rabbit_quorum_queue:status/1 function has been ported to use
this API instead as well as now includes a few new fields.


kjnilsson:
This shows how the `rabbitmq-queues quorum_status` command still returns promptly even though the member on rabbit-3 is in log recovery state and previously would have timed out and not shown any progress. The Last Applied column can be used to see the current recovery progress.

image

michaelklishin:
Rebased to hopefully address the root cause of the OCI build failure (#9568).

PR #9334, opened 2023-09-08
Revert "Settle unroutable message with released state (backport #8015) (backport #8736)"

michaelklishin:
Reverts rabbitmq/rabbitmq-server#8739.


michaelklishin:
We ended up agreeing that #8015 is the right thing to do and deciding to keep this behavior in v3.11.x as well.

So this PR was later reverted.

PR #9337, opened 2023-09-08
Try to avoid running out of space in the OCI workflow in actions

pjk25:
Sometimes the actions worker seems to run out of space


michaelklishin:
Our OCI builds run out of disk space so often today that I'm willing to try this just to compare :)

PR #9356, opened 2023-09-08
Improve consumer metric cleanup when a channel goes down

SimonUnge:
## Proposed Changes

This update improves metric cleanup of consumer data when a channel goes down by:
* No longer emitting consumer_deleted for each consumer of a channel
* bulk/pattern deletion of metric data in ets tables.

## Types of Changes

- [x] New feature (non-breaking change which adds functionality) (not really)
- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue_process.erl
+++ b/deps/rabbit/src/rabbit_amqqueue_process.erl
@@ -913,7 +913,7 @@ handle_ch_down(DownPid, State = #q{consumers                 = Consumers,
             {ok, State1};
         {ChAckTags, ChCTags, Consumers1} ->
             QName = qname(State1),
-            [emit_consumer_deleted(DownPid, CTag, QName, ?INTERNAL_USER) || CTag <- ChCTags],
+            [rabbit_core_metrics:consumer_deleted(DownPid, CTag, QName) || CTag <- ChCTags],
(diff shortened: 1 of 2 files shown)


michaelklishin:
With this PR, when I kill an app process that has 10K consumers on one channel, I observe a short spike of CPU usage from ≈ 0.3% to ≈ 5% which then instantly goes back to ≈ 0.3%.

michaelklishin:
Ah, the original test was with over 100K consumers on a channel.

michaelklishin:
This improves sample cleanup for a few channels with an extremely large number of consumers but regresses when there's a large number of channels with one consumer, which is more common.

Passing consumer info as context when a channel is closed would help avoid the scan introduced in this PR entirely.

michaelklishin:
Or, alternatively, the two approaches can be combined:

 * Do a scan when the number of consumers is > 1000 (or another constant)
 * Include consumer details when there are just a few

SimonUnge:
To clarify, its `ets:match_delete` that is the culprit then, and should be avoided if there are few consumers in the channel. Do we see a need to also revert the removal of `emit_consumer_delete` in rabbit_amqqueue_process, i.e add back the event notification?

PR #9376, opened 2023-09-12
Expose netsplit/partition info via Prometheus

### LINKED ISSUE #2508: Prometheus: expose a metric for network partition observed on the node
Hi,
I'm happy that RabbitMQ got its own Prometheus exporter. :+1: also for prepared Grafana dashboard, it looks great.
I was using [deadtrickster prometheus rabbitmq exporter](https://github.com/deadtrickster/prometheus_rabbitmq_exporter) before. It allowed me to monitor network partition with its `rabbitmq_node_up` metric https://github.com/deadtrickster/prometheus_rabbitmq_exporter#nodes because it shows whether one node is connected to another. I have three nodes in my cluster for now and thus I can write an alert
```
sum(rabbitmq_node_up{app="rabbitmq"}) != 9
```
To check that each RabbitMQ node is connected to the other two. When it is not I'm alerted. Would it be possible to add something similar to this plugin?

gomoripeti:
## Proposed Changes

Mnesia reports which nodes see another as being partitioned from during a network partiotion. This info is exposed today on the Management UI and via cli. This PR adds the same info to the Prometheus main endpoint in the format:
`rabbitmq_partitioned_from{peer="rabbit@<host>"} 1` for each element in the partitions list

Fixes #2508 

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments


deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_dynamic_collector.erl
```diff
@@ -0,0 +1,49 @@
...
+          ?METRICS
+         ),
+    ok.
+
+%%====================================================================
+%% Private Parts
+%%====================================================================
+
+values(partitioned_from) ->
+    %% Note: partition info comes from Mnesia under the hood
```
michaelklishin, line 47: It's not going to be just Mnesia very soon, so this should perhaps say "from the node monitor implementation"

gomoripeti: U just added that note so it is easy to grep for Mnesia related thins, and I did not know if this metric makes sense in a post-Mnesia world.
Will khepri have a similar notion (maybe info from aten)?


removing that comment then

michaelklishin (changes requested):
@gomoripeti when a new module is added or functions are moved between modules, please run

``` shell
bazel run gazelle
```

so that it regenerates the graph of module dependencies used by Bazel.

michaelklishin:
Also worth mentioning that for Khepri (optional in 3.13, the only game in town in 4.0), there won't be an equivalent source of information.

michaelklishin:
After an internal discussion we've decided to not accept this PR.

This metric won't be available (not in the same way anyway) for Khepri, and Khepri is a few months away from being the only option available (as of 4.0).

It does not make much sense to introduce this feature only to remove it in 4-6 months.

Node connectivity monitoring is a fundamental problem for which we have some tools already
(e.g. the Erlang link Grafana dashboard which is much more thorough than a single
value metric), and others may have to be developed for Khepri specifically.

Thank you for taking the time to contribute to RabbitMQ.

gomoripeti:
thanks @michaelklishin for the infos

>  it regenerates the graph of module dependencies used by Bazel.

Is this graph committed to the repo, or is it enough to just update app.bzl manually? (I don't have bazel locally)

> Also worth mentioning that for Khepri (optional in 3.13, the only game in town in 4.0), there won't be an equivalent source of information.

Would there be anything at least slightly similar? Or `rabbit_node_monitor:partitions/0` will always return an empty list. In that case this metric implementation can be entirely removed on 4.0

gomoripeti:
Did not seem your last comment, thanks anyway for discussing

michaelklishin:
@gomoripeti as an alternative suggested by @mkuratczyk, we can expose something like

``` erl
length(rabbit_nodes:list_reachable()).
```

as a metric. I don't know if it would be a good enough alternative for you but this is something that's not Mnesia or Khepri-specific, can be easily monitored (by comparing it to the expected number of online cluster nodes), and is very easy to compute and reason about.

gomoripeti:
- maybe `length(rabbit_nodes:list_unreachable()).`? then one does not have to know the expected number of nodes, just compare to zero?
- do you think this has added information compared to the Established Distribution Links that I see on the Erlang Distribution Dashboard? Maybe your suggestion is a subset of DB cluster members (so it has application level info) while the later is just any Erlang links including those towards remote_shells?

One of my motivations was to have a substitute for the Mgmt UI info after all stats are gone in 4.0. But as you clarified this Mnesia specific status won't make sense in 4.0.
The other thing I liked about the original proposal is that it provides an application level view which might be different from the Erlang distribution level status. But if there will be no such difference with Khepri then no need to add an extra metric.

michaelklishin:
@gomoripeti yes, using a list of unreachable nodes sounds even better.

The primary benefit of having a basic metric like this is that you don't have to worry about the nuances. Erlang distribution metrics provide a lot of nuance but not everyone will need it.

frittentheke:
> @gomoripeti yes, using a list of unreachable nodes sounds even better.
> 
> The primary benefit of having a basic metric like this is that you don't have to worry about the nuances. Erlang distribution metrics provide a lot of nuance but not everyone will need it.

Yes, that. A basic set of "warning light" style metrics indicating there is an issue would be awesome.
All the very detailed and implementation specifics are unimportant if this requires the person / operator / admin setting up the monitoring to know all of them to write alerts.

If I may compare this with Ceph Software defined storage exposing cluster health the Prometheus way:

`ceph_health_status` - Just a value indicating if the cluster is in OK, WARN or ERR condition.
 If you monitor nothing else, this will tell you to call someone to look at your storage cluster.

Then there is `ceph_health_detail` with labels for each health issue going into the `ceph_health_status`

```
[...]
ceph_health_detail{instance="az1", job="ceph", name="MON_DOWN", severity="HEALTH_WARN"} 0
ceph_health_detail{instance="az1", job="ceph", name="OSD_DOWN", severity="HEALTH_WARN"} 0
ceph_health_detail{instance="az1", job="ceph", name="OSD_HOST_DOWN", severity="HEALTH_WARN"} 0
ceph_health_detail{instance="az1", job="ceph", name="OSD_NEARFULL", severity="HEALTH_WARN"} 0
ceph_health_detail{instance="az1", job="ceph", name="OSD_SLOW_PING_TIME_BACK", severity="HEALTH_WARN"} 0
ceph_health_detail{instance="az1", job="ceph", name="OSD_SLOW_PING_TIME_FRONT", severity="HEALTH_WARN"} 0
ceph_health_detail{instance="az1", job="ceph", name="PG_AVAILABILITY", severity="HEALTH_WARN"} 0
ceph_health_detail{instance="az1", job="ceph", name="PG_DAMAGED", severity="HEALTH_ERR"} 0
ceph_health_detail{instance="az1", job="ceph", name="RECENT_CRASH", severity="HEALTH_WARN"} 0
ceph_health_detail{instance="az1", job="ceph", name="SLOW_OPS", severity="HEALTH_WARN"} 0
[...]
```
and which will return "1" is this particular condition is present. This might allow the monitoring to figure out if you just need to replace a drive, some servers require fixing or if there is a really big issue with the cluster.

Some goes for RabbitMQ ... is the cluster OK, in some degraded but working condition or is the service affected / unavailable.

I'd argue that such an abstraction of the RabbitMQ cluster health can and should be made no matter what happens to a particular implementation detail. Certainly, this does not take away the complexity of operational issues requiring all of the metrics and knowledgeable people to fix. But at least these abstract indication allow monitoring systems to alert on such condition without countless alerts on countless metrics and their correlation.

michaelklishin:
@frittentheke thanks for the example.

@gomoripeti you are welcome to submit a PR that would use

``` erl
length(rabbit_nodes:list_unreachable()).
```

and something like `*_unreachable_cluster_peers` (a gauge) for the metric name. We will discuss this on the core team, I personally don't expect much opposition since one of us had earlier suggested a similar counter above :) Thanks!

gomoripeti:
Thank you, I'm on it :)

Michael Klishin ***@***.***> ezt írta (időpont: 2023. szept.
13., Sze 20:49):

PR #9405, opened 2023-09-12
Forward compatibility with Erlang 27: distinguish between positive and negative 0.0

mkuratczyk:
Without this change, our tests against OTP master started failing since the negative zero change was introduced, for example: https://github.com/rabbitmq/rabbitmq-server/actions/runs/6141080013/job/16660857916

The warning (treated as error) is:
```
rabbit_numerical.erl:33:8: matching on the float 0.0 will no longer also match -0.0 in OTP 27. If you specifically intend to match 0.0 alone, write +0.0 instead.
```

As far as I can tell, we don't need special handling for this case in the first place - `digits(Float)` returns `"0.0"` for both `0.0` and `-0.0`.

### DIFF
--- a/deps/rabbit_common/src/rabbit_numerical.erl
+++ b/deps/rabbit_common/src/rabbit_numerical.erl
@@ -30,8 +30,6 @@
 %%       human-readable output, or compact ASCII serializations for floats.
 digits(N) when is_integer(N) ->
     integer_to_list(N);
-digits(0.0) ->
-    "0.0";
 digits(Float) ->
     {Frac1, Exp1} = frexp_int(Float),
     [Place0 | Digits0] = digits1(Float, Exp1, Frac1),
--- a/deps/rabbit_common/src/rabbit_numerical.erl
+++ b/deps/rabbit_common/src/rabbit_numerical.erl
@@ -30,7 +30,7 @@
 %%       human-readable output, or compact ASCII serializations for floats.
 digits(N) when is_integer(N) ->
     integer_to_list(N);
-digits(0.0) ->
+digits(+0.0) ->
     "0.0";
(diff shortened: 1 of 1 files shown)


michaelklishin:
Great find.

michaelklishin:
The OCI failures are due to something (Actions or BuildBuddy worker environments) running out of disk space again.

michaelklishin:
This genuinely affects

```
bazel test //deps/rabbit_common:eunit
```

I'm looking into it.

michaelklishin:
It is `math:log10(0.0).` that fails with this change, which is true, log10 of 0 is undefined.

michaelklishin:
```
rabbit_numerical: digits_test...*failed*
in function math:log10/1
  called as log10(0.0)
in call from rabbit_numerical:scale/7 (rabbit_numerical.erl, line 143)
in call from rabbit_numerical:digits/1 (rabbit_numerical.erl, line 35)
in call from rabbit_numerical:digits_test/0 (rabbit_numerical.erl, line 273)
in call from eunit_test:'-mf_wrapper/2-fun-0-'/2 (eunit_test.erl, line 273)
in call from eunit_test:run_testfun/1 (eunit_test.erl, line 71)
in call from eunit_proc:run_test/1 (eunit_proc.erl, line 531)
in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 356)
**error:badarith
  output:<<"">>
```

michaelklishin:
@mkuratczyk let's try matching both instead

michaelklishin:
Now we get

```
rabbit_numerical.erl:35:1: this clause cannot match because a previous clause at line 33 always matches
```

on Erlang 26 and 25 :)

PR #9411, opened 2023-09-13
Request more disk space for docker image jobs on RBE

pjk25:
On some occasions the OCI jobs fail on github actions due to lack of space


pjk25:
We could probably revert #9337 if this works

michaelklishin:
@pjk25 we can also keep both since our repo is of a non-trivial size, as are all Bazel build artifacts

PR #9414, opened 2023-09-13
Forward compatibility with Erlang 27: negative zero fix, take 2

mkuratczyk:
https://github.com/rabbitmq/rabbitmq-server/pull/9405 didn't actually solve the problem - OTP27 (git master) still complained about matching on the zero float. This does seem to make it happy.


michaelklishin:
The force-push above is a rebase to address the OCI build failures.

PR #9417, opened 2023-09-13
Rename quorum_queue_utils to queue_utils

dcorbacho:
Not quorum queue specific any longer.
It's used in many test suites to query any queue type.


dcorbacho:
@mergifyio backport v3.12.x

PR #9425, opened 2023-09-13
WIP Streams: advertise localhost to local clients

### LINKED ISSUE #9424: Streams: consider advertising localhost as connection hostname for clients that originally connect via localhost
This is one take-away from [this Spring-related discussion](https://github.com/spring-projects/spring-amqp/issues/2522).

On some hosts, a stream client that connects to `localhost` with default credentials, it will get back a a resolved local hostname (**not** `localhost` or `[::1]`). The client will then try to connect to that hostname using the same default credentials, which [will be rejected as a remote connection with default credentials](https://rabbitmq.com/access-control.html#loopback-users).

Instead, the server (or the client itself, this is a TBD) can notice that the connection target
was `localhost` and advertise `localhost` regardless of the node configuration.

Per discussion with @acogoluegnes @Zerpet.

michaelklishin:
Note that there is a way to opt out just in case.

Closes #9424.

### DIFF
--- a/deps/rabbitmq_stream/BUILD.bazel
+++ b/deps/rabbitmq_stream/BUILD.bazel
@@ -36,7 +36,8 @@ APP_ENV = """[
 	{frame_max, 1048576},
 	{heartbeat, 60},
 	{advertised_host, undefined},
-	{advertised_port, undefined}
+	{advertised_port, undefined},
+	{advertise_localhost_to_localhost_clients, true}
 ]"""
 
 all_beam_files(name = "all_beam_files")
--- a/deps/rabbitmq_stream/Makefile
+++ b/deps/rabbitmq_stream/Makefile
@@ -16,7 +16,8 @@ define PROJECT_ENV
 	{frame_max, 1048576},
 	{heartbeat, 60},
 	{advertised_host, undefined},
-	{advertised_port, undefined}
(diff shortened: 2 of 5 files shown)


acogoluegnes:
This is not working as expected, as the changes apply only to the hostname and port information carried out by the client connection, not where the information is actually advertised ([which is in the `metadata` command](https://github.com/rabbitmq/rabbitmq-server/blob/3f16a1c8822af5ca7fdd5adbd1c78b1b75c17338/deps/rabbitmq_stream/src/rabbit_stream_reader.erl#L2512-L2529)).

We would need to make the same check and skip the remote calls to always return the information from the local node. This assumes there's only one node in the cluster.

With the current changes, the broker returns the hostname, e.g. `foobar`, the client connects to `foobar`, the connection information contains `localhost`, this is not what the client expects, so it closes the connection, and retries. Infinite loop.

michaelklishin:
Instead of updating this PR to make the same change in two places we decided to try a client library solution. The downside of that is that we have to do it N times but the consensus on the team seems to be tilted that way.

michaelklishin:
https://github.com/rabbitmq/rabbitmq-stream-java-client/pull/420 is the first client library change that accomplishes the same goal.

PR #9509, opened 2023-09-21
Add amqp_client to the rabbitmq_web_mqtt deps in Make

pjk25:
as it is there in bazel


michaelklishin:
@mergifyio backport v3.12.x

PR #9523, opened 2023-09-22
Do not confirm MQTT messages if in partition

ansd:
Since MQTT publishers might publish to classic mirrored queues, add the same check as in `rabbit_channel:send_confirms_and_nacks/1`:

> If we are in a minority and pause_minority mode then a) we are
> going to shut down imminently and b) we should not confirm anything
> until then, since anything we confirm is likely to be lost.

I couldn't observe any performance degradation with this additional check using the following benchmark and the default `cluster_partition_handling` value `ignore`:
```
./mqtt-bm-latency -clients 50 -count 20000 -pubqos 1 -subqos 0 -size 12 -keepalive 120 -topic t
```


michaelklishin:
`pause_partition_guard/0` relies on a persistent term so yes, should be very cheap to call.

Issue #8758, opened 2023-07-04
HTTP API: DELETE /api/queues/{vhost}/{name} should use an internal API call


michaelklishin:
Currently it uses a direct Erlang client operation, which would fail if the queue in question is exclusive.

On the contrary, `rabbitmqctl delete_queue` can delete exclusive queues just fine because it does
not use a client to do so.

There is no reason why the HTTP API should not use `rabbit_amqqueue:delete/4` directly.

michaelklishin:
This was reverted in https://github.com/rabbitmq/rabbitmq-server/pull/10185

PR #9550, opened 2023-09-27
HTTP API: DELETE /api/queues/{vhost}/{name} use internal API call

dcorbacho:
A direct client operation fails if the queue is exclusive. This API should behave like the rabbitmqctl that can delete the queue even in that case

## Types of Changes
- [x] Bug fix (non-breaking change which fixes issue #8758 )
- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbitmq_management/src/rabbit_mgmt_wm_queue.erl
+++ b/deps/rabbitmq_management/src/rabbit_mgmt_wm_queue.erl
@@ -71,18 +71,31 @@ accept_content(ReqData, Context) ->
             rabbit_mgmt_util:bad_request(iolist_to_binary(io_lib:format(F ++ "~n", A)), ReqData, Context)
     end.
 
-delete_resource(ReqData, Context) ->
+delete_resource(ReqData, Context = #context{user = #user{username = ActingUser}}) ->
     %% We need to retrieve manually if-unused and if-empty, as the HTTP API uses '-'
     %% while the record uses '_'
     IfUnused = <<"true">> =:= rabbit_mgmt_util:qs_val(<<"if-unused">>, ReqData),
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1603,7 +1603,12 @@ delete_immediately_by_resource(Resources) ->
             rabbit_types:error('not_empty') |
             {protocol_error, Type :: atom(), Reason :: string(), Args :: term()}.
 delete(Q, IfUnused, IfEmpty, ActingUser) ->
-    rabbit_queue_type:delete(Q, IfUnused, IfEmpty, ActingUser).
+    rabbit_log:warning("TRACE delete queue ~p~n before ~p~n",
```
michaelklishin, line 1606: Looks like this should be turned into a debug statement or dropped entirely

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1603,7 +1603,12 @@ delete_immediately_by_resource(Resources) ->
...
+                       [amqqueue:get_name(Q), list()]),
+    Ret = rabbit_queue_type:delete(Q, IfUnused, IfEmpty, ActingUser),
+    rabbit_log:warning("TRACE delete queue ~p~n after ~p~n",
```
michaelklishin, line 1609: Same as above

michaelklishin (changes requested):

deps/rabbit/src/rabbit_amqqueue.erl
dcorbacho: Thanks, it completely slip through. It's gone now

michaelklishin (approved):

michaelklishin:
FYI, this PR is about to be reverted.

Replacing direct client's authorization checks ended up being not worth the gain in this PR.

Issue #9437, opened 2023-09-14
HTTP API: reduce the number of metrics served by GET /api/queues by default


michaelklishin:
`GET /api/queues` is widely used and abused without pagination and often to retrieve just a single metric. In environments with 10s or 100s of thousands of queues that produces enormously large JSON responses, burns a lot of CPU and maxes out network links for no good reason.

This is true for 3rd party tools such as monitoring ones.

In order to reduce this unnecessary resource waste we can a few things:

 * Drop some metrics, like almost all `backing_queue_status` fields, the `garbage_collection` field, from `GET /api/queues` responses
 * Potentially introduce a `/detailed` version of the endpoint that would return more metrics, or leave them to in the `rabbitmq-diagnostics observer`

We can also introduce pagination by default but that's likely going to be very confusing at first and will affect a lot of tools. Reducing the size of the response will be a lot less disruptive,
most tools and in most environments the vast majority of metrics are ignored.

dcorbacho:
@michaelklishin I wonder if in many cases it can be enough to use `GET /api/queues?disable_stats=true&enable_queue_totals=true`

```
[{
   "arguments":{"x-queue-type":"classic"},
   "auto_delete":false,
   "durable":true,
   "exclusive":false,
   "messages":8,
   "messages_ready":8,
   "messages_unacknowledged":0,
   "name":"q1",
   "node":"rabbit@dparracorb0633Q",
   "state":"running",
   "type":"classic",
   "vhost":"/"
}]
```

which already reduces a lot the datapoints returned vs `GET /api/queues`:

```
[{
   "arguments":{"x-queue-type":"classic"},
   "auto_delete":false,
   "backing_queue_status":{"version":2,"mode":"default","len":8,"qi_buffer_num_up":0,"qi_buffer_size":0,"qs_buffer_size":0,"q3":1,"delta":["delta",1,7,7,8],"num_pending_acks":0,"num_unconfirmed":0,"next_deliver_seq_id":0,"next_seq_id":8,"avg_ack_egress_rate":0.0,"avg_ack_ingress_rate":0.0,"avg_egress_rate":0.0,"avg_ingress_rate":0.670065251225457,"target_ram_count":"infinity","q4":0,"q2":0,"q1":0},
   "consumer_capacity":0,
   "consumer_utilisation":0,
   "consumers":0,
   "durable":true,
   "effective_policy_definition":{},
   "exclusive":false,
   "exclusive_consumer_tag":null,
[...]

PR #9578, opened 2023-09-28
Core metrics: remove backing_queue_status from the core metrics

dcorbacho:
They will not show in the HTTP API `GET /api/queues`, but are still available whenever the queue info is requested through the rabbit_amqqueue API.   

GET /api/queues is widely used and abused without pagination and often to retrieve just a single metric. This change will considerable reduce the size of the JSON response when querying 10s or 100s of queues, and it should not impact the operation of the system as `backing_queue_status` is an internal metric on which no monitoring tool should rely on.

References https://github.com/rabbitmq/rabbitmq-server/issues/9437

## Types of Changes
- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue_process.erl
+++ b/deps/rabbit/src/rabbit_amqqueue_process.erl
@@ -1205,7 +1205,7 @@ emit_stats(State, Extra) ->
      {messages, M},
      {reductions, R},
      {name, Name} | Infos]
-    = [{K, V} || {K, V} <- infos(statistics_keys(), State),
+    = [{K, V} || {K, V} <- infos(statistics_keys() -- [backing_queue_status], State),
                  not lists:member(K, ExtraKs)],
     rabbit_core_metrics:queue_stats(Name, Extra ++ Infos),
     rabbit_core_metrics:queue_stats(Name, MR, MU, M, R).


michaelklishin (approved):

gomoripeti:
I wonder how the mgmt UI queries individual queue info (on the single queue page)? As it gets the displayed "queue storage version" from the backing_queue_status field. (https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_management/priv/www/js/formatters.js#L198)

dcorbacho:
Thanks for the heads up @gomoripeti, that one is going to be reintroduced as a top-level key.

michaelklishin:
@H4rry12 you can obtain a queue's ingress and egress rates without using this leaked (and now removed) implementation detail. That's how management UI is still able to provide them, it does not use any internal or hidden API endpoints.

Start with [`localhost:15672/api/`](http://localhost:15672/api/) (assuming a locally running node). Do not use `GET /api/queues` to only get one metric of one queue, that's **extremely** expensive and wasteful.

jpanynines:
@michaelklishin  on the commit message is said that we can still get these `backing_queue_status` metrics using amqqueue API. Can you give us more information on how to do that? I used the bunny library to connect to an rabbitmq 3.13 server and did not saw any way to get this information. Maybe I am using the wrong lib? Thank you in advance for your time.

michaelklishin:
The metrics removed here are implementation details. They are not supposed to be used. In fact, when CQv1 are removed completely (most likely in 4.1), they won't be available.

michaelklishin:
Metrics are not consumed using client libraries. See [Monitoring](https://www.rabbitmq.com/docs/monitoring).

PR #9586, opened 2023-09-29
Make looking_glass an optional dep for deps/rabbit

pjk25:
https://github.com/rabbitmq/rules_erlang/tree/main/gazelle#directives


pjk25:
Follow up to #9585

michaelklishin (approved):

Issue #9406, opened 2023-09-12
Slow mnesia->khepri migration with many topic bindings


mkuratczyk:
### Describe the bug

Importing [1 queue with 100k topic bindings](https://github.com/rabbitmq/sample-configs/blob/main/topic-bindings/q1-with-100k-topic-bindings.json) takes 90s with Khepri and roughly twice as long with Mnesia. However, if imported into Mnesia, when a migration to Khepri is triggered, it takes 14-15 minutes.

### Reproduction steps

1. start a node
2. import https://github.com/rabbitmq/sample-configs/blob/main/topic-bindings/q1-with-100k-topic-bindings.json
3. `rabbitmqctl enable_feature_flags khepri_db`
4. notice how long it takes to enable

### Expected behavior

This is not a realistic scenario, so we don't necessarily need to optimize, but it's worth checking whether the behaviour it triggers could occur in other situations as well.

### Additional context

I've checked and this is not related to logging (the migration process currently logs a lot of debug info) - I set the log level to `error` to make sure logging can't cause this.

![migration](https://github.com/rabbitmq/rabbitmq-server/assets/9566114/bd4e14f9-0285-4e71-8d84-69e432b11ca5)

PR #9646, opened 2023-10-05
Migrate records from Mnesia to Khepri with async commands

the-mikedavis:
Migration from Mnesia to Khepri is currently much slower than importing equivalent definitions because the Ra commands to insert records into the Khepri store are handled in series with fsync calls for each command. We can instead pipeline commands (i.e. `ra:pipeline_command/4`) via Khepri's `async` command option to allow Ra to batch the commands. With this change, migration should take the same time or less than importing definitions.

This PR updates Khepri to take advantage of changes to the async API (https://github.com/rabbitmq/khepri/pull/228). `rabbit_db_m2k_converter`, the converter module which calls out to the converter modules for individual records, adds some functions each converter module calls to take correlation IDs and interact with the state. The other converter modules generally look like this now:

```erlang
%% rabbit_db_*_m2k_converter
copy_to_khepri(Table, Record, State) ->
    %% This function exerts backpressure if there are too many async requests
    %% in flight.
    rabbit_db_m2k_converter:with_correlation_id(
      fun(CorrId) ->
              Payload = convert_mnesia_record_to_khepri_payload(Record),
              %% See `khepri:async_option()'
              Extra = #{async => CorrId},
              rabbit_khepri:put(PathPattern, Payload, Extra)
      end, State).
```

Fixes https://github.com/rabbitmq/rabbitmq-server/issues/9406


deps/rabbit/src/rabbit_db_m2k_converter.erl
```diff
@@ -14,17 +14,78 @@
...
+         get_store_id/1]).
+
+%% `mnesia_to_khepri_converter' callbacks.
 -export([init_copy_to_khepri/4,
          copy_to_khepri/3,
-         delete_from_khepri/3]).
+         delete_from_khepri/3,
+         finish_copy_to_khepri/1]).
+
+-define(MAX_ASYNC_REQUESTS, 64).
```
the-mikedavis, line 27: @kjnilsson do you have a good intuition for how many in-flight `ra:pipeline_command/4`s we should allow? I picked this so that mnesia->khepri migration would be a little faster in my local testing than import_definitions (a worker pool of 32 workers) but I don't want to accidentally overload Ra if I'm setting this too high

michaelklishin: The value of `64` sounds reasonable to. We can generally assume that this migration will introduce cluster load, and the operators should plan the migration based on that. That is, migrate during a less busy period.

deps/rabbit/src/rabbit_db_binding_m2k_converter.erl
```diff
@@ -29,90 +28,90 @@
...
+                 [Table, Path, CorrId],
+                 #{domain => ?KMM_M2K_TABLE_COPY_LOG_DOMAIN}),
+              khepri:transaction(
+                StoreId,
+                fun() ->
+                        %% Store the exchange's type in the exchange name
+                        %% branch of the tree.
+                        XPath =
+                            rabbit_db_binding:khepri_route_exchange_path(
+                              XName),
```
dumbbell, line 70: I think this line can be moved outside of the transaction.

deps/rabbit/src/rabbit_db_m2k_converter.erl
```diff
@@ -63,65 +124,130 @@ init_copy_to_khepri(StoreId, MigrationId, _Tables, Migrations) ->
...
+                      case Resp of
+                          ok ->
+                              {ok, State1};
+                          {ok, _} ->
+                              {ok, State1};
+                          {error, _} = Error ->
+                              Error;
+                          {exception, _, _, _} = Exception ->
+                              khepri_machine:handle_tx_exception(
+                                Exception)
```
dumbbell, line 230: The `khepri_machine` module is not part of the public API.

Should we expose the same function from `khepri`? Perhaps it should be handled by `khepri:handle_async_ret/2`?

deps/rabbit/src/rabbit_db_m2k_converter.erl
```diff
@@ -63,65 +124,130 @@ init_copy_to_khepri(StoreId, MigrationId, _Tables, Migrations) ->
...
+            SeqNo = SeqNo0 + 1,
+            AsyncRequests1 = AsyncRequests#{SeqNo => Fun},
+            State = State0#?MODULE{seq_no = SeqNo,
+                                   async_requests = AsyncRequests1},
+            case Fun(SeqNo) of
+                ok ->
+                    {ok, State};
+                {error, _} = Error ->
+                    Error
+            end
```
dumbbell, line 247: This code is duplicated, the same exists in `with_correlation_id/2`. Can this be moved to a dedicated function?

deps/rabbit/src/rabbit_db_binding_m2k_converter.erl
```diff
@@ -29,90 +28,90 @@
...
-                        _ ->
-                            sets:new([{version, 2}])
-                    end,
-              case khepri_tx:put(Path, sets:add_element(Binding, Set)) of
-                  ok    -> {ok, State#?MODULE{exchanges = Xs}};
-                 Error -> Error
-              end
-      end, rw);
+    rabbit_db_m2k_converter:with_correlation_id(
+      fun(CorrId) ->
```
dumbbell, line 57: All modules do the `get_store_id/1` + `with_correlation_id/2` combination of functions. Perhaps the store identifier could be passed as an argument to the anonymous function given to `with_correlation_id/2`?

deps/rabbit/src/rabbit_db_binding_m2k_converter.erl
```diff
@@ -29,90 +28,90 @@
...
+              Extra = #{async => CorrId},
+              ?LOG_DEBUG(
+                 "Mnesia->Khepri data copy: [~0p] path: ~0p corr: ~0p",
+                 [Table, Path, CorrId],
+                 #{domain => ?KMM_M2K_TABLE_COPY_LOG_DOMAIN}),
+              khepri:transaction(
```
dumbbell, line 63: I think I would prefer we always use the `rabbit_khepri` wrapper in case we want to do generic things in the future.

The comment above about the store identifier becomes irrelevant if we do this.

dumbbell (changes requested):

deps/rabbit/src/rabbit_db_binding_m2k_converter.erl
the-mikedavis: As mentioned below, I switched to the `rabbit_khepri` functions and dropped StoreId

the-mikedavis:
I switched this back to a draft - we want to tune the async API a bit more to make sure we cover all of the Khepri internals: https://github.com/rabbitmq/khepri/pull/234

dumbbell (changes requested):
I just released Khepri 0.10.0, could you please update the references?

dumbbell (approved):

dumbbell:
Khepri 0.10.1 was released so that the `LICENSE-*` files are published to Hex.pm. Thank you @the-mikedavis for the quick pull request!

PR #9678, opened 2023-10-11
Osiris v1.6.8

kjnilsson:
This osiris release contains optimisations and bug fixes:

* Various index scanning operations have been substantially improved resulting in up to 10x improvement for certain cases.
* A bug which meant stream replication listener would fail if the TLS version was limited to `tlsv1.3` has been fixed.
* A bug where the log may be incorrectly truncated when filters are used has been fixed. (only relevant for RabbitMQ 3.13 and above).
* Startup handles one more case where a file has been corrupted after an unclean shutdown.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -43,7 +43,7 @@ bazel_dep(
 
 bazel_dep(
     name = "rabbitmq_osiris",
-    version = "1.6.7",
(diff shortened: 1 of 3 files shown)


michaelklishin:
Note: this is superseded by `1.6.9` which was also backported to `v3.11.x`.

PR #9697, opened 2023-10-13
Properly list Web-MQTT connections in CLI list_mqtt_connections

gomoripeti:
## Proposed Changes

Before this change only `:not_found` was printed for each Web-MQTT connection.

See https://github.com/rabbitmq/rabbitmq-server/discussions/9302 for details

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbitmq_mqtt/src/rabbit_mqtt.erl
+++ b/deps/rabbitmq_mqtt/src/rabbit_mqtt.erl
@@ -69,7 +69,13 @@ emit_connection_info(Items, Ref, AggregatorPid, Pids) ->
     rabbit_control_misc:emitting_map_with_exit_handler(
       AggregatorPid, Ref,
       fun(Pid) ->
-              rabbit_mqtt_reader:info(Pid, Items)
+              case rabbit_mqtt_reader:info(Pid, Items) of
+                  {error, not_found} ->
+                      %% a Web-MQTT connection
+                      rabbit_web_mqtt_handler:info(Pid, Items);
+                  Result ->
+                      Result
+              end
       end, Pids).
 
 -spec close_local_client_connections(atom()) -> {'ok', non_neg_integer()}.
--- a/deps/rabbitmq_web_mqtt/src/rabbit_web_mqtt_handler.erl
+++ b/deps/rabbitmq_web_mqtt/src/rabbit_web_mqtt_handler.erl
@@ -22,7 +22,8 @@
     terminate/3
 ]).
 
--export([conserve_resources/3]).
+-export([conserve_resources/3,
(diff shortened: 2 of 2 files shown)


deps/rabbitmq_mqtt/src/rabbit_mqtt.erl
```diff
@@ -69,7 +69,13 @@ emit_connection_info(Items, Ref, AggregatorPid, Pids) ->
     rabbit_control_misc:emitting_map_with_exit_handler(
       AggregatorPid, Ref,
       fun(Pid) ->
-              rabbit_mqtt_reader:info(Pid, Items)
+              case rabbit_mqtt_reader:info(Pid, Items) of
+                  {error, not_found} ->
+                      %% a Web-MQTT connection
```
johanrhodin, line 74: Are we sure there are no other ways to get `{error, not_found}`, or is it always a Web-MQTT connection?

michaelklishin: The logic seems to be "try obtaining connection metrics as if it was an MQTT connection, if not, do the same for Web MQTT". If both fail, then return `{error, not_found}`.

The original function returns `{error, not_found}` if so does `rabbit_mqtt_reader:info/2` so it seems tobe a safe thing to.

gomoripeti: if `Pid` is a `rabbit_mqtt_reader` gen_server (ie represents a regular MQTT connection), then there is no way it can return `{error, not_found}`. As Michael described then fall back to try Web-MQTT.

(if either of both calls exits with a "benign" reason, then `rabbit_control_misc:emitting_map_with_exit_handler` catches that and continues. Where "benign" is if the gen_server process does not exist, has shut down or cannot be contacted on a remote node)

gomoripeti:
CI fails because of xref/dialyzer, and they are right. currently rabbitmq_mqtt plugin calls rabbitmq_web_mqtt_plugin which is not a dependency. I'm not sure how to resolve this dependency. 
- I was thinking that somehow a process could register a callback to fetch info items. By default it is `rabbit_mqtt_reader` so only Web-MQTT connections should register an alternative. (Unfortunately `pg` used currently to register client_ids does not support attaching metadata to a registration)
- or maybe Web-MQTT connections should be registered separately and listed with a separate CLI command

michaelklishin:
@gomoripeti listing using a separate command would be by far the easiest option

gomoripeti:
> listing using a separate command would be by far the easiest option

I agree, will proceed on that route

PR #9708, opened 2023-10-16
Introduce a configurable limit to HTTP API request body size

michaelklishin:


qlong1505:
Hi @michaelklishin 
At file [deps/rabbitmq_management/src/rabbit_mgmt_util.erl](https://github.com/rabbitmq/rabbitmq-server/pull/9708/files#diff-5d38c63401e5c2e515037fd79794614756a9048322c560fe22b2457f44b6aedd), line 693
```cpp
case bit_size(Acc) > BodySizeLimit of
```
If I set in `rabbitmq.conf` file as `management.http.max_body_size = 20000000`, does it means that the maximum body size is only set at about 20Mbit base on the logic above? From the beginning I assume it is MiB but I always received 
```json
{
    "error": "bad_request",
    "reason": "Exceeded HTTP request body size limit"
}
```
when trying to push a msg size 15MB via http API. If I change the ``management.http.max_body_size = 160000000` then I receive 
```json
{"routed":true}
```

michaelklishin:
@qlong1505 the value is in bytes, as the name suggests. The computed body size is also in bytes.

The limit is meant to help you establish a guard rail, not allow for super fine grained control down to the byte.

Please use GitHub Discussions for questions. And consumer publishing using a regular messaging client and not the HTTP API, in particular for large messages.

PR #9715, opened 2023-10-17
Actually nack when using 'Nack message requeue true'

kjnilsson:
Option in mgmt UI.


Gsantomaggio (approved):
LGTM

I think that the atom `ack_requeue_true` is not correct here; maybe it would be better `nack_requeue_true`.  BTW I would not change it to avoid losing compatibility with external tools like `rabbitmqadmin get queue=my_quorum_q  ackmode=ack_requeue_true`

michaelklishin:
@Gsantomaggio note that we cannot easily change the atoms as it would change the API. Maybe for 3.13 but not in this PR which we will backport.

Gsantomaggio:
@michaelklishin yes, I agree.

PR #9734, opened 2023-10-19
Add WORKSPACE.bzlmod

pjk25:
This file shadows the WORKSPACE file when bzlmod is used, and reduces the fetching of unused dependencies


pjk25:
I forgot that I already slimmed down the `WORKSPACE` enough that the `WORKSPACE.bzlmod` file would be identical

PR #9767, opened 2023-10-24
AMQP 1.0 Erlang client: integrate credentials_obfuscation

michaelklishin:
`gen_statem:format_status/1` did not make any difference for logged crash reports, so we reach for the solution we already use in multiple places.

References #9763.


michaelklishin:
The failures on Erlang 25 seem to be repeatable. Converting to a draft.

michaelklishin:
I cannot reproduce these failures locally on `25.3.2.6`.

PR #9768, opened 2023-10-24
Optimise rabbit_registry:binary_to_type/1

kjnilsson:
By not roundtripping binary to list before converting to atom.


lukebakken (approved):
👍 👍 👍

PR #9782, opened 2023-10-25
rabbit_db: `join/2` now takes care of stopping/starting RabbitMQ

dumbbell:
## Why
Up until now, a user had to run the following three commands to expand a cluster:

1. stop_app
2. join_cluster
3. start_app

Stopping and starting the `rabbit` application and taking care of the underlying Mnesia application could be handled by `join_cluster` directly.

## How

After the call to `can_join/1` and before proceeding with the actual join, the code remembers the state of `rabbit`, the Feature flags controler and Mnesia.

After the join, it restarts whatever needs to be restarted to. It does so regardless of the success or failure of the join. One exception is when the node switched from Mnesia to Khepri as part of that join. In this case, Mnesia is left stopped.


michaelklishin (approved):

dumbbell:
This work is the basis for #9797. I will mark this pull request as ready to backport to 3.13.x once it is confirmed that both changes solve peer discovery with Khepri.

Issue #8364, opened 2023-05-26
minor: shovel cleanup_spec doesn't work as intended


gomoripeti:
### Describe the bug

`rabbit_shovel_dyn_worker_sup_sup:cleanup_specs` is called every minutes, and it is intended to check if there is any child of the supervisor for which there is no parameter definition. However there is a mismatch as `rabbit_runtime_parameters:list_component(<<"shovel">>)` returns a list of key-value lists not a list of vhost-name tuples, so the substraction will report all children as "undefined". Then `mirrored_supervisor:delete_child` is called for all children. (Un)fortunately this rejects the delete with `{error, running}` in case the child process is alive.

If the intention is to remove only not-running children (because PR https://github.com/rabbitmq/rabbitmq-server/pull/2708 which introduced the `cleanup_spec` function contains other safeguards which make sure a dynamic shovel worker is never resurrected) then there is just a unused filtering, dead code.

If the intention is to stop and delete even running workers which don't have a matching parameter definition, then the function should also call `terminate_child` but then the filtering also needs to be fixed to not delete valid shovel workers. (I've seen cases of running shovels without parameters, but that was on pre 3.8.10 versions which dont include PR #2708, so that scenario might not happen any more)

I can only think of a combination when a shovel worker is not running (maybe crashed) but it has a parameters definition. In this case the shovel child is currently deleted.

To sum up I just wanted to report two lines of dead code, which are probably harmless in most cases. I am happy to submit a PR if it is clarified whether adding `terminate_child` is desired or not.

### Reproduction steps

1. Create two queues and a non-autodelete shovel called `sh1`
2. In remote_shell `recon_trace:calls([{mirrored_supervisor, delete_child, fun([rabbit_shovel_dyn_worker_sup_sup, _]) -> message(caller()), return_trace() end}],10).`
The output is as follows
```
16:54:26.071162 <0.824.0> mirrored_supervisor:delete_child(rabbit_shovel_dyn_worker_sup_sup, {<<"/">>,<<"sh1">>}) {rabbit_shovel_dyn_worker_sup_sup,'-cleanup_specs/0-fun-1-',2}

16:54:26.072481 <0.824.0> mirrored_supervisor:delete_child/2 --> {error,running}

[...]

PR #9785, opened 2023-10-25
Speed up shovel start

mkuratczyk:
Previously whenever a shovel was declared, a full cleanup of all shovels was performed. Since some users have a lot of shovels, that would slow things down - especially import and node startup when we performed a cleanup of all shovels for each new shovel.

In a test with 1000 shovels, this change speeds up a definition import from a few minutes to a few seconds.

Fixes https://github.com/rabbitmq/rabbitmq-server/discussions/9714
Fixes https://github.com/rabbitmq/rabbitmq-server/issues/8364

### DIFF
--- a/deps/rabbitmq_shovel/src/rabbit_shovel_dyn_worker_sup_sup.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_shovel_dyn_worker_sup_sup.erl
@@ -84,7 +84,8 @@ stop_child({VHost, ShovelName} = Name) ->
 
 cleanup_specs() ->
     SpecsSet = sets:from_list([element(2, element(1, S)) || S <- mirrored_supervisor:which_children(?SUPERVISOR)]),
-    ParamsSet = sets:from_list(rabbit_runtime_parameters:list_component(<<"shovel">>)),
+    ParamsSet = sets:from_list([ {proplists:get_value(vhost, S), proplists:get_value(name, S)}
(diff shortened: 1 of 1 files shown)


michaelklishin:
Note that the full cleanup was introduced in https://github.com/rabbitmq/rabbitmq-server/commit/bd29642f04e62360ad2ce72797eaed7db5b4064e (#2708) for a reason. Perhaps we don't need to do it when a Shovel is added, or can make it conditional until after a definition import is done but I felt it was worth pointing out.

mkuratczyk:
Thanks. I made it a draft exactly because I didn't fully understand what we were safeguarding against. Reading more about it, it could be that a better solution is to fix https://github.com/rabbitmq/rabbitmq-server/issues/8364. If what Péter described is still the case, and I think it is, then fixing the filtering should make `cleanup_spec` much less expensive since in most cases it wouldn't attempt to cleanup anything. I will look into this today.

mkuratczyk:
Indeed, fixing the filtering gives us the same improvement in import / nodes joining (scaling from 1 to 3 nodes with 1000 shovels goes down from ~20 minutes to ~20 seconds for me) but without losing the full cleanup process on each shovel start. Additionally periodic cleanups (executed every minute) are faster as well.

michaelklishin (approved):
My tests confirm the claim that with this, 1K Shovels is imported in seconds, not minutes.

michaelklishin:
A follow-up issue https://github.com/rabbitmq/rabbitmq-server/issues/9894.

PR #9794, opened 2023-10-26
Fix quorum queue open files info

kjnilsson:
At some point Ra stopped updating the ra_open_files_metrics table in favour of using counters. This change updates the rabbit_quorum_queue:open_files/1 function to use ra counters instead of querying the deprecated ETS table.


michaelklishin:
This depends on a Ra `2.7.0`-specific function, so we cannot backport this as is. Which is acceptable.

PR #9809, opened 2023-10-27
Use new OTP 26 fun ets:lookup_element/4

ansd:
Routing to nowhere with the direct exchange is now a bit faster 🙃
```
java -jar target/perf-test.jar -x 2 -y 0 -k route-nowhere -z 60
```


michaelklishin:
About 2.5% faster according to my tests.

PR #9815, opened 2023-10-27
4.0: Remove classic mirrored queues

dcorbacho:
## Types of Changes
- [ ] Build system and/or CI

## Checklist


michaelklishin (approved):

ansd:
It looks like consumer argument `x-cancel-on-ha-failover` should have been deleted as part of this PR.

Likewise, MQTT connections don't need to perform failover for mirrored classic queues in https://github.com/rabbitmq/rabbitmq-server/blob/c67c940bcc5d14ff47ebc78e3d7203478ed734b4/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl#L2000-L2004 anymore.

michaelklishin:
Filed https://github.com/rabbitmq/rabbitmq-server/issues/11540 for those two.

michaelklishin:
It is not entirely obvious to me, what the `rabbit_channel` of `x-cancel-on-ha-failover` should do now: always recover? For classic (no longer replicated) queues the choice is less obvious.

ansd:
Copied from internal Slack:

> Channels should not try to re-register consumers.
> If the classic queue is down, consumption is stopped and the client gets notified.
> Likewise, MQTT connection should not try to re-consume. Instead, the MQTT connection will be terminated.
> I think it would be less confusing just to always cancel the consumer

PR #9875, opened 2023-11-03
Logger exchange: fix race condition during initialisation

dcorbacho:
The logger exchange needs to declare the exchange during initialisation, which requires the metadata store to be ready.
Metadata store initalisation happens in a rabbit boot step after logger initialisation in the second phase of the prelaunch. The spawned process that declares the exchange, should also wait for the store to be ready. Otherwise it enters a loop trying to decide which store to use which generates a huge log and delays initialisation:

```
Mnesia->Khepri fallback handling: Mnesia function failed because table
 'rabbit_vhost' is missing or read-only. Migration could be in progress;
 waiting for migration to progress and trying again... 
```

This commit gives it 60 seconds for the metadata store to boot, and only afterwards tries (and retries if needed) to declare the exchange.

Detected by frequent CI flakes in `logging_SUITE:logging_to_exchange_works` 

## Types of Changes
- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit.erl
+++ b/deps/rabbit/src/rabbit.erl
@@ -1079,6 +1079,7 @@ stop(State) ->
         [] -> rabbit_prelaunch:set_stop_reason(normal);
         _  -> rabbit_prelaunch:set_stop_reason(State)
     end,
+    rabbit_db:clear_init_finished(),
     rabbit_boot_state:set(stopped),
     ok.
 
--- a/deps/rabbit/src/rabbit_db.erl
+++ b/deps/rabbit/src/rabbit_db.erl
@@ -14,13 +14,17 @@
 
 -include_lib("rabbit_common/include/logging.hrl").
 
+-define(PT_KEY_INIT_FINISHED, {?MODULE, node(), initialisation_finished}).
+
 -export([init/0,
          reset/0,
          force_reset/0,
          force_load_on_next_boot/0,
          is_virgin_node/0, is_virgin_node/1,
          dir/0,
-         ensure_dir_exists/0]).
+         ensure_dir_exists/0,
+         is_init_finished/0,
+         clear_init_finished/0]).
 
 %% Exported to be used by various rabbit_db_* modules
 -export([
@@ -62,6 +66,7 @@ init() ->
                "DB: initialization successeful",
                #{domain => ?RMQLOG_DOMAIN_DB}),
 
+            init_finished(),
             post_init(IsVirgin),
 
             ok;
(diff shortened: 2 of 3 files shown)


deps/rabbit/src/rabbit_db.erl
```diff
@@ -72,6 +76,20 @@ init() ->
...
+    %% initialised. The initialisation happens on a rabbit boot step,
+    %% after the second phase of the prelaunch where the logger is
+    %% configured.
+    %% Using this persistent term the logger exchange can delay
+    %% declaring the exchange until the metadata store is ready.
+    persistent_term:put(?PT_KEY_INIT_FINISHED, true).
+
+is_init_finished() ->
+    persistent_term:get(?PT_KEY_INIT_FINISHED, false).
+
```
dumbbell, line 92: Could you please move these functions after the `init_with_*()` functions and before `reset()`? Otherwise, they sit in the "middle" of the init code.

deps/rabbit/src/rabbit_db.erl
```diff
@@ -62,6 +65,7 @@ init() ->
                "DB: initialization successeful",
                #{domain => ?RMQLOG_DOMAIN_DB}),
 
+            init_finished(),
```
dumbbell, line 68: Should there be an "undo" function executed when `rabbit` stops?

dumbbell (changes requested):

deps/rabbit/src/rabbit_db.erl
```diff
@@ -102,6 +107,23 @@ init_using_khepri() ->
...
+    %% configured.
+    %% Using this persistent term the logger exchange can delay
+    %% declaring the exchange until the metadata store is ready.
+    persistent_term:put(?PT_KEY_INIT_FINISHED, true).
+
+is_init_finished() ->
+    persistent_term:get(?PT_KEY_INIT_FINISHED, false).
+
+clear_init_finished() ->
+    persistent_term:erase(?PT_KEY_INIT_FINISHED).
```
dumbbell, line 125: We probably don't need to return the return value of `persistent_term:erase/1`.

Can you ignore the return value and always return `ok`?

deps/rabbit/src/rabbit.erl
```diff
@@ -467,6 +467,7 @@ stop() ->
                     Product = product_name(),
                     ?LOG_INFO("~ts is asked to stop...", [Product],
                               #{domain => ?RMQLOG_DOMAIN_PRELAUNCH}),
+                    rabbit_db:clear_init_finished(),
```
dumbbell, line 470: `rabbit:stop/0` is not the real stop function, it's a helper only. Could you please move the call to the real `rabbit` stop function, `stop/1`?

dumbbell (changes requested):

deps/rabbit/src/rabbit.erl
```diff
@@ -1080,6 +1080,7 @@ stop(State) ->
         _  -> rabbit_prelaunch:set_stop_reason(State)
     end,
     rabbit_boot_state:set(stopped),
+    rabbit_db:clear_init_finished(),
```
dumbbell, line 1083: Could you please move this line before `rabbit_boot_state:set(stopped)`?

This is to make sure that we set the boot state as the very last step in the function.

dumbbell (changes requested):

dumbbell (approved):

PR #7753, opened 2023-03-29
See #7593. Use connection_max to stop connections in rabbitmq

SimonUnge:
## Proposed Changes
Closes #7593 

Use connection_max to limit connections per node, from rabbitmq code, instead of ranch.

Add ranch_connection_max setting for setting ranch connection max.

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
If you're unsure about any of them, don't hesitate to ask on the mailing list.
We're here to help!
This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
- [x] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [x] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

## Further Comments

N/A

PR #9876, opened 2023-11-03
WIP: account other protocols in node connection limit.

ikvmw:
## Proposed Changes

Current node `connection_max` accounts only amqp0-9-1 connections and checked only inside amqp0-9-1 code as well.
This change fixes it. 

see also https://github.com/rabbitmq/rabbitmq-server/pull/7753 and https://github.com/rabbitmq/rabbitmq-server/pull/7788

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

Streams test TBD


ikavgo:
there must be a follow up PR for vhost connections limit everywhere

deps/rabbit/src/pg_local.erl
```diff
@@ -74,6 +76,12 @@ leave(Name, Pid) when is_pid(Pid) ->
     _ = ensure_started(),
     gen_server:cast(?MODULE, {leave, Name, Pid}).
 
+-spec count_unique_members(name()) -> non_neg_integer().
+
+count_unique_members(Name) ->
+    _ = ensure_started(),
+    ets:select_count(?TABLE, [{{{member, Name, '_'}, '_'}, [], [true]}]).
```
ansd, line 83: I fear this operation might become too CPU costly with many connections.
I know this table is an `ordered_set` and the connections are located in O(log n). However, this operation still matches a large number of connections.

Specifically https://blog.rabbitmq.com/posts/2023/03/native-mqtt/ was able to connect 9 millions connections.
I would like to see how https://github.com/rabbitmq/mqtt-testing/tree/main/many-conn compares before and after this PR: That test case deploys 400 pods with 25k connections each (=10 million) where each pod connects every 80ms. So, there are
* 1000ms / 80ms = 12.5 connections per second per pod
* 400 pods * 12.5 connections per second = 5000 connections per second
* since this is a 3 node cluster: 5000 connections per second / 3 nodes = 1666 connections per second against each node
* during end of the test case there will be > 3,000,000 connections on each node, and therefore this operation will match 1666 * 3,000,000 = 5 billion records per second.

I've seen in some scaling tests that ETS / Mnesia operations become too CPU intense, and removed all such bottlenecks. Once such as example is https://github.com/rabbitmq/rabbitmq-server/commit/35afeffceb9044082bbb4f060922dc150d6d4250 . This PR should not introduce any new such bottlenecks. We do want to continue connecting millions of MQTT connections efficiently to RabbitMQ.

Alternatively, look into more efficient methods in determining the total number of connections per node.
For example https://github.com/rabbitmq/rabbitmq-server/blob/b4804a8015d1276f883365e335955d058d3f5c23/deps/rabbit/src/rabbit_vhost_limit.erl#L94 just reads a counter.

Another idea is simply `ets:info(tracked_connection, size).` which works for all native connections, that is AMQP 0.9.1, MQTT, Steams, and from 4.0 onwards AMQP 1.0 and STOMP.

There might be other more efficient ways to query the total number of connections per node.

ansd (changes requested):

deps/rabbit/src/pg_local.erl
ikavgo: Hey, thanks for the review. I see what you mean with select vs simple size. 

First time I learn about rabbit_*tracking is from you today. Yet another registration facility it seems. 

The example of vhost_limit is a bit confusing at the moment since count_tracked_items_in does that rpc thingy I asked about on slack recently (what a coincidence!): https://github.com/rabbitmq/rabbitmq-server/blob/d258626148014f6463f48a5778c7ba00e3f74e30/deps/rabbit/src/rabbit_connection_tracking.erl#L161

michaelklishin:
@ikavgo this rebase it. 

It's also not obvious to me whether @ansd's feeback on the use of `ets:select_count` was addressed or not.

ikavgo:
I need to recall all details about this PR. ETS query could be slow, and RPC that currently happens can be slow too, thou now that CPU intensive, that's right. Regarding suggested `rabbit_connection_tracking:count_tracked_items_in({vhost, VirtualHost}),` I'm not sure because we are removing per vhost limit here, IIRC.

PR #9982, opened 2023-11-23
Make 'ctl node_health_check' a no-op

michaelklishin:
It's been deprecated over three years ago [1]
(before RabbitMQ 3.9).

This version for `3.13` intentionally does not remove the command or make it always fail.
Making it fail will make it easy to notice the deprecation but at the same time, seriously
annoy a lot of operators. Consider scenarios where the health check is used by external tools outside of operator's control.

So this seems to be the most natural deprecation progression for 3.13.

1. https://github.com/rabbitmq/rabbitmq-server/commit/e01753ed7adb4ae2d3fa217bfee2598736ca035f


lhoguin:
Should a message be printed saying this is a no-op and it has been deprecated, instead of doing nothing silently?

michaelklishin:
@lhoguin we already do that unless `-q` or `-s` are passed. Right now there is no way to "break through" `--quiet` and `--silent`. I assume if you explicitly silence output, you are OK with missing warnings and such.

PR #10021, opened 2023-12-01
rabbit_feature_flags: Fix registry reset vs. Erlang embedded mode

dumbbell:
## Why

If the Erlang VM is running in embedded mode, it will refuse to auto-load any module from disk. That's because all modules are loaded from the Erlang release right from the start.

## How

After purging the module, we explicitly reload the stub version from the disk.


dumbbell:
I forgot 3.11.x and 3.12.x didn't have the reset function. Closing the backport pull request and leaving this change for 3.13.0 only.

Issue #7508, opened 2023-03-02
Sync AMQP 1.0 Application Properties and AMQP 0.9.1 Properties/Headers in RabbitMQ Shovel


Haigutus:
https://groups.google.com/g/rabbitmq-users/c/2RhTKTYKc7I/m/_dAH1cUqBQAJ

RabbitMQ Shovel currently does not keep the AMQP 1.0 Application Properties and AMQP 0.9.1 Properties/Headers in sync when copying messages between exchanges or queues when transferring messages between AMQP 1.0 and AMQP 0.9.1 brokers.

To improve the functionality of RabbitMQ Shovel, I suggest that the Shovel should always copy the Application Properties and Properties/Headers of messages along with their body, and keep them in sync across all copies of the message. This will ensure that headers can be used for routing when transferring messages between AMQP 1.0 and AMQP 0.9.1 brokers.

michaelklishin:
Those properties do not have a perfect 1:1 match. At some point we plan to use AMQP 1.0 properties internally for all messages but we are not there yet.

So this is significantly more involved than it may seem.

Haigutus:
Hi those are user defined key and value pairs, so no need to map anything, just carry over.

It seems that for AMQP 0.9.1 -> AMQP 1.0 it is actually already done, but other way it seems not, but I am not familiar enough with the code base

https://github.com/rabbitmq/rabbitmq-server/blob/942c4ab5426a376f83fff32114b98306b9ddd6f2/deps/rabbitmq_shovel/src/rabbit_amqp10_shovel.erl#L397

Haigutus:
For reference

AMQP 1.0

Application Properties:

Any application-specific data, encoded as key-value pairs.

AMQP 0.9.1

Properties/Headers:
Additional metadata about the message, encoded as key-value pairs.

Haigutus:
Also it seems, that for AMQP 1.0 this has already been implemented.
https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_amqp1_0/README.md
![image](https://user-images.githubusercontent.com/11408965/236261072-e03be63c-7af6-4c42-87c6-d4bd9f6a3050.png)

Do these settings also apply for shovel?

michaelklishin:
This was addressed (without any configuration settings) in https://github.com/rabbitmq/rabbitmq-server/pull/10037 for 3.12.x.

[...]

PR #10022, opened 2023-12-01
Shovel: convert AMQP 1.0 props and app props to AMQP 0.9.1 props and headers

luos:
Hi, 

We've implemented Application Property conversion to AMQP 0.9.1 headers for shovel. We wanted this to work on 3.12 and we were not sure if this is appropriate for these message container changes. 

Let us know what you think.

- Timestamps are milliseconds in AMQP 1.0, but in AMQP 0.9.1 it is seconds. Fixed by multiplying the timestamp by 1 000.
- Shovel crashed if user_id was set in the message because the encoding was as utf8 while it should be a byte array.
- Negative integers were encoded as integers - therefore leading to incorrect positive values.
- Float values were not supported by the client.
- Fixed priority header encoding in AMQP 1.0. It was set as uint but it should be ubyte.
- Priority of the message is now in the Headers instead of Application Properties. This is potentially a breaking change.

GitHub issue: https://github.com/rabbitmq/rabbitmq-server/issues/7508.

A `v3.12.x` version of this PR: https://github.com/rabbitmq/rabbitmq-server/pull/10037.

## Proposed Changes

Implement conversion of Application Properties to Headers. 

Configurable by using the following key: 

```
shovel.convert_amqp10_props_to_amqp091 = true
```

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbit/src/mc_amqp.erl
+++ b/deps/rabbit/src/mc_amqp.erl
@@ -63,6 +63,10 @@
 init(Sections) when is_list(Sections) ->
     Msg = decode(Sections, #msg{}),
     init(Msg);
+init(Amqp10Msg) when element(1, Amqp10Msg) =:= amqp10_msg ->
+    [#'v1_0.transfer'{}| AmqpRecords] = amqp10_msg:to_amqp_records(Amqp10Msg),
+    Msg = decode(AmqpRecords, #msg{}),
+    init(Msg);
 init(#msg{} = Msg) ->
     %% TODO: as the essential annotations, durable, priority, ttl and delivery_count
     %% is all we are interested in it isn't necessary to keep hold of the
--- a/deps/rabbitmq_shovel/src/rabbit_amqp091_shovel.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_amqp091_shovel.erl
@@ -12,6 +12,7 @@
 -include_lib("amqp_client/include/amqp_client.hrl").
 -include("rabbit_shovel.hrl").
 
+
 -export([
          parse/2,
          source_uri/1,
@@ -31,7 +32,7 @@
          ack/3,
          nack/3,
          status/1,
-         forward/4
+         forward/3
         ]).
 
 %% Function references should not be stored on the metadata store.
@@ -167,8 +168,8 @@ forward_pending(State) ->
     case pop_pending(State) of
         empty ->
             State;
-        {{Tag, Props, Payload}, S} ->
-            S2 = do_forward(Tag, Props, Payload, S),
+        {{Tag, Msg}, S} ->
+            S2 = do_forward(Tag, Msg, S),
             S3 = control_throttle(S2),
             case is_blocked(S3) of
                 true ->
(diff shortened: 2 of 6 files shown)


michaelklishin:
Message containers were introduced in part to make this kind of conversions easier to pull off.

michaelklishin:
@luos why was the priority value moved to a header in AMQP 1.0? I doubt we can ship this in 3.12.x because of this change.

luos:
Hi,

I think we can definitely remove the priority change, I changed this because there is already a place for it in the AMQP 1.0 Headers and it gets placed into Application Properties. This is not critical for us. 

Regarding the MC, I'd be happy to make this more MC compatible, though I am quite unsure how as shovels are a level above that if I understand it correctly, or if somehow we can make this in a way that in the future it can be easily changed to use MC?

michaelklishin:
@luos can you please re-submit this for `v3.12.x` and put together a version for `main` that would use the `mc` conversion API? We have some pending docs for it in  https://github.com/rabbitmq/rabbitmq-website/pull/1747

michaelklishin:
@luos we can drop the configuration key and treat this as a bug fix for Shovels with AMQP 1.0 destinations.

luos:
We prepared the backporting here: https://github.com/rabbitmq/rabbitmq-server/pull/10037

We will work on the MC version at the end of this week.

luos:
Hi @michaelklishin ,

We've started looking at the conversion with MC. We are planning to remove the `Props` and `Payload` from the `rabbit_shovel_behaviour:forward` function, and replace it with an `mc` "object". 

We see that there is a somewhat related PR already which contains some AMQP 1.0 MC conversion functions, we may depend on these changes. What would be your preference, should we base our changes on that PR (#9022) by @ansd ? 

https://github.com/rabbitmq/rabbitmq-server/pull/9022/files#diff-fb45468cd5ac6820b904a3e908da68af368f0e70126a7df503efedc940e0d03b

michaelklishin:
@luos #9022 is vast in scope and while you can coordinate your changes with it, it is something for 4.0.

deps/rabbit/src/mc_amqp.erl
```diff
@@ -63,6 +63,10 @@
 init(Sections) when is_list(Sections) ->
     Msg = decode(Sections, #msg{}),
     init(Msg);
+init(Amqp10Msg) when element(1, Amqp10Msg) =:= amqp10_msg ->
+    [#'v1_0.transfer'{}| AmqpRecords] = amqp10_msg:to_amqp_records(Amqp10Msg),
```
luos, line 67: This conversions seems a little bit hacky, though we could not come up with a better way. 

By adding the `v1_0.transfer` record match at the beginning of the list, it effectively drops that list item.

deps/rabbitmq_shovel/src/rabbit_amqp10_shovel.erl
```diff
@@ -312,28 +317,27 @@ status(_) ->
...
           ack_mode := AckMode} = State) ->
     OutTag = rabbit_data_coercion:to_binary(Tag),
-    Msg0 = new_message(OutTag, Payload, State),
-    Msg = add_timestamp_header(
-            State, set_message_properties(
-                     Props, add_forward_headers(State, Msg0))),
-    case send_msg(Link, Msg) of
+    MsgCont = mc:convert(mc_amqp, InMsg),
+    AmqpMessage = mc:protocol_state(MsgCont),
+    NewMsg = amqp10_msg:from_amqp_records([#'v1_0.transfer'{} | AmqpMessage]),
```
luos, line 338: This is the other side of that conversion, we have to prepend the `v1_0.transfer` record to have a valid `amqp10_msg`.

deps/rabbitmq_shovel/src/rabbit_amqp10_shovel.erl
```diff
@@ -175,10 +176,14 @@ dest_endpoint(#{shovel_type := dynamic,
...
-    Tag = amqp10_msg:delivery_id(Msg),
-    Payload = amqp10_msg:body_bin(Msg),
-    rabbit_shovel_behaviour:forward(Tag, #{}, Payload, State);
+handle_source({amqp10_msg, _LinkRef, Amqp10Msg}, State) ->
+    Tag = amqp10_msg:delivery_id(Amqp10Msg),
+    Mc = mc:init(mc_amqp, Amqp10Msg, #{
+        % these are required values, however I don't think they make sense for AMQP 1.0
+        exchange => <<>>,
+        routing_keys => [<<>>]
+    }),
```
luos, line 185: `exchange` and `routing_keys` seem to be required by the `mc` implementation. If it is not set, then some crashes can be expected.

deps/rabbitmq_shovel/test/amqp10_dynamic_SUITE.erl
```diff
@@ -168,6 +170,169 @@ simple_amqp10_src(Config) ->
               ok
       end).
 
+message_prop_conversion(Config) ->
```
luos, line 173: These are the same tests as on 3.12.x. With minimal changes, they produce the same output.

deps/rabbitmq_shovel/src/rabbit_amqp091_shovel.erl
```diff
@@ -181,91 +182,51 @@ forward_pending(State) ->
...
                                 exchange = Exchange,
                                 routing_key = RoutingKey},
               #amqp_msg{props = Props0, payload = Payload}}, State) ->
-    Props = (map_from_props(Props0))#{exchange => Exchange,
-                                      routing_key => RoutingKey},
-    % forward to destination
-    rabbit_shovel_behaviour:forward(Tag, Props, Payload, State);
+    Anns = #{exchange => Exchange, routing_keys => [RoutingKey]},
+    Content = rabbit_basic:build_content(Props0, [Payload]),
+    Mc = mc:init(mc_amqpl, Content, Anns),
```
luos, line 228: This conversion as well looks a bit hacky. Let us know if you already have some better way to implement these.

luos (commented):
Hi @michaelklishin , we've updated the changes already merged into 3.12.x with the conversion to `mc`. Let us know what you think.

michaelklishin:
@kjnilsson @ansd this must be the first use of `mc` outside of core and outside of the core team. WDYT? Should Shovels simply do what they do in `v3.12.x`, at least for now?

deps/rabbit/src/mc_amqp.erl
michaelklishin: This introduce an implicit dependency on the AMQP 1.0 client to RabbitMQ core. We plan on making AMQP 1.0 support a core feature later this year (after https://github.com/rabbitmq/rabbitmq-server/pull/9022 is merged) but we cannot accept a dependency on an AMQP 1.0 client.

Just like at some point `rabbit_core` had to be introduced to avoid a dependency on `amqp_client` for AMQP 0-9-1.

michaelklishin (commented):
We have decided to postpone this until after 3.13 ships and then #9022 is merged.

PR #10037, opened 2023-12-04
Convert AMQP 1.0 props and app props to AMQP 0.9.1 props and headers

luos:
Hi,

As requested in https://github.com/rabbitmq/rabbitmq-server/pull/10022 this is a backport to 3.12.x. 

The configuration option was removed and now all the properties are always converted to headers or other appropriate properties.
The priority was moved to the Header section according to [AMQP 1.0 (section 3.2.1)](https://docs.oasis-open.org/amqp/core/v1.0/csprd01/amqp-core-messaging-v1.0-csprd01.html)

- Timestamps are milliseconds in AMQP 1.0, but in AMQP 0.9.1 it is seconds. Fixed by multiplying the timestamp by 1 000.
- Shovel crashed if user_id was set in the message because the encoding was as utf8 while it should be a byte array.
- Negative integers were encoded as integers - therefore leading to incorrect positive values.
- Float values were not supported by the client.
- Fixed priority header encoding in AMQP 1.0. It was set as uint but it should be ubyte.
- Priority of the message is now in the Headers instead of Application Properties. This is potentially a breaking change.

Fixes: #7508

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/amqp10_client/src/amqp10_msg.erl
+++ b/deps/amqp10_client/src/amqp10_msg.erl
@@ -306,7 +306,7 @@ set_headers(Headers, #amqp10_msg{header = Current} = Msg) ->
     H = maps:fold(fun(durable, V, Acc) ->
                           Acc#'v1_0.header'{durable = V};
                      (priority, V, Acc) ->
-                          Acc#'v1_0.header'{priority = {uint, V}};
+                          Acc#'v1_0.header'{priority = {ubyte, V}};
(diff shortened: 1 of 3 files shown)


michaelklishin:
@mergifyio backport v3.13.x

kjnilsson:
@Mergifyio backport main

PR #10058, opened 2023-12-06
Support stream filtering in STOMP

acogoluegnes:


ikavgo:
Do we need a test that generated header works, like full integration test? in python world?

acogoluegnes:
I created an integration test in the stream Java client test suite, there are already other tests related to stream filtering there. As stream filtering is not completely deterministic (it uses a Bloom filter internally), writing end-to-end tests is a bit tedious, but I can add some if you think it's necessary.

ikavgo:
This is how my concern is scoped - I'm pretty sure streams and filtering well tested elsewhere. I just want to make sure the process of translation from stomp headers to args works right. That code adheres to the streams conventions. That's all :-)

acogoluegnes:
You're right, this deserves at least some basic testing (the tests in the stream Java client are quite thorough, with randomness, etc), to make sure the headers are propagated correctly.

ikavgo:
Looking at the size of it that is a huge test :-). One final thing from me now - please please please write a couple of sentences describing what is going on in this test. I'm not sure I understand it fully now and I'm pretty sure I will understand it even less in the future.

PR #10062, opened 2023-12-07
A follow-up to #9550

michaelklishin:


michaelklishin:
Leaning towards simply reverting #9550 https://github.com/rabbitmq/rabbitmq-server/pull/10185.

PR #10088, opened 2023-12-11
rabbit_writer: Convert to a regular gen_server

dumbbell:
## Why

This process failed to implement properly the OTP principles. For instance, the mainloop always kept a reference on the module because it was not tail-recursive.

This prevents the module from being reloaded at runtime: because the process always keep that reference on the module, it is killed by the Code server as part of the code reloading.

### DIFF
--- a/deps/amqp_client/src/amqp_channel.erl
+++ b/deps/amqp_client/src/amqp_channel.erl
@@ -839,7 +839,11 @@ handle_channel_exit(Reason = #amqp_error{name = ErrorName, explanation = Expl},
     handle_shutdown({connection_closing, ReportedReason}, State);
 handle_channel_exit(Reason, State) ->
     %% Unexpected death of a channel infrastructure process
-    {stop, {infrastructure_died, Reason}, State}.
+    Reason1 = case Reason of
+                  {shutdown, R} -> R;
+                  _             -> Reason
+              end,
+    {stop, {infrastructure_died, Reason1}, State}.
 
 handle_shutdown({_, 200, _}, State) ->
     {stop, normal, State};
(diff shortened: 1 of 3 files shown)


deps/amqp_client/src/amqp_channel.erl
```diff
@@ -872,11 +876,7 @@ do(Method, Content, Flow, #state{driver = direct, writer = W}) ->
 
 
 flush_writer(#state{driver = network, writer = Writer}) ->
-    try
-        rabbit_writer:flush(Writer)
-    catch
-        exit:noproc -> ok
-    end;
+    _ = catch rabbit_writer:flush(Writer);
```
lhoguin, line 879: This won't always return `ok` anymore, is that intended? Before it was either returning `ok` or triggering an exception. Now it'll return `ok` or `{'EXIT', ...}` or other fancy tuples `catch` returns.

dumbbell: You're right, let me fix that.

dumbbell: Done!

lhoguin (approved):
Code looks good but I think it's a good idea to have a quic benchmark run against these changes just to make sure we aren't losing performance (in some cases special processes can be more efficient than gen_*, or at least they could be a few years ago). @mkuratczyk

mkuratczyk (approved):
I haven't found any regressions

lhoguin:
Thank you, good to know for future refactorings!!

dumbbell:
Thank you both for your comments and testing!

ansd:
I previously converted the AMQP 1.0 writer to a gen_server and there were no performance regressions.

I found that https://github.com/rabbitmq/rabbitmq-server/blob/a33768471f5e869f3974449c464f6f9b6bd2f801/deps/rabbit/src/rabbit_amqp_writer.erl#L32-L34 speeds up the writer, but that change is less relevant for AMQP 0.9.1 because AMQP 1.0 makes more use of iolists.

PR #10090, opened 2023-12-11
Support stream filtering in AMQP 1.0

acogoluegnes:
Use the x-stream-filter-value message annotation to carry the filter value in a published message.

Use the rabbitmq:stream-filter and rabbitmq:stream-match-unfiltered filters when creating a receiver that wants to filter out messages from a stream.


acogoluegnes:
Superseded by #10098.

PR #10091, opened 2023-12-11
Add a new local random exchange type.

kjnilsson:
Co-authored-by: Marcial Rosales <mrosales@pivotal.io>

This exchange type will only bind classic queues and will only return routes for queues that are local to the publishing connection. If more than one queue is bound it will make a random choice of the locally bound queues.

This exchange type is suitable as a component in systems that run highly available low-latency RPC workloads.

Fixes #8334

### DIFF
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -1063,6 +1063,14 @@ rabbitmq_integration_suite(
 )
 
 rabbitmq_integration_suite(
+    name = "rabbit_local_random_exchange_SUITE",
+    size = "small",
+    additional_beam = [
+        ":test_queue_utils_beam",
+    ],
+)
+
+rabbitmq_integration_suite(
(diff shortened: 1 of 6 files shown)


ansd (approved):

lhoguin (approved):
When testing I had troubles because the routing key must be empty, and I didn't know how to specify that in how I ran perf-test (long story). I managed to test thanks to kura. But perhaps the routing key should be ignored instead of forced empty? Like fanout. Just a suggestion, I don't have strong feelings either way.

All good otherwise!

kjnilsson:
> When testing I had troubles because the routing key must be empty, and I didn't know how to specify that in how I ran perf-test (long story). I managed to test thanks to kura. But perhaps the routing key should be ignored instead of forced empty? Like fanout. Just a suggestion, I don't have strong feelings either way.

So the binding key is validated to be empty as then we get the fastest routing query with khepri.

PR #10101, opened 2023-12-12
Optimise cluster_status cli command when not all nodes are online and responding.

kjnilsson:
If rabbitmqctl cluster_status was called when a node had just been force terminated such that the TCP connection was left dangling it could take several minutes to complete. This was mostly due to various uses of `rabbit_nodes:list_running/0` when retrieving listeners and feature flags and these have now been optimised to mostly avoid calls to this function.

### DIFF
--- a/deps/rabbit/src/rabbit_feature_flags.erl
+++ b/deps/rabbit/src/rabbit_feature_flags.erl
@@ -699,10 +699,9 @@ info(Options) when is_map(Options) ->
 
 get_state(FeatureName) when is_atom(FeatureName) ->
     IsEnabled = is_enabled(FeatureName),
-    IsSupported = is_supported(FeatureName),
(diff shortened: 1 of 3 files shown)


dumbbell (changes requested):
Just a "cosmetic" comment: could you please prefix the commit title of the commit to `rabbit_feature_flags.erl` with "rabbit_feature_flags: "?

dumbbell (changes requested):
There is a typo in the "rabbit_feature_flags:" prefix in the commit title :-)

dumbbell (approved):
I approve the change to `rabbit_feature_flags.erl`.

The rest of the patch looks fine, but I don't know this code well.

michaelklishin (approved):

PR #10102, opened 2023-12-12
Avoid expensive rabbit_nodes:list_running()

ansd:
for non-mirrored classic queues


michaelklishin:
I've took the liberty of replacing an `if` with a `case` since `if` in Erlang is weird and is very easy to misread (but `case` is effectively identical to any other function language, Rust, and so on).

PR #10104, opened 2023-12-12
Set bazel '--remote_instance_name' to invalidate the Buildbuddy cache

the-mikedavis:


the-mikedavis:
The tests are failing to build with the same error message so it looks like this isn't a cache problem

PR #10127, opened 2023-12-13
Unpin Bazel from 6.4.0

the-mikedavis:
This upgrades Bazel to 7.0.0. Previously this caused build failures involving dangling symlinks in cache entries but that error is no longer reproducible.


michaelklishin:
We still have a BEAM file discovery failure of sorts in CLI tools: https://github.com/rabbitmq/rabbitmq-server/actions/runs/7198125309/job/19606855677?pr=10127#step:5:776

the-mikedavis:
It's odd that it's reproducible in the mix version cluster tests but not the main tests anymore. I can reproduce it if I set a different cache key too: https://github.com/rabbitmq/rabbitmq-server/actions/runs/7198437454/job/19607875937

PR #10157, opened 2023-12-15
Test a branch of rules_erlang that does not create the symlinks

pjk25:
that appear to be problematic in bazel 7


pjk25:
Note that this should not be merged until (possibly) an new rules_erlang is released, and it's no longer referenced with a git_override

pjk25:
Closing in favor of #10338

PR #10186, opened 2023-12-20
Timestamp (annotate with a timestamp) all messages automatically

dcorbacho:
## Types of Changes
- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_priority_queue.erl
+++ b/deps/rabbit/src/rabbit_priority_queue.erl
@@ -413,6 +413,8 @@ info(backing_queue_status, #state{bq = BQ, bqss = BQSs}) ->
           end, nothing, BQSs);
 info(head_message_timestamp, #state{bq = BQ, bqss = BQSs}) ->
     find_head_message_timestamp(BQ, BQSs, '');
+info(oldest_message_received_timestamp, #state{bq = BQ, bqss = BQSs}) ->
+    find_oldest_message_received_timestamp(BQ, BQSs);
 info(online, _) ->
     '';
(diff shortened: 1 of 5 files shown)


deps/rabbit/src/rabbit_priority_queue.erl
```diff
@@ -689,6 +691,28 @@ find_head_message_timestamp(BQ, [{_, BQSN} | Rest], Timestamp) ->
...
+    Timestamps =
+        lists:foldl(
+          fun({_, BQSN}, Acc) ->
+                  case oldest_message_received_timestamp(BQ, BQSN) of
+                      '' -> Acc;
+                      Ts -> [Ts | Acc]
+                  end
+          end, [], BQs),
+    case Timestamps of
+        [] -> '';
```
michaelklishin, line 705: Should we return `undefined` and not an empty atom when the queue has no messages?

michaelklishin (changes requested):

michaelklishin:
The force-push was a rebase.

deps/rabbit/src/rabbit_priority_queue.erl
dcorbacho: It returns an empty atom to follow the same format as `head_message_timestamp` and any other empty values returned by `rabbit_amqqueue:info`.

michaelklishin: OK, I have suspected it might have been for consistency with certain decisions from years ago.

michaelklishin (approved):

Issue #2640, opened 2020-11-17
Make it possible to specify client_properties using a map


michaelklishin:
Proplists are so 2013.

PR #10256, opened 2023-12-29
[2640] client_properties accepted as either a proplists, or a map

ariel-anieli:
## Proposed Changes

- followed definition of `deps/rabbitmq_stream/src/rabbit_stream_reader.erl`
- did not introduce new variable: so the API stay the same.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #2640)
- [ ] Build system and/or CI

## Checklist

## Further Comments

```erlang
% awk '/prop/ && /binary/' deps/rabbitmq_stream/src/rabbit_stream_reader.erl
         client_properties = #{} :: #{binary() => binary()},
```

### DIFF
--- a/deps/amqp_client/src/amqp_connection.erl
+++ b/deps/amqp_client/src/amqp_connection.erl
@@ -153,12 +153,20 @@ start(AmqpParams) ->
 %% user specified connection name.
 start(AmqpParams, ConnName) when ConnName == undefined; is_binary(ConnName) ->
     ensure_started(),
+    Obfuscate   = fun(Password) -> credentials_obfuscation:encrypt(Password) end,
+    ToPropLists = fun(Props)    -> rabbit_data_coercion:to_proplist(Props) end,
     AmqpParams0 =
         case AmqpParams of
-            #amqp_params_direct{password = Password} ->
-                AmqpParams#amqp_params_direct{password = credentials_obfuscation:encrypt(Password)};
-            #amqp_params_network{password = Password} ->
-                AmqpParams#amqp_params_network{password = credentials_obfuscation:encrypt(Password)}
+            #amqp_params_direct{password = Password, client_properties = Props} ->
+                AmqpParams#amqp_params_direct{
+		  password          = Obfuscate(Password),
+		  client_properties = ToPropLists(Props)
+		 };
+            #amqp_params_network{password = Password, client_properties = Props} ->
+                AmqpParams#amqp_params_network{
+		  password          = Obfuscate(Password),
+		  client_properties = ToPropLists(Props)
+		 }
         end,
     AmqpParams1 =
         case AmqpParams0 of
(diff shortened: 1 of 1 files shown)


deps/amqp_client/src/amqp_connection.erl
```diff
@@ -153,12 +157,20 @@ start(AmqpParams) ->
 %% user specified connection name.
 start(AmqpParams, ConnName) when ConnName == undefined; is_binary(ConnName) ->
     ensure_started(),
+    Obfuscate   = fun(Password) -> credentials_obfuscation:encrypt(Password) end,
```
michaelklishin, line 160: You can reference an Erlang function like so: `fun credentials_obfuscation:encrypt/1`, and assign it to a binding:

``` erl
ObfuscateFn = fun credentials_obfuscation:encrypt/1,
```

deps/amqp_client/src/amqp_connection.erl
```diff
@@ -153,12 +157,20 @@ start(AmqpParams) ->
...
+    ToPropLists = fun(Props)    -> rabbit_data_coercion:to_proplist(Props) end,
```
michaelklishin, line 161: This says `ToPropLists` but in fact it takes a single value (a map or a proplist) and returns a single list. So perhaps the name should reflect that: `ToProplist`.

michaelklishin (changes requested):

deps/amqp_client/src/amqp_connection.erl
ariel-anieli: Clever indeed; thank you.

deps/amqp_client/src/amqp_connection.erl
ariel-anieli: Done.

ikavgo: what is the point of storing function references in variables anyway?

ikavgo: what i'm trying to say - instead of two lines for bootstrapping AmqpParams this has 6 lines which are likely slower (depending on how smart compiler is)

ariel-anieli: > what is the point of storing function references in variables anyway?

For the sake of code formatting, @ikavgo: with `credentials_obfuscation:encrypt/1` & `rabbit_data_coercion:to_proplist/1` into the case, the statements go over eighty characters.

Would you like me to change it? I didn't know there are compiling constraints.

ikavgo: we have enough lines over 80 :-) I think in this case it will be fine too.

ariel-anieli: :slightly_smiling_face: Fine, then; @ikavgo. There you are.

ikavgo (approved):

michaelklishin (approved):

PR #10275, opened 2024-01-04
Streams: two additional Prometheus metrics for connections

markus812498:
## Proposed Changes

This PR adds support for metrics from the `rabbit_stream_consumer_created` and `rabbit_stream_publisher_created` ETS tables to be exposed through prometheus through more user-friendly named metrics endpoints. Based off the conversation in a previous PR: https://github.com/rabbitmq/rabbitmq-server/pull/3043

The change is requested to ensure customers using stream connections can scrape metrics from the prometheus endpoints. 

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

If PR is accepted I think this should also be pushed to the documentation on the rabbitmq-website. Do I create a PR to that repo once this is merged? Or is there another procedure to follow?

Thanks!

### DIFF
--- a/deps/rabbitmq_prometheus/Makefile
+++ b/deps/rabbitmq_prometheus/Makefile
@@ -11,7 +11,7 @@ PROJECT_DESCRIPTION = Prometheus metrics for RabbitMQ
 PROJECT_MOD := rabbit_prometheus_app
 DEPS = accept cowboy rabbit rabbitmq_management_agent prometheus rabbitmq_web_dispatch
 BUILD_DEPS = amqp_client rabbit_common rabbitmq_management
-TEST_DEPS = rabbitmq_ct_helpers rabbitmq_ct_client_helpers eunit_formatters
+TEST_DEPS = rabbitmq_ct_helpers rabbitmq_ct_client_helpers eunit_formatters rabbitmq_stream
 
 EUNIT_OPTS = no_tty, {report, {eunit_progress, [colored, profile]}}
 
--- a/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
+++ b/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
@@ -200,6 +200,11 @@
         {4, undefined, connection_process_reductions_total, counter, "Total number of connection process reductions"}
     ]},
 
+    %% the family name for this metric is stream_consumer_metrics but the real table used for data is rabbit_stream_consumer_created.
+    {stream_consumer_metrics, [
+        {2, undefined, stream_consumer_max_offset_lag, gauge, "Current maximum of offset lag of consumers"}
+    ]},
+
     {connection_metrics, [
         {2, undefined, connection_incoming_packets_total, counter, "Total number of packets received on a connection", recv_cnt},
         {2, undefined, connection_outgoing_packets_total, counter, "Total number of packets sent on a connection", send_cnt},
@@ -578,6 +583,17 @@ get_data(channel_metrics = Table, false, _) ->
      [{Table, [{consumer_count, A1}, {messages_unacknowledged, A2}, {messages_unconfirmed, A3},
                {messages_uncommitted, A4}, {acks_uncommitted, A5}, {prefetch_count, A6},
                {global_prefetch_count, A7}]}];
+get_data(stream_consumer_metrics = MF, false, _) ->
+    Table = rabbit_stream_consumer_created, %% real table name
+    try ets:foldl(fun({_, Props}, OldMax) ->
+                          erlang:max(proplists:get_value(offset_lag, Props, 0), OldMax)
+                  end, 0, Table) of
+        MaxOffsetLag ->
+            [{MF, MaxOffsetLag}]
+    catch error:badarg ->
+            %% rabbitmq_stream plugin is not enabled
+            []
+    end;
 get_data(queue_consumer_count = MF, false, VHostsFilter) ->
     Table = queue_metrics, %% Real table name
     {_, A1} = ets:foldl(fun
@@ -708,6 +724,22 @@ get_data(MF, true, VHostsFilter) when is_map(VHostsFilter), MF == queue_metrics 
               end, [], Table);
 get_data(queue_consumer_count, true, _) ->
     ets:tab2list(queue_metrics);
+get_data(stream_consumer_metrics, true, _) ->
+    Table = rabbit_stream_consumer_created, %% real table name
+    try ets:foldl(fun({{QueueName, _Pid, _SubId}, Props}, Map0) ->
+                          Value = proplists:get_value(offset_lag, Props, 0),
+                          maps:update_with(
+                            QueueName,
+                            fun(OldMax) -> erlang:max(Value, OldMax) end,
+                            Value,
+                            Map0)
+                  end, #{}, Table) of
+        Map1 ->
+            maps:to_list(Map1)
+    catch error:badarg ->
+            %% rabbitmq_stream plugin is not enabled
+            []
+    end;
 get_data(vhost_status, _, _) ->
     [ { #{<<"vhost">> => VHost},
         case rabbit_vhost_sup_sup:is_vhost_alive(VHost) of
--- a/deps/amqp10_common/src/serial_number.erl
+++ b/deps/amqp10_common/src/serial_number.erl
@@ -15,9 +15,8 @@
          diff/2,
          foldl/4]).
 
--ifdef(TEST).
+%% For tests.
 -export([usort/1]).
--endif.
 
 -type serial_number() :: sequence_no().
 -export_type([serial_number/0]).
--- a/deps/rabbitmq_ct_helpers/Makefile
+++ b/deps/rabbitmq_ct_helpers/Makefile
@@ -1,7 +1,7 @@
 PROJECT = rabbitmq_ct_helpers
 PROJECT_DESCRIPTION = Common Test helpers for RabbitMQ
 
-DEPS = rabbit_common proper inet_tcp_proxy meck
+DEPS = rabbit_common amqp10_common rabbitmq_stream_common proper inet_tcp_proxy meck
 LOCAL_DEPS = common_test eunit inets
 #TEST_DEPS = rabbit
 
--- a/deps/rabbitmq_ct_helpers/src/stream_test_utils.erl
+++ b/deps/rabbitmq_ct_helpers/src/stream_test_utils.erl
@@ -1,0 +1,137 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
+%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%
+%% Copyright (c) 2007-2024 Broadcom. All Rights Reserved. The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries. All rights reserved.
+%%
+
+%% There is no open source Erlang RabbitMQ Stream client.
+%% Therefore, we have to build the Stream protocol commands manually.
+
+-module(stream_test_utils).
+
+-compile([export_all, nowarn_export_all]).
+
+-include_lib("amqp10_common/include/amqp10_framing.hrl").
+
+-define(RESPONSE_CODE_OK, 1).
+
+connect(Config, Node) ->
+    StreamPort = rabbit_ct_broker_helpers:get_node_config(Config, Node, tcp_port_stream),
+    {ok, Sock} = gen_tcp:connect("localhost", StreamPort, [{active, false}, {mode, binary}]),
+
+    C0 = rabbit_stream_core:init(0),
+    PeerPropertiesFrame = rabbit_stream_core:frame({request, 1, {peer_properties, #{}}}),
+    ok = gen_tcp:send(Sock, PeerPropertiesFrame),
+    {{response, 1, {peer_properties, _, _}}, C1} = receive_stream_commands(Sock, C0),
+
+    ok = gen_tcp:send(Sock, rabbit_stream_core:frame({request, 1, sasl_handshake})),
+    {{response, _, {sasl_handshake, _, _}}, C2} = receive_stream_commands(Sock, C1),
+    Username = <<"guest">>,
+    Password = <<"guest">>,
+    Null = 0,
+    PlainSasl = <<Null:8, Username/binary, Null:8, Password/binary>>,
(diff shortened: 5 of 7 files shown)


michaelklishin:
This has conflicts with `main` and needs rebasing.

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
```diff
@@ -414,6 +428,8 @@ label(M) when is_map(M) ->
               end, <<>>, M);
 label(#resource{virtual_host = VHost, kind = exchange, name = Name}) ->
     <<"vhost=\"", VHost/binary, "\",exchange=\"", Name/binary, "\"">>;
+label({#resource{virtual_host = VHost, kind = queue, name = Name}, P, _}) when is_pid(P) ->
+      <<"vhost=\"", VHost/binary, "\",queue=\"", Name/binary, "\"","channel=\"",(iolist_to_binary(pid_to_list(P)))/binary, "\"">>;
```
gomoripeti, line 432: recently another PR https://github.com/rabbitmq/rabbitmq-server/pull/9656 was merged which adds escaping to label values. I think you need to rebase to latest main and add escaping to this line as well (the vhost and queue name values)

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
```diff
@@ -694,7 +743,9 @@ accumulate_count_and_sum(Value, {Count, Sum}) ->
 
 empty(T) when T == channel_queue_exchange_metrics; T == channel_process_metrics; T == queue_consumer_count ->
     {T, 0};
-empty(T) when T == connection_coarse_metrics; T == auth_attempt_metrics; T == auth_attempt_detailed_metrics ->
+empty(T) when T == connection_coarse_metrics; T == rabbit_stream_publisher_created;
```
gomoripeti, line 746: similarly rabbit_stream_publisher_created is not needed

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
```diff
@@ -512,6 +528,35 @@ get_data(channel_metrics = Table, false, _, _) ->
      [{Table, [{consumer_count, A1}, {messages_unacknowledged, A2}, {messages_unconfirmed, A3},
                {messages_uncommitted, A4}, {acks_uncommitted, A5}, {prefetch_count, A6},
                {global_prefetch_count, A7}]}];
+get_data(rabbit_stream_publisher_created = Table, false, _, _) ->
```
gomoripeti, line 531: I think there is some rebase or copy-paste issue here. this function clause is not necessary, the next one covers publisher metrics.

gomoripeti (commented):
great work :)

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
markus812498: thanks @gomoripeti I did completely miss this!

gomoripeti:
there are valid test failures. We are working on fixing them...

markus812498:
after @gomoripeti 's push above, I believe this PR is ready for review as all Prometheus related tests passing and conflicts are resolved. (looks like failing tests are related to MQTT_V5)

luos:
Hi! 

This looks very good, thank you for implementing it. 

What do you think of the idea to expose the `segments` count as well for a stream, then the rough stream size could be calculated from `max-segment-size * segments`? (Maybe in a separate PR.)

michaelklishin:
Exposing the current segment count sounds OK to me.

gomoripeti:
force-push was a rebase to latest main. google-github-actions/auth fails in CI

michaelklishin:
CI fails because external PRs do not have access to the necessary secrets. We will run the tests locally.

michaelklishin (changes requested):
I cannot rebase this PR onto `main` without resolving a boatload of conflicts.

Specifically, I have given up on the 7th.

It's probably based on a revision that's a month or so out of date.

michaelklishin:
Indeed, https://github.com/cloudamqp/rabbitmq-server/ still uses `master` as its default branch and https://github.com/cloudamqp/rabbitmq-server/tree/main specifically is from May 15th.

michaelklishin:
Rebasing `cloudamqp/rabbitmq-server@main` onto `rabbitmq/rabbitmq-server@main` is a fast forward.

Then rebasing this PR onto the result will conflict with https://github.com/rabbitmq/rabbitmq-server/pull/11431 but that should be it.

gomoripeti:
I used the github UI button "Rebase" previously, sorry if it caused some trouble.

Now I manually rebased to latest main @ rabbitmq/rabbitmq-server and fixed the conflict with #11431

pad-master82:
Is this PR still active and can it be accepted? If not, can I create a new pull PR for the same purpose? I plan to use stream queues and really need to monitor consumer lag. If this PR is accepted, I am willing to create a new PR to collect this metric (consumer lag) in Datadog.

gomoripeti:
thanks for the ping. I guess this PR accidentally got forgotten by the Core Team. I rebased to latest main.

gomoripeti:
I noticed that the current implementation uses vhost, queue, and connection PID as labels. Instead or in addition to PID it would be better to include publisher_id/subscription_id as label to the per-object metrics. I'll work on this.

gomoripeti:
this is ready for another round of review!

I added publisher_id/subscription_id as label to the per-object metrics, keeping the connection label too, as I am not sure the pub/sub ids are unique per queue or per queue+connection.

@ansd I moved the quasi "erlang stream client" functions to a separate test module so they can be reused, does the change look ok? https://github.com/rabbitmq/rabbitmq-server/pull/10275/commits/9db4317c7e5e3035f18ef4cb229f71576755bfa5

Now `rabbitmq_prometheus` has `rabbitmq_stream` app as a test-dependency. I took the makefile change from https://github.com/rabbitmq/rabbitmq-server/commit/690b830e4355d2749385195d93bbc091a8cd739c#diff-0f6d8f1159e5fd049784ccea8d441d9d98eecf84a7c8b78e009da97daddef298R38-R42. I don't know if bazel is still used, I did not update any bazel files, and I would need some help with that, if a bazel change is necessary.

deps/rabbitmq_stream/test/stream_test_utils.erl
```diff
@@ -0,0 +1,139 @@
+%% FIXME this clause is copied from protocol_interop_SUITE.erl
```
gomoripeti, line 1: please advise on what header should be used in this new file

michaelklishin: The same thing as most others, [like `rabbit`](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit.erl).

```
%% This Source Code Form is subject to the terms of the Mozilla Public
%% License, v. 2.0. If a copy of the MPL was not distributed with this
%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
%%
%% Copyright (c) 2007-2024 Broadcom. All Rights Reserved. The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries. All rights reserved.
%%
```

ansd (commented):
[Prometheus cardinality must be kept low](https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels).
Every Prometheus 101 mentions that Prometheus labels must not be overused.

"Prometheus: Up & Running" mentions:

> Using context such as a user’s email address would be unwise, as they have an unbounded cardinality.

> As a metrics-based system, Prometheus is not suitable for storing event logs or individual events. Nor is it the best choice for high cardinality data, such as email addresses or usernames.

> It is usually obvious that email addresses, customers, and IP addresses are poor choices for label values on cardinality grounds.



This PR adds new time series each `{Stream, Connection, PublisherId}` and `{Stream, Connection, SubscriptionId}`.
```
label({stream_publisher, {#resource{virtual_host = VHost, kind = queue, name = Name}, P, PubId}}) ->
    %% stream_consumer_metrics {queue_id, connection_pid, subscription_id}
    <<"vhost=\"", (escape_label_value(VHost))/binary, "\",",
      "queue=\"", (escape_label_value(Name))/binary, "\",",
      "connection=\"", (iolist_to_binary(pid_to_list(P)))/binary, "\",",
      "publisher_id=\"", (integer_to_binary(PubId))/binary, "\"">>;
label({stream_consumer, {#resource{virtual_host = VHost, kind = queue, name = Name}, P, SubId}}) ->
    %% stream_consumer_metrics {queue_id, connection_pid, subscription_id}
    <<"vhost=\"", (escape_label_value(VHost))/binary, "\",",
      "queue=\"", (escape_label_value(Name))/binary, "\",",
      "connection=\"", (iolist_to_binary(pid_to_list(P)))/binary, "\",",
      "subscription_id=\"", (integer_to_binary(SubId))/binary, "\"">>;
```

From my point of view this cardinality is too high.

I envision a future RabbitMQ to emit "per-object" metrics only for long lived entities.
Such long lived entities generally include quorum queues, streams, non-temporary classic queues (i.e. classic queues which are not exclusive and not auto-delete), and optionally exchanges.
Therefore, in a future RabbitMQ version, I personally would like to see all Prometheus metrics that we currently expose per connection and per channel being removed. This PR does the opposite.

I leave the decision whether to merge this PR to other maintainers (cc @michaelklishin @kjnilsson @acogoluegnes @mkuratczyk), but I personally object this change.

I understand that tracking stream consumer offset lag is important. But we should try to brainstorm for a different solution first.

deps/rabbitmq_prometheus/Makefile
```diff
@@ -26,6 +26,12 @@ endif
 include ../../rabbitmq-components.mk
 include ../../erlang.mk
 
+# We are using stream_test_utils from rabbitmq_stream.
+CT_OPTS += -pa ../rabbitmq_stream/test/
+
+test-build::
+	$(verbose) $(MAKE) -C ../rabbitmq_stream test-dir
```
lhoguin, line 33: Don't do this unless there's no other choice. In this case it's probably better to move the module to rabbitmq_ct_helpers.

ikavgo:
> ```
> 
> ```
> 
> 
>     
>   
> 
> From my point of view this cardinality is too high.

ikavgo:
Huh, anyway. David is right, cardinality going to be too big.

gomoripeti:
thanks for the valuable input. As streams are long lived would per-stream (no connection and pub/sub id labels) metrics be acceptable?
- I realise that consumer offset does not make sense aggregated
- would be good to get some info on offset lag, definitely useful to see if it is zero or not. Sum of offset_lag might not be the best aggregation as it can be higher if there are a lot of consumers. Maybe "Max offset_lag" per-stream and max per node?  (a gauge)
- for published, confirmed and publish errors, the sum per stream sounds useful. (But this counter has the usual problem that if a publisher goes away these counters decrease)
- total number of messages consumed per stream might make sense (although with the non-destructive nature of streams this might be less meaningful) (Also has the problem that if a consumer goes away this counter decreases)

ikavgo:
yeah, I would start with describing user stories around these new metrics. I like max_offset_lag per stream since I can imagine setting an alert on this. The rest I don't know. 

So how they would be used? 

As for metrics granularity, "per-object", etc would it make more sense to have only reasonable aggregates in prometheus and referring to Management UI for granularity when an event occurred?

ansd:
> As streams are long lived would per-stream (no connection and pub/sub id labels) metrics be acceptable?

Exposing per-object Prometheus metrics for streams is acceptable from my point of view, similar to how exposing per-object metrics for quorum queues is acceptable.

gomoripeti:
I think the urge to add high cardinality metrics to the prometheus endpoint comes from the feeling that metrics in the mgmt plugin will be deprecated and removed eventually. (I think it is still not clear how and to what extent this will happen. As a user I really like the graphs with some history on per connection or per queue basis. As a maintainer or contributor it is quite a complexity so I agree it should be replaced somehow)

For use cases I think `stream_publisher_published_total` and `stream_consumer_consumed_total` has the some role as `queue_exchange_messages_published_total` and `channel_messages_delivered_total`: first we can look at global msgs received and delivered to get a sense of node traffic, then drill down to answer which were the busiest streams during a traffic spike (in the past - ie historic data)

lhoguin:
We don't have a clear idea on what we want to do about management metrics at this point so they are unlikely to be removed anytime soon. Prometheus was considered to be an alternative to get the metrics out, but this cardinality issue makes it a bad candidate for this. So we are still at square one in that regard, we want something better than management metrics currently are, but we don't know how to go about it.

michaelklishin:
Removal of management UI metrics is not something we have discussed in many months, not something planned for the foreseeable future, and the last time it was discussed, it was a discussion about undeprecating it.

gomoripeti:
so what I can do is add metrics which are sum per-stream:
- `stream_publisher_published_total`
- `stream_publisher_confirmed_total`
- `stream_publisher_error_messages`
- `stream_consumer_consumed_total`

and max per-stream for
- `stream_consumer_max_offset_lag`

Or if there is uncertainty if the former are useful enough, those can be postponed and only add `stream_consumer_max_offset_lag` for now?

deadtrickster:
I would start with max lag first

gomoripeti:
I updated this PR to only include max_offset_leg metric.
But have to fix the tests (issues with dependencies between apps)

gomoripeti:
`amqp10_common` test fails consistently with the below error.
`serial_number:usort/1` is only exported if compiled with `TEST` macro.
I wonder if the failure is because now that `amqp10_common` is a dependency of `rabbitmq_ct_helpers` it is first compiled without `TEST` and then it is not recompiled with `TEST` defined. Not sure if this is how erlang.mk works and if yes how to workaround this?

```
serial_number_SUITE > test_usort
    #1. {error,
            {undef,
                [{serial_number,usort,[[]],[]},
                 {serial_number_SUITE,test_usort,1,
                     [{file,"serial_number_SUITE.erl"},{line,65}]},
```

lhoguin:
It's not recompiled. You need to remove the test on TEST and instead add a comment just above this function's export that says `%% For tests.`

Functions that are hidden behind TEST in rabbit are often a misuse of test builds that we have noticed and begun correcting.

gomoripeti:
Rebased to latest main. This is ready for another round of review. The failing test cases seem unrelated to me.

michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/pull/12765

michaelklishin:
I have received no objections to this PR from other core team members.

michaelklishin:
This seems safe to backport to `v4.0.x`, too. It's a new metric in a node-local plugin.

Issue #10298, opened 2024-01-09
Get expiration time from authorization backend


acogoluegnes:
Add callback in `rabbit_authz_backend`:

```erlang
-callback expiry_timestamp(AuthUser :: rabbit_types:auth_user()) ->
    integer() | never.
```

Issue #10292, opened 2024-01-08
OAuth 2/JWT: take token expiration into account in stream connections


acogoluegnes:
Stream connections use an in-process cache for permissions. The cache has a size limitation, but it's never cleared. It should be cleared and refreshed, especially when credentials have an expiration date (e.g. JWT token).

Stream connections should be able to detect whether credentials have an expiration date and set a timer to close the connection when the token expires. When updating the secret (`sasl_authenticate` frame), the connection would clean the cache and re-evaluate the permissions for publishers and consumers. It would cancel any existing timer related to expiration and set a new timer based on the new expiration date.

PR #10299, opened 2024-01-09
Add expiry_timestamp/1 callback to authz backend behavior

acogoluegnes:
PR for #10298 (expiry timestamp support in authz backends) and #10292 (take expiry timestamp into account in stream connections).


deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -184,6 +99,9 @@
          tuned/3,
          open/3,
          close_sent/3]).
+%% for unit test
```
ansd, line 102: Instead of using a comment, can we check for the TEST macro instead as done in https://github.com/rabbitmq/rabbitmq-server/blob/6a3ba6210aa831bc8a35deb8c432dbef1ca0decf/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl#L19-L23 ?
This will "codify" the comment and enforce that these functions are only called externally in tests.

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -3265,6 +3245,34 @@ ensure_outstanding_requests_timer(#stream_connection{
 ensure_outstanding_requests_timer(C) ->
     C.
 
+ensure_token_expiry_timer(User, #stream_connection{token_expiry_timer = Timer} = Conn) ->
+    TimerRef =
+    maybe
+        rabbit_log:debug("Checking token expiry"),
+        true ?= rabbit_access_control:permission_cache_can_expire(User),
```
ansd, line 3252: I wonder whether we can remove function `rabbit_access_control:permission_cache_can_expire/1` in future? This function seems to be unnecessary given that we now have `rabbit_access_control:expiry_timestamp/1` which will return `never` if the permission cache cannot expire. Ideally `rabbit_channel` will use the same new approach as done by Stream connections in this PR. This change doesn't have to be part of this PR though. However, ideally we don't add new calls to `rabbit_access_control:permission_cache_can_expire/1` such that we can remove that funtion more easily in the future. WDYT?

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -3265,6 +3245,34 @@ ensure_outstanding_requests_timer(#stream_connection{
...
+        false ->
+            undefined;
+        {error, _} ->
+            undefined
+    end,
+    Cancel = case Timer of
+                 undefined ->
+                     ok;
+                 _ ->
+                     timer:cancel(Timer)
```
ansd, line 3272: Maybe we can use the newer `erlang:cancel_timer/2` API instead.

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
acogoluegnes: Sounds reasonable. Do we try to remove `rabbit_access_control:permission_cache_can_expire/1` in this PR? The sooner, the better.

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
acogoluegnes: I ended up using the timer-related functions in `erlang` instead of `timer`. I used `{async, false}` in `erlang:cancel_timer/2` to be able to test the cancellation. Asynchronous would be better, but I don't think it makes much of a difference. WDYT?

ansd: Thanks, that's perfect, synchronous cancellation (ie. `{async, false}`) is good enough in this rarely executed code path.

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
ansd: It's up to you :) I also think, the sooner the better, yes.

ansd (approved):

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
acogoluegnes: https://github.com/rabbitmq/rabbitmq-server/issues/10382

PR #10331, opened 2024-01-12
Stream coordinator: fixes to automatic membership changes.

kjnilsson:
Various bug fixes to make stream coordinator membership changes more reliable. Previously various errors could happen as well as partially successful attempts where the membership change command may fail but it leaves the new server running.

Also ensure that stream coordinator members are removed as part of the `forget_cluster_node` command.

Also added `rabbitmq-streams coordinator_status` command so that we can get a view of replication and other information in a nice tabular format, same as `rabbitmq-queues quorum_status`

This PR also contains a change to allow `rabbit_nodes:list_member/0` to return `[]` (rather than `[node()]`) when the query to fetch the current members fails. This allows us to detect this scenario in the code and avoid, e.g. shrinking the coordinator cluster incorrectly.

Also adding back a missing clause in the quorum queue tick handler that was accidentally removed in #10364

### DIFF
--- a/deps/rabbit/src/rabbit_db_cluster.erl
+++ b/deps/rabbit/src/rabbit_db_cluster.erl
@@ -295,6 +295,15 @@ members_using_mnesia() ->
     rabbit_mnesia:members().
 
 members_using_khepri() ->
+    %% This function returns the empty list when it encounters an error
+    %% trying to query khepri for it's members. As this function does not
+    %% return ok | error this is the only way for callers to detect this.
+    %% rabbit_mnesia:members/0 however _will_ still return at least the
+    %% current node making it impossible to detect the situation where
+    %% the current cluster members may not be correct. It is unlikely we
+    %% ever reach that as the mnesia cluster file probably always exists.
+    %% For khepri however it is a lot more likely to encounter an error
+    %% so we need to allow callers to be more defensive in this case.
     rabbit_khepri:locally_known_nodes().
 
 -spec disc_members() -> Members when
--- a/deps/rabbit/src/rabbit_db_cluster.erl
+++ b/deps/rabbit/src/rabbit_db_cluster.erl
@@ -295,10 +295,7 @@ members_using_mnesia() ->
     rabbit_mnesia:members().
 
 members_using_khepri() ->
-    case rabbit_khepri:locally_known_nodes() of
-        []      -> [node()];
-        Members -> Members
-    end.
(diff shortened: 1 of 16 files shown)


deps/rabbit/src/rabbit_db_cluster.erl
```diff
@@ -295,10 +295,7 @@ members_using_mnesia() ->
     rabbit_mnesia:members().
 
 members_using_khepri() ->
-    case rabbit_khepri:locally_known_nodes() of
-        []      -> [node()];
-        Members -> Members
-    end.
+    rabbit_khepri:locally_known_nodes().
```
dumbbell, line 298: Could you please add a comment about the difference of behavior/return value compared to Mnesia?

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
```diff
@@ -190,6 +190,13 @@ init_per_group(cluster_size_3_parallel = Group, Config) ->
         _ ->
             init_per_group1(Group, Config)
     end;
+init_per_group(unclustered_size_3_4 = Group, Config) ->
+    case rabbit_ct_helpers:is_mixed_versions() of
```
dumbbell, line 194: Is is incompatible with mixed verion cluster in general, or just incompatible with 3.12.x because it lacks something for the tests?

kjnilsson: it has assertions on behaviour that 3.12 doesn't have such as removal of stream coordinator members as part of `forget_cluster_node`

michaelklishin: Then the condition makes sense. Thanks for clarifying.

dumbbell: Isn’t there another way to check the actual condition you mention, “lack of stream coordinator members removal as part of `forget_cluster_node`”?

Because the `is_mixed_version()` will remain true even after the old version in the test is set to RabbitMQ 15.0.

acogoluegnes (approved):

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
kjnilsson: I managed to make the test mixed compat.

Issue #2572, opened 2016-06-17
rabbit_amqqueue:pid_of/1 clause for default vhost calls


Ayanda-D:
Thanks for rabbitmq/rabbitmq-server#826, useful & already using it :) & sorry to nitpick for a/another `pid_of/1` clause for `pid_of/2` calls with default vhost.

michaelklishin:
An attempt to address this almost introduced a critical regression that would break the message routing path.

I'd prefer to not have this extra helper than take on this risk.

PR #10335, opened 2024-01-14
#2572, added a clause for default vhost in pid_of/1

ariel-anieli:
## Proposed Changes

- added a clause for default vhost in `pid_of/1`
- re-formatting in _rabbit_amqqueue.erl_
- implemented second half of `lookup/1`: clause with input as list was specified, but not done.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #2572)
- [ ] Build system and/or CI

## Checklist

## Further Comments

None.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -1572,19 +1572,25 @@ get_queue_consumer_info(Q, ConsumerInfoKeys) ->
 stat(Q) ->
     rabbit_queue_type:stat(Q).
 
--spec pid_of(amqqueue:amqqueue()) ->
+-spec pid_of(binary() | amqqueue:amqqueue()) ->
           pid() | amqqueue:ra_server_id() | 'none'.
 
-pid_of(Q) -> amqqueue:get_pid(Q).
+pid_of(Queue) when is_binary(Queue) ->
+    DefaultVHost = rabbit_vhost:default_name(),
+    case lookup(Queue, DefaultVHost) of
+	{ok, Q}                -> pid_of(Q);
+	{error, not_found} = E -> E
+    end;
+pid_of(Queue) -> amqqueue:get_pid(Queue).
 
 -spec pid_of(rabbit_types:vhost(), rabbit_misc:resource_name()) ->
           pid() | rabbit_types:error('not_found').
 
-pid_of(VHost, QueueName) ->
-  case lookup(rabbit_misc:r(VHost, queue, QueueName)) of
-    {ok, Q}                -> pid_of(Q);
-    {error, not_found} = E -> E
-  end.
+pid_of(VHost, Queue) ->
+    case lookup(Queue, VHost) of
+	{ok, Q}                -> pid_of(Q);
+	{error, not_found} = E -> E
+    end.
 
 -spec delete_exclusive(qpids(), pid()) -> 'ok'.
 
@@ -2125,7 +2131,7 @@ queue_names(Queues)
     {ok, amqqueue:amqqueue()} | {error, not_found}.
 get_bcc_queue(Q, BCCName) ->
     #resource{virtual_host = VHost} = amqqueue:get_name(Q),
-    lookup(VHost, BCCName).
+    lookup(BCCName, VHost).
 
 is_queue_args_combination_permitted(Q) ->
     Durable = amqqueue:is_durable(Q),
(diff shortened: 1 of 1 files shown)


deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -339,7 +339,9 @@ is_server_named_allowed(Args) ->
             [amqqueue:amqqueue()].
 
 lookup(Name) when is_record(Name, resource) ->
-    rabbit_db_queue:get(Name).
+    rabbit_db_queue:get(Name);
+lookup(Names) when is_list(Names) ->
+    lists:map(fun rabbit_db_queue:get/1, Names).
```
michaelklishin, line 344: For better or worse, Team RabbitMQ has historically used list comprehensions over `lists:map/2` for basic map operations

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1578,11 +1580,13 @@ pid_of(Q) -> amqqueue:get_pid(Q).
...
 
-pid_of(VHost, QueueName) ->
-  case lookup(rabbit_misc:r(VHost, queue, QueueName)) of
+pid_of(VHost, QueueName) when is_binary(VHost) ->
+  case lookup(VHost, QueueName) of
     {ok, Q}                -> pid_of(Q);
     {error, not_found} = E -> E
-  end.
+  end;
+pid_of(default_vhost, QueueName) ->
```
michaelklishin, line 1588: This clause does not make much sense for an Erlang shell user. Please turn it into a `pid_of/1`.

deps/rabbit/src/rabbit_amqqueue.erl
ariel-anieli: :slightly_smiling_face: So be it, @michaelklishin.

deps/rabbit/src/rabbit_amqqueue.erl
ariel-anieli: I have done it; thanks for your feedback.

michaelklishin: My comment was not actually addressed.

`pid_of/2` with an artificial special case first argument does not make much sense.

``` erl
pid_of(QueueName) when is_binary(QueueName) ->
    VHost = rabbit_vhost:default_name(),
    case lookup(VHost, QueueName) of
    	{ok, Q}                -> pid_of(Q);
	    {error, not_found} = E -> E
    end.
```

If the virtual host is not explicitly given, and we only have a queue name, try looking it up in the default virtual host. If we are given a record, then it returns the pid value stored there.

Not very elegant but will break nothing and does not require us coming up with a clever short name for this new alternative to `pid_of/1` which accepts a record or `pid_of/2` which accepts a virtual host name.

michaelklishin (changes requested):

deps/rabbit/src/rabbit_amqqueue.erl
ariel-anieli: Thanks for this explanation, @michaelklishin; I did understand what you meant.

ariel-anieli:
:slightly_smiling_face: Inception-like, but it does compile.

ariel-anieli:
New push is about order of arguments in [`lookup/2`](https://github.com/rabbitmq/rabbitmq-server/blob/4c017a55d53feaf8bd369c58bfd180aeb7541f47/deps/rabbit/src/rabbit_amqqueue.erl#L358). It's now fixed.

michaelklishin:
This PR has enough unrelated changes that I'd prefer to not accept it. Having an extra head of `pid_of/1` is a nice-to-have at best.

This modifies **super sensitive** (performance and otherwise) code paths for no reason related to #2572. In fact, it introduced a critical regression by swapping the argument order in `get_bcc_queue/2`.

While this was caught by Dialyzer, the gain here is not worth the risk.

michaelklishin:
Thank you for taking the time to contribute. I'm afraid #2572 is not a good "new contributor issue" it seemed to be, in fact, I'd prefer to simply close it as a `wontfix`.

ariel-anieli:
I see you're concerned about the changes done on `pid_of/1` & `get_bcc_queue/2`;
Thank you for this thorough feedback, @michaelklishin; I appreciate it.

michaelklishin:
The changes around `pid_of/1` were harmless because it's a function invoked via an Erlang shell or `rabbitmqctl eval` only. Breaking them would be a annoyance at most.

ariel-anieli:
Okay; I understand your point.

PR #10339, opened 2024-01-15
Reduce per message disk overhead

ansd:
Message container annotation keys are stored on disk. By shortening them we save 95 - 58 = 37 bytes per message.
```
1> byte_size(term_to_binary(#{exchange => <<>>, routing_keys => [<<"my-key">>], durable => true, priority => 3, timestamp => 1000})).
95
2> byte_size(term_to_binary(#{x => <<>>, rk => [<<"my-key">>], d => true, p => 3, ts => 1000})).
58
```
This should somewhat reduce disk I/O and disk space.


michaelklishin:
I assume we will be documenting this "UNIX wizard"-ish format internally (that is, in the code) :)

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1111,12 +1112,12 @@ binary_to_msg(#resource{kind = queue,
     %% If exchange or routing_keys annotation isn't present the data most likely came
     %% from the rabbitmq-stream plugin so we'll choose defaults that simulate use
     %% of the direct exchange.
-    Mc = case mc:get_annotation(exchange, Mc0) of
-             undefined -> mc:set_annotation(exchange, <<>>, Mc0);
+    Mc = case mc:get_annotation(?EXCHANGE, Mc0) of
```
kjnilsson, line 1115: Given we have now exchange and routing key apis in `mc` could we do the defaulting in there instead so that it always returns a result ifyou call, e.g. `mc:exchange(Msg)`?

deps/rabbit/src/mc_amqp.erl
```diff
@@ -133,20 +132,6 @@ get_property(timestamp, Msg) ->
         _ ->
             undefined
     end;
-get_property(correlation_id, Msg) ->
```
kjnilsson, line 136: are these not needed?

deps/rabbit/src/mc.erl
```diff
@@ -223,9 +226,29 @@ routing_headers(#?MODULE{protocol = Proto,
...
+
+-spec routing_keys(state()) -> [rabbit_types:routing_key(),...].
+routing_keys(#?MODULE{annotations = #{?ROUTING_KEYS := RoutingKeys}}) ->
+    RoutingKeys;
+routing_keys(BasicMessage) ->
+    mc_compat:get_annotation(?ROUTING_KEYS, BasicMessage).
+
+-spec exchange_and_routing_keys(state()) ->
+    {rabbit_misc:resource_name(), [rabbit_types:routing_key(),...]}.
+exchange_and_routing_keys(#?MODULE{annotations = #{?EXCHANGE := Exchange,
```
kjnilsson, line 243: In the interest of keeping APIs from getting too big it feels a little bit like we could let this function go and just use `mc:exchange/1` and `mc:routing_keys/1`. WDYT?

deps/rabbit/include/mc.hrl
```diff
@@ -20,3 +20,11 @@
 %% "Short strings can carry up to 255 octets of UTF-8 data, but
 %% may not contain binary zero octets." [AMQP 0.9.1 $4.2.5.3]
 -define(IS_SHORTSTR_LEN(B), byte_size(B) < 256).
+
+%% We keep the following atom annotation keys short as they are stored per message on disk.
+-define(EXCHANGE, x).
+-define(ROUTING_KEYS, rk).
+-define(TIMESTAMP, ts).
```
kjnilsson, line 27: i am wondering if we're at risk from define collisions given these names are quite generic. Should we consider prefixing them with, e.g. `ANN_` or `MC_`?

kjnilsson (commented):
Mostly a couple of questions and suggestions to discuss before we commit.

kjnilsson (approved):

kjnilsson (approved):

PR #10348, opened 2024-01-16
Handle more information units (Mi, Gi, Ti), add rabbitmq.conf validation for IU values

### LINKED ISSUE #10310: vm_memory_high_watermark.absolute validator accepts unsupported information units (e.g. Mi and not MB or MiB)
As demonstrated in https://github.com/rabbitmq/rabbitmq-server/discussions/10308,

``` ini
vm_memory_high_watermark.absolute = 768Mi
```

will be accepted by the Cuttlefish (`rabbitmq.conf`) schema and then `rabbitmq-diagnostics status` will fail because it does not know how to format `Mi`.

Ideally the node should fail to start, although accepting `Mi`, `Gi` as reasonable approximations for `MiB`, `GiB` would be an improvement, too.

michaelklishin:
This revisits the information system conversion,
that is, support for suffixes like GiB, GB.

When configuration values like disk_free_limit.absolute, vm_memory_high_watermark.absolute are set, the value can contain an information unit (IU) suffix.

We now support several new suffixes and the meaning a few more changes.

First, the changes:

 * k, K now mean kilobytes and not kibibytes
 * m, M now mean megabytes and not mebibytes
 * g, G now means gigabytes and not gibibytes

This is to match the system used by Kubernetes.
There is no consensus in the industry about how
"k", "m", "g", and similar single letter suffixes
should be treated. Previously it was a power of 2, now a power of 10 to align with a very popular OSS project that explicitly documents what suffixes it supports.

Now, the additions:

 * `T`, `TB` (terabytes) and `Ti`, `TiB` (tebibytes)
 * `P`, `PB` (petabytes) and `Pi`, `PiB` (pebibytes)

Finally, the node will now validate these suffixes at boot time, so an unsupported value will cause
the node to stop with a rabbitmq.conf validation
error.

The message logged will look like this:

````
2024-01-15 22:11:17.829272-05:00 [error] <0.164.0> disk_free_limit.absolute invalid, supported formats: 500MB, 500MiB, 10GB, 10GiB, 2TB, 2TiB, 10000000000
2024-01-15 22:11:17.829376-05:00 [error] <0.164.0> Error preparing configuration in phase validation:
2024-01-15 22:11:17.829387-05:00 [error] <0.164.0>   - disk_free_limit.absolute invalid, supported formats: 500MB, 500MiB, 10GB, 10GiB, 2TB, 2TiB, 10000000000
````

Closes #10310.

## How to Test It

Using the following `rabbitmq.conf` lines:

``` ini
disk_free_limit.absolute = 5LOLWUT

# disk_free_limit.absolute = 5G

# disk_free_limit.absolute = 5GB

# disk_free_limit.absolute = 5Gi

# disk_free_limit.absolute = 5GiB

# disk_free_limit.absolute = 5000000000
```

``` shell
bazel run broker RABBITMQ_ENABLED_PLUGINS=rabbitmq_management RABBITMQ_CONFIG_FILE=/path/to/disk_free_limit.absolute.conf
```

The run `rabbitmq-diagnostics status` against it:

``` shell
./bazel-bin/broker-home/sbin/rabbitmq-diagnostics status
```


michaelklishin:
For context: [K8S docs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory), [#10308](https://github.com/rabbitmq/rabbitmq-server/discussions/10308).

PR #10351, opened 2024-01-16
Add channel limit per node

SimonUnge:
## Proposed Changes

A per node channel limit.

An opt in setting to set a hard max channel limit, per node, regardless of channel max per connection.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments


deps/rabbit/src/rabbit_reader.erl
```diff
@@ -915,11 +915,45 @@ create_channel(Channel,
...
+            {true, Limit, Fmt}
+    end.
+
+is_over_node_channel_limit() ->
+    case rabbit_misc:get_env(rabbit, channel_max_per_node, 0) of
+        0 ->
+            false;
+        NodeLimit ->
+            %% Only fetch this if a limit is set
+            CurrNodeChannels = length(rabbit_channel_tracking:list_on_node(node())),
```
dcorbacho, line 950: I'd suggest to implement a new function in `rabbit_channel_tracking` that gives you the number of channels. `ets:info(Tab, size)` gives you the size of the table without having to list all elements. With many channels, it's going to be a cheaper call

SimonUnge: Ah, smart will do!

SimonUnge: Done @dcorbacho

michaelklishin:
This was a rebase.

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -889,6 +889,9 @@ end}.
 
 {mapping, "channel_max", "rabbit.channel_max", [{datatype, integer}]}.
 
+{mapping, "channel_max_per_node", "rabbit.channel_max_per_node",
+          [{datatype, integer}, {validators, ["non_negative_integer"]}]}.
```
michaelklishin, line 893: This should be a positive integer, setting this value to 0 does not make much sense (unless you want to effectively block AMQP 0-9-1 clients specifically, which I don't think is what we're after here).

SimonUnge: Good point. Actually, 0 here means 'infinite', so I might just change that to be the default for readability.

michaelklishin: Ah, right. Then this validator makes sense. As for the default, we use `cuttlefish:unset/0` as a return value when we want to default in the code to kick in.

SimonUnge: Done, changed 0 -> infinity

michaelklishin: OK, I think it makes more sense than `0`, and for new settings we can adopt this new value. Most users who want no limits probably won't set it in the first place.

michaelklishin (approved):

PR #10353, opened 2024-01-17
Do not contact disconnected nodes in rabbit_nodes:list_running/0

kjnilsson:
As this will force erlang to attempt to set up a distribution connection to the down node. This can take some time, especially in cloud environments.

### DIFF
--- a/deps/rabbit/src/rabbit_node_monitor.erl
+++ b/deps/rabbit/src/rabbit_node_monitor.erl
@@ -168,7 +168,7 @@ notify_node_up() ->
 
 notify_joined_cluster() ->
     NewMember = node(),
-    Nodes = rabbit_nodes:list_running() -- [NewMember],
+    Nodes = alive_rabbit_nodes() -- [NewMember],
     gen_server:abcast(Nodes, ?SERVER,
                       {joined_cluster, node(), rabbit_db_cluster:node_type()}),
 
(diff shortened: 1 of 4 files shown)


dumbbell (approved):

michaelklishin:
This forced push was a rebase.

michaelklishin:
This has reproducible failures in

``` shell
bt //deps/rabbit:cluster_rename_SUITE
```

During the clustering step, one node logs

```
Mnesia: node jessica@localhost thinks it's clustered with node 'rename_cluster_one_by_one-3@localhost', but 'rename_cluster_one_by_one-3@localhost' disagrees
```

So the implicit connection behavior is important in a few more test suites that actively mess with nodes (kill them, shut them down, rename them).

michaelklishin:
To further confirm this, editing out the filtering line on this branch immediately makes the above test suite pass :(

deps/rabbit/src/rabbit_node_monitor.erl
```diff
@@ -1020,11 +1020,13 @@ alive_nodes() -> rabbit_nodes:list_reachable().
...
 alive_nodes(Nodes) -> rabbit_nodes:filter_reachable(Nodes).
 
-alive_rabbit_nodes() -> rabbit_nodes:list_running().
+alive_rabbit_nodes() ->
+    alive_rabbit_nodes(rabbit_nodes:list_members()).
 
 -spec alive_rabbit_nodes([node()]) -> [node()].
 
 alive_rabbit_nodes(Nodes) ->
+    ping_all(),
```
dumbbell, line 1029: Should we limit the ping to `Nodes`?

dumbbell (approved):

PR #10364, opened 2024-01-18
Reduce flakiness of certain Common Test suites

kjnilsson:
A variety of flaky test improvement and smaller bug fixes:

Bug fixes:

* Check that the rabbit app is running on the current node before modifying the stream coordinator cluster. This stops unwanted changes being triggered during shutdown
* `rabbit_nodes:list_running/0` returns `[]` whenever there is a failure so this PR also handles a few cases where that could cause unwanted side effects. This will be changed in a future version of the API.
* The wait command now keeps trying if it reads the empty binary from the file (race condition)


michaelklishin:
Awesome! Just out of curiosity, let's see if this backports to `v3.12.x` ;)

PR #10372, opened 2024-01-19
rabbit: Stop Khepri store and Ra systems in stop/1

dumbbell:
## Why

The Khepri store and the coordinators are Ra servers and thus run under the Ra supervision tree. Therefore we need to explicitly stop them otherwise they will continue to run after `rabbit` was stopped.

Likewise, the Ra systems are under the Ra supervision tree. We also need to stop them.

## How

We stop the Khepri store, then the Ra systems, in the opposite order they were started in during boot.


dumbbell:
Stopping Ra systems and servers with the `rabbit` application has some unfortunate consequences. We discovered that several areas of RabbitMQ willingly or unwillingly rely on the fact the these processes still run after `rabbit` has stopped.

One of them is the `forget_cluster_node` CLI command: changing the Ra cluster membership requires the Ra servers to run.

Ideally, we want to redesign the operations that require a `stop_app`/`start_app` pair of commands around them today to work. We would like to replace that with the maintenance mode and let `rabbit` run durring the operation otherwise. However to achieve, we need to stop using Mnesia.

Therefore this pull request is put on hold until we come back to it after Khepri is the only database we use.

Issue #140, opened 2015-05-04
Consider wrapping delivery tag around at 2^63-1


michaelklishin:
See https://github.com/rabbitmq/rabbitmq-java-client/issues/51. This is a protocol limitation but we may want to consider wrapping the value around.

PR #10379, opened 2024-01-20
#140 - Bound delivery tags to longlong

ariel-anieli:
## Proposed Changes

In [Section 1.1 of AMQP 0.9.1](https://www.rabbitmq.com/resources/specs/amqp-xml-doc0-9-1.pdf
) delivery tags defined as longlong.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #140)
- [ ] Build system and/or CI

## Checklist

## Further Comments

None.

### DIFF
--- a/deps/rabbit/src/rabbit_channel.erl
+++ b/deps/rabbit/src/rabbit_channel.erl
@@ -1991,7 +1991,11 @@ record_sent(Type, QueueType, Tag, AckRequired,
                 false ->
                     UAMQ
             end,
-    State#ch{unacked_message_q = UAMQ1, next_tag = DeliveryTag + 1}.
+    NextTag = case DeliveryTag=<math:pow(2,63)-1 of
+		  true  -> Tag + 1;
+		  false -> 1
+	      end,
+    State#ch{unacked_message_q = UAMQ1, next_tag = NextTag}.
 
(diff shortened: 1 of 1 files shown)


michaelklishin:

Erlang integers on 64-bit architectures are "auto-promoted" up to `2^60 - 1` or something like that. So this will not make any practical difference as far as I can tell.

Moreover, to a client library or application, an explicit reset to 1 will likely be as much of a surprise as an overflowed number.

So there isn't a productive way of addressing #140 without introducing a breaking protocol change that would require updating every client in existent for an unclear gain.

`2^60` as an upper limit for delivery tags sounds acceptable to me. In fact, in https://github.com/rabbitmq/rabbitmq-java-client/issues/51 there is a rough estimate that even at high delivery rates it should take a channel that keeps delivering and incrementing the number for years before you'd this issue.

Applications are usually redeployed more often than that. Equally importantly, we are aware of nodes that have 5, 6, even 8 years of uptime but their clients certainly lose connectivity more often than once every 6 years, therefore you can assume that these counters are naturally reset during connection recovery.

ariel-anieli:
Thanks for this explanation, @michaelklishin: I understand better the context.

PR #10384, opened 2024-01-22
Use read permission for store_offset

### LINKED ISSUE #10383: Use read permission for `store_offset`
Instead of write permission. Storing an offset in a stream is not like adding messages to the stream, so we consider the write permission to be too hard for this case. Read permission, like for consuming, is more appropriate.

The implementation will check there is a subscription for the target stream and fall back to an actual read permission call if there is not (e.g. when the subscription state goes away before `store_offset` is sent when closing a consumer).

acogoluegnes:
Instead of write permission. Storing an offset in a stream is not like adding messages to the stream, so we consider the write permission to be too hard for this case. Read permission, like for consuming, is more appropriate.

The implementation checks there is a subscription for the target stream and falls back to an actual read permission call if there is not (e.g. when the subscription state goes away before store_offset is sent when closing a consumer).

Fixes #10383

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
@@ -2201,29 +2201,25 @@ handle_frame_post_auth(Transport,
             {Connection, State}
     end;
 handle_frame_post_auth(_Transport,
-                       #stream_connection{user = User} = Connection,
+                       #stream_connection{stream_subscriptions = Subscriptions,
+                                          user = User} = Connection0,
                        State,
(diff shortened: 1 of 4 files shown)


ansd (approved):

ansd:
cc @thorhalvor IIRC this is something you were requesting when we talked at RabbitMQ Summit in Berlin :)

thorhalvor:
> cc @thorhalvor IIRC this is something you were requesting when we talked at RabbitMQ Summit in Berlin :)

Ahh! 🙌 perfect! Exactly what I requested and felt was «more correct» authz-wise. Thank you and keep up the good work!

PR #10394, opened 2024-01-23
Fix quorum queue status calculation to ignore non-voters

illotum:
## Proposed Changes

Fix: calculating qq status will ignore non-voters in cluster size calculation.
Add: list voters over HTTP API (via rabbit_quorum_queue:format/1).

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

-  I have mixed feelings about adding RPC for every queue in a potential list of hundreds, but couldn't think of an alternative.

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -1672,8 +1672,7 @@ online(Q) when ?is_amqqueue(Q) ->
                   is_pid(Pid)].
 
 format(Q, Ctx) when ?is_amqqueue(Q) ->
-    %% TODO: this should really just be voters
-    Nodes = lists:sort(get_nodes(Q)),
+    Voters = lists:sort(get_voters(Q)),
     Running = case Ctx of
                   #{running_nodes := Running0} ->
                       Running0;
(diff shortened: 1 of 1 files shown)


deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1783,6 +1784,15 @@ get_nodes(Q) when ?is_amqqueue(Q) ->
     #{nodes := Nodes} = amqqueue:get_type_state(Q),
     Nodes.
 
+get_voters(Q) when ?amqqueue_is_quorum(Q) ->
+    {_, Leader} = amqqueue:get_pid(Q),
+    %% FIXME replace with proper API pending rabbitmq/ra#411
+    case  ra_server_proc:state_query(Leader, voters, 1000) of
```
michaelklishin, line 1790: Timeouts this low are guaranteed to produce false positives in our experience. Please bump this to 5s.

michaelklishin (changes requested):

illotum:
Opening for review. This PR

- Pulls newly minted ra 2.8.0
- Bumps Elixir predicate to 1.17. I keep adding and reverting it on various branches, it's probably time to commit. But LMK if you prefer a standalone PR!
- Exposes voters in quorum queue `format`. I have mixed feelings about adding RPC for every queue in a potential list of hundreds, but couldn't think of an alternative.

michaelklishin:
@illotum can we use `ra_server_proc:local_state_query/3` directly in RabbitMQ and not bump Ra? We will bump it for 3.13 but before https://github.com/rabbitmq/ra/pull/411#issuecomment-1906535489 is addressed, likely cannot adopt a new version immediately.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1672,34 +1672,36 @@ online(Q) when ?is_amqqueue(Q) ->
                   is_pid(Pid)].
 
 format(Q, Ctx) when ?is_amqqueue(Q) ->
```
kjnilsson, line 1674: this function is called for every queue when using the /queues HTTP API, depending on sort columns this could mean _all_ queues in the system. Please compare HTTP API query times with and without this change with 5k quorum queues in a 3 node cluster.

I think we need to find another way of achieving this. Perhaps we need to persist the voting set in the queue record queue type state

kjnilsson (commented):
I think this change will make the /queues HTTP API slower when there are many queues. It will need testing but I can't approve as is.

michaelklishin:
Closing per discussion with @illotum @kjnilsson. Unlike https://github.com/rabbitmq/rabbitmq-server/pull/10304/, this PR needs to be reworked from scratch.

Since this change does not break anything and is arguably a bug fix, we can ship it in a 3.13.x patch in case a revised version does not make it in time for 3.13.0.

PR #10395, opened 2024-01-23
bazel run gazelle

pjk25:
Automated changes created by https://github.com/rabbitmq/rabbitmq-server/actions/runs/7620844786 using the [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action in the Run gazelle (Scheduled) workflow.


michaelklishin:
removes a `rabbitmq_stream` (`rabbitmq_stream_common`) dependency on `rabbit_common` which immediately breaks the build

PR #10409, opened 2024-01-24
Catch ETS error in stream management queries

acogoluegnes:
[Why]
A starting node can receive HTTP requests from a browser. The targeted ETS tables may not have been created yet, so errors and stack traces are logged, making the log file hard to read. There is nothing much to do at this point, so we should avoid unnecessary noise in the log file.

[How]
Errors from ets:select/2 are caught in a try/catch block. In case of error, an info message is logged and the function returns [].

### DIFF
--- a/deps/rabbitmq_stream_management/src/rabbit_stream_mgmt_db.erl
+++ b/deps/rabbitmq_stream_management/src/rabbit_stream_mgmt_db.erl
@@ -112,13 +112,13 @@ augment_entity(?ENTITY_PUBLISHER, {{Q, ConnPid, PubId}, Props}) ->
      | Props].
 
 consumers_by_vhost(VHost) ->
-    ets:select(?TABLE_CONSUMER,
+    ets_select(?TABLE_CONSUMER,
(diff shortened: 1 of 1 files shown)


deps/rabbitmq_stream_management/src/rabbit_stream_mgmt_db.erl
```diff
@@ -177,3 +177,13 @@ format_resource(_, unknown) ->
 format_resource(NameAs,
                 #resource{name = Name, virtual_host = VHost}) ->
     [{NameAs, Name}, {vhost, VHost}].
+
+ets_select(T, Spec) ->
+    try
+        ets:select(T, Spec)
+    catch error:Reason ->
+              %% badarg can occur if the table has no been created yet
+              rabbit_log:info("Error while querying ETS table '~tp': ~tp",
```
michaelklishin, line 186: Should this be a warning? An error sounds too harsh (this is a temporary harmless condition) but this is not an info message per se.

michaelklishin (approved):

PR #10415, opened 2024-01-25
#10330 /api/node/{node} should return 404 and not 500 when the node is not a cluster member

### LINKED ISSUE #10330: HTTP API request to `/api/nodes/{node_name}` where `{node_name}` isn't a valid RabbitMQ node fails with 500 error.
### Describe the bug

HTTP API request to `/api/nodes/{node_name}` where `{node_name}` isn't a valid RabbitMQ node fails with 500 error.

A 404 would be a more appropriate response.

### Reproduction steps

1. `curl -v http://localhost:15672/api/nodes/NOTANODENAME'`
2. observe 500 response code
3.
...

### Expected behavior

404 would be the expected response when a resource does not exist.

### Additional context

https://github.com/rabbitmq/rabbitmq-server/blob/9ff387e47363502d11f597dafb117841604ddacd/deps/rabbitmq_management/src/rabbit_mgmt_wm_node.erl#L74-L78

This function does not handle unknown node names correction, however the intire `resouce_exist` callback impl could be substantially optimised by just checking the the node is in `rabbit_nodes:list_members/0`

ikavgo:
## Proposed Changes

Fix #10330

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbitmq_management/src/rabbit_mgmt_wm_node.erl
+++ b/deps/rabbitmq_management/src/rabbit_mgmt_wm_node.erl
@@ -41,10 +41,11 @@ is_authorized(ReqData, Context) ->
 
 node0(ReqData) ->
     Node = list_to_atom(binary_to_list(rabbit_mgmt_util:id(node, ReqData))),
-    [Data] = node_data(Node, ReqData),
-    augment(ReqData, Node, Data).
+    augment(ReqData, Node, node_data(Node, ReqData)).
 
-augment(ReqData, Node, Data) ->
+augment(_ReqData, _Node, not_found) ->
+    not_found;
+augment(ReqData, Node, [Data]) ->
     lists:foldl(fun (Key, DataN) -> augment(Key, ReqData, Node, DataN) end,
                 Data, [memory, binary]).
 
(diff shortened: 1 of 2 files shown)


kjnilsson (changes requested):
How about just checking if the requested node in `resource_exists/2` is in `rabbit_nodes:list_members/0` ?

ikavgo:
> How about just checking if the requested node in `resource_exists/2` is in `rabbit_nodes:list_members/0` ?

@kjnilsson since there is this pattern of not only checking resource existence but also loading it I chose the less invasive method.

:node/memory for example works differently - it indeed checks node exists first. Well, then it does it second time in to_json. So anyway, I'd really advise against further changes here unless we want to touch :node/memory too for at least some degree of uniformity.

michaelklishin (approved):

michaelklishin:
I agree that this is a more generic solution. Any attempt at generalize this likely would have to worry about these functions, or introduce new helpers and use them in `resource_exists/2` carefully everywhere where a certain type of resource is present in the list of `:param`s.

michaelklishin:
At the same time, there are three modules that use this specific function:

``` erl
     {"/nodes/:node",                                          rabbit_mgmt_wm_node, []},
     {"/nodes/:node/memory",                                   rabbit_mgmt_wm_node_memory, [absolute]},
     {"/nodes/:node/memory/relative",                          rabbit_mgmt_wm_node_memory, [relative]},
     {"/nodes/:node/memory/ets",                               rabbit_mgmt_wm_node_memory_ets, [absolute]},
     {"/nodes/:node/memory/ets/relative",                      rabbit_mgmt_wm_node_memory_ets, [relative]},
     {"/nodes/:node/memory/ets/:filter",                       rabbit_mgmt_wm_node_memory_ets, [absolute]},
     {"/nodes/:node/memory/ets/:filter/relative",              rabbit_mgmt_wm_node_memory_ets, [relative]},
```

I'm giving extracting a new small "non-HTTP handler" module like we have for HSTS, for example.

michaelklishin:
I've combined part of Ilya's work in this PR and @kjnilsson's suggestion in https://github.com/rabbitmq/rabbitmq-server/pull/10417.

PR #10420, opened 2024-01-26
Made rabbit_fifo_client:pick_server/1 pick randomly

ariel-anieli:
## Proposed Changes

* instead of first server, picks randomly among the servers
* replaced case statement in `set_timer/2` with `pick_server/1`.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

None.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo_client.erl
+++ b/deps/rabbit/src/rabbit_fifo_client.erl
@@ -233,7 +233,6 @@ add_delivery_count_header(Msg, Count) ->
             Msg
     end.
 
-
 %% @doc Settle a message. Permanently removes message from the queue.
 %% @param ConsumerTag the tag uniquely identifying the consumer.
 %% @param MsgIds the message ids received with the {@link rabbit_fifo:delivery/0.}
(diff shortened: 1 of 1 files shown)


michaelklishin:
I'm afraid we will not accept a quorum queue (-related) change like this without a good justification.

Consider starting a discussion first when you want to touch an essential part of the codebase.

michaelklishin:
Also, if a change is on the hot code path of publishers and consumers, use PerfTest to prove that the change does not introduce regressions.

A nice little refactoring and duplication removal can result in a 30% throughput drop, we've been there before.

ariel-anieli:
> I'm afraid we will not accept a quorum queue (-related) change like this without a good justification.
> 
> Consider starting a discussion first when you want to touch an essential part of the codebase.

I will do that; thanks for the feedback.

> Also, if a change is on the hot code path of publishers and consumers, use PerfTest to prove that the change does not introduce regressions.
> 
> A nice little refactoring and duplication removal can result in a 30% throughput drop, we've been there before.

I didn't do it, indeed; I will check if the changes come with drawbacks.

ariel-anieli:
Herewith the discussion: https://github.com/rabbitmq/rabbitmq-server/discussions/10435

PR #10437, opened 2024-01-29
Update Cowboy to 2.11

lhoguin:
https://ninenines.eu/articles/cowboy-2.11.0/

I would suggest giving this version a few days of testing after merging since a fair amount of changes went in.


michaelklishin:
After testing things manually for an hour and running four of our HTTP API client test suites I could not spot any observable differences for HTTP clients.

michaelklishin (approved):

PR #10443, opened 2024-01-30
Use local Khepri queries only

dumbbell:
## Why

Ra consistent queries are currently fragile in the sense that the query function may run on a remote node and the function reference or MFA may not be valid on that node:

* A different Erlang compiler may produce difference function references for the same module source code. We observed a difference between Erlang/OTP 25.x and Erlang/OTP 26.x compilers for instance.

* There is no way to be sure that the remote function copy, whether it is described by a function reference of an MFA tuple, is the same as the copy local to the caller. Indeed, the remote node may run a different version after an upgrade to one of the local or remote nodes.

## How

That's why we force local queries for now. This is fine for now, especially that we use Khepri projections in many places and they are local by design.


ansd (approved):
It's nice how that switch happens in a single file!

dcorbacho (approved):

dumbbell:
This patch is currently blocked by a problem at the Bazel level which prevents me from updating Ra to 2.7.3.

Here is the error:
```
ERROR: Error computing the main repository mapping: in module dependency chain <root> -> rabbitmq_ra@2.7.3: module not found in registries: rabbitmq_ra@2.7.3
```

Ra is referenced twice in `MODULE.bazel`:

1.  ```
    bazel_dep(
        name = "rabbitmq_ra",
        version = "2.7.3",
        repo_name = "ra",
    )
    ```

2.  ```
    erlang_package.hex_package(
        name = "ra_alt",
        build_file = "@rabbitmq-server//bazel:BUILD.ra",
        pkg = "ra",
        sha256 = "6f2b912a779f4efa4deea762b65192ed6e87111c7d98cbbe8a29576964739147",
        version = "2.7.3",
    )
    ```

The error occurs after I changed the first entry.

The current workaround is to leave the first entry at 2.7.1. But this certainly incorrect.

dumbbell:
Ra 2.7.3 was published to that Bazel registry and the version pinning works.

Issue #9894, opened 2023-11-07
3.12.8 Shovel plugin crashes after upgrade with existing shovels


gomoripeti:
### Describe the bug

Commit ccc22cb86b01ea44af11986f1311533ce11f4b8 changed the id format of the children in the mirrored supervisor `rabbit_shovel_dyn_worker_sup`. However the child spec of a mirrored supervisor is stored in Mnesia and survives a rolling restart. During an upgrade with existing dynamic shovels the below crash was observed on the first node that is upgraded because of the new code hitting old id format.

```
BOOT FAILED
===========
Error during startup: {error,
                       {rabbitmq_shovel,
                        {{shutdown,
                          {failed_to_start_child,
                           rabbit_shovel_dyn_worker_sup_sup,
                           {'EXIT',
                            {function_clause,
                             [{rabbit_shovel_dyn_worker_sup_sup,id,
                               [<<"shovel1">>],
                               [{file,"rabbit_shovel_dyn_worker_sup_sup.erl"},
                                {line,100}]},
                              {rabbit_shovel_dyn_worker_sup_sup,
                               '-cleanup_specs/0-fun-2-',2,
                               [{file,"rabbit_shovel_dyn_worker_sup_sup.erl"},
                                {line,90}]},
                              {sets,fold_bucket,3,
                               [{file,"sets.erl"},{line,503}]},
                              {sets,fold_seg,4,[{file,"sets.erl"},{line,499}]},
                              {sets,fold_segs,4,
                               [{file,"sets.erl"},{line,495}]},
                              {rabbit_shovel_dyn_worker_sup_sup,
                               cleanup_specs,0,
                               [{file,"rabbit_shovel_dyn_worker_sup_sup.erl"},
                                {line,93}]},
                              {rabbit_shovel_dyn_worker_sup_sup,start_child,
                               2,
                               [{file,"rabbit_shovel_dyn_worker_sup_sup.erl"},
                                {line,42}]},
                              {rabbit_shovel_dyn_worker_sup_sup,
                               '-start_link/0-lc$^0/1-0-',1,
                               [{file,"rabbit_shovel_dyn_worker_sup_sup.erl"},
                                {line,28}]}]}}}},
                         {rabbit_shovel,start,[normal,[]]}}}}
```

For the record on another node:
```
1> mirrored_supervisor:which_children(rabbit_shovel_dyn_worker_sup_sup).
[{{<<"/">>,<<"shovel1">>},
[...]

PR #10472, opened 2024-02-02
Rework `mirrored_supervisor` child ID format

dumbbell:
[TBW]


gomoripeti:
I would love to use the testcase `rolling_upgrade` as a template in the future to add rolling upgrade test cases for other components. But how does it work? How does ct know that the "old" nodes have a different version than the "new" one. Aren't they all use the same version (that is the "current" commit whenever it is run)?

michaelklishin:
@gomoripeti we have a class of tests known as "mixed version tests". The test itself does not know which node is which, it depends on how you form the cluster. Tests can be skipped when run in mixed version mode.

The rest is up to the test author: run the correct assertions, avoid trying to use features that certain versions do not have, and so on.

dumbbell:
> @gomoripeti wrotes:
> I would love to use the testcase `rolling_upgrade` as a template in the future to add rolling upgrade test cases for other components. But how does it work? How does ct know that the "old" nodes have a different version than the "new" one. Aren't they all use the same version (that is the "current" commit whenever it is run)?

common_test doesn't know. It's part of our `rabbitmq_ct_helpers` library. To run tests in mixed-version mode, you can specify the path to a secondary "umbrella" (the term is historic):

```
gmake ct-exchange SECONDARY_UMBRELLA=/path/to/v3.12.x/clone RABBITMQ_FEATURE_FLAGS=
```

`/path/to/v3.12.x/clone` is the root of the Git clone.

If `rabbitmq_ct_helpers` sees this variable set, it will start nodes using both the clone you are running the test from and the secondary umbrella clone in a round-robin fashion:
* node 0 = current clone
* node 1 = secondary umbrella
* node 2 = current clone
* node 3 = secondary umbrella
* …

This new testcase here is built on that assumption.

I recommend that you precompile the secondary umbrella to avoid concurrent builds in `rabbitmq_ct_helpers`. For the context of this testcase:
```
gmake -C /path/to/v3.12.x/clone/deps/rabbitmq_federation test-dist
```

dumbbell:
> Tests can be skipped when run in mixed version mode.

They should return `{skip, …}` after they tried to enable the feature flag(s) they depend on. They should **never** **ever** use `rabbit_ct_helpers:is_mixed_version/1` except if there is a very good documented reason.

gomoripeti:
thank you both for clarifying how the mixed version tests work, very useful.
sorry for sidetracking comments of this PR.

dumbbell:
No problem @gomoripeti! In fact, this prompted me to add a comment in the testcase because indeed, the underlying behavior is not obvious from the testcase itself.

I updated the branch to restore the behavior for the shovel plugin and added a testcase to reproduce what was reported in #9894. I verified manually the new testcase failed when updating from 3.12.7 to 3.12.8.

dumbbell:
I think the patch is ready for more review and testing.

@gomoripeti: Could you please give it a try with both the shovel and federation plugins? The new testcases reproduce the scenarios you shared, but still, I would be more confident if you could give it a go.

gomoripeti:
I tested upgrading from both 3.12.7 and 3.12.12 to this branch, both on Mnesia and after migration to Khpepri did some modifications to shovels and federation. Couldn't find any issues.

What is the exit plan for removing the code that handles the temporary experimental ids for shovel? I'm not sure if this is a concern, but it sounds like a shovel created on 3.12.12 can survive until the end of time with rolling restarts.

michaelklishin:
@gomoripeti right now the plan is to ship 3.13, then we will see what would be the best way forward for 3.12. Likely it will simply support both formats in a simplified manner, that's it.

dumbbell:
I’m investigating a transieng failure in CI that I can reproduce locally where a shovel is not restarted as part of the simulated upgrade in the new testsuite.

dumbbell:
The problem in question is unrelated to this patch, but something in Khepri. Once CI is happy, I’ll mark this pull request as ready to review.

gomoripeti:
> What is the exit plan for removing the code that handles the temporary experimental ids

Yes, I absolutely meant on the long term, and outside of the scope of this PR. I don't think 3.12x and 3.13.x series should do anything more about them.

To be somewhat constructive, a draft idea:
On release series N, when the shovel plugin and `rabbit_shovel_dyn_worker_sup_sup` starts up, it can go through all children and restart (`rabbit_shovel_dyn_worker_sup_sup:stop_child`+`start_child`) those with temporary experimental IDs. Release series N is strictly after 3.13.x. At that point it is guaranteed that even in a mixed version cluster no other node creates temporary IDs (also no need for a feature flag). After an upgrade N - 1 -> N it is guaranteed that a node and sup_sup was restarted at least once, so on release series N + 1 there must be no temporary experimental IDs left and the code that handles them can be removed on N + 1. Going through the children is similar to `cleanup_spec` and most of the time it will find nothing to restart so it should be cheap.

michaelklishin:
@gomoripeti honestly, I think a long term solution should start with replacing the `mirrored_supervisor` we have with a Khepri-based version.

Which means we will have a migration to implement and that we will have a lot of design freedom with the new version.

michaelklishin (approved):
There is some ongoing Khepri-related work that seeks to make the new integration/CT test in this PR non-flaky or much less flaky but otherwise it seems to be good to go.

PR #10477, opened 2024-02-03
Removed explicit module references when they refer to functions in the same module

ariel-anieli:
## Proposed Changes

- 24 files are edited at once; if need be, I am willing to split this commit into many ones
- apart from comments, when called within the same module, replaced `module:function/arity` or `module:type/0`, by `function/arity` or `type/0`
- sef-references found with this command:

```shell
git grep -n : | perl -nE 'print if m;/([\w_]+)\.erl:.*[^\w_]\g1:[\w_]+\(; && !/%/'
```

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

None.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -404,7 +404,7 @@ maybe_rebalance({true, Id}, Type, VhostSpec, QueueSpec) ->
                     [Type, VhostSpec, QueueSpec]),
     Running = rabbit_maintenance:filter_out_drained_nodes_consistent_read(rabbit_nodes:list_running()),
     NumRunning = length(Running),
-    ToRebalance = [Q || Q <- rabbit_amqqueue:list(),
+    ToRebalance = [Q || Q <- list(),
                         filter_per_type(Type, Q),
                         is_replicated(Q),
                         is_match(amqqueue:get_vhost(Q), VhostSpec) andalso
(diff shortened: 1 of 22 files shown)


michaelklishin:

This kind of PRs presents a few serious headaches for us maintainers:

 * When something fails, like `bazel test //deps/rabbit_common:all` on Actions, it's difficult to narrow down what change may be causing this because they are all over the place
 * `git blame` output becomes harder to follow
 * Semi-automated or automated changes like this are risky because even the best IDEs sometimes cannot semantically correctly determine if two matching strings are indeed the same code element

I cannot reproduce the failure on two different machines but this PR does not really address any practical problem. The compiler will figure out that a `module:function/arity` call is local and all the same optimizations should apply.

So I would like to accept this contribution but it's too hard to justify.

This is now the 3rd or so time when your PRs are rejected. I suggest that you look into a recent practical problem to work on. Those are usually much more focussed and easier to evaluate.

michaelklishin:
I will try to port some of these manually to see where things may start breaking in that one `rabbit_common` suite.

michaelklishin:
The majority of changes apply without breaking any test suites: https://github.com/rabbitmq/rabbitmq-server/pull/10481.

ariel-anieli:
Thanks for having taken up this PR, @michaelklishin; your explanation made me understand it was a hassle to review.
I'll look into more recent and more scoped problems; I might ask for clarifications: I've only been contributing for a month.

PR #10487, opened 2024-02-05
4.0: use Ra checkpoints in rabbit_fifo for sub-linear time recovery of QQs on boot

the-mikedavis:
This is the companion PR for https://github.com/rabbitmq/ra/pull/415

`rabbit_fifo` currently has an ad-hoc checkpointing system where it saves `{release_cursor, RaftIdx, State}` effects in-memory periodically and emits them as the release cursor moves up. By building checkpointing into `ra` we can save the checkpoints on disk instead, reducing QQ memory footprint and enabling us to recover even very long queues in logarithmic time (w.r.t. the length of the queue). See https://github.com/rabbitmq/ra/pull/415 for in-depth details about checkpoints.

Connects #8261


the-mikedavis:
This has been folded into https://github.com/rabbitmq/rabbitmq-server/pull/10637

PR #10502, opened 2024-02-06
Streams: Soft remove policy configuration of max_segment_size_bytes

kjnilsson:
This configuration is not guaranteed to be safe to change after a stream has been declared and thus we'll remove the ability to change it after the initial declaration. Users should favour the x- queue arg for this config but it will still be possible to configure it as a policy but it will only be evaluated at declaration time.

This means that if a policy is set for a stream that re-configures the `stream-max-segment-size-bytes` key it will show in the UI as updated but the pre-existing stream will not use the updated configuration.

The key has been removed from the UI but for backwards compatibility it is still settable.

NB: this PR adds a new command `update_config` to the stream coordinator state machine. Strictly speaking this should require a new machine version but we're bypassing that by relying on the feature flag instead which avoids this command being committed before all nodes have the new code version. A new machine version can lower the availability properties during a rolling cluster upgrade so in this case it is preferable to avoid that given the simplicity of the change.

### DIFF
--- a/deps/rabbit/test/rabbit_stream_queue_SUITE.erl
+++ b/deps/rabbit/test/rabbit_stream_queue_SUITE.erl
@@ -249,6 +249,12 @@ init_per_group1(Group, Config) ->
                          Config2, stream_queue),
             case EnableFF of
                 ok ->
+                    if Clustered ->
+                           rabbit_ct_broker_helpers:enable_feature_flag(
+                             Config2, stream_update_config_command);
+                       true ->
+                           ok
+                    end,
                     ok = rabbit_ct_broker_helpers:rpc(
                            Config2, 0, application, set_env,
                            [rabbit, channel_tick_interval, 100]),
--- a/deps/rabbit/src/rabbit_core_ff.erl
+++ b/deps/rabbit/src/rabbit_core_ff.erl
@@ -148,3 +148,11 @@
                          post_enable =>
                          {rabbit_khepri, khepri_db_migration_post_enable}}
      }}).
+
+-rabbit_feature_flag(
+   {stream_update_config_command,
+    #{desc          => "A new internal command that is used to update streams as "
+                        "part of a policy.",
+      stability     => stable,
+      depends_on    => [stream_queue]
+     }}).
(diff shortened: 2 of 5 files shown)


deps/rabbit/test/rabbit_stream_queue_SUITE.erl
```diff
@@ -249,6 +249,12 @@ init_per_group1(Group, Config) ->
                          Config2, stream_queue),
             case EnableFF of
                 ok ->
+                    if Clustered ->
+                           rabbit_ct_broker_helpers:enable_feature_flag(
+                             Config2, stream_update_config_command);
+                       true ->
+                           ok
+                    end,
```
dumbbell, line 257: I don't think you need to verify if it's clustered or not.

However, you need to verify if the feature flag was successfully enabled. If it failed, you can return something like `{skip, "Failed to enable `stream_update_config_command`"}`. Otherwise, the testcase may fail later.

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
```diff
@@ -2364,6 +2370,7 @@ max_age_policy(Config) ->
     rabbit_ct_broker_helpers:rpc(Config, 0, ?MODULE, delete_testcase_queue, [Q]).
 
 update_retention_policy(Config) ->
+    rabbit_ct_broker_helpers:enable_feature_flag(Config, stream_update_config_command),
```
dumbbell, line 2373: Same as above.

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
```diff
@@ -2439,6 +2446,7 @@ max_segment_size_bytes_policy_validation(Config) ->
     ok.
 
 max_segment_size_bytes_policy(Config) ->
+    rabbit_ct_broker_helpers:enable_feature_flag(Config, stream_update_config_command),
```
dumbbell, line 2449: Same as above.

dumbbell (changes requested):

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
kjnilsson: If we enable this for the unclustered tests when they later try to cluster they will now fail (in mixed mode) which is why I don't enable it for all. The tests are written such they pass anyway, we just do things in another way internally.

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
kjnilsson: I will remove these as the flag is enabled in in init_per_group

acogoluegnes (approved):

PR #10512, opened 2024-02-07
Log message in case of stream publisher declaration failure

acogoluegnes:
The "precondition failed" error code is good enough for an application, but it may be useful when working on a client library to have more information about the reason of the failure of the publisher declaration.

This commit adds a warning log message that gives the reason(s) of the failure (publisher ID already taken and/or existing publisher with the same reference for the same stream in the connection).

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
@@ -1747,7 +1747,12 @@ handle_frame_post_auth(Transport,
                                internal_sequence = InternalSequence + 1},
                              State}
                     end;
-                {_, _} ->
+                {PublisherIdTaken, ReferenceTaken} ->
+                    rabbit_log:warning("Error while declaring publisher ~tp for stream '~ts', "
+                                       "with reference '~ts'. ID already taken: ~tp. "
+                                       "Reference already taken: ~tp.",
+                                       [PublisherId, Stream, WriterRef,
+                                        PublisherIdTaken, ReferenceTaken]),
                     response(Transport,
                              Connection0,
                              declare_publisher,


Gsantomaggio:
It looks good:
```
rabbit_node0-1  | 2024-02-08 08:32:20.144951+00:00 [warning] <0.1601.0> Error while declaring publisher for stream '<<"invoices-0">>', with reference '<<>>': ID already taken true, reference already taken false
rabbit_node0-1  | 2024-02-08 08:32:20.177854+00:00 [warning] <0.1601.0> Error while declaring publisher for stream '<<"invoices-0">>', with reference '<<>>': ID already taken true, reference already taken false
rabbit_node0-1  | 2024-02-08 08:32:20.198031+00:00 [warning] <0.1601.0> Error while declaring publisher for stream '<<"invoices-0">>', with reference '<<>>': ID already taken true, reference already taken false
```

Since we are here, can we print the `ID`? else it doesn’t matter.

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -1747,7 +1747,11 @@ handle_frame_post_auth(Transport,
                                internal_sequence = InternalSequence + 1},
                              State}
                     end;
-                {_, _} ->
+                {PublisherIdTaken, ReferenceTaken} ->
+                    rabbit_log:warning("Error while declaring publisher for stream '~tp', "
+                                       "with reference '~tp': ID already taken ~tp, "
+                                       "reference already taken ~tp",
```
michaelklishin, line 1753: Can we make these say "ID already taken: …, reference already taken: …", or perhaps "ID already taken? …, reference already taken? …"

And I agree with Gabriele that we probably should log the ID. Not sure if the reference would be useful to the user or to ourselves. But we can log both for completeness' sake.

acogoluegnes: I pushed a change to add the publisher ID:

```
2024-02-09 09:49:09.789164+01:00 [warning] <0.804.0> Error while declaring publisher 2 for stream '<<"declarePublisherUniqueness-15237b97-6ef2-4323-ab6b-91b950d36b79-s1">>', with reference '<<"ref-1">>'. ID already taken: false. Reference already taken: true.
```

What do you think?

Gsantomaggio: LGTM

acogoluegnes:
Do we want to backport this or keep it in 3.13?

Gsantomaggio:
I would backport to 3.12

acogoluegnes:
Any opinion @michaelklishin?

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
michaelklishin: Wait, surely we can format those binaries as "regular strings"?

michaelklishin:
@acogoluegnes I'd would backport all logging improvements unless we consider them risky. They improve our own quality of live during troubleshooting.

michaelklishin:
Per discussion with @acogoluegnes, we can log binaries and atoms (`undefined`) as `~ts`.

PR #10538, opened 2024-02-12
Fix consumer offset and offset lag calculation

acogoluegnes:
The offset lag for a consumer is the difference between the last committed offset (offset of the first message in the last chunk confirmed by a quorum of stream members) and the current offset of the consumer (offset of the first message in the last chunk dispatched to the consumer).

The calculation is simple for most cases, but it needs to be refined by using more context for edge cases (subscription to a stream that has not messages yet, subscription at the very end of a quiet stream).

Example: subscription at "next" (waiting for new messages) in a quiet stream (no messages published). The previous implementation would return consumer offset = 0 and offset lag = last committed offset, where we would expect to get consumer offset = next offset and offset lag = 0.

This commit fixes the calculation for these edge cases.

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_utils.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_utils.erl
@@ -35,7 +35,9 @@
          filter_spec/1,
          command_versions/0,
          filtering_supported/0,
-         check_super_stream_management_permitted/4]).
+         check_super_stream_management_permitted/4,
+         offset_lag/4,
+         consumer_offset/3]).
 
 -include_lib("rabbit_common/include/rabbit.hrl").
 -include_lib("rabbitmq_stream_common/include/rabbit_stream.hrl").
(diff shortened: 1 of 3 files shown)


deps/rabbitmq_stream/src/rabbit_stream_utils.erl
```diff
@@ -330,3 +332,26 @@ q(VirtualHost, Name) ->
...
+consumer_offset(0, 0, undefined) ->
+    0;
+consumer_offset(0, 0, LastListenerOffset) when LastListenerOffset > 0 ->
+    %% consumer at "next" waiting for messages most likely
+    LastListenerOffset;
+consumer_offset(ConsumerOffsetFromCounter, _, _) ->
+    ConsumerOffsetFromCounter.
+
+-spec offset_lag(CommittedOffset :: integer(),
+                      ConsumerOffsetFromCounter :: integer(),
```
kjnilsson, line 348: indentation

kjnilsson (approved):
1 nitpick

deps/rabbitmq_stream/src/rabbit_stream_utils.erl
```diff
@@ -330,3 +332,26 @@ q(VirtualHost, Name) ->
...
+    %% consumer at "next" waiting for messages most likely
+    LastListenerOffset;
+consumer_offset(ConsumerOffsetFromCounter, _, _) ->
+    ConsumerOffsetFromCounter.
+
+-spec offset_lag(CommittedOffset :: integer(),
+                 ConsumerOffsetFromCounter :: integer(),
+                 MessageConsumed :: non_neg_integer(),
+                 LastListenerOffset :: integer() | undefined) -> integer().
+offset_lag(-1, _, _, _) ->
```
michaelklishin, line 351: If we add a comment to the special case below, should we also have one for this special case of `-1`? Is it a new stream that'd have this committed offset value?

acogoluegnes: Right, this is the case of an empty stream. I will add a comment.

PR #10569, opened 2024-02-16
Configurable writer module

dcorbacho:
Make posible to configure the writer module used by the stream coordinator

## Types of Changes

- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_stream_coordinator.erl
+++ b/deps/rabbit/src/rabbit_stream_coordinator.erl
@@ -1049,7 +1049,7 @@ phase_delete_member(StreamId, #{node := Node} = Arg, Conf) ->
     fun() ->
             case rabbit_nodes:is_member(Node) of
                 true ->
-                    try osiris_server_sup:delete_child(Node, Conf) of
+                    try osiris:delete_member(Node, Conf) of
                         ok ->
                             rabbit_log:info("~ts: Member deleted for ~ts : on node ~ts",
                                             [?MODULE, StreamId, Node]),
@@ -1071,10 +1071,9 @@ phase_delete_member(StreamId, #{node := Node} = Arg, Conf) ->
             end
     end.
 
-phase_stop_member(StreamId, #{node := Node,
-                              epoch := Epoch} = Arg0, Conf) ->
+phase_stop_member(StreamId, #{node := Node, epoch := Epoch} = Arg0, Conf) ->
     fun() ->
-            try osiris_server_sup:stop_child(Node, StreamId) of
+            try osiris:stop_member(Node, Conf) of
                 ok ->
                     %% get tail
                     try get_replica_tail(Node, Conf) of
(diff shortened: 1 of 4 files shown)


deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -1900,7 +1901,7 @@ eval_replica(#{index := Idx} = Meta,
     case Current of
         undefined ->
             Args = Meta#{epoch => Epoch, node => Node},
-            Actions = [{aux, {delete_member, StreamId, Args, Conf}} |
+            Actions = [{aux, {delete_member, StreamId, Args, make_writer_conf(Conf)}} |
```
kjnilsson, line 1904: because of the different supervisor setting using the write conf for deletes means a running replica may not actually be deleted as it would try the wrong supervisor. I think we need to run all members under the osiris_server_sup

deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -2013,12 +2014,26 @@ make_replica_conf(LeaderPid,
 
 make_writer_conf(Node, #stream{epoch = Epoch,
```
kjnilsson, line 2015: This function is used inside the state machine which means we cannot reach outside the scope of the machine and into the environment as this is not deteministic.

deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -1109,9 +1110,9 @@ phase_stop_member(StreamId, #{node := Node,
     end.
 
 phase_start_writer(StreamId, #{epoch := Epoch,
-                               node := Node} = Args0, Conf) ->
+                               node := Node} = Args0, #{writer_mod := Mod} = Conf) ->
     fun() ->
-            try osiris_writer:start(Conf) of
+            try Mod:start(Conf) of
```
kjnilsson, line 1115: We need a behaviour module with a defined API for this as plain dynamic module calls are a pain to maintain and keep safe.

deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -1109,9 +1110,9 @@ phase_stop_member(StreamId, #{node := Node,
     end.
 
 phase_start_writer(StreamId, #{epoch := Epoch,
-                               node := Node} = Args0, Conf) ->
+                               node := Node} = Args0, #{writer_mod := Mod} = Conf) ->
```
kjnilsson, line 1113: Are all configs guaranteed to what the writer_mod key?

kjnilsson (changes requested):
Ok there are some issues with this PR that I will help to address. Doing a quick PR here to avoid a hasty merge.

kjnilsson (approved):

PR #10628, opened 2024-02-27
Use binary concatenation for classic queue segment filenames

mkuratczyk:
I noticed that in some scenarios with intense CQ workloads, we spend a few percent of the CPU time in `filename:join`. This PR simplifies how classic queues generate the file paths. I was unable to see measurable performance improvement because of this change (actual disk operations probably overshadow any CPU work here), but it reduces unnecessary CPU work and the amount of garbage generated, so it's still a reasonable change.

not to be backported

### DIFF
--- a/deps/rabbit/src/rabbit_classic_queue_index_v2.erl
+++ b/deps/rabbit/src/rabbit_classic_queue_index_v2.erl
@@ -196,9 +196,10 @@ init_for_conversion(#resource{ virtual_host = VHost } = Name, OnSyncFun, OnSyncM
 
 init1(Name, Dir, OnSyncFun, OnSyncMsgFun) ->
     ensure_queue_name_stub_file(Name, Dir),
+    DirBin = rabbit_file:filename_to_binary(Dir),
     #qi{
         queue_name = Name,
-        dir = rabbit_file:filename_to_binary(Dir),
+        dir = << DirBin/binary, "/" >>,
         on_sync = OnSyncFun,
         on_sync_msg = OnSyncMsgFun
     }.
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_classic_queue_index_v2.erl
```diff
@@ -1277,8 +1278,8 @@ queue_name_to_dir_name(#resource { kind = queue,
     rabbit_misc:format("~.36B", [Num]).
 
 segment_file(Segment, #qi{ dir = Dir }) ->
-    filename:join(rabbit_file:binary_to_filename(Dir),
-                  integer_to_list(Segment) ++ ?SEGMENT_EXTENSION).
+    N = integer_to_binary(Segment),
```
kjnilsson, line 1281: There is no need to convert Dir to a list, `filename:join/2` can take a binary as an arg and as soon as one of the args is a binary the output will also be binary.

lhoguin (approved):

deps/rabbit/src/rabbit_classic_queue_index_v2.erl
lhoguin: The original code likely was a copy paste from older rabbit code where it used to be necessary.

deps/rabbit/src/rabbit_classic_queue_index_v2.erl
```diff
@@ -1277,8 +1278,8 @@ queue_name_to_dir_name(#resource { kind = queue,
...
+    <<Dir/binary, N/binary, <<?SEGMENT_EXTENSION>>/binary>>.
```
lhoguin, line 1282: You can just do `<<Dir/binary, N/binary, ?SEGMENT_EXTENSION>>` it will be just as clear.

lhoguin (approved):

PR #10637, opened 2024-02-28
Quorum queues v4

kjnilsson:
Quorum queues v4

#8261 

Tasks:

- [x] Fair share high/low priorities using https://github.com/kjnilsson/hiloq/)
- [ ] ~Consumer timeouts implemented in the queue (rather than in the channel as currently done).~
- [x] SAC consumers honour consumer priorities 
- [x] Credited consumer [refactoring](https://github.com/rabbitmq/rabbitmq-server/blob/1e1a6848e1f2da46bc50c3995cdda44f08cb399e/deps/rabbit/src/rabbit_fifo.erl#L342-L346) to meet AMQP requirements.
- [x] Use checkpoints feature. https://github.com/rabbitmq/rabbitmq-server/pull/10487
- [x] Consumer cancel option that immediately removes consumer and returns all pending messages.
- [x] More compact commands of the most common commands such as `enqueue`, `settle` and `credit`
- [x] Track number of requeues in addition to delivery-count as this would allow us to calculate a correct delivery-count 
- [x] support the "modified" AMQP 1.0 outcome better. Both AMQP 1.0 and MQTT 5.0 can supply annotations along with the message outcome / rejection. If we add support for these in the queue type they could potentially be used to e.g. control dead-letter routing by supplying additional x- headers (message annotations) to a headers exchange or we could even allow routing key overrides by defining a new annotation: `x-opt-dead-letter-routing-key` that would be used for dead letter routing instead of the dead letter configuration. This would allow clients, e.g. to specify a different retry queue a message for re-processing that may have a different ttl setting. (short, medium, long)


kjnilsson:
Consumer timeouts postponed.

deps/rabbit/src/rabbit_fifo_client.erl
```diff
@@ -320,29 +328,45 @@ discard(ConsumerTag, [_|_] = MsgIds,
...
+                           {auto, CreditMode}
+                   end;
+               false ->
+                   case CreditMode of
+                       {credited, _} ->
+                           {auto, 0, credited};
+                       {simple_prefetch, 0} ->
+                           {auto, ?UNLIMITED_PREFETCH_COUNT, simple_prefetch};
+                       {simple_prefetch, Num} ->
+                           {auto, Num, simple_prefetch}
```
ansd, line 358: It seems always life `auto` is used. Can this tuple element then be omitted? In other words, where is life `once` used in the checkout spec?

deps/rabbit/src/rabbit_fifo_client.erl
```diff
@@ -320,29 +328,45 @@ discard(ConsumerTag, [_|_] = MsgIds,
...
+                       {simple_prefetch, 0} ->
+                           {auto, {simple_prefetch,
+                                   ?UNLIMITED_PREFETCH_COUNT}};
+                       _ ->
+                           {auto, CreditMode}
+                   end;
+               false ->
+                   case CreditMode of
+                       {credited, _} ->
+                           {auto, 0, credited};
```
ansd, line 354: When I enable feature flag `credit_api_v2` but leave feature flag `quorum_queues_v2` disabled and attach an AMQP 1.0 receiver to a quorum queue, the session proc crashes:
```
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>   crasher:
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>     initial call: rabbit_amqp_session:init/1
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>     pid: <0.705.0>
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>     registered_name: []
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>     exception error: no function clause matching
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>                      rabbit_amqp_session:'-handle_queue_actions/2-fun-1-'({send_credit_reply,
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>                                                                            0},
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>                                                                           {state,
...
(rabbit_amqp_session.erl, line 1940)
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>       in function  lists:foldl/3 (lists.erl, line 1594)
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>       in call from rabbit_amqp_session:handle_cast/2 (rabbit_amqp_session.erl, line 476)
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>       in call from gen_server:try_handle_cast/3 (gen_server.erl, line 1121)
2024-07-17 12:42:15.626243+00:00 [error] <0.705.0>       in call from gen_server:handle_msg/6 (gen_server.erl, line 1183)
```

This is because the session thinks it uses credit API v2 although quorum queue client and quorum queue server decide to use credit API v1 under the hood.

To solve this, we could either check whether `quorum_queues_v4` is enabled in the session proc and only then use credit API v2.
An alternative solution is to merge both feature flags `credit_api_v2` and `quorum_queues_v4` into a single feature flag `queue_api_4_0` which includes all new classic queue client to classic queue server and quorum queue client to quorum queue server interactions introduced in RabbitMQ 4.0
(Feature flag `credit_api_v2` is arguably misnamed because it includes already both credit API v2 and cancel API v2.)

ansd (changes requested):

deps/rabbit/src/rabbit_fifo_client.erl
ansd: Fixed by merging all feature flags introduced in RabbitMQ 4.0.0 into a single feature flag `rabbitmq_4.0.0`.

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -604,293 +601,189 @@ apply(Meta, {nodeup, Node}, #?MODULE{consumers = Cons0,
...
+    Returns = lqueue:from_list(
+                lists:map(fun (?MSG(I, #{delivery_count := DC} = H)) ->
+                                  ?MSG(I, H#{acquired_count => DC});
+                              (Msg) ->
+                                  Msg
+                          end, Returns0)),
+
+    Messages = rabbit_fifo_q:from_lqueue(Messages0),
+    #?STATE{cfg = rabbit_fifo_v3:get_field(cfg, StateV3),
+            messages = Messages,
```
ansd, line 676: v3 can set a `delivery_count` in `messages` when `#requeue{}` command is applied.
Looks like we therefore need to convert `delivery_count` to `acquired_count` in `messages` as well?

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -604,293 +601,189 @@ apply(Meta, {nodeup, Node}, #?MODULE{consumers = Cons0,
...
+convert_v3_to_v4(#{} = _Meta, StateV3) ->
+    %% TODO: consider emitting release cursors as checkpoints
+    Messages0 = rabbit_fifo_v3:get_field(messages, StateV3),
+    Returns0 = lqueue:to_list(rabbit_fifo_v3:get_field(returns, StateV3)),
+    Consumers0 = rabbit_fifo_v3:get_field(consumers, StateV3),
+    Consumers = maps:map(
+                  fun (_, #consumer{checked_out = Ch0} = C) ->
+                          Ch = maps:map(
+                                 fun (_, ?MSG(I, #{delivery_count := DC} = H)) ->
+                                         ?MSG(I, H#{acquired_count => DC});
```
ansd, line 659: Not too important, but I think deleting `delivery_count` makes sense (in addition to adding `acquired_count`) because then the new `delivery_count` is only ever set when the AMQP 1.0 client rejects or modifies the message (which is the new v4 behaviour).

deps/rabbit/src/rabbit_fifo.erl
kjnilsson: I don't think we need to do this as requeue can be considered a new entry point at which we start counting acquisitions from.

ansd (approved):

PR #10656, opened 2024-03-04
4.x: remove availability of CQv1

lhoguin:
Most of the v1 code is still around as it is needed for conversion to v2. It will be removed at a later time when conversion is no longer supported.

A few notes:

* `classic_queue_default_version`: currently it remains but is unused. We validate the value and if it selects v1 we abort with an informative error
* Both `x-queue-version` and `x-queue-mode` remain. We can't really remove them. They are ignored
* Most of the v1 code is still around as it is needed for the automatic conversion

## Type of Changes


lhoguin:
I do not understand why Peer Discovery AWS Integration Test is failing. It is likely related to the changes in this PR as I haven't seen this failure consistently elsewhere. I could use some help in figuring out the cause. The PR is otherwise ready for review. Thank you!

Issue #2544, opened 2015-06-11
A way to configure durability of the queues used by exchange federation


elalonde:
We are investigating the use of federation of an exchange shared throughout an upstream rabbitmq cluster. We observe that when we terminate the RabbitMQ instance that represents the location of the durable queue bound to the federated exchange, the downstream node cannot continue to federate even after it attempts to connect to a different URI in the upstream list. The logs contain messages such as:

```
Federation exchange '(redacted)' in vhost '(redacted)' did not connect to exchange '(redacted)' in vhost '(redacted)' on amqp://(redacted):7777
{{shutdown,{server_initiated_close,404,
     <<"NOT_FOUND - home node '(redacted)' of durable queue 'federation: (redacted)' in vhost '(redacted)' is down or inaccessible">>}},
```

From the documentation I gather that federation is implemented by the creation of a queue bound to the exchange, and messages are copied to that queue so they can be sent downstream. Documentation makes it clear that this queue resides on exactly one node, by default, so what we are seeing is understandable. 

We would like to configure federation such that the queue is not durable, and the behavior is that when the node hosting the queue terminates (or downstream disconnects), the queue is dynamically re-created and re-bound to the exchange on whatever node the downstream connects to at a later time. In essence, on-demand federation where queues are only bound to upstream exchanges if there is a downstream listening.

As you can likely tell, we have no need for message durability in the face of intermittent connectivity. We just want downstream message delivery to resume when the downstream selects another node in the URI list.

PR #10663, opened 2024-03-04
Add queue type setting for exchange federation

mkuratczyk:
This allows using a quorum queue as the upstream buffer for exchange federation.
While at it, `x-ha-policy` parameter was removed - it was unused since RabbitMQ 3.0.

Closes https://github.com/rabbitmq/rabbitmq-server/issues/4683
Closes https://github.com/rabbitmq/rabbitmq-server/issues/2544


deps/rabbitmq_federation/src/rabbit_federation_exchange_link.erl
```diff
@@ -504,15 +504,16 @@ consume_from_upstream_queue(
...
     #upstream_params{x_or_q = X,
                      params = Params} = UParams,
     Q = upstream_queue_name(name(X), vhost(Params), DownXName),
     Args = [A || {_K, _T, V} = A
                      <- [{<<"x-expires">>,          long,    Expiry},
                          {<<"x-message-ttl">>,      long,    TTL},
-                         {<<"x-ha-policy">>,        longstr, HA},
-                         {<<"x-internal-purpose">>, longstr, <<"federation">>}],
+                         {<<"x-internal-purpose">>, longstr, <<"federation">>},
+                         {<<"x-queue-type">>,       longstr, atom_to_binary(QueueType)}
```
michaelklishin, line 515: If `queue-type` is not set, this will set it to `<<"undefined">>` or something like that. Should we be more defensive here?

mkuratczyk: The default (`classic`) is set here: https://github.com/rabbitmq/rabbitmq-server/blob/exchange-federation-qq/deps/rabbitmq_federation/src/rabbit_federation_upstream.erl#L139.

michaelklishin (approved):

alphamonkey79:
Hello

Looking to implement this via our RabbitMQ policy definition(s) in Ansible code but running into issue...

'queue_type'              => 'quorum', 
[{<<"queue_type">>,<<"quorum">>}] are not recognised policy settings

'queue-type'              => 'quorum',
[{<<"queue-type">>,<<"quorum">>}] are not recognised policy settings

Not sure what I could be missing?

Tim

PR #10670, opened 2024-03-05
WIP Expose virtual host metadata to Prometheus

### LINKED ISSUE #10660: Add more useful data into rabbitmq_cluster_vhost_status metric

When a new virtual host is being created, a user can provide not only name, but also description and tags. Description and tags later are not visible in the GUI, but can be retrieved via `/api/definitions/{vhost}` API endpoint:
```{
    "name": "test",
    "description": "testDescription",
    "tags": [
        "testTags"
    ],
    "default_queue_type": "quorum",
    "metadata": {
        "description": "testDescription",
        "tags": [
            "testTags"
        ],
        "default_queue_type": "quorum"
    },
    "tracing": false,
    "cluster_state": {
        "rabbit@server.local": "running"
    }
}
```
However, the same data is not available using RabbitMQ monitoring tools like Prometheus. This limitation allows filtering only by vhost name. Sometimes that is a big disadvantageous when creating Prometheus alerting rules because filtering is only available by vhost name and can't be achieved by vhost tags or description.

Get RabbitMQ virtual host **tags** and **description** data when querying **rabbitmq_cluster_vhost_status** metric via Prometheus.

_No response_

### Additional context

_No response_

michaelklishin:
Closes #10660.


michaelklishin:
This has not been forgotten but also not a priority this month, and I am still fighting a test failure that is induced by this change, for which I still do not have an explanation.

PR #10688, opened 2024-03-06
Bump initial maximum frame size

ansd:
in AMQP 1.0 to allow large tokens to be sent in the SASL exchange. We choose the value 8192 defined in
https://docs.oasis-open.org/amqp/amqp-cbs/v1.0/csd01/amqp-cbs-v1.0-csd01.html#_Toc67999915


MarcialRosales (approved):
I have tried these changes and it addresses the issue I was having while using oauth2 token to authenticate.

PR #10720, opened 2024-03-12
rabbit_env: Correctly pass `TakeFromRemoteNode` argument to `update_context/3`

dumbbell:
## Why

The given `TakeFromRemoteNode` argument was unpacked and another tuple
was created to pass to `update_context/3`. However, the constructed
tuple would be:  

```
{{Node, Timeout}, Timeout}
```

… which is incorrect.


dumbbell:
@Mergifyio backport v3.13.x

PR #10736, opened 2024-03-14
rules_pkg 0.9.1 -> 0.10.1

mkuratczyk:
0.9.1 is not compatible with Python 3.12


WORKSPACE
```diff
@@ -7,8 +7,8 @@ http_archive(
     name = "rules_pkg",
     sha256 = "8f9ee2dc10c1ae514ee599a8b42ed99fa262b757058f65ad3c384289ff70c4b8",
```
pjk25, line 8: Need to update the hash, otherwise we'll get an incorrect cache hit

PR #10738, opened 2024-03-14
Update Cowboy to 2.12.0

lhoguin:
And Cowlib to 2.13.0 as a dependency of Cowboy 2.12.0.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -189,8 +189,8 @@ erlang_package.hex_package(
     patch_cmds = [
         "rm ebin/cowboy.app",
     ],
-    sha256 = "0fa395437f1b0e104e0e00999f39d2ac5f4082ac5049b67a5b6d56ecc31b1403",
-    version = "2.11.0",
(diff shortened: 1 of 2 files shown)


lhoguin:
PR isn't ready yet as it's passing despite me not publishing to hex yet...

lhoguin:
I have not updated the `sha256` but after forcing a Web-MQTT test to run it didn't give any errors. So not sure what I should put for the sha or if it's OK like this, or if the dependency update failed somehow.

MODULE.bazel
```diff
@@ -190,7 +190,7 @@ erlang_package.hex_package(
         "rm ebin/cowboy.app",
     ],
     sha256 = "0fa395437f1b0e104e0e00999f39d2ac5f4082ac5049b67a5b6d56ecc31b1403",
```
pjk25, line 192: If the hash is unchanged, this will produce an incorrect cache it. Comment out the line, build with `bazel build @cowboy//:erlang_app` and the new sha should be printed. Then you can update it.

PR #10742, opened 2024-03-14
Update durable queues outside of Khepri transactions

the-mikedavis:
@mkuratczyk found a bug that happens when the `khepri_db` feature flag is enabled and a vhost fails to recover. Specifically this block: https://github.com/rabbitmq/rabbitmq-server/blob/e8e50c8fae198d723774d58c16218b4a1dff5fb1/deps/rabbit/src/rabbit_vhost_process.erl#L49-L53

The `rabbit_amqqueue:mark_local_durable_queues_stopped/1` call fails when Khepri is enabled because it tries to create a transaction function which Khepri disallows - it finds queues which are not alive (via an RPC call) and marks them as stopped. That's unsafe to do in a Khepri transaction since it would be executed on each node and the value of `node()` would change, and side effects like RPC calls would be repeated by each Khepri cluster member. So that function currently errors. (For example start a 3.13.0 broker, enable `khepri_db` and execute `rabbit_amqqueue:mark_local_durable_queues_stopped(<<"/">>)` - this will error out).

We can work around this by using Khepri's advanced API to get the version number in the database of each queue, filter and update each queue and then use a transaction to apply the updates. This way we get transaction-like behavior of atomically updating all queues or none without the restrictions of Khepri transaction functions for FilterFun and UpdateFun.


mkuratczyk:
Thanks, I can no longer reproduce the problem.

PR #10764, opened 2024-03-18
rabbit_env_SUITE: Wait for application env. to be ready

dumbbell:
## Why

If the testcase run between the time the node is started and the time the `rabbit` application environment is loaded, it will fail.

## How

Instead of waiting for the node to be reachable only, we also wait for the `rabbit` application environment to be filled.


dumbbell:
@Mergifyio backport v3.12.x

Issue #10901, opened 2024-04-02
Unhandled function clause exception when retrieving queue for nonexistent virtual host


LoisSotoLopez:
### Describe the bug

Retrieving messages on a queue through `rabbit_mgmt_wm_queue_get.erl` for an invalid virtual host results on an unhandled `function_clause` exception.

This issue seems to have been introduced by commit https://github.com/rabbitmq/rabbitmq-server/commit/439c2b4a6a823462e76c014e4e02c94fb9d9b598 , more precisely by [this change ](https://github.com/rabbitmq/rabbitmq-server/commit/439c2b4a6a823462e76c014e4e02c94fb9d9b598#diff-dc706d775169364ab615292a0fab82e1b2f753061a7ecbef9c37a652b651bb2cL284).

```diff
-    mnesia:async_dirty(
-      fun () ->
-              case VHost of
-                  '_' -> ok;
-                  _   -> rabbit_vhost:assert(VHost)
-              end,
-              Match = #runtime_parameters{key = {VHost, Component, '_'},
-                                          _   = '_'},
-              [p(P) || #runtime_parameters{key = {_VHost, Comp, _Name}} = P <-
-                           mnesia:match_object(?TABLE, Match, read),
-                       Comp =/= <<"policy">> orelse Component =:= <<"policy">>]
-      end).
+    [p(P) || #runtime_parameters{key = {_VHost, Comp, _Name}} = P <-
+             rabbit_db_rtparams:get_all(VHost, Component),
+             Comp =/= <<"policy">> orelse Component =:= <<"policy">>].
```

While on the previous version of the file the `rabbit_vhost:assert/1` was called before searching for the vhost data on the Mnesia table, the current version calls [`rabbit_db_rtparams:get_all/2`](https://github.com/rabbitmq/rabbitmq-server/commit/439c2b4a6a823462e76c014e4e02c94fb9d9b598#diff-e85ed46932f74c38f11e4d775f9fdbc9d297c8cc8ce265a8339fceabadca9f8cR178-R194). As seen below, that function ends up calling to `rabbit_vhost:assert/1` on the vhost too (at `get_all_in_mnesia/2`) but only if the `VHostName` parameter passed to `get_all/2` is either a `binary()` or the `'_'` atom.

```diff
+get_all(VHostName, Comp)
+  when (is_binary(VHostName) orelse VHostName =:= '_') andalso
+       (is_binary(Comp) orelse Comp =:= '_') ->
[...]

PR #10908, opened 2024-04-03
Handle non-existent virtual host in several HTTP API endpoints

LoisSotoLopez:
Issue #10901

## Proposed Changes

When handling the retrieval of messages through the API at `rabbit_mgmt_wm_queue_get.erl`, substitute the parsing of the virtual host name for an assertion that might raise an exception. Raising an exception instead of propagating a `not_found` atom (current behaviour) prevents a function clause on further steps of the data flow. If the vhost does not exist we raise an exception that leads to a 404 HTTP error code being returned by cowboy, with a `Reason` describing the inexistence of the provided vhost, instead of the 500 HTTP error code that might be returned given the current code.

The reason for the changes are detailed on the issue description.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #10901)
- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbitmq_management/src/rabbit_mgmt_util.erl
+++ b/deps/rabbitmq_management/src/rabbit_mgmt_util.erl
@@ -10,7 +10,7 @@
 %% TODO sort all this out; maybe there's scope for rabbit_mgmt_request?
 
 -export([is_authorized/2, is_authorized_admin/2, is_authorized_admin/4,
-         is_authorized_admin/3, vhost/1, vhost_from_headers/1]).
+         is_authorized_admin/3, vhost/1, vhost_or_bad_request/2, vhost_from_headers/1]).
 -export([is_authorized_vhost/2, is_authorized_user/3,
          is_authorized_user/4, is_authorized_user/5,
          is_authorized_monitor/2, is_authorized_policies/2,
@@ -209,6 +209,14 @@ vhost_from_headers(ReqData) ->
 
 vhost(ReqData) ->
     rabbit_web_dispatch_access_control:vhost(ReqData).
+
+vhost_or_bad_request(ReqData, Context) ->
+    case rabbit_web_dispatch_access_control:vhost(ReqData) of
+        not_found -> 
+            bad_request(rabbit_data_coercion:to_binary("vhost_not_found"),
+                      ReqData, Context);
+        VHost -> VHost
+    end.
 
 user(ReqData) ->
   case id(user, ReqData) of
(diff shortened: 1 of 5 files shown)


michaelklishin (changes requested):
This check can and should be performed in `resource_exists/2`, a callback that runs much easier than `to_json/2` and friends.

If an endpoint supports `POST` requests, then `allow_missing_post/2` would have to return `false`, too.

deps/rabbitmq_management/src/rabbit_mgmt_util.erl
```diff
@@ -210,6 +210,14 @@ vhost_from_headers(ReqData) ->
 vhost(ReqData) ->
     rabbit_web_dispatch_access_control:vhost(ReqData).
 
+vhost_or_bad_request(ReqData, Context) ->
```
michaelklishin, line 213: With `resource_exists/2`, this function would be simpler (as it would only return a boolean) and can be made composable with other functions, e.g. those that verify that a queue exists, if this PR decides to go that far.

deps/rabbitmq_management/src/rabbit_mgmt_util.erl
```diff
@@ -210,6 +210,14 @@ vhost_from_headers(ReqData) ->
...
+    case rabbit_web_dispatch_access_control:vhost(ReqData) of
+        not_found -> 
+            bad_request(rabbit_data_coercion:to_binary("vhost_not_found"),
```
michaelklishin, line 216: If a virtual host exists, shouldn't the most suitable HTTP return code be a 404 Not Found and not a 400 Bad Request?

LoisSotoLopez: Yes, in those cases where both the vhost and the queue (or the vhost and a exchange, or a vhost and a binding) is the resource.

For instance, if a POST to `/api/queues/<vhost>/<queue>/get` failed with a 404 I would expect either the vhost or the queue for that host to be missing, because I understand the resource is the combination of both the vhost and the queue.

But on other more complex cases such as a POST to `/api/bindings/<vhost>/e/<src-exchange>/e/<dst-exchange>` a 404 would certainly not be enough. What's missing there? The source exchange on the current vhost, or the destination exchange on the current vhost? Or the vhost itself?

Regarding the proposed changes I limited myself to prevent 500 errors on posts, and since looking at previous code some cases were returning a 400, such as in [rabbit_mgmt_wm_definitions.erl](https://github.com/rabbitmq/rabbitmq-server/blob/dfddfe3fbb9070018b1be5d620f192f6d20b614b/deps/rabbitmq_management/src/rabbit_mgmt_wm_definitions.erl#L39), which solution is better, and coherent with the current design, wasn't clear.

I can let all changed callback modules fail with a 404 either using `resource_exists/2` and `allow_missing_post/2` or  `rabbit_mgmt_util:not_found`, which allows to specify the reason for the 404 (vhost_not_found); and revert `rabbit_mgmt_wm_definitions.erl` which already handled the missing vhost appropriately.

deps/rabbitmq_management/src/rabbit_mgmt_util.erl
LoisSotoLopez: The thing with `resource_exists/2` and `allow_missing_post/2` is that we don't reference that the vhost is the resource missing.

```
< HTTP/1.1 404 Not Found
< content-length: 49
< content-security-policy: script-src 'self' 'unsafe-eval' 'unsafe-inline'; object-src 'self'
< content-type: application/json
< date: Tue, 09 Apr 2024 11:42:21 GMT
< server: Cowboy
< vary: accept, accept-encoding, origin
<
* Connection #0 to host localhost left intact
{"error":"Object Not Found","reason":"Not Found"}
```
Considering what I stated on my [other comment](https://github.com/rabbitmq/rabbitmq-server/pull/10908#discussion_r1557549571) it felt more appropriate to trigger a `not_found` in a way I could make explicit that the vhost was the missing resource.

Maybe a better solution would be to put the call to `rabbit_mgmt_util:not_found` either on the `resource_exists/2` callback, if the vhost does not exist, or on the `allow_missing_post/2` callback.

Issue #10753, opened 2024-03-15
Khepri: timeouts when one of the nodes stops responding


mkuratczyk:
### Describe the bug

During chaos tests where one of the VMs/nodes is suddenly restarted, timeouts like this occur:
```
   crasher:
     initial call: rabbit_prequeue:init/1
     pid: <0.1007.0>
     registered_name: []
     exception exit: {{badrecord,
                          {error,
                              {timeout,
                                  {rabbitmq_metadata,
                                      'rabbit@foobar-s5000-server-1.foobar-s5000-nodes.chaos-tests'}}}},
                      [{dict,map_dict,2,[{file,"dict.erl"},{line,467}]},
                       {rabbit_amqqueue,internal_delete,3,
                           [{file,"rabbit_amqqueue.erl"},{line,1805}]},
                       {rabbit_amqqueue_process,'-terminate_delete/3-fun-1-',
                           7,
                           [{file,"rabbit_amqqueue_process.erl"},{line,332}]},
                       {rabbit_amqqueue_process,terminate_shutdown,2,
                           [{file,"rabbit_amqqueue_process.erl"},{line,362}]},
                       {gen_server2,terminate,3,
                           [{file,"gen_server2.erl"},{line,1158}]},
                       {gen_server2,handle_msg,2,
                           [{file,"gen_server2.erl"},{line,1048}]},
                       {proc_lib,wake_up,3,
                           [{file,"proc_lib.erl"},{line,251}]}]}
```
```
   crasher:
     initial call: rabbit_channel:init/1
     pid: <0.90831.0>
     registered_name: []
     exception exit: {{case_clause,
                          {error,
                              {timeout,
                                  {rabbitmq_metadata,
                                      'rabbit@foobar-s5000-server-2.foobar-s5000-nodes.chaos-tests'}}}},
                      [{rabbit_channel,binding_action,10,
                           [{file,"rabbit_channel.erl"},{line,1825}]},
                       {rabbit_channel,handle_method,3,
                           [{file,"rabbit_channel.erl"},{line,1614}]},
                       {rabbit_channel,handle_cast,2,
                           [{file,"rabbit_channel.erl"},{line,631}]},
                       {gen_server2,handle_msg,2,
                           [{file,"gen_server2.erl"},{line,1056}]},
                       {proc_lib,init_p_do_apply,3,
                           [{file,"proc_lib.erl"},{line,241}]}]}
[...]

PR #10915, opened 2024-04-03
Handle database timeouts from Khepri minority

the-mikedavis:
Operations like declaring/deleting queues fail when sent against a node that's part of a minority. We need to let the database failures (`{error, timeout}`) bubble up to the callers - usually the channel - so that these operations don't cause needless crash reports.

Closes https://github.com/rabbitmq/rabbitmq-server/issues/10753
This depends on a change upstream in Khepri: https://github.com/rabbitmq/khepri/pull/256

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -207,11 +207,11 @@ erlang_package.hex_package(
     version = "1.4.1",
 )
 
-erlang_package.hex_package(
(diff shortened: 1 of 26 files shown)


the-mikedavis:
These changes have been split out into other smaller PRs now

PR #10916, opened 2024-04-04
Make more TLS client options configurable for Kubernetes peer discovery

michaelklishin:
## Changes

 * Use file path validators to improve error messages when a certificate, key or another file does not exist or cannot be read by the node
 * Introduce a number of standard TLS options in addition to the Kubelet-provided CA certificate

## Why are they Necessary?

Starting with Erlang 26, [peer verification](https://www.rabbitmq.com/docs/ssl#peer-verification) is enabled by default for TLS clients. This may
be highly inconvenient in certain environments, and there is currently no way to disable
peer verification for K8S API requests. For example, see https://github.com/rabbitmq/cluster-operator/issues/1602.

While adding new TLS client `rabbitmq.conf` schema keys, I've improved validation of the path keys we already have.

### DIFF
--- a/deps/rabbitmq_peer_discovery_k8s/priv/schema/rabbitmq_peer_discovery_k8s.schema
+++ b/deps/rabbitmq_peer_discovery_k8s/priv/schema/rabbitmq_peer_discovery_k8s.schema
@@ -51,7 +51,7 @@ end}.
 %% (ACL) Token path
 
 {mapping, "cluster_formation.k8s.token_path", "rabbit.cluster_formation.peer_discovery_k8s.k8s_token_path", [
-    {datatype, string}
+    {datatype, string}, {validators, ["file_accessible"]}
 ]}.
 
 {translation, "rabbit.cluster_formation.peer_discovery_k8s.k8s_token_path",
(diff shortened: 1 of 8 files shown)


deps/rabbitmq_peer_discovery_k8s/priv/schema/rabbitmq_peer_discovery_k8s.schema
```diff
@@ -76,10 +80,73 @@ fun(Conf) ->
...
+
+{translation, "rabbit.cluster_formation.peer_discovery_k8s.ssl_options.cacertfile",
+fun(Conf) ->
+    case cuttlefish:conf_get("cluster_formation.k8s.tls.cacertfile", Conf, undefined) of
+        undefined -> cuttlefish:unset();
+        Value     -> Value
+    end
+end}.
+
+{mapping, "cluster_formation.k8s.tls.certfile", "rabbit.cluster_formation.peer_discovery_k8s.ssl_options.certfile",
```
michaelklishin, line 97: We use `tls` for prefix around definition import (from an HTTPS endpoint), so I figured it would make a bit more sense here compared to `ssl_options` if server listeners.

michaelklishin:
A use confirms that this change does work as expected https://github.com/rabbitmq/cluster-operator/issues/1602#issuecomment-2041305683 => merging and backporting.

PR #10918, opened 2024-04-04
Split OCI workflow into two workers

pjk25:
~~on an M1 worker~~

~~under qemu emulation, the build is over 90 minutes~~


pjk25:
> on an M1 worker
> 
> under qemu emulation, the build is over 90 minutes

Turns out this doesn't seem to be possible right now. Headless install of Docker Desktop, nor Rancher Desktop, seem to work on the M1 workers. For that matter, I couldn't get `docker buildx build ...` to work in any form on the M1 worker.

Instead, this PR just splits the workflow into two parallel jobs. Hopefully caching still works properly.

PR #10928, opened 2024-04-05
Prefer `erlang:monotonic_time/0` for elapsed time measurements

the-mikedavis:
This is now the recommended way to measure durations with the time API: <https://www.erlang.org/doc/apps/erts/time_correction#how-to-work-with-the-new-api>. This is mostly a cosmetic change though since it's only used in/for log messages.

\cc @dumbbell I've changed the log message for the time it takes to regen the feature flag registry to be milliseconds instead of microsecond as it takes ~35-40ms for me locally. (Similar to https://github.com/rabbitmq/rabbitmq-server/pull/10674)


michaelklishin (approved):

the-mikedavis:
I reverted the change to switch to milliseconds for the feature flag registry regen since #10988 might reduce the time it takes to generate the registry down to hundreds of microseconds. So now this patch just switches to `monotonic_time` - no changes to log messages

PR #10949, opened 2024-04-09
Have minimum queue TTL take effect for quorum queues

ansd:
For classic queues, if both policy and queue argument are set for queue TTL, the minimum takes effect.

Prior to this commit, for quorum queues if both policy and queue argument are set for queue TTL, the policy always overrides the queue argument.

This commit brings the quorum queue queue TTL resolution to classic queue's behaviour. This allows developers to provide a custom lower queue TTL while the operator policy acts an upper bound safe-guard.


michaelklishin:
This is also consistent with how the regular policy vs. operator policy tie break works for limits such as TTL and queue length limits: the lower (and safer w.r.t. resource use) wins.

michaelklishin (approved):

ansd:
https://github.com/rabbitmq/rabbitmq-server/blob/0172d0406a7e41cb5bcc906480881454d37ef207/deps/rabbit/src/rabbit_policies.erl#L216

ansd:
We can ship this PR only in 4.0 because it could otherwise come as a surprise to existing client applications which rely on today's 3.13 behaviour and therefore deleting quorum queues earlier than expected.
That's why backporting this PR to v3.13.x branch is too risky and a breaking change.

PR #10961, opened 2024-04-10
GitHub Actions: Add GHCR login step to `test-{authnz,selenium}.yaml` workflows

dumbbell:
## Why

Both workflows build and publish docker images but they currently fail to publish them because of a permission denied.

## How

This patch replicates (copy-pastes) what was done to the `oci.yaml` workflow.

### DIFF
--- a/.github/workflows/oci.yaml
+++ b/.github/workflows/oci.yaml
@@ -29,7 +29,7 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
-          password: ${{ secrets.GHCR_TOKEN }}
+          password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
--- a/.github/workflows/test-authnz.yaml
+++ b/.github/workflows/test-authnz.yaml
@@ -52,7 +52,7 @@ jobs:
       with:
         registry: ghcr.io
         username: ${{ github.actor }}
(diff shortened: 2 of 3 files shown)


dumbbell:
Thank you for the feedback @Zerpet!

I don’t really know what I’m doing in this patch, I just copy-pasted what I found in other workflows :-) Do you think we should do the authentication differently?

It was not enough anyway. The first run succeeded in CI but rebases of the patch (with no changes otherwise) fail: the AWS peer discovery workflow aborts before the test while setting the VMs with the error "An error occurred (InvalidParameterException) when calling the CreateService operation: Creation of service was not idempotent.".

As if the test can’t be executed twice for the same branch.

Zerpet:
@dumbbell a quirk I noticed from GitHub actions: if you rebase and force-push your branch, the actions may not pick up the updated workflow file, even tho the "Workflow manifest" tab shows the file as you expect. I don't know if that's the root cause of the failure, but I thought I would mention it just in case 🙂 I workaround this issue by making WIP commits, and rebasing at the very end, just before merging.

Regarding authentication, I _suspect_ the value of `GHCR_TOKEN` is a personal access token of some account, probably `rabbitmq-ci`. In that case, we should not use `github.actor` as the username, but `rabbitmq-ci` via a variable/secret/hard-coding. If my theory is correct, authentication will fail for any other user.

pjk25:
Regarding "Creation of service was not idempotent":

https://github.com/hashicorp/terraform-provider-aws/issues/20452

PR #10988, opened 2024-04-12
4.0: Use `persistent_term` for the feature flag registry

the-mikedavis:
This is an experiment for using `persistent_term` to hold the feature flag registry. `rabbit_ff_registry` is currently implemented with a module that is dynamically regenerated with `erl_syntax`, `compile:forms/2` and code reloading. The module approach has the advantage of maximizing read speed but it has a few drawbacks:

* We need workarounds to convince the dialyzer of the real typespecs of `rabbit_ff_registry` functions.
* We need to be careful to not deadlock around the code server.
* Regenerating the registry can be relatively slow: around 35-40ms for me locally per regeneration. With this patch I see a total savings of ~800ms during a regular single-node boot (`bazel run broker`) and a total savings of ~200ms with `rabbitmqctl enable_feature_flag khepri_db` for example.

The question is whether `persistent_term` will be fast enough. Running the usual `one_fast` perf-test flag setup I didn't see a difference between this branch and main but we'll want to run this through the full performance test suite before considering it.

This branch also includes a concurrency fix for `rabbit_feature_flags:inject_test_feature_flags/2` which can cause the `feature_flags_SUITE:registry_concurrent_reloads/1` case to misbehave.


lin72h:
very creative!

the-mikedavis:
I took some microbenchmarks of the two approaches - generating a module (albeit an Elixir module in the test) and reading from a `:persistent_term` - for the `rabbit_ff_registry:is_enabled/1` use-case:

[bench.exs.txt](https://github.com/rabbitmq/rabbitmq-server/files/15002547/bench.exs.txt)

(which can be run with `mv bench.exs.txt bench.exs` and `elixir bench.exs`)

<details><summary>Benchee results...</summary>

```
Operating System: Linux
CPU Information: AMD Ryzen Threadripper 3970X 32-Core Processor
Number of Available Cores: 64
Available memory: 125.64 GB
Elixir 1.16.2
Erlang 26.2.4
JIT enabled: true

Benchmark suite executing with the following configuration:
warmup: 1 s
time: 10 s
memory time: 2 s
reduction time: 2 s
parallel: 1
inputs: none specified
Estimated total run time: 30 s

Benchmarking dynamic module ...
Benchmarking persistent_term ...
Calculating statistics...
Formatting results...

Name                      ips        average  deviation         median         99th %
dynamic module        30.31 M       32.99 ns  ±7714.91%          30 ns          41 ns
persistent_term       17.67 M       56.58 ns  ±4861.17%          51 ns          70 ns

Comparison: 
dynamic module        30.31 M
persistent_term       17.67 M - 1.71x slower +23.59 ns

Memory usage statistics:

Name               Memory usage
dynamic module              0 B
persistent_term             0 B - 1.00x memory usage +0 B

**All measurements for memory usage were the same**

Reduction count statistics:

Name            Reduction count
dynamic module                2
persistent_term               2 - 1.00x reduction count +0

**All measurements for reduction count were the same**
```

</details>

`persistent_term` is ~1.7x slower than `is_enabled/1` on a module using the current set of feature flags. This is better than I expected: `persistent_term` is certainly slower but not by a very wide margin since both operations are so cheap.

the-mikedavis:
I also ran this branch against the performance tests vs. main: https://grafana.lionhead.rabbitmq.com/d/1OSIKB4Vk/rabbitmq-performance-tests?orgId=1&from=1713189778195&to=1713195192973&shareView=link

At a glance the two branches are hard to tell apart. The persistent_term may be slower but we might not read feature flags so often that it's noticeable. @mkuratczyk could I bug you to take a look and weigh in on these results when you get a chance?

mkuratczyk:
Yeah, I can't think of code paths where this could slow down normal operations. Small differences between environments are   unfortunately always there in these tests (there are multiple reasons for this, including intentional randomization in a few places, both in perf-test as well as on the broker side). I'm fairly sure the observed differences are due to such entropy, not this PR

deps/rabbit/test/feature_flags_SUITE.erl
```diff
@@ -551,7 +547,9 @@ registry_concurrent_reloads(_Config) ->
     MRef = erlang:monitor(process, Spammer),
     unlink(Spammer),
     exit(Spammer, kill),
-    receive {'DOWN', MRef, process, Spammer, _} -> ok end.
+    receive {'DOWN', MRef, process, Spammer, _} -> ok end,
+    %% All feature flags appeared.
+    ?assertEqual(FinalFFList, ?list_ff(all)).
```
the-mikedavis, line 552: This assertion fails on main but it's because of a bug in the test helpers (see 2e4e321c5a8eb80c3397053b5703f1d4ed495ccd) rather than a bug in the feature flag system

michaelklishin:
On an M1 machine with 10 cores and a fast SSD, this shaves off some 9% off of node-reported startup time 👏 on top of the improvements in #10989. This should nicely benefit all future CI runs where clusters are formed and nodes are restarted many times in tutoal.

deps/rabbit/src/rabbit_ff_controller.erl
```diff
@@ -1144,12 +1146,20 @@ collect_inventory_on_nodes(Nodes, Timeout) ->
       end, {ok, Inventory0}, Rets).
 
 inventory_rpcs(Nodes, Timeout) ->
-    %% We must use `rabbit_ff_registry_wrapper' if it is available to avoid
-    %% any deadlock with the Code server. If it is unavailable, we fall back
-    %% to `rabbit_ff_registry'.
+    %% The feature flag registry in `rabbit_ff_registry' was implemented with
+    %% a module which was dynamically regerated and reloaded in the past. To
```
dumbbell, line 1150: Typo: "regenerated"

deps/rabbit/src/rabbit_ff_registry.erl
```diff
@@ -178,7 +192,12 @@ is_enabled(FeatureName) ->
 %%   source code.
 
 is_registry_initialized() ->
-    always_return_false().
+    case inventory() of
+        init_required ->
+            false;
+        _Inventory ->
+            true
+    end.
```
dumbbell, line 200: This could be simplified with:
```erlang
inventory() =/= init_required.
```

But perhaps you prefer to keep the same logic as other functions?

I'm fine with either solutions.

deps/rabbit/src/rabbit_feature_flags.hrl
```diff
@@ -1,3 +1,12 @@
+-define(PT_INVENTORY_KEY, rabbit_ff_registry).
+%% The `persistent_term' key used to hold the feature flag inventory.
+%%
+%% `persistent_term:get(?PT_INVENTORY_KEY)' should return a value with the type
+%% `rabbit_feature_flags:inventory()' if the registry is initialized.
+%%
+%% Rather than fetching this key directly, use the functions in the
+%% `rabbit_ff_registry' module.
+
```
dumbbell, line 9: Could this be added to `src/rabbit_ff_registry.hrl` please?

`rabbit_feature_flags.*` being the public API of the Feature flags subsystem, we don't want this macro to be part of it.

deps/rabbit/src/rabbit_ff_registry_factory.erl
```diff
@@ -171,19 +168,23 @@ initialize_registry(NewSupportedFeatureFlags,
...
+                                       WrittenToDisk)
     catch
         throw:{error, _} = Error2 ->
             Error2
+    after
+        ?LOG_DEBUG(
+           "Feature flags: released lock after initializing registry (~tp)",
+           [self()],
+           #{domain => ?RMQLOG_DOMAIN_FEAT_FLAGS}),
+        true = global:del_lock(?FF_REGISTRY_LOADING_LOCK, [node()])
```
dumbbell, line 187: Technically, the lock is not "released" yet at the time of the log message :-)

So either the log message says "release" or it is moved after the unlock.

dumbbell (changes requested):

deps/rabbit/src/rabbit_ff_registry_factory.erl
the-mikedavis: Good point! Swapped to saying "releasing"

dumbbell (approved):

PR #11012, opened 2024-04-16
Store MQTT messages as non-durable if QoS 0

ansd:
By default, when the 'durable' message container (mc) annotation is unset, messages are interpreted to be durable.

Prior to this commit, MQTT messages that were sent with QoS 0 were stored durably in classic queues.
This commit takes the same approach for mc_mqtt as for mc_amqpl and mc_amqp: If the message is durable, the durable mc annotation will not be set. If the message is non-durable, the durable mc annotation will be set to false.


kjnilsson (approved):

ansd:
We should mention in the release notes that, in 3.13.0 and 3.13.1, MQTT QoS 0 messages that were sent to classic queues were always persisted to disk which is unnecessary and may result in poorer performance. This is fixed in 3.12.2.

PR #11030, opened 2024-04-18
Management UI: extend Get Message feature to streams

LoisSotoLopez:
## Proposed Changes

Extends the management endpoint for retrieving messages from a queue allowing this operation to be applied to stream queues. Alters the `queue.ejs` template to show the _Get messages_ box when showing a stream queue information.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbitmq_management/src/rabbit_mgmt_wm_queue_get.erl
+++ b/deps/rabbitmq_management/src/rabbit_mgmt_wm_queue_get.erl
@@ -7,6 +7,7 @@
 
 -module(rabbit_mgmt_wm_queue_get).
 
+-include_lib("kernel/include/logger.hrl").
 -export([init/2, resource_exists/2, is_authorized/2, allow_missing_post/2,
   allowed_methods/2, accept_content/2, content_types_provided/2,
   content_types_accepted/2]).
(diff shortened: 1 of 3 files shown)


deps/rabbitmq_management/src/rabbit_mgmt_wm_queue_get.erl
```diff
@@ -47,34 +48,124 @@ accept_content(ReqData, Context) ->
...
+                         delivery_tag = DeliveryTag,
+                         consumer_tag = ConsumerTag},
+        #amqp_msg{props = Props, payload = Payload}} ->
+            ok = amqp_channel:cast(Ch, #'basic.ack'{delivery_tag = DeliveryTag,
+                                                     multiple = false}),
+            [{payload_bytes, size(Payload)},
+             {redelivered, Redelivered},
+             {exchange, Exchange},
+             {routing_key, RoutingKey},
+             {consumer_tag, ConsumerTag},
```
gomoripeti, line 137: consumer tag is just hardcoded "ctag" so it does not need to be included.

deps/rabbitmq_management/src/rabbit_mgmt_wm_queue_get.erl
```diff
@@ -47,34 +48,124 @@ accept_content(ReqData, Context) ->
...
 
+subscription_get(Ch, Enc, Trunc) ->
+    receive 
+        {#'basic.deliver'{redelivered = Redelivered,
+                         exchange = Exchange,
+                         routing_key = RoutingKey,
+                         delivery_tag = DeliveryTag,
+                         consumer_tag = ConsumerTag},
+        #amqp_msg{props = Props, payload = Payload}} ->
+            ok = amqp_channel:cast(Ch, #'basic.ack'{delivery_tag = DeliveryTag,
```
gomoripeti, line 131: I dont think we want to ack the message. If the stream has 20 messages, and we ask for 10 (prefetch = 10) when we ack the first message, the broker will send the 11th message. (Prefetch count means the max number of outstanding ie unacknowledged messages)

deps/rabbitmq_management/src/rabbit_mgmt_wm_queue_get.erl
```diff
@@ -47,34 +48,124 @@ accept_content(ReqData, Context) ->
...
+    after
+        300 ->
+            none 
+    end.
 
+subscribe(Ch, Queue, NoAck, Offset, CTag) ->
+    amqp_channel:subscribe(
+        Ch, 
+        #'basic.consume'{queue = Queue,
+            no_ack = NoAck,
```
gomoripeti, line 149: NoAck is always false so we can hardcode it I think
Also hardcode the CTag

deps/rabbitmq_management/src/rabbit_mgmt_wm_queue_get.erl
michaelklishin: This feature should not arguably use AMQP 0-9-1 at all. And the original feature should never have used an AMQP 0-9-1 Erlang client to begin with.

michaelklishin:
Before you spend any more time on this feature, have you considered consulting with the core team as to whether this feature has any chance of being accepted? Or, perhaps, whether it should use the Erlang AMQP 0-9-1 client? Or any client at all and not an internal API?

I'd very much like to see contributors, even those who contribute about 1-2% of R&D, to spend their time on more fundamental issues like https://github.com/rabbitmq/rabbitmq-server/issues/11038 or https://github.com/rabbitmq/rabbitmq-server/pull/10763 (two examples of much more fundamental problems it does not take **that** deep an understanding of RabbitMQ internals to help).

deps/rabbitmq_management/src/rabbit_mgmt_wm_queue_get.erl
kjnilsson: Yes it would probably be better to use `rabbit_queue_type` directly to get a message off the stream.

gomoripeti: I dont say that an amqp connection should be used to implement this feature, just want to mention the aspect, or implicit benefit that the server side of an amqp connection does additional authorization checks.

Would like to cite the recent attempt for the delete queue endpoint to use an internal API https://github.com/rabbitmq/rabbitmq-server/pull/9550 which was reverted https://github.com/rabbitmq/rabbitmq-server/pull/10062#issuecomment-1864117758. Maybe can you recall what was the difficulty, why the path outlined in #10062 was abandoned?

Or maybe that restriction is not applicable for a simple reading from a queue?

carlhoerberg: Sounds like a way to not move anywhere.. why would not using the 0-9-1 protocol be ok expect it's maybe not 100% optimal? This solves a real world problem with minimal change.

gomoripeti: Dear Core Team! Do you know the difficulties of or things to consider when converting the Management plugin to not use an AMQP client, eg in context of #9550 and #10062? This would be helpful to move forward with this PR either using an amqp client or not.

PR #11045, opened 2024-04-19
rabbit_peer_discovery: Fixes and improvements for Consul and etcd

dumbbell:
### Node registration

After the rework of the peer discovery subsystem in RabbitMQ 3.13.0 to fix cases where clusters would be incorrectly initialized, there were regressions in the Consul and etcd backends causing nodes to hang during peer discovery and eventually boot as standalone nodes.

For Consul and etcd, the discovery process relies on the node registration: they return whatever was previously registered. With the new checks and failsafes added in RabbitMQ 3.13.0, the fact that registration happens after running discovery breaks these backends.

It used to work before because the first node would eventually time out waiting for a non-empty list of nodes from the backend and proceed as a standalone node, registering itself on the way. Following nodes would then discover that first node.

Among the new checks, the node running discovery expects to find itself in the list of discovered nodes. Because it didn't register yet, it will never find itself and reject the peer discovery results.

### Node selection

In addition to that, RabbitMQ 3.13 introduce a new nodes sorting to make sure the node selection was deterministic. However in Consul, the order in which nodes are registered should be considered in case registration happens in a different order that nodes booted in. Therefore, the patch now allows to let the backend select the node to join and return a single node instead of a list.

This is committed at the same time as the first part because the new testsuites may fail transiently without this.

## How

The solution is to register first, then run discovery. The node should at least get itself in the discovered nodes.

While here, add a testsuite in both the Consul and etcd plugins to actually test clustering against a real Consul/etcd server.

Finally, allow backends to select the node to join and return a single node name instead of a list. We still query that node’s properties to see if it is ready to be joined.


dumbbell:
As of this writing, the etcd testsuite is not finished and the Consul is yet to be extended.

dumbbell:
In the end, I put two peer discovery improvements in this patch that fix regressions in the Consul and etcd backends, plus new testsuites for these backends. Thus the scope expanded compared to the initial intent of this pull request. I'm going to rewrite the pull requestion description to explain everything.

frederikbosch:
I just found out that [my whole exercise in this discussion](https://github.com/rabbitmq/rabbitmq-server/discussions/11161) is what will probably be resolved in this PR. 

If I may, what I would prefer as a Consul / Nomad user, is to let RabbitMQ read what is already registered in Consul by Nomad. Hence, I would like to disable RabbitMQ registering the service. That's something that my container orchestrator (Nomad) already takes care of. So RabbitMQ sends a query to Consul, sees three nodes, and constructs the cluster including leader election. 

Config wise, this might be realized by.
```
cluster_formation.consul.svc_addr_auto = false
cluster_formation.consul.svc_addr =
```

And I would include the required policy in Consul in the documentation, which for 3.12 I believe is the following.

```hcl
node_prefix ""
{
  policy = "read"
}

service_prefix "rabbitmq"
{
  policy = "write"
}

agent_prefix ""
{
  policy = "read"
}

session_prefix ""
{
  policy = "write"
}

key_prefix "rabbitmq/"
{
  policy = "write"
}
```

michaelklishin:
We will backport this in a week or two once we have some evidence that this PR works as expected during CI runs in `main` and day-to-day core team development use (of `main`).

frederikbosch:
@michaelklishin What is the image name of the main branch? I am planning to test today.

frederikbosch:
I have tested this PR using a three node cluster with two scenario's.

1. Also register the service using my orchestration tool (Nomad) in Consul with the meta `erlang-node-name` set, under the same name (rabbitmq) as RabbitMQ will register the service.
2. Only let RabbitMQ do the registration in Consul.

The second scenario has one big downside: what if RabbitMQ did not close properly? Then the service remains in the registry. This could lead to an unrecoverable cluster. I actually ran into this scenario. What happened?

- I stopped the cluster, and RabbitMQ did not have time to shutdown properly, so the node was killed
- Leaving the service in the registry
- When I restarted all the nodes they query consul and see the left services (status passing), and try to join it.
- This results in RabbitMQ crashing, because the hostname does not resolve anymore. In my cluster Docker services only have a resolving FQDN when they are actually running.

```
=PROGRESS REPORT==== 14-May-2024::12:25:21.123386 ===
    supervisor: {local,inet_gethost_native_sup}
    started: [{pid,<0.97.0>},{mfa,{inet_gethost_native,init,[[]]}}]

=PROGRESS REPORT==== 14-May-2024::12:25:21.133443 ===
    supervisor: {local,kernel_safe_sup}
    started: [{pid,<0.96.0>},
              {id,inet_gethost_native_sup},
              {mfargs,{inet_gethost_native,start_link,[]}},
              {restart_type,temporary},
              {significant,false},
              {shutdown,1000},
              {child_type,worker}]

2024-05-14 12:25:21.171098+00:00 [notice] <0.44.0> Application mnesia exited with reason: stopped
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> 
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> BOOT FAILED
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> ===========
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> Exception during startup:
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> 
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> error:function_clause
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> 
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0>     rabbit_peer_discovery:select_node_to_join/1, line 873
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0>         args: [[]]
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0>     rabbit_peer_discovery:sync_desired_cluster/3, line 206
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0>     rabbit_db:init/0, line 66
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0>     rabbit_boot_steps:-run_step/2-lc$^0/1-0-/2, line 51
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0>     rabbit_boot_steps:run_step/2, line 58
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0>     rabbit_boot_steps:-run_boot_steps/1-lc$^0/1-0-/1, line 22
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0>     rabbit_boot_steps:run_boot_steps/1, line 23
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0>     rabbit:start/2, line 978
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> 
2024-05-14 12:25:22.173737+00:00 [debug] <0.253.0> Set stop reason to: {error,function_clause}
2024-05-14 12:25:22.174281+00:00 [debug] <0.253.0> Change boot state to `stopped`
```

Although I have a scenario I can work with, I would suggest to:

- do not let RabbitMQ crash when the hostname does not resolve, but rather skip the node to join
- allow RabbitMQ to not register at all (via a new config `cluster_formation.consul.svc_register = false` or skip registration when `cluster_formation.consul.svc` is left empty)

dumbbell:
Thank you for testing @frederikbosch! I think these are valid concerns. Could you please file an issue for this? The problem existed before anyway, so this is not a regression that should have been fixed by this pull request.

frederikbosch:
[Issue filed](https://github.com/rabbitmq/rabbitmq-server/issues/11233)

PR #11048, opened 2024-04-19
Fix dead letter cycle detection

ansd:
The desired correct dead letter cycle detection behaviour is:
> If there is a death with a source queue that is the same as the target
> queue name and there are no newer deaths with the 'rejected' reason then
> consider this a cycle.

This commit implements this desired behaviour.
Previously, we relied on the arbitrary order of keys returned from `maps:keys/1` which was wrong.


kjnilsson (changes requested):
what do you think of the idea of "simply" changing the death records to a list that is kept  in the modification order?

ansd:
While this PR improves cycle detection, it's still not accurate in all cases because we rely on wall clock time.
Yes, a list defining a clear order of death events (across nodes) is therefore better.

PR #11058, opened 2024-04-22
Fix failing auth test

ansd:
This fixes the failing GitHub action `Test Authentication/Authorization backends via mutiple messaging protocols / selenium` originally caused by https://github.com/rabbitmq/rabbitmq-server/pull/11023


michaelklishin:
FTR, this was per discussion with @MarcialRosales.

PR #11060, opened 2024-04-22
rabbitmq-diagnostics observer: fix accepting user input

mkuratczyk:
Follow-up to https://github.com/rabbitmq/rabbitmq-server/pull/10268.

There were still cases where input was ignored, eg. `rabbitmq-diagnostics -n node observer`


michaelklishin:
The forced push was a rebase.

PR #11076, opened 2024-04-24
WIP: CQ: Fix shared store crashes DO NOT MERGE

lhoguin:
Crashes could happen because compaction would wrongly write over valid messages, or truncate over valid messages, because when looking for messages into the files it would encounter leftover data that made it look like there was a message, which prompted compaction to not look for the real messages hidden within.

To avoid this we ensure that there can't be leftover data as a result of compaction. We get this guarantee by blanking data in the holes in the file before we start copying messages closer to the start of the file. This requires us to do a few more writes but we know that the only data in the files at any point are valid messages.

WIP because it's full of debugging lines left to ease attempts at reproducing the crashes.

Related discussion: #10902 

## Types of Changes


deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1976,9 +2016,7 @@ load_and_vacuum_message_file(File, State = #gc_state{ dir = Dir }) ->
                       Acc;
                   #msg_location { file = File, total_size = TotalSize,
                                   offset = Offset } = Entry ->
-                      [ Entry | Acc ];
-                  _ ->
-                      Acc
+                      [ Entry | Acc ]
```
lhoguin, line 2019: I have removed this clause as I believe it was hiding the real bug and we should really crash if the data we are getting in the file does not match our in-memory state.

lhoguin: The clause is not fully removed anymore, instead we match on what we expect exactly.

PR #11087, opened 2024-04-25
Replicate x_jms_topic_table Mnesia table

acogoluegnes:
The x_jms_topic_table Mnesia table must be on all nodes for messages to be published to JMS topic exchanges and routed to topic subscribers.

The table used to be only in RAM on one node, so it would be unavailable when the node was down and empty
when it came back up, losing the state for subscribers still online because connected to other nodes.

References #9005


michaelklishin (approved):

michaelklishin:
@mergifyio backport v3.13.x v3.12.x v3.11.x

PR #11098, opened 2024-04-26
Create rabbit_table:create_and_replicate_table/2

acogoluegnes:
Used for the maintenance and JMS topic exchange Mnesia tables.


michaelklishin:
Fixed in https://github.com/rabbitmq/rabbitmq-server/pull/11087.

PR #11112, opened 2024-04-29
4.x: Additional CQv2 message store optimisations

lhoguin:
Planning to include these for 4.0.


lhoguin:
After testing the commit [CQ: Don't scan shared store files before deleting them](https://github.com/rabbitmq/rabbitmq-server/pull/11112/commits/7427cfe2348537b3f9f34595b7bec26aefe9100a) does help avoid a backlog of compaction/delete operations in the store GC process. With 24 millions message, 4 queues (2 normal, 2 fan-out), I get about a minute and a half of backlog that is still being done after the queues have consumed all messages without the patch, and no backlog with the patch.

I will see if the other commit helps at all next.

lhoguin:
The other commit does not help improve the startup speed of RabbitMQ so I have removed it. It seems that binary pattern matching optimisations are enough that we don't need to do anything special on top.

lhoguin:
The second commit greatly improves dirty recovery times. On my machine, with data that's 24 million messages spread over many files and two queues, node recovery goes from 4min30 to less than 2min. The problem was that the old code was gathering messages from queues one by one (meaning 3 or 4 Erlang messages per AMQP message!!). Now it does so per segment file.

There are still parts that could be improved for making dirty recovery blazingly fast but they require storing additional state on disk and so will not be investigated fully for now.

lhoguin:
I will work on merging my 4.x PRs next week.

lhoguin:
Making one more addition to this PR so it is not ready to be merged yet.

lhoguin:
I have dropped the first commit. It turns out that the function doing the scanning also removes entries from the ets index, and we need to keep that. So unfortunately we cannot avoid scanning for the time being. To better handle that we have little choice other than reworking the file format on disk.

lhoguin:
I will do some refactoring that should help detect problems like https://github.com/rabbitmq/rabbitmq-server/pull/11288/files better as well as to clearly separate parts of the code that relate with each other.

lhoguin:
@mkuratczyk did a benchmark and a few checks to see that this PR works as intended. Merging, thanks!

PR #11121, opened 2024-04-29
4.x: Use compressed mem tables and set a wal max entries default for quorum queues

kjnilsson:
Compressed ETS tables may introduce a small throughput penalty (low single digit %) but can reduce peak Ra memory use by 30-50%.

Also set a default `wal_max_entries` value to avoid mem tables growing too large when using very small message sizes (as more than 1M tiny messages can easily fit into one WAL file).


mkuratczyk:
Indeed, the memory savings are significant in many cases (obviosuly, mostly with long queues). The throughput difference is small and actually the branch performed a bit better in some cases - it could be caused by "test entropy" (but even then, that means the differences are small if some randomness can flip the results) but could also be actually better thanks to lower memory usage.

## 100B messages

In this test we hit the known issue where `ra_log_segment_writer` can't keep up and therefore memory usage keeps going up as new ETS tables are created before the previous are deleted.
100B

1kb messages:
Same issue as above, but in this case it happened twice on `main` and once on the branch:
1kb

5kb messages:
5kb

Full results: https://grafana.lionhead.rabbitmq.com/goto/bLGbswfIR?orgId=1

mkuratczyk (approved):

PR #11125, opened 2024-04-30
Correct the plt for peer_discovery_etcd in bazel

HoloRin:
so that errors are correctly reported


dumbbell (approved):

michaelklishin:
The forced push was a rebase.

michaelklishin:
Another rebase.

PR #11177, opened 2024-05-06
jose dependency belongs to rabbitmq-components.mk

michaelklishin:
and not oauth2_client or the OAuth 2 plugin.

In follow-up to #10645.

Pair: @pjk25


michaelklishin:
https://github.com/Mergifyio backport v3.13.x

PR #11196, opened 2024-05-08
A way to configure a cluster wide queue limit

SimonUnge:
## Proposed Changes

Much simpler approach to https://github.com/rabbitmq/rabbitmq-server/discussions/10748.

Version for `main`: https://github.com/rabbitmq/rabbitmq-server/pull/11212.

This solution uses a limit conf setting. 

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -335,6 +335,14 @@ rabbitmq_integration_suite(
 )
 
 rabbitmq_integration_suite(
+    name = "cluster_limit_SUITE",
+    size = "medium",
+    additional_beam = [
+        ":test_queue_utils_beam",
+    ],
+)
+
+rabbitmq_integration_suite(
     name = "clustering_management_SUITE",
     size = "large",
     additional_beam = [
--- a/deps/rabbit/app.bzl
+++ b/deps/rabbit/app.bzl
@@ -2080,6 +2080,15 @@ def test_suite_beam_files(name = "test_suite_beam_files"):
         deps = ["//deps/amqp_client:erlang_app", "//deps/rabbitmq_ct_helpers:erlang_app"],
     )
     erlang_bytecode(
+        name = "cluster_limit_SUITE_beam_files",
+        testonly = True,
+        srcs = ["test/cluster_limit_SUITE.erl"],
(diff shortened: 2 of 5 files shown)


SimonUnge:
@michaelklishin @mkuratczyk this is a simpler approach. 

Another way to 'store' the cluster wide queue limit, instead of config, would perhaps be to store it as a global parameter?
i.e `rabbitmqctl set_global_parameter cluster_limit '{"queue":"5"}'`

michaelklishin:
No, we should not use global parameters for this that are not dynamic in nature, and knowing the background on this, this is something that an absolute majority of installations will set in stone and rarely change. Or not set in the first place.

SimonUnge:
OK, ill leave as is and add some tests.

michaelklishin:
@SimonUnge as for setting this cluster-wide setting on every node, since most cluster provisioning needs perform config generation one way or another, this won't be an issue. As long as the value is identical on all nodes anyway, which we can assume to be the case most of the time and/or eventually after a redeployment.

SimonUnge:
yes, I agree. Its the easiest and most straightforward approach to setting it.

mkuratczyk:
Thanks Simon. The limit is enforced in the wrong place however - it's only applicable to AMQP 0.9.1 and "non-native" protocols, so that'd be just STOMP in 4.0 or perhaps not even that. I can easily declare more queues than the limit I set by:
* importing definitions
* starting MQTT subscriptions
* using stream-perf-test
and probably a few other ways

I think these checks should be moved to `rabbit_queue_type:declare`

SimonUnge:
@mkuratczyk ah right you are, I'll move it (I basically just placed it where the vhost queue limit is). I'll move it to a better place.

SimonUnge:
@michaelklishin noticed I pointed this one to v3.13.x... Should I switch to main?

michaelklishin:
You need to create a new branch off of main, cherry-pick the commit(s) there and push it, then submit a PR against main.

SimonUnge:
@michaelklishin Done. Also squashed this commit.

SimonUnge:
Will rebase on https://github.com/rabbitmq/rabbitmq-server/pull/11222 once merged, to combine the logic a bit.

PR #11201, opened 2024-05-09
Fix rabbitmq.conf.example channel_max comment

jaehyung-lee2:
## Proposed Changes

It's a simple change in `channel_max` comment section in rabbitmq.conf.example, where it shows example value of `channel_max` as 128 (which is not the actual default value for channel_max).

Since other example configuration values are given with actual default value, I think the comment section can mislead developers to mistakenly think the default value of `channel_max` is 128.

Well at least it made me look through makefile of the source code to look at actual default channel_max value. And I believe this small change can save time for others as well.

Thank you.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

### DIFF
--- a/deps/rabbit/docs/rabbitmq.conf.example
+++ b/deps/rabbit/docs/rabbitmq.conf.example
@@ -352,7 +352,7 @@
 ## Set the max permissible number of channels per connection.
 ## 0 means "no limit".
 ##
-# channel_max = 128
+# channel_max = 2047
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you but the goal was never to show the default value.

The default value needs to be revisited down, so the least we can do before 4.x is to recommend a much more reasonable value between 100 and 200.

PR #11209, opened 2024-05-10
Move feature flag check outside of mc

kjnilsson:
the `mc` module is ideally meant to be kept pure and portable
 and feature flags have external infrastructure dependencies
as well as impure semantics.

Moving the check of this feature flag into the amqp session
simplifies the code (as no message containers with the new
format will enter the system before the feature flag is enabled).


ansd (approved):
Thank you!

Issue #11208, opened 2024-05-10
Don't close connection if channel without finished publish is closed


carlhoerberg:
### Describe the bug

RabbitMQ isn't respecting section 4.2.6 in the spec: https://www.rabbitmq.com/resources/specs/amqp0-9-1.pdf
 
> Note that any non-content frame explicitly marks the end of the content. Although the size of the content is well-known from the content header (and thus also the number of content frames), this allows for a sender to abort the sending of content without the need to close the channel.

Instead any non Body frame after a Header frame closes the whole connection.

### Reproduction steps

1. Send Basic.Publish
2. Send Basic.Header
3. Send Channel.Close

### Expected behavior

Should close the channel without closing the whole connection.

### Additional context

In [AMQProxy](https://github.com/cloudamqp/amqproxy) we pool channels from many different downstream client on a upstream single connection, but if any of those downstream client disconnect while not have finished a full publish the proxy tries to close the upstream channel but then RabbitMQ closes the whole connection and thus all other channels from all other downstream clients.

michaelklishin:
This is a highly hypothetical scenario that only a proxy can run into. Most RabbitMQ users do not use proxies of this kind, so we will only accept this if it introduces zero regressions for throughput (as reported by several long-ish PerfTest runs).

michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/pull/11210#issuecomment-2104979422

michaelklishin:
Sorry, the more I think about how this can be done (#11210), the more I find section 4.2.6 to be ridiculous and this change as inevitably risky. We have deviated from the AMQP 0-9-1 spec [in the past for practical implementation reasons](https://www.rabbitmq.com/amqp-0-9-1-errata) and questionable or ambiguous decisions in that fronze-in-time spec, and I have no problem with that.

## A Hypothetical Algorithm for Protocol-Aware Proxies

What a proxy could do to avoid an exception is something like this:

 * Knowing the size of the expected content (zero or more body frames), assume that the any frame arriving in the middle of a publish should be delayed
 * After the publish is done, publish all those pending frames

This will not allow clients to "terminate a publish" early but it will be a safe thing to do.

To handle potentially missing body frames, introduce a "flushing timeout" where all pending frames are sent. This would not be a safe thing to do, though, so such timeout would have to [match heartbeat or TCP keepalive timeouts](https://www.rabbitmq.com/docs/heartbeats).

[...]

PR #11210, opened 2024-05-10
Support channel.close in the middle of a publish

carlhoerberg:
## Proposed Changes

RabbitMQ isn't respecting section 4.2.6 in the spec: https://www.rabbitmq.com/resources/specs/amqp0-9-1.pdf

> Note that any non-content frame explicitly marks the end of the content. Although the size of the content is well-known from the content header (and thus also the number of content frames), this allows for a sender to abort the sending of content without the need to close the channel.

Instead any non Body frame after a Header frame closes the whole connection.

This PR allows channel.close frames to cancel an ongoing publish. 

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #11208)
- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

In [AMQProxy](https://github.com/cloudamqp/amqproxy) we pool channels from many different downstream client on a upstream single connection, but if any of those downstream client disconnect while not have finished a full publish the proxy tries to close the upstream channel but then RabbitMQ closes the whole connection and thus all other channels from all other downstream clients.

### DIFF
--- a/deps/rabbit/test/amqp_frames_SUITE.erl
+++ b/deps/rabbit/test/amqp_frames_SUITE.erl
@@ -1,0 +1,182 @@
+-module(amqp_frames_SUITE).
+
+-compile([export_all, nowarn_export_all]).
+
+-include_lib("amqp_client/include/amqp_client.hrl").
+-include_lib("amqp_client/include/amqp_client_internal.hrl").
+
+all() ->
+    [normal_publish,
+     publish_close_after_method_frame,
+     publish_close_after_header_frame,
+     publish_close_after_first_body_frame].
+
+%% -------------------------------------------------------------------
+%% Testsuite setup/teardown.
+%% -------------------------------------------------------------------
+
+init_per_suite(Config) ->
+    rabbit_ct_helpers:log_environment(),
+    Config1 = rabbit_ct_helpers:set_config(Config, [
+                                                    {rmq_nodename_suffix, ?MODULE}
+                                                   ]),
+    rabbit_ct_helpers:run_setup_steps(Config1,
+                                      rabbit_ct_broker_helpers:setup_steps() ++
+                                      rabbit_ct_client_helpers:setup_steps()).
+
+end_per_suite(Config) ->
+    rabbit_ct_helpers:run_teardown_steps(Config,
+                                         rabbit_ct_client_helpers:teardown_steps() ++
+                                         rabbit_ct_broker_helpers:teardown_steps()).
+
+init_per_group(_, Config) -> Config.
+
+end_per_group(_, Config) -> Config.
+
+init_per_testcase(Testcase, Config) ->
+    rabbit_ct_helpers:testcase_started(Config, Testcase).
+
+end_per_testcase(Testcase, Config) ->
+    rabbit_ct_helpers:testcase_finished(Config, Testcase).
+
+%% -------------------------------------------------------------------
+%% Testcases.
+%% -------------------------------------------------------------------
+
+%% According to section 4.2.6 in the spec: https://www.rabbitmq.com/resources/specs/amqp0-9-1.pdf
+%%
+%% Note that any non-content frame explicitly marks the end of the
+%% content. Although the size of the content is well-known from the
+%% content header (and thus also the number of content frames), this
+%% allows for a sender to abort the sending of content without the
+%% need to close the channel.
+
+%% The basic.publish request is encoded as several frames:
+%% 1. Publish Method frame
+%% 2. Header frame
+%% 3... Content Body frames (each containing 2 bytes of the payload,
+%%     as that is how we set up FrameMax)
+
+%% In the normal_publish test case we make sure that sending all the
+%% frames in this raw way works properly, and the broker considers it
+%% as a proper publish.
+
+%% In the consecutive tests we only send the first few frames and then
+%% send a channel.close which should mark the end of the previous
+%% content.
+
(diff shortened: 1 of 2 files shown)


carlhoerberg:
Tests are coming. Will maybe generalise to all method frames as per the spec. Just channel.close was the minimal change to solve our case.

michaelklishin:
This is a highly hypothetical scenario that hasn't come up since 2007 (that I recall).

I'd like to see evidence that it does not introduce any performance regressions.

michaelklishin:
The only scenario where a client can run into something like this is:

 * In one thread, start publishing
 * In another, concurrently close the channel

and sharing channels across threads like that is explicitly not supported in any of the clients. So it must be only proxies where N client connections are multiplexed to 1 actual upstream connection.

gomoripeti:
We will execute performance tests to ensure there is no performance impact. I am certain the code can be made like that, to only modify the code path, when it would hit the frame error anyway, and not the happy path.

In a followup commit I modified the code to not special case the channel.close method but act similarly for any non-content method according to the spec. However that might lead to behaviour which some would consider unexpected or confusing, eg prematurely terminating a publish with a queue.declare, the publish might seem to be lost while the channel is still up. Therefore I can accept if the behaviour is restricted to channel.close when the affected channel won't stay open.

I wasn't sure where to place the tests so I just put them in a new suite. For the record the suite fails without this patch the following way:
```
amqp_frames_SUITE > normal_publish (00:00.193)

amqp_frames_SUITE > publish_close_after_method_frame
    #1. {error,
            {{shutdown,
                 {connection_closing,
                     {server_initiated_close,505,
                         <<"UNEXPECTED_FRAME - expected content header for class 60, got non content header frame instead">>}}},
             {gen_server,call,
                 [<0.246.0>,{close,200,<<"Goodbye">>},infinity]}}}

amqp_frames_SUITE > publish_close_after_header_frame
    #1. {error,
            {{shutdown,
                 {connection_closing,
                     {server_initiated_close,505,
                         <<"UNEXPECTED_FRAME - expected content body, got non content body frame instead">>}}},
             {gen_server,call,
                 [<0.267.0>,{close,200,<<"Goodbye">>},infinity]}}}

amqp_frames_SUITE > publish_close_after_first_body_frame
    #1. {error,
            {{shutdown,
                 {connection_closing,
                     {server_initiated_close,505,
                         <<"UNEXPECTED_FRAME - expected content body, got non content body frame instead">>}}},
             {gen_server,call,
                 [<0.288.0>,{close,200,<<"Goodbye">>},infinity]}}}
```

michaelklishin:
@gomoripeti I expect that this version, with `try/catch`, is a lot more likely to introduce regressions. The original version made more sense to me.

Like I said earlier, I don't see how a realistic client can inject a `queue.declare` between a `basic.publish` and the subsequent content header and body frames. If the scope of this continues to creep, we will close this and ask you to solve this in your proxy, which is the only affected project if you think of it.

michaelklishin:
Sorry, the more I think about how this can be done, the more I find section 4.2.6 ridiculous and this change as inevitably risky. We have deviated from the AMQP 0-9-1 spec [in the past for practical implementation reasons](https://www.rabbitmq.com/amqp-0-9-1-errata) and questionable or ambiguous decisions in that fronze-in-time spec, and I have no problem with that.

michaelklishin:
What a proxy could do to avoid an exception is something like this:

 * Knowing the size of the expected content (zero or more body frames), assume that the any frame arriving in the middle of a publish should be delayed
 * After the publish is done, publish all those pending frames

This will not allow clients to "terminate a publish" early but it will be a safe thing to do.

To handle potentially missing body frames, introduce a "flushing timeout" where all pending frames are sent. This would not be a safe thing to do, though, so such timeout would have to [match heartbeat or TCP keepalive timeouts](https://www.rabbitmq.com/docs/heartbeats).

It's not the most straighforward algorithm to implement but I honestly don't see why  https://github.com/cloudamqp/amqproxy/issues/162 should be solved in RabbitMQ if the user very clearly states that directly connected clients are not affected.

carlhoerberg:
Sorry, the PR was made more generic than intended. We only want to support `channel.close` to not generate `connection.close`. I agree that introducing "non-content frame explicitly marks the end of the content" now can cause unintended consequences, but channel.close cleary shows the intention of the user.

It's not practical in a proxy or pool to buffer full messages, they can be tens of MB large (this is more common that you would think), exploding memory usage when you have potentially hundreds of thousands of downstream clients.

You're wrong to assume that https://github.com/cloudamqp/amqproxy/issues/162 is related, this is a different and a very generic issue for anyone implementing a channel pool.

The branch has been updated with https://github.com/rabbitmq/rabbitmq-server/commit/137315ee55a8f556d7bdf19604e606c5dc424bfa but it doesn't show up in the PR as it's already closed.

gomoripeti:
The generalisation was just an early experiment from my side, and by the time pushing the tests I realised it is a bad idea to interleave frames of multiple methods on the same channel. We don't have any use case for that. Let's forget about that and forget about the spec.

But as Carl wrote it's useful to keep connections long lived, avoid reconnect (which is expensive on TCP, TLS and AMQP levels) and creating connection churn.

As an alternative (only theoretical) the broker could support a channel.close(ChannelId) method on channel 0, so that a client side "channel manager" could terminate individual channels without keeping much state about them. But I think the proposed change in this branch has much smaller diff and less impact.

PR #11222, opened 2024-05-13
Enforce/honor per-vhost queue limit for all protocols

SimonUnge:
## Proposed Changes

The current vhost queue limit  it's only applicable to AMQP 0.9.1 and "non-native" protocols.

This PR moves the check to `rabbit_queue_type:declare`

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_queue_type.erl
+++ b/deps/rabbit/src/rabbit_queue_type.erl
@@ -6,6 +6,7 @@
 %%
 
 -module(rabbit_queue_type).
+-feature(maybe_expr, enable).
 
 -behaviour(rabbit_registry_class).
 
@@ -307,7 +308,7 @@ is_compatible(Type, Durable, Exclusive, AutoDelete) ->
 declare(Q0, Node) ->
     Q = rabbit_queue_decorator:set(rabbit_policy:set(Q0)),
     Mod = amqqueue:get_type(Q),
-    case check_vhost_queue_limit(Q) of
+    case check_queue_limits(Q) of
         ok ->
             Mod:declare(Q, Node);
         Error ->
@@ -770,6 +771,12 @@ known_queue_type_names() ->
     {QueueTypes, _} = lists:unzip(Registered),
     QTypeBins = lists:map(fun(X) -> atom_to_binary(X) end, QueueTypes),
     ?KNOWN_QUEUE_TYPES ++ QTypeBins.
+
+check_queue_limits(Q) ->
+    maybe
+        %% Prepare for more checks
+        ok ?= check_vhost_queue_limit(Q)
+    end.
 
 check_vhost_queue_limit(Q) ->
     #resource{name = QueueName} = amqqueue:get_name(Q),
--- a/deps/rabbit/src/rabbit_queue_type.erl
+++ b/deps/rabbit/src/rabbit_queue_type.erl
@@ -307,6 +307,7 @@ is_compatible(Type, Durable, Exclusive, AutoDelete) ->
 declare(Q0, Node) ->
     Q = rabbit_queue_decorator:set(rabbit_policy:set(Q0)),
     Mod = amqqueue:get_type(Q),
+    ok = check_vhost_queue_limit(Q),
     Mod:declare(Q, Node).
 
 -spec delete(amqqueue:amqqueue(), boolean(),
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_queue_type.erl
```diff
@@ -307,6 +307,7 @@ is_compatible(Type, Durable, Exclusive, AutoDelete) ->
 declare(Q0, Node) ->
     Q = rabbit_queue_decorator:set(rabbit_policy:set(Q0)),
     Mod = amqqueue:get_type(Q),
+    ok = check_vhost_queue_limit(Q),
```
michaelklishin, line 310: This will raise an exception if the limit is reached. The error will neither be logged nor propagated to the channel to be converted into a channel exception.

michaelklishin (changes requested):

deps/rabbit/src/rabbit_queue_type.erl
SimonUnge: Ah, I saw that `rabbit_channel` called the handle_method function from a try-catch so I thought it be OK to throw the error. But, I realize that `rabbit_queue_type:declare` and `rabbit_amqqueue:declare` might actually be called from other sources, so the right thing would be for `rabbit_queue_type:declare` to return a
`{protocol_error, Type :: atom(), Reason :: string(), Args :: term()}` if a limit is met?

kjnilsson: yes a `protocol_error` should be a ok to return as it is already in the return spec for the declare function.

deps/rabbit/src/rabbit_queue_type.erl
```diff
@@ -765,3 +766,13 @@ known_queue_type_names() ->
     {QueueTypes, _} = lists:unzip(Registered),
     QTypeBins = lists:map(fun(X) -> atom_to_binary(X) end, QueueTypes),
     ?KNOWN_QUEUE_TYPES ++ QTypeBins.
+
+check_vhost_queue_limit(Q) ->
+    #resource{name = QueueName} = amqqueue:get_name(Q),
+    VHost = amqqueue:get_vhost(Q),
+    case rabbit_vhost_limit:is_over_queue_limit(VHost) of
+        false         -> ok;
```
kjnilsson, line 774: redundant whitespace, there is no need to arbitrarily line up things like `->` just use a newline instead. Same for the clause below.

SimonUnge: Agree, I just moved the function from rabbit_channel, and didnt want to change it but I am happy to make it look better.

deps/rabbit/src/rabbit_queue_type.erl
```diff
@@ -765,3 +771,25 @@ known_queue_type_names() ->
...
+    maybe
+        %% Prepare for more checks
+        ok ?= check_vhost_queue_limit(Q)
+    end.
+
+check_vhost_queue_limit(Q) ->
+    #resource{name = QueueName} = amqqueue:get_name(Q),
+    VHost = amqqueue:get_vhost(Q),
+    case rabbit_vhost_limit:is_over_queue_limit(VHost) of
+        false->
```
michaelklishin, line 788: Sorry but can a whitespace be added before the `->` here?

michaelklishin (changes requested):

deps/rabbit/src/rabbit_queue_type.erl
SimonUnge: ah, yes of course!

SimonUnge: Done.

michaelklishin (approved):

ansd:
This PR adds a regression: `rabbit_vhost_limit:is_over_queue_limit/1` is now called twice for MQTT connections:
https://github.com/rabbitmq/rabbitmq-server/blob/ad6c1737462f45e4649cb9a9c1471161937a476a/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl#L1357-L1383

SimonUnge:
@ansd Hmmm, totally missed that check. It would only check if twice with a 'false' result I guess... Although something could slip in-between perhaps. I can update the mqtt_processor code to just handle the protocol error return value of rabbit_queue_type:declare?

PR #11225, opened 2024-05-14
Reorganize the Khepri tree

dumbbell:
[TBW]

### DIFF
--- a/deps/rabbit/app.bzl
+++ b/deps/rabbit/app.bzl
@@ -550,6 +550,7 @@ def all_srcs(name = "all_srcs"):
             "include/amqqueue_v2.hrl",
             "include/gm_specs.hrl",
             "include/internal_user.hrl",
+            "include/khepri.hrl",
             "include/mc.hrl",
             "include/rabbit_amqp.hrl",
             "include/rabbit_global_counters.hrl",
--- a/deps/rabbit/include/khepri.hrl
+++ b/deps/rabbit/include/khepri.hrl
@@ -1,0 +1,9 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
+%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%
+%% Copyright (c) 2024 Broadcom. All Rights Reserved. The term “Broadcom”
+%% refers to Broadcom Inc. and/or its subsidiaries. All rights reserved.
+%%
+
+-define(KHEPRI_ROOT_PATH, [rabbitmq]).
--- a/deps/rabbit/src/rabbit_db_exchange.erl
+++ b/deps/rabbit/src/rabbit_db_exchange.erl
@@ -10,6 +10,8 @@
 -include_lib("khepri/include/khepri.hrl").
 -include_lib("rabbit_common/include/rabbit.hrl").
 
+-include("include/khepri.hrl").
+
 -export([
          get_all/0,
          get_all/1,
@@ -41,6 +43,8 @@
          get_in_khepri_tx/1,
          update_in_mnesia_tx/2,
          update_in_khepri_tx/2,
+         clear_exchanges_in_khepri/0,
+         clear_exchange_serials_in_khepri/0,
          path/1
          ]).
 
@@ -49,9 +53,8 @@
 
 -export([
          khepri_exchange_path/1,
-         khepri_exchange_serial_path/1,
-         khepri_exchanges_path/0,
-         khepri_exchange_serials_path/0
+         khepri_exchange_path/2,
+         khepri_exchange_serial_path/1
         ]).
 
 -define(MNESIA_TABLE, rabbit_exchange).
@@ -80,7 +83,10 @@ get_all_in_mnesia() ->
     rabbit_db:list_in_mnesia(?MNESIA_TABLE, #exchange{_ = '_'}).
 
 get_all_in_khepri() ->
-    rabbit_db:list_in_khepri(khepri_exchanges_path() ++ [rabbit_khepri:if_has_data_wildcard()]).
+    rabbit_db:list_in_khepri(
+      khepri_exchange_path(
+        rabbit_khepri:if_has_data_wildcard(),
+        rabbit_khepri:if_has_data_wildcard())).
 
 -spec get_all(VHostName) -> [Exchange] when
       VHostName :: vhost:name(),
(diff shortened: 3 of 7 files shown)


dumbbell:
@michaelklishin: I removed the "4.0" prefix you added to the title because we will probably backport that to 3.13.x once it is stable.

dumbbell:
The layout is now in place.

The next challenge is the handling of deletions, especially the bindings. Currently the code does something like this:
1. In a transaction:
    1. A queue is deleted from the database
    2. Bindings that have that queue as their destination are deleted too
    3. If there are source exchanges with the auto-delete flag set and have no more bindings, they are deleted too
2. Outside of the transaction, deletions are "processed" to emit notifications for the queue, bindings and auto-deleted exchanges.

About the same occurs when a source exchange is deleted.

With this new organization, bindings are implicitly deleted with their source exchange.

Can we cover these situations with `keep_while` conditions in Khepri?
* One on bindings so that they are deleted when their destination (queue or exchange) is deleted?
* One on exchanges so that they are deleted when they no longer have bindings underneath?

The processing of the deletion should remain about the same as it already runs after records were deleted from the database. This would also allow us to drop the transaction and use a single delete.

A related idea: what about using `#if_exists{}` conditions when creating resources? For example, we could add " if vhost exists" in the path of an exchange. This way, the insert would be rejected if the vhost was deleted concurrently. This won't replace the `keep_while` conditions.

dumbbell:
After studying the code more deeply, I believe the current situation is fine:
* Exchanges and bindings that are supposed to be deleted during the deletion of another queue or exchange are already taker care of in a Khepri transaction, and the rest of the logic is handled in `process_deletions()` functions that do nod read the database (they get the old records as arguments).
* Exchange type callback modules are called from these `process_deletions()` functions, so their specific records were already deleted. Fortunately, the callbacks were also simply deleting them.
* The vhost code already calls explicitly the deletion function of everything that live underneath.

Therefore, I think we can defer the improvements to the deletion code to the the time we drop Mnesia.

What do you think @dcorbacho? Am I missing something?

dcorbacho:
@dumbbell Agree, we're fine with that.

dumbbell:
Here is an example of the content of Khepri in an unclustered RabbitMQ with a single PerfTest client running with default options:

```erlang
●
├── '__khepri_mnesia_migration'
│   ╰── m2k_table_copy
│       ╰── <<"rabbitmq_metadata">>
│             Data: {migration,finished,
│                              [rabbit_vhost,rabbit_user,rabbit_user_permission,
│                               rabbit_topic_permission,rabbit_runtime_parameters,
│                               rabbit_queue,rabbit_exchange,rabbit_exchange_serial,
│                               rabbit_route,rabbit_node_maintenance_states,
│                               mirrored_sup_childspec,rabbit_durable_queue,
│                               rabbit_durable_exchange,rabbit_durable_route,
│                               rabbit_semi_durable_route,rabbit_reverse_route,
│                               rabbit_index_route]}
│
╰── rabbitmq
    ├── node_maintenance
    │   ╰── rabbit@giotto
    │         Data: {node_maintenance_state,rabbit@giotto,regular,#{}}
    │
    ├── runtime_params
    │   ╰── internal_cluster_id
    │         Data: {runtime_parameters,internal_cluster_id,
    │                                   <<"rabbitmq-cluster-id-9tjXGNZE97EyiRBoH6sHTA">>}
    │
    ├── users
    │   ╰── <<"guest">>
    │         Data: {internal_user,<<"guest">>,
    │                              <<195,20,77,64,29,28,55,234,23,125,234,249,78,76,209,126,
    │                                72,20,135,236,235,100,222,244,185,61,113,44,7,25,67,198,
    │                                179,92,83,95>>,
    │                              [administrator],
    │                              rabbit_password_hashing_sha256,#{}}
    │
    ╰── vhosts
        ╰── <<"/">>
            │ Data: {vhost,<<"/">>,[],
            │              #{description => <<"Default virtual host">>,tags => []}}
            │
            ├── exchanges
            │   ├── <<>>
            │   │     Data: {exchange,{resource,<<"/">>,exchange,<<>>},
            │   │                     direct,true,false,false,[],undefined,undefined,undefined,
            │   │                     {[],[]},
            │   │                     #{user => <<"rmq-internal">>}}
            │   │
            │   ├── <<"amq.direct">>
            │   │     Data: {exchange,{resource,<<"/">>,exchange,<<"amq.direct">>},
            │   │                     direct,true,false,false,[],undefined,undefined,undefined,
            │   │                     {[],[]},
            │   │                     #{user => <<"rmq-internal">>}}
            │   │
            │   ├── <<"amq.fanout">>
            │   │     Data: {exchange,{resource,<<"/">>,exchange,<<"amq.fanout">>},
            │   │                     fanout,true,false,false,[],undefined,undefined,undefined,
            │   │                     {[],[]},
            │   │                     #{user => <<"rmq-internal">>}}
            │   │
            │   ├── <<"amq.headers">>
            │   │     Data: {exchange,{resource,<<"/">>,exchange,<<"amq.headers">>},
            │   │                     headers,true,false,false,[],undefined,undefined,undefined,
            │   │                     {[],[]},
            │   │                     #{user => <<"rmq-internal">>}}
            │   │
            │   ├── <<"amq.match">>
            │   │     Data: {exchange,{resource,<<"/">>,exchange,<<"amq.match">>},
            │   │                     headers,true,false,false,[],undefined,undefined,undefined,
            │   │                     {[],[]},
            │   │                     #{user => <<"rmq-internal">>}}
            │   │
            │   ├── <<"amq.rabbitmq.trace">>
            │   │     Data: {exchange,{resource,<<"/">>,exchange,<<"amq.rabbitmq.trace">>},
            │   │                     topic,true,false,true,[],undefined,undefined,undefined,
            │   │                     {[],[]},
            │   │                     #{user => <<"rmq-internal">>}}
            │   │
            │   ├── <<"amq.topic">>
            │   │     Data: {exchange,{resource,<<"/">>,exchange,<<"amq.topic">>},
            │   │                     topic,true,false,false,[],undefined,undefined,undefined,
            │   │                     {[],[]},
            │   │                     #{user => <<"rmq-internal">>}}
            │   │
            │   ╰── <<"direct">>
            │       │ Data: {exchange,{resource,<<"/">>,exchange,<<"direct">>},
            │       │                 direct,true,false,false,[],undefined,undefined,undefined,
            │       │                 {[],[]},
            │       │                 #{user => <<"guest">>}}
            │       │
            │       ╰── bindings
            │           ╰── queue
            │               ╰── <<"amq.gen-9chRARA1KM5g_0NqZ8OCQQ">>
            │                   ╰── <<"3c299664-94e7-443a-b642-f7dde58b759e">>
            │                         Data: #{{binding,{resource,<<"/">>,exchange,<<"direct">>},
            │                                          <<"3c299664-94e7-443a-b642-f7dde58b759e">>,
            │                                          {resource,<<"/">>,queue,<<"amq.gen-9chRARA1KM5g_0NqZ8OCQQ">>},
            │                                          []} =>
            │                                     []}
            │
            ├── queues
            │   ╰── <<"amq.gen-9chRARA1KM5g_0NqZ8OCQQ">>
            │         Data: {amqqueue,{resource,<<"/">>,queue,<<"amq.gen-9chRARA1KM5g_0NqZ8OCQQ">>},
            │                         false,true,none,[],<0.736.0>,[],[],[],undefined,undefined,[],
            │                         [],live,0,[],<<"/">>,
            │                         #{user => <<"guest">>},
            │                         rabbit_classic_queue,#{}}
            │
            ╰── user_permissions
                ╰── <<"guest">>
                      Data: {user_permission,{user_vhost,<<"guest">>,<<"/">>},
                                             {permission,<<".*">>,<<".*">>,<<".*">>}}
```

PR #11234, opened 2024-05-14
Fix a flake in amqp_client_SUITE

the-mikedavis:
`event_recorder` listens for events globally rather than per-connection so it's possible for `connection_closed` and other events from prior test cases to be captured and mixed with the events the test is interested in, causing a flake. For example: https://app.buildbuddy.io/invocation/f31d0f74-02f9-407c-887c-838e0952f60a?target=%2F%2Fdeps%2Frabbit%3Aamqp_client_SUITE&targetStatus=6#1

Instead we can assert that the events we gather from `event_recorder` contain the ones emitted by the connection.

### DIFF
--- a/deps/rabbit/test/amqp_client_SUITE.erl
+++ b/deps/rabbit/test/amqp_client_SUITE.erl
@@ -1226,27 +1226,29 @@ events(Config) ->
     end,
     ok = close_connection_sync(Connection),
 
-    [E0, E1, E2] = event_recorder:get_events(Config),
+    Events = event_recorder:get_events(Config),
     ok = event_recorder:stop(Config),
+    ct:pal("Recorded events: ~p", [Events]),
 
-    assert_event_type(user_authentication_success, E0),
     Protocol = {protocol, {1, 0}},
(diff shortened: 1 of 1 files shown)


deps/rabbit/test/amqp_client_SUITE.erl
```diff
@@ -4193,3 +4196,19 @@ has_local_member(QName) ->
...
+      Events :: [#event{}],
+      Ret :: {value, #event{}} | false.
+
+find_event(Type, Props, Events) when is_list(Props), is_list(Events) ->
+    lists:search(
+      fun(#event{type = EventType, props = EventProps}) ->
+              Type =:= EventType andalso
+                lists:all(
+                  fun({Key, _Value}) ->
+                          lists:keymember(Key, 1, EventProps)
```
ansd, line 4212: This commit removes assertions of the event property values (e.g. `<<"/">>` and `<<"guest">>`).
I'm adding the assertions back in https://github.com/rabbitmq/rabbitmq-server/pull/12302.

PR #11248, opened 2024-05-16
Read full message body at once when larger than 4MB

### LINKED ISSUE #11072: Optimise how rdq files are scanned at CQ shared message store recovery startup

The symptom is that after an unclean shutdown when RabbitMQ starts up and runs recovery steps, its memory usage spikes to a seemingly unreasonable way. On smaller servers with let's say only 2GB of memory it causes an OOM (cyclic restarts and the node cannot start up). We have good reason to believe that it is the recovery of the classic queue shared storage, when it reads in all the rdq files, that is responsible for this behaviour. Although there are large rdq files, the total message size of "ready" messages in queues is 1 or 2 orders less than the memory used.

This issue was observer since years from time to time including on 3.12.x

I would like to get early feedback from the Core Team if it is worth investigating (we definitely see this issue often enough) and if contributions would be accepted.

The function [rabbit_msg_store:scan_file_for_valid_messages](https://github.com/rabbitmq/rabbitmq-server/blob/0175c804a26be6769d1f73d4d19fe998ca050d64/deps/rabbit/src/rabbit_msg_store.erl#L1385-L1457) is used to get the MsgIds and their offsets in the file, but at the same time also the message bodies are read in which are thrown away. This might lead to extra large binaries which are not used and extra garbage.

- First it needs to be investigated that indeed this function is the source of excessive memory usage.
- If proven implement a scanning which does not concatenate binaries, and does not create large, unused binaries (instead skipping message bodies)

Part of the problem could be the presence of unreferenced messages in rdq files which are not yet removed from the file. When a server manages to start up after a successful recovery, we've seen the size of the rdq files shrunk. The GC of unreferenced entries changed considerably in 3.13 where the msg store uses a compaction mechanism instead of combining two files. But still there were seen issues (https://github.com/rabbitmq/rabbitmq-server/discussions/10681), so I believe this change could be still relevant. The `scan_file_for_valid_messages` function is not just used at recovery time, but also during compaction and rdq deletion.

_No response_

### Additional context

_No response_

gomoripeti:
## Proposed Changes

When a message (much) larger than 4MB is read from an rdq file in `rabbit_msg_store:scan/6)` adjust the read size to the full message size instead of appending it together from 4MB chunks, which leaves a lot of garbage and memory fragmentation behind.

This patch is intentionally limited to 3.13.x. More major changes coming in 4.0 will make it unnecessary.

Closes #11072 

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -1396,7 +1396,7 @@ scan_file_for_valid_messages(Path) ->
         {ok, Fd} ->
             {ok, FileSize} = file:position(Fd, eof),
             {ok, _} = file:position(Fd, bof),
-            Messages = scan(<<>>, Fd, 0, FileSize, #{}, []),
+            Messages = scan(<<>>, Fd, 0, FileSize, #{}, [], ?SCAN_BLOCK_SIZE),
             ok = file:close(Fd),
             case Messages of
                 [] ->
@@ -1412,8 +1412,8 @@ scan_file_for_valid_messages(Path) ->
                      Reason}}
     end.
 
-scan(Buffer, Fd, Offset, FileSize, MsgIdsFound, Acc) ->
-    case file:read(Fd, ?SCAN_BLOCK_SIZE) of
+scan(Buffer, Fd, Offset, FileSize, MsgIdsFound, Acc, ScanSize) ->
+    case file:read(Fd, ScanSize) of
         eof ->
             Acc;
         {ok, Data0} ->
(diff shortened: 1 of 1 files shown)


gomoripeti:
failing khepri tests seem to be unrelated - some timeouts in stream and shovel tests

lhoguin (approved):

michaelklishin:
Thank you. I trust your observations about lower GC pressure but just out of curiosity, what do the before/after numbers look like?

lhoguin:
It's hard to get numbers for this but this helps avoid OOMs in memory constrained environments. It's a good workaround for 3.13 and a proper fix will be included in 4.0.

gomoripeti:
yes, to clarify this change is about to reduce peak memory consumption, and not increasing speed or reducing cpu usage.
In a memory constrained env with a 105MiB message size the original function got OOM-killed with ~660MB RSS while for this patched version `/usr/bin/time -v` reported ~250MB maximum RSS.
I did not experiment with a lot more different sizes or a different server, and I did not fully understand how this looks from inside the Erlang VM (carriers) but tracing garbage_collection did hint at some differences (https://github.com/rabbitmq/rabbitmq-server/issues/11072#issuecomment-2098062224)

PR #11251, opened 2024-05-16
Wrap maybe function in quotes, so it is treated as an atom

robertoaloi:
## Proposed Changes

Since `maybe` is added as an optional keyword in OTP 25, quote its usage as an atom, so that it does not break syntax highlighting and other language service features.

## Types of Changes

- [X] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit_common/src/rabbit_heartbeat.erl
+++ b/deps/rabbit_common/src/rabbit_heartbeat.erl
@@ -19,7 +19,7 @@
 
 -export_type([heartbeaters/0]).
 
--type heartbeaters() :: {rabbit_types:maybe(pid()), rabbit_types:maybe(pid())}.
+-type heartbeaters() :: {rabbit_types:'maybe'(pid()), rabbit_types:'maybe'(pid())}.
 
 -type heartbeat_callback() :: fun (() -> any()).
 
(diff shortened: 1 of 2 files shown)


mkuratczyk:
Thanks! Please note that we have a branch with OTP27 support where a similar (but different) change was already made: https://github.com/rabbitmq/rabbitmq-server/tree/otp27-support

I opted for a different name, since we already use it in some other places

michaelklishin:
@mkuratczyk given that OTP 27 support will likely get into v3.13.x, I'm leaning towards keeping the `maybe` name. https://github.com/rabbitmq/rabbitmq-server/commit/7bf9af5256dbc9d635bd66d3a66eaa2a98fa421f uses `opt` but we have kept `maybe` for backwards compatibility when `rabbit_types:option/1` was introduced, IIRC.

lhoguin:
The maybe type is referring to https://wiki.haskell.org/Maybe which is used across languages and projects.

michaelklishin:
@lhoguin but so is `option` in `rabbit_types:option/1`. Java, Rust, Scala use `Option` for the same data type as Haskell's `Maybe`.

michaelklishin:
@robertoaloi thank you. We may change it soon, most likely by switching to `rabbit_types:option/1` which was introduced for the same reason. But I see no harm in accepting this PR for now to unblock the continuous experimentation with EPL and the RabbitMQ codebase.

PR #11269, opened 2024-05-19
rabbit_direct client: drop virtual host tree readiness check

michaelklishin:
The check won't be perfectly stable during the initial cluster formation. Moreover, it does not offer much to clients after the cluster is formed because virtual hosts are up the vast majority of the time.


michaelklishin:
The root cause turned out to be something different from what I originally expected, so closing this as more or less irrelevant.

PR #11297, opened 2024-05-22
Reduce inefficiency introduced in b636f41

pjk25:
Use "ERL_LIBS =" instead of "ERL_LIBS :=" so that $(shell ... is not called so often

as per @lhoguin 's suggestion


lhoguin (approved):
This seems to help but doesn't completely fix the extra time added in the build compared to v3.12.x.

PR #11301, opened 2024-05-22
Remove duplicated checks of vhost queue limit

SimonUnge:
## Proposed Changes

#11222 moved the check of vhost queue limit to `rabbit_queue_type:declare`. I did miss that there are checks from other sources too, so I have updated those sources to reflect the changes. Basically removed the check, and handle the precondition error better.

This 'might' have changed the logging msg some, which might or might not be OK?

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

- [x]  I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_amqp_management.erl
+++ b/deps/rabbit/src/rabbit_amqp_management.erl
@@ -142,7 +142,6 @@ handle_http_req(HttpMethod = <<"PUT">>,
         {ok, Result} ->
             Result;
         {error, not_found} ->
-            ok = check_vhost_queue_limit(QName),
             PermCache2 = check_dead_letter_exchange(QName, QArgs, User, PermCache1),
             case rabbit_amqqueue:declare(
                    QName, Durable, AutoDelete, QArgs, Owner, Username) of
@@ -160,6 +159,10 @@ handle_http_req(HttpMethod = <<"PUT">>,
                     %% Must have been created in the meantime. Loop around again.
                     handle_http_req(HttpMethod, PathSegments, Query, ReqPayload,
                                     Vhost, User, ConnPid, {PermCache2, TopicPermCache});
+                {error, queue_limit_exceeded, Reason, ReasonArgs} ->
+                    throw(<<"403">>,
+                          Reason,
+                          ReasonArgs);
                 {protocol_error, _ErrorType, Reason, ReasonArgs} ->
                     throw(<<"400">>, Reason, ReasonArgs)
             end;
@@ -675,17 +678,6 @@ prohibit_reserved_amq(Res = #resource{name = <<"amq.", _/binary>>}) ->
           [rabbit_misc:rs(Res)]);
 prohibit_reserved_amq(#resource{}) ->
     ok.
-
-check_vhost_queue_limit(QName = #resource{virtual_host = Vhost}) ->
-    case rabbit_vhost_limit:is_over_queue_limit(Vhost) of
-        false ->
-            ok;
-        {true, Limit} ->
-            throw(<<"403">>,
-                  "refused to declare ~ts because vhost queue limit ~b is reached",
-                  [rabbit_misc:rs(QName), Limit])
-    end.
-
 
 check_dead_letter_exchange(QName = #resource{virtual_host = Vhost}, QArgs, User, PermCache0) ->
     case rabbit_misc:r_arg(Vhost, exchange, QArgs, ?DEAD_LETTER_EXCHANGE_KEY) of
--- a/deps/rabbit/src/rabbit_amqp_session.erl
+++ b/deps/rabbit/src/rabbit_amqp_session.erl
@@ -71,7 +71,7 @@
 -export([init/1,
          terminate/2,
          handle_call/3,
-         handle_cast/2, 
+         handle_cast/2,
          handle_info/2,
          format_status/1]).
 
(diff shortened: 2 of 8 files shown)


SimonUnge:
@ansd as per discussed in #11222. This PR changes some more uses of the limit check. The actual logging msg might have changed slightly - but now uniformed. If thats not OK, I can fix that...

deps/rabbit/src/rabbit_amqp_management.erl
```diff
@@ -159,7 +159,7 @@ handle_http_req(HttpMethod = <<"PUT">>,
                     %% Must have been created in the meantime. Loop around again.
                     handle_http_req(HttpMethod, PathSegments, Query, ReqPayload,
                                     Vhost, User, ConnPid, {PermCache2, TopicPermCache});
-                {protocol_error, precondition_failed = _ErrorType, Reason, ReasonArgs} ->
+                {protocol_error, precondition_failed = _ErrorType, Reason, [_QueueName, _VHost, _Limit] = ReasonArgs} ->
```
michaelklishin, line 162: You don't have to destructure `ReasonArgs` if the individual parts are not used.

ansd: We could rename the returned error `{protocol_error, precondition_failed, Reason, ReasonArgs}` to something like `{error, queue_limit_exceeded, Reason, ReasonArgs}` so that each caller of `rabbit_queue_type:declare/2` can handle this error correctly?
Here, it maps better to status code `403 Forbidden` than to `412 Precondition Failed`.
https://datatracker.ietf.org/doc/html/rfc7232#section-4.2 doesn't make sense here:
>    The 412 (Precondition Failed) status code indicates that one or more
   conditions given in the request header fields evaluated to false when
   tested on the server.  This response code allows the client to place
   preconditions on the current resource state (its current
   representations and metadata) and, thus, prevent the request method
   from being applied if the target resource is in an unexpected state.

deps/rabbit/src/rabbit_amqp_session.erl
```diff
@@ -2628,6 +2627,11 @@ declare_queue(QNameBin,
             rabbit_core_metrics:queue_created(QName);
         {existing, _Q} ->
             ok;
+        {protocol_error, precondition_failed = _ErrorType, Reason, ReasonArgs} ->
```
ansd, line 2630: We could rename the returned error `{protocol_error, precondition_failed, Reason, ReasonArgs}` to something like `{error, queue_limit_exceeded, Reason, ReasonArgs}` so that each caller of `rabbit_queue_type:declare/2` can handle this error correctly?
"Blindly" mapping `precondition_failed` to `V_1_0_AMQP_ERROR_RESOURCE_LIMIT_EXCEEDED` is brittle. Returning a more correct error from `rabbit_queue_type:declare/2` makes more sense.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -1354,32 +1354,28 @@ create_queue(QNamePart, QOwner, QArgs, QType,
...
+                          QOwner,
+                          QArgs,
+                          VHost,
+                          #{user => Username},
+                          QType),
+        case rabbit_queue_type:declare(Q0, node()) of
+            {new, Q} when ?is_amqqueue(Q) ->
+                rabbit_core_metrics:queue_created(QName),
+                {ok, Q};
+            {protocol_error, precondition_failed = _ErrorType, Reason, ReasonArgs} ->
```
ansd, line 1371: We could rename the returned error `{protocol_error, precondition_failed, Reason, ReasonArgs}` to something like `{error, queue_limit_exceeded, Reason, ReasonArgs}` so that each caller of `rabbit_queue_type:declare/2` can handle this error correctly?
"Blindly" mapping `precondition_failed` to `{error, queue_limit_exceeded}` is brittle here.

ansd (changes requested):
Thanks @SimonUnge for spotting and updating all the other places too!
The changed logged message is no problem.

michaelklishin:
Returning a Precondition Failed to the client sounds perfectly fine to me simply because that's how most queue and stream declaration errors are communicated.

deps/rabbit/src/rabbit_amqp_management.erl
```diff
@@ -160,6 +159,10 @@ handle_http_req(HttpMethod = <<"PUT">>,
                     %% Must have been created in the meantime. Loop around again.
                     handle_http_req(HttpMethod, PathSegments, Query, ReqPayload,
                                     Vhost, User, ConnPid, {PermCache2, TopicPermCache});
+                {protocol_error, queue_limit_exceeded, Reason, ReasonArgs} ->
```
michaelklishin, line 162: This is not a protocol error per se, just a general `error` that has to do with the state (and specific configuration) of the cluster.

I'd use `{error, queue_limit_exceeded, Reason, ReasonArgs}` as @ansd suggested below.

michaelklishin (changes requested):

SimonUnge:
> @SimonUnge the Dialyzer failures seem legit and repeatable.
@michaelklishin 
On it.

UPDATE: Ah, missed adding a file, so should pass now.

michaelklishin (approved):

ansd (approved):
Thanks a lot @SimonUnge !

michaelklishin:
@SimonUnge @ansd AMQP 1.0 topology management does not exist in `v3.13.x` but we can backport other relevant bits, if any.

ansd:
Yes, I'll take care of the backport.

PR #4242, opened 2022-03-09
Make it possible to delete or restart a dynamic Shovel on/from any cluster node (take two)

michaelklishin:
## Proposed Changes

This builds on #4127 by @kostakal. Closes #2807.

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [x] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

PR #11321, opened 2024-05-24
Dynamic Shovel deletion: use a more effective way

michaelklishin:
Dynamic Shovels keep track of their status in
a node-local ETS table which is updated as
Shovels go through the lifecycle: start,
init (connect, declare the topology), stop.

This makes failing Shovels a bit special:
their status records will not be long lived,
which means it will be considered not to exist
by certain code paths.

In addition, for such Shovels we do not know what
node they are hosted on. But that's fine:
we just need to clear their runtime parameter
and a periodic Shovel cleanup will remove all
children for which no schema database entry
(a runtime parameter one) does not exist.

References #4242.


michaelklishin:
This should be safe to backport because

 * Names or signatures of key functions do not change
 * The primary gain comes from the fact that with this change, both CLI tools and the HTTP API endpoints for **deletion** will clear target shovel's runtime parameters. Shovels without runtime parameters will be reaped periodically by the plugin, this protection mechanism has been around for years
 * This revision has decent test coverage and it actually uses the HTTP API to delete Shovels, which previously was not the case
 * The way Shovels store their state or are started has not changed. It's only two explicit deletion operations that are now more effective irrespectively of Shovel state

PR #11329, opened 2024-05-26
Add filtering information to the documentation of RabbitMQ Stream protocol document

wrobell:
1. Link to up-to-date version of code describing Osiris chunk.
2. Update `Deliver` command description with Bloom filter entries.
3. Add information about use of `Publish` command with filter value. The hint about use of version 1 or 2 is based on the Java client implementation.

### DIFF
--- a/deps/rabbitmq_stream/docs/PROTOCOL.adoc
+++ b/deps/rabbitmq_stream/docs/PROTOCOL.adoc
@@ -289,6 +289,9 @@ Publish => Key Version PublisherId PublishedMessages
   Message => bytes
 ```
 
+1. Use version 1 if there is no filter value.
+2. Use version 2 if there is a filter value. 
+
 === PublishConfirm
 
 ```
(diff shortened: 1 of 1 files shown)


deps/rabbitmq_stream/docs/PROTOCOL.adoc
```diff
@@ -394,7 +397,9 @@ Deliver => Key Version SubscriptionId OsirisChunk
   ChunkCrc => int32
   DataLength => uint32
   TrailerLength => uint32
-  Reserved => unit32 // unused 4 bytes
+  BloomSize => uint8 // size of bloom filter data
+  Reserved => unit24 // unused 3 bytes
+  BloomFilterData => [uint8] // bloom filter data (no int32 for the size of this array, the size is defined by BloomSize
```
wrobell, line 402: @michaelklishin Unfortunately, this seems to be invalid.

I receive `BloomSize` default value (16), but `BloomFilterData` is missing.

Looking at Java client (24cc76a51), class `ServerFrameHandler` ignores `BloomSize`, and does not read `BloomFilterData` indeed.

wrobell: https://github.com/rabbitmq/rabbitmq-server/pull/11533/

acogoluegnes: That is right: Osiris does not send the bloom filter data to client applications. Sorry for the confusion.

Issue #10752, opened 2024-03-15
Shovel and AMQP 1.0 targets: expose capabilities needed for some brokers


madmalkav:
### Is your feature request related to a problem? Please describe.

When connecting to some AMQP 1.0 brokers, we cannot set capabilities on the shovel, so in some scenarios we are unable to give the shovel the proper configuration. In example, for IBM MQ , if you want to use a queue but don't set capabilities to queue, it will default to believe the object name provided is for a topic:

![imagen](https://github.com/rabbitmq/rabbitmq-server/assets/11020793/f54f4d02-89b6-45c2-a3d1-4a5e7f91fa7e)

This is documented on [IBM website](https://www.ibm.com/docs/en/ibm-mq/9.2?topic=applications-amqp-10-support):

![imagen](https://github.com/rabbitmq/rabbitmq-server/assets/11020793/69af11a7-35ff-4d5a-8918-bd20c54de37d)

I was able to replicate this same behavior with QPid Proton and this code sample, this code doesn't set capabilities so the same behavior happens:

```
from __future__ import print_function, unicode_literals
from proton import Message
from proton.handlers import MessagingHandler
from proton.reactor import Container

class HelloWorld(MessagingHandler):
    def __init__(self, server, address):
        super(HelloWorld, self).__init__()
        self.server = server
        self.address = address

    def on_start(self, event):
        print("connecting ...")
        conn = event.container.connect(self.server, sasl_enabled=True, allowed_mechs="PLAIN", allow_insecure_mechs=True, user="app", password="passw0rd")
        event.container.create_sender(conn, self.address)

    def on_sendable(self, event):
        print("sending command")
        event.sender.send(Message(body="Hello World!", address="DEV.QUEUE.2", content_type="text/plain"))
        event.sender.close()
        event.connection.close()

Container(HelloWorld("queue://172.17.0.2:5672", "DEV.QUEUE.2")).run()
```

This code set capabilities so it works OK:

```
from __future__ import print_function, unicode_literals
from proton import Message, symbol
from proton.handlers import MessagingHandler
from proton.reactor import Container, SenderOption

class HelloWorld(MessagingHandler):
    def __init__(self, server, address):
        super(HelloWorld, self).__init__()
        self.server = server
        self.address = address

    def on_start(self, event):
        print("connecting ...")
[...]

PR #11337, opened 2024-05-28
Set sender/receiver capabilities to attach frame in the amqp10_client

MarcialRosales:
## Proposed Changes

It partially addresses the feature request https://github.com/rabbitmq/rabbitmq-server/issues/10752. This PR lets a developer set the capabilities to the sender and receiver links in the amqp10_client. It will not add this capability to the shovel plugin though.

The system test group `ibmmq` requires a docker image for IBM MQ server. This docker image is built by a workflow added by this other PR https://github.com/rabbitmq/rabbitmq-server/pull/11419. The official IBM MQ docker image does not come with AMQP built-in capability hence we have to build one. And given the amount of time it takes to build this image it was better to build it from its own CI workflow. 

NOTE for reviewer:
- To run system test group `ibmmq`, it pulls a docker image based on amd64 arch. It does not take into account the local arch because the other possible architecture is arm64 and it is not possible to build a docker image for ARM64 with AMQP capabilities. (see https://github.com/rabbitmq/rabbitmq-server/blob/set-amqp10-capabilities/deps/amqp10_client/test/system_SUITE_data/ibmmq_runner#L11 for more context)

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/amqp10_client/src/amqp10_client.erl
+++ b/deps/amqp10_client/src/amqp10_client.erl
@@ -164,6 +164,8 @@ end_session(Pid) ->
 %% @doc Synchronously attach a link on 'Session'.
 %% This is a convenience function that awaits attached event
 %% for the link before returning.
+-spec attach_sender_link_sync(pid(), binary(), binary()) ->
+    {ok, link_ref()} | link_timeout.
 attach_sender_link_sync(Session, Name, Target) ->
     attach_sender_link_sync(Session, Name, Target, mixed).
 
@@ -275,7 +277,8 @@ attach_receiver_link(Session, Name, Source, SettleMode, Durability, Filter) ->
 -spec attach_receiver_link(pid(), binary(), binary(), snd_settle_mode(),
                            terminus_durability(), filter(), properties()) ->
     {ok, link_ref()}.
-attach_receiver_link(Session, Name, Source, SettleMode, Durability, Filter, Properties)
+attach_receiver_link(Session, Name, Source, SettleMode, Durability, Filter, 
+    Properties)
   when is_pid(Session) andalso
        is_binary(Name) andalso
        is_binary(Source) andalso
--- a/deps/amqp10_client/src/amqp10_client_session.erl
+++ b/deps/amqp10_client/src/amqp10_client_session.erl
@@ -833,8 +833,11 @@ send_attach(Send, #{name := Name, role := RoleTuple} = Args, {FromPid, _},
             #state{next_link_handle = OutHandle0, links = Links,
              link_index = LinkIndex} = State) ->
 
+    logger:warning("make_source ..."),
     Source = make_source(Args),
+    logger:warning("make_target ..."),
     Target = make_target(Args),
+    logger:warning("make properties ..."),    
     Properties = amqp10_client_types:make_properties(Args),
 
     {LinkTarget, InitialDeliveryCount, MaxMessageSize} =
@@ -866,7 +869,9 @@ send_attach(Send, #{name := Name, role := RoleTuple} = Args, {FromPid, _},
                             target = Target,
                             max_message_size = MaxMessageSize},
 
+    logger:warning("sending ..."),
     ok = Send(Attach, State),
+    logger:warning("sent ..."),
 
     Ref = make_link_ref(Role, self(), OutHandle),
     Link = #link{name = Name,
--- a/deps/amqp10_client/src/amqp10_client_session.erl
+++ b/deps/amqp10_client/src/amqp10_client_session.erl
@@ -255,6 +255,7 @@ unmapped(cast, {socket_ready, Socket}, State) ->
     {next_state, begin_sent, State1};
 unmapped({call, From}, {attach, Attach},
                       #state{early_attach_requests = EARs} = State) ->
+    logger:warning("on unmapped state, received call attach. Storing it in early_attach_requests"),
     {keep_state,
      State#state{early_attach_requests = [{From, Attach} | EARs]}}.
 
@@ -263,12 +264,13 @@ begin_sent(cast, #'v1_0.begin'{remote_channel = {ushort, RemoteChannel},
                                incoming_window = {uint, InWindow},
                                outgoing_window = {uint, OutWindow}} = Begin,
            #state{early_attach_requests = EARs} = State) ->
-    logger:warning("begin_sent send-attach "),
+    logger:warning("on state begin_sent, received v1_0.beging with remote channlel : ~p", [RemoteChannel]),
     State1 = State#state{remote_channel = RemoteChannel},
+    logger:warning("sending early attach requests ~p", [EARs]),
     State2 = lists:foldr(fun({From, Attach}, S) ->
-                                 logger:warning("begin_sent send-attach ~tp", [Attach]),
+                                 logger:warning("send early attach request ~tp", [Attach]),
                                  {S2, H} = send_attach(fun send/2, Attach, From, S),
-                                 logger:warning("begin_sent send-attach ~tp returning ~p", [Attach, H]),
+                                 logger:warning("sent attach request ~p with result ~p", [Attach, H]),
                                  gen_statem:reply(From, {ok, H}),
                                  S2
                          end, State1, EARs),
@@ -313,7 +315,7 @@ mapped(cast, #'v1_0.attach'{name = {utf8, Name},
                             max_message_size = MaybeMaxMessageSize} = Attach,
        #state{links = Links, link_index = LinkIndex,
               link_handle_index = LHI} = State0) ->
-
+    logger:warning("on mapped state, received v1_0.attach frame"),
     OurRoleBool = not PeerRoleBool,
     OurRole = boolean_to_role(OurRoleBool),
     LinkIndexKey = {OurRole, Name},
@@ -546,11 +548,8 @@ mapped({call, From},
     {keep_state, State, {reply, From, Res}};
 
 mapped({call, From}, {attach, Attach}, State) ->
-    logger:warning("amqp10_session: mapped send_attach ~p",
-                   [Attach]),
+    logger:warning("on mapped state, received call to attach ~p", [Attach]),
     {State1, LinkRef} = send_attach(fun send/2, Attach, From, State),
-    logger:warning("amqp10_session: mapped returned call attach ~p",
-                   [Attach]),
     {keep_state, State1, {reply, From, {ok, LinkRef}}};
 
 mapped({call, From}, Msg, State) ->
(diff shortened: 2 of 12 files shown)


deps/amqp10_client/src/amqp10_client_session.erl
```diff
@@ -771,6 +779,19 @@ filter_value_type({T, _} = V) when is_atom(T) ->
...
+    {utf8, Capabilities};
+translate_terminus_capabilities(CapabilitiesList) when is_list(CapabilitiesList) ->
+    {list, [filter_capability(V) || V <- CapabilitiesList]}.
+
+filter_capability(V) when is_binary(V) ->
+    {utf8, V};
+filter_capability({T, _} = V) when is_atom(T) ->
+    %% looks like an already tagged type, just pass it through
+    V.
+
```
ansd, line 794: Wrong AMQP types are used:
1. Capabilities are symbols, not strings.
2. Multiple values of a composite type are represented by an array, not by a list.

ansd (changes requested):
I suggest that we first test end-to-end that the RabbitMQ <-> IBM MQ integration works before merging this PR.
Right now, this PR adds a lot of boilerplate code that is not used and contains bugs.

deps/amqp10_client/src/amqp10_client.erl
```diff
@@ -20,14 +20,17 @@
...
          attach_sender_link_sync/4,
          attach_sender_link_sync/5,
+         attach_sender_link_sync/6,
          attach_receiver_link/3,
          attach_receiver_link/4,
          attach_receiver_link/5,
          attach_receiver_link/6,
          attach_receiver_link/7,
+         attach_receiver_link/8,
          attach_link/2,
```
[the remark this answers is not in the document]
MarcialRosales: I will look into using `attach_link`. When I was doing it I thought there could be another way of doing this like setting a structure like Attach_args and passing it . But the API already presented this pattern or design . I just followed it.

deps/amqp10_client/src/amqp10_client_session.erl
MarcialRosales: It is encoded as an array and as a symbol. See the tests https://github.com/rabbitmq/rabbitmq-server/blob/set-amqp10-capabilities/deps/amqp10_client/test/system_SUITE.erl#L935 where we send an array of symbols. 

The API exposes or accepts Erlang types from users where it accepts binaries (strings) and list of binaries but the library is serliazing/deserlizing it as an array of symbols .. It is more or less how we treat Filters ...

MarcialRosales: It is true that multiple values are encoded using `array` rather than `list`.

deps/amqp10_client/test/system_SUITE.erl
```diff
@@ -157,9 +177,11 @@ end_per_group(_, Config) ->
 %% -------------------------------------------------------------------
 
 init_per_testcase(_Test, Config) ->
+    ct:log("Setting per test case"),
```
ansd, line 180: These log statements seem to be useless. Can we remove them?

deps/amqp10_client/test/system_SUITE.erl
```diff
@@ -157,9 +177,11 @@ end_per_group(_, Config) ->
...
     case lists:keyfind(mock_port, 1, Config) of
         {_, Port} ->
             M = mock_server:start(Port),
+            ct:log("Setting mock server"),
```
ansd, line 184: These log statements seem to be useless. Can we remove them?

deps/amqp10_client/test/system_SUITE.erl
```diff
@@ -176,19 +198,20 @@ open_close_connection(Config) ->
...
-    {ok, Connection} = amqp10_client:open_connection(Hostname, Port),
-    {ok, Connection2} = amqp10_client:open_connection(OpnConf),
+    ct:log("Connecting to ~p", [OpenConf]),
+    {ok, Connection2} = amqp10_client:open_connection(OpenConf),
     receive
-        {amqp10_event, {connection, Connection2, opened}} -> ok
+        {amqp10_event, {connection, Connection2, opened}} -> ct:log("connection opened"), ok
     after 5000 -> exit(connection_timeout)
     end,
+    ct:log("Closing connection ..."),
```
ansd, line 212: These log statements seem to be useless. Can we remove them?

deps/amqp10_client/test/system_SUITE.erl
```diff
@@ -324,61 +347,107 @@ roundtrip_large_messages(Config) ->
...
-    ok = roundtrip(OpenConf, DataMb),
-    ok = roundtrip(OpenConf, Data8Mb),
-    ok = roundtrip(OpenConf, Data64Mb).
+    ok = roundtrip(OpenConf, [{body, DataKb}]),
+    ok = roundtrip(OpenConf, [{body, DataMb}]),
+    ok = roundtrip(OpenConf, [{body, Data8Mb}]),
+    ok = roundtrip(OpenConf, [{body, Data64Mb}]).
+
+basic_roundtrip_ibmmq(Config) ->
+    application:start(sasl),
```
ansd, line 356: Why do we need to start the Erlang `sasl` application here?
Is https://www.erlang.org/doc/apps/sasl/error_logging.html being confused with https://datatracker.ietf.org/doc/html/rfc4422 ?

ansd: I ran the test without this line, and the test succeeds.

deps/amqp10_client/test/system_SUITE.erl
```diff
@@ -324,61 +347,107 @@ roundtrip_large_messages(Config) ->
...
+                   role => {sender, #{address => Destination,
+                                      durable => unsettled_state,
+                                      capabilities => SenderCapabilities}},
+                   snd_settle_mode => settled,
+                   rcv_settle_mode => first,
+                   filter => #{},
+                   properties => #{}
+                   },
+    {ok, Sender} = amqp10_client:attach_link(Session, SenderAttachArgs),
+    %%await_link(Sender, credited, link_credit_timeout),
```
ansd, line 392: Delete commented code.

deps/amqp10_client/test/system_SUITE.erl
```diff
@@ -324,61 +347,107 @@ roundtrip_large_messages(Config) ->
...
+                   },
+    {ok, Receiver} = amqp10_client:attach_link(Session, ReceiverAttachArgs),
     {ok, OutMsg} = amqp10_client:get_msg(Receiver, 4 * 60_000),
     ok = amqp10_client:end_session(Session),
     ok = amqp10_client:close_connection(Connection),
 
     % ct:pal(?LOW_IMPORTANCE, "roundtrip message Out: ~tp~nIn: ~tp~n", [OutMsg, Msg]),
-    ?assertMatch(Props, amqp10_msg:properties(OutMsg)),
+    ActualProps = amqp10_msg:properties(OutMsg),
+    [ ?assertEqual(V, maps:get(K, ActualProps)) || K := V <- Props, K =/= creation_time],
```
ansd, line 430: The creation-time must not be changed by the broker as it's part of the properties which are part of the bare message and therefore immutable. If IBM MQ changes this creation time, this is a bug in IBM MQ. Can we maintain the assertion that creation time matches, and only omit this assertion for the buggy IBM MQ?

deps/amqp10_client/test/system_SUITE.erl
```diff
@@ -324,61 +347,107 @@ roundtrip_large_messages(Config) ->
...
     ?assertEqual([Body], amqp10_msg:body(OutMsg)),
     ok.
 
 filtered_roundtrip(OpenConf) ->
-    filtered_roundtrip(OpenConf, <<"banana">>).
+    filtered_roundtrip(OpenConf, []).
+
+filtered_roundtrip(OpenConf, Args) ->
+    Body = proplists:get_value(body, Args, <<"banana">>),
+    Destination = proplists:get_value(destination, Args, <<"test1">>),
```
ansd, line 444: Can we simplify this again as done prior to this PR?
Here, body is always `banana` and Destination is always `test1`.
There is no need to use `proplists:get_value/3` here.

ansd (changes requested):
I get sometimes the following error

```
system_SUITE > ibmmq > basic_roundtrip_ibmmq
    #1. {error,
            {shutdown,
                {gen_statem,call,
                    [<0.540.0>,
                     {attach,
                         #{name => <<"banana-sender">>,filter => #{},
                           properties => #{},rcv_settle_mode => first,
                           role =>
                               {sender,
                                   #{address => <<"DEV.QUEUE.3">>,
                                     capabilities => <<"queue">>,
                                     durable => unsettled_state}},
                           snd_settle_mode => settled}},
                     5000]}}}
```

when running

```
make -C deps/amqp10_client/ ct-system
```
locally on Ubuntu.
Might be worth checking this flake?

deps/amqp10_client/src/amqp10_client_session.erl
```diff
@@ -73,11 +73,14 @@
 -type rcv_settle_mode() :: first | second.
 
 -type terminus_durability() :: none | configuration | unsettled_state.
+-type terminus_capabilities() :: none | binary() | list().
```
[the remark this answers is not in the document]
MarcialRosales: nice refinement ! Thanks

deps/amqp10_client/test/system_SUITE.erl
MarcialRosales: Indeed . In fact, it is not needed in the function `basic_roundtrip` either. I removed it from there and the tests succeeds.

michaelklishin:
The forced push was a rebase.

ansd:
The build currently fails because IBM MQ can't be started:
```
Error: {error,1,
              "Docker image pivotalrabbitmq/ibm-mqadvanced-server-dev:9.3.5.1-amd64 ready\nUnable to find image 'pivotalrabbitmq/ibm-mqadvanced-server-dev:9.3.5.1-amd64' locally\ndocker: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.
```

deps/amqp10_client/src/amqp10_client_session.erl
```diff
@@ -771,6 +778,17 @@ filter_value_type({T, _} = V) when is_atom(T) ->
...
+translate_terminus_capabilities(Capabilities) when is_binary(Capabilities) ->
+    {symbol, Capabilities};
+translate_terminus_capabilities(CapabilitiesList) when is_list(CapabilitiesList) ->
+    {array, symbol, [filter_capability(V) || V <- CapabilitiesList]}.
+
+filter_capability(V) when is_binary(V) ->
+    {symbol, V};
+filter_capability(_) ->
+    %% Any other type is ignored
+    {}.
```
ansd, line 790: Is this code tested?
I'm pretty sure this tuple with 0 elements will crash later in `amqp10_binary_generator`.
See my comment in https://github.com/rabbitmq/rabbitmq-server/pull/11337#discussion_r1647195471

ansd (changes requested):

MarcialRosales:
This feature has already been incorporated via another PR.

PR #11338, opened 2024-05-28
Making make better

lhoguin:
See individual commits for details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ DEP_PLUGINS = rabbit_common/mk/rabbitmq-dist.mk \
 
 DISABLE_DISTCLEAN = 1
 
+ifeq ($(filter-out xref,$(MAKECMDGOALS)),)
 XREF_SCOPE = app deps
 
 # We add all the applications that are in non-standard paths
@@ -47,6 +48,7 @@ XREF_IGNORE = [ \
 
 # Include Elixir libraries in the Xref checks.
 xref: ERL_LIBS := $(ERL_LIBS):$(CURDIR)/apps:$(CURDIR)/deps:$(dir $(shell elixir --eval ":io.format '~s~n', [:code.lib_dir :elixir ]"))
+endif
 
 ifneq ($(wildcard deps/.hex/cache.erl),)
 deps:: restore-hex-cache-ets-file
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ PROJECT_DESCRIPTION = RabbitMQ Server
 # other components. If PROJECT_VERSION is unset, then an empty variable
 # is propagated and the default version will fallback to the default
 # value from rabbitmq-components.mk.
-export RABBITMQ_VERSION = $(PROJECT_VERSION)
+export RABBITMQ_VERSION := $(PROJECT_VERSION)
 
 # Release artifacts are put in $(PACKAGES_DIR).
 PACKAGES_DIR ?= $(abspath PACKAGES)
@@ -53,6 +53,12 @@ deps:: restore-hex-cache-ets-file
 endif
 
 include rabbitmq-components.mk
+
+# Set PROJECT_VERSION, calculated in rabbitmq-components.mk,
+# in stone now, because in this Makefile we will be using it
+# multiple times (including for release file names and whatnot).
+PROJECT_VERSION := $(PROJECT_VERSION)
+
 include erlang.mk
 include mk/github-actions.mk
 include mk/bazel.mk
--- a/deps/rabbit/Makefile
+++ b/deps/rabbit/Makefile
@@ -247,13 +247,6 @@ ifdef TRACE_SUPERVISOR2
 RMQ_ERLC_OPTS += -DTRACE_SUPERVISOR2=true
 endif
 
-ifndef USE_PROPER_QC
-# PropEr needs to be installed for property checking
-# http://proper.softlab.ntua.gr/
-USE_PROPER_QC := $(shell $(ERL) -eval 'io:format({module, proper} =:= code:ensure_loaded(proper)), halt().')
-RMQ_ERLC_OPTS += $(if $(filter true,$(USE_PROPER_QC)),-Duse_proper_qc)
-endif
-
 # --------------------------------------------------------------------
 # Documentation.
 # --------------------------------------------------------------------
--- a/deps/rabbit_common/mk/rabbitmq-dist.mk
+++ b/deps/rabbit_common/mk/rabbitmq-dist.mk
@@ -214,9 +214,11 @@ do-dist:: $(DIST_EZS)
 CLI_SCRIPTS_LOCK = $(CLI_SCRIPTS_DIR).lock
 CLI_ESCRIPTS_LOCK = $(CLI_ESCRIPTS_DIR).lock
 
+ifeq ($(MAKELEVEL),0)
 ifneq ($(filter-out rabbit_common amqp10_common rabbitmq_stream_common,$(PROJECT)),)
 app:: install-cli
 test-build:: install-cli
+endif
 endif
 
 install-cli: install-cli-scripts install-cli-escripts
--- a/deps/rabbit_common/mk/rabbitmq-dist.mk
+++ b/deps/rabbit_common/mk/rabbitmq-dist.mk
@@ -229,24 +229,12 @@ install-cli-scripts:
 	$(gen_verbose) \
 	if command -v flock >/dev/null; then \
 		flock $(CLI_SCRIPTS_LOCK) \
-		sh -c 'mkdir -p "$(CLI_SCRIPTS_DIR)" && \
-		for file in scripts/*; do \
-			cmp -s "$$file" "$(CLI_SCRIPTS_DIR)/$$(basename "$$file")" || \
-			cp -a "$$file" "$(CLI_SCRIPTS_DIR)/$$(basename "$$file")"; \
-		done'; \
+			sh -c 'mkdir -p "$(CLI_SCRIPTS_DIR)" && cp -a scripts/* $(CLI_SCRIPTS_DIR)/'; \
 	elif command -v lockf >/dev/null; then \
 		lockf $(CLI_SCRIPTS_LOCK) \
-		sh -c 'mkdir -p "$(CLI_SCRIPTS_DIR)" && \
-		for file in scripts/*; do \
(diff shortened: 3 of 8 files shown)


mkuratczyk:
Some modules can be hot reloaded, eg `c(rabbit_msg_store).` works for me after make run-broker, but some can't, for example:
```
(rabbit@K6L59PF0JR)4> c(rabbit_writer).
Recompiling /Users/mkuratczyk/rabbitmq/server/rabbit_writer/deps/rabbit_common/src/rabbit_writer.erl
/Users/mkuratczyk/rabbitmq/server/rabbit_writer/deps/rabbit_common/src/rabbit_writer.erl:29:10: can't find include file "rabbit.hrl"
%   29| -include("rabbit.hrl").
%     |          ^

/Users/mkuratczyk/rabbitmq/server/rabbit_writer/deps/rabbit_common/src/rabbit_writer.erl:188:6: undefined macro 'LG_PROCESS_TYPE/1'
%  188|     ?LG_PROCESS_TYPE(writer),
%     |      ^

/Users/mkuratczyk/rabbitmq/server/rabbit_writer/deps/rabbit_common/src/rabbit_writer.erl:32:2: function init/1 undefined
%   32| -export([init/1,
%     |  ^

/Users/mkuratczyk/rabbitmq/server/rabbit_writer/deps/rabbit_common/src/rabbit_writer.erl:422:10: record content undefined
%  422| msg_size(#content{payload_fragments_rev = PFR}) -> iolist_size(PFR);
%     |          ^

/Users/mkuratczyk/rabbitmq/server/rabbit_writer/deps/rabbit_common/src/rabbit_writer.erl:422:64: variable 'PFR' is unbound
%  422| msg_size(#content{payload_fragments_rev = PFR}) -> iolist_size(PFR);
%     |                                                                ^

/Users/mkuratczyk/rabbitmq/server/rabbit_writer/deps/rabbit_common/src/rabbit_writer.erl:423:10: record basic_message undefined
%  423| msg_size(#basic_message{content = Content})     -> msg_size(Content).
%     |          ^

/Users/mkuratczyk/rabbitmq/server/rabbit_writer/deps/rabbit_common/src/rabbit_writer.erl:423:61: variable 'Content' is unbound
%  423| msg_size(#basic_message{content = Content})     -> msg_size(Content).
%     |                                                             ^

error
```

This is with `PLUGINS_FROM_DEPS_DIR=1`. Without it, `c(rabbit_msg_store).` still works and `c(rabbit_writer).` just returns `{error, no_source}`

mkuratczyk:
Shortening OCI-based development/testing would be nice. Right now the OCI image is built by `make package-unix-generic && make docker-image`. Unfortunately:
`make package-unix-generic` takes a few minutes even without any changes whatsoever (just re-running the command)
`make docker-image` takes 40-60 seconds without any changes

So any modification of any file requires ~5 minutes of rebuilding the package and then the image to test the changes.
It'd be great if this could be reduced to seconds for a simple change (skip/short-circuit the `package-generic-unix` step).

For the time being, `bazel` offers a much faster alternative

michaelklishin:
This PR does speed up node startup with `gmake run-broker` and some 3-6 plugins (a very common workflow for me, with both Make and Bazel) significantly 👏

lhoguin:
I have a PoC patch locally that makes `make run-broker` basically instantaneous but that includes changes to Erlang.mk that only makes sense within the context of RabbitMQ: because RabbitMQ's applications are in deps/ instead of apps/. Normally deps/ shouldn't get rebuilt often like they are now. The patch is also slightly incorrect and will not pick up changes in deps of deps (if no changes were made to management, then changes to web dispatch won't be picked up).

If the applications were in apps/ then a proper fix could be made. The idea would be to have a "depend" file similar to the `$(PROJECT).d` file that would have rules or variables showing the dependency between applications in apps/. Then Erlang.mk could ensure that only the applications that have had changes were rebuilt, regardless of how deep they are in the dependency tree.

I don't plan to move the RabbitMQ files out of deps/ in this PR. That is too disruptive for now. But I will change the default to set `PLUGINS_FROM_DEPS_DIR=1` and `DIST_TARGET=all` so that changes do get picked up, but we avoid doing too much work (the DIST steps). And if `NOBUILD` is set then it can just launch the node immediately like before.

lhoguin:
I don't know why but the last commit no longer works for me. Investigating...

lhoguin:
Commit f6dd196 can be dropped once https://github.com/rabbitmq/rabbitmq-server/pull/10988 is merged.

lhoguin:
Starting the broker fully now takes less than 8s when no changes were made.

> make run-broker  4,79s user 3,16s system 108% cpu 7,297 total

If all is well I plan to merge this on Friday. I may do minor tweaks until then.

There's a lot more to do but I need to get some of my Erlang patches worked on as well. Including:

* Removing Mix usage (partially or fully)
* Reworking `package-generic-unix` for development, maybe via a new target
* Moving RabbitMQ applications to `apps/` and optimising `apps/` within Erlang.mk (long term), necessary to get rid of the "always recompile everything" problem

michaelklishin:
> Commit [f6dd196](https://github.com/rabbitmq/rabbitmq-server/commit/f6dd1960e9ea01edfc20351884642864b8f81189) can be dropped once #10988 is merged

#10988 was merged.

lhoguin:
With the two new commits I am now in the 4s to fully start the broker with `make -C deps/rabbit run-broker` and in the 6s with `make run-broker` which I believe includes all plugins.

dumbbell:
I didn't have a chance to test this pull request before it was merged as I was off for a few days. Unfortunately, it causes a regression with `gmake run-broker` and all testsuites that start a node (thus most of them).

Here is an example:

```
Error during startup: {error,
                       {rabbitmq_web_mqtt,
                        {bad_return,
                         {{rabbit_web_mqtt_app,start,[normal,[]]},
                          {'EXIT',
                           {undef,
                            [{rabbit_networking,ranch_refs,
                              [[{port,15675}]],
                              []},
                             {rabbit_web_mqtt_app,start_tcp_listener,2,
                              [{file,"rabbit_web_mqtt_app.erl"},{line,99}]},
                             {rabbit_web_mqtt_app,mqtt_init,0,
                              [{file,"rabbit_web_mqtt_app.erl"},{line,93}]},
                             {rabbit_web_mqtt_app,start,2,
                              [{file,"rabbit_web_mqtt_app.erl"},{line,36}]},
                             {application_master,start_it_old,4,
                              [{file,"application_master.erl"},
                               {line,293}]}]}}}}}}
```

This comes from the fact that the plugin's directory is set to the `deps` directory now. This directory is polluted with unwanted Erlang applications because it has all plugins even though I'm running e.g. `gmake -C deps/rabbit run-broker` (i.e. I explicitly start RabbitMQ without any plugins).

In this example, `rabbitmq_web_mqtt` was compiled a long time ago and is incompatible with the copy of `rabbit` I'm working with. But even after cleaning that plugin, I see that the broker is started with many plugins enabled (using old compiled copies), which is wrong based on what I expected. This may hide incorrectly defined dependencies or circular dependencies.

I see from the commit message that I can set `FAST_RUN_BROKER=0` to restore the correct behavior. Indeed, the broker works again with that. I will add that variable to my shell init file. I still believe this is the wrong default behavior because of the errors it will make invisible.

lhoguin:
I see, it starts everything it finds in the plugins directory. That was unexpected. Is there no way to specify what should be started instead? If not, we could add that. If an issue comes up with tests meanwhile we could add `FAST_RUN_BROKER=0` to them.

dumbbell:
> Is there no way to specify what should be started instead?

I'm not sure if this is what you are asking about, but there is a variable to set the list of plugins to start, `$RABBITMQ_ENABLED_PLUGINS`. It defaults to `ALL` if `$LEAVE_PLUGINS_DISABLED` is unset. However, this is not the same as plugins would still be discovered and their code available.

lhoguin:
A simple fix would be to set `FAST_RUN_BROKER=0` for tests. And set `RABBITMQ_ENABLED_PLUGINS=app` when we are running against `app`, not changing the behavior when in the top-level directory.

This won't be exactly the same as you said but halving the start time is probably more important for most people. Ideally we could use releases or at least `embedded` mode to load only what we want. This can be improved at a later time IMO.

lhoguin:
Something like this:

```
diff --git a/deps/rabbit_common/mk/rabbitmq-run.mk b/deps/rabbit_common/mk/rabbitmq-run.mk
index 7e09ae1e05..b97c5163e9 100644
--- a/deps/rabbit_common/mk/rabbitmq-run.mk
+++ b/deps/rabbit_common/mk/rabbitmq-run.mk
@@ -165,8 +165,15 @@ virgin-node-tmpdir:
 ifdef LEAVE_PLUGINS_DISABLED
 RABBITMQ_ENABLED_PLUGINS ?=
 else
+# When running "make -C deps/plugin run-broker" we only want
+# "plugin" to be enabled. See rabbitmq-components.mk for where
+# this variable comes from.
+ifdef deps_dir_overriden
+RABBITMQ_ENABLED_PLUGINS ?= $(PROJECT)
+else
 RABBITMQ_ENABLED_PLUGINS ?= ALL
 endif
+endif
 
 # --------------------------------------------------------------------
 # Run a full RabbitMQ.
diff --git a/deps/rabbitmq_ct_helpers/src/rabbit_ct_broker_helpers.erl b/deps/rabbitmq_ct_helpers/src>
index e1d6fd5327..9c6294e4b5 100644
--- a/deps/rabbitmq_ct_helpers/src/rabbit_ct_broker_helpers.erl
+++ b/deps/rabbitmq_ct_helpers/src/rabbit_ct_broker_helpers.erl
@@ -743,6 +743,7 @@ do_start_rabbitmq_node(Config, NodeConfig, I) ->
       {"RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS=+S 2 +sbwt very_short +A 24 ~ts", [AdditionalErlArgs]},
       "RABBITMQ_LOG=debug",
       "RMQCTL_WAIT_TIMEOUT=180",
+      "FAST_RUN_BROKER=0",
       {"TEST_TMPDIR=~ts", [PrivDir]}
       | ExtraArgs],
     Cmd = ["start-background-broker" | MakeVars],

lhoguin:
I opened https://github.com/rabbitmq/rabbitmq-server/pull/11438 with the fixes mentioned in the previous comment. Let's continue there!

PR #11342, opened 2024-05-29
Fix includes in many modules so that c(mod) works

lhoguin:
WIP


deps/amqp_client/src/amqp_client_internal.hrl
lhoguin, line 1: I have moved the file to `src/` rather than keeping it in `include/` since it's clearly marked as internal. Ideally internal include files would all be in `src/` and the public ones in `include/` (public with regards to other applications), but I am not investigating every single include file to make sure they should really be in `include/` as it's not always that obvious.

lhoguin:
TBH I am reconsidering most of these changes. The problem to be solved is enabling module reloading in run-broker, which may be achieved through a different method. I think I will see if that can be done there before I continue.

lhoguin:
I am dropping this PR in favor of the code reloading solution introduced in https://github.com/rabbitmq/rabbitmq-server/pull/11338

I still think better guidelines with regards to use of include files would be good to have. If time allows I will write a set of recommendations around that and perhaps other topics.

PR #11363, opened 2024-06-04
Periodically reconcile virtual host processes

michaelklishin:
for up to 10 times.

When a cluster is formed from scratch and a virtual host is declared in the process (can be via
definitions, or plugins, or any other way),
that virtual host's process tree will be started
on all reachable cluster nodes at that moment
in time.

However, this can be just a subset of all nodes
expected to join the cluster.

The most effective solution is to run this
reconciliation process on a timer for up to
5 minutes by default. This matches how long
some other parts of RabbitMQ (3.x) expect
cluster formation to take, at most.

Per discussion with @dcorbacho @mkuratczyk.


michaelklishin:
Investigating the `//deps/rabbit:dialyze` failure.

michaelklishin:
I will add a CLI command as discussed with @dcorbacho @mkuratczyk.

deps/rabbit/src/rabbit_vhosts.erl
```diff
@@ -0,0 +1,130 @@
...
+    reconcile/0,
+    reconcile_once/0,
+    start_processes_for_all/0,
+    start_on_all_nodes/2,
+    on_node_up/1
+]).
+
+%% Same as rabbit_vhost:exists/1.
+-spec exists(vhost:name()) -> boolean().
+exists(VirtualHost) ->
```
dcorbacho, line 29: Nothing against it, but why this function here? I can't see where it is being used.

dcorbacho: In fact, `exists/1`, and `list_names/1` already exist in `rabbit_vhost`. Why not use them?

mkuratczyk (approved):
Works as advertised.

deps/rabbit/src/rabbit_vhosts.erl
michaelklishin: Some are in `rabbit_vhost`, others are in the supervision tree-related modules. We need a place for all key functions.

There's also a comment at the top that explains why the module was originally introduced:

```
%% This module exists to avoid circular module dependencies between
%% several others virtual hosts-related modules.
```

michaelklishin:
Something is up with Actions, they are not picking up my latest changes. Will submit a new PR instead.

PR #11366, opened 2024-06-04
OTP27 support

mkuratczyk:
With these changes we should be able to run the broker and pass tests on OTP27. However, the OCI image is still built using OTP26 to allow easier performance regression testing vs 3.13, since we are aware of OTP27-specific regressions. I will maintain a [separate branch](https://github.com/rabbitmq/rabbitmq-server/tree/oci-otp27) for now with an OCI built using OTP27.

since `maybe` is now a keyword, it's quoted. However, I think we should consider renaming it. We already have a bunch of `opt` and `option` types in different modules that serve the same or very similar purpose (although they use `undefined` instead of `none`). We can unify this in a separate PR.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo_dlx_worker.erl
+++ b/deps/rabbit/src/rabbit_fifo_dlx_worker.erl
@@ -33,7 +33,7 @@
 %% gen_server callbacks
 -export([init/1, terminate/2, handle_continue/2,
          handle_cast/2, handle_call/3, handle_info/2,
-         code_change/3, format_status/2]).
+         code_change/3, format_status/1]).
 
 -define(HIBERNATE_AFTER, 4*60*1000).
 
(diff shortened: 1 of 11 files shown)


michaelklishin:
`rabbit_types:option/1` was added as a "keyword-safe" replacement for `rabbit_types:maybe/1`, so we should simply use it across the board.

deps/rabbit/src/rabbit_fifo_dlx_worker.erl
```diff
@@ -27,6 +27,7 @@
 -include_lib("rabbit_common/include/rabbit.hrl").
 % -include_lib("rabbit_common/include/rabbit_framing.hrl").
 
+-compile(nowarn_deprecated_callback).
```
lhoguin, line 30: This one should be simple enough to convert to `format_status/1`, just replace the `state` in the return value, with the map we are currently creating.

deps/rabbit/src/rabbit_tracking_store.erl
```diff
@@ -6,6 +6,7 @@
 %%
 -module(rabbit_tracking_store).
 
+-compile(nowarn_deprecated_callback).
```
lhoguin, line 9: The `format_status` here is empty it should be removed instead.

deps/rabbit_common/src/supervisor2.erl
```diff
@@ -60,6 +60,7 @@
 %%
 -module(supervisor2).
 
+-compile(nowarn_deprecated_callback).
```
lhoguin, line 63: We can safely drop this `format_status/2` as well instead. https://github.com/erlang/otp/blob/master/lib/stdlib/src/supervisor.erl doesn't implement it either.

lhoguin (changes requested):

lhoguin (approved):
Code review is good. Didn't test it but I trust you. 🥲

michaelklishin:
The forced push was a rebase.

mkuratczyk:
@mergify backport v3.13.x

PR #11377, opened 2024-06-05
Deflake AMQP connection close test

mkuratczyk:
The test expected GET /connections to not return
a connection immediately after it was closed. However, while the connection is closed synchronously during the DELETE request (the reader process is terminated), connection tracking is an asynchronous process and therefore the connection may still be listed for a brief moment after it was closed.

To deflake this test without redesigning connection tracking, we make sure that:
* the client connection process is no longer alive (it's closed because the server side of the connection is closed)
* /connections doesn't return this connection shorly after it was closed; however, we do allow this to take up to 50ms


ansd (approved):
Thank you!

PR #11389, opened 2024-06-06
Prefer monotonic time to measure durations

ansd:
Monotonic time instead of wall clock time should be used to measure a duration.

This is mostly a cosmetic change given that consumer timeouts are 30 mins by default.

See also
https://github.com/rabbitmq/khepri/issues/239
https://github.com/rabbitmq/rabbitmq-server/pull/10928

### DIFF
--- a/deps/rabbit/src/rabbit_channel.erl
+++ b/deps/rabbit/src/rabbit_channel.erl
@@ -117,7 +117,7 @@
                       delivery_tag,
                       %% consumer tag
                       tag,
-                      delivered_at,
+                      delivered_at :: integer(),
                       %% queue name
                       queue,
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):

michaelklishin (approved):

ansd:
`erlang:monotonic_time/1` is slower compared to `os:system_time/1`.
However this slowness is not noticeable with the following workload:
```
make run-broker FULL=1 RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+S 1"

java -jar target/perf-test.jar --predeclared --exchange amq.default -k cq -u cq -A 10000 --qos 10000 -z 90
```

PR #11392, opened 2024-06-06
Ensure rabbit_ff_controller doesn't block during boot

dcorbacho:
If it throws an exit, the supervisor calls terminate and blocks waiting for in-flight operations. Also, no information is provided in the logs.


michaelklishin (approved):

dcorbacho:
https://github.com/rabbitmq/rabbitmq-server/pull/11414 might be the right solution. Avoids the deadlock and the process/node terminates with an explanatory stacktrace

dumbbell:
I agree, #11414 seems a better solution. Closing this one.

PR #11411, opened 2024-06-07
Include vhost limits and metadata in definitions exported from the management UI

anhanhnguyen:
## Proposed Changes

The change is to include vhost limits and metadata in definitions exported from the management UI.

Fix the issue reported at #10515.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #10515)
- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbit/src/rabbit_vhost.erl
+++ b/deps/rabbit/src/rabbit_vhost.erl
@@ -604,6 +604,7 @@ i(cluster_state, VHost) -> vhost_cluster_state(vhost:get_name(VHost));
 i(description, VHost) -> vhost:get_description(VHost);
 i(tags, VHost) -> vhost:get_tags(VHost);
 i(default_queue_type, VHost) -> vhost:get_default_queue_type(VHost);
+i(limits, VHost) -> vhost:get_limits(VHost);
 i(metadata, VHost) -> vhost:get_metadata(VHost);
 i(Item, VHost)     ->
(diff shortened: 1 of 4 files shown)


michaelklishin:
Thank you but this PR won't be accepted (in its current form, the end goal sounds reasonable to me).

There are several serious omissions or incorrect assumptions:

 * You cannot just change the list of info keys willy-nilly, they usually exist for a reason
 * You have missed `rabbit_definitions` where you arguably should have started, therefore `rabbitmqctl export_definitions` may or may not return the limits
 * It does not add any tests to the definition import suites (both in core broker and in the management plugin)
 * Therefore it remains an open question whether such exported definitions can be imported back

michaelklishin:
Instead, limits should be exposed like so:

 * Start with `rabbit_definitions`, the management plugin code paths should arguably be removed by now and just use `rabbit_definitions`
 * [`definition_import_SUITE`](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/test/definition_import_SUITE.erl) and its [input cases](https://github.com/rabbitmq/rabbitmq-server/tree/main/deps/rabbit/test/definition_import_SUITE_data) should be covered first
 * Then test `rabbitmqctl export_definitions` and its importing counterpart manually
 * Test global definition import vs. single virtual host definition import
 * Then move on to the HTTP API

In addition, HTTP API clients — and there multiple popular ones — may have to be updated as some of them (Go, Ruby, Hop, Rust) include definition operations one way or another.

PR #11414, opened 2024-06-07
Avoid deadlock on terminating ff controller

dcorbacho:
If the global process is itself, it means it just crashed. It shouldn't wait but terminate.

As discussed with @dumbbell 

Probably supersedes https://github.com/rabbitmq/rabbitmq-server/pull/11392


dumbbell:
The patch looks good to me and I tested it successfully locally.

There are some (unrelated?) regressions in CI that we need to address before we can merge.

dumbbell:
@dcorbacho: Could you please prefix your commit message with `rabbit_feature_flags: `? You can then drop the `ff` part.

PR #11427, opened 2024-06-10
make restart-cluster

mkuratczyk:
Implement a rolling cluster restart similar
to what the Kubernetes Operator does


mkuratczyk:
@mergifyio backport v3.13.x

PR #11436, opened 2024-06-12
Remove most of the fd related FHC code

lhoguin:
Stats were not removed YET, including management UI stats relating to FDs.

Web-MQTT and Web-STOMP configuration relating to FHC were not removed.

The file_handle_cache itself must be kept until we remove CQv1.


lhoguin:
Thanks. There's likely a bunch of stuff broken regardless of flake improvements. I'm currently looking into some of them.

lhoguin:
Next steps are:

* [ ] Fix the overview Grafana dashboard as one of the values it uses was removed
* [ ] Ensure the clients are still working with the value that was removed (Java, Ruby, Go rabbit-hole, Rust, and .NET)

I will remove the `file_handle_cache_stats` module in a separate PR and replace it with a CQv2-specific stats module.

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -119,21 +118,6 @@
...
-warn_file_limit() ->
-    DurableQueues = find_recoverable_queues(),
-    L = length(DurableQueues),
-
-    %% if there are not enough file handles, the server might hang
-    %% when trying to recover queues, warn the user:
-    case file_handle_cache:get_limit() < L of
-        true ->
-            rabbit_log:warning(
-              "Recovering ~tp queues, available file handles: ~tp. Please increase max open file handles limit to at least ~tp!",
```
lhoguin, line 131: Maybe there can still be a warning at boot-up time that the FD limit looks low. If the limit is =< 1024 for example it's likely that the limit needs to be increased.

lhoguin:
Rebased, and sent a PR to the few clients noted previously.

This can be merged when green.

michaelklishin (approved):

Issue #11421, opened 2024-06-10
Load oauth javascript dependencies only when they are required


MarcialRosales:
### Is your feature request related to a problem? Please describe.

The management UI always load two javascript files only required when OAuth is enabled and required in the UI. 
The two files are : `js/oidc-oauth/helper.js` and `js/oidc-oauth/oidc-client-ts.js` with a total of 119kbytes. 

The management UI loads a third javascript file called `js/oidc-oauth/bootstrap.js`. But this is a javascript file generated by the management plugin which tells whether Oauth is enabled and required in the ui. Thus the management ui always need to load this file. When OAuth is disabled or not required in the ui, this file is very small, a few bytes long.

### Describe the solution you'd like

Load the first two javascript files only when OAuth in enabled in the management ui

PR #11448, opened 2024-06-13
Dynamically load oauth-related libraries 

MarcialRosales:
## Proposed Changes

It addresses one issue raised and two more issues found but not raised:
- https://github.com/rabbitmq/rabbitmq-server/issues/11421
- the  logged in user's name should only be a hyperlink to the user's record when the user is an administrator and it has been authenticated with the internal backend. In other words, if the user was authenticated with oauth or ldap or http, there is no internal record of the user hence the username should not be rendered as a hyperlink.
- disable browser login dialog-box and instead always redirect the user to the login html page when either the credentials are wrong or user is not authorized  or session has expired  

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/rabbitmq_management/priv/www/index.html
+++ b/deps/rabbitmq_management/priv/www/index.html
@@ -22,7 +22,7 @@
     <link href="favicon.ico" rel="shortcut icon" type="image/x-icon"/>    
     
     <script type="module">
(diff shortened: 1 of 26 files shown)


michaelklishin:
The forced push was a rebase.

PR #11455, opened 2024-06-14
Reduce default maximum message size to 16MB

### LINKED ISSUE #11187: 4.x: reduce default maximum message size further (e.g. to 64 or 50 MiB)
This is a follow-up to https://github.com/rabbitmq/rabbitmq-server/pull/1812.

We can and should lower the limit from its current 128 MiB. The idea in #1812 was to gradually lower it from its original 512 MiB default but we have missed the opportunity to go to, say, 64 MiB in 3.13.x. Oh well.

I guess 4.0 can use a 64 MiB or a 50 MiB limit.

Background: https://github.com/rabbitmq/rabbitmq-server/issues/11072.

dcorbacho:
Following this discussion https://github.com/rabbitmq/rabbitmq-server/issues/11187 we are setting the default maximum message size to 16MB.

This is just a default, and can be increased or decreased through the configuration setting `rabbit.max_message_size`.

Closes #11187.

### DIFF
--- a/deps/amqp10_client/test/system_SUITE.erl
+++ b/deps/amqp10_client/test/system_SUITE.erl
@@ -105,14 +105,20 @@ stop_amqp10_client_app(Config) ->
 init_per_group(rabbitmq, Config0) ->
     Config = rabbit_ct_helpers:set_config(Config0,
                                           {sasl, {plain, <<"guest">>, <<"guest">>}}),
-    rabbit_ct_helpers:run_steps(Config, rabbit_ct_broker_helpers:setup_steps());
+    Config1 = rabbit_ct_helpers:merge_app_env(Config,
+                                              [{rabbit,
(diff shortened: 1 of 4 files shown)


michaelklishin:
Given the heat of the discussion in #11187, we need to mention why the new value was set to 16 MiB. It is

 * Roughly the mid-point of what a few users with (very) large RabbitMQ cluster fleets and our team have considered
 * A value that CQv2 (maybe by then called CQv3) will handle well and reasonably efficiently as of 4.0 (per discussion with @lhoguin)

ansd:
The 4.0 docs do not reflect this change: https://www.rabbitmq.com/docs/configure#config-items

ansd:
Created https://github.com/rabbitmq/rabbitmq-website/pull/2073

PR #11477, opened 2024-06-18
MQTT: speed up shared_SUITE:many_qos1_messages

kjnilsson:
speed up tests.


michaelklishin (approved):
Note that some MQTT 3.1 (v3) tests were intentionally removed from the Erlang suites to reduce the time of their runs but that older version is still tested in the Java suite.

PR #11479, opened 2024-06-18
Use emqx/emqtt instead of a fork

mkuratczyk:


MODULE.bazel
```diff
@@ -374,7 +374,7 @@ erlang_dev_package.git_package(
     name = "emqtt",
     branch = "master",
     build_file = "@rabbitmq-server//bazel:BUILD.emqtt",
-    repository = "rabbitmq/emqtt",
+    repository = "emqx/emqtt",
```
ansd, line 377: Can we point to the latest tag https://github.com/emqx/emqtt/releases/tag/1.11.0 instead of `master` branch? This ensures RabbitMQ tests won't suddenly fail if `master` branch of emqx/emqtt gets updated.

ansd (changes requested):

MODULE.bazel
ansd: https://github.com/rabbitmq/rabbitmq-server/pull/7379/files shows how to use a tag instead of a branch

ansd (approved):
Thank you!

PR #11486, opened 2024-06-19
make: Don't propagate FULL=1 to "make test-dist"

lhoguin:
Both FULL and MAKEFLAGS env variables need to be unset as FULL=1 is present in both. This is a bit of a band-aid, it's possible that other variables get propagated that shouldn't be, but we'll fix them when they are detected.


ansd (approved):
```
make -C deps/rabbit ct-amqp_client t=[cluster_size_1]:multiple_sessions FULL=1
```
skipped the test prior to this PR with:
```
amqp_client_SUITE > cluster_size_1
    {skip,"rabbitmqctl required, please set RABBITMQCTL or 'rabbitmqctl_cmd' in ct config"}
```
but succeeds after this PR.

lhoguin:
JSP gave the OK in Slack.

PR #11507, opened 2024-06-20
Revert "rabbit_feature_flags: Retry after erpc:call() fails with `noconnection`"

dumbbell:
This reverts commit 8749c605f5d37112529df565201507f1bd4b19ae.

## Why

The patch was supposed to solve an issue that we didn't understand and that was likely a network/DNS problem outside of RabbitMQ. We know it didn't solve that issue because it was reported again 6 months after the initial pull request (#8411).

What we are sure however is that it increased the testing of RabbitMQ significantly because the code loops for 10+ minutes if the remote node is not running.

The retry in the Feature flags subsystem was not the right place either. The `noconnection` error is visible there because it runs earlier during RabbitMQ startup. But retrying there won't solve a network issue magically.

There are two ways to create a cluster:
1. peer discovery and this subsystem takes care of retries if necessary and appropriate
2. manually using the CLI, in which case the user is responsible for starting RabbitMQ nodes and clustering them

Let's revert it until the root cause is really understood.

### DIFF
--- a/deps/rabbit/src/rabbit_ff_controller.erl
+++ b/deps/rabbit/src/rabbit_ff_controller.erl
@@ -1390,32 +1390,9 @@ this_node_first(Nodes) ->
       Ret :: term() | {error, term()}.
 
 rpc_call(Node, Module, Function, Args, Timeout) ->
-    SleepBetweenRetries = 5000,
-    T0 = erlang:monotonic_time(),
     try
         erpc:call(Node, Module, Function, Args, Timeout)
     catch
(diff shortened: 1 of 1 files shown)


kjnilsson:
Additional observations: 

1. a clearly wrong command like `rabbitmqctl join_node this-is-not-a-node@argh` takes over a minute to return. This is bad UX.
2. It makes an already slow test suite: `clustering_managment_SUITE` take longer than it needs to. There are two tests that test this functionality and each take over a minute to complete.

kjnilsson (approved):

PR #11509, opened 2024-06-20
make: Fix circular dependency

lhoguin:
Running "make -C deps/rabbit ct-feature_flags_v2" on a clean repository would not work due to a circular dependency. Since the plugin in question is a no-op in 4.x, only used in tests, we make the plugin depend on nothing to avoid the issue.

This shouldn't be backported to 3.x because there this is a real plugin.


mkuratczyk (approved):

dumbbell:
What was the circular dependency?

lhoguin:
@dumbbell rabbit -> rabbitmq_amqp1_0 -> rabbit. Normally it doesn't cause issues but in this case it did because we wanted to build rabbit with TEST=1 and we compiled the dependency rabbit without TEST=1 first. The solution only works around the problem.

PR #11520, opened 2024-06-21
Fix regression introduced in 21619bba402

lhoguin:
Plugins must depend on 'rabbit' per JSP. But we don't want to build it. So the better solution is to use LOCAL_DEPS.


ansd (approved):

ansd:
Makes test
```
make -C deps/rabbit ct-amqp_client t=[cluster_size_1]:plugin
```
green again.

PR #11523, opened 2024-06-21
#11522 conflict resolution take 1

michaelklishin:
This is an attempt at resolving conflicts in #11522.


michaelklishin:
Some mixed version HTTP API tests fail with this resolution => I must have missed something, closing.

PR #11533, opened 2024-06-22
Remove Bloom filter data section for message delivery command in RabbitMQ Streams documentation

wrobell:
Also update OsirisChunk entry to reflect current state.


michaelklishin:
FTR, some of the places where this can be verified:

1. [`osiris_log`](https://github.com/rabbitmq/osiris/blame/main/src/osiris_log.erl#L111-L165)
2. [`osiris_log`](https://github.com/rabbitmq/osiris/blob/299495815a101b6c9471da13bdb80d203017dc24/src/osiris_log.erl#L173-L240) (specifically for chunks)

PR #11541, opened 2024-06-24
Make 'queue.declare' aware of virtual host DQT at validation time

michaelklishin:
DQT = default queue type.

Determining what queue type should be used is getting more and more involved.

It can come from four different places:

1. The client (as an `x-queue-type` argument value) or the HTTP API
2. Virtual host metadata
3. Node-wide default from `rabbitmq.conf`
4. If all else fails, the fallback (`"classic"`)

In addition, there are two ways to specify a type:

1. Clients use shortcuts such as `"quorum"`
2. Definitions and many places internally use module names such as `rabbit_classic_queue`

When a client provides no queue type, validation
should take the defaults (virtual host, global,
and the last resort fallback) into account
instead of considering the type to be "undefined".

References #11457 ##11528.


michaelklishin:
@mergifyio backport v3.13.x

michaelklishin:
Backported to `v3.13.x` manually.

michaelklishin:

PR #11547, opened 2024-06-24
Speed up rabbit tests suites

kjnilsson:
Improvements to many test suites in `deps/rabbit` see individual commits.


michaelklishin (approved):

deps/rabbit/.DS_Store
lhoguin, line 1: Careful not to include these. :)

PR #11549, opened 2024-06-24
Various make cleanup/consolidation

lhoguin:
The goal is to remove what we don't need, convert custom code to stock Erlang.mk features <s>(such as Hex.pm support which is now available in Erlang.mk)</s>, and consolidate the make code in one place now that we have a monorepo (this will make moving applications from deps/ to apps/ easier later on as well).


lhoguin:
This should be looked into as part of this PR: https://github.com/rabbitmq/rabbitmq-server/pull/11558#issuecomment-2191179383

PR #11555, opened 2024-06-24
Tracing: expose directory path setting as tracing.dir

### LINKED ISSUE #11554: Tracing: expose configuration to rabbitmq.conf
This is only really relevant if the plugin stays in the distribution (https://github.com/rabbitmq/rabbitmq-server/issues/11553).

`rabbitmq.conf` schema supports path validators. The tracing plugin only has one configurable setting which is a path, and it can be validated to be writeable similarly to how [TLS key and certificate paths are validated to be readable](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/priv/schema/rabbit.schema#L2694-L2709).

In fact, there's already a schema validator for that, `dir_writable`, which is used by the [log file path setting](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/priv/schema/rabbit.schema#L1611-L1613).

See https://github.com/rabbitmq/rabbitmq-server/discussions/11546 and https://github.com/rabbitmq/rabbitmq-server/issues/11551 for background.

michaelklishin:
in `rabbitmq.conf` using the same validator `log.dir` uses, and a similar
configuration key:

``` ini
tracing.dir = /path/to/a/writeable/directory
```

The error produced when the directory specified is not
writeable looks like this:

```
[error] <0.164.0> tracing.dir invalid, Directory must be writable
[error] <0.164.0> Error preparing configuration in phase validation:
[error] <0.164.0>   - tracing.dir invalid, Directory must be writable
```

Note that the path setting is not covered by the tests because
the test would have to use a writeable directory,
and it is not obvious what kind of cross-platform
path that is not computed programmatically they
could use.

It's a trivial schema file that uses an existing
core validator => let's leave it as is.

Closes #11554.

### DIFF
--- a/deps/rabbitmq_tracing/BUILD.bazel
+++ b/deps/rabbitmq_tracing/BUILD.bazel
@@ -90,6 +90,10 @@ eunit(
 broker_for_integration_suites()
 
 rabbitmq_integration_suite(
+    name = "config_schema_SUITE",
+)
+
+rabbitmq_integration_suite(
     name = "rabbit_tracing_SUITE",
 )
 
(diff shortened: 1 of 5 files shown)


razvanphp:
Thank you very much for taking the time to do this, really appreciated.

PR #11558, opened 2024-06-25
make: Better restrict FAST_RUN_BROKER

lhoguin:
Enable by default only when run-broker, run-tls-broker, start-brokers or start-cluster is explicitly used.

Should avoid issues with running the broker in tests.

I will work on some refactoring to ensure we can easily iterate dev run-broker without risking breaking tests.

### DIFF
--- a/deps/rabbit_common/mk/rabbitmq-run.mk
+++ b/deps/rabbit_common/mk/rabbitmq-run.mk
@@ -165,14 +165,7 @@ virgin-node-tmpdir:
 ifdef LEAVE_PLUGINS_DISABLED
 RABBITMQ_ENABLED_PLUGINS ?=
 else
-# When running "make -C deps/plugin run-broker" we only want
-# "plugin" to be enabled. See rabbitmq-components.mk for where
-# this variable comes from.
-ifdef deps_dir_overriden
-RABBITMQ_ENABLED_PLUGINS ?= $(filter-out rabbit,$(PROJECT))
-else
 RABBITMQ_ENABLED_PLUGINS ?= ALL
-endif
 endif
 
 # --------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


michaelklishin (approved):
Thank you, @lhoguin.

With this change,

``` bash
gmake run-broker RABBITMQ_ENABLED_PLUGINS="rabbitmq_federation,rabbitmq_stomp"
```

works the way

```
# this is a format that I used to use before Bazel
gmake run-broker PLUGINS="rabbitmq_federation rabbitmq_stomp"
```

used to, that is, the node is started from [the root] with only two plugins.

The comma-separated list of plugins mimics the Bazel convention and that of the `enabled_plugins` file, so I'm actually about the "format" change.

However, just for the sake of clarify, if I use

``` bash
# uses a space for plugin separator, by mistake
gmake run-broker RABBITMQ_ENABLED_PLUGINS="rabbitmq_federation rabbitmq_stomp"
```

then the node is started with zero plugins. This is not wrong
but should be at least pointed out. Maybe an error would be
a better outcome but I'm fine with either. For example,
if you generate an `enabled_plugins` file with an incorrect
format, the node will fail to parse it and as a consequence, fail to start.

lhoguin:
Right, `PLUGINS` is still around and works as you noted, with spaces. But it's only used when building a proper release in the top-level Makefile. It is used to create a plugins.mk file that contains the `PLUGINS` variable, and is included in the source-dist. I assume something somewhere makes use of it.

Perhaps we can add support for `PLUGINS` when `FAST_RUN_BROKER=1` as that is a more natural way of specifying plugins than comma lists.

lhoguin:
I will drop this PR in favor of what I am currently working on. I have managed to make the old method of running the broker about as fast if not faster than `FAST_RUN_BROKER` when `-j8` is used (parallel make). Therefore `FAST_RUN_BROKER` is no longer necessary. I will push that change in my other PR https://github.com/rabbitmq/rabbitmq-server/pull/11549 and look into honoring `PLUGINS` there as well.

PR #11565, opened 2024-06-25
Deprecate queue-master-locator

mkuratczyk:
* this should not be a breaking change - all validation should still pass
* CQs can now use `queue-leader-locator`
* `queue-leader-locator` takes precedence over `queue-master-locator` if both are used
* regardless of which name is used, effectively there are only two values: `client-local` (default) or `balanced`
* other values (`min-masters`, `random`, `least-leaders`) are mapped to `balanced`
* exclusive queues are always declared locally, as before


michaelklishin:
@mkuratczyk I have added the new suite to Bazel.

ansd (commented):
Since this PR deprecates `queue-master-locator`, how can we in future remove support for `queue-master-locator`? Should `queue-master-locator` be placed behind a deprecated feature flag such that removal can be enforced in a future RabbitMQ version?

mkuratczyk:
We can, even now, but I wouldn't rush with actually enforcing anything. The amount of code needed to support the old name is tiny and there can be applications that have that name in the codebase and we all know that for some users, application code changes are very hard (this particular setting is probably much more common in policies, that are easier to change, but still).

ansd:
> We can, even now, but I wouldn't rush with actually enforcing anything.

No, we can't remove support for `queue-master-locator` right now because - as you wrote - applications use `queue-master-locator`.
The question is how can we eventually remove support for `queue-master-locator` in RabbitMQ (e.g. version 5.0)? The deprecated feature subsystem enables "a way to declare deprecated features in the code, not only in our communication.".

Users might not read the documentation and might not even understand that `queue-master-locator` is deprecated.
As explained in https://github.com/rabbitmq/rabbitmq-server/pull/7390#issue-1595515010 deprecation of `queue-master-locator` would go through the following lifecycle:
* permitted_by_default
* denied_by_default
* disconnected
* removed

We already use this deprecated feature subsystem for various deprecated features.
It might be a good idea to also "properly" deprecate `queue-master-locator` using the deprecated feature subsystem.

dumbbell:
As I said on Slack, I also favor the use of deprecated features because it increases the chance that the information reaches the user who rely on that feature. It also allows them to test RabbitMQ as if the feature was gone.

deps/rabbit/src/rabbit_queue_location.erl
```diff
@@ -23,26 +32,33 @@
 queue_leader_locators() ->
     ?QUEUE_LEADER_LOCATORS.
 
--spec select_leader_and_followers(amqqueue:amqqueue(), pos_integer()) ->
+-spec select_leader_and_followers(amqqueue:amqqueue(), pos_integer(), atom()) ->
     {Leader :: node(), Followers :: [node()]}.
-select_leader_and_followers(Q, Size)
-  when (?amqqueue_is_quorum(Q) orelse ?amqqueue_is_stream(Q)) andalso is_integer(Size) ->
+select_leader_and_followers(Q, Size, QueueType)
```
ansd, line 37: Argument `QueueType` is redundant.
The queue type is already stored in the `Q` and can be queried using `amqqueue:get_type(Q)` as done in the deleted code.

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -502,15 +521,15 @@ recover_durable_queues(QueuesAndRecoveryTerms) ->
 capabilities() ->
     #{unsupported_policies => [%% Stream policies
                                <<"max-age">>, <<"stream-max-segment-size-bytes">>,
-                               <<"queue-leader-locator">>, <<"initial-cluster-size">>,
```
ansd, line 505: Can we add an integration that the new `queue-leader-locator` policy actually works with classic queues?

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -502,15 +521,15 @@ recover_durable_queues(QueuesAndRecoveryTerms) ->
...
+                               <<"initial-cluster-size">>,
                                %% Quorum policies
                                <<"delivery-limit">>, <<"dead-letter-strategy">>, <<"max-in-memory-length">>, <<"max-in-memory-bytes">>, <<"target-group-size">>],
       queue_arguments => [<<"x-expires">>, <<"x-message-ttl">>, <<"x-dead-letter-exchange">>,
                           <<"x-dead-letter-routing-key">>, <<"x-max-length">>,
                           <<"x-max-length-bytes">>, <<"x-max-priority">>,
                           <<"x-overflow">>, <<"x-queue-mode">>, <<"x-queue-version">>,
                           <<"x-single-active-consumer">>, <<"x-queue-type">>,
-                          <<"x-queue-master-locator">>],
+                          <<"x-queue-master-locator">>, <<"x-queue-leader-locator">>],
```
ansd, line 532: Can we add an integration that the new `x-queue-leader-locator` queue argument actually works with classic queues?

ansd (changes requested):

deps/rabbit/src/rabbit_queue_location.erl
```diff
@@ -63,40 +87,57 @@ leader_locator0(<<"client-local">>) ->
...
 leader_locator0(<<"least-leaders">>) ->
     <<"balanced">>;
+leader_locator0(<<"min-masters">>) ->
+    <<"balanced">>;
 leader_locator0(_) ->
     %% default
     <<"client-local">>.
 
--spec select_replicas(pos_integer(), [node(),...], [node(),...],
+-spec select_members(pos_integer(), term(), [node(),...], [node(),...],
```
ansd, line 101: Use more specific type, e.g. `module()`, instead of `term()`.

deps/rabbit/src/rabbit_queue_location.erl
```diff
@@ -63,40 +87,57 @@ leader_locator0(<<"client-local">>) ->
...
     {AllNodes, Fun};
+%% Classic queues: above the threshold, pick a random node
+%% For classic queues, when there's a lot of queues, if we knew that the
+%% distribution of queues between nodes is relatively even, it'd be better
+%% to declare this queue locally rather than randomly. However, currently,
+%% counting queues on each node is relatively expensive. Users can use
+%% the client-local strategy if they know their connections are well balanced
+select_members(1, rabbit_classic_queue, _AllNodes, RunningNodes, QueueCount, QueueCountStartRandom, GetQueues)
+  when QueueCount >= QueueCountStartRandom ->
+    {RunningNodes, GetQueues};
```
ansd, line 115: This whole function clause can be deleted given that the next function clause head matches for classic queues and returns the same value.

In general, it seems classic queue node selection can be a bit simpler than done in this PR.

ansd (changes requested):

deps/rabbit/src/rabbit_queue_location.erl
mkuratczyk: Clause removed. I also reverted one complication I introduced: since classic queues are handled by a dedicated function clause, there was no need to have classic queue support in a clause later on (when calculation `Counters`)

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -63,14 +64,61 @@
...
+validate_policy(Args) ->
+    Strategy = proplists:get_value(<<"queue-master-locator">>, Args, unknown),
+    case rabbit_queue_location:master_locator_permitted() of
+        true ->
+            case lists:member(Strategy, rabbit_queue_location:queue_leader_locators()) of
+                true -> ok;
+                false -> {error, "~tp is not a valid master locator", [Strategy]}
+            end;
+        false ->
+            {error, "use of deprecated queue-master-locator argument is not permitted", []}
```
ansd, line 89: Should we say here `policy` instead of `argument` given that this validates a policy and not a queue argument?

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -63,14 +64,61 @@
...
+
 -spec is_enabled() -> boolean().
 is_enabled() -> true.
 
 -spec is_compatible(boolean(), boolean(), boolean()) -> boolean().
 is_compatible(_, _, _) ->
     true.
 
+validate_arguments(Args) ->
+    case lists:keyfind(<<"x-queue-master-locator">>, 1, Args) of
```
ansd, line 100: `lists:keymember/3` is a better fit here since you don't use the returned tuple value.

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -63,14 +64,61 @@
...
+                    error
+            end
+    end.
+
 declare(Q, Node) when ?amqqueue_is_classic(Q) ->
+    case validate_arguments(amqqueue:get_arguments(Q)) of
+        ok -> do_declare(Q, Node);
+        error ->
+            Warning = rabbit_deprecated_features:get_warning(
+                        queue_master_locator),
```
ansd, line 117: It's better to return this warning and `protocol_error` directly in `validate_arguments/1` since right now `queue_master_locator` validation leaks out from function `validate_arguments/1` into `declare/2`. In other words, it's better when function `declare/2` doesn't know about `queue_master_locator`.

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -509,15 +555,18 @@ recover_durable_queues(QueuesAndRecoveryTerms) ->
...
+                               <<"initial-cluster-size">>,
                                %% Quorum policies
                                <<"delivery-limit">>, <<"dead-letter-strategy">>, <<"max-in-memory-length">>, <<"max-in-memory-bytes">>, <<"target-group-size">>],
       queue_arguments => [<<"x-expires">>, <<"x-message-ttl">>, <<"x-dead-letter-exchange">>,
                           <<"x-dead-letter-routing-key">>, <<"x-max-length">>,
                           <<"x-max-length-bytes">>, <<"x-max-priority">>,
                           <<"x-overflow">>, <<"x-queue-mode">>, <<"x-queue-version">>,
                           <<"x-single-active-consumer">>, <<"x-queue-type">>,
-                          <<"x-queue-master-locator">>],
+                          <<"x-queue-leader-locator">>] ++ case rabbit_queue_location:master_locator_permitted() of
```
ansd, line 566: When
```
deprecated_features.permit.queue_master_locator = false
```
is set, this line leads to the following error being logged by `rabbit_queue_type` module:
```
2024-07-08 09:46:56.396527+02:00 [error] <0.1265.0> Deprecated features: `queue_master_locator`: Feature `queue_master_locator` is deprecated.
2024-07-08 09:46:56.396527+02:00 [error] <0.1265.0> Its use is not permitted per the configuration (overriding the default, which is permitted):
2024-07-08 09:46:56.396527+02:00 [error] <0.1265.0>     "deprecated_features.permit.queue_master_locator = false"
2024-07-08 09:46:56.396527+02:00 [error] <0.1265.0> Its use will not be permitted by default in a future minor RabbitMQ version and the feature will be removed from a future major RabbitMQ version; actual versions to be determined.
2024-07-08 09:46:56.396527+02:00 [error] <0.1265.0> To continue using this feature when it is not permitted by default, set the following parameter in your configuration:
2024-07-08 09:46:56.396527+02:00 [error] <0.1265.0>     "deprecated_features.permit.queue_master_locator = true"
```
which is poor user experience as there is actually no error.
`rabbit_deprecated_features:is_permitted/1` assumes only to be called if the old feature is actually used. However, here, this function is also called when the old feature is not used.

ansd (changes requested):

deps/rabbit/test/classic_queue_SUITE.erl
```diff
@@ -0,0 +1,133 @@
...
+                                              {<<"x-queue-leader-locator">>, longstr, <<"client-local">>}])),
+                 {ok, Leader0} = rabbit_ct_broker_helpers:rpc(Config, Server, rabbit_amqqueue, lookup, [rabbit_misc:r(<<"/">>, queue, Q)]),
+                 Leader = amqqueue:qnode(Leader0),
+                 ?assertEqual(Server, Leader),
+                 ?assertMatch(#'queue.delete_ok'{},
+                              amqp_channel:call(Ch, #'queue.delete'{queue = Q}))
+             end || Server <- Servers] end.
+
+leader_locator_balanced(Config) ->
+    case rabbit_ct_helpers:is_mixed_versions() of
```
ansd, line 85: see above

deps/rabbit/test/classic_queue_SUITE.erl
```diff
@@ -0,0 +1,133 @@
...
+    rabbit_ct_helpers:run_steps(Config,
+                                rabbit_ct_client_helpers:teardown_steps() ++
+                                rabbit_ct_broker_helpers:teardown_steps()).
+
+%% -------------------------------------------------------------------
+%% Testcases.
+%% -------------------------------------------------------------------
+
+leader_locator_client_local(Config) ->
+    case rabbit_ct_helpers:is_mixed_versions() of
```
ansd, line 64: `rabbit_ct_helpers:is_mixed_versions/0` should not be called because this test will be skipped in the future for example in 4.1 / 4.0 mixed version clusters, which is not what we want.
Instead, you could call `rabbit_classic_queue:capabilities/0` and check whether `x-queue-leader-locator` is supported by the node.

ansd (changes requested):

deps/rabbit/test/classic_queue_SUITE.erl
```diff
@@ -0,0 +1,136 @@
...
+                rabbit_ct_client_helpers:setup_steps()),
+    Config2.
+
+end_per_group(_, Config) ->
+    rabbit_ct_helpers:run_steps(Config,
+                                rabbit_ct_client_helpers:teardown_steps() ++
+                                rabbit_ct_broker_helpers:teardown_steps()).
+
+init_per_testcase(T, Config) ->
+    case rpc(Config, rabbit_feature_flags, is_enabled, [classic_queue_leader_locator]) of
```
dumbbell, line 64: You should enable the feature flag, not check if it is already enabled.

The reason is that in mixed-version testing, nodes are started with feature flags disabled (except required ones).

dumbbell (changes requested):

deps/rabbit/src/rabbit_classic_queue.erl
```diff
@@ -63,14 +64,66 @@
...
+            ok;
+        _ ->
+            case rabbit_queue_location:master_locator_permitted() of
+                true ->
+                    case lists:member(Locator, rabbit_queue_location:queue_leader_locators()) of
+                        true -> ok;
+                        false -> {error, "~tp is not a valid master locator", [Locator]}
+                    end;
+                false ->
+                    {error, "use of deprecated queue-master-locator argument is not permitted", []}
```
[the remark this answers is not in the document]
dumbbell: I don’t have issues with the error message itself. I wanted to make sure that the behavior isn’t a temporary one that will change again once the feature is deleted.

ansd (changes requested):
When I want to create a classic queue in the Management UI, I only see the option `Master locator` although I set
```
deprecated_features.permit.queue_master_locator = false
```
in my `rabbitmq.conf`.

I would expect to see a `Leader locator` button in the Management UI.

Screenshot 2024-07-11 at 12 47 32

ansd:
same for policies in the Management UI:
Screenshot 2024-07-11 at 12 52 44

ansd:
I'm unable to test upgrades from `v3.13.x` to this PR branch using `make run-broker` because the node fails during boot with this PR branch:
```
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0>
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0> BOOT FAILED
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0> ===========
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0> Exception during startup:
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0>
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0> throw:{timeout,{rabbitmq_metadata,rabbit@V333JFK777}}
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0>
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0>     rabbit_khepri:-register_projections/0-lc$^9/1-0-/1, line 1087
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0>     rabbit_khepri:register_projections/0, line 1088
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0>     rabbit_khepri:setup/1, line 255
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0>     rabbit:run_prelaunch_second_phase/0, line 380
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0>     rabbit:start/2, line 902
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0>     application_master:start_it_old/4, line 293
2024-07-11 13:10:08.727825+02:00 [error] <0.262.0>
```

ansd:
If you have tested that upgrades from v3.13 to this PR branch work with classic queues with `x-queue-master-locator` queue argument or `queue-master-locator` policy set and a subsequent upgrade where the new deprecated feature is denied, and the classic queue declaration by the client keeps working without having to re-create the classic queue, I'm happy with this PR being merged.

PR #11572, opened 2024-06-27
Speed up rabbit tests backport

kjnilsson:


michaelklishin:
The `rabbitmq_management` failures are present in `v3.13.x` and not specific to this PR, I'm working on them.

PR #11597, opened 2024-07-01
CI: Add '-mixed' suffix to mixed-version job names

the-mikedavis:
Currently it's hard to tell apart the regular and mixed-version jobs in the GitHub UI - you need to check the "Set up job" or `bazelisk cquery` steps of a job. We can add a "-mixed" suffix to make it easy to tell them apart at a glance.

This also includes a fix for running the CLI tests in the mixed-version workflow: previously the CLI was running the regular plugin workflow but this has been fixed to use `test-plugin-mixed.yaml` instead.

Sample of a run where it's hard to tell that the job is a mixed-version test: https://github.com/rabbitmq/rabbitmq-server/actions/runs/9748547643/job/26903792526

### DIFF
--- a/.github/workflows/templates/test-mixed-versions.template.yaml
+++ b/.github/workflows/templates/test-mixed-versions.template.yaml
@@ -4,7 +4,7 @@
 #@ def job_names(plugins):
 #@   names = []
 #@   for p in plugins:
-#@     names.append("test-"+p)
+#@     names.append("test-"+p+"-mixed")
 #@   end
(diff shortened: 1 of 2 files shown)


the-mikedavis:
We should backport this to 3.13.x manually: we can cherry-pick the first two commits as-is and then use `make actions-workflows` to regenerate the workflow file for the suites available in that branch.

michaelklishin:
@the-mikedavis good catch!

PR #11598, opened 2024-07-01
3.13: Add '-mixed' suffix to mixed-version CI job names (Backport #11597)

the-mikedavis:
This is a backport of #11597 to the 3.13.x branch.


the-mikedavis:
@michaelklishin mentioned that the `rabbitmq_management` failures are legitimate but unrelated to this PR and should be handled in a later change

PR #11659, opened 2024-07-10
For 4.0: Require all stable feature flags added up to 3.13.0

ansd:
Since feature flag `message_containers` introduced in 3.13.0 is required in 4.0, we can also require all other feature flags introduced in or before 3.13.0 and remove their compatibility code for 4.0:

* `restart_streams`
* `stream_sac_coordinator_unblock_group`
* `stream_filtering`
* `stream_update_config_command`

Related to https://github.com/rabbitmq/rabbitmq-server/pull/11642

### DIFF
--- a/deps/rabbit/src/rabbit_stream_queue.erl
+++ b/deps/rabbit/src/rabbit_stream_queue.erl
@@ -96,8 +96,7 @@
                         soft_limit :: non_neg_integer(),
                         slow = false :: boolean(),
                         readers = #{} :: #{rabbit_types:ctag() => #stream{}},
-                        writer_id :: binary(),
-                        filtering_supported :: boolean()
+                        writer_id :: binary()
                        }).
 
 -import(rabbit_queue_type_util, [args_policy_lookup/3]).
@@ -286,8 +285,7 @@ consume(Q, #{no_ack := true,
 consume(Q, #{limiter_active := true}, _State)
   when ?amqqueue_is_stream(Q) ->
     {error, global_qos_not_supported_for_queue_type};
-consume(Q, Spec,
-        #stream_client{filtering_supported = FilteringSupported} = QState0)
+consume(Q, Spec, #stream_client{} = QState0)
   when ?amqqueue_is_stream(Q) ->
     %% Messages should include the offset as a custom header.
     case get_local_pid(QState0) of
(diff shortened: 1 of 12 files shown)


deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -96,8 +96,7 @@
                         soft_limit :: non_neg_integer(),
                         slow = false :: boolean(),
                         readers = #{} :: #{rabbit_types:ctag() => #stream{}},
-                        writer_id :: binary(),
-                        filtering_supported :: boolean()
```
dumbbell, line 100: Are we 100% sure that this record is never stored in Mnesia/Khepri and never sent to another node?

ansd: yes

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
```diff
@@ -1256,7 +1239,7 @@ recover_after_leader_and_coordinator_kill(Config) ->
 
 
     ct:pal("sys state ~p", [CState]),
-
+    rabbit_ct_broker_helpers:rpc(Config, 0, ?MODULE, delete_testcase_queue, [Q]),
```
dumbbell, line 1242: Is this a drive-by change? If this is a fix, should it be backported separately of this whole patch?

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
```diff
@@ -2218,6 +2194,7 @@ leader_locator_balanced_maintenance(Config) ->
       end, 60000),
 
     true = rabbit_ct_broker_helpers:unmark_as_being_drained(Config, Server3),
+    rabbit_ct_broker_helpers:rpc(Config, 0, ?MODULE, delete_queues, [[Q1, Q]]),
```
dumbbell, line 2197: Same question, drive-by change?

deps/rabbit/src/rabbit_stream_sac_coordinator.erl
```diff
@@ -629,32 +629,12 @@ handle_consumer_removal(Group0, Consumer, Stream, ConsumerName) ->
...
-        false ->
-            tuple
-    end.
-
 notify_consumer_effect(Pid, SubId, Stream, Name, Active) ->
     notify_consumer_effect(Pid, SubId, Stream, Name, Active, false).
 
 notify_consumer_effect(Pid, SubId, Stream, Name, Active, SteppingDown) ->
-    notify_consumer_effect(Pid, SubId, Stream, Name, Active, SteppingDown, message_type()).
+    notify_consumer_effect(Pid, SubId, Stream, Name, Active, SteppingDown, map).
```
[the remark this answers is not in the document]
ansd: Good spot, thanks!

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
ansd: As I already wrote in Slack:
> Test cases rebalance and restart_stream were always skipped prior to this PR because the feature flag restart_streams in these test cases is misspelled. So, that part should be backported to v3.13.x

This is not a "drive-by change" because this PR removes the code to enable the feature flag in https://github.com/rabbitmq/rabbitmq-server/pull/11659/files#diff-bd1bb71aee8de223a400a433496cefbd338052929a9ed71906665ca953b72746L2820-L2827 which causes test case `rebalance` to run for the first time. The test case was buggy. One bug is the line you are commenting: Test cases didn't clean up properly after each other which caused assertions in test case `rebalance` to fail. (In general, I prefer to randomise test case executions within a group such that test interference can be easily spotted.) So, these test case fixes are already in a separate commit in this PR and can be easily backported to `v3.13.x`.

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
ansd: same answer as above

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
ansd: Here is the commit I'm referring to: https://github.com/rabbitmq/rabbitmq-server/pull/11659/commits/6cdfb62a81c287f017a0b8907ee98ab759430583

dumbbell (approved):

deps/rabbit/test/rabbit_stream_queue_SUITE.erl
dumbbell: Thank you, I understand now.

acogoluegnes (approved):

PR #11661, opened 2024-07-10
rabbitmq_peer_discovery_{etcd,consul}: Fix error handling if Khepri is unsupported

dumbbell:
## How

We must check the return value of `rabbit_ct_broker_helpers:run_steps/2` because it could ask that the testsuite/testgroup/testcase should be skipped.


dumbbell:
This change is backported to `v3.13.x` as part of #11647.

PR #11247, opened 2024-05-16
Bump Khepri from 0.13.0 to 0.14.0

dumbbell:
Release notes:
https://github.com/rabbitmq/khepri/releases/tag/v0.14.0

While here, bump khepri_mnesia_migration from 0.4.0 to 0.5.0 as well.

PR #11663, opened 2024-07-10
Turn off mixed version tests against Khepri

the-mikedavis:
Khepri v0.14.0 creates problems for mixed-version tests: the new machine version causes commands and queries to fail on the nodes running the old version. We should separately fix these issues and eventually re-enable mixed version tests with Khepri as the feature stabilizes.

### DIFF
--- a/.github/workflows/test-plugin-mixed.yaml
+++ b/.github/workflows/test-plugin-mixed.yaml
@@ -29,7 +29,10 @@ jobs:
         - 26
         metadata_store:
           - mnesia
-          - khepri
+          # Khepri is currently skipped because Khepri is an unstable feature: we don't guarantee upgrability.
+          # Mixed-version tests currently fail with Khepri because of a new machine version introduced in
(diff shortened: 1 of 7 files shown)


the-mikedavis:
Part of this PR is updating suites to respect the `RABBITMQ_METADATA_STORE` env (via `rabbit_ct_broker_helpers:configured_metadata_store/1`) and conditionally skip cases, groups or suites. That's necessary to avoid some failures possible when mixed version tests would fail because of Khepri even though `RABBITMQ_METADATA_STORE=mnesia`. This should also save us a small amount of CI time that was spent on tests run for either value of `RABBITMQ_METADATA_STORE`.

Before the last force-push this branch included some of the changes to bump Khepri to 0.14.0 from #11247 which has mixed-test failures in mnesia jobs. There are more places we could use `rabbit_ct_broker_helpers:configured_metadata_store/1` but the changes so far were enough to have mixed versions tests pass for mnesia.

PR #11665, opened 2024-07-10
Fix direct reply to crash when tracing is enabled

ansd:
This commit fixes https://github.com/rabbitmq/rabbitmq-server/discussions/11662

Prior to this commit the following crash occurred when an RPC reply message entered RabbitMQ and tracing was enabled:
```
** Reason for termination ==
** {function_clause,
       [{rabbit_trace,'-tap_in/6-fun-0-',
            [{virtual_reply_queue,
                 <<"amq.rabbitmq.reply-to.g1h2AA5yZXBseUAyNzc5NjQyMAAAC1oAAAAAZo4bIw==.+Uvn1EmAp0ZA+oQx2yoQFA==">>}],
            [{file,"rabbit_trace.erl"},{line,62}]},
        {lists,map,2,[{file,"lists.erl"},{line,1559}]},
        {rabbit_trace,tap_in,6,[{file,"rabbit_trace.erl"},{line,62}]},
        {rabbit_channel,handle_method,3,
            [{file,"rabbit_channel.erl"},{line,1284}]},
        {rabbit_channel,handle_cast,2,
            [{file,"rabbit_channel.erl"},{line,659}]},
        {gen_server2,handle_msg,2,[{file,"gen_server2.erl"},{line,1056}]},
        {proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,241}]}]}
```

(Note that no trace message is emitted for messages that are delivered to direct reply to requesting clients (neither in 3.12, nor in 3.13, nor after this commit). This behaviour can be added in future when a direct reply virtual queue becomes its own queue type.)

### DIFF
--- a/deps/rabbit/src/rabbit_trace.erl
+++ b/deps/rabbit/src/rabbit_trace.erl
@@ -61,6 +61,8 @@ tap_in(Msg, QNames, ConnName, ChannelNum, Username, TraceX) ->
     RoutedQs = lists:map(fun(#resource{kind = queue, name = Name}) ->
                                  {longstr, Name};
                             ({#resource{kind = queue, name = Name}, _}) ->
+                                 {longstr, Name};
+                            ({virtual_reply_queue, Name}) ->
                                  {longstr, Name}
                          end, QNames),
     trace(TraceX, Msg, <<"publish">>, XName,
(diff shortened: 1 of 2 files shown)


michaelklishin (approved):

lukebakken (approved):
Great test!

michaelklishin:
I'm looking into the mixed version cluster failure. We don't really need to run this test in mixed version mode since the problem in #11662 was reproducible in an "all 3.13.x" cluster.

PR #11668, opened 2024-07-10
Bump `khepri` to 0.14.0, `khepri_mnesia_migration` to 0.5.0

the-mikedavis:
This version of Khepri includes a new Ra machine version which fails mixed version tests when `RABBITMQ_METADATA_STORE=khepri`. This depends on #11663 which turns off those tests.


dumbbell:
Oh sorry I wasn’t clear, the upgrade of Khepri isn’t ready. It depends on a version of Ra that isn’t released yet and there is work in progress in Khepri as well that we need for RabbitMQ.

Could you please revert this pull request?

PR #11685, opened 2024-07-11
Handle timeouts possible in Khepri minority in `rabbit_db_binding`

the-mikedavis:
These changes are split off of #10915 - I'd like to approach these changes incrementally and work through each module so that the diff is readable. This PR covers `rabbit_db_binding`: `create/2` and `delete/2` may return `{error, timeout}` when Khepri is in a minority and this PR updates the direct and indirect callers so that the error is handled gracefully.

### DIFF
--- a/deps/rabbit/src/rabbit_definitions.erl
+++ b/deps/rabbit/src/rabbit_definitions.erl
@@ -883,12 +883,17 @@ add_binding(VHost, Binding, ActingUser) ->
                     rv(VHost, DestType, destination, Binding), ActingUser).
 
 add_binding_int(Binding, Source, Destination, ActingUser) ->
-    rabbit_binding:add(
-      #binding{source       = Source,
-               destination  = Destination,
-               key          = maps:get(routing_key, Binding, undefined),
-               args         = args(maps:get(arguments, Binding, undefined))},
-      ActingUser).
+    case rabbit_binding:add(
+           #binding{source       = Source,
+                    destination  = Destination,
+                    key          = maps:get(routing_key, Binding, undefined),
+                    args         = args(maps:get(arguments, Binding, undefined))},
+           ActingUser) of
+        ok ->
+            ok;
+        {error, _} = Err ->
+            throw(Err)
+    end.
 
 dest_type(Binding) ->
     rabbit_data_coercion:to_atom(maps:get(destination_type, Binding, undefined)).
--- a/deps/rabbit/src/rabbit_binding.erl
+++ b/deps/rabbit/src/rabbit_binding.erl
@@ -41,7 +41,8 @@
 -type bind_ok_or_error() :: 'ok' | bind_errors() |
                             rabbit_types:error({'binding_invalid', string(), [any()]}) |
                             %% inner_fun() result
-                            rabbit_types:error(rabbit_types:amqp_error()).
+                            rabbit_types:error(rabbit_types:amqp_error()) |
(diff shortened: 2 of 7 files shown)


deps/rabbit/src/rabbit_definitions.erl
```diff
@@ -883,12 +883,17 @@ add_binding(VHost, Binding, ActingUser) ->
...
+    case rabbit_binding:add(
+           #binding{source       = Source,
+                    destination  = Destination,
+                    key          = maps:get(routing_key, Binding, undefined),
+                    args         = args(maps:get(arguments, Binding, undefined))},
+           ActingUser) of
+        ok ->
+            ok;
+        {error, _} = Err ->
+            throw(Err)
```
michaelklishin, line 895: I don't object to this change but throwing on the definition import path will halt the entire process. Perhaps we already do that in other places :( It's a good question whether we should continue or halt, "some definitions" can be better than none.

But for cases where the definitions are imported from an external tool on boot (and that happens to early), failing makes more sense.

michaelklishin (approved):

deps/rabbit/src/rabbit_definitions.erl
the-mikedavis: I was originally thinking we should refactor `rabbit_definitions` to import all definitions in one big transaction. Then the side effects like starting vhost or queue processes would happen after the transaction succeeded. It would be a really really big refactor though and definitions import is idempotent so if an import fails partially all you should need to do is re-run it

the-mikedavis: We do end up catching these throws here https://github.com/rabbitmq/rabbitmq-server/blob/ff5a3245645c5f0d66484a4a0b30d3a85050912e/deps/rabbit/src/rabbit_definitions.erl#L663-L671

and convert it into an `{error, _}` but it will halt the import after the first `concurrent_for_all/4` throws. It won't be a crash+stacktrace in the logs at least 😅

dcorbacho: If the metadata store is timing out, most likely it won't be the only failing operation. I wonder if this is one of the cases where we should retry the operation. Like if it fails during boot, retry a few times. Or just abort the boot process?

the-mikedavis: I'll add a bullet for looking into startup specifically to my todo list. It looks like currently we crash when trying to register projections if you start in a minority during `rabbit_khepri:setup/0` which comes before definitions import. (Start a 3-node cluster, enable khepri_db, stop the cluster, start one node.) I think we should have retries there so you have some time to form a majority before we error out and fail the boot. There are also hard matches on `ok` for when we insert default data https://github.com/rabbitmq/rabbitmq-server/blob/2679649ac5b53599411a3319c78d47568e8f1b55/deps/rabbit/src/rabbit.erl#L1139-L1168

so I think keeping definitions imports and default data behaving the same is probably ok for now. These operations happen right after we start up Khepri and send commands like `register_projections` so if we get to this part of the boot we have already been in a majority and probably still are, unless some of the nodes disappear right after boot.

dcorbacho: I thought we added retries at some point to wait for other nodes, as we do for Mnesia. I'll have a look

the-mikedavis:
Force-push was a rebase

PR #11559, opened 2024-06-25
Prometheus: a handful of new per-exchange and per-queue metrics

LoisSotoLopez:
## Proposed Changes

Solves the issue explained at https://github.com/rabbitmq/rabbitmq-server/issues/11217

- Defines 3 more `rabbit_core_metrics` tables, where per queue, per exchange and per queue/exchange metrics will be stored without including a channel ID on the key.
- Initializes those tables and configures `rabbit_core_metrics` to populate them on `channel_stats` calls, apart from populating those metrics per channel.
- Configures the `rabbit_mgmt` and `prometheus_rabbitmq_core` `_metrics_collector`s to collect those metrics.
- Provides a few tests in the `rabbit_prometheus_http_SUITE`

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist
- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

PR #11688, opened 2024-07-12
Revert "Prometheus: a handful of new per-exchange and per-queue metrics"

michaelklishin:
Reverts rabbitmq/rabbitmq-server#11559. Even with #11687 the Prometheus test suite still flakes most of the time with both Bazel and Make.


michaelklishin:
Will revert manually. Sigh.

PR #11696, opened 2024-07-12
Simplify

ansd:
This is follow-up of https://github.com/rabbitmq/rabbitmq-server/pull/11676
Protect only `rabbit_mqtt_processor:handle_queue_event/2` since only that call might throw a `{send_failed, Reason}`.


lukebakken (approved):
Nice use of `try` this way 🌟

PR #11700, opened 2024-07-12
Use 'try'/'catch' rather than 'ets:whereis/1' for Khepri projections

the-mikedavis:
See https://github.com/rabbitmq/rabbitmq-server/pull/11667#issuecomment-2224967310

`ets:whereis/1` adds some overhead - it's two ETS calls rather than one when `ets:whereis/1` returns a table identifier. It's also not atomic: the table could disappear between `ets:whereis/1` calls and the call to read data from a projection. We replace all `ets:whereis/1` calls on projection tables with `try`/`catch` and return default values when we catch the `badarg` `error` which ETS emits when passed a non-existing table name.

One special case though is `ets:info/2` which returns `undefined` when passed a non-existing table names. That block is refactored to use a `case` instead.

### DIFF
--- a/deps/rabbit/src/rabbit_db_binding.erl
+++ b/deps/rabbit/src/rabbit_db_binding.erl
@@ -413,11 +413,11 @@ get_all_in_mnesia() ->
       end).
 
 get_all_in_khepri() ->
-    case ets:whereis(?KHEPRI_BINDINGS_PROJECTION) of
-        undefined ->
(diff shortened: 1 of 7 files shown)


dcorbacho:
Good point about the overhead @dumbbell  I'm not so worried about tables disappearing but who knows!

michaelklishin (approved):

PR #11715, opened 2024-07-15
mc: increase utf8 scanning limit for longstr conversions.

kjnilsson:
The AMQP 0.9.1 longstr type is problematic as it can contain arbitrary binary data but is typically used for utf8 by users.

The current conversion into AMQP avoids scanning arbitrarily large longstr to see if they only contain valid utf8 by treating all longstr data longer than 255 bytes as binary. This is in hindsight too strict and thus this commit increases the scanning limit to 4096 bytes - enough to cover the vast majority of AMQP 0.9.1 header values.

This change also conversts the AMQP binary types into longstr to ensure that existing data (held in streams for example) is converted to an AMQP 0.9.1 type most likely what the user intended.

### DIFF
--- a/deps/rabbit/src/mc_amqpl.erl
+++ b/deps/rabbit/src/mc_amqpl.erl
@@ -42,6 +42,7 @@
 -define(AMQP10_FOOTER, <<"x-amqp-1.0-footer">>).
 -define(PROTOMOD, rabbit_framing_amqp_0_9_1).
 -define(CLASS_ID, 60).
+-define(LONGSTR_UTF8_LIMIT, 4096).
 
 -opaque state() :: #content{}.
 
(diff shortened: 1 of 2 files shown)


ansd:
> This change also conversts the AMQP binary types into longstr to ensure that existing data (held in streams for example) is converted to an AMQP 0.9.1 type most likely what the user intended.

... that's perfectly fine since according to the AMQP 0.9.1 spec:
> Long strings, used to hold chunks of binary data

ansd (approved):

michaelklishin:
Looks like if both labels are set, our greatly simplified Mergify backporting setup does nothing 😅 

@mergifyio backport v4.0.x v3.13.x

acogoluegnes:
@Mergifyio backport v4.0.x v3.13.x

PR #11719, opened 2024-07-15
rabbit_db_exchange: Remove dead `set/1`

the-mikedavis:
`rabbit_db_exchange:set/2` was only used by `rabbit_policy:recover/0` which was removed by 3.13.0 in c84bc83ce18b2741bd5c75fd4f70793a67efc445 (#7727). That function was part of the old upgrade system (predating feature flags) so this code path has been dead code for quite a while.

### DIFF
--- a/deps/rabbit/src/rabbit_db_exchange.erl
+++ b/deps/rabbit/src/rabbit_db_exchange.erl
@@ -20,7 +20,6 @@
          count/0,
          update/2,
          create_or_get/1,
-         set/1,
          peek_serial/1,
          next_serial/1,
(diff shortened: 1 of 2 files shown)


the-mikedavis:
Removing this is not quite as straightforward as I thought. `rabbit_db_exchange_SUITE:recover/1` uses `set/1` for setup and that setup can't be done with `create_or_get/1` instead since it writes to both tables and doesn't clear the `decorators` for the durable table.

It should be easy to land this once we've removed mnesia and there's no rush to remove it now

PR #11741, opened 2024-07-17
rabbit_khepri: Retry register_projections during boot

dcorbacho:
Gives some time to form a majority during the boot process, allowing nodes to boot more easily.

See discussion https://github.com/rabbitmq/rabbitmq-server/pull/11685#discussion_r1679756356

### DIFF
--- a/deps/rabbit/src/rabbit_khepri.erl
+++ b/deps/rabbit/src/rabbit_khepri.erl
@@ -306,7 +306,7 @@ wait_for_register_projections(Timeout, Retries) ->
     try
         register_projections()
     catch
-        throw : {timeout, _ServerId} ->
+        throw : timeout ->
             wait_for_register_projections(Timeout, Retries -1)
     end.
 
--- a/deps/rabbit/src/rabbit_khepri.erl
+++ b/deps/rabbit/src/rabbit_khepri.erl
@@ -252,7 +252,7 @@ setup(_) ->
     case khepri:start(?RA_SYSTEM, RaServerConfig) of
         {ok, ?STORE_ID} ->
             wait_for_leader(),
-            register_projections(),
+            wait_for_register_projections(),
             ?LOG_DEBUG(
                "Khepri-based " ?RA_FRIENDLY_NAME " ready",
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -295,6 +295,21 @@ wait_for_leader(Timeout, Retries) ->
...
+
+wait_for_register_projections(_Timeout, 0) ->
+    exit(timeout_waiting_for_khepri_projections);
+wait_for_register_projections(Timeout, Retries) ->
+    rabbit_log:info("Waiting for Khepri projections for ~tp ms, ~tp retries left",
+                    [Timeout, Retries - 1]),
+    try
+        register_projections()
+    catch
+        throw : {timeout, _ServerId} ->
```
the-mikedavis, line 309:
```suggestion
        throw : timeout ->
```

in Khepri 0.14.0+ this is simplified into just `{error, timeout}` https://github.com/rabbitmq/khepri/pull/256

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -295,6 +295,21 @@ wait_for_leader(Timeout, Retries) ->
...
+wait_for_register_projections() ->
+    wait_for_register_projections(retry_timeout(), retry_limit()).
+
+wait_for_register_projections(_Timeout, 0) ->
+    exit(timeout_waiting_for_khepri_projections);
+wait_for_register_projections(Timeout, Retries) ->
+    rabbit_log:info("Waiting for Khepri projections for ~tp ms, ~tp retries left",
+                    [Timeout, Retries - 1]),
+    try
+        register_projections()
```
the-mikedavis, line 307: This would retry registering all projections again if it fails. I think that's probably ok since I expect the first projection would be the only one to time out but we could make this a little more clever and fold over the list of projections to register and the retry number. Feel free to ignore though, it seems very unlikely to be hit in practice

the-mikedavis (approved):

PR #11745, opened 2024-07-18
Propagate errors in rabbit_registry register/unregister

the-mikedavis:
...TBW...


the-mikedavis:
Hmm on second look when we crash in `rabbit_registry` the whole node goes down since it's a child process of `rabbit_sup`. These callbacks are only called during boot and either way the error is fatal. `rabbit_registry` itself is quite old so I'd rather leave this be for the time being. If we see error reports about crashes involving `rabbit_registry` we can revisit this but for now let's focus on other possible errors.

PR #11755, opened 2024-07-19
Prometheus: some per-exchange/per-queue metrics aggregated per-channel (rebased)

michaelklishin:
This is #11726 rebased. Submitting to see if CI passes with Bazel and GCP-based Bazel caching.


ansd:
FTR commit https://github.com/rabbitmq/rabbitmq-server/commit/bb93e718c2b424bdb200aa67c96b37fa2bd19600 reduces end-to-end throughput by around 8%.

In my benchmark, I use
* v4.0.0-beta.5
* OTP 27.0.1
* Ubuntu box
* `make run-broker RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+S 3"`
* `java -jar target/perf-test.jar --predeclared --exchange amq.default --routing-key my-classic-queue --queue my-classic-queue -z 20 --size 12 --confirm 1000 --multi-ack-every 1000 --qos 5000`

Result for v4.0.0-beta.5, i.e. **including** above commit:
```
id: test-083126-681, sending rate avg: 113601 msg/s
id: test-083126-681, receiving rate avg: 113601 msg/s
```
Flame graph shows that 8.2% of CPU time is spent in `rabbit_core_metrics:channel_stats`.

Result with above commit reverted:
```
id: test-083019-691, sending rate avg: 123794 msg/s
id: test-083019-691, receiving rate avg: 123794 msg/s
```

If I switch off writing to `rabbit_core_metrics` completely, I get:
```
id: test-084728-218, sending rate avg: 139888 msg/s
id: test-084728-218, receiving rate avg: 139888 msg/s
```

I do not suggest to revert this commit, but I'd like to point out as an interesting that `ets:update_counter/4` is expensive when called multiple times on a per-message basis.

gomoripeti:
thanks David for uncovering this, 8% is relevant enough to look into it. (I wonder if this means that channel_stats already costed 8% and now they cost 8+8%? I also speculate that the newly added `ets:update_counter` would perform even worth in parallel when multiple channels publish to the same exchange or queue, as multiple processes try to update the same ETS entry. This was not the case when each channel had its own entry)

I'd open a discussion for the conversation about the details and ideas. 
I thought that `ets:update_counter` is "good enough" but apparently RabbitMQ reached a throughput where it matters.
I have in mind an implementation of core_metrics with atomic counters/seshat instead of ets:update_counter that wouldn't require too large code change.

gomoripeti:
thanks for the heads up, I'll hold off further efforts then. updating core metrics tables every few seconds only sounds good and probably won't need further optimisations.

I could not reproduce the difference, probably I did something wrong with the perf test
```
** perftest results
RMQ 4.0.0~beta.6
OTP 26.2.5.3
*** channel_stats does 2 ets:update_counter
id: test-152711-512, sending rate avg: 47211 msg/s
id: test-152711-512, receiving rate avg: 47211 msg/s

*** channel_stats does 1 ets:update_counter
id: test-152935-225, sending rate avg: 47844 msg/s
id: test-152935-225, receiving rate avg: 47844 msg/s

*** channel_stats does nothing
id: test-153508-370, sending rate avg: 46969 msg/s
id: test-153508-370, receiving rate avg: 46969 msg/s
```

PR #11763, opened 2024-07-19
Add stream replication port range in ini-style configuration

acogoluegnes:
This is more straightforward than configuring Osiris in the advanced configuration file.

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -2665,6 +2665,31 @@ fun(Conf) ->
     list_to_binary(cuttlefish:conf_get("amqp1_0.default_vhost", Conf))
 end}.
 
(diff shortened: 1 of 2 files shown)


deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -2665,6 +2665,31 @@ fun(Conf) ->
     list_to_binary(cuttlefish:conf_get("amqp1_0.default_vhost", Conf))
 end}.
 
+{mapping, "rabbit.stream_replication_port_range.min", "osiris.port_range", [
```
michaelklishin, line 2668: Do we really need this `rabbit.` prefix? How about `stream.replication.port_range.min` and `stream.replication.port_range.max`? `stream.*` is a suffix we already use in `rabbitmq_stream`.

michaelklishin (changes requested):

michaelklishin (approved):

PR #11767, opened 2024-07-19
maintenance_mode_SUITE: Skip leadership transfer case on mnesia (v4.0.x)

the-mikedavis:
This case only targets Khepri. Instead of setting the `metadata_store` config option we should skip the test when the configured metadata store is mnesia.


the-mikedavis:
The backport of this commit for v3.13.x is in https://github.com/rabbitmq/rabbitmq-server/pull/11765

PR #11785, opened 2024-07-22
Handle timeouts possible in Khepri minority in `rabbit_db_exchange`

the-mikedavis:
Continuation of the changes like in https://github.com/rabbitmq/rabbitmq-server/issues/11685 and https://github.com/rabbitmq/rabbitmq-server/pull/11706 but for exchanges.

The changes for exchanges are a bit more involved than prior PRs. Here's a summary of the changes:

* `rabbit_exchange:declare/7`'s spec is updated to return `{ok, #exchange{}} | {error, timeout}` where it returned just `#exchange{}` before
    * this is safe to change since the function is only ever called through RPC in the CLI's test helpers
* Callers of `rabbit_exchange:delete/3` need some small changes to handle an `{error, timeout}` return
    * I've added a `rabbit_exchange:ensure_deleted/3` wrapper that covers the common case for callers where `ok | {error, not_found}` are handled the same way.
* Set `timeout => infinity` for calls for updating the exchange serial
    * Serial updates are always made after another database operation (for example adding or deleting an exchange)
    * Ideally we would update the serial in a transaction but this would take a larger change, so I've left it for future work.

### DIFF
--- a/deps/rabbit/src/rabbit_amqp_management.erl
+++ b/deps/rabbit/src/rabbit_amqp_management.erl
@@ -285,8 +285,15 @@ handle_http_req(<<"DELETE">>,
     ok = prohibit_default_exchange(XName),
     ok = prohibit_reserved_amq(XName),
     PermCache = check_resource_access(XName, configure, User, PermCache0),
-    _ = rabbit_exchange:delete(XName, false, Username),
-    {<<"204">>, null, {PermCache, TopicPermCache}};
+    case rabbit_exchange:ensure_deleted(XName, false, Username) of
+        ok ->
+            {<<"204">>, null, {PermCache, TopicPermCache}};
+        {error, timeout} ->
+            throw(
+              <<"500">>,
+              "failed to delete exchange '~ts' due to a timeout",
+              [XNameBin])
+    end;
 
 handle_http_req(<<"POST">>,
                 [<<"bindings">>],
--- a/deps/rabbit/src/rabbit_amqp_management.erl
+++ b/deps/rabbit/src/rabbit_amqp_management.erl
@@ -218,9 +218,9 @@ handle_http_req(<<"PUT">>,
                     {error, timeout} ->
                         throw(
                           <<"503">>,
-                          "Could not create exchange '~ts' in vhost '~ts' "
-                          "because the operation timed out",
-                          [XName, Vhost])
+                          "Could not create ~ts because the operation "
+                          "timed out",
+                          [rabbit_misc:rs(XName)])
                 end
         end,
     try rabbit_exchange:assert_equivalence(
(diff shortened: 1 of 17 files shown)


michaelklishin (approved):

deps/rabbit/src/rabbit_amqp_management.erl
```diff
@@ -210,9 +210,18 @@ handle_http_req(<<"PUT">>,
...
-                  Internal, XArgs, Username)
+                case rabbit_exchange:declare(
+                       XName, XTypeAtom, Durable, AutoDelete,
+                       Internal, XArgs, Username) of
+                    {ok, DeclaredX} ->
+                        DeclaredX;
+                    {error, timeout} ->
+                        throw(
+                          <<"503">>,
+                          "Could not create exchange '~ts' in vhost '~ts' "
```
ansd, line 221: Will this crash because `~s` expects a string an `XName`  is a tuple?

ansd: Should we just use `rabbit_misc:rs/1` here?

the-mikedavis: Ah that's a resource record. Right let's use `rabbit_misc:rs/1` for that too

deps/rabbit/src/rabbit_amqp_management.erl
```diff
@@ -285,8 +294,15 @@ handle_http_req(<<"DELETE">>,
...
-    _ = rabbit_exchange:delete(XName, false, Username),
-    {<<"204">>, null, {PermCache, TopicPermCache}};
+    case rabbit_exchange:ensure_deleted(XName, false, Username) of
+        ok ->
+            {<<"204">>, null, {PermCache, TopicPermCache}};
+        {error, timeout} ->
+            throw(
+              <<"503">>,
+              "failed to delete exchange '~ts' due to a timeout",
+              [XNameBin])
```
ansd, line 304: Should we use `rabbit_misc:rs/1` here to log the vhost as well and pass `XName` as argument?

deps/rabbit/src/rabbit_channel.erl
```diff
@@ -2566,13 +2569,22 @@ handle_method(#'exchange.declare'{exchange    = XNameBin,
...
+                                             Args,
+                                             Username) of
+                    {ok, DeclaredX} ->
+                        DeclaredX;
+                    {error, timeout} ->
+                        rabbit_misc:protocol_error(
+                          internal_error,
+                          "failed to declare exchange '~ts' in vhost '~ts' "
+                          "because the operation timed out",
+                          [XNameBinStripped, VHostPath])
```
the-mikedavis, line 2586: We should be able to use `rabbit_misc:rs/1` on `ExchangeName` here too

Issue #11743, opened 2024-07-18
Provide detailed memory metrics via prometheus plugin


der-eismann:
### Is your feature request related to a problem? Please describe.

Hey everyone, we are currently working on replacing the soon-to-be EOL https://github.com/kbudde/rabbitmq_exporter with the built-in prometheus plugin. With that exporter it was possible to get detailed memory statistics from the management plugin, which have helped us debug issues:
https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_management/priv/www/js/tmpl/memory.ejs#L9-L31

Unfortunately I was unable to get these metrics from the prometheus plugin, the only thing that came close was `process_resident_memory_bytes` (https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl#L72)

### Describe the solution you'd like

Provide all memory metrics from the management UI via prometheus plugin

michaelklishin:
This is open source software, you are welcome to contribute what you find missing.

The data comes from [`rabbit_vm:memory/0`](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_vm.erl#L18) and the metrics belong [to this group](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl#L79-L87).

der-eismann:
Wow, I wasn't even able to finish my Erlang introductory course in that short time. Thanks for adding these metrics so quickly!

michaelklishin:
These metrics are fairly expensive with many queues and streams, so we will limit this to 4.0 and look for ways to optimize this or make this opt-in.

mkuratczyk:
@der-eismann We have now merged this into main/4.0 (but not 3.13). There's a dedicated endpoint for these metrics: https://www.rabbitmq.com/docs/next/prometheus#memory-breakdown-endpoint

However, I struggle to find a nice Grafana vizualization for these metrics. There are quite a few of them and multiple by the number of nodes in the cluster, you get a lot of data points. Are you currently visualizing these metrics from the exporter?
Can you share wha that looks like? Ideally, if you could contribute a panel for them, that'd be great.

[...]

PR #11788, opened 2024-07-23
Move memory breakdown metrics to new endpoint

mkuratczyk:
Move metrics introduced in https://github.com/rabbitmq/rabbitmq-server/pull/11746 to a dedicated endpoint (`/metrics/memory-breakdown`).

Collecting them on a large system (tens of thousands of processes or more) can be time consuming as we iterate over all processes. By putting them on a separate endpoint, we make that opt-in

Related docs:
https://github.com/rabbitmq/rabbitmq-website/pull/1985

I'm still playing around with the best visualization for these metrics, but what I have so far proves that things work:
Screenshot 2024-07-23 at 11 53 41


michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/issues/11743.

PR #11809, opened 2024-07-24
QQ: use a dedicated function for queue recovery after Ra system restart.

kjnilsson:
Previously we used the `registered` approach where all Ra servers that have a registered name would be recovered. This could have unintended side effects for queues that e.g. were deleted when not all members of a quorum queueu were running when the queue was deleted. In this case the Ra system would have recovered the members that were not deleted which is not ideal as a dangling member would just sit and loop in pre vote state and a future declaration of the queue may partially fail.

Instead we rely on the meta data store for the truth about which members should be restarted after a ra system restart.

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -17,6 +17,7 @@
          handle_event/3]).
 -export([is_recoverable/1,
          recover/2,
+         system_recover/1,
          stop/1,
          start_server/1,
          restart_server/1,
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -641,6 +642,24 @@ is_recoverable(Q) when ?is_amqqueue(Q) and ?amqqueue_is_quorum(Q) ->
...
+                            [?MODULE, ?FUNCTION_NAME, length(Queues)]),
+            {Recovered, Failed} = recover(<<>>, Queues),
+            rabbit_log:info("~s:~s/1 recovered ~b queues, "
+                            "failed to recover ~b queues",
+                            [?MODULE, ?FUNCTION_NAME, length(Recovered),
+                             length(Failed)]),
+            ok;
+        false ->
+            rabbit_log:info("~s:~s/1 rabbit not booted, skipping queue recovery",
+                            [?MODULE, ?FUNCTION_NAME]),
```
ansd, line 659: You could use macro `?FUNCTION_ARITY`.

kjnilsson: sure but `/1` is a lot more compact :)

kjnilsson: actually I will tweak it

michaelklishin (approved):

ansd (approved):

PR #11843, opened 2024-07-26
Report frame_max as integer

ansd:
Resolves https://github.com/rabbitmq/rabbitmq-server/issues/11838


lukebakken (approved):
Tested and works great, thanks!

PR #11672, opened 2024-07-11
Bump Ra from 2.13.6 to 2.14.0 + Khepri from 0.14.0 to 0.15.0

dumbbell:
Khepri is updated from version 0.14.0 to 0.15.0. This brings many improvements to queries and fixes to projections, along with several breaking changes. In particular, the `favor` default value is now `low_latency`. Therefore, we don't need to specify it explicitly anymore.

Ra is updated at the same time because this new version of Khepri depends on Ra 2.14.0.

Release notes:
* Ra: https://github.com/rabbitmq/ra/releases/tag/v2.14.0
* Khepri: https://github.com/rabbitmq/khepri/releases/tag/v0.15.0

PR #11879, opened 2024-08-01
Move projection registration out of Khepri setup

the-mikedavis:
`rabbit_khepri:setup/0` is always called during boot for the sake of clustering. Previously we registered projections during `rabbit_khepri:setup/0` meaning that even if you run Rabbit 3.13.x without enabling the `khepri_db` feature flag you will attempt to register projections. It's a benign bug - it doesn't cause any bad behavior other than a potential timeout on boot if a server is in a minority - but it modifies the Khepri store so we need to consider the projections when we reorganize the tree (#11225).

Instead of registering projections during `rabbit_khepri:setup/0` we should register them in two spots:

* In init (`rabbit_db:init/0`), a boot step run after `rabbit_khepri:setup/0` which has a Khepri-specific branch, and only if we detect that the database is blank. (We could always register on init when Khepri is enabled but there's no need to unless the database is blank.)
* In the enable callback for the `khepri_db` feature flag.

In the future when we want to unregister projections, we can add that step before `rabbit_khepri:register_projections/0` to the enable callback for the `khepri_db` feature flag (using https://github.com/rabbitmq/khepri/pull/282).

### DIFF
--- a/deps/rabbit/src/rabbit_db.erl
+++ b/deps/rabbit/src/rabbit_db.erl
@@ -67,8 +67,8 @@ init() ->
     end,
 
     Ret = case rabbit_khepri:is_enabled() of
-              true  -> init_using_khepri();
-              false -> init_using_mnesia()
+              true  -> init_using_khepri(IsVirgin);
+              false -> init_using_mnesia(IsVirgin)
           end,
     case Ret of
         ok ->
@@ -91,7 +91,7 @@ pre_init(IsVirgin) ->
     OtherMembers = rabbit_nodes:nodes_excl_me(Members),
     rabbit_db_cluster:ensure_feature_flags_are_in_sync(OtherMembers, IsVirgin).
 
-init_using_mnesia() ->
+init_using_mnesia(_IsVirgin) ->
     ?LOG_DEBUG(
       "DB: initialize Mnesia",
       #{domain => ?RMQLOG_DOMAIN_DB}),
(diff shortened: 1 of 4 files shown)


deps/rabbit/src/rabbit_db.erl
```diff
@@ -91,23 +91,29 @@ pre_init(IsVirgin) ->
...
                "Found the following metadata store members: ~p", [Members],
-               #{domain => ?RMQLOG_DOMAIN_DB})
+               #{domain => ?RMQLOG_DOMAIN_DB}),
+            case IsVirgin of
+                true ->
+                    rabbit_khepri:register_projections();
+                false ->
+                    ok
+            end
     end.
```
dumbbell, line 117: This starts to be a lot of Khepri-specific knowledge outside of `rabbit_khepri`. Would you mind moving this function to `rabbit_khepri`? E.g. `rabbit_khepri:init/1` like Mnesia.

While here, the log message level should be decreased to NOTICE or INFO probably. Perhaps we can add the equivalent of the "DB: initialize Mnesia" log message too.

dumbbell (changes requested):

the-mikedavis:
Yeah let's wait for that since we also want the `fence` API to await a leader (as [mentioned above](https://github.com/rabbitmq/rabbitmq-server/pull/11879#discussion_r1713455937))

dumbbell:
Khepri 0.15.0 is now released and used in RabbitMQ `main` branch.

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -1104,20 +1127,22 @@ register_projections() ->
...
-             throw(Error)
-     end || RegisterFun <- RegFuns],
-    ok.
+    rabbit_misc:for_each_while_ok(
+      fun(RegisterFun) ->
+              case RegisterFun() of
+                  ok ->
+                      ok;
+                  %% Before Khepri v0.13.0, `khepri:register_projection/1,2,3` would
+                  %% return `{error, exists}` for projections which already exist.
```
dumbbell, line 1136: The comments should be wrapped to fit into the 80-col limit.

dumbbell (changes requested):
The patch looks good to me. I just made a minor style-related comment.

dumbbell (approved):
Thank you!

the-mikedavis:
I'll set up the backports once we have the prerequisite PRs backported. In particular this PR needs https://github.com/rabbitmq/rabbitmq-server/pull/11672

PR #11898, opened 2024-08-05
Classic peer discovery: node list warnings

dcorbacho:
Log warnings when:
- Local node is not present. Even though we force it on the node list, this will not work for other cluster nodes if they have the same list.
- There are duplicated nodes

Issue https://github.com/rabbitmq/rabbitmq-server/issues/11586


lukebakken (approved):
Great!

Issue #11789, opened 2024-07-23
`pre_init failed in system quorum_queues` when upgrading from 3.13.4 to 3.13.5


headyj:
### Describe the bug

when changing from `rabbitmq:3.13.5-management` to `rabbitmq:3.13.5-management`, the following error appears:
```
rabbitmq 2024-07-23 10:10:48.033045+00:00 [error] <0.268.0> pre_init failed in system quorum_queues for UId ORCHESLVOCD9VDIAR9 with name *****_898e605a-33e3-4129-9906-f6f46a65a695 This error may need manual intervention
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>   crasher:
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     initial call: ra_log_pre_init:init/1
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     pid: <0.268.0>
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     registered_name: ]
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     exception exit: {error,{badmatch,false}}
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>       in function  gen_server:init_it/6 (gen_server.erl, line 961)
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     ancestors: [ra_log_sup,<0.261.0>,ra_systems_sup,ra_sup,<0.208.0>]
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     message_queue_len: 0
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     messages: ]
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     links: [<0.267.0>]
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     dictionary: ]
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     trap_exit: false
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     status: running
rabbitmq 2024-07-23 10:10:48.033175+00:00 [error] <0.268.0>     heap_size: 1598
[...]

PR #11967, opened 2024-08-09
Add `cluster_upgrade_SUITE` to check mixed-version upgrades

the-mikedavis:
This suite uses the mixed version secondary umbrella as a starting version for a cluster and then has a helper to upgrade the cluster to the current code. This is meant to ensure that we can upgrade from the previous minor. This suite fails when downgrading Ra to 2.13.4 or 2.13.3 for example, see https://github.com/rabbitmq/rabbitmq-server/issues/11789

### DIFF
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -1209,6 +1209,14 @@ rabbitmq_integration_suite(
 )
 
 rabbitmq_integration_suite(
+    name = "cluster_upgrade_SUITE",
+    size = "medium",
+    additional_beam = [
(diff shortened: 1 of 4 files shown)


the-mikedavis:
A run of the CI job that runs this test is: https://github.com/rabbitmq/rabbitmq-server/actions/runs/10325431081/job/28592434245?pr=11967

(I was curious if we need to adjust the CI workflows with `make actions-workflows` after adding a `rabbit` SUITE but the sharding seems to take care of that automatically.)

michaelklishin (approved):

PR #11979, opened 2024-08-12
Handle transient queue deletion in Khepri minority

the-mikedavis:
Transient queue deletion previously caused a crash if Khepri was enabled and a node with a transient queue went down while its cluster was in a minority. We need to handle the `{error,timeout}` return possible from `rabbit_db_queue:delete_transient/1`. In the `rabbit_amqqueue:on_node_down/1` callback we log a warning when we see this return.

We then try this deletion again during that node's `rabbit_khepri:init/0` which is called from a boot step after `rabbit_khepri:setup/0`. At that point we can return an error and halt the node's boot if the command times out. The cluster is very likely to be in a majority at that point since `rabbit_khepri:setup/0` waits for a leader to be elected (requiring a majority).

This fixes a crash report found in the `cluster_minority_SUITE`'s `end_per_group`.

### DIFF
--- a/deps/rabbit/src/rabbit_db.erl
+++ b/deps/rabbit/src/rabbit_db.erl
@@ -100,15 +100,10 @@ init_using_mnesia() ->
     rabbit_sup:start_child(mnesia_sync).
 
 init_using_khepri() ->
-    case rabbit_khepri:members() of
-        [] ->
-            timer:sleep(1000),
-            init_using_khepri();
-        Members ->
-            ?LOG_NOTICE(
-               "Found the following metadata store members: ~p", [Members],
-               #{domain => ?RMQLOG_DOMAIN_DB})
-    end.
+    ?LOG_DEBUG(
+      "DB: initialize Khepri",
+      #{domain => ?RMQLOG_DOMAIN_DB}),
+    rabbit_khepri:init().
 
 init_finished() ->
     %% Used during initialisation by rabbit_logger_exchange_h.erl
--- a/deps/rabbit/src/rabbit_db.erl
+++ b/deps/rabbit/src/rabbit_db.erl
@@ -105,7 +105,7 @@ init_using_khepri() ->
             timer:sleep(1000),
             init_using_khepri();
         Members ->
-            ?LOG_WARNING(
+            ?LOG_NOTICE(
                "Found the following metadata store members: ~p", [Members],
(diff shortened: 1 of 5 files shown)


deps/rabbit/src/rabbit_db.erl
```diff
@@ -100,15 +100,10 @@ init_using_mnesia() ->
...
-            init_using_khepri();
-        Members ->
-            ?LOG_WARNING(
-               "Found the following metadata store members: ~p", [Members],
-               #{domain => ?RMQLOG_DOMAIN_DB})
-    end.
+    ?LOG_DEBUG(
+      "DB: initialize Khepri",
+      #{domain => ?RMQLOG_DOMAIN_DB}),
+    rabbit_khepri:init().
```
the-mikedavis, line 106: @dumbbell I have addressed one of your reviews https://github.com/rabbitmq/rabbitmq-server/pull/11879#discussion_r1713487430 in this PR instead since it doesn't depend on the new Khepri version and also needs to hook into Khepri's init

deps/rabbit/src/rabbit_db_queue.erl
```diff
@@ -1073,26 +1074,63 @@ delete_transient_in_khepri(FilterFun) ->
...
+                           end
+                   end, [], Props),
+            case Qs of
+                [] ->
+                    %% If there are no changes to make, avoid performing a
+                    %% transaction. When Khepri is in a minority this avoids
+                    %% a long timeout waiting for the transaction command to
+                    %% be processed. Otherwise it avoids appending a somewhat
+                    %% large transaction command to Khepri's log.
+                    {[], []};
```
dumbbell, line 1100: Should this be handled as a function clause of `do_delete_transient_queues_in_khepri()` instead? The optimization would be part of the function, instead letting that to the caller(s).

deps/rabbit/src/rabbit_db.erl
dumbbell: Great!

dumbbell (changes requested):

deps/rabbit/src/rabbit_db_queue.erl
the-mikedavis: Good point, that makes `delete_transient_in_khepri/1` cleaner

dumbbell (approved):

PR #11998, opened 2024-08-14
Make it hard to accidentally enable an experimental feature flag

mkuratczyk:
Currently it's easy to enable an experimental feature flag like `khepri_db` by accident. Given the operation is irreversible, this leads to serious consequences when done in a production environment.

This PR implements the following changes:
* Management UI
  * Split the feature flags page into two tables - first with all but experimental feature flags, the second with only the experimental flags (and a warning)
  * To enable an experimental feature flag, a checkbox needs to be toggled first
* CLI
  * `rabbitmqctl enable_feature_flag khepri_db` won't work unless a new flag `--force` is provided

Screenshot 2024-08-14 at 12 19 57

### DIFF
--- a/deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/enable_feature_flag_command.ex
+++ b/deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/enable_feature_flag_command.ex
@@ -22,10 +22,15 @@ defmodule RabbitMQ.CLI.Ctl.Commands.EnableFeatureFlagCommand do
 
   use RabbitMQ.CLI.Core.RequiresRabbitAppRunning
 
-  def run(["all"], %{node: node_name}) do
-    case :rabbit_misc.rpc_call(node_name, :rabbit_feature_flags, :enable_all, []) do
-      {:badrpc, _} = err -> err
-      other -> other
+  def run(["all"], %{node: node_name, experimental: experimental}) do
+    case experimental do
+      true ->
+        {:error, RabbitMQ.CLI.Core.ExitCodes.exit_usage(), "`--experiemntal` flag is not allowed when enabling all feature flags.\nUse --experimental with a specific feature flag if you want to enable an experimental feature."}
+      false ->
+        case :rabbit_misc.rpc_call(node_name, :rabbit_feature_flags, :enable_all, []) do
+          {:badrpc, _} = err -> err
+          other -> other
+        end
     end
   end
 
--- a/deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/enable_feature_flag_command.ex
+++ b/deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/enable_feature_flag_command.ex
@@ -7,15 +7,18 @@
 defmodule RabbitMQ.CLI.Ctl.Commands.EnableFeatureFlagCommand do
   @behaviour RabbitMQ.CLI.CommandBehaviour
 
-  def merge_defaults(args, opts), do: {args, opts}
+  def switches(), do: [force: :boolean]
+  def aliases(), do: [f: :force]
 
-  def validate([], _), do: {:validation_failure, :not_enough_args}
-  def validate([_ | _] = args, _) when length(args) > 1, do: {:validation_failure, :too_many_args}
+  def merge_defaults(args, opts), do: { args, Map.merge(%{force: false}, opts) }
 
-  def validate([""], _),
+  def validate([], _opts), do: {:validation_failure, :not_enough_args}
+  def validate([_ | _] = args, _opts) when length(args) > 1, do: {:validation_failure, :too_many_args}
+
+  def validate([""], _opts),
     do: {:validation_failure, {:bad_argument, "feature_flag cannot be an empty string."}}
 
-  def validate([_], _), do: :ok
+  def validate([_], _opts), do: :ok
 
   use RabbitMQ.CLI.Core.RequiresRabbitAppRunning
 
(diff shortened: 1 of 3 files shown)


dumbbell:
What about naming the CLI flag `--experimental` to make sure the user knows it’s is an experimental feature flag?

Also, I feel that `--force` is too generic and could mean many things. I could interpret it as "I want this feature flag to be enabled even if some nodes are not running in the cluster" (which is denied today for instance).

michaelklishin:
I agree with @dumbbell and would use `--experimental` instead of `--force`.

michaelklishin:
The forced push was a rebase.

mkuratczyk:
I've renamed the flag and slightly changed the error message (+rebase)

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/enable_feature_flag_command.ex
```diff
@@ -29,32 +32,45 @@ defmodule RabbitMQ.CLI.Ctl.Commands.EnableFeatureFlagCommand do
...
-      # Server does not support feature flags, consider none are available.
-      # See rabbitmq/rabbitmq-cli#344 for context. MK.
-      {:badrpc, {:EXIT, {:undef, _}}} -> {:error, :unsupported}
-      {:badrpc, _} = err -> err
-      other -> other
+  def run([feature_flag], %{node: node_name, experimental: experimental}) do
+    case {experimental, :rabbit_misc.rpc_call(node_name, :rabbit_feature_flags, :get_stability, [
+      String.to_atom(feature_flag)
+    ])} do
+          {_, {:badrpc, {:EXIT, {:undef, _}}}} -> {:error, :unsupported}
```
dumbbell, line 39: I think you can drop the `undef` line: a node with this patch can’t talk to a 3.7 node anyway.

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/enable_feature_flag_command.ex
```diff
@@ -7,15 +7,18 @@
 defmodule RabbitMQ.CLI.Ctl.Commands.EnableFeatureFlagCommand do
   @behaviour RabbitMQ.CLI.CommandBehaviour
 
-  def merge_defaults(args, opts), do: {args, opts}
+  def switches(), do: [experimental: :boolean]
+  def aliases(), do: [f: :experimental]
```
dumbbell, line 11: I have a hard time reading Elixir, but does this means that `-f` is an alternative to `--experimental`?

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/enable_feature_flag_command.ex
```diff
@@ -29,32 +32,45 @@ defmodule RabbitMQ.CLI.Ctl.Commands.EnableFeatureFlagCommand do
...
 
   def usage_additional() do
     [
       [
         "<feature_flag>",
         "name of the feature flag to enable, or \"all\" to enable all supported flags"
+      ],
+      [
+        "--experimental",
+        "required to enable experimental feature flags (make sure you understand the risks!))"
```
dumbbell, line 73: There are two closing brackets in this sentence.

deps/rabbitmq_management/priv/www/js/tmpl/feature-flags.ejs
```diff
@@ -76,3 +79,74 @@
...
 </div>
+
+
+
+<div class="section">
+  <h2>Experimental Feature Flags</h2>
+  <div class="hider">
+<% if (feature_flags.length > 0) { %>
+<p class="warning">
+  Feature flags listed below are experimental. They should not be enabled in a produciton deployment.
```
dumbbell, line 90: There is a typo in "production".

dumbbell (changes requested):

mkuratczyk:
Thanks for the feedback, I've pushed another commit fixing all the issues you pointed out. One question that came to mind is what should happen when `rabbitmqctl enable_feature_flag --experimental all` is invoked. Currently it still won't enable experimental flags and while intuitively, it should, given the goal of this PR is to prevent accidental enablement of experimental features, I feel like ignoring `--experimental` when `all` is used is fine. I can easily be persuaded though. Thoughts?

dumbbell:
Instead of ignoring, I think we should either do what it says (enable everithing including experimental feature flags), or reject the command with an error.

As you said, given the goal, I would problably reject the command because the risks will be very specific to each experimental feature  flag.

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/enable_feature_flag_command.ex
```diff
@@ -7,54 +7,63 @@
...
-      other -> other
+  def run([feature_flag], %{node: node_name, experimental: experimental}) do
+    case {experimental, :rabbit_misc.rpc_call(node_name, :rabbit_feature_flags, :get_stability, [
+      String.to_atom(feature_flag)
+    ])} do
+          {_, {:badrpc, _} = err} -> err
+          {false, :experimental} ->
+              IO.puts("Feature flag #{feature_flag} is experimental. If you understand the risk, use --experimental to enable it.")
+          _ ->
+                case :rabbit_misc.rpc_call(node_name, :rabbit_feature_flags, :enable, [
```
dumbbell, line 40: As you removed the existing `undef` case clause, could you please put a note about why in the commit message? Because this is unrelated to the patch otherwise.

mkuratczyk:
`--experimental` is now rejected when `all` flags are to be enabled. Commit message improved

michaelklishin:
The forced push was a rebase.

michaelklishin (approved):

PR #12032, opened 2024-08-16
Maintain order of configured SASL mechanisms

ansd:
RabbitMQ should advertise the SASL mechanisms in the order as configured in `rabbitmq.conf`.

Starting RabbitMQ with the following `rabbitmq.conf`:
```
auth_mechanisms.1 = PLAIN
auth_mechanisms.2 = AMQPLAIN
auth_mechanisms.3 = ANONYMOUS
```

translates prior to this commit to:
```
1> application:get_env(rabbit, auth_mechanisms).
{ok,['ANONYMOUS','AMQPLAIN','PLAIN']}
```

and after this commit to:
```
1> application:get_env(rabbit, auth_mechanisms).
{ok,['PLAIN','AMQPLAIN','ANONYMOUS']}
```

In our 4.0 docs we write:
> The server mechanisms are ordered in decreasing level of preference.

which complies with https://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-security-v1.0-os.html#type-sasl-mechanisms.

### DIFF
--- a/deps/rabbitmq_ct_helpers/src/rabbit_ct_config_schema.erl
+++ b/deps/rabbitmq_ct_helpers/src/rabbit_ct_config_schema.erl
@@ -25,10 +25,14 @@ run_snippets(Config) ->
     {ok, [Snippets]} = file:consult(?config(conf_snippets, Config)),
     ct:pal("Loaded config schema snippets: ~tp", [Snippets]),
     lists:map(
-        fun({N, S, C, P})    -> ok = test_snippet(Config, {snippet_id(N), S, []}, C, P);
-           ({N, S, A, C, P}) -> ok = test_snippet(Config, {snippet_id(N), S, A},  C, P)
-        end,
-        Snippets),
+      fun({N, S, C, P}) ->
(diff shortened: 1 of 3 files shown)


deps/rabbitmq_ct_helpers/src/rabbit_ct_config_schema.erl
```diff
@@ -50,8 +54,12 @@ test_snippet(Config, Snippet = {SnipID, _, _}, Expected, _Plugins) ->
                     _ ->
                         generate_config(ConfFile, AdvancedFile)
                 end,
-    Gen = deepsort(Generated),
-    Exp = deepsort(Expected),
+    {Exp, Gen} = case Sort of
```
michaelklishin, line 57: I like this option.

michaelklishin (approved):

michaelklishin:
@ansd to be extra safe, we might want to re-sort one more time in the AMQP 1.0 plugin to make sure that `ANONYMOUS` is never in the first position. But I guess someone somewhere one day will argue that they "absolutely must use `ANONYMOUS` only in their environment…

PR #12035, opened 2024-08-16
Management: Add segment count to stream queue page

kjnilsson:
Also improve the help message for the 'Messages' count.

image


kjnilsson:
https://github.com/rabbitmq/osiris/pull/168

Fixes the issue where the segment count is 0 after an empty stream has been recovered.

michaelklishin (approved):

PR #12074, opened 2024-08-21
Cancel AMQP stream consumer when local stream member is deleted

### LINKED ISSUE #11915: Cancel AMQP stream consumer when local stream member is deleted
An AMQP stream consumer hangs when its local stream is deleted.

Step to reproduce:
 * spin up a 3-node cluster
 * create a stream
 * register an AMQP consumer on one of the nodes with a follower
 * use `rabbitmq-streams delete_replica` on the node
 * publish messages to the stream
 * the consumer should not receive anything

The consumer reader process is gone and there is no way to recover it as the node does not have a member of the stream anymore, so it should be cancelled/detached.

See #4133, #4217

dcorbacho:
The consumer reader process is gone and there is no way to recover it as the node does not have a member of the stream anymore, so it should be cancelled/detached.

Fixes #11915

### DIFF
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -832,7 +832,7 @@ rabbitmq_integration_suite(
     additional_beam = [
         ":test_queue_utils_beam",
     ],
-    shard_count = 19,
+    shard_count = 20,
     deps = [
         "@proper//:erlang_app",
     ],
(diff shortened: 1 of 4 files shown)


kjnilsson (changes requested):
The original idea with the code in the coordinator was that update_stream/3 should nto return any effects and that the detection of side effects would be done in the eva_* functions. Can this same logic not be added to `eval_listener/4`?

dcorbacho:
> The original idea with the code in the coordinator was that `update_stream/3` should not return any effects and that the detection of side effects would be done in the `eval_*` functions. Can this same logic not be added to `eval_listener/4`?

Sure, I wasn't aware of it. Just updated to generate the event in `eval_listener`

michaelklishin:
Rebased.

acogoluegnes (approved):
Tested against a 3-node cluster and PerfTest, the consumer is correctly canceled when the replica is deleted.

PR #12091, opened 2024-08-22
Run with default wal_sync_method

ansd:
...which is `datasync`

RA never pre-allocates the WAL anymore unless explicitly configured to.


michaelklishin:
FTR, this is [how Erlang's `file:datasync/1` behaves](https://www.erlang.org/docs/19/man/file#datasync-1) on platforms where `fdatasync(2)` is not available.

The wording is a bit cryptic but in combination with [this part](https://github.com/erlang/otp/blob/34e3822a3ca6e3794ca0bd2758cb941702b0dc1e/erts/emulator/nifs/unix/unix_prim_file.c#L542-L551) it leads me to believe that `file:datasync/1` will try to do its best but otherwise should not be harmful on Windows, macOS and older Linux kernels.

michaelklishin:
Here's the [Windows implementation](https://github.com/erlang/otp/blob/34e3822a3ca6e3794ca0bd2758cb941702b0dc1e/erts/emulator/nifs/win32/win_prim_file.c#L754-L766). My reading of it is that it will fall back to its `fsync(2)` ("full sync") equivalent.

PR #12109, opened 2024-08-24
Default queue type (DQT): fall back to node-wide default when virtual host has no metadata set

michaelklishin:
when virtual host does not have any metadata.

With this change, an `x-queue-type` header will be injected into optional queue arguments
and returned as part of definitions and the HTTP API in virtually all cases.

This also relaxes the `assert_list/2` assertion a bit: it
no longer requires the size of an actually returned list element
to be exactly equal to the size of the expected one.

Sometimes it makes perfect sense to not assert on
every single key but only a subset, and with this
change, it now will be possible.

Individual tests may choose to assert on all
keys by listing them explicitly.

References #11541 #11457 #11528


michaelklishin:
I'm adding more redeclaration tests for classic queues just in case:

 * For HTTP API, such as https://github.com/michaelklishin/rabbitmq-http-api-rs/commit/4cf4acd80dc60aeba43a2e9cb7a35bd5a3512572
 * For AMQP 0-9-1, such as https://github.com/ruby-amqp/bunny/commit/8e79ee852d16a5d10db5b5717712a066f32ca20e
 
 In particular, the behavior of redeclaration of a classic queue with `x-queue-type` and without any `x-queue-type` set does not change according to the above tests and my manual testing.

michaelklishin:

PR #12118, opened 2024-08-25
MQTT and Streams: handle connection shutdown via CLI command gracefully

### LINKED ISSUE #11985: Crashes when closing MQTT or Stream connections via CLI
### Describe the bug

`mqtt_unexpected_call` exception when closing MQTT connection

### Reproduction steps

1. start an MQTT connection (idle consumer in this example)
```omq mqtt -x 0```

1. Close all `guest` connections
```rabbitmqctl  close_all_user_connections guest bye```

1. The connection is closed, but there's an exception in the logs
```
{mqtt_unexpected_call,{shutdown,<<"bye">>},
                      {<0.1238.0>,#Ref<0.3075749135.3916431361.188844>}}
Client <0.1238.0> stacktrace
[{gen,do_call,4,[{file,"gen.erl"},{line,241}]},
 {gen_server,call,3,[{file,"gen_server.erl"},{line,1122}]},
 {rabbit_networking,close_connection,2,
                    [{file,"rabbit_networking.erl"},{line,521}]},
 {rabbit_networking,'-close_all_user_connections/2-lc$^1/1-1-',2,
                    [{file,"rabbit_networking.erl"},{line,535}]},
 {rabbit_networking,close_all_user_connections,2,
                    [{file,"rabbit_networking.erl"},{line,535}]},
 {erpc,execute_call,4,[{file,"erpc.erl"},{line,1250}]}]
```

### Expected behavior

Connection closed, the fact logged without a stacktrace

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by dcorbacho**
`{shutdown, Reason}` must be handled into handle_call and not handle_info `rabbitmqctl close_all_user_connections` calls rabbit_reader which does a call into the process, the same as rabbitmq_management

Fixes #11985

### DIFF
--- a/deps/rabbit/src/rabbit_connection_tracking.erl
+++ b/deps/rabbit/src/rabbit_connection_tracking.erl
@@ -427,11 +427,7 @@ close_connection(#tracked_connection{pid = Pid, type = direct}, Message) ->
     %% Do an RPC call to the node running the direct client.
     Node = node(Pid),
     rpc:call(Node, amqp_direct_connection, server_close, [Pid, 320, Message]);
-close_connection(#tracked_connection{pid = Pid,
-                                     protocol = {'Web MQTT', _}}, Message) ->
-    % this will work for connections to web mqtt plugin
-    Pid ! {shutdown, Message};
 close_connection(#tracked_connection{pid = Pid}, Message) ->
-    % best effort, this will work for connections to the stream plugin
-    Node = node(Pid),
-    rpc:call(Node, gen_server, call, [Pid, {shutdown, Message}, infinity]).
+    %% Best effort will work for following plugins:
+    %% rabbitmq_stream, rabbitmq_mqtt, rabbitmq_web_mqtt
+    Pid ! {shutdown, Message}.
--- a/deps/rabbit/src/rabbit_connection_tracking.erl
+++ b/deps/rabbit/src/rabbit_connection_tracking.erl
@@ -427,6 +427,10 @@ close_connection(#tracked_connection{pid = Pid, type = direct}, Message) ->
     %% Do an RPC call to the node running the direct client.
     Node = node(Pid),
     rpc:call(Node, amqp_direct_connection, server_close, [Pid, 320, Message]);
+close_connection(#tracked_connection{pid = Pid,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by ansd — CHANGES_REQUESTED** [review]
1. This PR introduces a crash as can be seen in test case `management_plugin_connection` in the `shared_SUITE`.
For example running
```
make -C deps/rabbitmq_mqtt ct-shared t=[mqtt,cluster_size_1,v5]:management_plugin_connection
```
will make the MQTT connection crash with:
```
2024-08-26 08:47:10.764564+02:00 [error] <0.1022.0> ** Reason for termination ==
2024-08-26 08:47:10.764564+02:00 [error] <0.1022.0> ** {mqtt_unexpected_msg,{shutdown,"Closed via management plugin"}}
```

2. This PR doesn't handle WebMQTT connections.

3. It would be nice to add a test case for this change.

4. Ideally, we refactor such that closing connections via the Management API and via the CLI command both cause the same logic to happen in RabbitMQ, i.e. cause either `cast` or `call` for (Web)MQTT connections.

ansd:
Good spot. That's because
5: This PR returns `{stop,Reason,NewState}` instead of `{stop,Reason,Reply,NewState}`.

dcorbacho:
Thanks. I've just seen the issue. `rabbitmq_management` calls `rabbit_networking:close_connection` as expected for `network` connections, but on the same case sends a message for everything else :( https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_management/src/rabbit_mgmt_wm_connection.erl#L101

dcorbacho:
@ansd I've restored the original `handle_info` clause. Changing the management to perform a call when forcing close the connections would be a breaking change for mqtt and streams, it would require a feature flag for mixed-version clusters. Not sure it is worth it.

mkuratczyk (approved):
Functionality-wise I see no problems now, thanks. I'll let David comment on the implementation.

ansd:
Thank you @dcorbacho.
I noticed that Stream connections also crash.
I pushed two commits which simplify a lot (see commit messages for details) and fix the Stream connection crash as well (manually tested).

ansd (approved):

PR #12140, opened 2024-08-28
Remove some dead code and configuration options

mkuratczyk:
* ` memory_high_watermark_paging_ratio` and `memory_monitor_interval` are not used in CQv2 - the only version supported in 4.0
* classic_queue_default_version is now ignored, since only version 2 is supported
* I've never heard of `halt_on_upgrade_failre` and `autoscaler` and they don't appear to be used anywhere
* HiPE was removed from OTP in 2020


deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -1149,6 +1149,8 @@ fun(Conf) ->
     end
 end}.
 
+%% DEPRECATED. Not used since RabbitMQ 4.0
+%%
```
ansd, line 1153: I wonder whether we can just remove all deprecated Cuttlefish translations?
The node will fail to boot if this setting in configured in `rabbitmq.conf`.
However, if 4.0 is not the right time for such a breaking change, when else is the right time? 🙂

michaelklishin: There will be fewer changes overall in 4.1.0 because release scope generally goes "in waves."

Let's not make it harder for the users to upgrade to 4.0. In fact, by removing these immediately we would not save ourselves much time in terms of support and such. Such small things can be removed fairly slowly. For example, we very slowly deprecated legacy CLI health checks, they are still around, just do nothing (and print messages that say as much). Removing them completely in 4.0 would not benefit us much.

michaelklishin (approved):

deps/rabbit/priv/schema/rabbit.schema
ansd: Fair enough

Issue #12142, opened 2024-08-28
Khepri metrics are showing up without any tags (v4 beta 5)


luos:
### Is your feature request related to a problem? Please describe.

Hi,

We're testing out Khepri and reviewing how we could monitor its behaviour and performance. 

Today, we monitor mnesia transaction counters to see if there is a high amount of churn in the system - mostly because mnesia can cause some issues if the transaction count / restarts are very high. 

We've noticed, that the metrics in the metric family `ra_metrics` show up without any tags, which I think potentially should be either in a different family, ie. `khepri` or `metadata` , or they should have proper tags, ie. for the `rabbit_metadata|quorum_queues, etc`. 

Test setup:

1. Deploy v4 beta 5
2. Enable the khepri_db feature flag
3. Create a quorum queue
4. Call the metrics api `curl localhost:15692/metrics/detailed?family=ra_metrics`
5. Call `rabbitmqctl eval 'rabbit_khepri:status().'`

Excerpt from the output:

```
$ curl localhost:15692/metrics/detailed?family=ra_metrics
rabbitmq_detailed_raft_log_last_written_index{vhost="/",queue="qq1"} 2
rabbitmq_detailed_raft_log_last_written_index 49
```

```
$ rabbitmqctl eval 'rabbit_khepri:status().'
...  {<<"Last Written">>,49}, ...

```

### Describe the solution you'd like

- Metric is tagged with something like `ra_process=rabbit_metadata` 
- Metric is served under a different family and different name. 

### Describe alternatives you've considered

I tried to look at the metric collection code, but from my cursory review I could not figure out how to add the tag, and not even sure that would be the preferred way to go about it. 😄 

### Additional context

Due to Khepri's consistency behaviour with the projections, It would be good to know if a node is falling behind.

michaelklishin:
We can both use a tag (if we do so for other Ra machine process types) and provide a new endpoint or a metric family.

luos:
Thanks, yeah, I was thinking similar to what the-mikedavis proposed in the PR. 

I think it is worth considering a different family, ie. we may have a many thousands of queues but only one/few metadata processes, and I expect we will care more about khepri than QQ indexes, though I can't say for sure at this point in time. :-)

At the same time, I do not expect the khepri process to take a lot of traffic - but I am sure it will happen.

mkuratczyk:
Closing as implemented. This is what this looks like these days:
```
[...]

PR #12148, opened 2024-08-28
Add prometheus tags "raft_cluster" to non-QQ raft metrics

the-mikedavis:
Adds a `<<"raft_cluster">>` tag for ra_metrics emitted the stream coordinator and Khepri. By default Ra sets the cluster name as the `metrics_key` (if not set in config, like it is for QQs). The prometheus plugin ignores atoms that aren't protocol names when rendering labels so these cluster name values disappear. We can convert them into a map `#{<<"raft_cluster">> => atom_to_binary(ClusterName, utf8)}` to render properly.

Before:

    $ curl -s "localhost:15692/metrics/detailed?family=ra_metrics" | grep rabbitmq_raft_term_total
    # TYPE rabbitmq_raft_term_total counter
    # HELP rabbitmq_raft_term_total Current Raft term number
    rabbitmq_raft_term_total{vhost="/",queue="qq"} 9
    rabbitmq_raft_term_total 10

After:

    $ curl -s "localhost:15692/metrics/detailed?family=ra_metrics" | grep rabbitmq_raft_term_total
    # TYPE rabbitmq_raft_term_total counter
    # HELP rabbitmq_raft_term_total Current Raft term number
    rabbitmq_raft_term_total{vhost="/",queue="qq"} 9
    rabbitmq_raft_term_total{raft_cluster="rabbitmq_metadata"} 10

Connects https://github.com/rabbitmq/rabbitmq-server/discussions/9665
Connects #12142

### DIFF
--- a/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
+++ b/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
@@ -640,6 +640,19 @@ get_data(Table, false, VHostsFilter) when Table == channel_exchange_metrics;
         _ ->
             [Result]
     end;
+get_data(ra_metrics = Table, true, _) ->
+    ets:foldl(
+      fun ({#resource{kind = queue}, _, _, _, _, _, _} = Row, Acc) ->
+              %% Metrics for QQ records use the queue resource as the table
+              %% key. The queue name and vhost will be rendered as tags.
+              [Row | Acc];
+          ({ClusterName, _, _, _, _, _, _} = Row, Acc) when is_atom(ClusterName) ->
+              %% Other Ra clusters like Khepri and the stream coordinator use
+              %% the cluster name as the metrics key. Transform this into a
+              %% value that can be rendered as a "raft_cluster" tag.
+              Row1 = setelement(1, Row, #{<<"raft_cluster">> => atom_to_binary(ClusterName, utf8)}),
+              [Row1 | Acc]
+      end, [], Table);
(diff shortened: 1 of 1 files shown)


deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -1256,11 +1256,12 @@ make_ra_conf(Node, Nodes) ->
     Members = [{?MODULE, N} || N <- Nodes],
     TickTimeout = application:get_env(rabbit, stream_tick_interval,
                                       ?TICK_TIMEOUT),
+    MetricsKey = #{<<"raft_cluster">> => atom_to_binary(?MODULE)},
     #{cluster_name => ?MODULE,
       id => {?MODULE, Node},
       uid => UId,
       friendly_name => atom_to_list(?MODULE),
-      metrics_key => ?MODULE,
+      metrics_key => MetricsKey,
```
the-mikedavis, line 1264: This looks like a breaking change but non-protocol atoms are omitted from labels: https://github.com/rabbitmq/rabbitmq-server/blob/afa28cbdb3af99e60943159a815a669bfa16367a/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl#L500-L504

(Unless something else is reading out of the `ra_metrics` table and expecting this value)

the-mikedavis:
This doesn't quite solve #12142 but fixes the issue of these Raft metrics showing up without a label for Khepri and the stream coordinator. There are some larger changes @kjnilsson and I have been looking at for seshat, ra and here to improve how we expose Raft metrics: switching from the `ra_metrics` table to Ra counters, improving seshat's usability for Prometheus. In the meantime though this is a nearly trivial fix that we can make and easily backport so I will merge this down as-is and leave the rest for future work.

PR #12161, opened 2024-08-29
Set the default vm_memory_high_watermark to 0.6

mkuratczyk:
The default of 0.4 was very conservative even when it was set years ago. Since then:
- we moved to CQv2, which have much more predictable memory usage than (non-lazy) CQv1 used to
- we removed CQ mirroring which caused large sudden memory spikes in some situations
- we removed the option to store message payload in memory in quorum queues

For the past two years or so, we've been running all our internal tests and benchmarks using the value of 0.8 with no OOMkills at all (note: we do this on Kubernetes where the Cluster Operators overrides the available memory leaving some additional headroom, but effectively we are still using  more than 0.6 of memory).

Corresponding docs PR: https://github.com/rabbitmq/rabbitmq-website/pull/2032

### DIFF
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -34,7 +34,7 @@ _APP_ENV = """[
 	    {ssl_listeners, []},
 	    {num_ssl_acceptors, 10},
 	    {ssl_options, []},
-	    {vm_memory_high_watermark, 0.4},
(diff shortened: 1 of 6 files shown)


ansd (approved):
Related: https://github.com/rabbitmq/rabbitmq-server/discussions/10518

PR #12163, opened 2024-08-29
rabbit_db_cluster: Reset feature flags immediately after a failure to join

dumbbell:
## Why

If a node failed to join a cluster, `rabbit` was restarted then the feature flags were reset and the error returned. I.e., the error handling was in a single place at the end of the function.

We need to reset feature flags after a failure because the feature flags states were copied from the remote node just before the join.

However, resetting them after restarting `rabbit` was incorrect because feature flags were initialized in a way that didn't match the rest of the state. This led to crashes during the start of `rabbit`.

## How

The feature flags are now reset after the failure to join but before starting `rabbit`.

A new testcase was added to test this scenario.

### DIFF
--- a/deps/rabbit/BUILD.bazel
+++ b/deps/rabbit/BUILD.bazel
@@ -1162,7 +1162,7 @@ rabbitmq_integration_suite(
 rabbitmq_integration_suite(
     name = "metadata_store_clustering_SUITE",
     size = "large",
-    shard_count = 18,
+    shard_count = 19,
     sharding_method = "case",
 )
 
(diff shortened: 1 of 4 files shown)


michaelklishin:
@mergify backport v4.0.x

michaelklishin:
The repeatable failures in https://github.com/rabbitmq/rabbitmq-server/pull/12192 suggests this PR has more `main` dependencies than meets the eye => it will be backported another day or maybe even will have to wait till `4.1.0`.

dumbbell:
It is not ready to be backported anyway. It will be with the series of pull requests that go together.

dumbbell:
@Mergifyio backport v4.0.x

PR #12208, opened 2024-09-04
Adjust vheap sizes for message handling processes in OTP 27

kjnilsson:
OTP 27 reset all assumptions on how the vm reacts to processes that
buffer and process a lot of large binaries.

Substantially increasing the vheap sizes for such process restores
most of the same performance by allowing processes to hold more binary
data before major garbage collections are triggered.

This introduces a new module to capture process flag configurations.

The new vheap sizes are only applied when running on OTP 27 or
above.


mkuratczyk:
64B messages (represented differently within OTP than binaries above 64B):
msg-size-64B

65B messages:
msg-size-65B

100B messages:
msg-size-100B

1kb messages:
msg-size-1kb

5kb messages:
msg-size-5kb

Full results:
https://grafana.lionhead.rabbitmq.com/goto/aDIKeWeIR?orgId=1

michaelklishin:
My interpretation of the benchmarks can be summarized as: with this QQ process configuration the throughput becomes comparable, if not better (with smallest messages), on Erlang 27 but latency is worse, in particular the 99% percentile/max.

I think it's a reasonable outcome and will allow us to support Erlang 27 in 4.0 without major regressions.

michaelklishin (approved):

mkuratczyk (approved):

michaelklishin:
The forced push was a rebase. @mkuratczyk I assume this change is still relevant, correct?

mkuratczyk:
Yes, I think we can merge it. @kjnilsson anything else you wanted to do here?

michaelklishin:
Another rebase.

PR #12221, opened 2024-09-05
Enforce AMQP 1.0 channel-max

ansd:
Enforce AMQP 1.0 field `channel-max` in the `open` frame by introducing a new more user friendly setting called `session_max`:
> The channel-max value is the highest channel number that can be used on the connection.
> This value plus one is the maximum number of sessions that can be simultaneously active on the connection.

We set the default value of `session_max` to 64 such that, by default, RabbitMQ 4.0 allows maximum 64 AMQP 1.0 sessions per AMQP 1.0 connection.

More than 64 AMQP 1.0 sessions per connection make little sense. See also https://www.rabbitmq.com/blog/2024/09/02/amqp-flow-control#session

Limiting the maximum number of sessions per connection can be useful to protect against
* applications that accidentally open new sessions without ending old sessions (session leaks)
* too many metrics being exposed, for example in the future via the "/metrics/per-object" Prometheus endpoint with timeseries per session being emitted.

This commit does not make use of the existing `channel_max` setting because:
1. Given that `channel_max = 0` means "no limit", there is no way for an operator to limit the number of sessions per connections to 1.
2. Operators might want to set different limits for maximum number of AMQP 0.9.1 channels and maximum number of AMQP 1.0 sessions.
3. The default of `channel_max` is very high: It allows using more than 2,000 AMQP 0.9.1 channels per connection. Lowering this default might break existing AMQP 0.9.1 applications.

This commit also fixes a bug in the AMQP 1.0 Erlang client which, prior to this commit used channel number 1 for the first session. That's wrong if a broker allows maximum 1 session by replying with `channel-max = 0` in the `open` frame. Additionally, the spec recommends:
> To make it easier to monitor AMQP sessions, it is RECOMMENDED that implementations always assign the lowest available unused channel number.

Note that in AMQP 0.9.1, channel number 0 has a special meaning:
> The channel number is 0 for all frames which are global to the connection and 1-65535 for frames that
refer to specific channels.


michaelklishin:
This is much lower than the current (leave alone historic) channel max used for AMQP 0-9-1 but this is a better idea than, say, 256. Exactly because more than 64 usually indicates a leak or a misunderstanding of how channels/sessions work.

deps/rabbit/test/amqp_client_SUITE.erl
```diff
@@ -5621,6 +5622,28 @@ tcp_back_pressure_rabbitmq_internal_flow(QType, Config) ->
     ok = end_session_sync(Session),
     ok = amqp10_client:close_connection(Connection).
 
+session_max(Config) ->
+    App = rabbit,
+    Par = ?FUNCTION_NAME,
```
michaelklishin, line 5627: Should we really depend on `?FUNCTION_NAME` here? Test functions do get renamed sometimes :)

michaelklishin (approved):

deps/rabbit/test/amqp_client_SUITE.erl
ansd: My intent was that the test case tests the config value its named after: `session_max`.
I removed `?FUNCTION_NAME` 👍

PR #12232, opened 2024-09-06
WIP A PBKDF2 implementation for internal database password hashing

### LINKED ISSUE #3350: RabbitMQ should support pbkdf2 for password hashing by default
Currently by default, RabbitMQ only supports sha256 and sha512 for password hashing. These two hasing algorithms are considered having security risks nowadays. The pbkdf2 algorithm is considered more secure and is widely use. Maybe RabbitMQ should support it by default instead of letting users develop their own plugin, since erlang does not have implement pdkdf2 hashing, and many users who use RabbitMQ don't have the ability to implement it by their own.

iinuwa:
## Proposed Changes

This implements PBKDF2 password hash storage for internal authentication. The only built-in hash algorithms available for storing passwords are SHA256 and SHA512, which, while salted with a 32-bit random salt, are not secure for password storage. If an attacker gains access to the hashes stored in RabbitMQ, it would be fairly trivial to crack the salted passwords stored in RabbitMQ.

This change implements PBKDF2 HMAC-SHA512 using `crypto:pbkdf2_hmac/5`, using the [OWASP-recommended iteration count](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2) of 210,000. (This OTP method is available since Erlang 24, so it should be available for all currently supported and future versions of RabbitMQ.) The scheme is recorded as `_v1` so that future increases to the iteration count can be increased without breaking changes or having to store extra data for the user.

This is more secure against stolen hashes, though decreases performance for evaluating passwords. This means that new AMQP connections, API calls and Management UI sessions will all be more expensive. If merged, the documentation should clearly state this performance impact of using this algorithm, perhaps recommending not to use this algorithm for users that access the Management UI. Since this does not change the default password hash, and will require importing user definitions to apply to a RabbitMQ instance, this can be considered advanced usage.

Fixes #3350.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

- This could also be implemented as a plugin instead of being added to the core.
- I have not added tests, as I'm not very familiar with Erlang, Bazel or this codebase, but I have tested this locally.


lhoguin:
With this many iterations it is not surprising that it has such an impact. Perhaps in the future we can have session ID-based authentication for management UI, whenever we get to modernise it, or earlier if a good solution is figured out.

The PR looks fine but needs tests.

michaelklishin (changes requested):
Please drop the `_v1` suffix, make the number of iterations configurable and provide some numbers on what the impact on CPU consumption and opening, say, of 50K connections is.

Connection churn is not something we recommend but it is not at all unheard of. 210K iterations may be really excessive but we should not be guessing as to what the impact is.

deps/rabbit_common/src/rabbit_password_hashing_pbkdf2_hmac_sha512_v1.erl
```diff
@@ -0,0 +1,21 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
+%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%
+%% Copyright (c) 2007-2024 Broadcom. All Rights Reserved. The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries. All rights reserved.
+%%
+
+-module(rabbit_password_hashing_pbkdf2_hmac_sha512_v1).
+
+%% TODO: I don't know if I should extend this behaviour, or change the other
```
michaelklishin, line 10: Behaviors can have optional callbacks. Changing other implementations does not seem to be necessary.

lhoguin:
Closing as it has been superseded by https://github.com/rabbitmq/rabbitmq-server/pull/17316 which has been merged. Thanks!

PR #12250, opened 2024-09-09
QQ: a delivery-limit of -1 disables the delivery limit.

kjnilsson:
For cases where users want to live a bit more dangerously this commit maps a delivery limit of -1 (or any negative value) such that it disables the delivery limit and restores the 3.13.x behaviour.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -195,7 +195,14 @@ update_config(Conf, State) ->
     Overflow = maps:get(overflow_strategy, Conf, drop_head),
     MaxLength = maps:get(max_length, Conf, undefined),
     MaxBytes = maps:get(max_bytes, Conf, undefined),
-    DeliveryLimit = maps:get(delivery_limit, Conf, undefined),
+    DeliveryLimit = case maps:get(delivery_limit, Conf, undefined) of
+                        DL when is_number(DL) andalso
+                                DL < 0 ->
+                            undefined;
+                        DL ->
+                            DL
+                    end,
+
     Expires = maps:get(expires, Conf, undefined),
     MsgTTL = maps:get(msg_ttl, Conf, undefined),
     ConsumerStrategy = case maps:get(single_active_consumer_on, Conf, false) of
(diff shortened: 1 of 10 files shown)


ansd (changes requested):
If an operator wants to enforce a maximum delivery-limit (e.g. by setting an operator policy with `delivery-limit=5`), a client app can now disable the delivery limit completely:

Screenshot 2024-09-09 at 15 47 38

I don't think this is what we want?

michaelklishin:
@ansd you are right, for this setting we should simply pick the operator policy value and not the lower or higher one. So, we need to [tweak the resolving function](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_policies.erl#L211-L221).

kjnilsson:
> @ansd you are right, for this setting from now on we should simply pick the operator policy value and not the lower or higher one. So, we need to [tweak the resolving function](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_policies.erl#L211-L221).

unfortunately it isn't a simple as that as the policy resolution code doesn't include queue arg values. i'd have to add some code to a do a resolution based on all of queue arg, policy and operator policy. 

Alternatively we treat all values above some limit, say 65535, as `unlimited` rather than special casing -1. In this case the current resolution logic would work fine.

kjnilsson:
Alt we could just say that _any_ policy would override the queue arg if the queue arg is negative.

ansd:
> Alt we could just say that any policy would override the queue arg if the queue arg is negative.

Yes, I think the "more conservative" value should always win for `(x-)delivery-limit` where "more conservative" means lower value except for `-1` which means "unlimited".

Examples:
1. Queue argument = -1, Policy = 5 should result in 5
2. Queue argument = 5, Policy = -1 should result in 5
3. Queue argument = -1, Policy = -1, Operator Policy = 5 should result in 5
4. Queue argument = 0, Policy = -1, Operator Policy = -1 should result in 0

This allows operators to set guardrails via operator policies while applications can optionally choose a more conservative setting.

michaelklishin (approved):

PR #12276, opened 2024-09-11
Increase the timeout when batch deleting permissions

dcorbacho:
Deleting a vhost with many user/topic permissions (10k-100k users with one permission each) is very slow in the current implementation of Khepri. `khepri_tree:eval_keep_while_conditions` is particularly slow, mainly on OS X but it is also a bit slower in Linux (tested in Ubuntu) than the equivalent Mnesia operation to delete the permissions. 

This PR allows the `rabbitmqctl delete_vhost <vhost>` command to finish without a timeout, even if still equally slow.

### DIFF
--- a/deps/rabbit/src/rabbit_db_user.erl
+++ b/deps/rabbit/src/rabbit_db_user.erl
@@ -640,7 +640,7 @@ clear_all_permissions_for_vhost_in_khepri(VHostName) ->
                             TopicProps,
                             rabbit_khepri:collect_payloads(UserProps)),
               {ok, Deletions}
-      end, rw).
+      end, rw, #{timeout => infinity}).
 
 %% -------------------------------------------------------------------
 %% get_topic_permissions().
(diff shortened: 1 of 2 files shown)


the-mikedavis (commented):
This is the first thing deleted in `rabbit_vhost:delete/2` so calling `rabbitmqctl delete_vhost myvhost` against a node in a minority will hang with this change. I think that's ok but for cluster_minority_SUITE we will need to change the `delete_vhost` case to pass a timeout. Something like:

```erlang
delete_vhost(Config) ->
    ?assertError(
      {erpc, timeout},
      rabbit_ct_broker_helpers:rpc(
        Config, 0,
        rabbit_vhost, delete, [<<"vhost1">>, <<"acting-user">>], 1_000)).
```

the-mikedavis (approved):

michaelklishin (approved):

PR #12286, opened 2024-09-12
Tweak wording around experimental feature flags (Khepri)

michaelklishin:
This updates Khepri FF description to be more correct and to the point.

It also tweaks the management UI copywriting so
that it does not recommend against the use of
Khepri in production as it is much more mature
in 4.0.

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/tmpl/feature-flags.ejs
+++ b/deps/rabbitmq_management/priv/www/js/tmpl/feature-flags.ejs
@@ -87,7 +87,8 @@
   <div class="hider">
 <% if (feature_flags.length > 0) { %>
 <p class="warning">
-  Feature flags listed below are experimental. They should not be enabled in a production deployment.
+  Feature flags listed below are experimental (maturing). They can be enabled in production deployments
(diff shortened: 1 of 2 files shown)


dcorbacho (approved):

the-mikedavis (approved):

deps/rabbitmq_management/priv/www/js/tmpl/feature-flags.ejs
HowardTwine, line 1: Could we change the wording to 
 "Feature flags listed below are optional, these are new features to this release and will not be automatically enabled. Once enabled Khepri cannot be disabled. "
Any more wording that this gives the user a reason not to use Khepri. Sometimes less is more ;-)
The term 'experimental' in itself is enough of a reason for users to stay away.

michaelklishin: @HowardTwine most of the team have "mob-programmed" to arrive at the following wording changes https://github.com/rabbitmq/rabbitmq-server/pull/12294. There will be some doc guide changes of a similar kind.

PR #12348, opened 2024-09-19
Make rabbit_table:wait/2 silent when checking if cmq are used

dcorbacho:


Zerpet (commented):
The changes look reasonable to me, but I know no Erlang, so I'll defer the approval to @michaelklishin

michaelklishin (approved):

PR #12392, opened 2024-09-26
CQ: Fix shared store scanner missing messages

lhoguin:
Fix for https://github.com/rabbitmq/rabbitmq-server/issues/12367

### DIFF
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -16,7 +16,7 @@
 
 -export([compact_file/2, truncate_file/4, delete_file/2]). %% internal
 
--export([scan_file_for_valid_messages/1]). %% salvage tool
+-export([scan_file_for_valid_messages/1, scan_file_for_valid_messages/2]). %% salvage tool
 
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2,
          code_change/3, prioritise_call/4, prioritise_cast/3,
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2096,28 +2105,28 @@ delete_file(File, State = #gc_state { file_summary_ets = FileSummaryEts,
...
-              end
-      end, {[], 0}, Messages).
+    Path = form_filename(Dir, filenum_to_name(File)),
+    {ok, Messages, _FileSize} = scan_file_for_valid_messages(Path,
+        fun ({MsgId, TotalSize, Offset}) ->
+            case index_lookup(IndexEts, MsgId) of
+                #msg_location { file = File, total_size = TotalSize,
+                                offset = Offset, ref_count = 0 } = Entry ->
+                    index_delete_object(IndexEts, Entry),
+                    invalid;
```
gomoripeti, line 2115: isn't this a valid message from scanning perspective? So we don't want to scan byte-by-byte just because ref count is zero? Maybe need a 3rd status `valid_but_ignore` (I might be wrong)

lhoguin: It is but since we are deleting it in this clause it becomes no longer valid. Maybe I can add a comment.

lhoguin:
This should be better. If not, I will fix tomorrow.

lhoguin:
All good. Just need more testing and if possible some real world testing.

michaelklishin:
This PR can be tested with the following Docker image tag https://hub.docker.com/layers/pivotalrabbitmq/rabbitmq/sha-acb04e929f9b1030dd077c62991147876b08fff5/images/sha256-353eb18c3b4d98fd393eacaf98234a90bfc582f0c6d1f38232f8a4317b0cf4be?context=explore

deps/rabbit/src/rabbit_msg_store.erl
gomoripeti: what I meant is something like the bellow where we jump to `Offset + TotalSize` instead of `Offset + 1` but we don't add the `Entry` to the `Acc` in `scan_data/7`
```
                %% Confirmed to be a message by the provided fun
                %% but we don't want to keep it.
                valid_but_ingore ->
                    scan_data(Rest, Fd, Fun, Offset + TotalSize, FileSize,
                              MsgIdsFound#{MsgIdInt => true}, Acc);
```
(This is just an efficiency optimisation, not a correctness issue)

lhoguin: Oh fair enough, yes that would be better. Maybe it'll even help with file delete performance. Will check on Monday.

lhoguin: Done, and that solves the slow delete. Thanks for noticing!!

gomoripeti: great

lhoguin:
This should be the most up to date image: https://hub.docker.com/layers/pivotalrabbitmq/rabbitmq/loic-fix-cq-scan/images/sha256-ab439bb28985d6699f8b5a7c6ca91ca0d183202521f808b888c874b41146c02e?context=explore

lhoguin:
Rebased and marked ready for review. But I recommend merging this after https://github.com/rabbitmq/rabbitmq-server/pull/12425 just to be sure that everything is green.

michaelklishin:
Two positive reports from two users in https://github.com/rabbitmq/rabbitmq-server/issues/12367:

1. One had an exception before roughly every 18-24 hours ("every night"), not any more after 24 hours
2. Another no longer sees the exception after several days in a few clusters

michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/pull/12425 was merged.

michaelklishin (approved):

Issue #7863, opened 2023-04-06
Quorum queues periodic policy repair


kjnilsson:
### Is your feature request related to a problem? Please describe.

If a quorum queue is unavailable when a policy is changed it may never apply the resulting configuration command and thus be out of sync with the matching policy. To handle this potential scenario we should do periodic checks to see if the active QQ configuration matches the policy it should have and if not issue another configuration change command.

### Describe the solution you'd like

Quorum queues already do periodic repair type of work in `rabbit_quorum_queue:handle_tick/3` so it would make sense to extend this function to also compare selected configuration fields with the currently active policy and process a new config command as is done when a policy changes. (see `rabbit_quorum_queue:policy_changed/1`)

### Describe alternatives you've considered

Currently if this failure is even detected the only solution is to remove and re-add the policy for a given queue.

### Additional context

VESC-1065

gomoripeti:
We will look into implementing this one shortly

PR #12412, opened 2024-10-01
Periodically check for unapplied policies on QQs

LoisSotoLopez:
## Proposed Changes

As documented https://github.com/rabbitmq/rabbitmq-server/issues/7863 :

_If a quorum queue is unavailable when a policy is changed it may never apply the resulting configuration command and thus be out of sync with the matching policy._ 

This PR provides a function in `rabbit_quorum_queue.erl` that checks whether the current Ra Machine configuration for a queue corresponds to the expected configuration to be in use based on defined policies. That function is called by each queue process on tick (`handle_tick`).

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments
.


deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -707,6 +719,21 @@ system_recover(quorum_queues) ->
...
+maybe_apply_policies(Q, #{config := CurrentConfig}) ->
+    NewPolicyConfig = gather_policy_config(Q, _ShoudLog = false),
+
+    RelevantKeys = maps:keys(NewPolicyConfig),
+    CurrentPolicyConfig = maps:with(RelevantKeys, CurrentConfig),
+
+    ShouldUpdate = NewPolicyConfig =/= CurrentPolicyConfig,
+    case ShouldUpdate of
+        true ->
+            rabbit_log:debug("Re-applying policies to ~p", [amqqueue:get_name(Q)]),
```
michaelklishin, line 731: Log messages should use `rabbit_misc:rs/2`.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1528,35 +1555,35 @@ reclaim_memory(Vhost, QueueName) ->
     ra_log_wal:force_roll_over({?RA_WAL_NAME, Node}).
 
 %%----------------------------------------------------------------------------
-dead_letter_handler(Q, Overflow) ->
+dead_letter_handler(Q, Overflow, ShouldLog) ->
```
michaelklishin, line 1558: I strongly dislike this extra argument and how it changes existing functions. If an invalid overflow strategy is used, I don't see a problem with logging that periodically.

michaelklishin (changes requested):
The "maybe log" changes the internal API in ways that are difficult to justify.

deps/rabbit/test/quorum_queue_SUITE.erl
```diff
@@ -1151,6 +1152,166 @@ single_active_consumer_priority(Config) ->
...
+                Srv,
+                rabbit_quorum_queue,
+                recover,
+                [foo, [Queue]]
+            )
+        end,
+        Servers),
+
+    % Wait for the queue to be available again. 
+    lists:foreach(fun(Srv) ->
```
michaelklishin, line 1278: This is reinventing `rabbit_ct_helpers:await_condition/2`.

deps/rabbit/test/quorum_queue_SUITE.erl
```diff
@@ -4153,3 +4314,28 @@ lists_interleave([Item | Items], List)
...
+        case amqp_channel:call(Ch, #'basic.get'{queue = QQ}) of
+            {#'basic.get_ok'{}, Msg} ->
+                C([Msg | Acc]);
+            _ ->
+                Acc
+        end
+    end,
+    Consume([]).
+
+ensure_qq_proc_dead(Config, Server, RaName) ->
```
michaelklishin, line 4331: If the target process recovers in fewer than 500ms, this function will loop forever.

deps/rabbit/src/rabbit_quorum_queue.erl
kjnilsson: yes I agree, I'm not sure what value they add to this PR

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -315,29 +315,27 @@ declare_queue_error(Error, Queue, Leader, ActingUser) ->
...
-    Overflow = overflow(OverflowBin, drop_head, QName),
+    Overflow = overflow(OverflowBin, drop_head, QName, ShouldLog),
     MaxBytes = args_policy_lookup(<<"max-length-bytes">>, fun min/2, Q),
     DeliveryLimit = case args_policy_lookup(<<"delivery-limit">>,
                                             fun resolve_delivery_limit/2, Q) of
                         undefined ->
-                            rabbit_log:info("~ts: delivery_limit not set, defaulting to ~b",
-                                             [rabbit_misc:rs(QName), ?DEFAULT_DELIVERY_LIMIT]),
+                            maybe_log(ShouldLog, info,
+                                      "~ts: delivery_limit not set, defaulting to ~b",
```
gomoripeti, line 329: dear Core Team, should this message be logged unconditionally as well? This is not a misconfiguration, if a user is happy with the default value and does not set an explicit delivery-limit, this will be logged all the time for all the quorum queues.

deps/rabbit/test/quorum_queue_SUITE.erl
gomoripeti: ra supervisor has a max restart intensity of 2 restarts per 5 seconds https://github.com/rabbitmq/ra/blob/main/src/ra_server_sup.erl#L36-L37. So supervisor will give up eventually.
Otoh if the process restart takes more than 500ms then this loop would stop before the process is dead completely. But I think this is highly unlikely for a test queue.

deps/rabbit/src/rabbit_quorum_queue.erl
kjnilsson: It's there to be clear that there is a default set. We can perhaps lower it to debug in 4.1 or remove it completely but I think making users aware of this potentially breaking change doesn't harm.

michaelklishin: We can move this to a function that is not called periodically and remove it from this function.

michaelklishin: In fact, we don't necessarily have to do it in `rabbit_quorum_queue` if there's a more suitable alternative where the delivery limit is known.

LoisSotoLopez:
Is there anything we can do about this PR?

michaelklishin:
@LoisSotoLopez according to `git`, this PR is over 100 commits behind `main`. Please rebase it.

michaelklishin (approved):
To summarize this change: the optional policy re-application happens as part of the periodic QQ replica "tick", also used by the [continuous membership reconciliation](https://www.rabbitmq.com/docs/quorum-queues#replica-reconciliation) feature.

In cases where effective policy keys do not change, the new operation is a no-op.

michaelklishin (changes requested):
```
cd deps/rabbit
gmake ct-quorum_queue
```

as well as

```
bazel test //deps/rabbit:quorum_queue_SUITE
```

fails, specifically one policy-based test does, on CI and locally, with Khepri and with Mnesia, in other words, reliably:

 * `quorum_queue_SUITE > single_node > queue_ttl`
 * `quorum_queue_SUITE > clustered > cluster_size_3 > queue_ttl`
 
Since `maybe_apply_policies/2` is the only new function that
seems relevant, I suggest adding some debug logging there and comparing what it logs vs. `main`.

michaelklishin:
Some details:

```
=== Ended at 2024-10-24 21:06:50
=== Location: [{quorum_queue_SUITE,'-queue_ttl/1-AwaitMatchFilter/1-0-',[3776](https://github.com/rabbitmq/rabbitmq-server/pull/quorum_queue_suite.src.html#3776)},
              {quorum_queue_SUITE,queue_ttl,[3765](https://github.com/rabbitmq/rabbitmq-server/pull/quorum_queue_suite.src.html#3765)},
              {test_server,ts_tc,1794},
              {test_server,run_test_case_eval1,1303},
              {test_server,run_test_case_eval,1235}]
=== === Reason: {awaitMatch,
                     [{module,quorum_queue_SUITE},
                      {line,3776},
                      {expression,
                          "catch amqp_channel : call ( Ch , # 'queue.declare' { queue = QQ , passive = true , durable = true , auto_delete = false , arguments = QArgs } )"},
                      {pattern,
                          "{ 'EXIT' , { { shutdown , { server_initiated_close , 404 , << \"NOT_FOUND - no queue 'queue_ttl' in vhost '/'\" >> } } , _ } }"},
                      {value,{'queue.declare_ok',<<"queue_ttl">>,0,0}}]}
  in function  quorum_queue_SUITE:'-queue_ttl/1-AwaitMatchFilter/1-0-'/3 (quorum_queue_SUITE.erl, line 3776)
  in call from quorum_queue_SUITE:queue_ttl/1 (quorum_queue_SUITE.erl, line 3765)
```

While this could have been be a flake, it is too persistent to conclude that.

gomoripeti:
indeed this also fails for me locally with the same error, thanks for the details. we will look at it on Monday

michaelklishin:
@gomoripeti were you able to collect some more data from the tests/test suite in question?

https://github.com/rabbitmq/rabbitmq-server/pull/12412/commits/2577b7e2843a571392227a2e31c4c971e8b2c7e7 does not explain why the key were removed.

gomoripeti:
The new ` gather_policy_config` sub-function's purpose is to gather configs that can be changed by a policy. The problem was that the keys `single_active_consumer_on` and `created` are not one of those and they are set later in `ra_machine_config`. So they were accidentally set two times. The result was that the config was always reapplied even though real policy config did not change (but `created` changed at every call). Lois knowns the details of why the test case failed, but as far as I understand there was some race condition because of this and the TTL setting was overridden or ignored.

michaelklishin:
In other words, this is ready for another round 👍

gomoripeti:
do I see correctly that the `policy_repair` test case only fails in mixed cluster runs?
What version of RabbitMQ is used for the "old" version in those cases? (is it v4.0.2? https://github.com/rabbitmq/rabbitmq-server/blob/main/bazel/bzlmod/secondary_umbrella.bzl#L34)
Is it possible that sometimes the QQ leader is on an "old" version node, and the code to repair policy is simply not present in that version?

michaelklishin:
@gomoripeti it turned out to be a flake, it seems. The old version is the latest patch of the previous series, so, `3.13.7`.

michaelklishin:
> Is it possible that sometimes the QQ leader is on an "old" version node, and the code to 
> repair policy is simply not present in that version?

This is a very good hypothesis. In which case the test should be skipped for mixed version clusters because it won't be deterministic by definition. [Like here](https://github.com/rabbitmq/rabbitmq-server/blob/033a79336fb2890830a1bea61a2a3de01c6f8381/deps/rabbit/test/definition_import_SUITE.erl#L307-L314), for example.

LoisSotoLopez:
> ... test should be skipped for mixed version clusters because it won't be deterministic by definition

@michaelklishin Maybe you prefer to add that change yourself. In case you don't I PRed it here: https://github.com/rabbitmq/rabbitmq-server/pull/12665 . Feel free to close that one if contributing it yourself eases the process :+1:

michaelklishin:
@LoisSotoLopez your changes seems fine, I have submitted #12666.

PR #12415, opened 2024-10-01
Support AMQP filter expressions

ansd:
## What?

This PR implements the following property filter expressions for AMQP clients
consuming from streams as defined in
[AMQP Filter Expressions Version 1.0 Working Draft 09](https://groups.oasis-open.org/higherlogic/ws/public/document?document_id=66227):
* properties filters [section 4.2.4]
* application-properties filters [section 4.2.5]

String prefix and suffix matching is also supported.

This PR also fixes a bug where RabbitMQ would accept wrong filters.
Specifically, prior to this PR the values of the filter-set's map were
allowed to be symbols. However, "every value MUST be either null or of a
described type which provides the archetype filter."

## Why?

**This feature adds the ability to RabbitMQ to have multiple concurrent clients
each consuming only a subset of messages while maintaining message order.**

**This feature also reduces network traffic between RabbitMQ and clients by
only dispatching those messages that the clients are actually interested in.**

Note that AMQP filter expressions are more fine grained than the [bloom filter based
stream filtering](https://www.rabbitmq.com/blog/2023/10/16/stream-filtering) because
* they do not suffer false positives
* the unit of filtering is per-message instead of per-chunk
* matching can be performed on **multiple** values in the properties and
  application-properties sections
* prefix and suffix matching on the actual values is supported.

Both, AMQP filter expressions and bloom filters can be used together.

## How?

If a filter isn't valid, RabbitMQ ignores the filter. RabbitMQ only
replies with filters it actually supports and validated successfully to
comply with:
"The receiving endpoint sets its desired filter, the sending endpoint
[RabbitMQ] sets the filter actually in place (including any filters defaulted at
the node)."


michaelklishin (approved):

mkuratczyk:
I've played with this feature and everything works nicely.

If anyone is interested, I've added support for application property filters to [`omq`](https://github.com/rabbitmq/omq). For example, the following command publishes 100 msgs/s, each message has one of the three values set on the `key` application property, and the consumer applies a prefix filter, so that it receives only messages with `key` starting with `"b"`. As expected, the consumption rate is two thirds of the publishing rate:
```
$ omq amqp --queues stream -t /queues/stream -T /queues/stream --rate 100 --amqp-app-property key=foo,bar,baz --amqp-app-property-filter key='$p:b'
2024/10/04 13:48:26 INFO consumer started id=1 terminus=/queues/stream
2024/10/04 13:48:26 INFO publisher started id=1 rate=100 destination=/queues/stream
2024/10/04 13:48:27 published=95/s consumed=62/s
2024/10/04 13:48:28 published=100/s consumed=66/s
2024/10/04 13:48:29 published=100/s consumed=65/s
```

The only issue I ran into is that property based filters require keys to be symbols, not strings. This requires updates to go-amqp and hasn't been added to `omq` yet.

acogoluegnes (approved):

PR #12416, opened 2024-10-01
Represent `rabbit_binding:deletions()` with a map instead of dict

the-mikedavis:
This is a mainly cosmetic refactor that makes the types of `rabbit_binding:deletions()` clearer and easier to work with.

The `dict:dict()` typing of `rabbit_binding` appears to be a historical artifact. `dict` has been superseded by `maps`. Switching to a map makes deletions easier to inspect manually and faster. Though if deletions grow so large that the map representation is important, manipulation of the deletions is unlikely to be expensive compared to any other operations that produced them, so performance is probably irrelevant.

This commit refactors the bottom section of the `rabbit_binding` module to switch to a map, switch the `deletions()` type to an opaque, eliminating a TODO created when using Erlang/OTP 17.1, and the deletion value to a record. We eliminate some historical artifacts and "cruft":

* Deletions taking multiple forms needlessly, specifically the shape `{X, deleted | not_deleted, Bindings, none}` no longer being handled. `process_deletions/2` was responsible for creating this shape. Instead we now use a record to clearly define the fields.
* Clauses to catch `{error, not_found}` are unnecessary after minor refactors of the callers. Removing them makes the type specs cleaner.
* `rabbit_binding:process_deletions/1` has no need to update or change the deletions. This function uses `maps:foreach/2` instead and returns `ok` instead of mapped deletions.
* Remove `undefined` from the typespec of deletions. This value is no longer possible with a refactor to `maybe_auto_delete_exchange_in_*` functions for Mnesia and Khepri. The value was nonsensical since you cannot delete bindings for an exchange that does not exist.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -1818,8 +1818,8 @@ internal_delete(Queue, ActingUser, Reason) ->
         {error, timeout} = Err ->
             Err;
         Deletions ->
-            _ = rabbit_binding:process_deletions(Deletions),
-            rabbit_binding:notify_deletions(Deletions, ?INTERNAL_USER),
+            ok = rabbit_binding:process_deletions(Deletions),
+            ok = rabbit_binding:notify_deletions(Deletions, ?INTERNAL_USER),
             rabbit_core_metrics:queue_deleted(QueueName),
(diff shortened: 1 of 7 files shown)


michaelklishin (approved):
The elimination of functions such as `anything_but/3` indeed makes the code easier to follow.

PR #12466, opened 2024-10-07
rabbit_feature_flags: Introduce hard vs. soft required feature flags

dumbbell:
## Why

Before this patch, required feature flags were basically checked during boot: they must have been enabled when they were mere stable feature flags. If they were not, the node refused to boot.

This was easy for the developer because making a feature flag required allowed to remove the entire compatibility code. Very satisfying.

Unfortunately, this was a pain point to end users, especially those who did not pay attention to RabbitMQ and the release notes and were just asking their package manager to update everything. They could end up with a node that refuse to boot. The only solution was to downgrade, enable the disabled stabte feature flags, upgrade again.

## How

This patch introduces two levels of requirement to required feature flags:
* `hard`: this corresponds to the existing behavior where a node will refuse to boot if a hard required feature flag is not enabled before the upgrade.
* `soft`: such a required feature flag will be automatically enabled during the upgrade to a version where it is marked as required.

The level of requirement is set in the feature flag definition:

```erlang
-rabbit_feature_flag(
   {my_feature_flag,
    #{stability     => required,
      require_level => hard
     }}).
```

The default requirement level is `soft`. All existing required feature flags have now a requirement level of `hard`.

The handling of soft required feature flag is done when the cluster feature flags states are verified and synchronized. If a required feature flag is not enabled yet, it is enabled at that time.

This means that as developers, we will have to keep compatibility code forever for every soft required feature flag, like the feature flag definition itself.

### DIFF
--- a/deps/rabbit/src/rabbit_core_ff.erl
+++ b/deps/rabbit/src/rabbit_core_ff.erl
@@ -10,14 +10,16 @@
 -rabbit_feature_flag(
    {classic_mirrored_queue_version,
     #{desc          => "Support setting version for classic mirrored queues",
-      stability     => required
+      stability     => required,
+      require_level => hard
      }}).
 
 -rabbit_feature_flag(
    {quorum_queue,
     #{desc          => "Support queues of type `quorum`",
       doc_url       => "https://www.rabbitmq.com/docs/quorum-queues",
-      stability     => required
+      stability     => required,
+      require_level => hard
      }}).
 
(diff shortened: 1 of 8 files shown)


dumbbell:
The `hard` requirement level is mostly introduced to support existing required feature flags. I expect all future required feature flags will be soft (which is the default).

mkuratczyk (approved):
Thanks!

PR #12427, opened 2024-10-02
Shutdown QQ peer members on force-shrink-to-single-member execution to allow restoring QQ clusters on healthy nodes

Ayanda-D:
## Proposed Changes

Hi 

This is a follow up to #12344 (bouncing off part of our approach to some DR scenarios in case of any unforeseen issues).

We rely on force-shrink commands to rescue QQ nodes in extreme disaster scenarios. Current behavior, when QQs are shrunk to a single node, connected peers keep their local FSMs running, which prevents re-growing the shrunk queues back onto connected nodes which would/could be in a healthy state. We expect peers to update their log to that of the leader on `grow` and/or the next received `#append_entries_rpc{}`.

Example, doing `force_all_queues_shrink_member_to_current_member/0`.

```
$> rabbitmq-queues quorum_status Q1
Status of quorum queue Q1 on node rabbit@host ...
┌─────────────────────┬────────────┬───────────┬──────────────┬────────────────┬──────┬─────────────────┐
│ Node Name           │ Raft State │ Log Index │ Commit Index │ Snapshot Index │ Term │ Machine Version │
├─────────────────────┼────────────┼───────────┼──────────────┼────────────────┼──────┼─────────────────┤
│ rabbit_2@host │ follower   │ 233       │ 233          │ undefined      │ 16   │ 3               │
├─────────────────────┼────────────┼───────────┼──────────────┼────────────────┼──────┼─────────────────┤
│ rabbit_1@host │ follower   │ 233       │ 233          │ undefined      │ 16   │ 3               │
├─────────────────────┼────────────┼───────────┼──────────────┼────────────────┼──────┼─────────────────┤
│ rabbit@host   │ leader     │ 233       │ 233          │ undefined      │ 16   │ 3               │
└─────────────────────┴────────────┴───────────┴──────────────┴────────────────┴──────┴─────────────────┘


$> rabbitmqctl eval 'rabbit_quorum_queue:force_all_queues_shrink_member_to_current_member().'
ok


$> rabbitmq-queues quorum_status Q1
Status of quorum queue Q1 on node rabbit@host ...
┌───────────────────┬────────────┬───────────┬──────────────┬────────────────┬──────┬─────────────────┐
│ Node Name         │ Raft State │ Log Index │ Commit Index │ Snapshot Index │ Term │ Machine Version │
├───────────────────┼────────────┼───────────┼──────────────┼────────────────┼──────┼─────────────────┤
[...]

PR #12467, opened 2024-10-07
By @Ayanda-D: stop QQ replicas when a QQ is forced to shrink to a single replica

michaelklishin:
This is #12427 by @Ayanda-D.

Shrinking operations did not stop QQ replicas. This was easy to miss because QQs are usually shrunk before a node is removed from the cluster.

However, there is a scenario where this is not the case. If some nodes (replicas) need to be replaced, in particular when a majority of nodes cannot be recovered for any reasons, the recovery process will involve shrinking a QQ to just one member so that it has an online quorum (of 1 node out of 1) before new replicas can be added.

For this to succeed, the older replicas must be stopped and deleted from the (QQ) cluster.


michaelklishin:
This will require a manual backport.

michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/pull/12468

PR #12468, opened 2024-10-07
4.1, by @Ayanda-D: stop QQ replicas when a QQ is forced to shrink to a single replica 

michaelklishin:
This is #12427 by @Ayanda-D.

Shrinking operations did not stop QQ replicas. This was easy to miss because QQs are usually shrunk before a node is removed from the cluster.

However, there is a scenario where this is not the case. If some nodes (replicas) need to be replaced, in particular when a majority of nodes cannot be recovered for any reasons, the recovery process will involve shrinking a QQ to just one member so that it has an online quorum (of 1 node out of 1) before new replicas can be added.

For this to succeed, the older replicas must be stopped and deleted from the (QQ) cluster.

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -74,6 +74,7 @@
 -export([validate_policy/1, merge_policy_value/3]).
 
 -export([force_shrink_member_to_current_member/2,
+         force_vhost_queues_shrink_member_to_current_member/1,
          force_all_queues_shrink_member_to_current_member/0]).
 
 %% for backwards compatibility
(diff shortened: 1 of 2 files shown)


michaelklishin:
@Ayanda-D the failure on Erlang 27 in `ct_logs` have nothing to do with this change. Nonetheless, this PR may have to wait until more CI-related changes land into `main`.

michaelklishin:
@mergifyio backport v4.0.x v3.13.x

michaelklishin:
The failures seem to be unrelated to the changes here.

See https://github.com/erlang/otp/issues/8908 and https://github.com/erlang/otp/issues/8911.

Ayanda-D:
@michaelklishin Thank you! 🙏

Ayanda-D:
cc: @willhoy @ahsanabbas123 @25schmekles @mlongob @thisisnsh

michaelklishin:
@mergifyio backport v4.0.x

michaelklishin:
FTR, this was backported to `v4.0.x` manually.

[`v3.13.x` is now only covered by commercial support](https://www.rabbitmq.com/release-information) so if we backport this change to `v3.13.x`, it will ship in the commercial edition (Tanzu RabbitMQ).

PR #12480, opened 2024-10-08
4.0.x: update systemd to 0.6.2

michaelklishin:
The change is by @SimonUnge.


michaelklishin:
`systemd` is not listed in `rabbitmq-components.mk` in `v4.0.x`, instead, it is set in [`deps/rabbit/Makefile`](https://github.com/rabbitmq/rabbitmq-server/blob/v4.0.x/deps/rabbit/Makefile#L138), which makes some sense since it is a core server dependency.

michaelklishin:
@SimonUnge FYI.

PR #12520, opened 2024-10-15
AMQP filter expressions: handle invalid filters

mkuratczyk:
An invalid application property filter could cause a function_clause that was returned to the client, eg `omq` output looked like this:
```
2024/10/15 13:10:55 ERRO consumer failed to create a receiver id=1
  error=
  │ *Error{Condition: amqp:internal-error, Description: Session error: function_clause
  │ [{rabbit_amqp_filtex,'-validate0/2-fun-0-',
  │                      [{{symbol,<<"subject">>},{utf8,<<"var">>}}],
  │                      [{file,"rabbit_amqp_filtex.erl"},{line,119}]},
  │  {lists,map,2,[{file,"lists.erl"},{line,2077}]},
  │  {rabbit_amqp_filtex,validate0,2,[{file,"rabbit_amqp_filtex.erl"},{line,119}]},
  │  {rabbit_amqp_filtex,validate,1,[{file,"rabbit_amqp_filtex.erl"},{line,28}]},
  │  {rabbit_amqp_session,parse_filters,2,
  │                       [{file,"rabbit_amqp_session.erl"},{line,3068}]},
  │  {rabbit_amqp_session,parse_filter,1,
  │                       [{file,"rabbit_amqp_session.erl"},{line,3014}]},
  │  {rabbit_amqp_session,'-handle_attach/2-fun-0-',21,
  │                       [{file,"rabbit_amqp_session.erl"},{line,1371}]},
  │
  {rabbit_misc,with_exit_handler,2,[{file,"rabbit_misc.erl"},{line,465}]}],
  Info: map[]}
```

Now such filters are ignored.

### DIFF
--- a/deps/rabbit/src/rabbit_amqp_filtex.erl
+++ b/deps/rabbit/src/rabbit_amqp_filtex.erl
@@ -119,9 +119,16 @@ validate0(Descriptor, KVList0) when
     KVList = lists:map(fun({{utf8, Key}, {utf8, String}}) ->
                                {Key, parse_string_modifier_prefix(String)};
                           ({{utf8, Key}, TaggedVal}) ->
-                               {Key, unwrap(TaggedVal)}
+                               {Key, unwrap(TaggedVal)};
+                          (_) ->
+                               invalid_filter
                        end, KVList0),
-    {ok, {application_properties, KVList}};
+    case lists:member(invalid_filter, KVList) of
+        false ->
+            {ok, {application_properties, KVList}};
+        true ->
+            error
+    end;
 validate0(_, _) ->
     error.
 
--- a/deps/rabbit/test/amqp_filtex_SUITE.erl
+++ b/deps/rabbit/test/amqp_filtex_SUITE.erl
@@ -32,7 +32,8 @@
 
 all() ->
     [
-     {group, cluster_size_1}
+     {group, cluster_size_1},
+     {group, filter_validation}
     ].
 
 groups() ->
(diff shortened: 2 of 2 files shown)


deps/rabbit/src/rabbit_amqp_filtex.erl
```diff
@@ -119,9 +119,11 @@ validate0(Descriptor, KVList0) when
     KVList = lists:map(fun({{utf8, Key}, {utf8, String}}) ->
                                {Key, parse_string_modifier_prefix(String)};
                           ({{utf8, Key}, TaggedVal}) ->
-                               {Key, unwrap(TaggedVal)}
+                               {Key, unwrap(TaggedVal)};
+                          (_) ->
+                               invalid_filter
```
ansd, line 124: Your modification violates the spec when sending the `filter` field from RabbitMQ to the consumer:
> the sending endpoint sets the filter actually in place

In your change, the desired filter as sent from the client to RabbitMQ is echoed back from RabbitMQ to the client while RabbitMQ silently modifies the filter.
This function should return `error` if the filter is invalid such that the entire filter doesn't take effect. 
For example, the Java client compares only the filter names that were desired (i.e. sent from consumer to RabbitMQ) with the filter names actually in place (i.e. sent from RabbitMQ to the consumer).

ansd (changes requested):

ansd:
Once https://github.com/rabbitmq/rabbitmq-server/pull/12541 is merged, we can easily add an integration test since it will allow the Erlang client to be configured to notify about the `attach` reply from RabbitMQ such that we can check the list of filters returned from RabbitMQ within the test case.

ansd:
Closed in favour of https://github.com/rabbitmq/rabbitmq-server/pull/12549

PR #12521, opened 2024-10-15
QQ: improve `rabbit_quorum_queue:status/2` efficiency by allowing it to use `ra:key_metrics/1`

kjnilsson:
Currently the `rabbit_quorum_queue:status/2` function always falls back to the (slower) compatibility code and never gets the benefit of using `ra:key_metrics/1` due to incorrect use of the map update operatior `:=` instead of the insert operator `=>`.


michaelklishin:
The failure is known flake in Common Test with parallel runs. There are no failing tests in the results.

PR #12526, opened 2024-10-16
Make CI: Use ubuntu-22.04 for the time being

lhoguin:
There are .NET related issues in 24 that we don't
want to look at right now.


lhoguin:
Not necessary.

Issue #12531, opened 2024-10-16
AMQP 1.0 Add Non-Normative Connection Properties `topic-prefix` and `queue-prefix`


alaendle:
### Is your feature request related to a problem? Please describe.

Client code should be independent from broker internals - therefore the prefixes should be encapsulated by the protocol. Unfortunately there is no standardisation of the connection properties (http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-open).
But most clients respect the mentioned prefixes (e.g. Apache.NMS.AMQP (https://github.com/apache/activemq-nms-amqp/blob/5e4e137b071101463b8083f8a8a1cddada5b6f8a/src/NMS.AMQP/Provider/Amqp/AmqpConnection.cs#L161)) and many brokers provide this information (e.g. Solace (https://docs.solace.com/API/AMQP/AMQP-Protocol-Conformance.htm#section-2-7-1-open) & ActiveMQ (https://github.com/apache/activemq/blob/4ff78ddab993d62408c8e63ce63a63d5757fe039/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/protocol/AmqpConnection.java#L209))

### Describe the solution you'd like

I think it would be a good idea to include the prefix information into the transmitted connection properties:

Just to make it clear - here a response from an RabbitMQ connection:
![grafik](https://github.com/user-attachments/assets/733c173a-e673-4a9b-8c3d-7593a3ae4307)

And here Solace:
![grafik](https://github.com/user-attachments/assets/c428f36a-292d-492e-b321-fc6d63f64717)

So basically I request to add `topic-prefix` and `queue-prefix`.

### Describe alternatives you've considered

At least there seems to be an easy workaround - as the server properties are extensible via an environment variable:
```
RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="-rabbit server_properties [{'topic-prefix',\"/topic/\"},{'queue-prefix',\"/queue/\"}]"
```

### Additional context

Please let me know if I was unclear or if I could contribute further information!

ansd:
Thank you @alaendle , this is a great suggestion.

> But most clients respect the mentioned prefixes

Which other clients handle these properties?

alaendle:
[...]

PR #12541, opened 2024-10-17
Return queue-prefix in open properties

ansd:
Closes https://github.com/rabbitmq/rabbitmq-server/issues/12531

See commit messages for more details.

### DIFF
--- a/deps/rabbit/src/rabbit_amqp_reader.erl
+++ b/deps/rabbit/src/rabbit_amqp_reader.erl
@@ -136,7 +136,15 @@ system_code_change(Misc, _Module, _OldVsn, _Extra) ->
 server_properties() ->
     Props0 = rabbit_reader:server_properties(amqp_1_0),
     Props1 = [{{symbol, K}, {utf8, V}} || {K, longstr, V} <- Props0],
-    Props = [{{symbol, <<"node">>}, {utf8, atom_to_binary(node())}} | Props1],
+    Props = [
+             {{symbol, <<"node">>}, {utf8, atom_to_binary(node())}},
+             %% queue-prefix is not part of the AMQP spec.
+             %% However, we include this property because it is also returned by
+             %% ActiveMQ and Solace and understood by some client libs,
+             %% e.g. ActiveMQ NMS.AMQP and Qpid JMS.
+             %% https://github.com/rabbitmq/rabbitmq-server/issues/12531
+             {{symbol, <<"queue-prefix">>}, {utf8, <<"/queues/">>}}
+            ] ++ Props1,
     {map, Props}.
 
 %%--------------------------------------------------------------------------
--- a/deps/amqp10_client/src/amqp10_client.erl
+++ b/deps/amqp10_client/src/amqp10_client.erl
@@ -144,6 +144,8 @@ begin_session_sync(Connection, Timeout) when is_pid(Connection) ->
             receive
                 {amqp10_event, {session, Session, begun}} ->
                     {ok, Session};
+                {amqp10_event, {session, Session, {begun, #'v1_0.begin'{}}}} ->
+                    {ok, Session};
                 {amqp10_event, {session, Session, {ended, Err}}} ->
                     {error, Err}
             after Timeout -> session_timeout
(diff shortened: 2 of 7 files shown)


deps/rabbit/src/rabbit_amqp_reader.erl
```diff
@@ -136,7 +136,15 @@ system_code_change(Misc, _Module, _OldVsn, _Extra) ->
...
     Props1 = [{{symbol, K}, {utf8, V}} || {K, longstr, V} <- Props0],
-    Props = [{{symbol, <<"node">>}, {utf8, atom_to_binary(node())}} | Props1],
+    Props = [
+             {{symbol, <<"node">>}, {utf8, atom_to_binary(node())}},
+             %% queue-prefix and topic-prefix are not part of the AMQP spec.
+             %% However, we include these properties because they are also returned by
+             %% ActiveMQ and Solace and understood by some client libs (e.g. ActiveMQ NMS.AMQP Client).
+             %% https://github.com/rabbitmq/rabbitmq-server/issues/12531
+             {{symbol, <<"queue-prefix">>}, {utf8, <<"/queues/">>}},
+             {{symbol, <<"topic-prefix">>}, {utf8, <<"/exchanges/amq.topic/">>}}
```
alaendle, line 146: @ansd - I've tested and for me it seems that there is a difference in behaviour if this is `/exchanges/amq.topic/` or `/topic/`. If I just subscribe to `test.*` topic the call succeeds with `/topic/`, but it fails if the prefix is `/exchanges/amq.topic/` - `Attach rejected: {unknown_destination,"/exchanges/amq.topic/test.*"}`

ansd: Thanks for testing @alaendle.

Going forward RabbitMQ will support only address format v2: https://www.rabbitmq.com/docs/amqp#addresses
Therefore, including `/topic/` is not an option.

So far my understanding has been that the client libs use these prefixes only for **publishing** messages, i.e. as prefixes for **target** addresses. If these client libs also use these prefixes for **consuming** messages, i.e as **source** addresses, then RabbitMQ should advertise only the `/queues/` prefix.

ansd: @alaendle where in the client libs to do you see these prefixes being used for **consuming** messages, i.e. as **source** address prefix?

ansd: I changed the PR to include only the queue-prefix to be on the safe side.

alaendle: You are right! Because it started to work again if I remove the `s` from `exchanges` - means switching back to address format v1 😄. Unfortunately the translation need to work in both directions - so you're implementation is correct, but only supports v2 addresses. Only advertising the `queue-prefix` isn't a good solution - simple because of it's inconsistency - this would tell the user code to manage topic prefixes but not queue prefixes. So be honest I have to think about this a few minutes... if I use a queue for consumption (what I SHOULD do according to the documentation) everything works as expected; so two downsides of this PR - I could potentially break code that currently uses v1-prefixes like `/topic/` for consumption - because prefixes won't match and the client lib would build a address like `/exchanges/amq.topic//topic/test.*`; it would work if the client already follows v2 pattern on `amq.topic` - because if it is already named `/exchanges/amq.topic/test.cell` the prefix matches and the resulting address is identical. However code might break in another scenario - what if the client explicitly addresses another exchange like getting a topic for `/exchanges/amq.direct/test`? This would lead to: `Attach rejected: {bad_address_string,<<"/exchanges/amq.topic//exchanges/amq.direct/test">>`. I'm still pretty sure that it is worth to expose the properties and that this is exactly the behaviour all would like in 99% percent of all cases - however there are two caveats, a) code using v1-addresses could break, b) code using v2-addresses with exchanges other than `amq.topic` could break.

alaendle: > I changed the PR to include only the queue-prefix to be on the safe side.

You are too fast for me ⚡

ansd:
```
make -C deps/rabbit ct-amqp_system t=java:roundtrip
```
fails with this PR because the JMS client starts using `/queues/my-queue` although the client does not previously declare the queue, which may be dangerous for existing client apps upgrading to RabbitMQ 4.1.

Also note that setting the `topic-prefix` only works for **target** address `/exchanges/amq.topic/` but not for **source** address.

We can re-evaluate whether RabbitMQ returning a queue-prefix `/queues/` makes sense in the future when address format v1 is completely removed. For the time being RabbitMQ returning a queue-prefix in the open properties introduces more problems than it solves.

The 1st commit of this PR got extracted into https://github.com/rabbitmq/rabbitmq-server/pull/12547 since it will help writing test cases for other features such as https://github.com/rabbitmq/rabbitmq-server/pull/12520

PR #12551, opened 2024-10-18
Try running test suites on Ubuntu 24.04

michaelklishin:
to avoid slapd(8) segfaulting on Ubuntu 22.04,
which "ubuntu-latest" uses despite [1].

Per discussion with @lukebakken.

1. https://github.com/actions/runner-images/issues/9848

### DIFF
--- a/.github/workflows/test-make-target.yaml
+++ b/.github/workflows/test-make-target.yaml
@@ -20,7 +20,7 @@ on:
 jobs:
   test:
     name: ${{ inputs.plugin }} (${{ inputs.make_target }})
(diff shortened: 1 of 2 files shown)


michaelklishin:
The problem persists on 24.04 with `slapd` 2.6.x.

A relevant finding: https://lists.openldap.org/hyperkitty/list/openldap-software@openldap.org/thread/DCHKVCJKK5OUFFTMHRJ4D7TNYDI5OX6Z/#DCHKVCJKK5OUFFTMHRJ4D7TNYDI5OX6Z

PR #12557, opened 2024-10-21
Use OS trusted CA certs if not provided by the user

LoisSotoLopez:
## Proposed Changes
Provides a specific function to fix client ssl options, i.e.: apply all fixes that are applied for TLS listeneres and clients on previous versions but also sets `cacerts` option to CA certificates obtained by `public_key:cacerts_get/0`, only when no `cacertfile` or `cacerts` are provided.

Addressing #10519 

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

## Further Comments

Should we also apply this fix for TLS server options? Or is it fine if we only use OS trusted CA certs for TLS client options, as done in the provided changes?

PR #12502, opened 2024-10-10
4.1/main: Make CI: Fix and enhance ct_master

lhoguin:
Most of this will be sent as a PR to OTP.

PR #12564, opened 2024-10-21
#12557: fall back to system-wide CA certificates (if available) when none are configured for AMQP 1.0 and AMQP 0-9-1 clients such as shovels

michaelklishin:
This is #12557 by @LoisSotoLopez.


michaelklishin:
The forced push was a rebase to make sure that #12502 is included (it addresses one specific kind of annoying CT suite flakes).

PR #12566, opened 2024-10-22
Bump observer_cli to 1.8.0

zhongwencool:
🎉 https://github.com/zhongwencool/observer_cli/releases/tag/1.8.0 
Now we can  see the number of bytes in the output distribution queue in **System** view
iShot_2024-10-21_17 13 10


michaelklishin (approved):

michaelklishin:
@zhongwencool thank you!

michaelklishin:
@zhongwencool I'm afraid I'll have to revert this. This version relies on a function that does not exist in Erlang 26, as this `xref` failure https://github.com/rabbitmq/rabbitmq-server/actions/runs/11451731575/job/31861435010#step:8:329 correctly points out.

zhongwencool:
@michaelklishin This seems a bit strange, and I added xref.ignore in observer_cli's rebar.conf 
 but rabbitmq-server didn’t use this xref configuration.
https://github.com/zhongwencool/observer_cli/blob/master/rebar.config#L43C2-L44
the Ci on 26 is also passed.
https://github.com/zhongwencool/observer_cli/actions/runs/11437029547
the code is ok on otp without proc_lib:get_label/1

https://github.com/zhongwencool/observer_cli/blob/master/src/observer_cli.erl#L785-L790
```erlang
erlang:function_exported(proc_lib, get_label, 1) andalso
            proc_lib:get_label(Pid)
```

michaelklishin:
@zhongwencool our team does not use rebar. It can be a peculiarity of how our `erlang.mk`-based build system drives `xref`.

I suspect that an explicit exclude (in the code) should work. We used to not apply `xref` for dependencies, that's another option but we'd prefer for everything to pass `xref` and `dialyze` with dependencies included.

zhongwencool:
ok，I move the xref_ignore into module. so erlang.mk and rebar3 should support now. 

https://github.com/zhongwencool/observer_cli/pull/106

I will update this when observer_cli 1.8.1 is ready :0

michaelklishin:
@zhongwencool sounds good. Thanks again for your help!

PR #12585, opened 2024-10-25
Re-submitted #12412

michaelklishin:
Re-submitting #12412 so that it has access to GH secrets.


michaelklishin:
This required one more conflict resolution after https://github.com/rabbitmq/rabbitmq-server/commit/b1169d06ba7dc22059cddc3dfb645ca620e3c132, so will have to be backported to `v4.0.x` manually.

michaelklishin:
A relevant test repeatedly fails in every environment I try, so this goes back to the contributors for now, see #12412.

PR #12617, opened 2024-10-30
Use log macros for AMQP

ansd:
Using a log macro has the benefit that location data is added as explained in https://www.erlang.org/doc/apps/kernel/logger.html#t:metadata/0

This PR is based on top of https://github.com/rabbitmq/rabbitmq-website/pull/2113

### DIFF
--- a/deps/rabbit/src/rabbit_access_control.erl
+++ b/deps/rabbit/src/rabbit_access_control.erl
@@ -249,7 +249,7 @@ check_user_id0(ClaimedUserName, #user{username = ActualUserName,
     end.
 
 -spec update_state(User :: rabbit_types:user(), NewState :: term()) ->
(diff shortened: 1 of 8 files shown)


michaelklishin:
You likely mean https://github.com/rabbitmq/rabbitmq-server/pull/12599 and not its corresponding documentation PR :)

michaelklishin:
FTR, as discussed internally, this is primarily useful for structured logging. Text file logging may benefit from it if we choose to change the format one day but structured (e.g. [JSON logging](https://www.rabbitmq.com/docs/logging#json)) stands to gain the most.

ansd:
> You likely mean https://github.com/rabbitmq/rabbitmq-server/pull/12599 and not its corresponding documentation PR :)

Yes, thanks :)

michaelklishin (approved):

PR #12643, opened 2024-11-04
rabbit_feature_flags: Rework the management UI page

dumbbell:
## Why

The "Feature flags" admin section had several issues:
* It was not designed for experimental feature flags. What was done for RabbitMQ 4.0.0 was still unclear as to what a user should expect for experimental feature flags.
* The UI uses synchronous requests from the browser main thread. It means that for a feature flag that has a long running migration callback, the browser tab could freeze for a very long time.

## How

The feature flags table is reworked and now displays:
* a series of icons to highlight the following:
    * a feature flag that has a migration function and thus that can take time to be enabled
    * a feature flag that is experimental
    * whether this experimental feature flag is supported or not
* a toggle to quickly show if a feature flag is enabled or not and let the user enable it at the same time.

Here is how the new table looks like:

Screenshot 2024-10-31 at 17-45-21 RabbitMQ Management

For stable feature flags, when a user click on the toggle, the toggle goes into an intermediate state while waiting for the response from the broker. If the response is successful, the toggle is green. Otherwise it goes back to red and the error is displayed in a popup as before.

For experimental feature flags, when a user click on the toggle, a popup is displayed to let the user know of the possible constraints and consequences, with one or two required checkboxes to tick so the user confirms they understand the message. The feature flag is enabled only after the user validates the popup. The displayed message and the checkboxes depend on if the experimental feature flag is supported or not (it is a new attribute of experimental feature flags).

Here is a screenshot of the confirmation popup for unsupported experimental feature flags:

Screenshot 2024-10-31 at 17-45-39 RabbitMQ Management

The request to enable feature flags now uses the modern `fetch()` API. Therefore it uses Javascript promises and does not block the main thread: the UI remains responsive while a migration callback runs.

Finally, an "Enable all stable feature flags" button has been added to the warning that tells the user some stable feature flags are still disabled.

Here is a screenshot of the warning with the added button:

image

### DIFF
--- a/deps/rabbit/src/rabbit_feature_flags.erl
+++ b/deps/rabbit/src/rabbit_feature_flags.erl
@@ -348,6 +348,8 @@
               feature_state/0,
               feature_states/0,
               stability/0,
+              require_level/0,
+              experiment_level/0,
               callback_fun_name/0,
               callbacks/0,
               callback_name/0,
--- a/deps/rabbit/src/rabbit_feature_flags.erl
+++ b/deps/rabbit/src/rabbit_feature_flags.erl
@@ -719,13 +719,17 @@ info() ->
 info(Options) when is_map(Options) ->
     rabbit_ff_extra:info(Options).
 
--spec get_state(feature_name()) -> enabled | disabled | unavailable.
+-spec get_state(feature_name()) -> enabled |
+                                   state_changing |
+                                   disabled |
+                                   unavailable.
 %% @doc
 %% Returns the state of a feature flag.
 %%
 %% The possible states are:
 %% <ul>
 %% <li>`enabled': the feature flag is enabled.</li>
+%% <li>`state_changing': the feature flag is being enabled.</li>
 %% <li>`disabled': the feature flag is supported by all nodes in the
 %%   cluster but currently disabled.</li>
 %% <li>`unavailable': the feature flag is unsupported by at least one
@@ -733,16 +737,20 @@ info(Options) when is_map(Options) ->
 %% </ul>
 %%
 %% @param FeatureName The name of the feature flag to check.
-%% @returns `enabled', `disabled' or `unavailable'.
+%% @returns `enabled', `state_changing', `disabled' or `unavailable'.
 
 get_state(FeatureName) when is_atom(FeatureName) ->
-    IsEnabled = is_enabled(FeatureName),
+    IsEnabled = is_enabled(FeatureName, non_blocking),
     case IsEnabled of
-        true  -> enabled;
-        false -> case is_supported(FeatureName) of
-                     true  -> disabled;
-                     false -> unavailable
-                 end
+        true  ->
+            enabled;
+        state_changing ->
+            state_changing;
+        false ->
+            case is_supported(FeatureName) of
+                true  -> disabled;
+                false -> unavailable
+            end
     end.
 
 -spec get_stability
--- a/deps/rabbit/src/rabbit_feature_flags.erl
+++ b/deps/rabbit/src/rabbit_feature_flags.erl
@@ -106,6 +106,7 @@
          get_state/1,
          get_stability/1,
          get_require_level/1,
+         get_experiment_level/1,
          check_node_compatibility/1, check_node_compatibility/2,
          sync_feature_flags_with_cluster/2,
          refresh_feature_flags_after_app_load/0,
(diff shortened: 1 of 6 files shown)


dumbbell:
As of this writing, the branch contains a commit that adds three feature flags just to help test the branch. They will be removed before the merge.

These feature flags are:
* `a_stable_ff`, a stable feature flag :-)
* `exp_sup_ff`, a supported experimental feature flag
* `exp_unsup_ff`, an unsupported experimental feature flag

They all use an `enable` callback that throws an exception after sleeping two seconds.

michaelklishin:
This looks great.

deps/rabbit/src/rabbit_feature_flags.erl
```diff
@@ -219,6 +222,24 @@
...
+%% The level of support of an experimental feature flag.
+%%
+%% At first, an experimental feature flag is offered to give a chance to users
+%% to try it and give feedback as part of the design and development of the
+%% feature. At this stage, it is unsupported: it must not be enabled in a
+%% production environment and upgrade to a later version of RabbitMQ while
+%% this experimental feature flag is enabled is not supported.
+%%
+%% Then, the experimental feature flag becomes supported. At this point, it is
+%% stable enough that upgrading is guarantied and help will be provided.
```
michaelklishin, line 235: "guarantied" is a word that does exist but you likely mean "guaranteed" here :)

deps/rabbit/src/rabbit_feature_flags.erl
```diff
@@ -809,6 +838,45 @@ get_require_level(FeatureProps) when ?IS_DEPRECATION(FeatureProps) ->
...
+      ExperimentLevel :: experiment_level() | none.
+%% @doc
+%% Returns the experimental level of an experimental feature flag.
+%%
+%% The possible experiment levels are:
+%% <ul>
+%% <li>`unsupported': the experimental feature flag must not be enabled in
+%%   production and upgrades with it enabled is unsupported.</li>
+%% <li>`supported': the experimental feature flag is not yet stable enough but
+%%   upgrades are guarantied to be possible. This is returned too if the
```
michaelklishin, line 858: "guarantied" is a word that does exist but you likely mean "guaranteed" here :)

deps/rabbitmq_management/priv/www/js/tmpl/feature-flags.ejs
```diff
@@ -1,145 +1,303 @@
...
+           <% } %>
+           <% if (feature_flag.stability == 'experimental') { %>
+           <svg width="800px" height="800px" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg" class="experimental_icon">
+               <title>This is an experimental feature flag</title>
+               <path d="M10 4V10L5.20285 16.8531C4.27496 18.1786 5.22327 20 6.84131 20H17.1587C18.7767 20 19.725 18.1786 18.7972 16.8531L14 10V4M10 4H14M10 4H8M14 4H16" stroke="#000000" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+           </svg>
            <% } %>
+           <% if (feature_flag.experiment_level == 'unsupported') { %>
+           <svg width="800px" height="800px" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+               <title>This experimental feature flag is not yet supported at this stage and an upgrade path is not guarantied</title>
```
michaelklishin, line 286: "guarantied" is a word that does exist but you likely mean "guaranteed" here :)

deps/rabbit/src/rabbit_feature_flags.erl
dumbbell: Thank you! I never know how to write it…

mkuratczyk:
There are two issues related to auto-refresh:
1. (discussed yesterday with @dumbbell) When using the "enable all FFs button", if there's an FF migration that takes more than the page refresh, feature flags that will be enabled later (are "queued" to be enabled) will turn red (because their state is not changing - they are still disabled)
2. when the experimental feature flag dialog appears with the warning, it disappears on page refresh

Both issues are relatively minor and the overall experience is already better. I don't insist we have to fix these.

dumbbell:
@mkuratczyk: I finally found and implemented a fix for the issues you list where the auto-refresh is involved: I pause the auto-refresh timer and resume it at the end of handling feature flags.

I think there are still issues in this area. For example, if the user switch to another page and comes back. Or if they play with the auto-refresh dropdown menu at the top right. But I feel fixing this becomes out of scope for this patch. We can wait that a user reports them :)

mkuratczyk (approved):
Works great now

PR #12646, opened 2024-11-04
metrics_SUITE: wait for tables in proper test

dcorbacho:
More attempts to fix CI flakes


michaelklishin:
The failure in

```

Test (make) / Test mixed clusters (26, 1.17, khepri) / Test rabbit (ct-quorum_queue)
```

is present in `main` and not related.

PR #12659, opened 2024-11-04
Make it possible to set some cluster metadata besides the name using tags

### LINKED ISSUE #12552: Introduce a way to tag clusters (and, optionally, nodes)
Per discussion with @SimonUnge but also @stefanmoser @mkuratczyk.

## Problem Definition

Larger users can have thousands of clusters used for all kinds of purposes. Sometimes it may be
necessary to perform certain operations (e.g. upgrades) on a subset of them, e.g. those that are
used by development environments only.

Right now there only so many ways of doing it:

1. Keep this cluster metadata in an external store. This is not always an option
2. Prefix cluster name with `production-*` or something like that. This is fragile and allows for only a couple of tags

## Cluster Tags

One solution can be configuring cluster metadata with a map (a set of pairs) in `rabbitmq.con`:

```ini
cluster_tags.environment = production

cluster_tags.region = us-east
cluster_tags.az = us-east-3
```

This map can be then added to `GET /api/overview` HTTP API responses. In Prometheus responses, this could look something like this (**please read the comments**, with Prometheus it is much more nuanced than it sounds):

```
rabbitmq_identity_cluster_tags{environment = "production", region = "us-east", az = "us-east-3"}
```

This data can be stored in the node's application environment or in a global runtime parameter
(for "last write wins" consistency).

## Node Tags

The same idea can be extended to node tags that will only be stored in the application
environment since this data is local to the node. 

```ini
node_tags.environment = production

node_tags.region = us-east
node_tags.az = us-east-3
```

Which could look approximately like this in Prometheus output (**please read the comments**, with Prometheus it is much more nuanced than it sounds):

```
rabbitmq_identity_node_tags{environment = "production", region = "us-east", az = "us-east-3"}
```

The usefulness of this is less clear.

SimonUnge:
## Proposed Changes

Implements #12552.
Closes #12552.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -2688,6 +2688,20 @@ fun(Conf) ->
     end
 end}.
 
+{mapping, "cluster_tags.$tag", "rabbit.cluster_tags", [
+    {datatype, [binary]}
+]}.
+
+{translation, "rabbit.cluster_tags",
+fun(Conf) ->
+    case cuttlefish:conf_get("cluster_tags", Conf, undefined) of
+        none -> [];
+        _ ->
+            Settings = cuttlefish_variable:filter_by_prefix("cluster_tags", Conf),
+            [ {list_to_binary(K), V} || {[_, K], V} <- Settings]
+    end
+end}.
+
 % ===============================
 % Validators
 % ===============================
--- a/deps/rabbit/src/rabbit.erl
+++ b/deps/rabbit/src/rabbit.erl
@@ -11,8 +11,6 @@
 -include_lib("kernel/include/logger.hrl").
 -include_lib("rabbit_common/include/logging.hrl").
 
--feature(maybe_expr, enable).
-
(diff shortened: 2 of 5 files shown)


deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -2688,6 +2688,20 @@ fun(Conf) ->
     end
 end}.
 
+{mapping, "cluster_tags.$tag", "rabbit.cluster_tags", [
+    {datatype, [string]}
```
michaelklishin, line 2692: You can change this to `binary` as of https://github.com/Kyorai/cuttlefish/commit/d367bb702471f9e13b1a06f5a5d47a3e5d766eb7 (already used by RabbitMQ) and avoid a conversion to binary in other places.

SimonUnge: good point, will do.

PR #12665, opened 2024-11-05
Exclude policy_repair QQ test on mixed versions

LoisSotoLopez:
## Proposed Changes

As suggested here https://github.com/rabbitmq/rabbitmq-server/pull/12412#issuecomment-2457254782 the recently merged QQ policy repair mechanism included some tests that are not passing in the CI for mixed versions clusters. This PR prevents that test execution for the mixed versions cluster tests.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #NNNN)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

## Further Comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution
you did and what alternatives you considered, etc.

PR #12666, opened 2024-11-05
By @LoisSotoLopez: Exclude policy_repair QQ test on mixed versions

michaelklishin:
This is #12665 by @LoisSotoLopez. References #12412.

### DIFF
--- a/deps/rabbit/test/quorum_queue_SUITE.erl
+++ b/deps/rabbit/test/quorum_queue_SUITE.erl
@@ -1307,154 +1307,159 @@ force_vhost_queues_shrink_member_to_current_member(Config) ->
 % that affects such queue, when the process is made available again, the policy
 % will eventually get applied. (https://github.com/rabbitmq/rabbitmq-server/issues/7863)
 policy_repair(Config) ->
-    [Server0, _Server1, _Server2] = Servers =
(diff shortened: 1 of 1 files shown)


mkuratczyk:
If possible, the test condition should not be "if it's a mixed version test" but something specific to 4.0-4.1 upgrade.
Once mixed version tests run 4.1-4.2 or anything beyond, it should run without further actions (performing which we are likely to forget)

michaelklishin:
@mkuratczyk you are right but we do not have such a mechanism in CT helpers. And some of our team members have discussed revisiting such excluded tests for 4.1.x specifically, so I think we won't necessarily forget about them.

michaelklishin:
Given that the majority of such excluded tests can be re-enabled when a release series rolls over, I don't think we necessarily need an exclusion mechanism for a specific part of versions.

There are so many build system-related changes going on that I'd rather not try to introduce it in the next few months.

michaelklishin:
Per discussion on the team: we will have to revisit such checks after every release, because in a (future) 4.1/4.0 combination, this test would pass and can be run in mixed version mode.

In the future we will likely support N-2 upgrades (this is not a promise of delivery but just an idea on the team), in which case we'd need new helpers that detect/exclude specific combinations of release series, e.g. 4.1.x and 3.13.x but not 4.1.x and 4.0.x.

Issue #12619, opened 2024-10-30
Methods to determine deprecated features in use report incomplete/inaccurate information


Zerpet:
### Describe the bug

The command `rabbitmq-diagnostics check_if_any_deprecated_features_are_used` and the API endpoint `/api/deprecated-features/used` do not report accurate information on actual state of the system. The command/endpoint only seem to check for classic queue mirroring, and it doesn't report the state of other deprecated features in use.

### Reproduction steps

1. Start RabbitMQ 4.0.3 (latest at this time): `docker run --name bunny -p 15672:15672 --rm rabbitmq:4.0-management`
2. List deprecated features: `docker exec bunny rabbitmq-diagnostics --formatter=pretty_table list_deprecated_features`
    Observe `transient_nonexcl_queues` as deprecated feature, permitted by default
3. Create a transient non-exclusive classic queue:
    ```
    curl -v -u guest:guest -H "Content-Type: application/json" -d '{"auto_delete":false,"durable":false}' -X PUT http://localhost:15672/api/queues/%2F/myqueue
    ```
4. Query the state of deprecated features used: `curl -v -u guest:guest http://localhost:15672/api/deprecated-features/used`
5. Observe that none `[]` are reported
6. Observe that server logs print __Deprecated features: `transient_nonexcl_queues`: Feature `transient_nonexcl_queues` is deprecated.__
7. Run `docker exec bunny rabbitmq-diagnostics --formatter=pretty_table check_if_any_deprecated_features_are_used`
8. Observe the reported output: `Cluster reported no deprecated features in use`

### Expected behavior

The command `rabbitmq-diagnostics --formatter=pretty_table check_if_any_deprecated_features_are_used` reports that `transient_nonexcl_queues` deprecated feature is used.

The API endpoint `/api/deprecated-features/used` reports that `transient_nonexcl_queues` deprecated feature is used.

### Additional context

Setting `deprecated_features.permit.transient_nonexcl_queues = false` causes the queue declare operation to fail right away.

michaelklishin:
It's the CLI command that specifically only checks for CMQs.

Zerpet:
> It's the CLI command that specifically only checks for CMQs.

Strange, I can reproduce the issue using the HTTP API as well 😕

michaelklishin:
Maybe the HTTP API endpoints do the same thing, the CLI command was the first thing I've taken a look at.

dumbbell:
[...]

PR #12674, opened 2024-11-06
rabbit_amqqueue: Add `is_feature_used` callback to `transient_nonexcl_queues` depr. feature

dumbbell:
## Why

Without this callback, the deprecated features subsystem can't report if the feature is used or not.

This reduces the usefulness of the HTTP API endpoint or the CLI command that help verify if a cluster is using deprecated features.

## How

The callback counts transient non-exclusive queues and return `true` if there are one or more of them.

References #12619.

### DIFF
--- a/deps/rabbit/src/rabbit_db_queue.erl
+++ b/deps/rabbit/src/rabbit_db_queue.erl
@@ -46,7 +46,8 @@
 
 %% Used by on_node_up and on_node_down.
 %% Can be deleted once transient entities/mnesia are removed.
--export([foreach_transient/1,
+-export([list_transient/0,
+         foreach_transient/1,
          delete_transient/1]).
 
 %% Only used by rabbit_amqqueue:forget_node_for_queue, which is only called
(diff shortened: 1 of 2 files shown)


michaelklishin (approved):

deps/rabbit/src/rabbit_db_queue.erl
```diff
@@ -965,6 +966,40 @@ set_in_khepri(Q) ->
...
+%%
+%% @private
+
+list_transient() ->
+    rabbit_khepri:handle_fallback(
+      #{mnesia => fun() -> list_transient_in_mnesia() end,
+        khepri => fun() -> list_transient_in_khepri() end
+       }).
+
+list_transient_in_mnesia() ->
```
gomoripeti, line 985: sorry for the late comment, but to me it looks like if mnesia is used, this will list all queues, not just transient ones.
I get a "Deprecated features are being used" "transient_nonexcl_queues | In use Permitted by default" warning when I have a single durable quorum queue.

```
> rabbit_db_queue:list_transient().
{ok,[#amqqueue{name = #resource{virtual_host = <<"/">>,
                                kind = queue,name = <<"qq1">>},
               durable = true,auto_delete = false,exclusive_owner = none,
               ...
```

dumbbell: Then there is a long-standing bug in this function. Could you please file an issue for it? I can’t work on this one right now.

PR #12683, opened 2024-11-07
Make sure Management UI JavaScript files are forced to reload after an upgrade

mkuratczyk:
## What

Users should no longer run into issues with the Management UI when upgrading RabbitMQ. There are numerous reports
of JavaScript errors, blank pages, or just header+footer pages (without the main content).

## How?

1. When build RMQ, a version is injected into index.html
2. `index.html` refers to all our JS and CSS files with query string `v=$VERSION` and therefore the URL changes from version to version
3. If Management UI is disconnected from the server (red error at the top) and then reconnects, it'll trigger a full reload (likely we are now connected to a different node and/or different server version)
4. If the Management UI is open on the main page, it checks the `management_version` returned by the `/api/overview` endpoint and compares it with its own version (defined in index.html). If there's a difference - a full page reload is performed (this is clearly an upgrade)

Breaking change: we no longer support MSIE 8 :)  (`evil.css` was a piece of history: https://github.com/rabbitmq/rabbitmq-server/commits/v4.0.x/deps/rabbitmq_management/priv/www/css/evil.css)

### DIFF
--- a/deps/rabbitmq_management/Makefile
+++ b/deps/rabbitmq_management/Makefile
@@ -45,3 +45,6 @@ list-dist-deps::
 prepare-dist::
 	$(verbose) sed 's/%%VSN%%/$(PROJECT_VERSION)/' bin/rabbitmqadmin \
 		> $(EZ_DIR)/priv/www/cli/rabbitmqadmin
+	mv $(EZ_DIR)/priv/www/css/main.css $(EZ_DIR)/priv/www/css/main-$(PROJECT_VERSION).css
+	$(verbose) sed "s#css/main.css#css/main-$(PROJECT_VERSION).css#" priv/www/index.html \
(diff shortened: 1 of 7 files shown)


michaelklishin (approved):
This is a widely used approach and given our "minimalistic" (read: simplistic) UI code, this seems like the best approach.

Works as expected, including the full reload when the API reports a different version.

michaelklishin:
Not sure why Selenium suites begin to flake again but given the scope of the change, I assume that if some of them pass, then all of them should and these are, indeed, flakes.

mkuratczyk:
The OAuth2 issue looks legitimate - seems to be a race condition based on whether `prefs.js` is fetched before bootstrap.js is executed

mkuratczyk:
Closing in favour of https://github.com/rabbitmq/rabbitmq-server/pull/12749/

PR #12708, opened 2024-11-12
4.0.x: Avoid an exception when an AMQP 0-9-1-originating message with expiration set is converted for an MQTT consumer

michaelklishin:
Note: this is submitted directly against `v4.0.x` because the [`mc_mqtt` module](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_mqtt/src/mc_mqtt.erl) in `main` is different and an equivalent line must be located first, so there might be a separate future PR for that.

See https://github.com/rabbitmq/rabbitmq-server/discussions/12707.

Version for `main`: https://github.com/rabbitmq/rabbitmq-server/pull/12710.

To reproduce the exception:

1. Start a node with `rabbitmq_mqtt` enabled
2. Tail its logs
3. Install `github.com/rabbitmq/omq`
4. `omq amqp-mqtt --message-ttl 10s -t /topic/my_topic -T my_topic -C 1`

### DIFF
--- a/deps/rabbitmq_mqtt/src/mc_mqtt.erl
+++ b/deps/rabbitmq_mqtt/src/mc_mqtt.erl
@@ -426,7 +426,7 @@ protocol_state(Msg = #mqtt_msg{props = Props0,
                 undefined ->
                     Props2;
                 Ttl ->
-                    case maps:get(?ANN_TIMESTAMP, Anns) of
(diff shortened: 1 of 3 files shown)


ansd:
Closed in favour of the Mergify backport in https://github.com/rabbitmq/rabbitmq-server/pull/12711

PR #12709, opened 2024-11-12
Ensure only alive leaders and followers when fetching QQ replica states

Ayanda-D:
## Proposed Changes

This addresses issues reported in https://github.com/rabbitmq/rabbitmq-server/discussions/12701 and https://github.com/rabbitmq/rabbitmq-server/discussions/10423 where quorum queue leader and follower FSMs have been observed to be down (`noproc`) while reporting tools such as `check_if_node_is_quorum_critical` still pass. With this patch, such terminated processes should now be reported by existing CLI tools to help avoid risks of carrying out operations such as node restarts/upgrades, when some QQ members aren't in a good state.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -442,17 +442,25 @@ become_leader0(QName, Name) ->
 all_replica_states() ->
     Rows0 = ets:tab2list(ra_state),
     Rows = lists:map(fun
-                         ({K, follower, promotable}) ->
-                             {K, promotable};
-                         ({K, follower, non_voter}) ->
-                             {K, non_voter};
-                         ({K, S, _}) ->
-                             %% voter or unknown
-                             {K, S};
-                         (T) ->
-                             T
+                        (T = {K, _, _}) ->
+                            case rabbit_process:is_registered_process_alive(K) of
+                                true ->
+                                    to_replica_state(T);
+                                false ->
+                                    []
+                            end;
+                        (_T) ->
+                            []
                      end, Rows0),
-    {node(), maps:from_list(Rows)}.
+    {node(), maps:from_list(lists:flatten(Rows))}.
+
+to_replica_state({K, follower, promotable}) ->
+    {K, promotable};
+to_replica_state({K, follower, non_voter}) ->
+    {K, non_voter};
+to_replica_state({K, S, _}) ->
+    %% voter or unknown
+    {K, S}.
 
 -spec list_with_minimum_quorum() -> [amqqueue:amqqueue()].
 list_with_minimum_quorum() ->
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -441,18 +441,19 @@ become_leader0(QName, Name) ->
 -spec all_replica_states() -> {node(), #{atom() => atom()}}.
 all_replica_states() ->
     Rows0 = ets:tab2list(ra_state),
-    Rows = lists:map(fun
+    Rows = lists:filtermap(
+                    fun
                         (T = {K, _, _}) ->
-                            case rabbit_process:is_registered_process_alive(K) of
-                                true ->
-                                    to_replica_state(T);
-                                false ->
-                                    []
+                            case whereis(K) of
+                                undefined ->
+                                    false;
+                                P when is_pid(P) ->
+                                    {true, to_replica_state(T)}
                             end;
                         (_T) ->
-                            []
-                     end, Rows0),
-    {node(), maps:from_list(lists:flatten(Rows))}.
+                            false
+                    end, Rows0),
+    {node(), maps:from_list(Rows)}.
 
 to_replica_state({K, follower, promotable}) ->
     {K, promotable};
--- a/deps/rabbit/test/quorum_queue_SUITE.erl
+++ b/deps/rabbit/test/quorum_queue_SUITE.erl
@@ -97,7 +97,8 @@ groups() ->
                                             force_all_queues_shrink_member_to_current_member,
                                             force_vhost_queues_shrink_member_to_current_member,
                                             policy_repair,
-                                            gh_12635
+                                            gh_12635,
+                                            replica_states
                                            ]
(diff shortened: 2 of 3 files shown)


deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -442,17 +442,25 @@ become_leader0(QName, Name) ->
...
-                             {K, promotable};
-                         ({K, follower, non_voter}) ->
-                             {K, non_voter};
-                         ({K, S, _}) ->
-                             %% voter or unknown
-                             {K, S};
-                         (T) ->
-                             T
+                        (T = {K, _, _}) ->
+                            case rabbit_process:is_registered_process_alive(K) of
```
kjnilsson, line 446: `whereis/1` will do here, you are already on the node where the process is running.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -442,17 +442,25 @@ become_leader0(QName, Name) ->
...
+                                true ->
+                                    to_replica_state(T);
+                                false ->
+                                    []
```
kjnilsson, line 450: I'm not sure about returning nil here. is that expected?

kjnilsson (changes requested):

deps/rabbit/src/rabbit_quorum_queue.erl
Ayanda-D: Ok, i avoided the `rabbit_process:is_process_alive/1` which does the extra distribution work. But also noticed we have `rabbit_process:is_registered_process_alive/1` which is already doing `whereis/1` with `is_pid/1` guard check on the [local node](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_process.erl#L77-L78), which i'd still have to do on the case statement. seems to be doing same thing (if i'm seeing correct)

deps/rabbit/src/rabbit_quorum_queue.erl
Ayanda-D: hmm, I think the `noproc` issue, if my understanding this right, is when we have an entry in `ra_state` of a member which had already terminated, but wasnt cleaned up from ets from FSM terminate callback. Then its still listed here as if it were still active, yet not alive. But if alive=false, we should handle it as if it had been removed from `ra_state`, so returning nil here. (i'm not sure if we should also be doing a deletion from `ra_state` here when the fsm proc is not alive)

kjnilsson: all other code paths return a tuple `{RegName, Status}` so returning `[] feels a bit odd, could you either return `{K, noproc}` or use lists:filtermap to remove the entry all together?

deps/rabbit/src/rabbit_quorum_queue.erl
kjnilsson:
```case whereis(K) of
         undefined -> ...
         Pid -> to_replica_state(T)
    end
```

I'm not sure using a helper function for the above is worth it, just causes read indirection.

Ayanda-D: k cool, definitely easier to read. sorry for the back-&-forth on this small item (tendency has been check/use rabbit helpers 1st 😊 after finding myself rewriting some existing utils in the past 😅 ). Yep, the read indirection not necessary here 👍

deps/rabbit/src/rabbit_quorum_queue.erl
Ayanda-D: Ok,  `lists:filtermap` sounds good. this was removing/flattening out the `[]` on the conversion to a `map`, but no need for that extra step. updated 👍

kjnilsson (approved):
looks good, thanks @Ayanda-D

michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/pull/12727

michaelklishin (changes requested):
```shell
gmake ct-unit_quorum_queue
```

[fails repeatably](https://github.com/rabbitmq/rabbitmq-server/actions/runs/11843844517/job/33009072541?pr=12727), both on Actions and for locally for me, with

```
Node: rabbit_shard1@localhost
Case: unit_quorum_queue_SUITE:all_replica_states_includes_nonvoters
Reason: {error,
            {{badmatch,{rabbit_shard1@localhost,#{}}},
             [{unit_quorum_queue_SUITE,all_replica_states_includes_nonvoters,
                  1,
                  [{file,"unit_quorum_queue_SUITE.erl"},{line,111}]},
              {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1793}]},
                  [{file,"test_server.erl"},{line,1302}]},
                  [{file,"test_server.erl"},{line,1234}]}]}}

Node: rabbit_shard4@localhost
Case: mirrored_supervisor_SUITE:{end_per_group,broker_tests,[]}
Reason: {error,
            {{assertEqual,
                 [{module,rabbit_ct_broker_helpers},
                  {line,1218},
                  {expression,"CrashesCount"},
                  {expected,0},
                  {value,21}]},
             [{rabbit_ct_broker_helpers,find_crashes_in_logs,2,
                  [{file,"rabbit_ct_broker_helpers.erl"},{line,1218}]},
              {rabbit_ct_broker_helpers,stop_rabbitmq_nodes,1,
                  [{file,"rabbit_ct_broker_helpers.erl"},{line,1174}]},
                  [{file,"rabbit_ct_helpers.erl"},{line,136}]},
              {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1793}]},
                  [{file,"test_server.erl"},{line,1390}]},
                  [{file,"test_server.erl"},{line,1234}]}]}}
```

Ayanda-D:
didnt know there were more QQ tests elsewhere :) updated, thanks for the review & QA @kjnilsson @michaelklishin  👍

michaelklishin:
Thank you, @Ayanda-D.

michaelklishin:
Merged in https://github.com/rabbitmq/rabbitmq-server/pull/12727.

PR #12710, opened 2024-11-12
4.1: Avoid an exception when an AMQP 0-9-1-originating message with expiration set is converted for an MQTT consumer

michaelklishin:
when an AMQP 0-9-1 publisher publishes a message
that has expiration set.

Stack trace was contributed in #12707 by @rdsilio.

Originally submitted against `v4.0.x` as https://github.com/rabbitmq/rabbitmq-server/pull/12708.

To reproduce the exception:

1. Start a node with `rabbitmq_mqtt` enabled
2. Tail its logs
3. Install `github.com/rabbitmq/omq`
4. `omq amqp-mqtt --message-ttl 10s -t /topic/my_topic -T my_topic -C 1`


ansd (approved):
Thank you @michaelklishin !

PR #12712, opened 2024-11-13
QQ: reduce memory use when dropping many messages at once.

### LINKED ISSUE #12608: Quorum queues: memory spike when applying a `max-length` policy to a long queue
### Describe the bug

Given a long quorum queue, if I apply a policy to limit the queue's length (in a real-world scenario, likely with the intention of preventing further queue growth and running out of memory), a significant memory spike occurs to drop the messages above the new threshold. This can easily cause the opposite effect than intended: I run out of memory because I was trying to prevent running out of memory...

In my particular case, it was even "funnier": I had a cluster on Kubernetes, applied a policy, the leader was OOMkilled, a new leader was elected and tried to apply the policy, so it was OOMkilled. The remaining node survived because a leader could not be elected, but as soon as one of the nodes restarted, a leader was elected and OOMkilled. A policy that was meant to limit memory usage, cause an OOMkill-loop. :)

### Reproduction steps

1. `make run-broker` (tested on `main`)
2. Publish a significant number of messages: `perf-test -qq -u qq -x 4 -y 0 -c 100 -s 5000 -ms -C 1250000`
3. Apply a policy that sets the limit to a low value: `rabbitmqctl set_policy max qq '{"max-length": 1234}'`
4. Observe memory usage

Screenshot 2024-10-29 at 12 35 47

set_policy-main

### Expected behavior

Ideally there should be no significant spike when dropping messages.

### Additional context

_No response_

kjnilsson:
As may happen when a max_length configuration change is made when there are many messages on the queue.
Fixes #12608

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -1596,10 +1596,29 @@ drop_head(#?STATE{ra_indexes = Indexes0} = State0, Effects) ->
             #?STATE{cfg = #cfg{dead_letter_handler = DLH},
                     dlx = DlxState} = State = State3,
             {_, DlxEffects} = rabbit_fifo_dlx:discard([Msg], maxlen, DLH, DlxState),
-            {State, DlxEffects ++ Effects};
+            {State, combine_effects(DlxEffects, Effects)};
         empty ->
             {State0, Effects}
     end.
+
+%% combine global counter update effects to avoid bulding a huge list of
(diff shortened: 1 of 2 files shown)


mkuratczyk (approved):
There's still a spike, but significantly smaller. Given this test case is pretty extreme (drops almost 5M messages), this should be fine for pretty much all normal situations.

Screenshot 2024-11-13 at 11 11 06

the-mikedavis (approved):

michaelklishin (approved):

PR #12713, opened 2024-11-13
Ra v2.16.0

kjnilsson:
This version of Ra contains a substantially refactored Ra log implementation that provides higher throughput and lower memory use in several scenarios.

Changes:

* QQ: Use new `log_ext` effect

  This offloads the work of reading messages from on-disk segments
  to the interacting process (typically a session or channel process)
  rather than doing this blocking, performance affecting work in the 
  Ra server process.

* Ensure opened segments are closed after some time of inactivity

  Processes that have received messages that had to be read from disks
  may keep a segment open indefinitely. This introduces a timer which
  after some time of inactivity will close all opened segments to ensure
  file descriptors are not kept open indefinitely.

* Use `initial_machine_version` config to avoid initialising from rabbit_fifo version 0.

* Use `initial_machine_version` config to avoid initialising from rabbit_stream_coordinator version 0.

* Avoid dead lock in queue federation.

  When processing the queue federation startup even the process
  may call back into the Ra process causing a deadlock. in this
  case we spawn a temporary process to avoid this.


deps/rabbit/src/rabbit_fifo_client.erl
```diff
@@ -863,7 +915,39 @@ handle_delivery(_QName, _Leader, {delivery, Tag, [_ | _] = IdMsgs},
...
+                     Ref = erlang:send_after(?CACHE_SEG_TIMEOUT, self(),
+                                             {'$gen_cast',
+                                              {queue_event, QName,
+                                               {Leader, close_cached_segments}}}),
+                     {Ref, now_ms(), Cached1};
+                 true ->
+                     {TRef, now_ms(), Cached1};
+                 false ->
+                     if TRef =/= undefined ->
+                            _ = erlang:cancel_timer(TRef, [{async, true}]),
```
ansd, line 942: Is this line tested?
Given that `Info` defaults to `true`, I would this line expect to crash the channel process from what I read in https://www.erlang.org/doc/apps/erts/erlang.html#cancel_timer/2
> a message on the form {cancel_timer, TimerRef, Result} is sent to the caller of erlang:cancel_timer() when the cancellation operation has been performed

kjnilsson: no I doubt this line is tested, it would be very hard to write a reliable test for it such that the timer is set but the cache is empty. In fact I'm not 100% sure it can happen unless we can trigger some weird combo of a missing key combined with some entries in the mem table and some not.
 
 I will add `{info, false}` though

kjnilsson: good catch

MODULE.bazel
the-mikedavis, line 1: I believe the `bazel/BUILD.ra` will also need some updates for the new modules, [`c79bf01` (#12753)](https://github.com/rabbitmq/rabbitmq-server/pull/12753/commits/c79bf01b2e74c837b28944f020f66a8b3cdafd2e#diff-c6e1acb0c8da78d9c8155254f86ae400fd281ffd697eae1ab4ddd7150aa792b0)

kjnilsson: is there a bazel task to do this or a manual job? gazelle appears not to have touched this file when I ran it

kjnilsson: I did a manual update, hopefully it is right now?

the-mikedavis (approved):

MODULE.bazel
the-mikedavis: I believe it has to be done manually

michaelklishin: @kjnilsson other than `bazel run gazelle`, there are no tasks that update the Bazel build files, so you did the right thing 👍

michaelklishin:
FTR, the reason why `bazel run gazelle` did not update the file in question is likely as simple as: it does not mess with any dependency build files, only the "proper RabbitMQ source files", by design.

PR #12723, opened 2024-11-14
Use preprocessor macros for building Khepri path patterns

the-mikedavis:
This is two mostly cosmetic refactors:

* Move `include/khepri.hrl` to `include/rabbit_khepri.hrl` in `rabbit`. This helps ErlangLS find the header. It seems to confuse this header with one of the same name from `khepri` itself.
* Use preprocessor macros for building path patterns. This isn't a meaningful change yet but will be used in the future to pattern match on paths in a `khepri_adv:node_props_map()`.


deps/rabbit/include/rabbit_khepri.hrl
```diff
@@ -0,0 +1,63 @@
...
+%%
+%% If you are creating a path pattern to use in a call to the Khepri API (for
+%% example `rabbit_khepri:get/1') you should prefer using the
+%% `khepri_<entity>_path' function from the `rabbit_db_<entity>' modules
+%% instead, for example `rabbit_db_queue:khepri_queue_path/2', since those
+%% functions have guards to assert that the variables passed are valid pattern
+%% components.
+
+-define(KHEPRI_ROOT_PATH, ?KHEPRI_ROOT_PATH([])).
+-define(KHEPRI_ROOT_PATH(Rest), [rabbitmq | Rest]).
```
dumbbell, line 22: Now that I think of it, we should not use the `KHEPRI_` prefix here. Because there is a risk that it pollutes Khepri's namespace. Could please rename them to use another prefix such as `RABBITMQ_KHEPRI_`?

dumbbell (changes requested):

deps/rabbit/include/rabbit_khepri.hrl
dumbbell: I mean all macros, not just the two I quoted.

dumbbell (approved):
Thank you!

PR #12737, opened 2024-11-15
rabbitmq_ct_helpers: Change how Mnesia/Khepri is selected

dumbbell:
## Why

Once `khepr_db` is enabled by default, we need another way to disable it to select Mnesia instead.

## How

We use the new relative forced feature flags mechanism to indicate if we want to explicitly enable or disable `khepri_db`. This way, we don't touch other stable feature flags and only mess with Khepri.

However, this mechanism is not supported by RabbitMQ 4.0.x and older. They will ignore the setting. Therefore, to make this work in mixed-version testing, after a node has been started, we try to enable `khepri_db` if we detect that it doesn't support this mechanism.

At the end, we compare the effective metadata store to the expected one. If they don't match, we skip the test.

While here, change `rjms_topic_selector_SUITE` to only choose Khepri without specifying any feature flags.

### DIFF
--- a/.github/workflows/test-make-target.yaml
+++ b/.github/workflows/test-make-target.yaml
@@ -57,7 +57,7 @@ jobs:
       uses: dsaltares/fetch-gh-release-asset@master
       if: inputs.mixed_clusters
       with:
-        version: 'tags/v4.0.3'
+        version: 'tags/v4.0.5'
         regex: true
         file: "rabbitmq-server-generic-unix-\\d.+\\.tar\\.xz"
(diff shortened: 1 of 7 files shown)


dumbbell:
This pull request depends on the release of RabbitMQ because it needs the following fixes already backported to `v4.0.x`:
* one in the Khepri reset code backported in #12739 and published in RabbitMQ 4.0.4.
* one in the quorum queue code backported in #12850 and published in RabbitMQ 4.0.5.

PR #12743, opened 2024-11-16
Revival of #9620 Pass the message to `rabbit_backing_queue:discard` callback

noxdafox:
## Proposed Changes

This is a refresher of PR #9620. It was requested to come back with these changes after HA/Mirroring queues were deprecated.

Here I include the previous PR description:

The `rabbit_backing_queue:discard` callback is passing the message ID to the implementer. This is often not enough to carry on some necessary work as it's seen in the rabbit_priority_queue [comment](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_priority_queue.erl#L252).

In the above issue a consumer starts consuming with noAck over an empty queue. A publisher publishes a single message which gets forwarded directly to the consumer. In this case, the discard callback is invoked instead of publish_delivered and therefore it's header is not removed from the deduplication cache.

Problem is the discard callback only forwards the message ID and not the whole message not providing enough context for the implementer.

This patch adjust the rabbit_backing_queue behaviour passing the whole message to the discard callback instead of its sole ID.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue_process.erl
+++ b/deps/rabbit/src/rabbit_amqqueue_process.erl
@@ -648,7 +648,7 @@ discard(#delivery{confirm = Confirm,
                true  -> confirm_messages([MsgId], MTC, QName);
                false -> MTC
            end,
-    BQS1 = BQ:discard(MsgId, SenderPid, BQS),
+    BQS1 = BQ:discard(Msg, SenderPid, BQS),
     {BQS1, MTC1}.
 
 run_message_queue(ActiveConsumersChanged, State) ->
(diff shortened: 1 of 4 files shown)


michaelklishin:
Accepted in https://github.com/rabbitmq/rabbitmq-server/pull/12921.

noxdafox:
@michaelklishin, are you sure this was handled correctly? I don't see the commits in #12921. Neither I see the changes in `main` at the moment.

noxdafox:
@michaelklishin is there any blocker for this PR? I would like to get it through so I can make a new release of the plugin.

michaelklishin:
Rebased (the branch was 430 commits behind) and pushed to https://github.com/rabbitmq/rabbitmq-server/pull/13374.

PR #12749, opened 2024-11-18
Insert control-cache headers to every resource

MarcialRosales:
## Proposed Changes

Add response's header that tells the browser to always check whether a web resource has been modified (using eTag) before delivering to the user.


michaelklishin:
@MarcialRosales I cannot reproduce the file reload behavior in `main` (using the acceptance tests outlined) but I can on this branch. So, for me, it works as expected in two different browsers on two different machines.

MarcialRosales:
Thanks a lot @michaelklishin for testing it. I will make it ready for view soon.

michaelklishin:
@MarcialRosales I have chatted with @mkuratczyk about this and we think this PR is ready to go.

michaelklishin:
Backporting to `v4.0.x` will take some time, so this will not ship in `4.0.4` but rather `4.0.5`.

PR #12753, opened 2024-11-18
Bump Khepri to 0.17.0

the-mikedavis:
This is not ready to be merged yet, Khepri 0.17.0 hasn't yet been released. (And we want to make more changes before then anyways.)

This PR handles the breaking changes in Khepri 0.17.0.


deps/rabbit/src/rabbit_db_queue.erl
```diff
@@ -409,17 +416,26 @@ delete_in_khepri(QueueName) ->
...
+        %% Khepri 0.16 and below returned `khepri:node_props()' for
+        %% adv queries and commands targeting one node:
+        {ok, #{data := _}} ->
+            %% we want to execute some things, as decided by rabbit_exchange,
+            %% after the transaction.
+            rabbit_db_binding:delete_for_destination_in_khepri(QueueName, OnlyDurable);
+        %% Khepri 0.17+ returns `khepri_adv:node_props_map()`
+        %% instead.
+        {ok, #{Path := #{data := _}}} ->
+            rabbit_db_binding:delete_for_destination_in_khepri(QueueName, OnlyDurable);
```
dumbbell, line 434: Oh, that’s a good point! This probably needs a fix in Khepri. I see two solutions:
1. The typespec is updated to document the old return value (which it can return)
2. The function converts the old return value to the new one (I didn’t check if it was possible though)

What do you think?

the-mikedavis: Do we need a change in Khepri? https://github.com/rabbitmq/khepri/pull/303 only touched the "client" code (i.e. not `khepri_machine`) so older versions should continue to see `khepri:node_props()` while newer versions consistently see `khepri_adv:node_props_map()`, even if the leader were an older versioned node.

The typespec change is unfortunate and the dialyzer had a few complaints that I worked around with `-dialyzer({nowarn_function, [..]}).` attributes. I think this is ok for compatibility code like this. We should be able to delete these branches and remove the dialyzer ignores for 4.2.x once we don't need compatibility with Khepri 0.16. (Assuming Khepri 0.17 would ship with 4.1.x.)

dumbbell: > newer versions consistently see khepri_adv:node_props_map(), even if the leader were an older versioned node.

Right, we have `reply_from => LocalMember` for commands, so the return value is created by the local version of Khepri. Then, why do we need to have the `case` for both return values?

With the discussions around long-term support for RabbitMQ, compatibility code can stay for a long time as we will have to support upgrades that skip versions. That said, it might not be the case with RabbitMQ 4.0.x. This is more of a general comment.

the-mikedavis: Oh actually that's a really good point, I suppose the old return value isn't possible and we shouldn't need this compatibility code

dumbbell:
Thank you for updating the branch!

I’m working on updating Ra in Khepri but there are small changes of behaviors that cause test flakes. I fixed one of them (Ra can reture `{error, normal}` after a Ra server exited), but not the second one yet (that new error causes a retry loop to retry until timeout).

dumbbell:
I fixed the bug in `khepri_cluster.erl` and merged the bump of Ra to 2.16.0.

Do you think we can update Khepri in RabbitMQ now? Perhaps we should focus on that and release Khepri 0.17.0 this week.

deps/rabbit/src/rabbit_db_queue.erl
```diff
@@ -1129,6 +1113,24 @@ do_delete_transient_queues_in_khepri(Qs, FilterFun) ->
             Error
     end.
 
+do_delete_transient_queues_in_khepri_tx([], Acc) ->
+    {ok, Acc};
+do_delete_transient_queues_in_khepri_tx([{Path, Vsn, QName} | Rest], Acc) ->
+     %% Also see `delete_in_khepri/2'.
+     VersionedPath = khepri_path:combine_with_conditions(
+                       Path, [#if_payload_version{version = Vsn}]),
```
dumbbell, line 1121: That’s a good idea to use the payload version here, given the read is outside of the transaction!

But now, I wonder if we need the transaction at all :-)

deps/rabbit/src/rabbit_db_exchange.erl
```diff
@@ -711,13 +710,20 @@ delete_all_in_khepri_tx(VHostName) ->
...
-              {deleted, #exchange{name = XName}, Bindings, XDeletions} =
-                rabbit_db_binding:delete_all_for_exchange_in_khepri(
-                  X, false, true),
-              Deletions1 = rabbit_binding:add_deletion(
-                             XName, X, deleted, Bindings, XDeletions),
-              rabbit_binding:combine_deletions(Deletions, Deletions1)
+      fun(Path, Props, Deletions) ->
+              case {Path, Props} of
+                  {?RABBITMQ_KHEPRI_EXCHANGE_PATH(VHostName, _),
+                   #{data := X}} ->
```
dumbbell, line 716: Mostly a note to myself: I wonder if `?RABBITMQ_KHEPRI_EXCHANGE_PATH()` can be use in the function clause directly instead of having a `case`.

the-mikedavis: I was hopeful we could do that but we would need https://github.com/erlang/eep/blob/master/eeps/eep-0055.md to be implemented - we would want to pin bindings like `^VHostName`. Without EEP55 I believe we can't do that, it would be a compile error to re-use the binding if I remember correctly

deps/rabbit/src/rabbit_db_queue.erl
the-mikedavis: I think the transaction would only be necessary for the sake of performance - it's faster to send the one transaction command than many small delete commands (since Ra handles them in serial). We might be able to make it work with async commands though, or a batch API.

The hope with some of the changes in 0.17 is to remove the transaction for regular queue deletion too (rather than just the transient ones here) since that will save resources + be faster

acogoluegnes:
The upgrade to RA 2.16 in the broker seems to cause a flake (#13172) when the stream plugin uses a consistent read to check if some nodes are in maintenance node. Feel free to ping me when you think this PR is ready, I can run the stream plugin test suite against it to make sure the flake is gone.

deps/rabbit/src/rabbit_db_queue.erl
dumbbell: I see, good point.

dumbbell:
I submitted #13217 to remove `force_reset` with Khepri. It will "fix" some test failures :-)

dumbbell:
I prepared a patch to fix the failures with peer discovery plugins that is pending CI approval.

I have another patch locally that is unfinished to address `clustering_management_SUITE`.

PR #12765, opened 2024-11-20
By @gomoripeti: Streams: two additional Prometheus metrics for connections

michaelklishin:
This is #10275 by @gomoripeti re-submitted so that Actions have access to all the GH secrets.

### DIFF
--- a/deps/amqp10_common/src/serial_number.erl
+++ b/deps/amqp10_common/src/serial_number.erl
@@ -15,9 +15,8 @@
          diff/2,
          foldl/4]).
 
--ifdef(TEST).
+%% For tests.
 -export([usort/1]).
--endif.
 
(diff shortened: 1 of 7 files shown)


michaelklishin:
I have received no objections to this PR from other core team members

michaelklishin:
This seems safe to backport to `v4.0.x`, too. It's a new metric in a node-local plugin.

michaelklishin:
@gomoripeti backporting to `v4.0.x` turned out to be more involved. If you would like to contribute a `v4.0.x` version of this PR, it will be considered, otherwise it will ship next year in `4.1.0`.

gomoripeti:
thank you. This feature is not super urgent for us so 4.1.0 sounds good. And I take a note of manually backporting to 4.0.x if time permits.

PR #12799, opened 2024-11-25
rabbit_vhosts: Don't reconcile vhosts if `rabbit` is stopped

dumbbell:
## Why

That timer was started during boot and continued regardless if `rabbit` was running or stopped.

This caused the reconsiliation to crash if the `rabbit` app was stopped before the it ended because it tried to access the database even though it was stopped or even reset.

## How

We just check if `rabbit` is running before running one reconciliation and scheduling a new one.


michaelklishin:
Good catch.

PR #12809, opened 2024-11-25
Peer discovery cleanup: be more defensive

michaelklishin:
See discussion #12807 for details. This is to avoid an exception when
the cleanup functions perform peer discovery and may get a single node back
instead of a list.

`rabbit_peer_discovery:normalize/1` can be
changed to only return lists of nodes but then
there is a number of core code paths that
treat a single node as a special "preselected" value.

So let's keep that part and convert both
sets of nodes to lists before computing the
difference.

### DIFF
--- a/deps/rabbitmq_peer_discovery_common/src/rabbit_peer_discovery_cleanup.erl
+++ b/deps/rabbitmq_peer_discovery_common/src/rabbit_peer_discovery_cleanup.erl
@@ -28,6 +28,8 @@
 -compile(export_all).
 -endif.
 
+-import(rabbit_data_coercion, [as_list/1]).
+
 -define(CONFIG_MODULE, rabbit_peer_discovery_config).
 -define(CONFIG_KEY, node_cleanup).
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
Per discussion with @dumbbell: indeed, the "single value" results of peer discovery is a special case introduced intentionally to allow for a modified seeding (see RabbitMQ 4.0 release notes for details) by certain backends.

The cleanup module simply wasn't updated to account for that.

deps/rabbitmq_peer_discovery_common/src/rabbit_peer_discovery_cleanup.erl
```diff
@@ -241,7 +243,7 @@ maybe_cleanup(State, UnreachableNodes) ->
        "Peer discovery: cleanup discovered unreachable nodes: ~tp",
        [UnreachableNodes],
        #{domain => ?RMQLOG_DOMAIN_PEER_DIS}),
-    case lists:subtract(UnreachableNodes, service_discovery_nodes()) of
+    case lists:subtract(as_list(UnreachableNodes), as_list(service_discovery_nodes())) of
```
dumbbell, line 246: This can't work with a single seed node. The cleanup should be skipped because we don't have the list of members.

If we use the seed node as a list, this code will uncluster everyone.

dumbbell (changes requested):

PR #12821, opened 2024-11-26
Definition export: inject default queue type into virtual host metadata

### LINKED ISSUE #12776: DQT defined in `rabbitmq.conf` is not injected into exported definitions
default_queue_type is not reflected in definitions export nor /api/vhosts. Additionally, it seems ineffective in 3.13.

Affected versions: 3.13, 4.0, main.

Reproduction steps:

```
echo 'default_queue_type = quorum' > /tmp/dqt.conf
make run-broker RABBITMQ_CONFIG_FILE=/tmp/dqt.conf

# export_definitions doesn't mention DQT
rabbitmqctl export_definitions - | jq | rg quorum

# /api/vhosts returns DQT as undefined
curl -u guest:guest -s localhost:15672/api/vhosts | jq | rg default_queue_type
    "default_queue_type": "undefined",

# the setting is effective nonetheless
rabbitmqadmin declare queue name=foo
rabbitmqctl -q list_queues name type
name	type
foo	quorum
```

On v3.13.x, I get `foo classic` in that last command, so it seems like
`default_queue_type` is not being applied at all. `rabbitmqctl environment`
returns `{default_queue_type,rabbit_quorum_queue}` as expected.

### LINKED ISSUE #12835: Introduce definition file type markers
This is a continuation to https://github.com/rabbitmq/rabbitmq-server/issues/10078.

Per discussions of https://github.com/rabbitmq/rabbitmq-server/pull/12821 and #10078, several members of the core team concluded that our definition files need a "type" (single virtual host vs. cluster-wide export) metadata field.

It then can be used for filtering out single virtual host files at definition boot, with a sensible error message.

michaelklishin:
With this change, `GET /api/definitions` and `GET /api/vhosts` will inject a default queue type (DQT) into virtual host metadata when `default_queue_type` is set in `rabbitmq.conf` (`v4.0.x`-only) and the virtual host in question does not have [DQT defined for it](https://www.rabbitmq.com/docs/vhosts):

``` ini
default_queue_type = quorum
```

Closes #12776.
Closes #12835.

### DIFF
--- a/deps/rabbit/src/rabbit_definitions.erl
+++ b/deps/rabbit/src/rabbit_definitions.erl
@@ -1071,10 +1071,13 @@ list_vhosts() ->
     [vhost_definition(V) || V <- rabbit_vhost:all()].
 
 vhost_definition(VHost) ->
+    Name = vhost:get_name(VHost),
+    DQT = rabbit_queue_type:short_alias_of(rabbit_vhost:default_queue_type(Name)),
     #{
-        <<"name">> => vhost:get_name(VHost),
+        <<"name">> => Name,
         <<"limits">> => vhost:get_limits(VHost),
-        <<"metadata">> => vhost:get_metadata(VHost)
+        <<"metadata">> => vhost:get_metadata(VHost),
+        <<"default_queue_type">> => DQT
     }.
 
 list_users() ->
--- a/deps/rabbit/src/rabbit_queue_type.erl
+++ b/deps/rabbit/src/rabbit_queue_type.erl
@@ -23,6 +23,7 @@
          feature_flag_name/1,
          to_binary/1,
          default/0,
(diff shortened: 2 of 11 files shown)


michaelklishin:
Yes, `rabbitmqctl list_vhosts` also needs to be updated.

michaelklishin:
@mkuratczyk ready for another round.

mkuratczyk (approved):
Works great, thanks. The only thing that I wasn't 100% sure about is that
if you export definitions with a DQT set and import them to a fresh node (that just defaults to `classic`), the DQT doesn't get set to the value from the definitions JSON.
But given that imports are additive, I think that's ok - we are not overriding an existing value.

michaelklishin:

Issue #12802, opened 2024-11-25
"Deprecated features are being used" for transient_nonexcl_queues even when there are no transient queues in the cluster


daanwa:
### Describe the bug

After a clean install of RabbitMQ 4.0.4 and configuring `deprecated_features.permit.transient_nonexcl_queues = false` in `/etc/rabbitmq/rabbitmq.conf`, creating a queue via the RabbitMQ Management Interface, I get the message  "Deprecated features are being used".

If I look in "Admin -> Deprecated features" I see that `transient_nonexcl_queues` is "in_use". This also occurs when I use a stream queue or quorum queue.

I would expect this message to not show when I create a new queue (of any type).

Perhaps I'm misunderstanding how these feature flags work. But I would expect this feature to not be enabled (?) or usable upon a fresh install of RabbitMQ 4.0.4

### Reproduction steps

1. Install RabbitMQ
2. Add deprecated_features.permit.transient_nonexcl_queues = false to the config file
3. Restart RabbitMQ
4. Create a queue of any type
5. See the message
6. See that `transient_nonexcl_queues` is "in_use"

### Expected behavior

To not have `transient_nonexcl_queues` is "in_use" when creating a new queue in a fresh RabbitMQ 4.0.4 install

Dutchy-:
I just upgraded to 4.0.4 and I also get this message. I do not have transient_nonexcl_queues explicitly disabled but I cannot find any queue that uses it. I only have quorum queues.

michaelklishin:
Yes, the function that detects if transient non-exclusive queues are used does not do it correctly at least with Mnesia.

This was the conclusion of https://github.com/rabbitmq/rabbitmq-server/discussions/12793.

This is a harmless warning, although I understand that it is annoying.

Tazmaniac:
All my classic queues are either "exclusive: true" or "auto-delete:true  x-expires:10000" but after the upgrade to from 4.0.4 to 4.0.5 of my cluster, I still have the deprecated feature message in the interface.
Is it because I did not completely shutdown the server ?
Are "auto-delete:true  x-expires:10000" classic queues considered "transient_nonexcl_queues" queues ? In this case which alternatives I have ?
Or is the fix incomplete ?

mkuratczyk:
I've checked and I don't see an error here. I guess your "auto-delete:true x-expires:10000" are non-durable (in other words: transient) and that's what triggers the warning.

michaelklishin:
[...]

PR #12840, opened 2024-11-27
Fix transient queue detection

mkuratczyk:
This fixes two separate issues in the same area:

1. Invalid pattern leading to an incorrect warning about transient queues being used (because all durable queues were counted).
Fixes https://github.com/rabbitmq/rabbitmq-server/issues/12802

2. A fresh deployment wouldn't boot with `deprecated_features.permit.transient_nonexcl_queues = false`
Fixes https://github.com/rabbitmq/rabbitmq-server/discussions/12793

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -118,12 +118,19 @@
      }}).
 
 are_transient_nonexcl_used(_) ->
-    case rabbit_db_queue:list_transient() of
-        {ok, Queues} ->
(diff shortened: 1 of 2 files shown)


michaelklishin (approved):

michaelklishin (approved):
I have confirmed the effectiveness of this.

With both

``` ini
deprecated_features.permit.transient_nonexcl_queues = true
```

and

```ini
deprecated_features.permit.transient_nonexcl_queues = false
```

the behavior is now what I'd expect: only transient, non-exclusive classic queues trigger a warning or, with the 2nd configuration file, the node refuses to perform the declaration.

Transient exclusive or durable queues do not trigger the warning and can be declared in both cases.

PR #12858, opened 2024-11-29
rabbitmq_peer_discovery_consul: Set `bind_addr` to 127.0.0.1 in test config

dumbbell:
## Why

The test configuration was querying a network interface IP address based on its name. However, the name, "eth0", is very specific to Linux. This broke the test on other systems.

## How

We still have to set an explicit `bind_addr` because Consul refuses to start if the host has multiple private IPv4 addresses, as it is the case in CI.

Therefore, we hard-code 127.0.0.1 as the IPv4 address to use because it has a great chance to exist about anywhere.

**V2**: We abandon the idea of not setting `bind_addr` because GitHub Actions default hosts have several private IPv4 addresses configured.

### DIFF
--- a/deps/rabbitmq_peer_discovery_consul/test/system_SUITE.erl
+++ b/deps/rabbitmq_peer_discovery_consul/test/system_SUITE.erl
@@ -197,7 +197,11 @@ start_consul(Config) ->
     ct:pal("Starting Consul daemon"),
     ConsulBin = ?config(consul_bin, Config),
     ConsulConfDir = ?config(consul_conf_dir, Config),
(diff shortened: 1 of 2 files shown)


michaelklishin:
Allowing Consul to bind to all interfaces is definitely a better idea.

dumbbell:
> Allowing Consul to bind to all interfaces is definitely a better idea.

Unfortunately that didn’t work. GitHub Actions hosts have several IPv4 addresses configured and the Consul daemon doesn’t like it.

michaelklishin:
@dumbbell should we backport this to `v4.0.x`?

dumbbell:
I will backport the pull request in order tomorrow. Thanks!

PR #12894, opened 2024-12-04
CLI: Error on compiler warnings in tests

the-mikedavis:
This is similar to https://github.com/rabbitmq/rabbitmq-server/pull/10459 but focused on `make` instead. #10459 also only focused on warnings in `lib/` files while this change causes CLI tests to fail when the test modules contain warnings. We need to add the `--warnings-as-errors` flag to `mix test` (added in Elixir 1.12). We can at the same time check that the `lib/` files compile without warning by compiling with `--warnings-as-errors` as well.

Most compiler warnings fixed were benign. The shadowing diagnostics like unused bindings are useful though - one fix covered an unused binding that should've been a match (i.e. needed the pin operator `^`).

### DIFF
--- a/deps/rabbitmq_cli/Makefile
+++ b/deps/rabbitmq_cli/Makefile
@@ -17,7 +17,7 @@ VERBOSE_TEST ?= true
 MAX_CASES ?= 1
 
 MIX_TEST_OPTS ?= ""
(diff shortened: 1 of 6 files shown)


michaelklishin:
@the-mikedavis do I understand it correctly that this change will be effective for `gmake tests` and not `gmake dialyze`? The latter is looking for `.erl` files and in the case of CLI, of course, it won't find any.

PR #12907, opened 2024-12-10
By @gomoripeti: Restore credit_flow between AMQP 0.9.1 channel/MQTT connection -> CQ processes

michaelklishin:
See #12885 for background. This is #12906 by @gomoripeti.


michaelklishin:
I will create a backport PR manually.

PR #12910, opened 2024-12-10
Backport #12907 to v4.0.x

michaelklishin:
This backports #12906 #12907 to `v4.0.x`.


michaelklishin:
The Selenium failures have nothing to do with this change. They need to be migrated to Make in `v4.0.x`, just like they were in `main`.

PR #12919, opened 2024-12-11
Use correct AMQP filter expression string modifier prefix

ansd:
Section 4.1.1 of AMQP Filter Expressions Working Draft 09 defines `&` (ampersand) instead of `$` (dollar) as the string modifier prefix.


lukebakken (approved):

michaelklishin:
To be extra clear, this is not a breaking change because this feature has not yet shipped in a GA release (it is `4.1.0`-specific).

PR #12921, opened 2024-12-12
By @noxdafox: Improve rabbit_backing_queue:is_duplicate behaviour (#12913), take 2

michaelklishin:
This is #12913 by @noxdafox, take 2, re-submitted so that Actions have access to all the secrets.


michaelklishin:
In case (hopefully) we decide to backport this to `v4.0.x`: ignore the Selenium-based suite failures, those workflows need to be updated to use Make in `v4.0.x`.

michaelklishin:
There were some Actions build environment changes that we need to address first before we can backport this potentially risky change.

PR #12922, opened 2024-12-12
Native Elixir support in Erlang.mk (SEE DESCRIPTION BEFORE MERGE)

lhoguin:
I would recommend not merging this until after 4.1 is released, or at least to keep this in `main`.

The related Erlang.mk PR is at https://github.com/ninenines/erlang.mk/pull/1020 and is currently not merged, therefore the Erlang.mk in this PR is experimental. It would be good to have some feedback before merging into Erlang.mk, and RabbitMQ `main` is a good candidate for that as it has a low impact compared to release branches.


lhoguin:
This should not be merged before I am back from PTO but we are getting to a point where it can be tested as everything should work fine and tests are green.

Mix remains used but only when testing rabbitmq_cli (I do not see much value in converting that part).

With Mix removed this opens the way to moving to a proper apps/ structure and avoid going through all rabbit apps every time we compile but this will be done at a later time, priorities!

PR #12931, opened 2024-12-12
Backport #12921 by @noxdafox

michaelklishin:
This is a manual backport of #12921 by @noxdafox needed by  https://github.com/noxdafox/rabbitmq-message-deduplication.


michaelklishin:
There were some Actions build environment changes that we need to address first before we can backport this potentially risky change.

PR #12936, opened 2024-12-13
Pin Bazel to v7.4.1

the-mikedavis:
This works around incompatibilities from the recent Bazel 8.0.0 release.


michaelklishin:
Thank you, @the-mikedavis 👏

PR #12944, opened 2024-12-14
4.0.6: backport #12921 by @noxdafox, take 2

michaelklishin:


michaelklishin:
This is potentially risky change so it will have to wait until `4.0.6`.

PR #12951, opened 2024-12-16
WIP Fix v4.0.x CI

ansd:
This reverts commit bdce755c4ef42434f97459a23855e11d7244d0ef because it broke CI. `make` tests are not run anymore.


michaelklishin:
FTR, we have applied both of these reverts separately. They did make a difference.

PR #12954, opened 2024-12-16
v4.0.x: AMQP 1.0 .NET-based integration suite: an attempt to fix 'dotnet restore'

michaelklishin:
As a result recent Actions environment upgrade, `dotnet restore` currently throws

```
Process terminated. Couldn't find a valid ICU package installed on the system.
Set the configuration flag System.Globalization.Invariant to true if you want
to run with no globalization support.
```

on `v4.0.x`.


michaelklishin:
Per discussion with @ansd: perhaps an even better idea would be to upgrade .NET to 8 in `v4.0.x`, there is no reason why it should stay on `3.1.x` :)

PR #12960, opened 2024-12-16
Remove actions/setup-dotnet

ansd:
According to https://github.com/rabbitmq/rabbitmq-server/pull/12955#pullrequestreview-2506932356 and https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#net-tools dotnet 8.0 is already installed on Ubuntu 24.04.


lukebakken (approved):
👍

michaelklishin (approved):

PR #12980, opened 2024-12-19
Upgrade eetcd and gun

ansd:
## Why?

To introduce AMQP over WebSocket, we will add gun to the Erlang AMQP
1.0 client. We want to add the latest version of gun for this new
feature. Since rabbitmq_peer_discovery_etcd depends on the outdated
eetcd 0.3.6 which in turn depends on the outdated gun 1.3.3, this commit
first upgrades eetcd and gun.

## How?
See https://github.com/zhongwencool/eetcd?tab=readme-ov-file#migration-from-eetcd-03x-to-04x

## Breaking Changes

This commit causes the following breaking change:
`rabbitmq.conf` settings
* `cluster_formation.etcd.ssl_options.fail_if_no_peer_cert`
* `cluster_formation.etcd.ssl_options.dh`
* `cluster_formation.etcd.ssl_options.dhfile`

are unsupported because they are not valid `ssl:tls_client_option()`.

See https://github.com/erlang/otp/issues/7497#issuecomment-1636012198


lhoguin (approved):
It looks fine BUT it would be much better if it had TLS tests to make sure.

PR #12985, opened 2024-12-19
rabbitmq_peer_discovery_etcd: Wait for etcd start in system_SUITE

### LINKED ISSUE #12981: Wait for etcd listener before starting tests
Poll the etcd listener until it's up before running the client to avoid the `timer:sleep/1` call in https://github.com/rabbitmq/rabbitmq-server/commit/c23c632753ef329fd728e0a48c60f9e386d9acb6.

dumbbell:
## Why

It was possible that testcases were executed before the etcd daemon was ready, leading to test failures.

## How

There was already a santy check to verify that the etcd daemon was working correctly, but it was itself a testcase.

This patch moves this code to the etcd start code to wait for it to be ready.

This replaces the previous workaround of waiting for 2 seconds.

While here, log anything printed to stdout/stderr by etcd after it exited.

Fixes #12981.

### DIFF
--- a/deps/rabbitmq_peer_discovery_etcd/test/system_SUITE.erl
+++ b/deps/rabbitmq_peer_discovery_etcd/test/system_SUITE.erl
@@ -26,7 +26,6 @@
          init_per_testcase/2,
          end_per_testcase/2,
 
-         etcd_connection_sanity_check_test/1,
          init_opens_a_connection_test/1,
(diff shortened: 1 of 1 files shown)


michaelklishin:
@mergifyio backport v4.0.x

ansd:
Thank you @dumbbell !

PR #12994, opened 2024-12-23
Upgrade eetcd to v0.5.0

ansd:
eetcd v0.5.0 is compatible with OTP 27 and depends on gun 2.1.0


ansd:
FTR: Unless we see any breakage of `eetcd 0.3.6` on `v4.0.x` with OTP 27, @michaelklishin and I decided to keep using `eetcd 0.3.6` including `gun 1.3.3` on `v4.0.x`.

In other words, for now, we backport neither this PR nor https://github.com/rabbitmq/rabbitmq-server/pull/12980 because the risks of TLS config breakage of backporting https://github.com/rabbitmq/rabbitmq-server/pull/12980 to `v4.0.x` is higher than the benefits of using the latest `eectd` and `gun` versions.

PR #13014, opened 2025-01-02
v3.13.x: update copyright year

michaelklishin:
There are no functional changes in this PR.

This change is backported to reduce the delta and the probability of silly conflicts during backports.

References #13008.


michaelklishin:
I have verified most test suites locally.

`v3.13.x` needs a Make-on-Actions to be backported.

PR #13015, opened 2025-01-02
4.1.0: Introduce a way to protect a virtual host from deletion

### LINKED ISSUE #12772: Protection of virtual hosts from "fat finger" deletion
## Problem Definition

Every so often we see users delete virtual hosts by accident. This can happen both over the HTTP API
and with `rabbitmqctl delete_vhost`.

When this happen, [certain kinds of virtual host data](https://www.rabbitmq.com/docs/definitions) can be
restored relatively easily. The definitions data set is small (compared to tens or hundreds of
GiBs a stream can contain, for example). Unfortunately, for messages stored in queues (not streams),
the backup and restore problem is a **very** different in scope.

[Tanzu RabbitMQ](https://docs.vmware.com/en/VMware-Tanzu-RabbitMQ/index.html) has a Warm Standby Replication feature that replicates both schema and messages (for QQs, streams, CQs) to a remote warm standby cluster or multiple clusters. This does help with certain fat finger errors but not all of them,
and only if a reasonably large schema synchronization interval is used.

Since virtual host deletion is one of the most destructive operations (up there with node reset),
it could use some extra protection, similarly to how opt-in (experimental) feature flag enablement was made much more difficult to do accidentally in https://github.com/rabbitmq/rabbitmq-server/pull/11998, https://github.com/rabbitmq/rabbitmq-server/pull/12643.

In addition, specific clusters, systems, and RabbitMQ-based distributions such as Tanzu RabbitMQ can
have system virtual host. "System" here means .

## Proposed Solution

Virtual hosts can be marked with additional [metadata](https://www.rabbitmq.com/docs/vhosts#metadata). This metadata today stores the DQT (default queue type), a description, a and set of tags.

The same metadata map can be used to enable protection from deletion. If set,
`rabbitmqctl delete_vhost` and `DELETE /api/vhosts/{name}` will fail with a reasonably specific
error asking the user to delete the metadata key from the target virtual host first, then retry.

Deletion of the key will act as a confirmation for CLI tools, and will require the same permissions
as `DELETE /api/vhosts/{name}` in the case of HTTP API.

Dedicated CLI commands and HTTP API endpoints will be provided for locking and unlocking deletion.

## Currently Available Options

`DELETE /api/vhosts/{name}` requires the user to be tagged as `administrator`.

For CLI tools, such "roles" do not exist, so an explicit confirmation is particularly important to have.

michaelklishin:
Accidental "fat finger" virtual deletion accidents #12772 would be easier to avoid if there was a protection mechanism that would apply equally even to CLI tools and external applications that do not use confirmations for deletion operations.

This introduce the following changes:

 * Virtual host metadata now supports a new queue, 'protected_from_deletion', which, when set, will be considered by key virtual host deletion function(s)
 * `DELETE /api/vhosts/{name}` was adapted to handle such blocked deletion attempts to respond with a 412 Precondition Failed status
 * `rabbitmqctl list_vhosts` and `rabbitmqctl delete_vhost` were adapted accordingly
 * `DELETE /api/vhosts/{name}/deletion/protection` is a new endpoint that can be used to remove the protective seal (the metadata key)
 * `POST /api/vhosts/{name}/deletion/protection` marks the virtual host as protected

In the case of the HTTP API, all operations on
virtual host metadata require administrative
privileges.

Other considerations:

 * When a virtual host does not exist, the behavior remains the same: the original, protection-unaware code path is used to preserve backwards compatibility

Closes #12772.


mkuratczyk:
`protected_from_deletion` key is ignored when importing definitions. For example, after importing the attached file I get:
```
> rabbitmqctl list_vhosts name protected_from_deletion
Listing vhosts ...
name    protected_from_deletion
new     false
/       false
```
[export.json](https://github.com/user-attachments/files/18298967/export.json)

mkuratczyk:
I pushed a commit that adds the deletion protection status to the Management UI vhost's page:
Screenshot 2025-01-03 at 12 22 08

michaelklishin:
The implementation turned out to be more straightforward than I thought and we have decent test coverage, so I think we can backport this change.

PR #13046, opened 2025-01-10
Prevent node from booting if credit_flow_default_credit is not valid

JimmyWang6:
## Proposed Changes
As mention in previous [discussion](https://github.com/rabbitmq/rabbitmq-server/discussions/13041) thread, this PR add adds validation for
param `credit_flow_default_credit`, and change it to default value for the following two cases:
1. MoreCreditAfter larger than InitialCredit
2. Negative value

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [x] Bug fix (non-breaking change which fixes issue #NNNN)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
If you're unsure about any of them, don't hesitate to ask on the mailing list.
We're here to help!
This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

## Further Comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution
you did and what alternatives you considered, etc.

PR #13055, opened 2025-01-12
Prevent node from booting if credit_flow_default_credit is not valid by @JimmyWang6

michaelklishin:
This is #13046 by @JimmyWang6 with a small edit from me.


michaelklishin:
I will backport this manually.

PR #13085, opened 2025-01-16
Delete stream consumer metrics when AMQP 091 connection closes

acogoluegnes:
To avoid rogue consumer records.


mkuratczyk (approved):

ansd:
Not sure whether it's relevant or not, but commenting here just in case:
In https://github.com/rabbitmq/rabbitmq-server/pull/9356 we decided to omit the `consumer_deleted` events in classic queues since they seem to have caused performance issues.

PR #13092, opened 2025-01-17
Emit cancellation event only when stream consumer is cancelled

acogoluegnes:
Not when the channel or the connection is closed.

References #13085, #9356

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_metrics.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_metrics.erl
@@ -22,7 +22,7 @@
 -export([init/0]).
 -export([consumer_created/9,
          consumer_updated/9,
-         consumer_cancelled/3]).
+         consumer_cancelled/4]).
 -export([publisher_created/4,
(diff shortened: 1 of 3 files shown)


deps/rabbitmq_stream/src/rabbit_stream_metrics.erl
```diff
@@ -104,16 +104,20 @@ consumer_updated(Connection,
...
                                          StreamResource),
-    rabbit_event:notify(consumer_deleted,
-                        [{consumer_tag, consumer_tag(SubscriptionId)},
-                         {channel, self()}, {queue, StreamResource}]),
+    case Notify of
+      true ->
+            rabbit_event:notify(consumer_deleted,
+                                [{consumer_tag, consumer_tag(SubscriptionId)},
+                                 {channel, self()}, {queue, StreamResource}]);
+            _ -> ok
```
ansd, line 119: Can we fix the formatting here?

ansd (approved):

PR #13104, opened 2025-01-20
rabbitmq_ct_helpers: Always give `$RABBITMQ_FEATURE_FLAGS` a value

dumbbell:
… even if `skip_metadata_store_configuration` is given.

## Why

If `skip_metadata_store_configuration` is set to true, we let the node boot with the stable feature flags all enabled. This may prevent clustering in mixed-version testing if the secondary umbrella misses a stable feature flag.

## How

If `skip_metadata_store_configuration` is set, we restore the previous behavior of starting the node with the required feature flags only. I.e. all stable ones are disabled at first.

This is a follow-up to #13077.


dumbbell:
This pull request is abandonned and replaced by #13099.

PR #13071, opened 2025-01-15
AMQP over WebSocket

ansd:
Support AMQP 1.0 over WebSocket complying with https://docs.oasis-open.org/amqp-bindmap/amqp-wsb/v1.0/cs01/amqp-wsb-v1.0-cs01.html.

This PR contains two commits:
1. Support AMQP over WebSocket in the Erlang client.
2. The necessary changes outside of plugins `rabbitmq_web_amqp` and `rabbitmq_web_amqp_examples`.

PR #13115, opened 2025-01-21
Support exchange federation with MQTT 5.0 subscribers

### LINKED ISSUE #13040: Federated MQTT topic exchange with MQTT 5.0 subscribers
### Describe the bug

Federation crashes if the MQTT topic exchange (`amq.topic` by default) is federated and MQTT 5.0 clients subscribed. That's because bindings are sent from downstream to upstream as described in https://www.rabbitmq.com/docs/federated-exchanges#details and binding arguments containing Erlang record `mqtt_subscription_opts` cannot be encoded in AMQP 0.9.1.

### Reproduction steps

See https://github.com/rabbitmq/rabbitmq-server/discussions/13033#discussion-7793634

### Expected behavior

Federation should work.

### Additional context

Exporting definitions in JSON format also omits the `mqtt_subscription_opts` record from the binding arguments.

One solution is to encode the MQTT subscription options as an AMQP 0.9.1 table in the binding arguments when storing the binding in the database.

ansd:
## What?
This commit fixes #13040.

Prior to this commit, exchange federation crashed if the MQTT topic exchange
(`amq.topic` by default) got federated and MQTT 5.0 clients subscribed on the
downstream. That's because the federation plugin sends bindings from downstream
to upstream via AMQP 0.9.1. However, binding arguments containing Erlang record
`mqtt_subscription_opts` (henceforth binding args v1) cannot be encoded in AMQP 0.9.1.

 ## Why?
Federating the MQTT topic exchange could be useful for warm standby use cases.

 ## How?
This commit makes binding arguments a valid AMQP 0.9.1 table (henceforth
binding args v2).

Binding args v2 can only be used if all nodes support it. Hence binding
args v2 comes with feature flag `rabbitmq_4.1.0`. Note that the AMQP
over WebSocket
[PR](https://github.com/rabbitmq/rabbitmq-server/pull/13071) already
introduces this same feature flag. Although the feature flag subsystem
supports plugins to define their own feature flags, and the MQTT plugin
defined its own feature flags in the past, reusing feature flag
`rabbitmq_4.1.0` is simpler.

This commit also avoids database migrations for both Mnesia and Khepri
if feature flag `rabbitmq_4.1.0` gets enabled. Instead, it's simpler to
migrate binding args v1 to binding args v2 at MQTT connection establishment
time if the feature flag is enabled. (If the feature flag is disabled at
connection etablishment time, but gets enabled during the connection
lifetime, the connection keeps using bindings args v1.)

This commit adds two new suites:
1. `federation_SUITE` which tests that federating the MQTT topic
   exchange works, and
2. `feature_flag_SUITE` which tests the binding args migration from v1 to v2

### DIFF
--- a/deps/rabbit/src/rabbit_core_ff.erl
+++ b/deps/rabbit/src/rabbit_core_ff.erl
@@ -205,3 +205,10 @@
       stability     => stable,
       depends_on    => [message_containers]
      }}).
+
+-rabbit_feature_flag(
+   {'rabbitmq_4.1.0',
+    #{desc          => "Allows rolling upgrades to 4.1.x",
+      stability     => stable,
+      depends_on    => ['rabbitmq_4.0.0']
+     }}).
--- a/deps/rabbitmq_mqtt/Makefile
+++ b/deps/rabbitmq_mqtt/Makefile
@@ -43,7 +43,7 @@ export BUILD_WITHOUT_QUIC
 
 LOCAL_DEPS = ssl
 DEPS = ranch rabbit amqp10_common
-TEST_DEPS = emqtt ct_helper rabbitmq_ct_helpers rabbitmq_ct_client_helpers rabbitmq_management amqp_client rabbitmq_consistent_hash_exchange rabbitmq_amqp_client rabbitmq_stomp rabbitmq_stream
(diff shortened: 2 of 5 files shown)


kjnilsson (approved):

ansd:
The two CI checks fails for the same reason as in https://github.com/rabbitmq/rabbitmq-server/pull/13071#issuecomment-2595217323

michaelklishin (approved):

PR #13128, opened 2025-01-22
rabbit_peer_discovery: Cache computed node start time

dumbbell:
## Why

The computation of a node's system-time-based start time is based on its monotonic-time-based start time plus the current time offset.

However, since Erlang/OTP 26, that time offset is volatile by default. Therefore the computed system-time-based start time can change too. It happens in CI and this affects the seed node selection in the peer discovery code: because the sorting is unstable when the start time is the criteria, different instances of the peen discovery code could select a different node.

## How

The same code is used to compute the start time. However, the computed value is cached in a persistent_term for future queries. This way we ensure that the value is computed once and thus all queries will return the same result.

### DIFF
--- a/deps/rabbit/src/rabbit_peer_discovery.erl
+++ b/deps/rabbit/src/rabbit_peer_discovery.erl
@@ -26,7 +26,8 @@
          normalize/1,
          append_node_prefix/1,
          node_prefix/0]).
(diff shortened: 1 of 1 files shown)


dumbbell:
Superseeded by #13490.

PR #13138, opened 2025-01-23
Simplify Direct Reply-To

ansd:
This commit is no change in functionality and mostly deletes dead code.

1. Code targeting Erlang 22 and below is deleted since the mininmum required Erlang version is higher nowadays. "In OTP 23 distribution flag DFLAG_BIG_CREATION became mandatory. All pids are now encoded using NEW_PID_EXT, even external pids received as PID_EXT from older nodes." https://www.erlang.org/doc/apps/erts/erl_ext_dist.html#new_pid_ext
2. All v1 encoding and decoding of the Pid is deleted since the lower version RabbitMQ node supports the v2 encoding nowadays.

### DIFF
--- a/deps/rabbit/src/pid_recomposition.erl
+++ b/deps/rabbit/src/pid_recomposition.erl
@@ -7,7 +7,6 @@
 
 -module(pid_recomposition).
 
-
 %% API
 -export([
     to_binary/1,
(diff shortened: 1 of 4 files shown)


acogoluegnes (approved):
Ran successfully direct reply-to tests from AMQP 091 Java client and JMS client.

michaelklishin (approved):

PR #13160, opened 2025-01-27
Simplified OCI builds

mkuratczyk:
NOTE: this PR only affects the image we build for our testing purposes (`pivotalrabbitmq/rabbitmq`) and not the one used by users (called `rabbtimq` and maintained by Docker)

~TODO: currently the PR pushes images to `pivotalrabbitmq/rabbitmq-test`. After the review, I'll change it to push to `pivotalrabbitmq/rabbitmq`.~  DONE

Changes included in this PR:
* simplification of the action and Dockerfile (+60 -554 lines)
* less maintenance - since we use the `erlang` image as the base, we get the regular updates for free (in the past we had to manually bump Erlang/OpenSSL)
* the actions will no longer recompile erlang/openssl (which happened more often than it should)
* an image build can now be triggered manually with any Erlang version and with an option to build an ARM64 image
* by default we now build the image with OTP27 (OTP27 is used for compilation and at runtime)

If we backport this PR to `v4.0.x`, we should probably change the default OTP version to 26 in the backport


michaelklishin (approved):

michaelklishin:
@mergifyio backport v4.0.x

PR #13172, opened 2025-01-29
Depends on #12753: Use local read to check node is in maintenance

acogoluegnes:
In stream reader. The consistent read may cause some timeout for the metadata frame (when Khepri is used as the data store).

Depends on #12753.

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
@@ -2417,12 +2417,10 @@ handle_frame_post_auth(Transport,
         lists:sort(
             maps:keys(NodesMap)),
     %% filter out nodes in maintenance
-    Nodes =
-        lists:filter(fun(N) ->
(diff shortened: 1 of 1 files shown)


michaelklishin:
A local read should be perfectly sufficient in this case.

acogoluegnes:
It is likely to be due to the version mismatch between Khepri and RA (we upgraded RA to 2.16 in RabbitMQ, but Khepri is not fully compatible with it). I agree the stream plugin could use a local read in this case, even though the metadata frame can be used when there is instability in the cluster (client applications recovering their connection during a rolling upgrade/restart).

Maybe we should wait for #12753 to be merged and make sure it fixes the flake before changing the stream plugin to use a local read.

michaelklishin:
Yes, we can wait for #12753.

michaelklishin:
This should no longer be necessary with Khepri `0.17.x`.

Per discussion with @acogoluegnes @kjnilsson.

PR #13180, opened 2025-01-30
Fix flake on rabbitmq_mqtt auth_SUITE

MarcialRosales:
## Proposed Changes

Separate invalid client test from the valid one onto separate group and this way they use different SSL certs

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/rabbitmq_mqtt/test/auth_SUITE.erl
+++ b/deps/rabbitmq_mqtt/test/auth_SUITE.erl
@@ -591,8 +591,8 @@ client_id_from_cert_dn(Config) ->
 invalid_client_id_from_cert_san_dns(Config) ->
     MqttClientId = <<"other_client_id">>,
     {ok, C} = connect_ssl(MqttClientId, Config),
-    {error, {client_identifier_not_valid, _}} = emqtt:connect(C),    
-    unlink(C).
+    unlink(C),
+    {error, {client_identifier_not_valid, _}} = emqtt:connect(C).        
 
 ssl_user_vhost_parameter_mapping_success(Config) ->
     expect_successful_connection(fun connect_ssl/1, Config).
--- a/deps/rabbitmq_mqtt/test/auth_SUITE.erl
+++ b/deps/rabbitmq_mqtt/test/auth_SUITE.erl
@@ -68,9 +68,11 @@ sub_groups() ->
        ssl_user_vhost_parameter_mapping_vhost_does_not_exist,
        ssl_user_cert_vhost_mapping_takes_precedence_over_port_vhost_mapping
       ]},
+     {ssl_user_with_invalid_client_id_in_cert_san_dns, [],
+       [invalid_client_id_from_cert_san_dns
+       ]},
      {ssl_user_with_client_id_in_cert_san_dns, [],
-       [client_id_from_cert_san_dns,
-        invalid_client_id_from_cert_san_dns
+       [client_id_from_cert_san_dns        
        ]},
      {ssl_user_with_client_id_in_cert_san_dns_1, [],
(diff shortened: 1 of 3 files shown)


ansd (changes requested):
CI fails

deps/rabbitmq_mqtt/test/auth_SUITE.erl
```diff
@@ -588,7 +591,7 @@ client_id_from_cert_dn(Config) ->
 invalid_client_id_from_cert_san_dns(Config) ->
     MqttClientId = <<"other_client_id">>,
     {ok, C} = connect_ssl(MqttClientId, Config),
-    ?assertMatch({error, _}, emqtt:connect(C)),
+    {error, {client_identifier_not_valid, _}} = emqtt:connect(C),
```
ansd, line 594: This PR doesn't fix the flake. I can repro the flake locally as follows:
```erl
    {error, {client_identifier_not_valid, _}} = emqtt:connect(C),
    timer:sleep(1),
    unlink(C).
```
which makes the test fail with:
```
=== === Reason: {'EXIT',{shutdown,client_identifier_not_valid}}
```
That's the flake which we also observed in CI.

The correct fix is to unlink first, then connect.

MarcialRosales: i cannot reproduce it locally and i dont see CI failing because of this . I can modify the test as you suggest though.

MarcialRosales: Can you send me the link to the failed job in CI where this test case is failing ?

ansd: > Can you send me the link to the failed job in CI where this test case is failing ?

Yes, here are two flakes failing due to `{'EXIT',{shutdown,client_identifier_not_valid}}`:
* https://github.com/rabbitmq/rabbitmq-server/actions/runs/12769170662/job/35591148832
* https://github.com/rabbitmq-pro/tanzu-rabbitmq/actions/runs/12768915497/artifacts/2428404146

ansd: > i cannot reproduce it locally and i dont see CI failing because of this .

The CT test case process runs concurrently with the MQTT connection process. CI can fail if the CT test case process executes `unlink(C)` after receiving the EXIT signal.

ansd: > I still dont understand what you are suggesting. 

I'm suggesting to flip the two lines, i.e. to unlink **before** the expected failure as done in many other places in this test suite, e.g. in:
* https://github.com/rabbitmq/rabbitmq-server/blob/3a61274f0311bd4ff26fbc98e4e5cc9b855c9837/deps/rabbitmq_mqtt/test/auth_SUITE.erl#L810-L812
* https://github.com/rabbitmq/rabbitmq-server/blob/3a61274f0311bd4ff26fbc98e4e5cc9b855c9837/deps/rabbitmq_mqtt/test/auth_SUITE.erl#L945-L947
* https://github.com/rabbitmq/rabbitmq-server/blob/3a61274f0311bd4ff26fbc98e4e5cc9b855c9837/deps/rabbitmq_mqtt/test/auth_SUITE.erl#L1227-L1228

MarcialRosales: those ci failures are because this change has not been merged yet

MarcialRosales: Regardless i have committed that change.

ansd (approved):

Issue #11233, opened 2024-05-14
Consul peer discovery: nodes can leave behind a service record in case of an unresolvable address


frederikbosch:
### Describe the bug

- RabbitMQ crashes when a hostname that is listed in Consul cannot be resolved to an IP address.
- When RabbitMQ closes unexpectedly, a service might be left in the Consul service registry. When using orchestration tools like Nomad or Kubernetes, the orchestration tool should be made responsible for registering and deregistering services. Hence, RabbitMQ should only be reading from Consul.

### Reproduction steps

In order tot test PR #11045, I used a three node cluster with two scenario's.

1. Also register the service using my orchestration tool (Nomad) in Consul with the meta `erlang-node-name` set, under the same name (rabbitmq) as RabbitMQ will register the service.
2. Only let RabbitMQ do the registration in Consul.

The second scenario has one big downside: what if RabbitMQ did not close properly? Then the service remains in the registry. This could lead to an unrecoverable cluster. I actually ran into this scenario. What happened?

- I stopped the cluster, and RabbitMQ did not have time to shutdown properly, so the node was killed
- Leaving services in the registry
- When I restarted all the nodes they query consul and see the left services (status passing), and try to join it.
- This results in RabbitMQ crashing, because the hostname does not resolve anymore. In my cluster Docker services only have a resolving FQDN when they are actually running.

```
=PROGRESS REPORT==== 14-May-2024::12:25:21.123386 ===
    supervisor: {local,inet_gethost_native_sup}
    started: [{pid,<0.97.0>},{mfa,{inet_gethost_native,init,[[]]}}]

=PROGRESS REPORT==== 14-May-2024::12:25:21.133443 ===
    supervisor: {local,kernel_safe_sup}
    started: [{pid,<0.96.0>},
              {id,inet_gethost_native_sup},
              {mfargs,{inet_gethost_native,start_link,[]}},
              {restart_type,temporary},
              {significant,false},
              {shutdown,1000},
              {child_type,worker}]

2024-05-14 12:25:21.171098+00:00 [notice] <0.44.0> Application mnesia exited with reason: stopped
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> 
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> BOOT FAILED
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> ===========
2024-05-14 12:25:21.171789+00:00 [error] <0.253.0> Exception during startup:
[...]

PR #13194, opened 2025-02-01
Peer discovery: add an option to opt out of registration

frederikbosch:
## Proposed Changes

When forming a cluster, registration of the node joining the cluster might be left to (container) orchestration tools like Nomad or Kubernetes. This PR add the config option `cluster_formation.registration` which defaults to `true`. When set to `false` node registration at the backend is disabled.

When using container orchestration - we are using Nomad, it is the orchestrator that deploys containers with all requirements like network, storage, and also the registration in the service registry (Consul) with a per-service health check. There is at least one important advantage using this over the application doing the registration. When the application is not stopped gracefully for whatever reason, e.g. its OOM killed, it cannot deregister the service. This leaves behind an unlinked service entry in the registry. This caused the bug in #11233, also see [my remarks in PR #11045](https://github.com/rabbitmq/rabbitmq-server/pull/11045#issuecomment-2110144234) and [the exchange I had with @dumbbell in this discussion](https://github.com/rabbitmq/rabbitmq-server/discussions/11161#discussioncomment-9307332). When leaving registration and deregistration to the orchestrator, you cannot run in such problems.

## Types of Changes

- [X] Bug fix (non-breaking change which fixes issue #11233)
 
## Checklist

- [x] If relevant, I have added [necessary documentation in a PR](https://github.com/rabbitmq/rabbitmq-website/pull/2177)

## Further Comments

When trying to solve #11233 I was looking how I should disable node registration for Consul peer discovery. I found out that the `maybe_register` function inside `rabbit_peer_discovery` was the best option. This does however makes the option to disable registration a global peer discovery option rather specifically for Consul. Before continuing I'd like to know how this direction is perceived.

Please take into account, this is my first code in Erlang. I have additional work to do to make this a complete PR.

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -1353,6 +1353,13 @@ end}.
     {datatype, {enum, [disc, disk, ram]}}
 ]}.
 
+%% Register node during cluster formation.
+%%
+
+{mapping, "cluster_formation.registration", "rabbit.cluster_formation.register", [
+    {datatype, {enum, [true, false]}}
+]}.
+
 {translation, "rabbit.cluster_formation.node_type",
 fun(Conf) ->
     %% if peer discovery backend isn't configured, don't generate
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -1353,13 +1353,6 @@ end}.
     {datatype, {enum, [disc, disk, ram]}}
 ]}.
 
-%% Register node during cluster formation.
-%%
-
-{mapping, "cluster_formation.registration", "rabbit.cluster_formation.register", [
-    {datatype, {enum, [true, false]}}
-]}.
-
 {translation, "rabbit.cluster_formation.node_type",
 fun(Conf) ->
     %% if peer discovery backend isn't configured, don't generate
@@ -1376,6 +1369,13 @@ fun(Conf) ->
             end
     end
 end}.
+
+%% Register node during cluster formation when backend supports registration.
+%%
+
+{mapping, "cluster_formation.registration", "rabbit.cluster_formation.register", [
+    {datatype, {enum, [true, false]}}
+]}.
 
 %% Cluster formation: lock acquisition retries as passed to https://erlang.org/doc/man/global.html#set_lock-3
 %%
--- a/deps/rabbit/src/rabbit_peer_discovery.erl
+++ b/deps/rabbit/src/rabbit_peer_discovery.erl
@@ -46,6 +46,9 @@
 %% a new cluster as a virgin node
 -define(DEFAULT_NODE_TYPE, disc).
 
+%% register node by default
+-define(DEFAULT_REGISTRATION, true).
+
 %% default node prefix to attach to discovered hostnames
 -define(DEFAULT_PREFIX, "rabbit").
 
@@ -80,6 +83,16 @@ node_type() ->
             proplists:get_value(node_type, Proplist, ?DEFAULT_NODE_TYPE);
         undefined ->
             ?DEFAULT_NODE_TYPE
+    end.
+
+-spec registration() -> true | false.
+
+registration() ->
+    case application:get_env(rabbit, cluster_formation) of
+        {ok, Proplist} ->
+            proplists:get_value(registration, Proplist, ?DEFAULT_REGISTRATION);
+        undefined ->
+            ?DEFAULT_REGISTRATION
     end.
 
 -spec lock_acquisition_failure_mode() -> ignore | fail.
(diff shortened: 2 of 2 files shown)


michaelklishin:
@frederikbosch thank you for taking the time to contribute!

At first glance, this is a reasonable PR. We will discuss this with the team. Specifically @mkuratczyk and @dumbbell have been working on peer discovery-related things.

In my opinion, this is a good option to have. Can you please, however, document a justification for this feature to someone who would not be aware of #11233? We will need it both for release notes and the [doc guides](https://github.com/rabbitmq/rabbitmq-website).

frederikbosch:
@michaelklishin I have changed the text under Proposed Changes. It now includes justification for this feature and references earlier exchanges I had with @dumbbell on this topic.

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -1370,6 +1370,13 @@ fun(Conf) ->
     end
 end}.
 
+%% Register node during cluster formation when backend supports registration.
+%%
+
+{mapping, "cluster_formation.registration", "rabbit.cluster_formation.register", [
```
dumbbell, line 1376: I would prefer a name that implicitly tells the user it wants a boolean. For instance `register` (like application environment name you picked) or `perform_registration` because "register" is both a verb and a noun.

deps/rabbit/src/rabbit_peer_discovery.erl
```diff
@@ -82,6 +85,16 @@ node_type() ->
             ?DEFAULT_NODE_TYPE
     end.
 
+-spec registration() -> true | false.
```
dumbbell, line 88: You can use `boolean()` instead of `true | false`. That’s the builtin type spec for this.

deps/rabbit/priv/schema/rabbit.schema
michaelklishin, line 1376: The key used in `rabbit_peer_discovery` is `registration`, not `register`. I will correct this on your behalf.

deps/rabbit/src/rabbit_peer_discovery.erl
```diff
@@ -82,6 +85,16 @@ node_type() ->
...
+
+registration() ->
+    case application:get_env(rabbit, cluster_formation) of
+        {ok, Proplist} ->
+            proplists:get_value(registration, Proplist, ?DEFAULT_REGISTRATION);
```
dumbbell, line 93: If the value is set to something else that a list, this code will crash.

You can do something like:
```erlang
    case application:get_env(rabbit, cluster_formation) of
        {ok, Proplist} when is_list(Proplist) ->
            proplists:get_value(registration, Proplist, ?DEFAULT_REGISTRATION);
        {ok, Other} ->
            ?LOG_ERROR(
               "Peer discovery: ignoring invalid value for `register`: ~0p",
               [Other],
               #{domain => ?RMQLOG_DOMAIN_PEER_DISC}),
            ?DEFAULT_REGISTRATION;
```

dumbbell: Also, you use the name `registration` but the schema uses `rabbit.cluster_formation.register`. I think there is an inconsistency here.

deps/rabbit/src/rabbit_peer_discovery.erl
```diff
@@ -968,35 +981,51 @@ error_description({invalid_cluster_node_type, BadType}) ->
...
+                true ->
+                    ?LOG_DEBUG(
+                       "Peer discovery: registering this node",
+                       #{domain => ?RMQLOG_DOMAIN_PEER_DISC}),
+                    register(Backend),
+                    _ = Backend:post_registration(),
+                    ok;
+                false ->
+                    ?LOG_DEBUG(
+                       "Peer discovery: registration unsupported, skipping register",
```
dumbbell, line 997: You may need to split the string to fit in the 80 columns:
```erlang
                       "Peer discovery: registration unsupported, skipping "
                       "register",
```

michaelklishin (changes requested):
This PR as it is does not compile:

```
rabbit_peer_discovery.erl:1000:16: syntax error before: '.'
rabbit_peer_discovery.erl:1002:14: ?FUNCTION_NAME can only be used within a function
rabbit_peer_discovery.erl:1025:16: syntax error before: '.'
rabbit_peer_discovery.erl:1027:14: ?FUNCTION_NAME can only be used within a function
rabbit_peer_discovery.erl:19:2: function maybe_register/0 undefined
rabbit_peer_discovery.erl:19:2: function maybe_unregister/0 undefined
rabbit_peer_discovery.erl:981:2: spec for undefined function maybe_register/0
rabbit_peer_discovery.erl:1008:2: spec for undefined function maybe_unregister/0
rabbit_peer_discovery.erl:90:1: function registration/0 is unused
rabbit_peer_discovery.erl:1050:1: function register/1 is unused
rabbit_peer_discovery.erl:1071:1: function unregister/1 is unused
```

I will finish it and add a `rabbitmq.conf` schema test.

dumbbell (changes requested):
The overal patch looks good to me, thank you very much! That’s pretty cool for your first time with Erlang :-)

I made a few minor comments.

deps/rabbit/src/rabbit_peer_discovery.erl
```diff
@@ -968,35 +981,51 @@ error_description({invalid_cluster_node_type, BadType}) ->
...
+                       #{domain => ?RMQLOG_DOMAIN_PEER_DISC}),
+                    ok
+            end.
```
michaelklishin, line 1000: This is syntactically incorrect Erlang.

deps/rabbit/src/rabbit_peer_discovery.erl
```diff
@@ -968,35 +981,51 @@ error_description({invalid_cluster_node_type, BadType}) ->
...
+                    ?LOG_DEBUG(
+                       "Peer discovery: unregistering this node",
+                       #{domain => ?RMQLOG_DOMAIN_PEER_DISC}),
+                    unregister(Backend);
+                false ->
+                    ?LOG_DEBUG(
+                       "Peer discovery: registration unsupported, skipping unregister",
+                       #{domain => ?RMQLOG_DOMAIN_PEER_DISC}),
+                    ok
+            end.
```
michaelklishin, line 1025: Same as above, this is syntactically incorrect Erlang.

michaelklishin:
@dumbbell will do, I am finishing this PR.

dumbbell:
@frederikbosch: One more thing, could you please add a testcase?

michaelklishin:
@frederikbosch I have corrected the compilation issues, added a `rabbitmq.conf` schema test, renamed the setting and tested the changes against `etcd`.

The result is https://github.com/rabbitmq/rabbitmq-server/pull/13201.

I hope you don't mind us finishing the PR for you. You will be given full credit in release notes, I just wanted to speed up the process because we are working towards a round of (preview) releases for `4.1.0` and `4.0.x`.

frederikbosch:
@michaelklishin Not at all, please go ahead. No need for full credit, but a mention is nice. I am very happy with the feature as it will probably make the setup in my cluster easier. Thanks for your efforts.

michaelklishin:
@frederikbosch we cannot possibly not give you credit since you have come up with the idea and wrote at least some of the code :)

Let's continue in #13201.

PR #13219, opened 2025-02-08
Bump observer_cli to 1.8.2

zhongwencool:
https://github.com/zhongwencool/observer_cli/releases
```
1.8.2

Fix unit of fullsweep_after, The value of fullsweep_after is a number, not bytes.
1.8.1

Show node name in system pane.
1.8.0

Support <Pid to jump to specific pid.
Show process's label if it's set with [proc_lib:set_label(Label)](https://www.erlang.org/doc/apps/stdlib/proc_lib.html#set_label/1)
Show The number of bytes in the output distribution queue on System View. This queue sits between the Erlang code and the port driver, using undocumented functionerlang:dist_get_stat/1.
Fix Doc View not show when otp version = 27
```


michaelklishin:
@zhongwencool thank you!

zhongwencool:
Wow,  RabbitMQ's build and CI pipeline is lightning fast!

michaelklishin:
@zhongwencool they simply run in parallel on Actions (and can be run in parallel with CT). A few suites take close to 30 minutes.

PR #13328, opened 2025-02-13
Ra v2.16.2

kjnilsson:
This is a bugfix release of Ra:

* Fix last_index counter lag
* Fix off by one in follower assertion
* Fix log divergence bug


michaelklishin:
FTR, the same (recent) change in Ra [passed in the `2.15.2` PR](https://github.com/rabbitmq/rabbitmq-server/pull/13326) for `v4.0.x`.

PR #13336, opened 2025-02-13
rabbit_channel: Ignore DOWN message from monitored process if it exited normally

dumbbell:
## Why

It happens in CI from time to time and it was crashing the channel process. There is always a `channel.close` method pending in the channel mailbox.

## How

For now, log something and ignore the DOWN message. The channel will exit after handling the pending `channel.close` method anyway.

### DIFF
--- a/deps/rabbit/src/rabbit_channel.erl
+++ b/deps/rabbit/src/rabbit_channel.erl
@@ -36,6 +36,8 @@
 %% When a queue is declared as exclusive on a channel, the channel
 %% will notify queue collector of that queue.
 
+-include_lib("kernel/include/logger.hrl").
+
 -include_lib("rabbit_common/include/rabbit_framing.hrl").
 -include_lib("rabbit_common/include/rabbit.hrl").
 -include_lib("rabbit_common/include/rabbit_misc.hrl").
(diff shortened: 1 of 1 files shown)


ansd (approved):

deps/rabbit/src/rabbit_channel.erl
```diff
@@ -729,6 +731,10 @@ handle_info({{'DOWN', QName}, _MRef, process, QPid, Reason},
             handle_eol(QRef, State)
     end;
 
+handle_info({'DOWN', _MRef, process, Pid, normal}, State) ->
+    ?LOG_DEBUG("Process ~0p monitored by channel ~0p exited", [Pid, self()]),
```
lhoguin, line 735: As mentioned in chat this should include `process_info` information, as well as `process_info(Pid, current_stacktrace)` to figure out where the message is coming from.

dumbbell: The reason is also `normal`, so it didn’t crash.

lhoguin: No you're right if we have the DOWN message then we most likely won't get anything out of it. What would help is a test run that has tracing enabled but only for the spawn events, this way we have the MFA of all processes that start, and we can see which process it was that DOWN when we reproduce.

dumbbell: I never played with tracing. How should I enable it for the entire set of testsuites?

michaelklishin:

PR #13351, opened 2025-02-17
CLI: Don't use regex as module attributes

mkuratczyk:
This is not sufficient to build RabbitMQ on OTP 28.0-rc1 but it fixes the first problem that we hit. It might be that Elixir will support regex module attributes once it supports OTP 28, but we don't have to wait - this minor refactor just makes our code compatible with OTP28 without changing anything of substance.


michaelklishin (approved):

michaelklishin:
We might have to support Erlang 28 for `v4.0.x` during its "commercial support only" phase, so let's backport this further.

michaelklishin (approved):

PR #13360, opened 2025-02-18
Remove set_stream_retention_policy command

acogoluegnes:
It is not working as expected. Policies are the way to change data retention for stream.


michaelklishin:
If this is indeed to be backported to `v4.1.x`, we currently set the milestone to `4.1.0` right on the `main` PR.

`4.2.0` would make sense starting with an RC1 when we won't backport everything from `main` any more :)

PR #13363, opened 2025-02-18
Add dynamic socket buffer functionality to rabbit_reader

lhoguin:


lhoguin:
Applies to both old AMQP and AMQP 1.0.

ansd (approved):

PR #13374, opened 2025-02-20
Rabbitmq server 12743

michaelklishin:
This is #12743 #9620 by @noxdafox rebased and re-submitted by me.

### DIFF
--- a/deps/rabbit/src/rabbit_backing_queue.erl
+++ b/deps/rabbit/src/rabbit_backing_queue.erl
@@ -105,7 +105,7 @@
 
 %% Called to inform the BQ about messages which have reached the
 %% queue, but are not going to be further passed to BQ.
--callback discard(rabbit_types:msg_id(), pid(), state()) -> state().
+-callback discard(rabbit_types:basic_message(), pid(), state()) -> state().
 
 %% Return ids of messages which have been confirmed since the last
 %% invocation of this function (or initialisation).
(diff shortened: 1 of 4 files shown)


deps/rabbit/src/rabbit_backing_queue.erl
```diff
@@ -105,7 +105,7 @@
 
 %% Called to inform the BQ about messages which have reached the
 %% queue, but are not going to be further passed to BQ.
--callback discard(rabbit_types:msg_id(), pid(), state()) -> state().
+-callback discard(rabbit_types:basic_message(), pid(), state()) -> state().
```
kjnilsson, line 108: should this be `mc:state()` instead of `rabbit_types:basic_message()`?

michaelklishin: @noxdafox can you please see if you can refactor this to `mc:stat/0` and make it work for the deduplication plugin? `rabbit_types:basic_message/0` is AMQP 0-9-1-specific so we'd prefer to use the more generic `mc` module (which is AMQP 1.0-inspired but works for all protocols).

Thank you!

noxdafox: Will do. Will submit another PR.

noxdafox: Done in #13382

PR #13381, opened 2025-02-20
Run `rabbit_registry` boot step after `pre_boot`

### LINKED ISSUE #13352: Broker does not restart with binding to exclusive queue and Khepri activated
This has been detected by activating Khepri in the RabbitMQ AMQP 091 Java client test suite.

A test does the following:

```java
ch.exchangeDeclare(x, "fanout");
ch.queueDeclare(q, false, false, true, null);
ch.queueBind(q, x, "");
```

Then `stop_app / start_app` the broker (this is part of the connection recovery test). The broker fails to restart with the following error:

```
error:badarg

    ets:lookup/2
        args: [rabbit_registry,{exchange,fanout}]
    rabbit_registry:lookup_module/2, line 57
    rabbit_exchange:type_to_module/1, line 542
    rabbit_exchange:serialise_events/1, line 85
    rabbit_exchange:serial/1, line 90
    rabbit_binding:-process_deletions/1-fun-0-/2, line 494
    maps:foreach_1/3, line 391
    rabbit_amqqueue:notify_queue_binding_deletions/1, line 1968
```

Adding a 1-second delay between `stop_app` / `start_app` in the test does not help. Doing `start_app` manually after the test execution does work though.

Tested on 4.0.6 and `main`.

[Java test](https://github.com/rabbitmq/rabbitmq-java-client/blob/2bf9a9fa8407d8eda771221341c9bdc8f8d200ee/src/test/java/com/rabbitmq/client/test/functional/ConnectionRecovery.java#L370-L390)

the-mikedavis:
The `rabbit_registry` boot step starts up the `rabbit_registry` gen server from `rabbit_common`. This is a registry somewhat similar to the feature flag registry - it's meant to protect an ETS table used for looking up implementers of behaviors. The registry and its ETS table should be available as early as possible: the step should enable external_infrastructure rather than require it.

This is a quick fix for #13352. The crash in #13352 occurs because the `database` boot step runs before the `rabbit_registry` and Khepri deletes transient queues for the local node on during init (and that operation reads from the registry).

Fixes #13352


michaelklishin (approved):

michaelklishin:
Good catch!

michaelklishin:
This revealed a broken (or, well, 10 years out of date) boot step attributes in `rabbitmq_delayed_message_exchange` https://github.com/rabbitmq/rabbitmq-delayed-message-exchange/issues/298.

It was a good opportunity to revisit what boot steps does that plugin use.

PR #13409, opened 2025-02-24
amqp_client_SUITE: Fix frequent test failures

dumbbell:
* amqp_client_SUITE: Use a dedicated CI job for this testsuite
* amqp10_client: Handle `close` message in the `open_sent` state
* amqp10_client: Fix crash in close_sent
* amqp_client_SUITE: Retry connection in two testcases
* amqp_client_SUITE: Ensure `idle_time_out_on_server` restores heartbeat value
* amqp_client_SUITE: Use a dedicated AMQP-0-9-1 connection per testcase
* amqp_client_SUITE: Close all connections in `end_per_testcase/2`

### DIFF
--- a/deps/amqp10_client/src/amqp10_client_connection.erl
+++ b/deps/amqp10_client/src/amqp10_client_connection.erl
@@ -287,6 +287,22 @@ open_sent({call, From}, begin_session,
           #state{pending_session_reqs = PendingSessionReqs} = State) ->
     State1 = State#state{pending_session_reqs = [From | PendingSessionReqs]},
     {keep_state, State1};
+open_sent(_EvtType, {close, Reason}, State) ->
+    %% TODO: stop all sessions writing
+    %% We could still accept incoming frames (See: 2.4.6)
+    case send_close(State, Reason) of
+        ok ->
+            %% "After writing this frame the peer SHOULD continue to read from the connection
+            %% until it receives the partner's close frame (in order to guard against
+            %% erroneously or maliciously implemented partners, a peer SHOULD implement a
+            %% timeout to give its partner a reasonable time to receive and process the close
+            %% before giving up and simply closing the underlying transport mechanism)." [§2.4.3]
+            {next_state, close_sent, State, {state_timeout, ?TIMEOUT, received_no_close_frame}};
+        {error, closed} ->
+            {stop, normal, State};
+        Error ->
+            {stop, Error, State}
+    end;
 open_sent(info, {'DOWN', MRef, process, _, _},
           #state{reader_m_ref = MRef}) ->
     {stop, {shutdown, reader_down}}.
--- a/deps/amqp10_client/src/amqp10_client_connection.erl
+++ b/deps/amqp10_client/src/amqp10_client_connection.erl
@@ -288,8 +288,6 @@ open_sent({call, From}, begin_session,
     State1 = State#state{pending_session_reqs = [From | PendingSessionReqs]},
     {keep_state, State1};
 open_sent(_EvtType, {close, Reason}, State) ->
-    %% TODO: stop all sessions writing
-    %% We could still accept incoming frames (See: 2.4.6)
     case send_close(State, Reason) of
         ok ->
             %% "After writing this frame the peer SHOULD continue to read from the connection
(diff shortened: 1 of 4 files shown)


ansd:
I wonder whether instead of using `try ... after` in each test case, it would be better to force close any connections in `end_per_testcase`. This way the test cases are more readable. WDYT?

deps/amqp10_client/src/amqp10_client_connection.erl
```diff
@@ -287,6 +287,22 @@ open_sent({call, From}, begin_session,
...
+    %% TODO: stop all sessions writing
+    %% We could still accept incoming frames (See: 2.4.6)
+    case send_close(State, Reason) of
+        ok ->
+            %% "After writing this frame the peer SHOULD continue to read from the connection
+            %% until it receives the partner's close frame (in order to guard against
+            %% erroneously or maliciously implemented partners, a peer SHOULD implement a
+            %% timeout to give its partner a reasonable time to receive and process the close
+            %% before giving up and simply closing the underlying transport mechanism)." [§2.4.3]
+            {next_state, close_sent, State, {state_timeout, ?TIMEOUT, received_no_close_frame}};
```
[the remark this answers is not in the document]
dumbbell: I didn’t think a lot about this one. The client gets a `close` message while being in the `open_sent` state and was crashing with a function clause exception. I simply copy-pasted the [handling of that same message in the `opened` state](https://github.com/rabbitmq/rabbitmq-server/blob/5a467934e394a32d2ea23420aa8dea7984686af2/deps/amqp10_client/src/amqp10_client_connection.erl#L298-L313) :-)

dumbbell:
> I wonder whether instead of using `try ... after` in each test case, it would be better to force close any connections in `end_per_testcase`. This way the test cases are more readable. WDYT?

We discussed this with David outside of GitHub and I agreed this is a good solution.

deps/amqp10_client/src/amqp10_client_connection.erl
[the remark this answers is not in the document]
dumbbell: Should the client ignore all frames is `close_sent` other than the `close` one?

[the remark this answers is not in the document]
ansd: We could add another state `CLOSE_PIPE`, which might be an overkill.

I think it's best to explicitly ignore the `open` frame in `close_sent`. There is  no need ignore other frames as frames other than `open` and `close` are not handled by the connection process.

deps/amqp10_client/src/amqp10_client_connection.erl
```diff
@@ -287,6 +287,22 @@ open_sent({call, From}, begin_session,
           #state{pending_session_reqs = PendingSessionReqs} = State) ->
     State1 = State#state{pending_session_reqs = [From | PendingSessionReqs]},
     {keep_state, State1};
+open_sent(_EvtType, {close, Reason}, State) ->
+    %% TODO: stop all sessions writing
```
ansd, line 291: This comment is wrong because no sessions have begun at this point.

ansd: I pushed a commit which removes this comment.

deps/amqp10_client/src/amqp10_client_connection.erl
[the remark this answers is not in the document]
ansd: I pushed a commit which ignores the `open` frame.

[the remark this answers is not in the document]
dumbbell: Thank you!

deps/amqp10_client/src/amqp10_client_connection.erl
dumbbell: Thanks!

deps/rabbit/test/amqp_client_SUITE.erl
```diff
@@ -358,7 +359,11 @@ end_per_testcase(Testcase, Config) ->
     %% Assert that every testcase cleaned up.
     rabbit_ct_broker_helpers:rpc(Config, 0, ?MODULE, delete_queues, []),
     eventually(?_assertEqual([], rpc(Config, rabbit_amqqueue, list, []))),
-    %% Wait for sessions to terminate before starting the next test case.
+    %% Terminate all connections and wait for sessions to terminate before
+    %% starting the next test case.
+    _ = rabbit_ct_broker_helpers:rpc(
```
ansd, line 364: This function is already imported, so you can omit the module name.

deps/rabbit/test/amqp_client_SUITE.erl
```diff
@@ -358,7 +359,11 @@ end_per_testcase(Testcase, Config) ->
     %% Assert that every testcase cleaned up.
```
ansd, line 359: This comment can now be deleted. We don't assert anymore that testcases cleaned up. We just force cleanup from now on by force terminating any connection and sessions and force deleting any queues.

ansd (approved):

deps/rabbit/test/amqp_client_SUITE.erl
ansd: fixed

deps/rabbit/test/amqp_client_SUITE.erl
ansd: fixed

michaelklishin:

PR #13430, opened 2025-02-26
Fix CQ shared store files not deleted with large messages

lhoguin:
We must consider whether the previous current file is empty (has data written, but was already removed) when writing large messages and opening a file specifically for the large message. If we don't, then the file will never get deleted as we only consider files for deletion when a message gets removed (and there are none).

This is only an issue for large messages. Small messages write a message than roll over to a new file, so there is at least one valid message. Large messages close the current file first, regardless of there being a valid message.

@gomoripeti Please review.


gomoripeti:
great (looks reasonable to me) thank you for finding this subtle bug!

PR #13433, opened 2025-02-26
CLI: new health check that detects QQs without an elected leader

Ayanda-D:
## Proposed Changes

Hi Team 👋 

The following changes are for a CLI diagnostics tool used for checking the health of Quorum Queue leaders. The main use is to detect presence quorum queues which are in a bad state (e.g. in particular, leaderless, where no messaging can be done) as reported and discussed here: https://github.com/rabbitmq/rabbitmq-server/discussions/13101. There are some edge cases where quorum queues are observed to "lose leaders" and re-election is not triggered leaving such queues in bad state, and where any node restarts could result in complete message loss. This diagnostics tool allows us to carry out quick QQ leader health checks per vhost, or globally across all vhosts on a node/cluster for a specific queue match-spec. The command signature is as follows:

```
rabbitmq-diagnostics check_for_quorum_queues_without_an_elected_leader [--vhost <vhost>] [--across-all-vhosts] <pattern>
```

Output on a healthy node would be as follows:

```
rabbitmq-diagnostics check_for_quorum_queues_without_an_elected_leader --across-all-vhosts ".*"
Checking availability and health status of leaders for quorum queues matching .* in all vhosts ...
┌────────────────────────────────────────────────────────────────────────────────────────┐
│ Output                                                                                 │
├────────────────────────────────────────────────────────────────────────────────────────┤
│ Node rabbit@host-1 reported all quorum queue leaders as healthy                        │
└────────────────────────────────────────────────────────────────────────────────────────┘
```

Output on a node with unhealthy quorum queue leaders would be as follows (json formatted):

```
rabbitmq-diagnostics check_for_quorum_queues_without_an_elected_leader --across-all-vhosts ".*"  --formatter json
{"message":"Node rabbit@host-2 reported unhealthy quorum queue leaders","queues":[{"name":"TEST-QQ-1","readable_name":"queue 'TEST-QQ-1' in vhost 'VHOST-A'","type":"rabbit_quorum_queue","virtual_host":"VHOST-A"},{"name":"TEST-QQ-2","readable_name":"queue 'TEST-QQ-2' in vhost 'VHOST-B'","type":"rabbit_quorum_queue","virtual_host":"VHOST-B"}],"result":"error"}

```

In addition to the command implementation, few items to note:
- The command spawns and executes leader-health-check procedures concurrently per QQ and gathers results for fast execution
- Each leader-health-check procedure queries the leaderboard and executes a `ping` as an aliveness and health check per leader
- A `ProcessLimitThreshold` is set to 40% of the node's `process_limit` to ensure `process_count` when executing these checks will never reach the node's `process_limit` (and halt the node).
- The CLI output.ex is extended to support `:check_passed` results with a payload. This is simply to help for clarity for commands such as these in which the command check is a successful operation, while the result itself is an error (list of unhealthy QQs).

This diagnostics tool would help catch leaderless queues, which on the Management UI, are listed as follows:

LEADERLESS-QUORUM-QUEUE-1

Some environments are quite risky to carrying out maintenance operations such as complete node shutdown. The main goal is to help avoid situations where nodes/queues are assumed to be in a good state (where other checks such as quorum critical checks which rely on quorum node count pass) while QQ leaders may be in an unhealthy state. If nodes pass both these checks, then we can deem them healthy and OK for various maintenance procedures. Following on, on this detection tool, are investigations to also fix the underlying problem reported in https://github.com/rabbitmq/rabbitmq-server/discussions/13101. 

Please take a look/review. We're keen on having these capabilities available upstream to catch and monitor this recurring issue^ 👍 

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -2194,13 +2194,9 @@ run_leader_health_check(ClusterName, QResource, HealthCheckRef, From) ->
     end,
     ok.
 
-wait_for_leader_health_checks(_Ref, 0, UnhealthyAcc = []) -> UnhealthyAcc;
+wait_for_leader_health_checks(_Ref, 0, UnhealthyAcc) -> UnhealthyAcc;
 wait_for_leader_health_checks(Ref, N, UnhealthyAcc) ->
     receive
-        {ok, Ref, _QResource} when N == 1 ->
-            UnhealthyAcc;
-        {error, Ref, QResource} when N == 1 ->
-            [amqqueue:to_printable(QResource, ?MODULE) | UnhealthyAcc];
         {ok, Ref, _QResource} ->
             wait_for_leader_health_checks(Ref, N - 1, UnhealthyAcc);
         {error, Ref, QResource} ->
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -2154,12 +2154,22 @@ leader_health_check(QueueNameOrRegEx, VHost) ->
     %% we cannot spawn any new processes for executing QQ leader health checks.
     ProcessLimitThreshold = round(0.4 * erlang:system_info(process_limit)),
 
+    leader_health_check(QueueNameOrRegEx, VHost, ProcessLimitThreshold).
+
+leader_health_check(QueueNameOrRegEx, VHost, ProcessLimitThreshold) ->
+    Qs =
+        case VHost of
+            global ->
+                rabbit_amqqueue:list();
+            VHost when is_binary(VHost) ->
+                rabbit_amqqueue:list(VHost)
+        end,
     ParentPID = self(),
     HealthCheckRef = make_ref(),
     HealthCheckPids =
         lists:flatten(
             [begin
-                {resource, VHost, queue, QueueName} = QResource = amqqueue:get_name(Q),
+                {resource, _VHostN, queue, QueueName} = QResource = amqqueue:get_name(Q),
                 case check_process_limit_safety(ProcessLimitThreshold) of
                     true ->
                         case re:run(QueueName, QueueNameOrRegEx, [{capture, none}]) of
@@ -2173,7 +2183,7 @@ leader_health_check(QueueNameOrRegEx, VHost) ->
                         rabbit_log:warning("Leader health check failed from exceeded process limit threshold"),
                         throw({error, leader_health_check_process_limit_exceeded})
                 end
-            end || Q <- rabbit_amqqueue:list(VHost), amqqueue:get_type(Q) == ?MODULE]),
+            end || Q <- Qs, amqqueue:get_type(Q) == ?MODULE]),
     wait_for_leader_health_checks(HealthCheckRef, length(HealthCheckPids), []).
 
 run_leader_health_check(ClusterName, QResource, HealthCheckRef, From) ->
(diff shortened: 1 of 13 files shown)


michaelklishin (changes requested):
It is easy to see from

```shell
rabbitmq-diagnostics help | grep chec
```

that all existing health check commands except for the deprecated no-op one begin with `check_`, so this one should be named `check_for_quorum_queues_without_an_elected_leader`.

I'd rename `--global` to `--across-all-vhosts`.

michaelklishin (changes requested):

gomoripeti (commented):
great initiative!

What do you think if there would be a function `check_local_leader_health()` that would only check local leaders therefore it would avoid a lot of potential inter-node communication (a ping for each remote queue) and a wrapper function which calls `check_local_leader_health()` via rpc on every node? Similar to a common pattern in RabbitMQ

michaelklishin:
@gomoripeti we even have separate health checks, e.g. for local vs. cluster-wide alarms. 

I don't know if we need two separate health checks in this case. I'd try to get this command right first.

FTR, `rabbit_quorum_queue:list_with_local_promotable/0` demonstrates how to filter only the local QQs (that have a replica on `node()`).

Ayanda-D:
@gomoripeti @michaelklishin thanks for the feedback. For local or remote call optimization, `gen_statem` should already do that under-hood (if I'm understanding the question correctly). For this case though, large cluster sizes, purpose of this command is to list leaderless queues without any awareness of whether its local or remote, and only executed at specific times. Didnt see the need to filter locally or remote? Unless you see the need for this extension.

michaelklishin (changes requested):
```shell
cd deps/rabbit;
gmake dialyze
```

now fails with

```shell
rabbit_quorum_queue.erl:2189:5: Expression produces a value of type 
          {'error', _, _} | {'ok', _, _}, but this value is unmatched
```

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -2186,12 +2186,13 @@ leader_health_check(QueueNameOrRegEx, VHost, ProcessLimitThreshold) ->
 
 run_leader_health_check(ClusterName, QResource, HealthCheckRef, From) ->
     Leader = ra_leaderboard:lookup_leader(ClusterName),
-    case ra_server_proc:ping(Leader, ?LEADER_HEALTH_CHECK_TIMEOUT) of
-        {pong,leader} ->
-            From ! {ok, HealthCheckRef, QResource};
-        _ ->
-            From ! {error, HealthCheckRef, QResource}
-    end,
+    {_, _, _} =
```
michaelklishin, line 2189: Sounds like `_ = …` then would do?

Ayanda-D: yeah, dialyzer didnt complain for both. switching to `_` if thats preferred 👍

deps/rabbitmq_cli/lib/rabbitmq/cli/diagnostics/commands/check_for_quorum_queues_without_an_elected_leader_command.ex
```diff
@@ -0,0 +1,105 @@
...
+  end
+
+  def output({:error, unhealthy_queues}, %{silent: true}) when is_list(unhealthy_queues) do
+    {:ok, :check_passed}
+  end
+
+  def output({:error, unhealthy_queues}, %{vhost: _vhost}) when is_list(unhealthy_queues) do
+    lines = queue_lines(unhealthy_queues)
+
+    {:ok, :check_passed, Enum.join(lines, line_separator())}
```
michaelklishin, line 70: @Ayanda-D if this is an error, why does this return an `:ok` and `:check_passed`?

As things stand right now, this command will always exit with a code of zero. That's not at all the convention we have for all the other health checks.

It should have been

```elixir
{:error, :check_failed}
```

or one of its variations.

michaelklishin:
Moved to https://github.com/rabbitmq/rabbitmq-server/pull/13487.

PR #13435, opened 2025-02-27
Handle mc_amqp 3.13 `msg` record in 4.x

ansd:
The `msg` record was used in 3.13. This commit makes 4.x understand this record for backward compatibility, specifically for the rare case where:
1. a 3.13 node internally parsed a message from a stream via
```
Message = mc:init(mc_amqp, amqp10_framing:decode_bin(Bin), #{})
```
2. published this Message to a queue
3. RabbitMQ got upgraded to 4.x

(This commit can be reverted in some future RabbitMQ version once it's safe to assume that these upgraded messages have been consumed.)

The changes were manually tested as described in Jira RMQ-1525.

### DIFF
--- a/deps/rabbit/src/mc_amqp.erl
+++ b/deps/rabbit/src/mc_amqp.erl
@@ -50,6 +50,29 @@
                               Val :: term()}].
 -type opt(T) :: T | undefined.
 
+%% This representation was used in v3.13.7. 4.x understands this record for
+%% backward compatibility, specifically for the rare case where:
+%% 1. a 3.13 node internally parsed a message from a stream via
+%% ```
+%% Message = mc:init(mc_amqp, amqp10_framing:decode_bin(Bin), #{})
+%% ```
+%% 2. published this Message to a queue
+%% 3. RabbitMQ got upgraded to 4.x
+%%
+%% This record along with all its conversions in this module can therefore
+%% be deleted in some future RabbitMQ version once it's safe to assume that
+%% these upgraded messages have all been consumed.
+-record(msg,
+        {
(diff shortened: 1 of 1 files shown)


dcorbacho (approved):

deps/rabbit/src/mc_amqp.erl
```diff
@@ -50,6 +50,29 @@
...
+%% ```
+%% Message = mc:init(mc_amqp, amqp10_framing:decode_bin(Bin), #{})
+%% ```
+%% 2. published this Message to a queue
+%% 3. RabbitMQ got upgraded to 4.x
+%%
+%% This record along with all its conversions in this module can therefore
+%% be deleted in some future RabbitMQ version once it's safe to assume that
+%% these upgraded messages have all been consumed.
+-record(msg,
```
michaelklishin, line 65: Should we reflect the fact that this is an older/3.13 format? With versioning, similarly to what we do with `amqqueue`?

ansd: We can't. The name of this record must be the same as in 3.13 in https://github.com/rabbitmq/rabbitmq-server/blob/v3.13.7/deps/rabbit/src/mc_amqp.erl#L43

The `msg` record got enqueued into the quorum queue and the state machine crashed as follows:
```
exception error: no function clause matching
                 mc_amqp:size({msg,
                               {'v1_0.header',true,undefined,undefined,
                                undefined,undefined},
                               ...

```

michaelklishin: Right, good point.

michaelklishin (approved):

PR #13470, opened 2025-03-07
Fix test flakes in various testsuites

dumbbell:
[TBW]


deps/rabbitmq_mqtt/test/v5_SUITE.erl
```diff
@@ -206,10 +206,27 @@ end_per_testcase(T, Config) ->
...
+          rabbit_networking, close_all_connections, [<<"test finished">>]),
+    _ = rabbit_ct_broker_helpers:rpc_all(
+          Config,
+          rabbit_mqtt, close_local_client_connections, [normal]),
+    eventually(?_assertEqual(
+                  [],
+                  rpc(Config, rabbit_mqtt, local_connection_pids, []))),
     %% Assert that every testcase cleaned up their MQTT sessions.
+    rabbit_ct_broker_helpers:rpc(Config, 0, ?MODULE, delete_queues, []),
     eventually(?_assertEqual([], rpc(Config, rabbit_amqqueue, list, []))),
```
ansd, line 222: Should this line
```erl
eventually(?_assertEqual([], rpc(Config, rabbit_amqqueue, list, []))),
```
be deleted now that you force delete all queues? It seems redundant to first delete all queues, and then again list them to check that they are indeed deleted.

PR #13490, opened 2025-03-12
rabbit_peer_discovery: Compute start time once

dumbbell:
… and cache it.

## Why

It happens at least in CI that the computed start time varies by a few seconds. I think this comes from the Erlang time offset which might be adjusted over time.

This affects peer discovery's sorting of RabbitMQ nodes which uses that start time to determine the oldest node. When the start time of a node changes, it could be considered the seed node to join by some nodes but ignored by the other nodes, leading to troubles with cluster formation.


michaelklishin:

PR #13493, opened 2025-03-12
Add missing selenium function

MarcialRosales:
## Proposed Changes

Add. missing javascript function to selenium. And add a bash function that allows to initialize an idp without starting it.

## Types of Changes

- [] Bug fix (non-breaking change which fixes issue #NNNN)
- [X] Build system and/or CI

### DIFF
--- a/selenium/bin/suite_template
+++ b/selenium/bin/suite_template
@@ -544,8 +544,8 @@ determine_required_components_excluding_rabbitmq() {
 }
 initOnly() {
   for (( i=1; i<=$#; i++)) {
-    if [[ $i != "rabbitmq" ]]; then
-      eval val='$'$i
+    eval val='$'$i
+    if [[ $val != "rabbitmq" ]]; then  
       init="init_$val"
       $init
     fi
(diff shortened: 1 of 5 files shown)


selenium/bin/suite_template
```diff
@@ -536,6 +542,15 @@ determine_required_components_excluding_rabbitmq() {
     fi
   }
 }
+initOnly() {
+  for (( i=1; i<=$#; i++)) {
+    if [[ $i != "rabbitmq" ]]; then
+      eval val='$'$i
+      init="init_$val"
+      $init
+    fi
```
Zerpet, line 551: I'm not sure the `if` condition ever evaluates to false. `$i` expands to numbers e.g. 1, 2, 3, therefore, the comparison is always `<some-number> != "rabbitmq"`, which is always true. If I understand the intention correctly, you can rewrite this loop as:

```bash
for j in "$@"; do
  if [[ "$j" != "rabbitmq" ]]; then
    $j
  fi
done
```

MarcialRosales: You are dead right ! well spotted . Although no one will ever include rabbitmq in the `initOnly` function but it was not correct.

PR #13512, opened 2025-03-13
Change browser tab/window title according to currently loaded view (tab)

dcorbacho:
It is very hard now to distinguish different tabs. With this addition we have titles like 'RabbitMQ - Queue vhost/name', 'RabbitMQ - Exchanges'. To be continued...


michaelklishin (approved):
I have extended the same idea to a few more tabs (all top-level).

Will backport to `v4.1.x` manually due to the rebase.

PR #13175, opened 2025-01-29
Add force checkpoint functions for quorum queues and command line tool

aaron-seo:
## Proposed Changes

Addresses #13137.

Adds functions to force checkpoint for a quorum queue or for all quorum queues matching a `VhostSpec` and `QueueSpec`.

Also adds a CLI tool to invoke that function.

Hand-in-hand with doc change PR: https://github.com/rabbitmq/rabbitmq-website/pull/2175

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
If you're unsure about any of them, don't hesitate to ask on the mailing list.
We're here to help!
This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
- [x] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

## Further Comments

The test case for `force_checkpoint_on_queue` is a bit sensitive to changes in `rabbit_fifo:aux...` and `rabbit_fifo:checkpoint` records. Please feel free to suggest a different testing approach.

### Some local testing for CLI tool

`gmake run-broker RABBITMQ_NODENAME="force_checkpoint2"`

Create these queues and vhost:

```
aaronseo: ~/workplace/rabbitMq/rabbitmq-server/sbin $ ./rabbitmqctl -n force_checkpoint2 list_queues --vhost test name,type
Timeout: 60.0 seconds ...
Listing queues for vhost test ...
name    type
qwe     quorum
qq1     quorum
qq2     quorum
cq1     classic
aaronseo: ~/workplace/rabbitMq/rabbitmq-server/sbin $ ./rabbitmqctl -n force_checkpoint2 list_queues name,type
[...]

PR #13548, opened 2025-03-17
For 4.1.x, by @aaron-seo: introduce a command that would force QQs to take a checkpoint and truncate its segments

michaelklishin:
This is #13175 by @aaron-seo plus some changes from me that to address the QA feedback from @kjnilsson.

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -159,6 +159,7 @@
 -define(RPC_TIMEOUT, 1000).
 -define(START_CLUSTER_TIMEOUT, 5000).
 -define(START_CLUSTER_RPC_TIMEOUT, 60_000). %% needs to be longer than START_CLUSTER_TIMEOUT
+-define(FORCE_CHECKPOINT_RPC_TIMEOUT, 15_000).
 -define(TICK_INTERVAL, 5000). %% the ra server tick time
 -define(DELETE_TIMEOUT, 5000).
 -define(MEMBER_CHANGE_TIMEOUT, 20_000).
@@ -2126,7 +2127,7 @@ force_checkpoint_on_queue(QName) ->
         {ok, Q} when ?amqqueue_is_quorum(Q) ->
             {RaName, _} = amqqueue:get_pid(Q),
             rabbit_log:debug("Sending command to force ~ts to take a checkpoint", [QNameFmt]),
-            rpc:call(Node, ra, cast_aux_command, [{RaName, Node}, force_checkpoint]);
+            rpc:call(Node, ra, cast_aux_command, [{RaName, Node}, force_checkpoint], ?FORCE_CHECKPOINT_RPC_TIMEOUT);
         {ok, _Q} ->
             {error, not_quorum_queue};
         {error, _} = E ->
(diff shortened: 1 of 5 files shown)


michaelklishin:
@aaron-seo @kjnilsson note that due to how `rabbit_fifo:do_checkpoints/4` works, I had to adapt this test to publish more data (which is consistent with how we expect this new command to be used) and wait for a minimum amount of time between checkpoints.

Otherwise the checkpoint is not actually taken in the test, even if the `aux` state is updated.

@kjnilsson I have rolled back the `rabbit_fifo.hrl` change that wasn't necessary. Anything else that you'd like to see changed?

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -2089,6 +2092,38 @@ force_all_queues_shrink_member_to_current_member(ListQQFun) when is_function(Lis
...
+         case force_checkpoint_on_queue(QName) of
+             ok ->
+                 {QName, {ok}};
+             {error, Err} ->
+                 rabbit_log:warning("~ts: failed to force checkpoint, error: ~w",
+                                    [rabbit_misc:rs(QName), Err]),
+                 {QName, {error, Err}}
+         end
+     end
+     || Q <- rabbit_db_queue:get_all_durable_by_type(?MODULE),
```
kjnilsson, line 2123: this will include queues that do not have any members on the current node but the `force_checkpoint_on_queue` will only ask any local member to force a checkpoint.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -2089,6 +2092,38 @@ force_all_queues_shrink_member_to_current_member(ListQQFun) when is_function(Lis
...
+force_checkpoint_on_queue(QName) ->
+    Node = node(),
+    QNameFmt = rabbit_misc:rs(QName),
+    case rabbit_db_queue:get_durable(QName) of
+        {ok, Q} when ?amqqueue_is_classic(Q) ->
+            {error, classic_queue_not_supported};
+        {ok, Q} when ?amqqueue_is_quorum(Q) ->
+            {RaName, _} = amqqueue:get_pid(Q),
+            rabbit_log:debug("Sending command to force ~ts to take a checkpoint", [QNameFmt]),
+            rpc:call(Node, ra, cast_aux_command, [{RaName, Node}, force_checkpoint], ?FORCE_CHECKPOINT_RPC_TIMEOUT);
```
kjnilsson, line 2104: there is no need to do an rpc here as this function always addresses the quorum queue member on the node it runs on. is this the intent? to force a checkpoint on a _single_ member if there is one locally (there are no checks to ensure that).

deps/rabbitmq_cli/lib/rabbitmq/cli/queues/commands/force_checkpoint_command.ex
```diff
@@ -0,0 +1,88 @@
...
+  def usage_doc_guides() do
+    [
+      DocGuide.quorum_queues()
+    ]
+  end
+
+  def help_section, do: :replication
+
+  def description,
+    do: "Forces checkpoints for all matching quorum queues"
```
kjnilsson, line 83: no it will only force a checkpoint on the local member of the matching queues, if there is a local member.

kjnilsson (changes requested):
Should this functionality force a checkpoint on a given member or all members of a queue? I'm not sure it works correctly in all cases but am somewhat unsure of the goal of this new command.

michaelklishin:
When it comes to triggering a checkpoint for all replicas on the given node vs. all replicas, period, I guess the right thing to do is to do it on all replicas.

I'll look into it.

michaelklishin:
We have agreed to finish this most likely after `4.1.0` ships. We are still interested in this feature but it's not a blocker for `4.1.0`.

Issue #13541, opened 2025-03-15
AMQP 0-9-1 connections: accommodate longer JWT tokens


michaelklishin:
### Is your feature request related to a problem? Please describe.

`connection.start-ok`'s SASL `response` field was not originally designed for JWT
tokens, which can be longer than, say, a typical string length of 256 bytes.

While it's not common to see JWT tokens that exceed the length limit and
force the AMQP 0-9-1 parser to consider the frame to be too long,
it would be nice to see if the limit can be increased to, say, 8192 bytes in a backwards-compatible way.

With many HTTP servers, the limit is that of a header length, which seems to be around 8192 bytes, for example, that's the value mentioned in [`large_client_header_buffers` documentation in Nginx](https://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers).

See https://github.com/rabbitmq/rabbitmq-server/discussions/13537 for one example.

### Describe the solution you'd like

If feasible in terms of backwards compatibility, AMQP 0-9-1's `connection.start-ok`'s `response` field should allow for longer payloads.

### Describe alternatives you've considered

The only alternative seems to be adding an early validation of the password/secret field length in client libraries, plus an extension to the [OAuth 2 Troubleshooting guide](https://www.rabbitmq.com/docs/troubleshooting-oauth2): https://github.com/rabbitmq/rabbitmq-website/issues/2212.

michaelklishin:
My findings suggest that this is a matter of the negotiated maximum connection frame size, which with many clients is 128 KiB already:

1. https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/Makefile#L23
2. https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_reader.erl#L1093-L1099

The `rabbitmq_codegen`-generated parser does not place a limit on the `response` field specifically:

```erl
decode_method_fields('connection.start_ok', <<F0Len:32/unsigned, F0Tab:F0Len/binary, F1Len:8/unsigned, F1:F1Len/binary, F2Len:32/unsigned, F2:F2Len/binary, F3Len:8/unsigned, F3:F3Len/binary>>) ->
  F0 = rabbit_binary_parser:parse_table(F0Tab),
  rabbit_binary_parser:assert_utf8(F1),
  rabbit_binary_parser:assert_utf8(F3),
[...]

PR #13549, opened 2025-03-17
Increase FRAME-MIN-SIZE in AMQP 0-9-1 code generation file

acogoluegnes:
References #13541


michaelklishin:
Oh, it was a constant produced by codegen. Thank you very much, @acogoluegnes!

PR #13573, opened 2025-03-19
Allow the rabbitmq_auth_backend_http plug-in to be used only for user authentication and authorization against the HTTP backend by setting vhost_path, resource_path, and topic_path to 'disabled'

mqhase:
## Proposed Changes

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

- [] All tests pass locally with my changes

## Further Comments

### DIFF
--- a/deps/rabbitmq_auth_backend_http/README.md
+++ b/deps/rabbitmq_auth_backend_http/README.md
@@ -93,7 +93,7 @@ For instance, if the user accessed RabbitMQ via the MQTT protocol, it is expecte
 * `vhost`: the name of the virtual host being accessed
 * `ip`: the client ip address
 
-Note that you cannot create arbitrary virtual hosts using this plugin; you can only determine whether your users can see / access the ones that exist. By setting `vhost_path` to `disabled`, this authentication request is skipped, and  unrestricted access is granted.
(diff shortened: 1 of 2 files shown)


mqhase:
@lukebakken - In our production environment we are using the rabbitmq_auth_backend_http plug-in mainly for user authentication against a custom, central HTTP backend. Once the user is authenticated and authorized, there is theoretically no further need to e.g. authenticate/authorize the bindings and publish requests. Until now we solve(d) this situation by deploying httpd on the same machine returning as static file with content "allow".

But just recently a feature was introduced, which relies on highly per-mutating routing keys and we discovered that  rabbitmq_auth_backend_cache isn't of much help in this case, thus we ended up in a situation where or each and every message published resulted in a HTTP request (only to return static allow). This additional HTTP request caused that throughput decreased due to flow control and/or added some additional latency.
And the second motivation is that the need for a local httpd instance (even if only serving static content) introduces another point of failure we'd like to avoid.

Therefore the most performant and reliable solution, from our point of view, is to granularly disable certain backend calls (e.g. only by setting  topic_path to disabled) requiring only allow/deny (being fully aware that it's not applicable for  user_path as the tag is required).

michaelklishin:

This feature repeats something that's already possible with very little effort. Your backend can return `allow` for the endpoints you don't want to implement. That's what our example apps do, in fact.

If the intent is to cut down the upstream HTTP service traffic, this plugin should only be used with [`rabbitmq_auth_backend_cache`](https://www.rabbitmq.com/docs/auth-cache-backend).

PR #13587, opened 2025-03-21
QQ: Revise checkpointing logic to take more frequent checkpoints for large message workloads

kjnilsson:
Lower the `min_checkpoint_interval` substantially to allow quorum queues better control over when checkpoints are taken.

Tracking message bytes written to the log and use this to request a checkpoint every 64MB if no other checkpoint condition was met. This ensures that queues where the messages are very large (1MB+) are checkpointed based on their data ingress rather than indexes.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -932,7 +932,7 @@ which_module(5) -> ?MODULE.
                      smallest_index :: undefined | ra:index(),
                      messages_total :: non_neg_integer(),
                      indexes = ?CHECK_MIN_INDEXES :: non_neg_integer(),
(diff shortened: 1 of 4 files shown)


kjnilsson:
Testing:

Try various workloads with both small and large messages and monitor the disk to ensure that there is no undue segment buildup that is retained when workload is paused.

Also ensure segments are cleared up reasonably after queue is emptied / purged.

mkuratczyk:
We've run lots of tests with different variations of the tweaks. This clearly improves the situation with many slow queues, especially as messages get larger. In this test, all queues are empty (publishers are slow and consumers are present so they immediately get the messages). We can see that `main` uses more and more disk space, while this branch does not.
Screenshot 2025-03-26 at 11 29 29

PR #13604, opened 2025-03-24
Switch is_<queue_type> to using queue.type field in Management UI

ikavgo:
x-queue-type is not mandatory. Using queue.type is more robust. 

See also https://github.com/rabbitmq/rabbitmq-server/pull/13603


michaelklishin:
This looks good at first glance. I will give it a shot later today.

michaelklishin (approved):

PR #13625, opened 2025-03-26
Fix crash when consuming from unavailable quorum queue

ansd:
Prior to this commit, when a client consumed from an unavailable quorum queue, the following crash occurred:
```
{badmatch,{error,noproc}}
[{rabbit_quorum_queue,consume,3,[{file,\"rabbit_quorum_queue.erl\"},{line,993}]}
```

This commit fixes this bug by returning any error when registering a quorum queue consumer to rabbit_queue_type.

This commit also refactors errors returned by
rabbit_queue_type:consume/3 to simplify and ensure seperation of concerns.

For example prior to this commit, the channel did error formatting specifically for consuming from streams. It's better if the channel is unaware of what queue type it consumes from and have each queue type implementation format their own errors.

Jira: RMQ-1582

### DIFF
--- a/deps/rabbit/src/rabbit_queue_type.erl
+++ b/deps/rabbit/src/rabbit_queue_type.erl
@@ -522,7 +522,7 @@ consume(Q, Spec, State) ->
     case Mod:consume(Q, Spec, CtxState0) of
         {ok, CtxState} ->
             {ok, set_ctx(Q, Ctx#ctx{state = CtxState}, State)};
-        Err = {error, _Type, _Fmt, _FmtArgs} ->
+        {error, _Type, _Fmt, _FmtArgs} = Err->
             Err
     end.
 
--- a/deps/rabbit/src/rabbit_queue_type.erl
+++ b/deps/rabbit/src/rabbit_queue_type.erl
@@ -211,8 +211,7 @@
                   consume_spec(),
                   queue_state()) ->
     {ok, queue_state(), actions()} |
-    {error, term()} |
-    {protocol_error, Type :: atom(), Reason :: string(), Args :: term()}.
+    {error, Type :: atom(), Format :: string(), FormatArgs :: [term()]}.
 
 -callback cancel(amqqueue:amqqueue(),
                  cancel_spec(),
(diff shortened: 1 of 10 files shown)


deps/rabbit/test/amqp_client_SUITE.erl
```diff
@@ -6596,8 +6597,55 @@ bad_x_cc_annotation_exchange(Config) ->
...
+                        Session0, <<"receiver">>, Address),
+    receive
+        {amqp10_event,
+         {session, Session0,
+          {ended,
+           #'v1_0.error'{
+              condition = ?V_1_0_AMQP_ERROR_INTERNAL_ERROR,
+              description = {utf8, Desc}}}}} ->
+            ?assertEqual(
+               <<"failed consuming from quorum queue 'q-down' in vhost '/': noproc">>,
```
michaelklishin, line 6632: Should we replace the "noproc" with something like "the leader replica was unavailable or not elected"?

ansd: Ideally yes. However, `noproc` is an exit reason returned from https://www.erlang.org/doc/apps/stdlib/gen_statem.html#call/3. That function can return other exit reasons too. I'd like to avoid parsing and translating any possible exit reason from `gen_statem:call/3`. The alternative would be to log this Erlang term only in the server and not pass it down to the client app.

FWIW, I removed the `noproc` from the test expectation.

deps/rabbit/src/rabbit_queue_type.erl
```diff
@@ -516,15 +515,14 @@ new(Q, State) when ?is_amqqueue(Q) ->
...
-    {protocol_error, Type :: atom(), Reason :: string(), Args :: term()}.
+    {error, Type :: atom(), Format :: string(), FormatArgs :: [term()]}.
 consume(Q, Spec, State) ->
     #ctx{state = CtxState0} = Ctx = get_ctx(Q, State),
     Mod = amqqueue:get_type(Q),
     case Mod:consume(Q, Spec, CtxState0) of
         {ok, CtxState} ->
             {ok, set_ctx(Q, Ctx#ctx{state = CtxState}, State)};
-        Err ->
+        Err = {error, _Type, _Fmt, _FmtArgs} ->
```
kjnilsson, line 525: the assingment should got to the right of the pattern match.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1008,46 +1010,58 @@ consume(Q, Spec, QState0) when ?amqqueue_is_quorum(Q) ->
...
-            %% get the leader from state
-            case rabbit_fifo_client:query_single_active_consumer(QState) of
-                {ok, SacResult} ->
-                    ActivityStatus = case SacResult of
-                                         {value, {ConsumerTag, ChPid}} ->
-                                             single_active;
-                                         _ ->
-                                             waiting
-                                     end,
+    case rabbit_fifo_client:checkout(
```
kjnilsson, line 1013: Try to favour left to right formatting styles as used in this module. so arguments should as far as possible go to the right of the function name.

kjnilsson (approved):
This looks good to me. Just a bit of formatting feedback.

PR #13641, opened 2025-03-27
Incoming message interceptors

LoisSotoLopez:
## Proposed Changes

This PR enables users to provide custom message interceptor modules,
i.e. modules to process incoming and outgoing messages. The
`rabbit_message_interceptor` behaviour defines a `intercept/4` callback,
for those modules to implement.

```erlang
-callback intercept(Msg, MsgInterceptorCtx, Group, Config) -> Msg when
    Msg :: mc:state(),
    MsgInterceptorCtx :: msg_interceptor_ctx(),
    Group :: incoming_message_interceptors | outgoing_message_interceptors,
    Config :: #{atom() := term()}.
```

The `intercept/4` callback receives a `Msg`; a context map provided by
the protocol-specific module that invoked the interception; a `Group`
specifying whether `Msg` is a incoming or a outgoing message; and a
configuration map specific to the interceptor.

Some of those protocol-specific modules that invoke the interception and
provide the context map are `rabbit_channel`, `rabbit_amqp_session` and
`rabbit_mqtt_processor`. The `msg_interceptor_ctx()` type makes
mandatory to provide certain information on the context map, such as the
protocol used to send/get the message, the virtual host where it
arrived/departed or the username publishing/getting the message. Other
pieces of information such as the name of the connection used to
publish/get the message may be added to the context map but are not
mandatory, so users providing custom interceptors should be aware of
this when implementing such modules.

For an interceptor to be invoked it needs to be properly configured
through the `.conf` file.

For just enabling an interceptor it would be enough to add a single line
to the `.conf` file. For an interceptor module named `foo_mod.erl` that
line would be:

```
message_interceptors.incoming.foo_mod.enabled=true
```

Notice this configuration line only enables the module as an interceptor
for incoming messages. To also make the same module intercept outgoing
messages two lines are needed.

```
message_interceptors.incoming.foo_mod.enabled=true
message_interceptors.outgoing.foo_mod.enabled=true
```

Thid last two-line configuration will generate an Erlang configuration
like this:

```erlang
[{rabbit, [{message_interceptors,
            #{incoming => [{foo_mod, #{}}],
              outgoing => [{foo_mod, #{}}]}},
           ...]},
  ...].
```

The empty map after `foo_mod` is the configuration map that will be
received by the `invoke/4` function as fourth argument. To populate this
map users can add more lines to the `.config` file. For example:

```
message_interceptors.incoming.foo_mod.enabled=true
message_interceptors.incoming.foo_mod.delay=100
message_interceptors.incoming.foo_mod.period=3000
message_interceptors.outgoing.foo_mod.delay=200
message_interceptors.outgoing.foo_mod.mode=idle
```

would result on a config like the following.

```erlang
[{rabbit, [{message_interceptors,
            #{incoming => [{foo_mod, #{delay => 100, period => 3000}}],
              outgoing => [{foo_mod, #{delay => 200, mode => idle}}]}},
           ...]},
  ...].
```

The configuration map can be different for the same interceptor used as
incoming or outgoing interceptor. Also, the `enabled` key never gets
added to the resulting configuration map. In fact, if other keys are
used for the same group (incoming or outgoing) you don't need to specify
the `enabled` key. Therefore the configuration above could be simplified
to:

```
message_interceptors.incoming.foo_mod.delay=100
message_interceptors.incoming.foo_mod.period=3000
message_interceptors.outgoing.foo_mod.delay=200
message_interceptors.outgoing.foo_mod.mode=idle
```

Users providing custom interceptors should also provide the Cuttlefish
schemas providing _mappings_ for the configuration parameters of their
interceptors. This commit already provides a Cuttlefish _translation_
for those parameters. Therefore, those user-provided mappings should
keep the expected format:

```erlang
{mapping,
 "message_interceptor.<incoming|outgoing>.<interceptor_module_name>.<config_key>",
 "rabbit.<incoming|outgoing>_message_interceptors",
 <mapping_options>}.
```

For compatibility reasons, an alias can be used in
`<interceptor_module_name>` instead of the actual module name, but only
for the interceptors that already existed in RabbitMQ before this
commit. The `set_header_timestamp` and `set_header_routing_node` aliases
can be used as `<interceptor_module_name>`. However, using the actual
module names, wich are `rabbit_message_interceptor_timestamp` and
`rabbit_message_interceptor_routing_node` respectively is recommended.

The order in which the interceptors are invoked is the same as the order
in which they appear in the `.conf` file. For instance, the following
config lines

```
message_interceptors.outgoing.mod1.enabled=true
message_interceptors.incoming.mod3.config1=1
message_interceptors.incoming.mod1.config1=1
message_interceptors.incoming.mod2.config2=2
```

would result in the following Erlang configuration

```
[{rabbit, [{message_interceptors,
            #{incoming => [{mod3, _Cfg1}, {mod1, _Cfg1}, {mod2, _Cfg2}],
              outgoing => [{mod1, _Cfg4}]}}
           ...]},
  ...].
```

and the following order of invocation for incoming messages:
`mod3 (first) -> mod1 -> mod2 (last)`.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

- [x] All tests pass locally with my changes.


deps/rabbit/src/rabbit_incoming_message_interceptor.erl
```diff
@@ -0,0 +1,38 @@
+-module(rabbit_incoming_message_interceptor).
```
kjnilsson, line 1: i don't think we need to distinguish incoming and outgoing interceptors are the module / behaviour level.
We should have a single behaviour `rabbit_message_interceptor` with a single `intercept` function that takes the message container, the protocol context map (see another comment) and the config.

Instead they would be distinguined by which configuration they are added to. It is fine if this PR only implements incoming interceptors.

LoisSotoLopez:
> I think the behaviour can be reduced to a single behaviour for both incoming and outgoing interceptors and the protocol behaviour can be replaced by a context map.

Sounds good to me. The context map could be set on initialization by each _protocol accessor_ (AMQP 0.9.1 channel, AMQP 1.0 session, or MQTT connection process), and then passed to `rabbit_incoming_message_interceptor` (or `rabbit_message_interceptor`). If we go for the `rabbit_message_interceptor` (both incoming and outgoing messages) a keyword `incoming | outgoing` should be passed along the context map so that the right list of interceptors gets invoked.

Any additional comments on not using the _getters_ approach and going for both a incoming+outgoing approach, @ansd ?

deps/rabbit/src/rabbit_message_interceptor.erl
```diff
@@ -1,50 +1,43 @@
...
+    MsgInterceptorCtx :: map(),
+    MsgDirection:: incoming | outgoing,
+    Resp :: mc:state().
+intercept(Msg, MsgInterceptorCtx, MsgDirection) ->
+    InterceptorsList = list_to_atom(atom_to_list(MsgDirection)
+                                    ++ "_message_interceptors"),
+    Interceptors = persistent_term:get(InterceptorsList, []),
+    lists:foldl(fun({Module, Config}, Msg0) ->
+        try
+            Module:intercept(Msg0, MsgInterceptorCtx, Config)
```
kjnilsson, line 25: I feel the direction should be passed to the interceptor, you may want to just have a single interceptor for both incoming and outgoing.

deps/rabbit/src/rabbit_message_interceptor.erl
```diff
@@ -1,50 +1,43 @@
...
-%% now be configured to add such headers. This enables non-AMQP 0.9.1 protocols (that
-%% do not use rabbit_channel) to also add AMQP 0.9.1 headers to incoming messages.
 -module(rabbit_message_interceptor).
--include("mc.hrl").
-
--export([intercept/1]).
 
--define(HEADER_TIMESTAMP, <<"timestamp_in_ms">>).
--define(HEADER_ROUTING_NODE, <<"x-routed-by">>).
+-export([
```
kjnilsson, line 3: format as:

```
-export([intercept/3,
         set_msg_annotation/4]).

```

deps/rabbit/src/rabbit_message_interceptor.erl
```diff
@@ -1,50 +1,43 @@
...
 
--spec intercept(mc:state()) -> mc:state().
-intercept(Msg) ->
-    Interceptors = persistent_term:get(incoming_message_interceptors, []),
-    lists:foldl(fun({InterceptorName, Overwrite}, M) ->
-                        intercept(M, InterceptorName, Overwrite)
-                end, Msg, Interceptors).
+-callback intercept(
+    Msg :: mc:state(),
+    MsgInterceptorCtx :: map(),
```
kjnilsson, line 10: Can we define some context properties that we'd always pass, e.g. `vhost`, `username` and the current protocol itself?

e.g.

```
#{protocol := amqp091 | amqp | mqttv3 | mqttv5, %% tags to be defined
   vhost := binary(),
   username := binary(),
   conn_name => binary() % optional
   atom() -> term()}
```

deps/rabbit/src/rabbit_message_interceptor.erl
```diff
@@ -1,50 +1,43 @@
...
+    set_msg_annotation/4
+]).
 
--spec intercept(mc:state()) -> mc:state().
-intercept(Msg) ->
-    Interceptors = persistent_term:get(incoming_message_interceptors, []),
-    lists:foldl(fun({InterceptorName, Overwrite}, M) ->
-                        intercept(M, InterceptorName, Overwrite)
-                end, Msg, Interceptors).
+-callback intercept(
```
kjnilsson, line 8: formatting is too squished here, try:

```
-callback intercept(Msg :: mc:state(),
                    MsgInterceptorCtx :: map(),
                    Config :: #{atom() := term()}) -> mc:state().

```

deps/rabbit/src/rabbit_message_interceptor.erl
```diff
@@ -1,50 +1,43 @@
...
+                                    ++ "_message_interceptors"),
+    Interceptors = persistent_term:get(InterceptorsList, []),
+    lists:foldl(fun({Module, Config}, Msg0) ->
+        try
+            Module:intercept(Msg0, MsgInterceptorCtx, Config)
+        catch
+            error:undef ->
+                Msg0
+        end
+    end, Msg , Interceptors).
```
kjnilsson, line 30: `end` should line up with it's corresponding `fun`

deps/rabbitmq_mqtt/src/rabbit_mqtt_client_id_interceptor.erl
```diff
@@ -0,0 +1,18 @@
...
+
+-behaviour(rabbit_message_interceptor).
+
+-define(ANN_CLIENT_ID, <<"client_id">>).
+
+-export([
+    intercept/3
+]).
+
+intercept(Msg, #{client_id := ClientId}, _Config) ->
```
kjnilsson, line 11: a more left to right formatting fits well within 80 chars:

e.g.
```
intercept(Msg, #{client_id := ClientId}, _Config) ->
    rabbit_message_interceptor:set_msg_annotation(Msg, ?ANN_CLIENT_ID,
                                                  ClientId, true);
intercept(Msg, _MsgInterceptorCtx, _Config) ->
    Msg.

```

deps/rabbit/src/rabbit_channel.erl
```diff
@@ -509,7 +510,10 @@ init([Channel, ReaderPid, WriterPid, ConnPid, ConnName, Protocol, User, VHost,
                             consumer_timeout = ConsumerTimeout,
                             authz_context = OptionalVariables,
                             max_consumers = MaxConsumers,
-                            writer_gc_threshold = GCThreshold
+                            writer_gc_threshold = GCThreshold,
+                            msg_interceptor_ctx = #{user => User,
+                                                   vhost => VHost,
```
kjnilsson, line 515: indentation is slightly off

deps/rabbit/src/rabbit_header_routing_node_interceptor.erl
```diff
@@ -0,0 +1,17 @@
+-module(rabbit_header_routing_node_interceptor).
+-behaviour(rabbit_message_interceptor).
+
+-define(HEADER_ROUTING_NODE, <<"x-routed-by">>).
+
+-export([
```
kjnilsson, line 6: reformat as per prev comment

kjnilsson (changes requested):
we're getting closer I think a few suggested changes and some formatting advise.

LoisSotoLopez:
@kjnilsson Thanks on the formatting advices :) . I'm using neovim and I've tried to find out a proper solution on this, but I haven't found yet a solution that would allow me to format my contributions w/out re-formatting previous code. Will try to figure this out.

deps/rabbit/src/rabbit_message_interceptor.erl
LoisSotoLopez: I assumed an interceptor would do the same independently of the direction of the message. It is the user the one that decides (if the interceptor allows it by providing the appropriate `message_interceptors.<direction>.` cuttlefish mapping) in which direction should the interceptor be applied, or if it should be applied in both.

deps/rabbit/src/rabbit_message_interceptor.erl
```diff
@@ -1,50 +1,43 @@
...
-    Ts = mc:get_annotation(?ANN_RECEIVED_AT_TIMESTAMP, Msg0),
-    Msg = set_annotation(Msg0, ?HEADER_TIMESTAMP, Ts, Overwrite),
-    set_timestamp(Msg, Ts, Overwrite).
+-spec intercept(Msg, MsgInterceptorCtx, MsgDirection) -> Resp when
+    Msg :: mc:state(),
+    MsgInterceptorCtx :: map(),
+    MsgDirection:: incoming | outgoing,
+    Resp :: mc:state().
+intercept(Msg, MsgInterceptorCtx, MsgDirection) ->
+    InterceptorsList = list_to_atom(atom_to_list(MsgDirection)
```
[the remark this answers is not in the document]
LoisSotoLopez: This needs for ad hoc code to be put in [rabbit:persist_static_configuration](https://github.com/rabbitmq/rabbitmq-server/blob/dc5a703c237265baaa6ce0b67033d58c0c4d4665/deps/rabbit/src/rabbit.erl#L1654)

```erl
    persist_static_configuration(
      [classic_queue_index_v2_segment_entry_count,
       classic_queue_store_v2_max_cache_size,
       classic_queue_store_v2_check_crc32,
       incoming_message_interceptors,
       outgoing_message_interceptors
      ]),
```
needs to become something like
```erl
    persist_static_configuration(
      [classic_queue_index_v2_segment_entry_count,
       classic_queue_store_v2_max_cache_size,
       classic_queue_store_v2_check_crc32,
       {incoming_message_interceptors, {message_interceptors, incoming}
       outgoing_message_interceptors, {message_interceptors, outgoing}
      ]),
```

and

```erl
persist_static_configuration(Params) ->
    lists:foreach(
      fun(Param) ->
              case application:get_env(?MODULE, Param) of
                  {ok, Value} ->
                      ok = persistent_term:put(Param, Value);
                  undefined ->
                      ok
              end
      end, Params).
```

needs to become

```erl
persist_static_configuration(Params) ->
    lists:foreach(
        fun({Param, Alias}) ->
            case application:get_env(?MODULE, Param) of
                {ok, Value} ->
                    ok = persistent_term:put(Alias, Value);
                undefined ->
                    ok
            end;
           (Param) ->
            case application:get_env(?MODULE, Param) of
                {ok, Value} ->
                    ok = persistent_term:put(Param, Value);
                undefined ->
                    ok
            end
        end, Params).

```

Is that fine? What about this other option:

The cuttlefish translations could generate a `rabbit.message_interceptors.<incoming|outgoing>` config instead of `rabbit.<incoming_message_interceptors|outgoing_message_interceptors>` and the `rabbit:persist_static_configuration` would only receive `message_interceptors` instead of both `<incoming|outgoing>_message_interceptors`. The interceptor module would read app conf. with `message_interceptors` and then then get either the incoming or the outgoing ones from the read term. 

The app conf would look like
```
[{rabbit, [{message_interceptors, #{incoming => [...], outgoing => [...] ...
```
instead of
```
[{rabbit, [...
    {incoming_message_interceptors, ...},
    {outgoing_message_interceptors,...},
...
```

deps/rabbit/src/rabbit_message_interceptor.erl
LoisSotoLopez: As suggested by @gomoripeti , the `rabbit_message_interceptor` module could also add the message direction in the `MsgInterceptorCtx` map. How does that sound?

kjnilsson: that sounds good.

deps/rabbit/src/rabbit_message_interceptor.erl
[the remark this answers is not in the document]
kjnilsson: I see. that seems complicated. Perhaps just resolve `incoming` to `incoming_message_interceptors` etc in a function rather than doing atom to list conversions.

```
resolve(incoming) ->
   incoming_message_interceptors;
resolve(outgoing) ->
   outgoing_message_interceptors.
```

kjnilsson:
> @kjnilsson Thanks on the formatting advices :) . I'm using neovim and I've tried to find out a proper solution on this, but I haven't found yet a solution that would allow me to format my contributions w/out re-formatting previous code. Will try to figure this out.

It is an art rather than a science really. I use nvim too and it is fine for indentation - just hit '=' for your selection and it mostly does the right thing.

deps/rabbit/src/rabbit_message_interceptor.erl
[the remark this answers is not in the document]
ansd: Yes, calling `atom_to_list` for each message that RabbitMQ receives creates lots of garbage, I wouldn't be surprised if performance drops by a few percent.

What Karl suggested is great.

Alternatively, it might be simpler to omit these translations entirely. For example instead of
```erl
-spec intercept(Msg, MsgInterceptorCtx, MsgDirection) -> Resp when
    Msg :: mc:state(),
    MsgInterceptorCtx :: map(),
    MsgDirection:: incoming | outgoing,
    Resp :: mc:state().
```
we could have
```erl
-spec intercept(Msg, Context, Group) -> Msg when
    Msg :: mc:state(),
    Context :: map(),
    Group :: incoming_message_interceptors | outgoing_message_interceptors.
```

kjnilsson:
> > he only downside of the context map I see is that we duplicate some state information, for example each MQTT connection has higher memory usage after this PR. However, this context map is cleaner and simpler, so I prefer the context map.
> 
> MQTT could add the context map to a persistent term to avoid this if necessary

no crap that, the context map is per vhost / user.

kjnilsson:
> The only downside of the context map I see is that we duplicate some state information, for example each MQTT connection has higher memory usage after this PR. However, this context map is cleaner and simpler, so I prefer the context map.

MQTT could consider building it on the fly to keep memory overhead low

deps/rabbit/src/rabbit_header_routing_node_interceptor.erl
```diff
@@ -0,0 +1,17 @@
+-module(rabbit_header_routing_node_interceptor).
```
ansd, line 1: Can we rename this module to `rabbit_message_interceptor_routing_node`?
This way all `rabbit_message_interceptor*` modules show up next to each other in the file menu.

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -2658,23 +2658,94 @@ end}.
...
+                                #{Key => Value},
+                                Acc),
+                        {NewAcc, [Module| Order]}
+                    end,
+                    {#{}, []},
+                    InterceptorsConfig
+                ),
+                Order = lists:uniq(Order0),
+                [{O, maps:without([enabled], maps:get(O, Result))} || O <- Order]
+        end
```
ansd, line 2748: Can you add unit tests?
`deps/rabbit/test/config_schema_SUITE_data/rabbit.snippets` is the file to add these tests.

ansd (changes requested):

deps/rabbitmq_mqtt/src/rabbit_mqtt_client_id_interceptor.erl
```diff
@@ -0,0 +1,18 @@
+-module(rabbit_mqtt_client_id_interceptor).
+
+-behaviour(rabbit_message_interceptor).
+
+-define(ANN_CLIENT_ID, <<"client_id">>).
```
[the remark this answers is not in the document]
kjnilsson: `x-opt-mqtt-client-id` gets my vote.

deps/rabbit/src/rabbit_reader.erl
```diff
@@ -645,7 +645,22 @@ handle_other(handshake_timeout, State) ->
 handle_other(heartbeat_timeout, State = #v1{connection_state = closed}) ->
     State;
 handle_other(heartbeat_timeout,
-             State = #v1{connection = #connection{timeout_sec = T}}) ->
+             State = #v1{connection = #connection{timeout_sec = T, name = ConnectionName},
+                         sock = Socket}) ->
+    case ssl:connection_information(Socket, [keylog]) of
```
kjnilsson, line 650: is this related to message interceptors?

deps/rabbit/src/rabbit_message_interceptor.erl
```diff
@@ -1,50 +1,49 @@
...
-%% now be configured to add such headers. This enables non-AMQP 0.9.1 protocols (that
-%% do not use rabbit_channel) to also add AMQP 0.9.1 headers to incoming messages.
 -module(rabbit_message_interceptor).
--include("mc.hrl").
 
--export([intercept/1]).
+-export([intercept/3,
+         set_msg_annotation/4]).
+
+-type protocol() :: amqp091 | amqp | mqtt310 | mqtt311 | mqtt50.
```
kjnilsson, line 6: change `amqp` to `amqp10` as it then matches the global metric labels we currently emit

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -2607,3 +2609,15 @@ mc_env() ->
         MqttX ->
             #{mqtt_x => MqttX}
     end.
+
+build_msg_interceptor_ctx(#state{cfg = #cfg{client_id = ClientId,
```
kjnilsson, line 2613: this is great and what we discussed as we think lower memory overhead is preferrable. It would be good if it at all possible you could run some throughput tests so we can get a view on how much this might affect performance. https://www.rabbitmq.com/blog/2023/03/21/native-mqtt#latency-and-throughput

kjnilsson (changes requested):
Nearly there! couple of tweaks

deps/rabbit/src/rabbit_reader.erl
LoisSotoLopez: Definitely not, my bad.

deps/rabbit/src/rabbit_message_interceptor_timestamp.erl
```diff
@@ -0,0 +1,28 @@
...
+
+-export([
+    intercept/4
+]).
+
+intercept(Msg0, _MsgInterceptorCtx, _Group, Config) ->
+    Ts = mc:get_annotation(?ANN_RECEIVED_AT_TIMESTAMP, Msg0),
+    Overwrite = maps:get(overwrite, Config, false),
+    Msg = rabbit_message_interceptor:set_msg_annotation(
+        Msg0,
```
kjnilsson, line 16: select this funcction and hit `=` in nvim

deps/rabbit/src/rabbit_message_interceptor_routing_node.erl
```diff
@@ -0,0 +1,16 @@
+-module(rabbit_message_interceptor_routing_node).
+-behaviour(rabbit_message_interceptor).
+
+-define(HEADER_ROUTING_NODE, <<"x-routed-by">>).
+
+-export([
```
kjnilsson, line 6: select and hit `=` in nvim

deps/rabbit/src/rabbit_message_interceptor_timestamp.erl
```diff
@@ -0,0 +1,28 @@
+-module(rabbit_message_interceptor_timestamp).
+-behaviour(rabbit_message_interceptor).
+
+-include("mc.hrl").
+
+-define(HEADER_TIMESTAMP, <<"timestamp_in_ms">>).
+
+-export([
```
kjnilsson, line 8: `=`

kjnilsson (changes requested):
This looks good to me, a couple of lines to indent with `=`.

Then just squash all commits and write a nice single commit for the feature and update PR description and we can merge I think.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
LoisSotoLopez: Run the benchmarks against the last commit (_build on publish_) and with the last one commit building the context map on init (_build on init_).

On an Arch Linux w/ 10 cores and 24 GB of RAM.

```
QOS1 build on publish

================= TOTAL PUBLISHER (100) =================
Total Publish Success Ratio:   100.000% (1000000/1000000)
Total Runtime (sec):           63.151
Average Runtime (sec):         62.670
Pub time min (ms):             0.134
Pub time max (ms):             39.215
Pub time mean mean (ms):       6.258
Pub time mean std (ms):        0.044
Average Bandwidth (msg/sec):   159.574
Total Bandwidth (msg/sec):     15957.402

================= TOTAL SUBSCRIBER (100) =================
Total Forward Success Ratio:      100.000% (1000000/1000000)
Forward latency min (ms):         0.126
Forward latency max (ms):         36.001
Forward latency mean std (ms):    0.020
Total Mean forward latency (ms):  3.128


QOS0 build on publish

================= TOTAL PUBLISHER (100) =================
Total Publish Success Ratio:   100.000% (1000000/1000000)
Total Runtime (sec):           1.352
Average Runtime (sec):         0.966
Pub time min (ms):             0.002
Pub time max (ms):             52.091
Pub time mean mean (ms):       0.036
Pub time mean std (ms):        0.014
Average Bandwidth (msg/sec):   14429.931
Total Bandwidth (msg/sec):     1442993.119

================= TOTAL SUBSCRIBER (100) =================
Total Forward Success Ratio:      75.687% (756868/1000000)
Forward latency min (ms):         7.968
Forward latency max (ms):         4167.565
Forward latency mean std (ms):    345.354
Total Mean forward latency (ms):  1803.073


QOS1 build on init

================= TOTAL PUBLISHER (100) =================
Total Publish Success Ratio:   100.000% (1000000/1000000)
Total Runtime (sec):           63.087
Average Runtime (sec):         62.568
Pub time min (ms):             0.142
Pub time max (ms):             31.975
Pub time mean mean (ms):       6.248
Pub time mean std (ms):        0.054
Average Bandwidth (msg/sec):   159.839
Total Bandwidth (msg/sec):     15983.907

================= TOTAL SUBSCRIBER (100) =================
Total Forward Success Ratio:      100.000% (1000000/1000000)
Forward latency min (ms):         0.138
Forward latency max (ms):         19.824
Forward latency mean std (ms):    0.025
Total Mean forward latency (ms):  3.112


QOS0 build on init

================= TOTAL PUBLISHER (100) =================
Total Publish Success Ratio:   100.000% (1000000/1000000)
Total Runtime (sec):           1.423
Average Runtime (sec):         0.961
Pub time min (ms):             0.002
Pub time max (ms):             38.604
Pub time mean mean (ms):       0.030
Pub time mean std (ms):        0.008
Average Bandwidth (msg/sec):   12087.794
Total Bandwidth (msg/sec):     1208779.411

================= TOTAL SUBSCRIBER (100) =================
Total Forward Success Ratio:      75.167% (751674/1000000)
Forward latency min (ms):         1.671
Forward latency max (ms):         4043.116
Forward latency mean std (ms):    207.787
Total Mean forward latency (ms):  1765.429
```

I get that the under-100% _Total Forward Success Ratio_ on subscribers for QOS 0 comes from me using not a boxed environment + reduced RAM compared to the one used on the doc page you linked.

kjnilsson: So unexpectedly the qos0 test is faster creating the map dynamically?

LoisSotoLopez: It seems it is, but does not really make too much sense to me. I'll repeat this benchmark on a dedicated machine just in case after addressing the squashed commit writing.

ansd (changes requested):

deps/rabbitmq_mqtt/src/rabbit_mqtt_message_interceptor_client_id.erl
```diff
@@ -0,0 +1,17 @@
...
+          #{client_id := ClientId},
+          incoming_message_interceptors,
+          #{annotation_key := AnnotationKey}
+         ) ->
+    rabbit_message_interceptor:set_msg_annotation(Msg,
+                                                  AnnotationKey,
+                                                  ClientId,
+                                                  true);
+intercept(Msg, _MsgInterceptorCtx, _Group, _Config) ->
+    Msg.
```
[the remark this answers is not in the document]
ansd: ... let me know if you have difficulties adding such a test case (I can add it as well)

LoisSotoLopez:
It might be better to leave full support for outgoing interceptors out of this PR. It lays the groudwork for those interceptors, but triggering them is not included in this PR to not grow the scope.  Wdyt?

ansd:
> It might be better to leave full support for outgoing interceptors out of this PR. It lays the groudwork for those interceptors, but triggering them is not included in this PR to not grow the scope. Wdyt?

That's okay, we can add them at a later point.

deps/rabbitmq_mqtt/src/rabbit_mqtt_message_interceptor_client_id.erl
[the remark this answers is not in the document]
LoisSotoLopez: As you can see from my last commit, I'm struggling to get the published message with the `amqp10_client`. If you want to finish that up, I won't come back to this until  tomorrow ~0830.

deps/rabbitmq_mqtt/Makefile
```diff
@@ -43,7 +43,7 @@ export BUILD_WITHOUT_QUIC
 
 LOCAL_DEPS = ssl
 DEPS = ranch rabbit amqp10_common
-TEST_DEPS = cowlib emqtt ct_helper rabbitmq_ct_helpers rabbitmq_ct_client_helpers rabbitmq_management amqp_client rabbitmq_consistent_hash_exchange rabbitmq_amqp_client rabbitmq_stomp rabbitmq_stream rabbitmq_federation
+TEST_DEPS = cowlib emqtt ct_helper rabbitmq_ct_helpers rabbitmq_ct_client_helpers rabbitmq_management amqp_client rabbitmq_consistent_hash_exchange rabbitmq_amqp_client rabbitmq_stomp rabbitmq_stream rabbitmq_federation amqp10_client
```
ansd, line 46: I don't think this addition is necessary because `rabbitmq_amqp_client` already depends on `amqp10_client`.

ansd (changes requested):

deps/rabbitmq_mqtt/priv/schema/rabbitmq_mqtt.schema
```diff
@@ -303,3 +303,9 @@ end}.
     {datatype, integer},
     {validators, ["non_negative_integer"]}
 ]}.
+
+{mapping, "message_interceptor.incoming.rabbit_mqtt_message_interceptor_client_id.annotation_key",
```
ansd, line 307: 1. Why is this called `message_interceptor` instead of `message_interceptors` as in file `deps/rabbit/priv/schema/rabbit.schema`?
2. I think it's not great that this Cuttleflish mapping defined in the MQTT plugin sets the environment in app `rabbit`. Do you see this approach done for any other configuration in any other plugin?
3. I tried it out by setting the following in `rabbitmq.conf`:
```
message_interceptor.incoming.rabbit_mqtt_message_interceptor_client_id.annotation_key = my-key
```
and this setting won't be set in the application environment of app `rabbit`.

Did you test this once that this Cuttlefish translation actually works?

ansd (changes requested):

deps/rabbitmq_mqtt/priv/schema/rabbitmq_mqtt.schema
ansd: Thanks for the clarification. Ok, after fixing 1. I now see my above setting taking effect successfully.

However, the following should work as well if I write my own third party plugin, correct?
```
message_interceptors.incoming.my_third_party_module.my_custom_key = some-value
```
With this PR, I get the following error:
```
2025-04-15 13:57:39.511692+00:00 [error] <0.162.0> You've tried to set message_interceptors.incoming.my_third_party_module.my_custom_key, but there is no setting with that name.
2025-04-15 13:57:39.511723+00:00 [error] <0.162.0>   Did you mean one of these?
2025-04-15 13:57:39.731574+00:00 [error] <0.162.0>     message_interceptors.incoming.$interceptor.overwrite
2025-04-15 13:57:39.731619+00:00 [error] <0.162.0>     message_interceptors.incoming.$interceptor.enabled
2025-04-15 13:57:39.731645+00:00 [error] <0.162.0>     message_interceptors.incoming.set_header_routing_node.overwrite
2025-04-15 13:57:39.731662+00:00 [error] <0.162.0> Error preparing configuration in phase transform_datatypes:
2025-04-15 13:57:39.731676+00:00 [error] <0.162.0>   - Conf file attempted to set unknown variable: message_interceptors.incoming.my_third_party_module.my_custom_key
```

ansd: Another thing to consider is: If I disable a plugin, I would like the plugin's interceptors to be removed as well. rabbit_registry is an advantage here.
What do you think about the following?
Each plugin that defines interceptors (e.g. MQTT) has its own Cuttlefish translation not interfering with app `rabbit`. Then upon plugin activation and deactivation the plugin itself ensures that it adds and removes its interceptors to/from persistent term.

gomoripeti: No, I don't think `my_custom_key` should just work. You need to define a type (mapping) for it in the schema. `annotation_key` is an example of that.

michaelklishin: There is a test suite called `config_schema[_SUITE]` which uses an Erlang term file with translation pairs. These `*.schema` translations are, should be and can trivially be tested.

gomoripeti: I imagined the list of interceptors to be quite static and set up by the operators. (Also their order of execution is explicitly defined so in the unlikely event of conflict it is defined what will happen) (I personally don't really like how channel interceptors work, they register dynamically, but as plugin startup order is not defined the channel interceptor order is also not known)

I understand from a user's perspective that it would be expected that if a plugin is disabled then its interceptors are also deactivated, but this is quite hard to implement without the dynamic registration similar to channel interceptors. Currently (maybe accidentally) RabbitMQ can still call code from disabled plugins, this has its pros and cons. But if the persistent_term registration is bound to plugin startup and stop, but the list of message interceptors is hardcoded in config, this detaches the two and an interceptor might be called without a config. That's why I think its simpler to have the list of interceptor callbacks and their config in one place, all handled by rabbit app. (With the caveat that a disabled plugin's interceptor will still work if the plugin is disabled, it will only not be invoked if the plugin is deleted or not available)

gomoripeti: There is existing example of callbacks and config being separated: the list of auth_backends are set in `rabbit` app env but their config is in the plugin apps' env and the plugin also must be started. In case of auth_backend_http if it is configured as an auth_backend but the rabbitmq_auth_backend_http plugin is not enabled, it wont start `inet` and the backend wont work. In this case disabling the plugin won't automagically disable the auth_backend. But the order of auth_backends is well defined. So it has pros and cons.

ansd:
Thank you for your great work and great collaboration @LoisSotoLopez and @gomoripeti! I think we all settled with a good solution 🙂

Since I can't push to your `cloudamqp` fork, I created a new branch/PR (https://github.com/rabbitmq/rabbitmq-server/pull/13760) with your original commit and take the liberty to close this PR. From my point of view, we can merge the new PR. Let me know what you think.

LoisSotoLopez:
Changes you proposed on the new PR seem reasonable to me. I'll be running the benchmarks again today so you can expect to get those results before 4 pm CET.

LoisSotoLopez:
Ah, I had the brilliant idea of running the benchmarks on a burstable instance. Also less performant than my laptop :clap: . Will try to run those benchmarks during the holidays.

PR #13344, opened 2025-02-14
Redirect to end_session_endpoint for idp_initiated logon when it is configured

MarcialRosales:
## Proposed Changes

Addresses https://github.com/rabbitmq/rabbitmq-server/issues/12545

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [X] Bug fix (non-breaking change which fixes issue #NNNN)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

PR #13644, opened 2025-03-28
Rabbitmq server 13344 for v4.1.x

michaelklishin:
This is a manual backport of #13344 to `v4.1.x`.


michaelklishin:
#13344 was reverted.

PR #13646, opened 2025-03-28
queue info metric: guard against whereis returning `undefined`

mkuratczyk:
Follow up to https://github.com/rabbitmq/rabbitmq-server/pull/13583. It crashes with `called as is_process_alive(undefined)` under some conditions.


kjnilsson (approved):

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
```diff
@@ -433,7 +433,8 @@ membership(Pid, Members) when is_pid(Pid) ->
 membership({Name, Node}, Members) ->
     case Node =:= node() of
         true ->
-            case is_process_alive(whereis(Name)) of
+            Pid = whereis(Name),
+            case is_pid(Pid) andalso is_process_alive(Pid) of
```
kjnilsson, line 437: the is_pid/1 check is enough to determine liveness so the `is_process_alive/1` call isn't necessary

PR #13661, opened 2025-03-31
AMQP 1.0 client: make it easier to pass in a virtual host

michaelklishin:
while AMQP 1.0 does not have the concept of virtual hosts, RabbitMQ still does, and so do shovels
and other components.

Currently the way to pass a virtual host is via
a query parameter named 'hostname', which is
very counterintuitive to most users.

With this PR, there are now two better options:

1. The URI path, with the leading slash stripped off
2. The 'vhost' query parameter which is formatted correctly

Both still feed the 'hostname' connection parameter but are much easier to guess and remember
for, say, AMQP 1.0 Shovel users.

### DIFF
--- a/deps/amqp10_client/src/amqp10_client.erl
+++ b/deps/amqp10_client/src/amqp10_client.erl
@@ -39,7 +39,9 @@
          link_handle/1,
          get_msg/1,
          get_msg/2,
-         parse_uri/1
+         parse_uri/1,
+         %% for tests
+         binary_without_leading_slash/1
         ]).
 
 -type snd_settle_mode() :: amqp10_client_session:snd_settle_mode().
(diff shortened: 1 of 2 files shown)


michaelklishin:
I don't feel strongly about the URI path segment part, it can be rolled back. But asking the user to pass in a `hostname` via query parameters to pass in the virtual host is very counterintuitive.

lukebakken (approved):
👍

michaelklishin:
The conclusion is: there are plans to change way we propagate this additional piece of information from the AMQP 1.0 Erlang client to the AMQP 1.0 implementation in RabbitMQ (the change won't affect other AMQP 1.0 brokers).

We'd still need a simplification of URI parsing, possibly in the Shovel plugin, but the `hostname` field is a part of the AMQP 1.0 documentation and some believe that AMQP 1.0 users must read it, or at least read an overview of sorts.

I'll update Shovel guides a bit to make extra clear what needs to be done at the moment, a `hostname` URI query parameter must be set to `vhost:{name}`.

ansd:
> I don't think this is a good change. The client is a generic AMQP client not a RabbitMQ specific wrapper.

I agree. Any RabbitMQ specific things should ideally go into the RabbitMQ specific wrapper: https://github.com/rabbitmq/rabbitmq-server/tree/main/deps/rabbitmq_amqp_client

PR #13672, opened 2025-04-02
Introduce stream SAC status instead of active flag

acogoluegnes:


Gsantomaggio:
I can't validate the code but I can't break the SAC anymore.  The PR looks to me. 
cc  @jonnepmyra

deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -803,22 +854,51 @@ maybe_resize_coordinator_cluster() ->
...
+                                  [] ->
+                                      ok;
+                                  [Old | _]  when length(RabbitNodes) > 0 ->
+                                      %% this ought to be rather rare as the stream
+                                      %% coordinator member is now removed as part
+                                      %% of the forget_cluster_node command
+                                      rabbit_log:info("~ts: Rabbit node(s) removed "
+                                                      "from the cluster, "
+                                                      "deleting: ~w", [?MODULE, Old]),
+                                      remove_member(Leader, Members, Old)
```
kjnilsson: instead of the outer `_` just add `_` before remove_member/4 and add a new line returning `ok`

PR #13703, opened 2025-04-08
QQ: handle tick tweaks

kjnilsson:
Move leader repair earlier in tick function to ensure more
timely update of meta data store record after leader change.

Also use RPC_TIMEOUT macro for metric/stats multicalls to improve
liveness when a node is connected but partitioned / frozen.


ansd:
This PR fixes the test failure in https://github.com/rabbitmq/rabbitmq-amqp-java-client/blob/e4af739a56587ff611967bfa70f51ce9eac321f2/src/test/java/com/rabbitmq/client/amqp/impl/ClusterTest.java#L294.

michaelklishin (approved):

PR #13718, opened 2025-04-09
Start coordination Ra system before quorum_queues

kjnilsson:
So that quorum_queues shuts down _before_ coordination where khepri run inside. Quorum queues depend on khepri so need to be shut down first.


dumbbell:
It’s a good idea. However, shouldn’t we just `lists:reverse/1` the list in `rabbit_ra_systems:ensure_stopped/0`? This way, we ensure that the systems list corresponds to the starting order and the stop function should be fixed to stop in opposite order as the starte one.

dumbbell (approved):

PR #13723, opened 2025-04-10
Remove log level tests

mkuratczyk:
When debug logging is enabled, we log something at each log level to test if logs are emitted. I don't think this is particularly useful, but it's certainly annoying, because I constantly need to filter out these logs when searching if any errors happened during tests.


ansd (approved):
I agree, this fake logging has also annoyed me multiple times.

PR #13752, opened 2025-04-15
Use RABBITMQ_ENABLED_PLUGINS over RABBITMQ_ENABLED_PLUGINS_FILE

MarcialRosales:
gmake was ignoring RABBITMQ_ENABLED_PLUGINS_FILE. The majority of the suites were working because all plugins were enabled. But only when I wanted to specify just a few it did not work. It was always enabling all plugins.
With this change, the content of the file `enabled_plugins` in selenium is converted into the value of the env var RABBITMQ_ENABLED_PLUGINS.

This change is necessary for tanzu wsr testing.

## Types of Changes

- [x] Build system and/or CI


selenium/test/exchanges/management.js
```diff
@@ -44,7 +44,7 @@ describe('Exchange management', function () {
       ["/", "amq.fanout", "fanout"],
       ["/", "amq.headers", "headers"],
       ["/", "amq.match", "headers"],
-      ["/", "amq.rabbitmq.event", "topic"],
+//      ["/", "amq.rabbitmq.event", "topic"],
```
michaelklishin, line 47: FTR, this exchange is declared and used by a plugin that is no longer enabled as of this PR.

michaelklishin (approved):

PR #13765, opened 2025-04-16
[skip ci] Grafana: update dashboards for RabbitMQ 4.1.x

mkuratczyk:
Key changes:
- endpoint variable to handle scraping multiple endpoints
- message size panels (new metric in 4.1)
- panels at the top of the Overview dashboard should be more up to date
  (they show the latest value)
- values should be accurate if multiple endpoints are scraped
  (previously, many would be doubled)
- Nodes table shows fewer columns and shows node uptime


ansd:
I checked the `message size panels (new metric in 4.1)`. Looks great, thank you!

PR #13809, opened 2025-04-26
Management UI: add a TLS column to the listeners and ports table

razvanphp:
## Proposed Changes

While working on a custom plugin, I found this to be missing. The new cowboy sockets don't appear in Web context, so one does not know if it's TLS or not, except for deriving from the name of the protocol.

Would be nice to add also the Path, but not sure if it's worth the code changes for only the web mqtt/stomp.

## Types of Changes

- [ ] Build system and/or CI

Here it is how it looks:

Screenshot 2025-04-26 at 19 18 28

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/tmpl/overview.ejs
+++ b/deps/rabbitmq_management/priv/www/js/tmpl/overview.ejs
@@ -252,6 +252,7 @@
 <% } %>
     <th>Bound to</th>
     <th>Port</th>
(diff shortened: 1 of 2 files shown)


michaelklishin:
@razvanphp thank you for contributing. I have renamed the field and column name to say "TLS", tested it with several listeners and merged.

Will backport to `v4.1.x` manually, so this will ship in `4.1.1`.

PR #13831, opened 2025-04-30
v4.0.x: Backport erlang.mk plus a few rabbitmqct_ct_helpers changes

michaelklishin:
This PR backports a few non-functional changes that collectively fix a few test failures in certain environments.


michaelklishin:
We need to start with an `erlang.mk` backport. It's a web of dependencies.

PR #13836, opened 2025-04-30
Management UI: new page and elements for superstreams (partitioned streams)

dcorbacho:


michaelklishin (changes requested):
Right now, when you add a superstream, the form clears and there are no visible changes. The actual partitions will only be visible on the Queues and Streams page.

Should we redirect to the Queues and Streams page in the UI code?

And yes, we should validate virtual host for existence similarly
to what `rabbit_mgmt_wm_queues` does:

```erl
resource_exists(ReqData, {Mode, Context}) ->
    %% just checking that the vhost requested exists
    {case rabbit_mgmt_util:all_or_one_vhost(ReqData, fun (_) -> [] end) of
         vhost_not_found -> false;
         _               -> true
     end, ReqData, {Mode, Context}}.
```

dcorbacho:
@michaelklishin changes done

michaelklishin (approved):

PR #13837, opened 2025-04-30
Modify default queue type injection logic

mkuratczyk:
There were a few issues:
1. BUG: `rabbitmqctl export definitions -` would export the `default_queue_type` within vhost's metadata if set explicitly (correct) but also outside of the `metadata` block even if not set explicitly for the vhost (not correct)
2. BUG/consistency issue: HTTP API / `rabbitmqadmin` would always export the default queue type in both places
- outside of the `metadata` block is just redundant
- inside the `metadata` block is inconsistent with `rabbitmqctl export_definitions` (the difference stems from the fact that CTL only exports existing vhosts' metadata, while HTTP API injects the configuration-level DQT into each vhost's `metadata` upon export)
3. The aforementioned inject-upon-export logic is prone to confusion: I can have a system with no DQT configuration and export definitions from it, then restart with a DQT on the node configuration level and export again. The exports will be different because the configuration-level default changed and the current default is injected into the export.

This PR changes the DQT logic: no DQT is injected upon export because it's injected/inherited from the node's configuration into vhost's metadata when a vhost is declared. Therefore each vhost has its own DQT explicitly defined and it's always present in any export (both CTL and HTTP API).

The downsides of this approach is that export/import between systems with a different configuration-level DQT will now require adjusting the DQT in the JSON file

However the logic is should be easier now and more consistent with how queue-level DQT works: when a queue is declared without a type, the default queue type is injected. Going forward, DQT changes don't affect the queue. Similarly, now if a vhost with no DQT is declared, configuration-level DQT is injected and configuration changes don't affect existing vhosts.

### DIFF
--- a/deps/rabbit/src/rabbit_definitions.erl
+++ b/deps/rabbit/src/rabbit_definitions.erl
@@ -1081,12 +1081,10 @@ list_vhosts() ->
 
 vhost_definition(VHost) ->
     Name = vhost:get_name(VHost),
-    DQT = rabbit_queue_type:short_alias_of(rabbit_vhost:default_queue_type(Name)),
     #{
         <<"name">> => Name,
         <<"limits">> => vhost:get_limits(VHost),
-        <<"metadata">> => vhost:get_metadata(VHost),
-        <<"default_queue_type">> => DQT
+        <<"metadata">> => vhost:get_metadata(VHost)
     }.
 
 list_users() ->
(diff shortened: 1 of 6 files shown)


michaelklishin:
@lukebakken @mkuratczyk and I have discussed this in detail, there is also at least one related change in `rabbitmqadmin` that we'd like to do.

With this change, the injection of the DQT moves from definition export time to virtual host creation time, which is easier to reason about and avoids inconsistent or surprising behavior in a number of scenarios.

dcorbacho (approved):
LGTM. It seems the cleanest and simplest way to present the info.

michaelklishin (approved):

michaelklishin:

PR #13873, opened 2025-05-08
QQ grow to a target quorum cluster size

Ayanda-D:
## Proposed Changes

Hello team 👋 

The following changes are a proposed extension to the `rabbitmq-queues grow` command. Currently this command only allows growing queues to a single target node per execution, for a set of queues matching the `--queue-pattern`. This means the command needs to be executed a number of times to fully restore a quorum queue (replicas) back to N-number of nodes (which can also be expensive from the multiple RPCs from the temporary CLI execution node). As an extension to the `rabbitmq-queues grow` command, we'd like to allow growing queues to a specified number of nodes N / target quorum cluster size (allowing the broker to do all the grow processing to multiple nodes in one CLI execution). The command signature becomes:

```
rabbitmq-queues grow <node | quorum_cluster_size> .... 
``` 

When quorum queues get into a bad state, e.g. become leaderless (https://github.com/rabbitmq/rabbitmq-server/discussions/12701,  https://github.com/rabbitmq/rabbitmq-server/discussions/13101) and shrink operations (https://github.com/rabbitmq/rabbitmq-server/pull/12427) to a single node to attempt rescue are applied, fast recovery/growing back to N nodes / to a target quorum cluster size for the set of queues becomes a critical requirement. 

Please take a look. Thanks!

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -1528,25 +1528,14 @@ shrink_all(Node) ->
 grow(Node, VhostSpec, QueueSpec, Strategy) ->
     grow(Node, VhostSpec, QueueSpec, Strategy, promotable).
 
--spec grow(node(), binary(), binary(), all | even, membership()) ->
+-spec grow(node() | integer(), binary(), binary(), all | even, membership()) ->
     [{rabbit_amqqueue:name(),
       {ok, pos_integer()} | {error, pos_integer(), term()}}].
-grow(Node, VhostSpec, QueueSpec, Strategy, Membership) ->
+grow(Node, VhostSpec, QueueSpec, Strategy, Membership) when is_atom(Node) ->
     Running = rabbit_nodes:list_running(),
     [begin
          Size = length(get_nodes(Q)),
-         QName = amqqueue:get_name(Q),
-         rabbit_log:info("~ts: adding a new member (replica) on node ~w",
-                         [rabbit_misc:rs(QName), Node]),
-         case add_member(Q, Node, Membership) of
-             ok ->
-                 {QName, {ok, Size + 1}};
-             {error, Err} ->
-                 rabbit_log:warning(
-                   "~ts: failed to add member (replica) on node ~w, error: ~w",
-                   [rabbit_misc:rs(QName), Node, Err]),
-                 {QName, {error, Size, Err}}
-         end
+         maybe_grow(Q, Node, Membership, Size)
      end
      || Q <- rabbit_amqqueue:list(),
         amqqueue:get_type(Q) == ?MODULE,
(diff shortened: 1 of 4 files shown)


deadtrickster:
Hi, what happens when add member call errors in the middle of the loop?

michaelklishin (changes requested):
@Ayanda-D from the early days of Ra and QQs, we have decided to not add N members, by design. Some other Raft implementations explicitly prohibit such parallel membership changes, or the logic of the spec becomes really hard to follow at times.

So at best we can do this after a pause of say, 10 seconds minimum. And ideally after learning that the previous membership change has completed.

Ayanda-D:
hi folks, thanks for taking a look.

@michaelklishin i agree "parallel membership changes" would be a bit tricky and hard to follow, here. Everything is still sequential. We are still retaining the exact same behaviour of `grow <node>` the main extension for `grow N`, is that the loop simply runs `N` times executing `grow <node-N>` where `<node-N>` is chosen for the CLI caller (which users are currently, manually having to do over RPC, and we just want the broker to do this for us for faster queue recovery). This is still the same behaviour, just [running N-times (or QuorumClusterSize-times)](https://github.com/rabbitmq/rabbitmq-server/pull/13873/files#diff-aafd1be4bf1d8c8af7462438372594b2b2253cfb3187687df6d30f92526e2f75R1575). For each matched queue, if `QuorumClusterSize` has already been reached, then [there's no need to grow](https://github.com/rabbitmq/rabbitmq-server/pull/13873/files#diff-aafd1be4bf1d8c8af7462438372594b2b2253cfb3187687df6d30f92526e2f75R1571-R1572) and the loop continues to the next matched queue from `--queue-pattern` (rest of the behaviour still the same). After loop runs N-times, results are just sent back to the CLI caller as current grow implementation.

@deadtrickster if grow fails middle of the loop, still the same behaviour of current `grow <node>` loop is retained, [we just add the failure](https://github.com/rabbitmq/rabbitmq-server/pull/13873/files#diff-aafd1be4bf1d8c8af7462438372594b2b2253cfb3187687df6d30f92526e2f75R1600) to the list of results, which are returned and [formatted on the CLI grow command callback](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_cli/lib/rabbitmq/cli/queues/commands/grow_command.ex#L94-L100). 

Nothing much's changed functionality wise, we're just running the same grow command,  `N times` or `TargetQuorumClusterSize times`, each run an available node is being chosen and attempted for grow for the caller. List of success and failures are returned to the caller as per current behaviour. Might be slower than doing in parallel, but I yeah keeping sequential is safe. We're [still seeing these issues](https://github.com/rabbitmq/rabbitmq-server/discussions/13101#discussioncomment-12315927) and need this option for faster recovery of the broken queues.

michaelklishin:
@Ayanda-D per discussion with @kjnilsson, we suggest doing the following before adding more members (replicas).

We need to make sure that every QQ member is a voter. That means that they have caught up enough to assume that it's really safe to proceed. If that's not the case yet, the algorithm should skip the queue where it's been the case. So it will be "best effort" but I'm not sure what else we can do.

We should also do that before we start the process (if we don't do so already).

That variation of this change we are willing to accept.

michaelklishin:
@Ayanda-D that's an excellent point. I think we should conclude a combination of those options to be invalid. If you want to explicitly add a non-voter member, do it one at a time.

Ayanda-D:
fyi, there's another update here, target cluster size need be parsed to `integer` - we had missed this, the unit tests already assumed `int` hence was PASSING. this is now updated to reflect expected IO input format.

michaelklishin:
@Ayanda-D can this be moved to a separate command, e.g. `grow_to_count`. This will both make the command less complex and allow you to introduce more validations.

michaelklishin:
I have completely missed the most recent round of changes. @Ayanda-D thank you for introducing a new command, I'm sure as a new command we'll find a way to get it in after the `4.2.0` release.

michaelklishin:
Merged with some tweaks in https://github.com/rabbitmq/rabbitmq-server/pull/15166.

PR #13885, opened 2025-05-12
Ra v3

kjnilsson:
PR to test Ra v3 changes


deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1678,6 +1725,53 @@ reclaim_memory(Vhost, QueueName) ->
     ra_log_wal:force_roll_over({?RA_WAL_NAME, Node}).
 
 %%----------------------------------------------------------------------------
+
+get_consumer_disconnected_timeout(Q) ->
+    case args_policy_lookup(<<"consumer-disconnected-timeout">>, fun min/2, Q) of
```
ansd, line 1871: Do we need to add this new policy to `unsupported_policies` for classic queues and streams?
(Ideally every queue type should advertise their supported policies rather advertising their unsupported policies.)

kjnilsson: yes we probably do, and I agree unsupported_policies should go really.

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -1674,6 +1674,17 @@ end}.
...
 
+%% Time to wait before returning messages to the queue when a consumer's
+%% node becomes unreachable (due to network partition), in milliseconds.
+%% Only applies to quorum queues.
+%%
+%% {consumer_disconnected_timeout, 10000},
+
+{mapping, "consumer_disconnected_timeout", "rabbit.consumer_disconnected_timeout", [
+    {datatype, integer},
+    {validators, ["non_zero_positive_integer"]}
```
ansd, line 1643: The corresponding `policy_validator` allows a timeout of 0.
Should we allow or disallow a timeout of 0?
Allowing a timeout of 0 is equivalent to what we have today in QQs, so I guess it should be allowed?

kjnilsson: yes I think it should be alllowed

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -1043,12 +1342,14 @@ handle_aux(_RaftState, cast, {#return{msg_ids = MsgIds,
...
-                          fun (MsgId, ?MSG(Idx, Header), {RA0, Acc}) ->
+                          fun (MsgId, ?C_MSG(Msg), {RA0, Acc}) ->
+                                  Idx = get_msg_idx(Msg),
+                                  Header = get_msg_header(Msg),
                                   %% it is possible this is not found if the consumer
                                   %% crashed and the message got removed
                                   case ra_aux:log_fetch(Idx, RA0) of
                                       {{_Term, _Meta, Cmd}, RA} ->
-                                          Msg = get_msg(Cmd),
+                                          Msg = get_msg_from_cmd(Cmd),
```
ansd: Looks like this crashes given that `Msg` is bound

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -2194,42 +2779,45 @@ checkout_one(#{system_time := Ts} = Meta, ExpiredMsg0, InitState0, Effects0) ->
...
+                                                        delivery_count = DelCnt},
+                                    Size = get_header(size, get_msg_header(Msg)),
+                                    State1 =
+                                        State0#?STATE{service_queue = SQ1,
+                                                      msg_bytes_checkout = BytesCheckout + Size,
+                                                      msg_bytes_enqueue = BytesEnqueue - Size,
+                                                      next_consumer_timeout = min(Timeout, NextConTimeout)},
+                                    Effects = maybe_add_consumer_timeout_effect(Timeout,
+                                                                                NextConTimeout,
+                                                                                Effects1),
```
ansd, line 2822: Instead of being called here for every message being checked out, should `maybe_add_consumer_timeout_effect/3` instead be called just once at the end of function `checkout/5`?

It's worth double checking what happens if messages are assigned to different consumers with decreasing consumer timeouts. Looks like multiple timer effects will be emitted:
```
[{timer,...,100}, ..., {timer,...,200}, ..., {timer,...,300}, ...]
```
which seems to result in the timer to fire late (at the longest timeout instead of the shortest) ? (although this will self correct with the next checkout round)

deps/rabbit/src/rabbit_fifo.erl
ansd: fixed

PR #13930, opened 2025-05-22
Add CQ and QQ Observer CLI plugins from boot steps

ikavgo:
Do not hard code them, also preserve user-provided plugins list

### DIFF
--- a/deps/rabbit/src/rabbit_observer_cli.erl
+++ b/deps/rabbit/src/rabbit_observer_cli.erl
@@ -10,10 +10,8 @@
 -export([init/0, add_plugin/1]).
 
 init() ->
-    application:set_env(observer_cli, plugins, [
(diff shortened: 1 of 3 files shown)


mkuratczyk:
There are Observer plugins, not RabbitMQ plugins. I don't think we need any configurability for this - these plugins are only executed if you do `rabbitmq-diagnostics observer -> P -> CQ/QQ`. I think they should be always available, in case you want to take a look at these metrics when debugging an issue

ikavgo:
They are always available - the place of registration with observer cli changes closer to specific plugins.

mkuratczyk:
ok, I must have misunderstood the intention. There is a typo (twice): `plugun`

michaelklishin (approved):

michaelklishin:
@mergifyio backport v4.1.x

PR #13946, opened 2025-05-24
QQ Reconciliator - remove hardcoded triggers, switch to Logger, comment/refactor SUITE

ikavgo:
Additionally to what the title says - Reconciliator has two operation modes that are interwinded - timer and trigger (say node_down), So testing say, triggers alone is actually tricky and involves tweaking Interval and sleeps. I summarized my findings in the new comments for SUITE.

Two things I'd like to fix in the future:

1. there is no trigger for drain/undrain - in the degenerated case full Interval must pass. OTOH it is tricky to do since gen_events AFAIK are scoped by node. So possibly something like `gen_server:abcast` must be utilized to let the whole cluster know about maintenance. But maybe there is some yet-to-be-discovered-by-me facility inside Rabbit that does just that.
2. there is a DOS opportunity I see here: 
```
handle_cast({membership_reconciliation_trigger, Reason}, #state{timer_ref = OldRef,
                               trigger_interval = Time} = State) ->
    ?LOG_DEBUG("Quorum Queue membership reconciliation scheduled: ~p", [Reason]),
    _ = erlang:cancel_timer(OldRef),
    Ref = erlang:send_after(Time, self(), ?EVAL_MSG),
    {noreply, State#state{timer_ref = Ref}};
```
Since previous timer removed and actual work happens on ?EVAL_MSG if we have steady stream of triggers just withing the TriggerInterval the timer will be removed and message will never arrive.


michaelklishin:
@ikavgo this reconciliation kicks in when a replica is lost permanently. So I don't think that maintenance mode events are relevant, they are transient QQ cluster member changes.

deadtrickster:
@michaelklishin thank you for the review. Your wording is much better!

Regarding maintenance events - the SUITE has drain test where cluster size changed so I assumed they are important, just not easily subscribable.

michaelklishin (approved):

michaelklishin:
@SimonUnge @lukebakken @the-mikedavis FYI. This feature was originally contributed by AWS and I'd like your team to be aware of its evolution.

PR #13947, opened 2025-05-25
Run rabbit_ct_hook before cth_styledout so that removing default logger handler works

ikavgo:
As a follow-up to my GChat thread about removing default logger handler to clean CT stdout, I was looking at
injecting logger config with undefined default handler to `ct_run`. It is possible but breaks `cth_styledout` - no nice green things whatsoever. Then I found `rabbit_ct_hook` which calls [redirect_logger_to_ct_logs](https://github.com/rabbitmq/rabbitmq-server/blob/c5d150a7ef43b35426c43bed550f15e5106f5583/deps/rabbitmq_ct_helpers/src/rabbit_ct_helpers.erl#L156) which in turn calls `logger:remove_handler(default)` apparently with zero effect! To cut story short - turned out `rabbit_ct_hook` must run before `cth_styledout` for `remove_handler` line to have any effect.

Doesn't work for parallel targets yet - https://github.com/rabbitmq/rabbitmq-server/blob/77477dee6e7d1fba40f67e9626388c3e3ad3c575/deps/rabbit/Makefile#L227 But for manual local runs like `make -C deps/rabbit ct-<>` it really shines.


ikavgo:
@michaelklishin It works for individual suites, but not for parallel - too much magic. I can add `rabbit_ct_hook` to other deps if this approach is fine.

michaelklishin:
Unless @dumbbell @lhoguin have objections, I think we can extend this to a few subprojects with large test suites (see above) and merge this change for `main`, then see how it goes.

dumbbell:
Looks good to me as well.

ikavgo:
I added crude support for parallel test (ct_master is forked anyway), and replaces some ct:pals with ct:log.

Before: https://github.com/rabbitmq/rabbitmq-server/actions/runs/15240078576/job/42860845449#step:9:13156
After: https://github.com/rabbitmq/rabbitmq-server/actions/runs/15254128554/job/42897594880

ikavgo:
I gave up on rabbit_ct_hook for rabbit_common. ct_helpers depend on rabbit_common and it is all sorts of ugliness

ikavgo:
It is green, finally. If looks good, I think I'll add to the rest of the deps and if you are flying I can merge. Thoughts?

PR #13959, opened 2025-05-27
CQ shared store: Delete from index on remove or roll over

lhoguin:
It is expensive to delete files because we must clean up the index and to get the messages in the file we have to scan it.

Instead of cleaning up the index on file delete this commit deletes from the index as soon as possible. There are two scenarios: messages that are removed from the current write file, and messages that are removed from other files. In the latter case, we
can just delete the index entry on remove. For messages in the current write file, we want to keep the entry in case fanout is used, because we don't want to write the fanout message multiple times if we can avoid it. So we keep track of removes in the current write file and do a cleanup of these entries on file roll over.

Compared to the previous implementation we will no longer increase the ref_count of messages that are not in the current write file, meaning we may do more writes in fanout scenarios. But at the same time the file delete operation is much cheaper.

@mkuratczyk

### DIFF
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -77,8 +77,10 @@
           current_file,
           %% current file handle since the last fsync?
           current_file_handle,
-          %% file handle cache
+          %% current write file offset
           current_file_offset,
+          %% messages that were potentially removed from the current write file
+          current_file_removes = [],
           %% TRef for our interval timer
           sync_timer_ref,
           %% files that had removes
@@ -1150,7 +1152,8 @@ write_message(MsgId, Msg, CRef,
               end, CRef, State1)
     end.
 
-remove_message(MsgId, CRef, State = #msstate{ index_ets = IndexEts }) ->
+remove_message(MsgId, CRef, State = #msstate{
+        index_ets = IndexEts, current_file = CurrentFile }) ->
     case should_mask_action(CRef, MsgId, State) of
         {true, _Location} ->
             State;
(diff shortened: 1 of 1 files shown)


gomoripeti:
sorry to keep on nitpicking :)
a small note, that if I see it correctly compaction might become slower with this change (as during `scan_file_for_valid_messages` before this change there were ref-count=0 index entries which resulted in a `previously_valid` status while now `not_found` entries result in `invalid` status, and causing a scan_next_byte scanning mode)

during healthy operation when messages are mostly consumed in the same order as they are written, most of the rdq files can be deleted before any compaction happens on them. So I agree that its more important to make the delete_file more efficient than compact_file. Just wanted to mention the trade off in the performance of compaction.

lhoguin:
Right it's likely to make compaction itself slower. But it might be better overall when compaction+delete are both happening, because before delete was very slow.

There's other cases where it is likely slower (like mass deleting messages).

deps/rabbit/src/rabbit_msg_store_gc.erl
```diff
@@ -51,6 +51,9 @@ delete(Server, File) ->
 stop(Server) ->
     gen_server2:call(Server, stop, infinity).
 
+prioritise_cast({delete, _}, _Len, _State) -> 5;
+prioritise_cast(_, _Len, _State) -> 0.
```
lhoguin, line 55: Not a bad idea. Worth adding a comment that in the future we should switch to using priority messages introduced in OTP-28. Then we can just remove gen_server2 from that process.

mkuratczyk:
Test results (with this change applied on top of `v4.1.x`)
https://grafana.lionhead.rabbitmq.com/goto/rkYSf0BHg?orgId=1

Highlights:

1. Disk space is reclaimed much faster (v4.1.x env, green, required 12 additional minutes to reclaim the space):


2. Much lower memory usage throughout the tests


The throughput is comparable in most cases.

The downside is slower compaction, since we no longer skip over consumed messages. This should be less of a problem since we prioritised deletion requests now, avoiding unnecessary compactions (for files we can already delete). Deletions are generally much more common, so it feels like it's worth to have faster deletions and slower compactions than the other way round. On my machine we can still do a few compactions per second.

mkuratczyk:
@lhoguin This is now ready for review. While we discussed ideas for speeding up compaction, I feel like it's a too risky change to include here. With 4.1.1 coming soon, I'd rather include this change without further complications.

For future reference, the idea to speed up compaction was to:
1. During recovery, zero the file in places that contain invalid messages
2. Then, on compaction, we could change `not_found -> invalid` to `not_found -> previously_valid` in `scan_and_vacuum_message_file/2`

But any bug in zeroing means data corruption on potentially valid data so I don't want to rush it.

mkuratczyk (approved):

PR #13963, opened 2025-05-27
Selenium: minor refactoring

MarcialRosales:
## Proposed Changes

Use /var over /tmp to store conf files for each suite. Without this change, it is not possible to run selenium unless docker or docker desktop is used.
Rename doWhile to doUntil as it is more appropriate name.

## Types of Changes

- [x] Build system and/or CI

### DIFF
--- a/.github/workflows/test-authnz.yaml
+++ b/.github/workflows/test-authnz.yaml
@@ -72,22 +72,22 @@ jobs:
         docker build -t mocha-test --target test .
 
     - name: Run Suites
-      run: |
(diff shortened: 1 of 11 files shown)


Zerpet:
I don't have a strong opinion, but I would suggest to squash some of the commits, to have a more meaningful commit history. For example, 52788b8 57b1dde 95a2e37 could be squashed into a single commit, since they all apply the same idea to different parts of the code. Same for ad104cd and 53d3b46.

PR #13967, opened 2025-05-28
DO NOT MERGE: v41x msg store index delete on remove

mkuratczyk:
https://github.com/rabbitmq/rabbitmq-server/pull/13959/ on top of v4.1.x branch for testing purposes

### DIFF
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -77,8 +77,10 @@
           current_file,
           %% current file handle since the last fsync?
           current_file_handle,
-          %% file handle cache
+          %% current write file offset
           current_file_offset,
+          %% messages that were potentially removed from the current write file
+          current_file_removes = [],
           %% TRef for our interval timer
           sync_timer_ref,
           %% files that had removes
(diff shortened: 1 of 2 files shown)


mkuratczyk:
Here are the results (after pushing the `Prioritise delete calls` commit):
https://grafana.lionhead.rabbitmq.com/goto/rkYSf0BHg?orgId=1

Observations:
* disk spaced is freed much faster
* much lower memory usage, which was unexpected
* in most cases, similar throughput

Slower compactions/scanning remains a problem, although with prioritised deletion requests, less so. Without this change, sometimes the node would be busy compacting files, minutes after queues were deleted and therefore, the files could have been deleted altogether. Now, such files will be deleted and never compacted.

Some highlights:
1. Disk space is reclaimed much faster (v4.1.x env, green, required 12 additional minutes to reclaim the space):
Screenshot 2025-05-29 at 13 18 01

2. Much lower memory usage throughout the tests
Screenshot 2025-05-29 at 13 19 27

gomoripeti:
thanks a lot @mkuratczyk, for the detailed tests and results, they are really reassuring to read

PR #13979, opened 2025-05-29
Log ranch timeout and tls errors

LoisSotoLopez:
## Proposed Changes

This PR addresses https://github.com/rabbitmq/rabbitmq-server/issues/11171 

The proposed changes catch the exit signal emitted by `ranch:handshake` in `rabbit_networking:handshake` and log an error message in case the reason for shutting down the connection is other than `closed`. 

The reason for not logging `closed` shutdown exit exceptions has been added as a context comment in the source.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbit/src/rabbit_networking.erl
+++ b/deps/rabbit/src/rabbit_networking.erl
@@ -33,7 +33,7 @@
          close_all_user_connections/2,
          force_connection_event_refresh/1, force_non_amqp_connection_event_refresh/1,
          handshake/2, handshake/3, tcp_host/1,
-         ranch_ref/1, ranch_ref/2, ranch_ref_of_protocol/1,
+         ranch_ref/1, ranch_ref/2, ranch_ref_of_protocol/1, ranch_ref_to_protocol/1,
          listener_of_protocol/1, stop_ranch_listener_of_protocol/1,
          list_local_connections_of_protocol/1]).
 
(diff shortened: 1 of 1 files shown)


deps/rabbit/src/rabbit_networking.erl
```diff
@@ -562,16 +577,36 @@ handshake(Ref, ProxyProtocolEnabled, BufferStrategy) ->
...
+        %% and closes a connection, as mentioned in
+        %% https://github.com/rabbitmq/rabbitmq-server/pull/12304
+        exit:{shutdown, {closed, _} = Reason} ->
+            exit({shutdown, Reason});
+        exit:{shutdown, {Reason, {PeerIp, PeerPort} = PeerInfo}} ->
+            PeerAddress = io_lib:format("~ts:~tp", [rabbit_misc:ntoab(PeerIp), PeerPort]),
+            Protocol = ranch_ref_to_protocol(Ref),
+            rabbit_log:error("~p error during handshake for protocol ~p and peer ~ts",
+                             [Reason, Protocol, PeerAddress]),
+            exit({shutdown, {Reason, PeerInfo}})
```
lhoguin, line 607: It's better in both catch cases to call `erlang:raise/3` so that we don't lose the stacktrace which may be propagated to other processes.

deps/rabbit/src/rabbit_networking.erl
```diff
@@ -562,16 +577,36 @@ handshake(Ref, ProxyProtocolEnabled, BufferStrategy) ->
...
-            {ok, Sock} = ranch:handshake(Ref),
+            Sock = try_ranch_handshake(Ref),
             ok = tune_buffer_size(Sock, BufferStrategy),
             {ok, Sock}
     end.
 
+try_ranch_handshake(Ref) ->
+    try ranch:handshake(Ref) of
+        {ok, Sock} ->
+            Sock
```
lhoguin, line 593: There's no reason no to return `{ok, Sock}`. You could also use the name `ranch_handshake` as the "try" doesn't tell us much.

michaelklishin:
The failures are not specific to this PR.

PR #13981, opened 2025-05-30
Management UI: fix MQTT qos0 queue [type] rendering

MarcialRosales:
## Proposed Changes

Fix issue rendering mqtt qos0 queue type.

Although there is a selenium test that reproduces the issue (selenium/test/queuesAndStreams/view-mqtt-qos0.js) these are the reproduction steps:
1. open an mqtt 5.0 connection with Session Expiry Interval of 0
2. subscribe to a quorum queue using QoS 0 
3. Go to the management ui
4. Go to the queue and Streams tab
5. Click on the queue whose type is `rabbit_mqtt_qos0_queue`
6. You get an exception 

This issue is only reproducible on `main`. 4.1.x uses a different mechanism to render this type of queue.

There is another PR, https://github.com/rabbitmq/rabbitmq-server/pull/14014, which backports all the changes relative to selenium except the fix which is only applicable to main.

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/main.js
+++ b/deps/rabbitmq_management/priv/www/js/main.js
@@ -1762,7 +1762,14 @@ function is_internal(queue) {
 }
 
 function get_queue_type (queue) {    
-    return queue.type;
+    switch(queue.type) {
+        case "classic":
+        case "quorum":
+        case "stream":
+            return queue.type;
+        default:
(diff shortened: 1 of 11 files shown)


michaelklishin:
The forced push was a rebase on top of `main` to benefit from a CI fix in #14018.

PR #13984, opened 2025-05-30
DO NOT MERGE: Revert https://github.com/rabbitmq/rabbitmq-server/pull/13950

mkuratczyk:
Revert the rabbit_fifo.erl change from https://github.com/rabbitmq/rabbitmq-server/pull/13950

We noticed a strange test flake(?) and it seems like it first happened after that PR.


mkuratczyk:
Same failure, despite the change reverted, so it's not that.

Issue #13785, opened 2025-04-24
MQTT consumer should get disconnected when its queue is deleted


mkuratczyk:
### Describe the bug

Queues are automatically declared for MQTT consumers. Currently we assume they don't get deleted externally (eg. manually by a user) but technically they can be. The consumer should be disconnected in such case, because it has no way of knowing this happened - from its perspective there are simply no messages to consume.

In RabbitMQ 3.11 the consumer was disconnected in such situations because that's what AMQP does. This behaviour changed with native MQTT, which doesn't use AMQP internally.

### Reproduction steps
```
> omq mqtt -r 1
...
2025/04/24 09:47:02 published=1/s consumed=1/s
2025/04/24 09:47:03 published=1/s consumed=1/s
2025/04/24 09:47:04 published=1/s consumed=1/s
2025/04/24 09:47:05 published=1/s consumed=1/s
2025/04/24 09:47:06 published=1/s consumed=1/s
2025/04/24 09:47:07 published=1/s consumed=0/s <---- rabbitmqct delete_queue ...
2025/04/24 09:47:08 published=1/s consumed=0/s
2025/04/24 09:47:09 published=1/s consumed=0/s
```

### Expected behavior

When a queue declared for an MQTT consumer is deleted, the consumer should be disconnected.

ansd:
Reported in https://groups.google.com/g/rabbitmq-users/c/3SuGHNAhylQ

vitalerter:
Hi
Just checking in on the status of this bug. Is there any plan or estimated version for when a fix might be released?

PR #13996, opened 2025-06-01
MQTT: disconnect consumer when queue is deleted

dcorbacho:
Queues are automatically declared for MQTT consumers, but they can be externally deleted. The consumer should be disconnected in such case, because it has no way of knowing this happened - from its perspective there are simply no messages to consume.

In RabbitMQ 3.11 the consumer was disconnected in such situation. This behaviour changed with native MQTT, which doesn't use AMQP internally.

Fixes https://github.com/rabbitmq/rabbitmq-server/issues/13785

### DIFF
--- a/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
+++ b/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
@@ -1984,7 +1984,7 @@ handle_down({{'DOWN', QName}, _MRef, process, QPid, Reason},
                 State ->
                     {ok, State}
             catch throw:consuming_queue_down ->
-                      {error, consuming_queue_down}
+                    {error, consuming_queue_down}
             end;
         {eol, QStates1, QRef} ->
             {ConfirmPktIds, U} = rabbit_mqtt_confirms:remove_queue(QRef, U0),
@@ -1992,12 +1992,25 @@ handle_down({{'DOWN', QName}, _MRef, process, QPid, Reason},
             State = State0#state{queue_states = QStates,
                                  unacked_client_pubs = U},
             send_puback(ConfirmPktIds, ?RC_SUCCESS, State),
-            {ok, State}
+            try handle_queue_down(QName, State) of
+                State ->
+                    {ok, State}
+            catch throw:consuming_queue_down ->
+                    {error, consuming_queue_down}
+            end
     end.
 
 -spec handle_queue_event(
         {queue_event, rabbit_amqqueue:name() | ?QUEUE_TYPE_QOS_0, term()}, state()) ->
     {ok, state()} | {error, Reason :: any(), state()}.
+handle_queue_event({queue_event, ?QUEUE_TYPE_QOS_0, {queue_down, QName}},
+                   State0) ->
+    try handle_queue_down(QName, State0) of
+        State ->
+            {ok, State}
+    catch throw:consuming_queue_down ->
+            {error, consuming_queue_down, State0}
+    end;
 handle_queue_event({queue_event, ?QUEUE_TYPE_QOS_0, Msg},
                    State0 = #state{qos0_messages_dropped = N}) ->
     State = case drop_qos0_message(State0) of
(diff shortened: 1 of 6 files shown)


michaelklishin:
@dcorbacho I have updated `web_mqtt_shared_SUITE` to propagate `notify_consumer_classic_queue_deleted` to `mqtt_shared_SUITE`.

This should address the `undef` failure.

dcorbacho:
Thank you @michaelklishin. I had totally forgotten about that suite even though I had to fix that very same issue on CI a few times before. I just added also `notify_consumer_quorum_queue_deleted` and `notify_consumer_qos0_queue_deleted`, now it should be complete

michaelklishin:
Per @dcorbacho, the CI failure has to do with an MQTT plugin test dependency (`rabbitmq_federation`) that unintentionally blocks cluster shutdown in a specific scenario.

michaelklishin (approved):
Tested with

```shell
mosquitto_sub -u guest -P guest -I mosquitto_sub -d -t rabbitmq-server-13996
```

and `rabbitmqctl list_queues`, `rabbitmqctl delete_queue`.

deps/rabbitmq_mqtt/src/rabbit_mqtt_qos0_queue.erl
```diff
@@ -126,6 +126,8 @@ delete(Q, _IfUnused, _IfEmpty, ActingUser) ->
     log_delete(QName, amqqueue:get_exclusive_owner(Q)),
     case rabbit_amqqueue:internal_delete(Q, ActingUser) of
         ok ->
+            Pid = amqqueue:get_pid(Q),
+            delegate:invoke_no_result([Pid], {gen_server, cast, [{queue_event, ?MODULE, {queue_down, QName}}]}),
```
ansd, line 130: Why is `delegate` being used here? To me this seems an overkill. Just calling `gen_server:cast` directly is good enough.

ansd: Can we emit an `eol` "end of life"  instead of `queue_down` event?
This way, we use a consistent naming with classic queues and quorum queues, and streams: They all emit an `eol` event if the queue is known to be deleted.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -2018,13 +2031,17 @@ handle_queue_event({queue_event, QName, Evt},
...
-            send_puback(ConfirmPktIds, ?RC_SUCCESS, State),
-            {ok, State};
+            try
+                State1 = handle_queue_actions(Actions ++ [{queue_down, QName}], State0),
+                {ConfirmPktIds, U} = rabbit_mqtt_confirms:remove_queue(QName, U0),
+                QStates = rabbit_queue_type:remove(QName, QStates0),
+                State = State1#state{queue_states = QStates,
+                                     unacked_client_pubs = U},
+                send_puback(ConfirmPktIds, ?RC_SUCCESS, State),
+                {ok, State}
```
ansd, line 2041: When is this code executed, given that the additional `queue_down` action will throw?

deps/rabbitmq_mqtt/src/rabbit_mqtt_qos0_queue.erl
dcorbacho: The `delegate` is there because it is replicating 0.9.1 behaviour. In this case, it ended up having just one pid to notify and I didn't realised it is unnecessary now.

dcorbacho: Same as above, in 0.9.1 the channel emits a queue_down event in this case

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
dcorbacho: Good point. It is executed for non consuming queues, but skipped for consuming ones. I guess the handing of queue_down should be done after sending the ack?

deps/rabbitmq_mqtt/src/rabbit_mqtt_qos0_queue.erl
ansd: > The delegate is there because it is replicating 0.9.1 behaviour

Yes, for AMQP 0.9.1 all protocol methods have to go through `delegate` such that the causal order of protocol methods from the client's point of view is guaranteed to stay the same (including the AMQP 0.9.1 `queue.delete` method).

ansd: > Same as above, in 0.9.1 the channel emits a queue_down event in this case

Where exactly is this happening? The fact that the channel handles 'DOWN' monitor messages is a relict from prior to the `rabbit_queue_type` API.
The more general direction after the `rabbit_queue_type` API got introduced is that an `eol` queue event/action denotes a queue deletion.

dcorbacho: https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_channel.erl#L2750

ansd: The line you are referencing is not executed when a classic queue got deleted. Instead the classic queue emits an `eol` queue action in https://github.com/rabbitmq/rabbitmq-server/blob/8e78c102c60397b936bf120eaf6b65aa217421f3/deps/rabbit/src/rabbit_classic_queue.erl#L417-L418

PR #13998, opened 2025-06-02
TLS: Hide password in config

dcorbacho:
See https://github.com/rabbitmq/rabbitmq-server/discussions/13958

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -233,7 +233,8 @@ end}.
 
 {translation, "rabbit.definitions.ssl_options.password",
 fun(Conf) ->
-    rabbit_cuttlefish:optionally_tagged_string("definitions.tls.password", Conf)
+    Pass = rabbit_cuttlefish:optionally_tagged_string("definitions.tls.password", Conf),
+    fun() -> Pass end
 end}.
 
 {mapping, "definitions.tls.secure_renegotiate", "rabbit.definitions.ssl_options.secure_renegotiate",


michaelklishin:
@dcorbacho I have considered this approach and concluded that it is likely to break when encrypted values are used in `rabbitmq.conf`.

I am talking a different approach in `rabbit_networking`:

``` erl
start_ssl_listener(Listener, SslOpts0, NumAcceptors, ConcurrentConnsSupsCount) ->
    SslOpts = case proplists:get_value(password, SslOpts0) of
                undefined -> SslOpts0;
                Password  ->
                  %% A password can be a value or a function returning that value.
                  %% See the key_pem_password/0 type in https://github.com/erlang/otp/pull/5843/files.
                  NewOpts = proplists:delete(password, SslOpts0),
                  Fun = fun() -> Password end,
                  [{password, Fun} | NewOpts]
              end,
    start_listener(Listener, NumAcceptors, ConcurrentConnsSupsCount, 'amqp/ssl',
                   "TLS (SSL) listener", tcp_opts() ++ SslOpts).
```

michaelklishin:
We have gone over the reason why I haven't touched the `rabbit.schema` on my branch with @dcorbacho. I will document it here for other contributors:

1. We [support encrypted values in `rabbitmq.conf`](https://www.rabbitmq.com/docs/configure#using-encrypted-values-in-rabbitmqconf-and-advancedconfig) in 4.x
2. After Cuttlefish generates a final `erlang.config` from `rabbitmq.conf` and `advanced.config`, the prelaunch app will decrypt all `{encrypted, …}` values and it cannot decrypt an `{encrypted, fun(() -> binary())`

Therefore we need to update the sites where these values are used for `ssl_options.password` and `definitions.tls.password`.

I will create a PR with that approach since I've had a branch ready.

PR #14006, opened 2025-06-02
Delete mqtt qos0 when connection is closed

MarcialRosales:
## Proposed Changes

Fix issue where an qos0 queue is left alive when using mqtt 5.0 and [Session Expiry Interval](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901048) of 0. 

## Types of Changes

- [ ] Build system and/or CI


deps/rabbitmq_mqtt/test/mqtt_shared_SUITE.erl
```diff
@@ -1583,6 +1586,18 @@ clean_session_disconnect_client(Config) ->
     L = rpc(Config, rabbit_amqqueue, list, []),
     ?assertEqual(0, length(L)).
 
+zero_session_expiry_interval_disconnect_client(Config) ->
+    C = connect(?FUNCTION_NAME, Config, [{properties, #{'Session-Expiry-Interval' => 0}}]),
```
ansd: This test case doesn't test the fix you provided.
In fact this test case succeeds even without your fix.

Specifically the bug is that an MQTT QoS 0 queue is not deleted by the MQTT plugin if an MQTT 5.0 client connects with CleanStart=false and a SessionExpiryInterval of 0.

Therefore, this test case should use `{clean_start, false}`.

I recommend moving this test case into the `v5_SUITE` since the combination of these two settings is an MQTT 5.0 specific feature.

deps/rabbitmq_mqtt/test/mqtt_shared_SUITE.erl
```diff
@@ -1583,6 +1586,18 @@ clean_session_disconnect_client(Config) ->
...
+    {ok, _, _} = emqtt:subscribe(C, <<"topic0">>, qos0),
+    QsQos0 = rpc(Config, rabbit_amqqueue, list_by_type, [rabbit_mqtt_qos0_queue]),
+    ?assertEqual(1, length(QsQos0)),
+
+    ok = emqtt:disconnect(C),
+    %% After terminating a clean session, we expect any session state to be cleaned up on the server.
+    timer:sleep(200), %% Give some time to clean up exclusive classic queue.
```
ansd: You copied this comment wrongly. It's not a classic queue that gets created, it's an MQTT QoS 0 queue.

ansd:
To be extra clear: The bug was that an MQTT QoS 0 queue was not deleted by the MQTT plugin for an MQTT 5.0 client that connected with Clean Start 0 **and** Session Expiry Interval 0.

deps/rabbitmq_mqtt/test/mqtt_shared_SUITE.erl
MarcialRosales: Thanks @ansd I was following the 3 conditions in the docs https://www.rabbitmq.com/docs/mqtt#qos0-queue-type and it does not mention `clean_start` setting .. should it also be mentioned in the docs?

I started doing this test in v5_SUITE but then I found mqtt_shared_SUITE and I thought this suite would be the best place because both versions are tested ...

ansd: > should it also be mentioned in the docs?

No, the docs are correct.

If anything, we can delete the 1st condition since it's no longer relevant. This feature flag has become required: https://github.com/rabbitmq/rabbitmq-website/pull/2284

MarcialRosales: I am reading (http://www.steves-internet-guide.com/mqttv5-clean-start-clean-sessions-and-session-expiry/) that `clean_start` is used in conjunction with `session expiry interval` . why do you think the documentation should not mention `clean_start`  in the conditions?

ansd: > why do you think the documentation should not mention clean_start in the conditions?

If no queue exists for an MQTT 5.0 client, and the client connects with a Session Expiry Interval of 0 subsequently subscribing with QoS 0, the MQTT plugin will create an MQTT QoS 0 queue for that client irrespective of whether the client connected with Clean Start 0 or 1.

deps/rabbitmq_mqtt/test/v5_SUITE.erl
```diff
@@ -389,6 +401,21 @@ session_expiry_quorum_queue_disconnect_decrease(Config) ->
...
+    C = connect(ClientId, Config, [
+        {clean_start, false},
+        {properties, #{'Session-Expiry-Interval' => 0}}]),
+    {ok, _, _} = emqtt:subscribe(C, <<"topic0">>, qos0),
+    QsQos0 = rpc(Config, rabbit_amqqueue, list_by_type, [rabbit_mqtt_qos0_queue]),
+    ?assertEqual(1, length(QsQos0)),
+
+    ok = emqtt:disconnect(C),
+    %% After terminating a clean session, we expect any session state to be cleaned up on the server.
+    timer:sleep(200), %% Give some time to clean up exclusive classic queue.
```
ansd: Can we fix this comment? It's not an ` exclusive classic queue` that gets cleaned up, it's an MQTT QoS 0 queue.

ansd (changes requested):

deps/rabbitmq_mqtt/test/v5_SUITE.erl
michaelklishin: Done

michaelklishin:
I have addressed the comment and trailing whitespace feedback.

ansd:
Thank you for this fix @MarcialRosales !

PR #14036, opened 2025-06-05
CI: workflow to build container image nightly

Zerpet:
## Proposed Changes

It's a middle ground between building on every commit, and not building
at all. We currently have a workflow to build the OCI on every PR.
However, building on every PR does not cover some use cases. For
example, providing an image to a user to preview some changes coming in
the next minor or patch. Another use case: compare main with a PR in
Kubernetes.

It's better to have a separate workflow, even at the expense of
duplication, because the "on schedule" trigger only runs on the default
branch of the repository. This "limitation" makes it complicated to
extend the current "build OCI on PRs" to also build nightly for main and
release branches.

## Types of Changes

- [x] Build system and/or CI

### DIFF
--- a/.github/workflows/oci-make-nightly.yaml
+++ b/.github/workflows/oci-make-nightly.yaml
@@ -1,0 +1,117 @@
+name: Nightly OCI (make)
+on:
+  schedule:
+    # at 2:20am Mon-Fri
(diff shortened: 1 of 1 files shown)


Zerpet:
This doesn't need backporting because "on schedule" only runs on the default branch i.e. `main`

PR #14039, opened 2025-06-05
Mqtt tests: start just required dependencies

dcorbacho:
MQTT tests depend on a few plugins, which are just used in 1 or 2 suites each. These have caused issues in CI, triggering a bug in rabbitmq_federation where the mirrored supervisor submits a transaction while the cluster is being shut down. The transaction hangs and the whole rabbitmq_mqtt job times out.

This bug has been addressed, however it is best to start just the required plugins on each SUITE.


michaelklishin:
@dcorbacho the `web_mqtt` failures seem repeatable.

dcorbacho:
Depends on https://github.com/rabbitmq/rabbitmq-server/pull/14057

PR #14060, opened 2025-06-11
Avoid using the size/1 BIF

ansd:
Avoid using the size/1 BIF for performance critical code because according to
https://whatsapp.github.io/erlang-language-platform/docs/erlang-error-index/w/W0050/ 
> The BIF is not optimized by the JIT


michaelklishin:
Good catch.

PR #14079, opened 2025-06-14
Like internal/protected queues, but shovels

ikavgo:
Introduces optional "internal" and "internal_owner" parameters. Also adds "--force" flag to Delete Shovel command.

Management integration is yet to be done.


ikavgo:
green!

michaelklishin (approved):

PR #14086, opened 2025-06-16
WIP: Add per-queue-type disk limits

the-mikedavis:
This is an extension of the idea of the free disk space alarm. That option enables blocking publishers when the free disk space on the data dir's disk falls below some threshold. This feature blocks publishers of individual queue types when the disk space taken by the queue type exceeds the configured threshold.

This change is incomplete: it only affects QQs and streams and AMQP 0-9-1 so far. I thought I'd open this as a draft early though to hear if anyone has thoughts on this as a feature.

<details><summary>How to demo the changes...</summary>

1. `make run-broker RABBITMQ_CONFIG_FILE=example.conf`
2. In that shell start the monitor `rabbit_queue_type_disk_monitor:start_link()` (should become part of a sup tree eventually)
3. Start a QQ producer/consumer `perf-test -qq -u qq -x 1 -y 1 --rate 5 --confirm 5`
4. Start a stream producer/consumer `perf-test -sq -u sq -x 1 -y 1 --rate 5 --confirm 5`
5. Go to the QQ data dir: `cd /tmp/rabbitmq-test-instances/<node>/mnesia/<node>/quorum/<node>/`
6. Create a big file larger than the QQ disk limit: `dd if=/dev/zero of=ballast.bin bs=1G count=3`
7. Wait a few seconds and notice that the QQ perf-test command has stopped but not the stream one.
8. `rm ballast.bin` and the QQ perf-test command should continue

---

</details>

### DIFF
--- a/deps/rabbit/src/rabbit_disk_usage.erl
+++ b/deps/rabbit/src/rabbit_disk_usage.erl
@@ -1,0 +1,76 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
+%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%
+%% Copyright (c) 2007-2025 Broadcom. All Rights Reserved. The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries. All rights reserved.
+%%
+
+-module(rabbit_disk_usage).
+
+%% Functions for calculating disk usage of a given directory.
+
+-include_lib("kernel/include/file.hrl").
+
+-export([scan/1]).
+
+%% @doc Calculates the disk usage in bytes of the given directory.
+%%
+%% On especially large directories this can be an expensive operation since
+%% each sub-directory is scanned recursively and each file's metadata must be
+%% read.
+-spec scan(Dir) -> {ok, Size} | {error, Error} when
+    Dir :: filename:filename_all(),
+    Size :: non_neg_integer(),
+    Error :: not_directory | file:posix() | badarg.
+scan(Dir) ->
+    case file:read_file_info(Dir) of
+        {ok, #file_info{type = directory, size = S}} ->
+            {ok, Gatherer} = gatherer:start_link(),
+            scan_directory(Dir, Gatherer),
+            Size = sum(Gatherer, S),
+            gatherer:stop(Gatherer),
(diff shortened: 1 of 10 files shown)


ikavgo:
I like this.
I would change configuration from `quorum_queue_disk_limit.absolute = 2GiB` to `queue_disk_limit.quorum.absolute = 2GiB` and work together with type registry. 

I wonder if it is possible to set such limits via policies and configs for individual queues? 
If this alarm reported via status() we could also highlight blocked queues in the Management UI.

deps/rabbit/src/rabbit_disk_usage.erl
```diff
@@ -0,0 +1,76 @@
...
+
+-include_lib("kernel/include/file.hrl").
+
+-export([scan/1]).
+
+%% @doc Calculates the disk usage in bytes of the given directory.
+%%
+%% On especially large directories this can be an expensive operation since
+%% each sub-directory is scanned recursively and each file's metadata must be
+%% read.
```
lhoguin, line 20: Might be better for queue types to increment/decrement an internal counter (that may be stored in persistent_term).

kjnilsson:
> This is an extension of the idea of the free disk space alarm. That option enables blocking publishers when the free disk space on the data dir's disk falls below some threshold. This feature blocks publishers of individual queue types when the disk space taken by the queue type exceeds the configured threshold.
> 
> This change is incomplete: it only affects QQs and streams and AMQP 0-9-1 so far. I thought I'd open this as a draft early though to hear if anyone has thoughts on this as a feature.
> 
> How to demo the changes...

I am wondering if it would be better (by some definition of better) that if you want this type of feature the different data directories were mounted as separate disk volumes which should make usage querying a lot simpler and faster.

Also making this queue type specific isn't necessarily always going to work. For example we have experimented with running multiple quorum queue Ra systems which would use different data directories (ideally on different volumes). In this case it it is much harder to work out which channels etc related to particular quorum queues in different systems. Not impossible but not a straight map to a queue type.

the-mikedavis:
Thanks for the quick feedback on this! Sorry it took me so long to get back to it. I have opened up https://github.com/rabbitmq/rabbitmq-server/discussions/14590 to discuss it further and I will continue in a new branch, most likely with the `df`-like approach used by `rabbit_disk_monitor` currently.

Issue #13957, opened 2025-05-27
AMQP 1.0 Application Properties lost when shovelling to an AMQP 0.9.1 broker


DHoskin-Aurizon:
### Describe the bug

RabbitMQ Shovel currently does not keep the AMQP 1.0 Application Properties and AMQP 0.9.1 Properties/Headers in sync when copying messages between exchanges or queues when transferring messages between AMQP 1.0 and AMQP 0.9.1 brokers as per https://www.rabbitmq.com/docs/conversions

Properties/Headers are retained when shovelling from AMQP 0.9.1 -> AMQP 1.0 but are lost when shovelling from AMQP 1.0 -> AMQP 0.9.1 in RabbitMQ V4.*.

Properties/Headers are retained in both directions in RabbitMQ V3.12.11+ and V3.13.5+.

### Reproduction steps

1. Set up a shovel from AMPQ 1.0 to AMQP 0.9.1
2. Publish a message to the AMQP 1.0 broker with Application Properties
3. The received message on the AMQP 0.9.1 broker has no Properties/Headers

### Expected behavior

The message received from the shovel contains the converted Properties/Headers.

### Additional context

Refer to https://github.com/rabbitmq/rabbitmq-server/issues/7508 and https://github.com/rabbitmq/rabbitmq-server/pull/11729

ansd:
As explained in https://github.com/rabbitmq/rabbitmq-server/pull/11729#issuecomment-2911345020, the solution for RabbitMQ 4.x should use the `rabbit_queue_type` interface directly and should convert between protocols using the `mc` (message container) module.

Advantages:

* the well documented conversions in https://www.rabbitmq.com/docs/conversions are followed
* no duplication of conversion rules how messages are converted from AMQP 0.9.1 to AMQP 1.0 (or vice versa)

DHoskin-Aurizon:
Hi @ansd, my apologies I misunderstood your prior message. I thought you were referring to a different method of fixing the bug, not that there already exists different functionality to achieve this.

Can you please direct me to the documentation on how to set this up as I have search the RabbitMQ docs and tried a few different configurations with no success.

Below is my setup:
* RabbitMQ 4.1.0
* Messages generated in Azure Service Bus with Broker Properties (correlationID) & Custom Properties (Type, Version and routing_key) (AMQP 1.0)
* RabbitMQ shovel used to read from Service Bus and route messages to a RabbitMQ queue.
[...]

PR #14096, opened 2025-06-18
Shovel: use message containers

dcorbacho:
See #13957


dcorbacho:
I think the only breaking change is the translation of `type` from amqp0.9.1 into the message annotation`x-basic-type` in amqp1.0, as translated by our message container implementation. Previously it was translated to `type` in the application properties, which is probably a bug. See https://www.rabbitmq.com/docs/conversions#amqpl-amqp

ansd (commented):
When a message is shovelled from RabbitMQ to another AMQP 1.0 broker, I would expect the following to happen:

RabbitMQ **encodes** and sends the message.

This can be achieved by using the `rabbit_queue_type` API directly, as suggested by @kjnilsson initially and as described in https://github.com/rabbitmq/rabbitmq-server/issues/13957#issuecomment-2911764754: The shovel plugin would consume directly from the queue, encode and then send the message.

If I read the code correctly, the following happens with this PR:

RabbitMQ **encodes** the message -> sends the message to shovel plugin -> shovel plugin **decodes** the message in the AMQP 1.0 client -> shovel plugin **encodes** the message to create a message container -> `mc_amqp` on the shovel plugin then **decodes** the message ->  in `rabbit_amqp10_shovel.erl:forward/3` the shovel plugin then **encodes** the message  -> the shovel plugin then **decodes** the message for a format suitable by the AMQP 1.0 client, the AMQP 1.0 client then **encodes** the message before finally sending it.

Therefore, with this PR there are 7 (!) different encode/decode steps instead of the final goal to have only 1 step (and instead of 3 steps prior to this PR).

So, while this PR uses message containers for conversions and therefore solves one problem, it seems that performance gets worse compared to prior to this PR.

Have you done any performance tests?

That said, it's up to you to decide whether we want to accept a potentially severe performance drop as an interim solution until using `rabbit_queue_type` API.

dcorbacho:
@ansd How we are going to use the `rabbit_queue_type` API directly and skip the client? Unless I'm missing something, for the queue type to be used the shovel should be in the same cluster.

deps/amqp10_client/src/amqp10_msg.erl
```diff
@@ -270,14 +270,23 @@ new(DeliveryTag, Bin, Settled) when is_binary(Bin) ->
...
+               %% This lib is safe by default.
+               header = #'v1_0.header'{durable = true},
+               body = Body};
+        false ->
+            Transfer = #'v1_0.transfer'{
+                          delivery_tag = {binary, DeliveryTag},
+                          settled = Settled,
+                          message_format = {uint, ?MESSAGE_FORMAT}},
+            from_amqp_records([Transfer | Body])
+    end.
```
ansd: The function spec is now wrong: The function spec expects currently a body, but not other sections.

ansd: The creation of the `'v1_0.transfer'` record got duplicated. Avoid code duplication.

ansd (changes requested):

ansd:
> How we are going to use the rabbit_queue_type API directly and skip the client? Unless I'm missing something, for the queue type to be used the shovel should be in the same cluster.

Correct, the assumption of using `rabbit_queue_type` API directly is that either the shovel source or destination is located in the cluster where the shovel runs. This is the common case. Shovels usually forward broker local messages to a remote broker, or the other way around from a remote broker to the local broker. In both cases, using `rabbit_queue_type` API directly will greatly speed up performance.

The case where `rabbit_queue_type` cannot be used at all is if the shovel runs on neither the source nor the destination, but on some 3rd broker. This is a rather uncommon setup.

kjnilsson:
> Correct, the assumption of using `rabbit_queue_type` API directly is that either the shovel source or destination is located in the cluster where the shovel runs. This is the common case. Shovels usually forward broker local messages to a remote broker, or the other way around from a remote broker to the local broker. In both cases, using `rabbit_queue_type` API directly will greatly speed up performance.

I think that would need to be a new "internal" shovel protocol impl which may be slightly beyond the scope of this PR. But I agree that is where we _want_ to be.

dcorbacho:
Superseeded by https://github.com/rabbitmq/rabbitmq-server/pull/14256

PR #14100, opened 2025-06-20
rabbit_fifo: Activate next SAC deterministically

the-mikedavis:
This change extends the changes in #13971 (making `rabbit_fifo` deterministic). Determining the next active consumer uses a `maps:iterator()` which has undefined ordering. Like calls to `fold/3` or `keys/1` we should ensure ordering of the iterator so that two replicas are guaranteed to find the same consumer.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo_maps.erl
+++ b/deps/rabbit/src/rabbit_fifo_maps.erl
@@ -8,7 +8,8 @@
 -module(rabbit_fifo_maps).
 
 -export([keys/2,
-         fold/4]).
+         fold/4,
+         iterator/2]).
 
 -spec keys(Map, ra_machine:version()) -> Keys when
     Map :: #{Key => _},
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_fifo_maps.erl
```diff
@@ -29,13 +30,18 @@ keys(Map, Vsn) ->
...
     Map :: #{Key => Value}.
 fold(Fun, Init, Map, Vsn)  ->
-    Iterable = case is_deterministic(Vsn) of
-                   true ->
-                       maps:iterator(Map, ordered);
-                   false ->
-                       Map
-               end,
-    maps:fold(Fun, Init, Iterable).
+    maps:fold(Fun, Init, iterator(Map, Vsn)).
```
the-mikedavis, line 33: Note that `maps:fold/3` creates an iterator with `maps:iterator/1` by default, so this is equivalent to the old code which passed a map directly: https://github.com/erlang/otp/blob/d4c78e0414664564ae29d4d6a09bbaf7d2c2110b/lib/stdlib/src/maps.erl#L881-L882

the-mikedavis:
Aha I see now, this path is only used for the single-active consumer. I think I see now why @kjnilsson mentioned that using an iterator here is odd 😅. I assume it makes sense performance-wise to find the SAC this way.

I'll close this out then - this is already deterministic since there can be only one SAC. Thanks @ansd!

ansd:
Yes 🙂 Thank you though for following because it's always good to double check these different places where map traversal with undefined order in the Ra machine's `apply` callback takes place. For example, there are also some places in https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_stream_coordinator.erl and https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_stream_sac_coordinator.erl which use undefined map iteration and therefore look non-deterministic, however we discussed internally that they are all fine. JSP also checked the Ra state machine in Khepri IIRC.

PR #14114, opened 2025-06-24
Improve Erlang/OTP 28 compatibility, use OTP 28 in the pipelines

mkuratczyk:
Test with OTP28


michaelklishin:
This also means that we will have to produce Elixir 1.18 Debian packages sooner rather than later but that's OK.

michaelklishin:
Debian and RPM packages now mark Erlang/OTP 28 as supported (for RabbitMQ 4.2 only): https://github.com/rabbitmq/rabbitmq-packaging/pull/57

PR #14178, opened 2025-07-02
CQ: Retry opening write file when writing messages

lhoguin:
Followup to ff8ecf1cf7cfd22981668cbed374a5572560dd80 only this time it's for the index.


michaelklishin:
Thanks to @tomerrt, we have learned that this filesystem API behavior is not only Windows-specific, it is specific to a certain set of Windows versions, and the newest versions
[should not be affected](https://github.com/rabbitmq/rabbitmq-server/discussions/14231#discussioncomment-13769535).

PR #14190, opened 2025-07-04
Very minor: Remove outdated comment from rabbit_msg_store

gomoripeti:
## Proposed Changes

Sorry for providing a PR instead of asking the question: Is that correct that "The value `close` is never stored in the file handle ets since commit 32816c0a."?

I found this comment while investigating https://github.com/rabbitmq/rabbitmq-server/discussions/14181. As far as I can see the value can never be `close` so it cannot be the cause of the bug.

During the same investigation I got another question. Comment says (https://github.com/rabbitmq/rabbitmq-server/blob/v4.1.2/deps/rabbit/src/rabbit_msg_store.erl#L520-L522)
> Get index for all Msgids in File.
> It's possible that we get no results here if compaction
> was in progress. That's OK: we will try again with those
> MsgIds to get them from the new file.

But compaction (as opposed to combination) doesn't create new files. Only moves messages within the same file. How can a message be moved to a "new file" (assuming it has positive refcount)?

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -1498,8 +1498,7 @@ writer_close(#writer{fd = Fd}) ->
     file:close(Fd).
 
 mark_handle_open(FileHandlesEts, File, Ref) ->
-    %% This is fine to fail (already exists). Note it could fail with
-    %% the value being close, and not have it updated to open.
+    %% This is fine to fail (already exists).
     ets:insert_new(FileHandlesEts, {{Ref, File}, erlang:monotonic_time()}),
     true.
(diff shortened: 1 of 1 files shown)


michaelklishin:
@lhoguin any objections from you?

lhoguin (approved):

lhoguin:
> But compaction (as opposed to combination) doesn't create new files. Only moves messages within the same file. How can a message be moved to a "new file" (assuming it has positive refcount)?

Not only that but compaction now doesn't prevent reading from the file at all so we should not reach the `[]` clause anymore.

PR #14210, opened 2025-07-10
Minor: Don't use Mnesia in rabbitmq_mqtt/test/processor_SUITE

gomoripeti:
## Proposed Changes

Soon Mnesia will be gone from RabbitMQ, so better make the test suite metadata store agnostic. (The suite contains unit tests so mocking on a higher level seemed acceptable)

Also refactored mqtt_processor get_vhost functions, in order to clarify preference of different methods. This commit does
not change functionality. (I wanted to prepare an actual change in this area but abandoned the idea. But I think this refactoring makes the code a bit more readable

This highlighted some differences (which I initially thought as inconsistencies, but after digging the history they seem to be intentional and the behaviour is documented in the processor_SUITE)

- If both User/Password and SslLogin are provided, in
  `check_credentials` User/Password takes precedence while in
  `get_vhost` SslLogin

- If SslLogin is provided (but no mapping is found) vhost from port
  mapping has precedence over vhost from UserName, while in case of
  no ssl it is the other way around.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
+++ b/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
@@ -1185,33 +1185,31 @@ get_vhost(UserBin, SslLogin, Port) ->
     get_vhost_ssl(UserBin, SslLogin, Port).
 
 get_vhost_no_ssl(UserBin, Port) ->
-    case vhost_in_username(UserBin) of
-        true  ->
-            {vhost_in_username_or_default, get_vhost_username(UserBin)};
-        false ->
-            PortVirtualHostMapping = rabbit_runtime_parameters:value_global(
-                                       mqtt_port_to_vhost_mapping
-                                      ),
-            case get_vhost_from_port_mapping(Port, PortVirtualHostMapping) of
+    case get_vhost_username(UserBin) of
+        undefined ->
+            case get_vhost_from_port_mapping(Port) of
                 undefined ->
-                    {plugin_configuration_or_default_vhost, {rabbit_mqtt_util:env(vhost), UserBin}};
-                VHost ->
-                    {port_to_vhost_mapping, {VHost, UserBin}}
-            end
+                    VhostFromConfig = rabbit_mqtt_util:env(vhost),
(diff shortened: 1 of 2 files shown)


deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -1185,65 +1185,58 @@ get_vhost(UserBin, SslLogin, Port) ->
...
                 [Vhost, UserName] -> {Vhost,  UserName};
-                [UserBin]         -> Default;
-                []                -> Default
+                [UserBin]         -> undefined;
+                []                -> undefined
             end
     end.
 
+get_vhost_from_user_mapping(none) ->
+    undefined;
```
ansd: Looks like this is dead code?

deps/rabbitmq_mqtt/test/processor_SUITE.erl
```diff
@@ -31,38 +31,27 @@ suite() ->
 
 init_per_suite(Config) ->
     ok = application:load(rabbitmq_mqtt),
+    meck:new(rabbit_runtime_parameters, [passthrough, no_link]),
```
ansd, line 34: If this mock is needed only for test case `get_vhost`, then this line should be moved into the `init_per_testcase` for that single test.

gomoripeti: the second commit makes it needed in `ignores_colons_in_username_if_option_set` as well.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
gomoripeti: yes dead code, originally wanted to merge `get_vhost_no_ssl` and `get_vhost_ssl`, it's a leftover from that. thanks for noticing, deleting

ansd:
Thank you @gomoripeti, the code is simpler to read now after your refactor!

PR #14232, opened 2025-07-15
Consistent logging

mkuratczyk:
Make our logging more consistent:
1. get rid of rabbit_log, rabbit_shovel_log and similar modules
2. use logger macros directly
3. remove `pid => self()` metadata, since that's automatically added by logger
4. instead of passing the logging domain on every call, set the domain on the process level

The domain should be set on more processes, but we can do that over time. The fact they are missing shouldn't affect anything anyway, since currently we don't log the domain on the logger level, but rather many log strings contain the domain "informally" (eg. peer discovery sets the domain but also prefixes all logs with "Peer discovery:")


deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -129,7 +130,7 @@
 -define(DEFAULT_DELIVERY_LIMIT, 20).
 
 -define(INFO(Str, Args),
-        rabbit_log:info("[~s:~s/~b] " Str,
+        ?LOG_INFO("[~s:~s/~b] " Str,
```
ansd, line 133: The `INFO` macro should be deleted given that the `LOG_INFO` macro already includes module, function name, and arity.

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -332,7 +333,7 @@ consume(Q, Spec, #stream_client{} = QState0)
               args := Args,
               ok_msg := OkMsg,
               acting_user := ActingUser} = Spec,
-            rabbit_log:debug("~s:~s Local pid resolved ~0p",
+            ?LOG_DEBUG("~s:~s Local pid resolved ~0p",
```
ansd, line 336: The `LOG_DEBUG` already includes module, function name, and arity. There is no need to pass the module and function name again as parameter to the format string.

deps/rabbit/include/rabbit_amqp.hrl
```diff
@@ -3,7 +3,7 @@
 -ifdef(TRACE_AMQP).
 -warning("AMQP tracing is enabled").
 -define(TRACE(Format, Args),
-        rabbit_log:debug(
+        ?LOG_DEBUG(
```
ansd, line 6: Same here

ansd (commented):
There are many places where information such as module, function name, arity, line number are now passed into the format string **in addition to** the logger macros included by file `logger.hrl` providing this same information:
https://github.com/erlang/otp/blob/c8ee2093fce4766d52e4f246123818d2273bee92/lib/kernel/include/logger.hrl#L61-L63

Should this PR remove the now duplicated provided infos from the format strings?

EDIT: let's do these changes in a follow up PR

ansd (approved):

PR #14237, opened 2025-07-15
rabbitmq-queues: a command to display a member with highest (commit, log, snapshot) index

Ayanda-D:
## Proposed Changes

Hi RMQ team 👋 

We're still facing severe problems with Quorum Queues getting into a non-responsive state with no leader https://github.com/rabbitmq/rabbitmq-server/discussions/13101. To recover these queues from a bad state (and to avoid deleting them and losing messages), we'd like to have the following capabilities to:
1. Quickly pick the member/node with the highest index (`commit|log||snapshot`) - to recover to/from. This PR.
2. [Force shrink the damaged quorum queue](https://github.com/rabbitmq/rabbitmq-server/pull/12427) to the member with the highest chosen index from step-1.
3. Recover/[re-grow the quorum queue](https://github.com/rabbitmq/rabbitmq-server/pull/13873) to the target-quorum-cluster size (PR open)

The changes proposed here provide the CLI tool to be used for step-1 above, a command to get the `member_with_highest_index` to use for recovery, for a given QQ e.g:

```
ayandad@rmachine-1> rabbitmq-queues member_with_highest_index "Q.2" --index commit
Member with highest commit index for queue Q.2 in vhost / on node rabbit@rmachine-1...
┌───────────────────┬────────────┬────────────┬────────────────┬──────────────┬──────────────┬──────────────┬────────────────┬──────┬─────────────────┐
│ Node Name         │ Raft State │ Membership │ Last Log Index │ Last Written │ Last Applied │ Commit Index │ Snapshot Index │ Term │ Machine Version │
├───────────────────┼────────────┼────────────┼────────────────┼──────────────┼──────────────┼──────────────┼────────────────┼──────┼─────────────────┤
│ rabbit@rmachine-2 │ leader     │ voter      │ 17             │ 17           │ 17           │ 17           │ -1             │ 13   │ 7               │
└───────────────────┴────────────┴────────────┴────────────────┴──────────────┴──────────────┴──────────────┴────────────────┴──────┴─────────────────┘

```

With this procedure, we can confirm that we able to restore the broken quorum queues back into a usable state, with all messages retained (depending on the member selected with the highest log/commit index - retaining all in the log). Internally, this using the queue's quorum-status to acquire the member highest index (seemed the least complex approach).

Please take a look - we'd really appreciate having these tools available. Currently, this is the only way we are able to fix/recover/revive broken Quorum Queues which are seeing quite regularly, back to a usable state. These procedure(s) all working well on our end, and restoring service as expected.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -68,7 +68,8 @@
          notify_decorators/3,
          spawn_notify_decorators/3]).
 
--export([get_member_with_highest_index/4]).
+-export([get_member_with_highest_index/3,
+         get_member_with_highest_index/4]).
 
 -export([is_enabled/0,
          is_compatible/3,
@@ -1336,6 +1337,11 @@ get_sys_status(Proc) ->
             {error, other}
 
     end.
+
+-spec get_member_with_highest_index(rabbit_types:vhost(), rabbit_misc:resource_name(), atom()) ->
+    [[{binary(), term()}]] | {error, term()}.
+get_member_with_highest_index(Vhost, QueueName, IndexName) ->
+    get_member_with_highest_index(Vhost, QueueName, IndexName, false).
 
 -spec get_member_with_highest_index(rabbit_types:vhost(), rabbit_misc:resource_name(), atom(), boolean()) ->
     [[{binary(), term()}]] | {error, term()}.


deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1335,6 +1337,38 @@ get_sys_status(Proc) ->
...
+                        lists:foldl(
+                            fun(Entry, {PreviousIndexValue, _PreviousEntry} = Acc) ->
+                                case rabbit_misc:pget(Key, Entry) of
+                                    CurrentIndexValue when is_integer(CurrentIndexValue),
+                                                           CurrentIndexValue > PreviousIndexValue ->
+                                        {CurrentIndexValue, Entry};
+                                    _ ->
+                                        Acc
+                                end
+                            end, {-100, []}, Status),
```
Ayanda-D, line 1369: just FYI, the `-100` here is just an arbitrary number to initialise the accumulator, to ensure its less than the initial index value of `-1`, e.g. initial snapshot index = `-1`: https://github.com/rabbitmq/ra/blob/v2.16.11/src/ra_snapshot.erl#L195

michaelklishin:
@Ayanda-D there's nothing wrong with having this command. Can we rename it to just `member_with_highest_index`, though.

Ayanda-D:
@michaelklishin ok, i've updated name to `member_with_highest_index`

kjnilsson:
> We're still facing severe problems with Quorum Queues getting into a non-responsive state with no leader #13101.

Do you still have this issue on a supported OSS version (i.e. 4.1)?

Ayanda-D:
hi @kjnilsson (sorry for the delay) we are seeing this on 3.12.14. However, the issue seems very closely related to https://github.com/rabbitmq/ra/discussions/514 and https://github.com/rabbitmq/rabbitmq-server/discussions/13131 (which seem to have not been resolved yet?) - we see logs flooded with the same crash when the QQs get into this leader-less state.

Then, I managed to to produce a very similar state from `main` branch https://github.com/rabbitmq/rabbitmq-server/commit/cdd9ba1bb6a82f33ff3eb52aa51a8f986d88d0f8 (using same procedures listed here https://github.com/rabbitmq/rabbitmq-server/discussions/13131 + full broker restart):

```
ayandad@host rabbit % sbin/rabbitmq-queues quorum_status qq1 --node rabbit-1
Status of quorum queue qq1 on node rabbit-1@host ...
┌─────────────────────┬────────────┬────────────┬────────────────┬──────────────┬──────────────┬──────────────┬────────────────┬──────┬─────────────────┐
│ Node Name           │ Raft State │ Membership │ Last Log Index │ Last Written │ Last Applied │ Commit Index │ Snapshot Index │ Term │ Machine Version │
├─────────────────────┼────────────┼────────────┼────────────────┼──────────────┼──────────────┼──────────────┼────────────────┼──────┼─────────────────┤
│ rabbit-1@host       │ noproc     │ unknown    │ 976            │ 976          │ 976          │ 976          │ 974            │ 977  │ 7               │
├─────────────────────┼────────────┼────────────┼────────────────┼──────────────┼──────────────┼──────────────┼────────────────┼──────┼─────────────────┤
│ rabbit-2@host       │ noproc     │ unknown    │ 977            │ 977          │ 976          │ 976          │ 975            │ 978  │ 7               │
├─────────────────────┼────────────┼────────────┼────────────────┼──────────────┼──────────────┼──────────────┼────────────────┼──────┼─────────────────┤
│ rabbit-3@host       │ pre_vote   │ voter      │ 10006          │ 10006        │ 10006        │ 10006        │ 10006          │ 978  │ 7               │
└─────────────────────┴────────────┴────────────┴────────────────┴──────────────┴──────────────┴──────────────┴────────────────┴──────┴─────────────────┘
```

We are ntill not precisely sure whats causing this, hence these workarounds to at least restore such queues.

kjnilsson:
> Then, I managed to to produce a very similar state from `main` branch [cdd9ba1](https://github.com/rabbitmq/rabbitmq-server/commit/cdd9ba1bb6a82f33ff3eb52aa51a8f986d88d0f8) (using same procedures listed here #13131 + full broker restart):

Do you delete queues whilst a node is down then re-declare them with the same name?

This is a very specific and a bit odd use case (why delete and re-declare queues with the same name?) that is causing the issues you mention.

I just feel this utility is a bit odd as it takes seconds to work this out using the existing `rabbitmq-queues quorum_status` command and it feels like it is addressing issues in unsupported RabbitMQ versions.

Ayanda-D:
> Do you delete queues whilst a node is down then re-declare them with the same name?

Not really, e.g. an example is routine sanity checks that run periodically, sometimes when one or more nodes are undergoing restarts (even if `queue.delete` and `queue.declare` are carried out between - for same queue name). We expect Quorum Queues to remain stable in such cases - but they can be in this leaderless state indefinitely (until deletion).

We are also not always in full control of what thousands of client applications do. Same SLA guarantees need to be on Quorum Queues as with Classic Mirrored Queues for general AMQP operations, despite behind the scenes operations such as node restarts taking place with majority of nodes always available (we understand there are some unsupported features, but getting into a state of potential message loss / queue deletion is a blocker).

> I just feel this utility is a bit odd as it takes seconds to work this out using the existing

Doing the computation on the node is still the fastest, we just loop through `ClusterSize` times (which is typically 5 or 7) to get the member with highest index - time incurred (which should be a few milliseconds) should be sufficient on the operator end - compared to doing this from external scripts, parsing outputs, etc which is much slower (I'm happy to change this for a faster suggested approach? we desperately need this or similar). 

> it feels like it is addressing issues in unsupported RabbitMQ versions

We dont have specific references addressing these problems to justify that upgrading will fix this issue:
- https://github.com/rabbitmq/ra/discussions/514 
- https://github.com/rabbitmq/rabbitmq-server/discussions/13131
- https://github.com/rabbitmq/rabbitmq-server/discussions/13101

And also managed to produce the same state from `main` branch. This tool to help recover the queue, even from what we're observing from latest `main` branch.

kjnilsson:
> We dont have specific references addressing these problems to justify that upgrading will fix this issue:

There is very little evidence linking any of what you have seen to these issues. We have however fixed _many_ other issues that could result in broken Ra clusters since 3.12. Too many to list. Upgrading to the latest 4.1.x release _will_ help. Also it would help us as we can then justify spending time investigating any further issues you see.

Ayanda-D:
We understand there is no clear path to fixing this issue - hence we are adding/contributing operator tools to recover from these currently unrecoverable states, i.e. in this case "a new CLI command to help us choose the safest member to shrink to" to use with existing "force shrink" and "grow" commands (and QQ leader-health-checks https://github.com/rabbitmq/rabbitmq-server/pull/13433 which help us detect the problem).

We are able to clone rabbit-server and reproduce the same "crash" and broken "quorum-status" on the latest _main_ branch. I thought this is enough to justify that there's a bug / problem in the most latest release. Is there a specific patch for this crash at least, or a recommended recovery strategy?

We want to upgrade, but 3.12.x gives us 2 options for HA (QQ and CMQ. In case one fails we have a backup HA plan). Risk of upgrading when we're still seeing this same leaderless quorum-status issue on `main` is too high (and too scary tbh 😬). But with these cli tools and procedures, we can justify a clear recovery plan in case things fail.

kjnilsson:
> We are able to clone rabbit-server and reproduce the same "crash" and broken "quorum-status" on the latest _main_ branch. I thought this is enough to justify that there's a bug / problem in the most latest release. Is there a specific patch for this crash at least, or a recommended recovery strategy?

This issue is being worked on here https://github.com/rabbitmq/rabbitmq-server/pull/14241 - taking the highest index as a "best target for recovery" isn't necessarily going to work for this case as the old "revived" member is likely to have a higher index but contain older data.

> We want to upgrade, but 3.12.x gives us 2 options for HA (QQ and CMQ. In case one fails we have a backup HA plan). Risk of upgrading when we're still seeing this same leaderless quorum-status issue on main is too high (and too scary tbh 😬). But with these cli tools and procedures, we can justify a clear recovery plan in case things fail.

You are using a very old unsupported version. I cannot stress this fact enough. It is not a good idea to stay on 3.12.x, in fact, given the fixes and improvements available in 4.1.x I think it is reckless to stay on an old version that you know you have problems with.

Ayanda-D:
> This issue is being worked on here https://github.com/rabbitmq/rabbitmq-server/pull/14241

ok great, so the version that'll make sense for us to upgrade to, from 3.12.x is most likely to be 4.1.3. Assuming this patch is merged and does the job (needs to be tested). We're keen to upgrade, but confidence levels on using QQs entirely are not yet high on our side atm till this problem is clearly fixed (any chance of complete queue unavailability would be very detrimental). We also see need for these cli tools to offer a recovery path to try revive the queue and prevent complete outages if things fail. 

> highest index as a "best target for recovery" isn't necessarily going to work for this case as the old "revived" member is likely to have a higher index but contain older data

we'd care about what's been committed at this point. unless there's another metric we use to choose the most reliable online member? we also expect publish-confirms should nack client the moment the queue is unreachable.

kjnilsson:
> we'd care about what's been committed at this point. unless there's another metric we use to choose the most reliable online member? we also expect publish-confirms should nack client the moment the queue is unreachable.

The specific issue you keep referring to (despite not showing that you have ever encountered it) happens when a member from a different incarnation of the queue (with the same name) is revived - so it will have completely different indexes as they refer to a different incarnation. It could have a much higher index despite containing data that is much older. that is why looking a the committed index is not a reliable indicator of freshness (in this specific case).

> confidence levels on using QQs entirely are not yet high on our side atm

I cannot possibly guess why you have experienced issues with quorum queues as you haven't provided much in terms of specifics. I am confident that staying on an old, unsupported version is a very bad idea and most likely a major contributor to your experience. Support contracts are available where we can engage in a much better way and help you diagnose and remediate any issues you encounter. We cannot help with community issues encountered on 3.12, 3.13 or even 4.0 versions of RabbitMQ. Only RabbitMQ 4.1.x has community support

Ayanda-D:
Sorry to sound as if we want to persist with an old version. It's not the case 😃 We are keen to move forward. It's just questions we **_always_** have to have answers for (both for us and folks up the hierarchy) before turning on the full switch:

_"In the event of a broken QQ, what is/are the recovery procedure/s, to restore the QQ without complete queue deletion (or complete message loss)"._

We cant assume QQs will never break (is there a link to some notes on handling QQs with corrupt state perhaps?).

Thats what we are trying to answer at the moment, and to contribute. We totally agree that upgrading will clear lots of underlying issues, but we also need to design for such failure handling as well, even for 4.x. or newer (we dont expect things to always be perfect, just as 3.12.x was released such issues weren't anticipated, likewise 4.x. & future versions - such/similar issues may not be anticipated but can occur - however if we have tools and recovery procedures in place to help restore a critical QQ when things go bad - then we're confident to proceed). Currently, there's a grey area around recovery of a corrupt QQ. With old CMQ, we have procedures if things dont look "right", we are preparing the same for QQs. 

These proposed recovery procedures from our end are working, from tests ran on `main` branch and 3.12.x - _a totally broken QQ being restored back to normal operation._ We just want to have these tools available in 4.x. and later versions (in the event of "any" QQ failure that can corrupt the state in like manner) - and also happy to contribute notes on the docs for handling these disastrous problems.

Commit index may not be best indicator - but can we have this CLI command available to retrieve a member with highest index available (for use at operators discretion and purpose).

michaelklishin:
Let's wait for https://github.com/rabbitmq/rabbitmq-server/pull/14241 to be finished. I suspect that it has a chance of being backported to `4.1.x`.

I have no objections to such observability-related commands, they can be useful when you least expect them to, but if #14241 addresses the fundamental issue, the three-step workaround might simply not be necessary.

gomoripeti:
just noting that I've seen a quorum queue in a strange state on 4.1.0

```
rabbitmq-queues quorum_status <queue> 
Status of quorum queue <queue> on node rabbit@<hostname1> ...
┌───────────────────────────────────┬─────────────────┬────────────┬────────────────┬──────────────┬──────────────┬──────────────┬────────────────┬──────┬────────────
─────┐
│ Node Name                         │ Raft State      │ Membership │ Last Log Index │ Last Written │ Last Applied │ Commit Index │ Snapshot Index │ Term │ Machine Version │
├───────────────────────────────────┼─────────────────┼────────────┼────────────────┼──────────────┼──────────────┼──────────────┼────────────────┼──────┼────────────
─────┤
│ rabbit@hostname1 │ await_condition │ voter      │ 28             │ 28           │ 25           │ 25           │ -1             │ 26   │ 5               │
├───────────────────────────────────┼─────────────────┼────────────┼────────────────┼──────────────┼──────────────┼──────────────┼────────────────┼──────┼────────────
─────┤
│ rabbit@hostname2 │ leader          │ voter      │ 25             │ 25           │ 19           │ 19           │ -1             │ 26   │ 5               │
├───────────────────────────────────┼─────────────────┼────────────┼────────────────┼──────────────┼──────────────┼──────────────┼────────────────┼──────┼────────────
─────┤
│ rabbit@hostname3 │ noproc          │ unknown    │ 21             │ 21           │ 18           │ 18           │ -1             │ 24   │ 5               │
└───────────────────────────────────┴─────────────────┴────────────┴────────────────┴──────────────┴──────────────┴──────────────┴────────────────┴──────┴────────────
─────┘
```

The first member has the highest value for all indexes. But it is completely stuck. Force shrinking to that node failed.

```
erpc:call('rabbit@hostname1', rabbit_quorum_queue, force_shrink_member_to_current_member, [<<"vhost">>, <<"queue">>]).
** exception error: {exception,{badmatch,timeout},
                               [{rabbit_quorum_queue,force_shrink_member_to_current_member,
                                                     2,
                                                     [{file,"rabbit_quorum_queue.erl"},{line,2065}]}]}
     in function  erpc:call/5 (erpc.erl, line 1368)
```

However force shrinking to the second node succeeded.

What I'm saying is selecting the member with the highest index does not always work and we might need either a human's intuition or more sophisticated heuristics for automated recovery.

Ayanda-D:
@gomoripeti Thanks for trying out 👍 The `timeout` makes me wonder if the `member_with_highest_index` was/would've been responsive to `ra_server_proc:ping/2` as a sanity check before attempting the shrink + re-grow. (i.e. similar motivation for this diagnostics command: https://github.com/rabbitmq/rabbitmq-server/pull/13433 for the QQ leaders). 

> However force shrinking to the second node succeeded.

great! 🎉 

> we might need either a human's intuition or more sophisticated heuristics for automated recovery

Indeed. This cli command is only just to assist to get better visibility as a first step. I agree with your feedback (& @kjnilsson) above that this cant fully guarantee that that's the member to immediately use for recovery.

PR #14241, opened 2025-07-16
Reconcile QQ node dead during delete and redeclare

LoisSotoLopez:
## Proposed Changes

This PR implements the suggested solution for the issue described in discussion https://github.com/rabbitmq/rabbitmq-server/discussions/13131

Currently, when a QQ is deleted and re-declared while one of its nodes is dead, this dead node won't be able to reconcile with the new queue. 

In this PR we add the list of ra UIds for the cluster to each node queue record, so that when a Rabbit node recovers a queue it will be able to detect the situation described above and properly reconcile .

## Types of Changes

- [ ] Build system and/or CI

## Checklist

We're here to help!

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -275,9 +275,12 @@ start_cluster(Q) ->
     {LeaderNode, FollowerNodes} =
         rabbit_queue_location:select_leader_and_followers(Q, QuorumSize),
     LeaderId = {RaName, LeaderNode},
+    UIDs = maps:from_list([{Node, ra:new_uid(ra_lib:to_binary(RaName))}
+                           || Node <- [LeaderNode | FollowerNodes]]),
     NewQ0 = amqqueue:set_pid(Q, LeaderId),
     NewQ1 = amqqueue:set_type_state(NewQ0,
-                                    #{nodes => [LeaderNode | FollowerNodes]}),
+                                    #{nodes => [LeaderNode | FollowerNodes],
+                                      uids => UIDs}),
 
     Versions = [V || {ok, V} <- erpc:multicall(FollowerNodes,
                                                rabbit_fifo, version, [],
@@ -791,6 +794,24 @@ recover(_Vhost, Queues) ->
          ServerId = {Name, node()},
          QName = amqqueue:get_name(Q0),
          MutConf = make_mutable_config(Q0),
+         RaUId = ra_directory:uid_of(?RA_SYSTEM, Name),
+         QTypeState0 = amqqueue:get_type_state(Q0),
+         RaUIds = maps:get(uids, QTypeState0, undefined),
+         QTypeState = case RaUIds of
+             undefined ->
+                 %% Queue is not aware of node to uid mapping, do nothing
+                 QTypeState0;
+             #{node() := RaUId} ->
+                 %% Queue is aware and uid for current node is correct, do nothing
+                 QTypeState0;
+             _ ->
+                 %% Queue is aware but either current node has no UId or it
+                 %% does not match the one returned by ra_directory, regen uid
+                 maybe_delete_data_dir(RaUId),
+                 NewRaUId = ra:new_uid(ra_lib:to_binary(Name)),
+                 QTypeState0#{uids := RaUIds#{node() => NewRaUId}}
+         end,
+         Q = amqqueue:set_type_state(Q0, QTypeState),
          Res = case ra:restart_server(?RA_SYSTEM, ServerId, MutConf) of
                    ok ->
                        % queue was restarted, good
@@ -803,7 +824,7 @@ recover(_Vhost, Queues) ->
                                           [rabbit_misc:rs(QName), Err1]),
                        % queue was never started on this node
                        % so needs to be started from scratch.
-                       case start_server(make_ra_conf(Q0, ServerId)) of
+                       case start_server(make_ra_conf(Q, ServerId)) of
                            ok -> ok;
                            Err2 ->
                                rabbit_log:warning("recover: quorum queue ~w could not"
@@ -825,8 +846,7 @@ recover(_Vhost, Queues) ->
          %% present in the rabbit_queue table and not just in
          %% rabbit_durable_queue
          %% So many code paths are dependent on this.
-         ok = rabbit_db_queue:set_dirty(Q0),
-         Q = Q0,
+         ok = rabbit_db_queue:set_dirty(Q),
          case Res of
              ok ->
                  {[Q | R0], F0};
@@ -1207,12 +1227,17 @@ cleanup_data_dir() ->
 maybe_delete_data_dir(UId) ->
     _ = ra_directory:unregister_name(?RA_SYSTEM, UId),
     Dir = ra_env:server_data_dir(?RA_SYSTEM, UId),
-    {ok, Config} = ra_log:read_config(Dir),
-    case maps:get(machine, Config) of
-        {module, rabbit_fifo, _} ->
-            ra_lib:recursive_delete(Dir);
-        _ ->
-            ok
+    case filelib:is_dir(Dir) of
+        false ->
+            ok;
+        true ->
+            {ok, Config} = ra_log:read_config(Dir),
+            case maps:get(machine, Config) of
+                {module, rabbit_fifo, _} ->
+                    ra_lib:recursive_delete(Dir);
+                _ ->
+                    ok
+            end
     end.
 
 policy_changed(Q) ->
@@ -1378,16 +1403,30 @@ add_member(Q, Node, Membership) ->
     do_add_member(Q, Node, Membership, ?MEMBER_CHANGE_TIMEOUT).
 
 
-do_add_member(Q, Node, Membership, Timeout)
-  when ?is_amqqueue(Q) andalso
-       ?amqqueue_is_quorum(Q) andalso
+do_add_member(Q0, Node, Membership, Timeout)
+  when ?is_amqqueue(Q0) andalso
+       ?amqqueue_is_quorum(Q0) andalso
        is_atom(Node) ->
-    {RaName, _} = amqqueue:get_pid(Q),
-    QName = amqqueue:get_name(Q),
+    {RaName, _} = amqqueue:get_pid(Q0),
+    QName = amqqueue:get_name(Q0),
     %% TODO parallel calls might crash this, or add a duplicate in quorum_nodes
     ServerId = {RaName, Node},
-    Members = members(Q),
-
+    Members = members(Q0),
+    QTypeState0 = amqqueue:get_type_state(Q0),
+    RaUIds = maps:get(uids, QTypeState0, undefined),
+    QTypeState = case RaUIds of
+        undefined ->
+            %% Queue is not aware of node to uid mapping, do nothing
+            QTypeState0;
+        #{Node := _} ->
+            %% Queue is aware and uid for targeted node exists, do nothing
+            QTypeState0;
+        _ ->
+            %% Queue is aware but current node has no UId, regen uid
+            NewRaUId = ra:new_uid(ra_lib:to_binary(RaName)),
+            QTypeState0#{uids := RaUIds#{Node => NewRaUId}}
+    end,
+    Q = amqqueue:set_type_state(Q0, QTypeState),
     MachineVersion = erpc_call(Node, rabbit_fifo, version, [], infinity),
     Conf = make_ra_conf(Q, ServerId, Membership, MachineVersion),
     case ra:start_server(?RA_SYSTEM, Conf) of
@@ -1477,7 +1516,11 @@ delete_member(Q, Node) when ?amqqueue_is_quorum(Q) ->
                     Fun = fun(Q1) ->
                                   update_type_state(
                                     Q1,
-                                    fun(#{nodes := Nodes} = Ts) ->
+                                    fun(#{nodes := Nodes,
+                                          uids := UIds} = Ts) ->
+                                            Ts#{nodes => lists:delete(Node, Nodes),
+                                                uids => maps:remove(Node, UIds)};
+                                       (#{nodes := Nodes} = Ts) ->
                                             Ts#{nodes => lists:delete(Node, Nodes)}
                                     end)
                           end,
@@ -1986,7 +2029,15 @@ make_ra_conf(Q, ServerId, TickTimeout,
     QName = amqqueue:get_name(Q),
     RaMachine = ra_machine(Q),
     [{ClusterName, _} | _] = Members = members(Q),
-    UId = ra:new_uid(ra_lib:to_binary(ClusterName)),
+    {_, Node} = ServerId,
+    UId = case amqqueue:get_type_state(Q) of
+        #{uids := #{Node := Id}} ->
+            Id;
+        _ ->
+            %% Queue was declared on an older version of RabbitMQ
+            %% and does not have the node to uid mappings
+            ra:new_uid(ra_lib:to_binary(ClusterName))
+    end,
     FName = rabbit_misc:rs(QName),
     Formatter = {?MODULE, format_ra_event, [QName]},
     LogCfg = #{uid => UId,
--- a/deps/rabbit/test/quorum_queue_SUITE.erl
+++ b/deps/rabbit/test/quorum_queue_SUITE.erl
@@ -105,7 +105,9 @@ groups() ->
                                             force_checkpoint,
                                             policy_repair,
                                             gh_12635,
-                                            replica_states
+                                            replica_states,
+                                            restart_after_queue_reincarnation,
+                                            no_messages_after_queue_reincarnation
                                            ]
                        ++ all_tests()},
                       {cluster_size_5, [], [start_queue,
@@ -2802,15 +2804,21 @@ add_member_wrong_type(Config) ->
                           [<<"/">>, SQ, Server, voter, 5000])).
 
 add_member_already_a_member(Config) ->
-    [Server | _] = rabbit_ct_broker_helpers:get_node_configs(Config, nodename),
+    [Server, Server2 | _] = rabbit_ct_broker_helpers:get_node_configs(Config, nodename),
     Ch = rabbit_ct_client_helpers:open_channel(Config, Server),
     QQ = ?config(queue_name, Config),
     ?assertEqual({'queue.declare_ok', QQ, 0, 0},
                  declare(Ch, QQ, [{<<"x-queue-type">>, longstr, <<"quorum">>}])),
+    R1 = rpc:call(Server, rabbit_amqqueue, lookup, [{resource, <<"/">>, queue, QQ}]),
     %% idempotent by design
     ?assertEqual(ok,
                  rpc:call(Server, rabbit_quorum_queue, add_member,
-                          [<<"/">>, QQ, Server, voter, 5000])).
+                          [<<"/">>, QQ, Server, voter, 5000])),
+    ?assertEqual(R1, rpc:call(Server, rabbit_amqqueue, lookup, [{resource, <<"/">>, queue, QQ}])),
+    ?assertEqual(ok,
+                 rpc:call(Server, rabbit_quorum_queue, add_member,
+                          [<<"/">>, QQ, Server2, voter, 5000])),
+    ?assertEqual(R1, rpc:call(Server, rabbit_amqqueue, lookup, [{resource, <<"/">>, queue, QQ}])).
 
 add_member_not_found(Config) ->
     [Server | _] = rabbit_ct_broker_helpers:get_node_configs(Config, nodename),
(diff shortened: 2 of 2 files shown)


kjnilsson:
thanks @LoisSotoLopez - this looks like it will do what we discussed a while back. I feel unsure about adding another key to the queue type state for this, mainly becuause we'd have to keep `uids` and `nodes` in sync. It would be nicer if `nodes` turned from a list into a map although even this is a bit controversial and could become a source of bugs. Let me consider it for a day or two.

kjnilsson (changes requested):
I don't like having two keys with similar information (nodes) that will need to be kept in sync. I think we need to move the current `nodes` list value to the the `#{node() => uid()}` map format and handle the two formats in the relevant places, mostly in `rabbit_queue_type`. we do need a list of the member nodes in a few places but we could add a convenience functio: `rabbit_queue_type:nodes/1` that takes a queue record and returns a list of member nodes. internally it could just call `rabbit_queue_type:info(Q, [members])` and extract the result from that then update all places where we explicity use get_type_state to extract the member nodes.

In addition I think we need to put the use of `nodes` as a map behind a feature flag to avoid new queue records with `nodes` map values being created in a mixed versions cluster.

LoisSotoLopez:
@kjnilsson Thanks for the suggestions. Just wanted to let you know we are working on this. Had some incident we had to take care of this week but I'll be pushing this PR forward next week.

michaelklishin:
@LoisSotoLopez we have to ask your employer to sign the [Broadcom CLA](https://github.com/rabbitmq/cla/) before we can accept this contribution (or its future finished version).

It is about one page long, nothing particularly unusual or onerous.

LoisSotoLopez:
That commit below is just for showing current progress. Have been struggling to understand why a few of the remaining tests fail.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -791,6 +810,23 @@ recover(_Vhost, Queues) ->
          ServerId = {Name, node()},
          QName = amqqueue:get_name(Q0),
          MutConf = make_mutable_config(Q0),
+         RaUId = ra_directory:uid_of(?RA_SYSTEM, Name),
```
LoisSotoLopez, line 880: Figured out what I was doing wrong but I'm not sure what's the best approach to fix it. Let me recap and raise some questions.

So when a QQ is declared I am setting the `#{node() => ra_uid()}` map in the Queue Type State. I'll refer to those values in the map as `NodeRaUid`s, as opposed to the UId associated to each QQ as a ra cluster ( the one generated by calling `rabbit_quorum_queue:make_ra_conf`), which I'll refer to as `ClusterRaUid`. 

When a queue is deleted and re-declared the Queue Type state gets re-generated (no doubt about that because it's a new queue). Therefore between queue reincarnations the Queue Type state will change. However, the `ClusterRaUid` will not change for those nodes that were dead during the delete+redeclare. 

I was checking, on member recovery, whether the ClusterRaUid, as retrieved with `ra_directory:uid_of`, was associated to the current node in the map of `NodeRaUid`s. That's not right. 

My current approach is storing the `NodeRaUid`s map in disk, as it is done for `ClusterRaId`, and on recover compare that in-disk map with the one in the `QueueTypeState`.

**My questions for you guys are**: does that last paragraph sound right? and if it does, what API should I use to store that in-disk copy of the `NodeRaUid`s map? Is there any `ra_` module that I could use?

michaelklishin: If we already store `ClusterRaId` that way, I guess we can persist more metadata that the Ra members won't otherwise have/preserve.

michaelklishin:
@LoisSotoLopez sorry, any updates or feedback on the [new CLA](https://github.com/rabbitmq/cla)? We cannot accept any contributions without a signed CLA at the moment, and this change won't qualify for a "trivial" one, even if those get exempt in the future.

We are trying to make it digital one way or another but there's a risk that the process will stay what it currently is (just a document to sign and email).

LoisSotoLopez:
> @LoisSotoLopez sorry, any updates or feedback on the [new CLA](https://github.com/rabbitmq/cla)? We cannot accept any contributions without a signed CLA at the moment, and this change won't qualify for a "trivial" one, even if those get exempt in the future.
> 
> We are trying to make it digital one way or another but there's a risk that the process will stay what it currently is (just a document to sign and email).

Yes, sorry about not having this already addressed. We are on summer vacations right now so the people who will be signing it for the whole company won't be able to do it until next week. I would do it myself but can't due to a intellectual property clause on  my contract that wasn't there the last time I signed the CLA.

Will try to get it signed asap.

deps/rabbit/src/rabbit_quorum_queue.erl
LoisSotoLopez: The `ClusterRaUid` gets persisted through `ra_directory.erl` which seems designed to strictly storing Ra cluster's specific pieces of related information. Maybe we just need a `qq_nodes_uids` dets for the sole purpose of storing those `NodeRaUid`s. Or using Rabbit's metadata storage.

Edited: forget about this, found the error that led to this confusion.

LoisSotoLopez:
Hi! The CLA should be signed now. This should be ready to be reviewed again.

michaelklishin:
@LoisSotoLopez I'm afraid I do not see any inbound emails with the signed document. Perhaps it was not sent to `teamrabbitmq <@> gmail`, as the [CLA repo suggests](https://github.com/rabbitmq/cla) but rather straight to `opensource <@> broadcom`.

LoisSotoLopez:
> @LoisSotoLopez I'm afraid I do not see any inbound emails with the signed document. Perhaps it was not sent to `teamrabbitmq <@> gmail`, as the [CLA repo suggests](https://github.com/rabbitmq/cla) but rather straight to `opensource <@> broadcom`.

Should be there now. 🫡

michaelklishin:
@LoisSotoLopez thank you, we should be all set in terms of the CLA. 

I'll defer merging to @kjnilsson. That can take another week or two.

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1234,6 +1235,12 @@ list() ->
 count() ->
     rabbit_db_queue:count().
 
+-spec get_nodes(amqqueue:amqqueue_v2()) -> [node(),...].
+
```
kjnilsson: redundant blank line

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -276,9 +276,17 @@ start_cluster(Q) ->
     {LeaderNode, FollowerNodes} =
         rabbit_queue_location:select_leader_and_followers(Q, QuorumSize),
     LeaderId = {RaName, LeaderNode},
+    UIDs = maps:from_list([{Node, ra:new_uid(ra_lib:to_binary(RaName))}
+                           || Node <- [LeaderNode | FollowerNodes]]),
     NewQ0 = amqqueue:set_pid(Q, LeaderId),
-    NewQ1 = amqqueue:set_type_state(NewQ0,
-                                    #{nodes => [LeaderNode | FollowerNodes]}),
+    NewQ1 = case rabbit_feature_flags:is_enabled(track_qq_members_uids) of
```
kjnilsson, line 276: `case` indentation isn't correct.

deps/rabbit/src/rabbit_quorum_queue.erl
kjnilsson: There is no such thing as a ClusterRaUId? the UId that is generated for each member is it's local identifier which is what we store in the nodes map.

LoisSotoLopez:
Just rebased and squashed this PR. Let me know if there's anything new I can do to move it forward.

michaelklishin:
@kjnilsson

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -790,6 +809,23 @@ recover(_Vhost, Queues) ->
          ServerId = {Name, node()},
          QName = amqqueue:get_name(Q0),
          MutConf = make_mutable_config(Q0),
+         RaUId = ra_directory:uid_of(?RA_SYSTEM, Name),
+         #{nodes := Nodes} = QTypeState0 = amqqueue:get_type_state(Q0),
+         QTypeState = case Nodes of
```
kjnilsson: `case` formatting is off

michaelklishin:
@LoisSotoLopez given that this changes how member nodes are represented in the queue state, it is too late for us to get it into `4.2.0` and the earliest this PR can be considered is for `4.3.0`, the work on which will begin in earnest in November.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1378,16 +1418,29 @@ add_member(Q, Node, Membership) ->
...
+    {RaName, _} = amqqueue:get_pid(Q0),
+    QName = amqqueue:get_name(Q0),
     %% TODO parallel calls might crash this, or add a duplicate in quorum_nodes
     ServerId = {RaName, Node},
-    Members = members(Q),
-
+    Members = members(Q0),
+    QTypeState0 = #{nodes := Nodes} = amqqueue:get_type_state(Q0),
+    NewRaUId = ra:new_uid(ra_lib:to_binary(RaName)),
+    QTypeState = case Nodes of
```
kjnilsson, line 1485: `case` formatting is off

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -790,6 +809,23 @@ recover(_Vhost, Queues) ->
...
+         #{nodes := Nodes} = QTypeState0 = amqqueue:get_type_state(Q0),
+         QTypeState = case Nodes of
+             List when is_list(List) ->
+                 %% Queue is not aware of node to uid mapping, do nothing
+                 QTypeState0;
+             #{node() := RaUId} ->
+                 %% Queue is aware and uid for current node is correct, do nothing
+                 QTypeState0;
+             _ ->
+                 %% Queue is aware but either current node has no UId or it
```
kjnilsson: We're going to need some nice info logging at this point.

kjnilsson: I also think here to avoid having to update the queue type state in the db you can just use the Uid in the queue type state for the node if present (I think it will be present for any queues declared after ff enabled). That would save you creating a new UId for this member if it is already configured in the amqqueue record.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -824,8 +860,7 @@ recover(_Vhost, Queues) ->
          %% present in the rabbit_queue table and not just in
          %% rabbit_durable_queue
          %% So many code paths are dependent on this.
-         ok = rabbit_db_queue:set_dirty(Q0),
-         Q = Q0,
+         ok = rabbit_db_queue:set_dirty(Q),
```
kjnilsson, line 935: this isn't enough to persist the new queue type state in the database. if we re-use the already generated uid for this member we should be good and not even need this. 

This line of code can be removed for RabbitMQ 4.3

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1999,7 +2058,15 @@ make_ra_conf(Q, ServerId, TickTimeout,
     #resource{name = QNameBin} = QName,
     RaMachine = ra_machine(Q),
     [{ClusterName, _} | _] = Members = members(Q),
-    UId = ra:new_uid(ra_lib:to_binary(ClusterName)),
+    {_, Node} = ServerId,
+    UId = case amqqueue:get_type_state(Q) of
```
kjnilsson, line 2169: `case` formatting is off

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1999,7 +2058,15 @@ make_ra_conf(Q, ServerId, TickTimeout,
...
+        #{nodes := #{Node := Id}} ->
+            Id;
+        _ ->
+            %% Queue was declared on an older version of RabbitMQ
+            %% and does not have the node to uid mappings
```
kjnilsson: should be "_or_ doesn't have the node to uid mappings"

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1226,6 +1227,12 @@ list() ->
...
+get_nodes(Q) ->
```
kjnilsson: I think this function should be in `rabbit_queue_type`

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1403,8 +1456,12 @@ do_add_member(Q, Node, Membership, Timeout)
...
-                                         Q1, fun(#{nodes := Nodes} = Ts) ->
-                                                     Ts#{nodes => lists:usort([Node | Nodes])}
+                                         Q1, fun(#{nodes := NodesList} = Ts) when is_list(NodesList) ->
+                                                     Ts#{nodes => lists:usort([Node | NodesList])};
+                                                (#{nodes := #{Node := _}} = Ts) ->
+                                                     Ts;
+                                                (#{nodes := NodesMap} = Ts) when is_map(NodesMap) ->
+                                                     Ts#{nodes => maps:put(Node, NewRaUId, NodesMap)}
                                              end),
                                   amqqueue:set_pid(Q2, Leader)
```
kjnilsson: this line is surely not needed.

gomoripeti:
It is responsible to not rush something released that changes the queue record.

Because 4.3 is 6 months away I wonder if it would have any benefit to still include that part of the PR in 4.2 which moves `get_name` from `amqqueue` to `rabbit_amqqueue` module. That change is just a refactoring but touches multiple modules while the problematic part is restricted to `rabbit_quorum_queue`. I'm suggesting this because maybe it avoids some conflicts and makes it easier to backport other changes to 4.2 when the whole of this PR is merged into main but not into 4.2.x. Or if the whole of this PR is not merged for a long time it would make it easier to rebase it to a future main.
Just an idea, what do you think?

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -726,7 +734,18 @@ repair_amqqueue_nodes(Q0) ->
             %% update amqqueue record
             Fun = fun (Q) ->
                           TS0 = amqqueue:get_type_state(Q),
-                          TS = TS0#{nodes => RaNodes},
+                          TS = case rabbit_feature_flags:is_enabled(track_qq_members_uids)
+                                    andalso has_uuid_tracking(TS0)
+                               of
+                                   false ->
```
kjnilsson: I think in the case where the feature flag is enabled and there is no existing uuid trackign we should perform a state update to ensure there is tracking.

deps/rabbit/src/rabbit_quorum_queue.erl
kjnilsson: not part of the PR though, just an observation

kjnilsson (changes requested):
A few more changes, this feature is a bit trickier than I had initially expected. I am happy take over if you feel like we're going back and forth too much. Just let me know.

michaelklishin:
Test failures are not flakes. They key part is:

```
{badmatch,[]}

[{rabbit_queue_type,get_nodes,1,[{file,\"rabbit_queue_type.erl\"}, {line,428}]},

{rabbit_amqp_management,queue_topology,1, [{file,\"rabbit_amqp_management.erl\"},{line,474}]},\n 

{rabbit_amqp_management,encode_queue,3, [{file,\"rabbit_amqp_management.erl\"},{line,438}]}
```

LoisSotoLopez:
There's some test that passes locally but fails in CI. Expect a few tiny commits while I figure out what's going on.

LoisSotoLopez:
Just one tiny commit was needed :) I think it's now ready for review again.

deps/rabbit/src/rabbit_core_ff.erl
```diff
@@ -219,3 +219,10 @@
       depends_on    => ['rabbitmq_4.1.0'],
       callbacks     => #{enable => {rabbit_khepri, enable_feature_flag}}
      }}).
+
+-rabbit_feature_flag(
+   {'track_qq_members_uids',
+    #{desc         => "Track queue members UIDs in the metadata store",
```
kjnilsson: it might be a lot better not to add superfluous whitespace before `=>` at all - I _think_ this formatting style expects the `=>` symbols to line up but here they don't even do that. And that is the primary problem with  the superfluous whitespace formatting style: they tend to drift and just make stuff harder to read in the end.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -726,7 +734,16 @@ repair_amqqueue_nodes(Q0) ->
             %% update amqqueue record
             Fun = fun (Q) ->
                           TS0 = amqqueue:get_type_state(Q),
-                          TS = TS0#{nodes => RaNodes},
+                          TS = case rabbit_feature_flags:is_enabled(track_qq_members_uids) of
+                                   false ->
+                                       TS0#{nodes => RaNodes};
+                                   true ->
+                                       RaUids = maps:from_list([{N, erpc:call(N, ra_directory, uid_of,
```
kjnilsson: `ra_directory:uid_of/2` can return `undefined` - I think we need to guard against that somehow

deps/rabbit/src/rabbit_core_ff.erl
kjnilsson: Rant aside you just need to line up the `=>`

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -785,11 +802,28 @@ maybe_apply_policies(Q, #{config := CurrentConfig}) ->
...
-      fun (Q0, {R0, F0}) ->
-         {Name, _} = amqqueue:get_pid(Q0),
+      fun (Q, {R0, F0}) ->
+         {Name, _} = amqqueue:get_pid(Q),
          ServerId = {Name, node()},
-         QName = amqqueue:get_name(Q0),
-         MutConf = make_mutable_config(Q0),
+         QName = amqqueue:get_name(Q),
+         MutConf = make_mutable_config(Q),
+         RaUId = ra_directory:uid_of(?RA_SYSTEM, Name),
```
LoisSotoLopez, line 880: > ra_directory:uid_of/2 can return undefined - I think we need to guard against that somehow

Also here then.

deps/rabbit/src/rabbit_quorum_queue.erl
LoisSotoLopez: Also, in `recover/2` then.

I feel like an `undefined` could appear only if a node appears in the `ra_leaderboard` (so returned by `lookup_members/1`) but for some reason is still not registered in the local `ra_directory`. 

I don't think that would be something that should happen on a normal scenario without some major error happening in some earlier execution phase. But agree, we could maybe log with a warning if `undefined`s appear there.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -725,7 +733,28 @@ repair_amqqueue_nodes(Q0) ->
             %% update amqqueue record
             Fun = fun (Q) ->
                           TS0 = amqqueue:get_type_state(Q),
-                          TS = TS0#{nodes => RaNodes},
+                          TS = case rabbit_feature_flags:is_enabled(track_qq_members_uids) of
+                                   false ->
+                                       TS0#{nodes => RaNodes};
+                                   true ->
+                                       RaUidsList = [begin
+                                                         Uid = erpc:call(N, ra_directory, uid_of,
```
kjnilsson: It looks like this is done irrespectively of whether the nodes currently include uids or not, I think this part should only be done if the queue record currently has the old list format.

gomoripeti: you are right that in the happy path when all queues have the new list format, there are unnecessary rpc calls.
But this code is executed when the members changed (`lists:sort(RaNodes) =:= lists:sort(Nodes)` is false) so if there are some new nodes, then some rpc-ing is necessary. So maybe need to split this to two sub-cases: 1. nodes are in old list format 2. new list format but new members added

gomoripeti:
Hi Karl, I just wanted to double check, what should happen with the nodes list of queues that were created before the feature flag is enabled (https://github.com/rabbitmq/rabbitmq-server/pull/14241#discussion_r2390766440)

1. nothing. This makes it very easy to reason about "All queues created after the feature flag was enabled have uid tracking and none of the queues created before the feature flag was enabled has uid tracking". This is not a problem from the original issue's point of view which requires a QQ to be deleted and redeclared, so no problem with long lived queues. The drawback is that there will be some queues which might have the old nodes list format forever.
2. Update nodes list to new format when repair finds a mismatch of members in metadata store (current implementation). Drawback is that this will happen to some queues, but we never know if this will ever happen to all queues. So we still might have some queues with the old nodes list format forever.
3. Update nodes list to new format in tick/repair even if no mismatch in nodes list. This has the advantage that it will happen to all queues within a tick interval after the feature flag is enabled. Drawback is that this might flood khepri and create a lot of rpc calls
4. Convert all QQs as a feature flag migration step. This also has the advantage that it will happen to all queues, when the feature flag is enabled. It can also use one rpc per node calling `ra_directory:list_registered/1`. Drawback is that it can still create a lot of updates in khepri. (But it cannot be more updates than the mnesia->khepri migration itself) Apart from the khepri migration Core Team avoided such "table conversions" recently.

kjnilsson:
I think 3 but 4 could also work although it might be prone to race conditions where the local check and the migration task disagree.

gomoripeti:
while Lois is away I pushed his changes in the last commit. Also rebased to latest main.

It implements point 3. variant of the nodes list to map conversion. Also added some unit-test-like test cases for repair_amqqueue_nodes.

If this will make it into 4.3 maybe the standalone feature flag can be merged into the generic `rabbitmq_4.3.0` feature flag.

michaelklishin:
@gomoripeti can you please rebase this branch? Thank you.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -785,11 +871,35 @@ maybe_apply_policies(Q, #{config := CurrentConfig}) ->
...
+         #{nodes := Nodes} = amqqueue:get_type_state(Q),
+         case Nodes of
+             List when is_list(List) ->
+                 %% Queue is not aware of node to uid mapping, do nothing
+                 ok;
+             #{node() := RaUId} ->
+                 %% Queue is aware and uid for current node is correct, do
+                 %% nothing
+                 ok;
+             #{node() := _NewRaUId} ->
```
kjnilsson, line 892: would we end up here if RaUId for some reason was `undefined`?

gomoripeti: yes. and while `maybe_delete_data_dir` is a noop in this case, the logging and comment is misleading. Would be good to add another case clause or rearrange the case to be more explicit about the `undefined` RaUid case

kjnilsson:
@LoisSotoLopez there are 5 failing tests according to ci - please can you take a look?

michaelklishin:
Those can be flakes we've addressed earlier, so I'd make sure that the PR is rebased first.

But just to compare, I've re-triggered those QQ tests.

LoisSotoLopez:
Done, couldn't identify it from the CI tests, but it was the `grow_queue` test in `quorum_queue_SUITE` the one failing. Maybe `message_ttl_policy` too, but that was flaky and ended up succeeding (not sure if it was flaky or it ended up succeeding because of my changes to fix the other test case).

LoisSotoLopez:
Mmmm, tests pass locally even when running with the same command as ci `make -C deps/rabbit ct-quorum_queue RABBITMQ_METADATA_STORE=khepri NON_DETERMINISTIC=1`. Not sure how to identify which one is failing from the CI, but maybe you have some idea of what the cause of the inconsistency might be.

michaelklishin:
@LoisSotoLopez not sure if you can see the CT logs archive in the Actions build but the `gmake ct-quorum_queue` failures are:

 * `cluster_size_3:repair_metadata_nodes_list_to_map`
 * `cluster_size_3:repair_metadata_nodes_added_member`
 * `cluster_size_3:repair_metadata_nodes_removed_member`
 * `cluster_size_3:repair_metadata_nodes_added_removed_member`

Here is one example:

```
=== Location: [{erpc,call,1380},
              {quorum_queue_SUITE,repair_metadata_nodes,1628},
              {test_server,ts_tc,1796},
              {test_server,run_test_case_eval1,1305},
              {test_server,run_test_case_eval,1237}]

=== Reason: {exception,
                 {badmap,
                     ['rmq-ct-cluster_size_3-1-21162@localhost',
                      'rmq-ct-cluster_size_3-2-21216@localhost',
                      'rmq-ct-cluster_size_3-3-21270@localhost']},
                 [{maps,keys,
                      [['rmq-ct-cluster_size_3-1-21162@localhost',
                        'rmq-ct-cluster_size_3-2-21216@localhost',
                        'rmq-ct-cluster_size_3-3-21270@localhost']],
                      [{error_info,#{module => erl_stdlib_errors}}]},
                  {quorum_queue_SUITE,
                      '-repair_metadata_nodes_list_to_map/1-fun-0-',1,
                      [{file,"test/quorum_queue_SUITE.erl"},{line,1585}]},
                  {rabbit_db_queue,update_in_khepri,3,
                      [{file,"src/rabbit_db_queue.erl"},{line,677}]},
                  {rabbit_amqqueue,update,2,[]}]}
```

Perhaps you want to call `rabbit_data_coercion:to_map/1` on some of the values in the function in question.

LoisSotoLopez:
Last test failures seem unrelated:
- `Test plugins (rabbitmq_shovel)...` : seen errors like this `Error: Reason: {thrown,{timeout_waiting_for_deliver1,[...]}}` on CI logs and  the following one in rmq server logs
```
2026-01-09 09:13:43.325473+00:00 [error] <0.746.0> Channel error on connection <0.647.0> (127.0.0.1:48722 -> 127.0.0.1:21000, vhost: '/', user: 'guest'), channel 1:
2026-01-09 09:13:43.325473+00:00 [error] <0.746.0> operation basic.publish caused a channel exception not_found: no exchange 'test_exchange' in vhost '/'
```
- `Test plugins(rabbitmq_trust_store)...` : this error on the ci logs
```
fatal: unable to access 'https://github.com/rabbitmq/inet_tcp_proxy/': Failed to connect to github.com port 443 after 133427 ms: Couldn't connect to server
```

> not sure if you can see the CT logs archive in the Actions build but ...

@michaelklishin yep, I can. Missed the zip files in the Actions Summary page for the PR.

michaelklishin:
Yeah, we again have accumulated a few flakes in `main`. I have restarted those failed jobs.

The Selenium one requires secret access that external contributors do not have.

LoisSotoLopez:
Regarding the latest tests failing:

### For the `ct-quorum_queue` error:

which happened on `force_shrink_member_to_current_member` testcase of `quorum_queue_SUITE`, it seems like when re-growing the queue after shrinking, and successfully detecting that the number of replicas is reduced to 1, the queue was not able to grow because `non-voter` nodes were detected.

I would expect non-voter nodes to be present, particularly the node where we shrinked if it has enough time to become leader.

Several points in this regard:
- The node where we just shrinked is expected to be leader by the time we grow the queue, since we wait for messages to be ready which feels to me should be time enough to be come leader.
```erl
            rabbit_ct_broker_helpers:rpc(Config, 0, rabbit_quorum_queue,
                                         force_shrink_member_to_current_member, [<<"/">>, QQ]),

            wait_for_messages_ready([Server0], RaName, 3),
            queue_utils:assert_number_of_replicas(
              Config, Server0, <<"/">>, QQ, 1),

            %% grow queues back to all nodes
            [rpc:call(Server0, rabbit_quorum_queue, grow, [S, <<"/">>, <<".*">>, all]) || S <- [Server1, Server2]],
```
- The logic behind getting nodes changed slightly but we are just taking the list of nodes from the queue state, so I don't think that has affected the test results.

@michaelklishin do you think this has been just a race condition? I ran the tests locally and passed all green so...

_Edit: k, maybe those two were flaky ones_

michaelklishin:
@LoisSotoLopez yeah, there are flakes in `main` again. We fix most of them every so often, then merge a lot of stuff with a lot of tests and some new flakes appear. It's a "grass cutting" exercise :)

We will get to review this PR likely in the next couple of weeks when Ra v3 lands. That'd give us two more months before the anticipated `4.3.0` release date.

LoisSotoLopez:
After this last rebase tests seem to pass without more intervention. Is there anything else I can do?

michaelklishin:
@LoisSotoLopez thanks. The only thing that's left is a decision to merge on our team, I am trying to push this PR forward.

michaelklishin:
@LoisSotoLopez the only remaining step is to test rolling upgrades (mixed version clusters) with this change.

michaelklishin:
Merged with some changes in https://github.com/rabbitmq/rabbitmq-server/pull/15615 for `4.3.0` (depends on the `rabbitmq_4.3.0` feature flag).

PR #14256, opened 2025-07-21
By @dcorbacho and me: introduce [cluster-]local shovels, adopt message containers for shovels

michaelklishin:
This introduces several changes for shovels for `4.2.0`:

 * Cross-protocol message translation is now done using the `mc` module
 * A new "local" shovel type ("protocol") that uses the internal API (namely `rabbit_queue_type`) and the AMQP 1.0-inspired consumer and publisher side implementations that avoid using an AMQP 1.0 or AMQP 0-9-1 Erlang client
 * AMQP 1.0 shovel no longer mutates messages where it should not do so

## Cluster-Local Shovel Type

This PR introduces a new shovel "protocol": `local`. Instead of using an AMQP 1.0 or AMQP 0-9-1 Erlang client for consuming and re-publishing the messages, this shovel implementation
uses various internal API modules, namely `rabbit_queue_type`, and can skip network serialization and cross-protocol message conversion.

For this reason, this shovel protocol **can only be used to with a source or destination in the same cluster**. However, the other part of shovel can still use AMQP 1.0 or AMQP 0-9-1.

A shovel can use the `local` protocol on both sides within the same cluster.

## Message Containers for Message Translation

This is self-explanatory: adopt the `mc` module.

## AMQP 1.0 Message Immutability: a Potentially Breaking Change

AMQP 1.0 shovels modified message application property, even though it arguably should not.

Instead, we modify message annotations. **This is a potentially breaking change**: a number of metadata fields have been renamed.

 * `shovelled-by` => `x-opt-shovelled-by`
 * `shovel-type` => `x-opt-shovel-type`
 * `shovel-name` => `x-opt-shovel-name`


deps/rabbitmq_shovel/src/rabbit_local_shovel.erl
```diff
@@ -0,0 +1,752 @@
...
+                            acting_user => ?SHOVEL_USER},
+                   case remaining(Q, State) of
+                       0 ->
+                          {0, {error, autodelete}};
+                       Remaining ->
+                           {Remaining, rabbit_queue_type:consume(Q, Spec, QState0)}
+                   end
+           end) of
+        {Remaining, {ok, QState1}} ->
+            {ok, QState, Actions} = rabbit_queue_type:credit(QName, CTag, ?INITIAL_DELIVERY_COUNT, MaxLinkCredit, false, QState1),
```
ansd, line 180: Granting a fix value of `MaxLinkCredit=170` directly after registering the consumer doesn't make sense (even if configurable via the environment).
The number of link credit to be granted should be determined by the shovel destination/receiver, specifically by:
* the link credit granted , if the the shovel receiver uses AMQP 1.0
* the AMQP 0.9.1 consumer prefetch, if the shovel receiver uses AMQP 0.9.1
* the MQTT 5.0 `Receive Maximum`, if the shovel receiver uses MQTT 5.0 with QoS 1

As an example, imagine that the AMQP 1.0 shovel receiver grants 5 link credits. In this case this line here shouldn't grant 170 link credits (since this would mean 165 messages will be held in memory unnecessarily).

dcorbacho: There is no info about the link credit granted amount in the AMQP 1.0 shovel, it just receives an `insuficiente_credit` error from the client and waits for the next `credited` notification. 

It could use the prefetch count for both AMQP 0.9.1 and 1.0, as that is how both shovels are initialised.

MQTT 5.0 is not supported in shovel.

ansd: > There is no info about the link credit granted amount in the AMQP 1.0 shovel

This info is not yet available at the time the link is attached (`attach` frame). This info becomes available with the first `flow` frame.

> MQTT 5.0 is not supported in shovel

I'm aware of that. I suggested that because we want a future shovel version to support MQTT 5.0 and the same idea applies.

dcorbacho: Could you point where in the AMQP 1.0 shovel is that credit available? https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_shovel/src/rabbit_amqp10_shovel.erl
Shovel doesn't seem to handle any AMQP 1.0 frames, that is internal to the client.

ansd: It requires changes to the client, for example in https://github.com/rabbitmq/rabbitmq-server/blob/e35909a7b488f7594880c130e5fa210ef90b6418/deps/amqp10_client/src/amqp10_client_session.erl#L1041 the actual number of credits need to be sent to the application instead of just `credited`.

Given the amount of other shovel features and 4.2 approaching, we concluded that AMQP 1.0 credit flow control can be optimised after 4.2 shipped in a separate PR.

deps/rabbitmq_shovel/priv/schema/rabbitmq_shovel.schema
```diff
@@ -9,3 +9,7 @@
 {mapping, "shovel.topology.predeclared", "rabbitmq_shovel.topology.predeclared", [
      [{datatype, {enum, [true, false]}}]
 ]}.
+
+{mapping, "max_local_shovel_credit", "rabbitmq_shovel.max_local_shovel_credit", [
```
michaelklishin: We have a `shovel.*` prefix, for example, the key right above this one uses it.

michaelklishin:
Merging per discussion with @dcorbacho @mkuratczyk.

PR #14304, opened 2025-07-29
Add a configurable limit for number of exchanges

the-mikedavis:
This is like @SimonUnge's past work for [vhosts](https://github.com/rabbitmq/rabbitmq-server/pull/7798) and [queues](https://github.com/rabbitmq/rabbitmq-server/pull/11196) but for limiting the total number of exchanges.

Vhosts and queues are more expensive since they run processes while exchanges are just metadata. A limit would be good for exchanges too though, because if you create a really large number of them you can increase the memory footprint by quite a bit - through taking more space in the metadata store and making more work for monitoring. By default the limit is unset and it can be enabled with `cluster_exchange_limit = 100` or explicitly disabled with `cluster_exchange_limit = infinity`, like the vhost limit.

I've also added some small changes to the Khepri impls for `rabbit_db_exchange`'s `count/0` and `exists/1` so that we use the projection ETS table rather than a query, to make this as cheap as possible.

### DIFF
--- a/deps/rabbit/src/rabbit_exchange.erl
+++ b/deps/rabbit/src/rabbit_exchange.erl
@@ -102,9 +102,13 @@ serial(X) ->
       Internal :: boolean(),
       Args :: rabbit_framing:amqp_table(),
       Username :: rabbit_types:username(),
-      Ret :: {ok, rabbit_types:exchange()} | {error, timeout}.
+      Ret :: {ok, rabbit_types:exchange()} |
+             {error, timeout} |
+             %% May exit with `#amqp_error{}` if validations fail:
+             rabbit_types:channel_exit().
 
 declare(XName, Type, Durable, AutoDelete, Internal, Args, Username) ->
+    ok = check_exchange_limits(XName),
     X = rabbit_exchange_decorator:set(
           rabbit_policy:set(#exchange{name        = XName,
                                       type        = Type,
@@ -139,6 +143,25 @@ declare(XName, Type, Durable, AutoDelete, Internal, Args, Username) ->
             ?LOG_WARNING("ignoring exchange.declare for exchange ~tp,
                                 exchange.delete in progress~n.", [XName]),
             {ok, X}
+    end.
+
+check_exchange_limits(XName) ->
+    Limit = rabbit_misc:get_env(rabbit, cluster_exchange_limit, infinity),
+    case rabbit_db_exchange:count() >= Limit of
+        false ->
+            ok;
+        true ->
+            case rabbit_db_exchange:exists(XName) of
+                true ->
+                    %% Allow re-declares of existing exchanges when at the
+                    %% exchange limit.
+                    ok;
+                false ->
+                    rabbit_misc:protocol_error(
+                      precondition_failed,
+                      "cannot declare ~ts: exchange limit of ~tp is reached",
+                      [rabbit_misc:rs(XName), Limit])
+            end
     end.
 
 %% Used with binaries sent over the wire; the type may not exist.
--- a/deps/rabbit/test/cluster_limit_SUITE.erl
+++ b/deps/rabbit/test/cluster_limit_SUITE.erl
@@ -12,6 +12,7 @@
 -include_lib("amqp_client/include/amqp_client.hrl").
 -compile([nowarn_export_all, export_all]).
 
+-define(EXCHANGE_LIMIT, 10).
 
 all() ->
     [
@@ -22,7 +23,8 @@ groups() ->
     [
      {clustered, [],
       [
-       {size_2, [], [queue_limit]}
+       {size_2, [], [queue_limit,
+                     exchange_limit]}
       ]}
     ].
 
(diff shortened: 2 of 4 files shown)


deps/rabbit/src/rabbit_exchange.erl
```diff
@@ -141,6 +142,25 @@ declare(XName, Type, Durable, AutoDelete, Internal, Args, Username) ->
...
+    Limit = rabbit_misc:get_env(rabbit, exchange_max, infinity),
+    case rabbit_db_exchange:count() >= Limit of
+        false ->
+            ok;
+        true ->
+            case rabbit_db_exchange:exists(XName) of
+                true ->
+                    %% Allow re-declares of existing exchanges when at the
+                    %% exchange limit.
+                    ok;
```
the-mikedavis, line 158: Note: allowing redeclares is important because if you reach the exchange limit and restart a Rabbit node, it will fail to boot due to the exchange declarations done in boot steps

ansd (approved):

SimonUnge (approved):

deps/rabbit/src/rabbit_exchange.erl
```diff
@@ -105,6 +105,7 @@ serial(X) ->
       Ret :: {ok, rabbit_types:exchange()} | {error, timeout}.
```
Ayanda-D: Nice! small comment, can we update the `rabbit_exchange:declare/7` spec to reflect that this can now also return a channel exit / protocol error?

the-mikedavis: I added `no_return()` here and a comment :+1:

Note that this exit was possible before this change too if the exchange's validations failed: https://github.com/rabbitmq/rabbitmq-server/blob/5c26be69857eb9310d44062fc47779e015c417c1/deps/rabbit/src/rabbit_exchange_type.erl#L35-L36

(And this already logs a warning, so there's no need to also log in addition to exiting [like we do for the vhost limit](https://github.com/rabbitmq/rabbitmq-server/blob/5c26be69857eb9310d44062fc47779e015c417c1/deps/rabbit/src/rabbit_vhost.erl#L452-L456))

michaelklishin (changes requested):
`exchange_max` is not consistent with the queue equivalent, `cluster_queue_limit`. Perhaps this new setting should be named `cluster_exchange_limit`. This is a global limit.

Speaking of which, I guess we should also introduce a [per-vhost limit](https://www.rabbitmq.com/docs/vhosts#configuring-max-number-of-queues) as well? Possibly as a follow-up change.

the-mikedavis:
Yeah adding per-vhost limits as a follow up sounds good to me.

The naming is a bit tricky: config options [in this section](https://github.com/rabbitmq/rabbitmq-server/blob/5c26be69857eb9310d44062fc47779e015c417c1/deps/rabbit/priv/schema/rabbit.schema#L898-L1010) tend to follow the `<thing>_max` formula whether they are scoped as per-node like connections or globally like vhosts. I don't have a strong opinion either way - the option name for queues is easy to understand

deps/rabbit/src/rabbit_exchange.erl
Ayanda-D: Cool. I was thinking `rabbit_types:channel_exit()` gives a bit more context here for the type? Can we use that instead? https://github.com/rabbitmq/rabbitmq-server/blob/5c26be69857eb9310d44062fc47779e015c417c1/deps/rabbit_common/src/rabbit_types.erl#L207

michaelklishin:
`*_max` are some of the oldest settings RabbitMQ has. The closest limit equivalent is `cluster_*_limit`.

michaelklishin:
@the-mikedavis thank you for renaming the setting.

I'm afraid we have to wait for a [signed CLA](https://github.com/rabbitmq/cla) to be returned to us by one of your colleagues, which is a work in progress and hopefully won't take too long.

deps/rabbit/test/cluster_limit_SUITE.erl
```diff
@@ -101,48 +104,102 @@ queue_limit(Config) ->
...
+
+    %% Trying to declare the next exchange fails.
+    OverLimitXName = <<"over-limit-x">>,
+    ExpectedError = list_to_binary(rabbit_misc:format(
+                                     "PRECONDITION_FAILED - cannot declare "
+                                     "exchange '~s' in vhost '/': "
+                                     "exchange limit of ~b is reached",
+                                     [OverLimitXName, ?EXCHANGE_LIMIT])),
+    ?assertExit(
+       {{shutdown, {server_initiated_close, 406, ExpectedError}}, _},
```
michaelklishin: It's not a big deal but I would not depend on a specific message in the test. The message can change, arguably the `406 PRECONDITION_FAILED` response is all we care about.

However, I don't feel strongly about this.

michaelklishin (approved):

michaelklishin:
We don't have an obvious place for this new setting to be documented in. 

Well, besides `rabbitmq.conf.example`, which is presently missing all such new-ish global limits.

On the website, there is no obvious guide for such limits. We only have the per virtual host limits in the [Virtual hosts guide](https://www.rabbitmq.com/docs/vhosts) but global limits do not really fit there.

I'll have to think about the documentation part https://github.com/rabbitmq/rabbitmq-website/issues/2328

PR #14348, opened 2025-08-08
Add config option for enabling local_random_exchange

udeeksha30-netizen:
## Proposed Changes

This adds config option which can be used to prevent creation of local random exchanges. If RabbitMQ is behind a load balancer then local random exchanges can't be used effectively, so it would be useful for operators to set enable_local_random_exchange = false to prevent them from being created.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -2592,7 +2592,13 @@ fun(Conf) ->
     end
 end}.
 
+%% Enable or disable local random exchange
+%%
+%% {enable_local_random_exchange, false},
 
+{mapping, "enable_local_random_exchange", "rabbit.enable_local_random_exchange", [
+    {datatype, {enum, [true, false]}}
+]}.
 
 %%
 %% Backing queue version
--- a/deps/rabbit/src/rabbit_exchange_type_local_random.erl
+++ b/deps/rabbit/src/rabbit_exchange_type_local_random.erl
@@ -58,11 +58,12 @@ info(_X) -> [].
 info(_X, _) -> [].
 serialise_events() -> false.
 validate(_X) ->
-    case rabbit_feature_flags:is_enabled(?MODULE) of
+    case rabbit_feature_flags:is_enabled(?MODULE) andalso
+         rabbit_misc:get_env(rabbit, enable_local_random_exchange, true) of
         true ->
             ok;
         false ->
-            rabbit_misc:amqp_error(
+            rabbit_misc:protocol_error(
               precondition_failed,
               "x-local-random exchange feature not available", [],
               'exchange.declare')
--- a/deps/rabbit/test/rabbit_local_random_exchange_SUITE.erl
+++ b/deps/rabbit/test/rabbit_local_random_exchange_SUITE.erl
@@ -21,7 +21,8 @@ groups() ->
     [
      {non_parallel_tests, [], [
                                routed_to_one_local_queue_test,
-                               no_route
+                               no_route,
+                               enable_local_random_exchange_config_test
                               ]}
     ].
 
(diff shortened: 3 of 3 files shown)


deps/rabbit/src/rabbit_exchange_type_local_random.erl
```diff
@@ -58,11 +58,12 @@ info(_X) -> [].
...
-    case rabbit_feature_flags:is_enabled(?MODULE) of
+    case rabbit_feature_flags:is_enabled(?MODULE) andalso
+         rabbit_misc:get_env(rabbit, enable_local_random_exchange, true) of
         true ->
             ok;
         false ->
-            rabbit_misc:amqp_error(
+            rabbit_misc:protocol_error(
               precondition_failed,
               "x-local-random exchange feature not available", [],
```
michaelklishin: It would be great to make this error more specific when the feature is disabled and not just "not available" (due to a disabled feature flag).

deps/rabbit/src/rabbit_exchange_type_local_random.erl
```diff
@@ -58,11 +58,12 @@ info(_X) -> [].
 info(_X, _) -> [].
 serialise_events() -> false.
 validate(_X) ->
-    case rabbit_feature_flags:is_enabled(?MODULE) of
+    case rabbit_feature_flags:is_enabled(?MODULE) andalso
+         rabbit_misc:get_env(rabbit, enable_local_random_exchange, true) of
```
michaelklishin: With modern Erlang versions you can use `application:get_env/3` directly. `rabbit_misc:get_env/3` is from the days when `application:get_env/3` did not yet exist (only `application:get_env/2` did).

`rabbit_misc:get_env/3` these days is defined as

```erl
%% application:get_env/3 is available in R16B01 or later.
get_env(Application, Key, Def) ->
    application:get_env(Application, Key, Def).
```

deps/rabbit/src/rabbit_exchange_type_local_random.erl
```diff
@@ -58,11 +58,12 @@ info(_X) -> [].
...
         true ->
             ok;
         false ->
-            rabbit_misc:amqp_error(
+            rabbit_misc:protocol_error(
```
the-mikedavis, line 73: Note for more context: `amqp_error/4` returns an `#amqp_error{}` while `protocol_error/4` exits with an `#amqp_error{}`. There's a doc in `rabbit_exchange_type` that `validate/1` can exit - this is caught elsewhere and handled gracefully

https://github.com/rabbitmq/rabbitmq-server/blob/edfd8ffedebc0341535a5645b461c65f605c1f11/deps/rabbit/src/rabbit_exchange_type.erl#L35-L36

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -2592,7 +2592,13 @@ fun(Conf) ->
     end
 end}.
 
+%% Enable or disable local random exchange
+%%
+%% {exchange_types.local_random.enabled, false},
```
the-mikedavis:
```suggestion
```

I think we can just remove this part of the comment. Some of the older options at the top of the file show what is set in advanced.config syntax within the rabbit application, but we can see what the option name is in rabbit within the `mapping` below (the `rabbit.local_random_exchange_enabled` part)

deps/rabbit/test/rabbit_local_random_exchange_SUITE.erl
```diff
@@ -21,7 +21,8 @@ groups() ->
     [
      {non_parallel_tests, [], [
                                routed_to_one_local_queue_test,
-                               no_route
+                               no_route,
+                               enable_local_random_exchange_config_test
```
michaelklishin: This test does not enable the local random exchange, it disables it.

michaelklishin (changes requested):
`rabbit_exchange_type_local_random` now uses uses two different app environment keys (`rabbit.local_random_exchange_enabled`, `rabbit.enable_local_random_exchange`), and both `application:get_env/3` and `rabbit_misc:get_env/3`.

michaelklishin (approved):

PR #14356, opened 2025-08-11
Move `rabbit_auth*` from rabbit_common to rabbit

dumbbell:
## Why

These modules are not used by `amqp_client`. Therefore, they shouldn't be in `rabbit_common`.


michaelklishin:
@mergifyio backport `v4.1.x`

dumbbell:
It’s a breaking change at the amqp_client API level, it can’t be backported.

michaelklishin:
@dumbbell yes but if those modules are not actually used by the AMQP 0-9-1 Erlang client, it's in our best interest to reduce the delta between the branches.

PR #14359, opened 2025-08-11
rabbit_access_control: Check configured auth backends are enabled at boot time

dumbbell:
## Why

If a user configures an auth backend module, but doesn't enabled the plugin that provides it, it will get a crash and a stacktrace when authentication is performed. The error is not helpful to understand what the problem is.

## How

We add a boot step that go through the configured auth backends and query the core of RabbitMQ and the plugins. If an auth backend is provided by a plugin, the plugin must be enabled to consider the auth backend to be valid.

In the end, at least one auth backend must be valid, otherwise the boot is aborted.

If only some of the configured auth backends were filtered out, but there are still some valid auth backends, we store the filtered list in the application environment variable so that authentication/authorization doesn't try to use them later.

We also report invalid auth backends in the logs:

* Info message for a single invalid auth backend:

    ```
    [info] <0.213.0> The `rabbit_auth_backend_ldap` auth backend module is configured. However, the `rabbitmq_auth_backend_ldap` plugin must be enabled in order to use this auth backend. Until then it will be skipped during authentication/authorization
    ```

* Warning message when some auth backends were filtered out:

    ```
    [warning] <0.213.0> Some configured backends were dropped because their corresponding plugins are disabled. Please look at the info messages above to learn which plugin(s) should be enabled. Here is the list of auth backends kept after filering:
    [warning] <0.213.0> [rabbit_auth_backend_internal]
    ```

* Error message when no auth backends are valid:

    ```
    [error] <0.213.0> None of the configured auth backends are usable because their corresponding plugins were not enabled. Please look at the info messages above to learn which plugin(s) should be enabled.
    ```

Fixes #13783.

### DIFF
--- a/deps/rabbit/src/rabbit.erl
+++ b/deps/rabbit/src/rabbit.erl
@@ -52,6 +52,14 @@
                     {requires,    pre_boot},
                     {enables,     external_infrastructure}]}).
 
+-rabbit_boot_step({auth_backend_plugins_check,
+                   [{description, "check configured auth plugins are enabled"},
+                    {mfa,         {rabbit_access_control,
+                                   ensure_auth_backends_are_enabled,
+                                   []}},
+                    {requires,    pre_boot},
+                    {enables,     external_infrastructure}]}).
+
 %% rabbit_alarm currently starts memory and disk space monitors
 -rabbit_boot_step({rabbit_alarm,
                    [{description, "alarm handler"},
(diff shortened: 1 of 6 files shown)


dumbbell:
While adding a testsuite, I discovered that the definition of what an "enabled plugin" is varies from function to function in `rabbit_plugins`… Sometimes, it means the plugin was enabled by a user (e.g. using the CLI), sometimes, it means the plugin is running regardless of whether it was enabled or not.

`rabbit_plugins:is_enabled/1` uses the latter definition unfortunately. I need to modily the patch to use `rabbit_plugins:enabled_plugins/0` instead. The drawback is that if an auth backend is enabled implicitly because it is a dependency of another explicitly enabled plugin, the broken won’t start because the new check will consider that the plugin is disabled.

lukebakken (commented):
Hi JSP! This is great 👍

PR #14386, opened 2025-08-14
Quorum queues Raft machine: preserve order for missing deliveries (get_checked_out)

ikavgo:
When I publish some messages, then call checkout in the another client and get {delivery} Effect, consumer client goes into missing messages and the command doesn't keep the order.

So the test:

publish m1
publish m2
publish m3
checkout
publish m4

Before I got:
m3, m2, m1, m4

And now I get:
m1, m2, m3, m4.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -1091,7 +1091,7 @@ handle_aux(_, _, {get_checked_out, ConsumerKey, MsgIds}, Aux0, RaAux0) ->
                                   {S, Acc}
                           end
                   end, {RaAux0, []}, maps:with(MsgIds, Checked)),
-            {reply, {ok,  IdMsgs}, Aux0, RaState};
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -1091,7 +1091,7 @@ handle_aux(_, _, {get_checked_out, ConsumerKey, MsgIds}, Aux0, RaAux0) ->
                                   {S, Acc}
                           end
                   end, {RaAux0, []}, maps:with(MsgIds, Checked)),
-            {reply, {ok,  IdMsgs}, Aux0, RaState};
+            {reply, {ok,  lists:reverse(IdMsgs)}, Aux0, RaState};
```
kjnilsson, line 1094: it needs to be lists:sort/1 instead of lists:reverse, we can't make any assumptions about the maps ordering

ansd:
Closing in favour of https://github.com/rabbitmq/rabbitmq-server/pull/16008

PR #14401, opened 2025-08-19
QQ: when invoking drain only shut down small batches at a time

kjnilsson:
Then wait for elections to complete before shutting further members down.

This should help avoid election storms when enabling maintenance mode.


mkuratczyk (approved):
This works for me but only if I increase the AWAIT_TIMEOUT in ra_log_segment_writer. I think we should increase that timeout as well, otherwise with enough QQs, node startup fails.

michaelklishin:
@mkuratczyk fair enough. What `AWAIT_TIMEOUT` value in `ra_log_segment_writer` did you use? Would doubling it to 60s be enough?

michaelklishin:
For example, this PR doubles the aforementioned timeout https://github.com/rabbitmq/ra/pull/556.

michaelklishin:
@mergifyio backport `v4.1.x`

PR #14425, opened 2025-08-25
Shovel: introduce operating modes

michaelklishin:
Sometimes you want a plugin to act as a library
and not an application. That is, for its modules
to be available at compile time and on a running
node but, say, the actual runtime parameter
handling and supervision of shovels to be
handled by another plugin.

Since we do not currently have a concept of
"library plugins" or "library dependencies",
this approach demonstrates one example of how
some plugins can be used effectively as libraries.

Per discussion with @mkuratczyk @Zerpet.

### DIFF
--- a/deps/rabbitmq_shovel/src/rabbit_shovel_operating_mode.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_shovel_operating_mode.erl
@@ -1,0 +1,34 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
+%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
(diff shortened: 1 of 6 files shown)


deps/rabbitmq_shovel/src/rabbit_shovel_operating_mode.erl
```diff
@@ -0,0 +1,34 @@
...
+
+%%
+%% API
+%%
+
+-spec operating_mode() -> operating_mode().
+operating_mode() ->
+    application:get_env(?SHOVEL_APP, operating_mode, standard).
+
+-spec is_standard() -> boolean().
```
lhoguin: You don't need the two "is" functions, you can just match on `standard` / `alternative` in case/function clauses. This avoids boolean blindness.

PR #14438, opened 2025-08-27
Fix issue where leader is not returned after stream declaration

kjnilsson:
This would affect the meta data that is returned when declaring a queue through AMQP.


Zerpet:
I confirm that this PR fixes the test error reported in https://github.com/rabbitmq/rabbitmq-amqp-dotnet-client/pull/131. The test was failing because `leader` was missing in the response of queue declaration over AMQP. In this PR, the test passes.

PR #14445, opened 2025-08-28
DO NOT MERGE v4.1.x consistent logging

mkuratczyk:
Backport of https://github.com/rabbitmq/rabbitmq-server/pull/14232 for v4.1.x.
There were many conflicts so it's not a straightforward backport. Some automation
was used to resolve conflicts. Some indentation is wrong, but hopefully we can ignore this
(it's good, or at least better, on `main`).


michaelklishin:
I agree, the only reason for backporting #14232 is to avoid conflicts in many other PRs, both present and future ones, because logging touches every subsystem and every module.

Thank you!

PR #14449, opened 2025-08-28
Fix in OCI trigger for PRs

Zerpet:
The output is of type bool, and the comparison was checking for string
'true'. In JS 'true' is not equal true (bool) :shrug:

I can't quite prove the fix because I have permissions to push. We will see if this
fix is effective when a new external PR is submitted.


michaelklishin:
@mergifyio backport `v4.2.x`

PR #14454, opened 2025-08-29
rabbit_ff_controller: Raise log levels of first-time boot messages

the-mikedavis:
This is a very minor change to adjust log levels of some lines I found useful recently. These lines can be very useful for determining why a subset of feature flags is/isn't being enabled on boot. They're currently at debug level. However a first-time boot of an unclustered node is a rare event, so these lines shouldn't be noisy.


michaelklishin:
@mergifyio backport v4.1.x

Issue #14202, opened 2025-07-08
Quorum queue delivery count set to unlimited can cause the number segment files to increase without truncation


Mott88:
### Describe the bug

As described in the documentation you can set the delivery limit for quorum queues to unlimited for rare use cases: https://www.rabbitmq.com/docs/quorum-queues#disabling-the-limit
"The old unlimited delivery-limit behaviour can be restored by setting a queue argument or policy with a delivery limit of -1. It is not recommended to do so but may be needed for 3.13.x compatibility in some rare cases."

If you add a policy that sets the delivery limit for a quorum queue to -1, RabbitMQ starts re-applying policies to the queue every 5 second, which causes the number of segment files to increase because there is added a "update_config" log event every time re-apply policies is run.

I'm no developer and have little Erlang experience, but I think this happens because the **maybe_apply_policies** is run every tick
https://github.com/rabbitmq/rabbitmq-server/blob/3c28315e83328d639704d1809d6e7ae5798a0275/deps/rabbit/src/rabbit_quorum_queue.erl#L683

Then it seems like the -1 in deliverylimit causes the NewPolicyConfig not to be equal to CurrentPolicyConfig" which  triggers the update_config event:
https://github.com/rabbitmq/rabbitmq-server/blob/3c28315e83328d639704d1809d6e7ae5798a0275/deps/rabbit/src/rabbit_quorum_queue.erl#L776

### Reproduction steps

1. Install RabbitMQ 4.1.2
2. Add raft.wal_max_size_bytes = 1000 to rabbitmq.conf to make the write-ahed-log rotate quickly.
3. Start RabbitMQ.
4. Create a single quorum queue named q1 and publish a message to the queue.
5. Enable debug logging for RabbitMQ.
6. Add a quorum queue policy that sets the delivery limit for the queue you created to -1.
7. Log entries with " Re-applying policies to queue" will start to show up in the log every 5 second.
8. The write-ahed-log will start to grow every 5 second, because there is added an update_config log event everytime it is trying to re apply policies. When the wal gets rotated these events is added to the queue segment files. If the queue has no traffic (no active consumers or messages published), the number of segment files will continue to increase with 4096 update_config events for each segment file. To list the number of update_config events in a segment file run the command `strings 0000000000000002.segment | grep -o "update_config" | wc -l`

### Expected behavior

Don't expect RabbitMQ to apply the policy every 5 second.

kjnilsson:
[...]

PR #14458, opened 2025-08-29
QQ: exclude delivery_limit from policy repair comparison.

kjnilsson:
As it cannot be updated as part of a policy when it is disabled (-1)
 it needs to be excluded from comparison

Fixes https://github.com/rabbitmq/rabbitmq-server/issues/14202

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -770,7 +770,10 @@ system_recover(quorum_queues) ->
     end.
 
 maybe_apply_policies(Q, #{config := CurrentConfig}) ->
-    NewPolicyConfig = gather_policy_config(Q, false),
+    %% delivery_limit can't be updated from a policy and thus has to be
+    %% excluded from the comparison
+    NewPolicyConfig = maps:without([delivery_limit],
+                                   gather_policy_config(Q, false)),
 
     RelevantKeys = maps:keys(NewPolicyConfig),
     CurrentPolicyConfig = maps:with(RelevantKeys, CurrentConfig),
@@ -778,7 +781,8 @@ maybe_apply_policies(Q, #{config := CurrentConfig}) ->
     ShouldUpdate = NewPolicyConfig =/= CurrentPolicyConfig,
     case ShouldUpdate of
         true ->
-            ?LOG_DEBUG("Re-applying policies to ~ts", [rabbit_misc:rs(amqqueue:get_name(Q))]),
+            ?LOG_DEBUG("Re-applying policies for ~ts",
+                       [rabbit_misc:rs(amqqueue:get_name(Q))]),
             policy_changed(Q),
             ok;
         false -> ok


gomoripeti:
is it only when it is disabled (-1) that it cannot be updated from a policy?
can it be updated eg from 20 to 40? if it can then this change will prevent the periodic policy repair to notice such change?

I really did not investigate thoroughly to find out what prevents the disabled delivery limit to be changed by a policy but this code part seems to allow it https://github.com/rabbitmq/rabbitmq-server/blob/v4.2.0-beta.2/deps/rabbit/src/rabbit_fifo.erl#L227.

isn't to problem of continuous update_config that rabbit_fifo converts a negative delivery limit to `undefined` so when `maybe_apply_policies` compares current config with new config, the former contains `undefined` while the later contains a negative value for the same unchanged "disabled" delivery limit? 
https://github.com/rabbitmq/rabbitmq-server/blob/4d7a3fed1e6fb6fbe0695ad5acf18d34d157039b/deps/rabbit/src/rabbit_quorum_queue.erl#L781

I will shortly link a commit to show what I mean. (Sorry I just came back from vacation and couldn't look at this issue earlier)

kjnilsson:
> is it only when it is disabled (-1) that it cannot be updated from a policy?
> can it be updated eg from 20 to 40? if it can then this change will prevent the periodic policy repair to notice such change?

yes good point, this will probably disable "valid" repairs. that said with the defaulting logic I feel it is probably safest to exclude this key irrespectively.

gomoripeti:
This change https://github.com/rabbitmq/rabbitmq-server/commit/f855eb5676e6d97940aefb2de3ab2913e57b7d41 prevents update_config every tick if delivery-limit is negative, but still allows the periodic check to detect when delivery-limit is changed via policy. What do you think?

kjnilsson:
> This change [f855eb5](https://github.com/rabbitmq/rabbitmq-server/commit/f855eb5676e6d97940aefb2de3ab2913e57b7d41) prevents update_config every tick if delivery-limit is negative, but still allows the periodic check to detect when delivery-limit is changed via policy. What do you think?

yes I think this is fine - do we need the `not IsQueueDeclaration` guard?

gomoripeti:
> do we need the `not IsQueueDeclaration` guard?

Not necessarily. Just wanted to highlight that the conversion is necessary only for the comparison, for the actual config update it does not matter.
Does this conversion (https://github.com/rabbitmq/rabbitmq-server/blob/v4.2.0-beta.2/deps/rabbit/src/rabbit_fifo.erl#L203-L205) still needed if the guard is removed and the fifo_client passes the disabled value as `undefined`? (I guess yes because of mixed version clusters or when applying older raft logs)

I will add tests and submit a PR early next week.

kjnilsson:
I have been doing some testing and your change allows the delivery_limit to be disable _after_ it has been active, this didn't use to be the case and was a deliberate choice. We could ofc extract the delivery_limit key in the comparison and compare them using adjusted logic but I'm not convinced it is worth it. Especially as we are probably going to make some adjustments to deliver limit defaulting once we introduce Ra v3 (Which is better able to cope with repeated returns).

gomoripeti:
thanks for testing and looking Karl. Don't want to delay the fix for https://github.com/rabbitmq/rabbitmq-server/issues/14202 so I'm fine with merging your PR which has well understood behaviour.

To continue the discussion do you mean it is not allowed to disable after active on main or by your PR?

My change (especially with the `not IsQueueDeclaration` guard) and your change as well only modify when a config update is triggered in case the queue would miss the `policy_changed` notification, but does not change what values the config is updated with. So I don't think it is modified which transitions are allowed.

kjnilsson:
yes I think just removing the key from comparison is fine for now, Once we ship Ra v3 (with log compaction) the restrictions on delivery limits may well be lifted and we will start allowing unlimited returns to the queue (potentially). So we can keep it simple for now even if it means that we'd not repair the case where only the delivery limit has changed.

PR #14480, opened 2025-09-02
rabbitmq_shovel_prometheus: handle single atom shovel states

michaelklishin:
Namely, 'starting'.

Previously, a Prometheus scraping endpoint hit
just at the right moment (when a shovel was
in the 'starting' state) would produce
a function_clause exception.


michaelklishin:
@mergifyio backport `v4.2.x`

michaelklishin:
Manually backported to `v4.2.x`.

PR #14488, opened 2025-09-02
Local shovels: send confirms and rejections before autodelete

dcorbacho:


michaelklishin:
@dcorbacho note that `main` now targets `4.3.0`, so the most logical thing IMO is to set the milestone for `main` PRs accordingly :) and then the `v4.2.x` backport's milestone must be set to `4.2.0` so that we consider them for the release notes.

michaelklishin (approved):

PR #14489, opened 2025-09-02
Delete amqp_filtex_SUITE from ct.test.spec

ansd:
This test suite doesn't exist anymore.


lhoguin:
ct.test.spec is not used in CI, it's used with its own target to run "everything" locally. But if nobody uses it we can remove it and the target.

PR #14493, opened 2025-09-02
rabbit_fifo: Expose each smallest raft index in `overview/1`

the-mikedavis:
This is meant to be used as a debugging tool via an eval command. It can be hard to tell from the outside why a QQ is not snapshotting. Looking at the `overview/1` is a nice debugging tool. The `smallest_raft_index` key (and other info like the `release_cursors`) point out why a snapshot isn't being taken. The `smallest_raft_index` is the minimum of three criteria though:

* oldest ready message
* oldest checked out message
* oldest message in the dlx state

This change adds all of those indexes so that it is clear which criteria is keeping the `smallest_raft_index` down.


kjnilsson (approved):

michaelklishin:
@the-mikedavis thank you. I will backport this all the way to `v4.1.x` but please note that the next `4.1.x` release is now expected around `4.2.0` in the 2nd half of October.

PR #14495, opened 2025-09-02
Shovel: extract shovel_test_utils

michaelklishin:
Just like stream_test_utils, they arguably
belong to rabbitmq_ct_helpers.


michaelklishin:
The module depends on `rabbit_shovel_status` and cannot be moved out without changes.

PR #14507, opened 2025-09-04
Speed up Direct Reply-To

ansd:
Speed up Direct Reply-To in AMQP 0.9.1. Prior to this PR sending each RPC reply was bottlenecked by the slowest network link between the RabbitMQ node the responder connected to and all other RabbitMQ nodes due to calling `erpc:multicall`. This regression was introduced in RabbitMQ 3.12.0 via https://github.com/rabbitmq/rabbitmq-server/commit/d65637190a48129925eec4da16bc9aa01459e90a


ansd:
This should be backported all the way back to `v3.13.x`.

michaelklishin (approved):

kjnilsson (approved):

PR #14530, opened 2025-09-12
Optimise Khepri for concurrent reads

ansd:
This commit makes read operations for the following Khepri projections much cheaper:
* rabbit_khepri_queue
* rabbit_khepri_exchange
* rabbit_khepri_index_route
* rabbit_khepri_topic_trie

Entries in these ETS tables are read for every message entering RabbitMQ. Some messages entering RabbitMQ will cause even multiple reads from these ETS tables, e.g. multiple reads from `rabbit_khepri_queue` if a message is routed to more than one queue or multiple reads from `rabbit_khepri_index_route` if a message has multiple routing keys.

On a busy RabbitMQ node, these tables are read concurrently (on multiple physical processors) hundreds of thousands of times per second.

### DIFF
--- a/deps/rabbit/src/rabbit_khepri.erl
+++ b/deps/rabbit/src/rabbit_khepri.erl
@@ -1344,7 +1344,7 @@ register_rabbit_exchange_projection() ->
                     _VHost = ?KHEPRI_WILDCARD_STAR,
                     _Name = ?KHEPRI_WILDCARD_STAR),
     KeyPos = #exchange.name,
-    register_simple_projection(Name, PathPattern, KeyPos).
+    register_simple_projection(Name, PathPattern, KeyPos, true).
 
 register_rabbit_queue_projection() ->
     Name = rabbit_khepri_queue,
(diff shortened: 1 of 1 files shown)


dumbbell:
Did you measure the impact?

Other than that, the patch looks good to me.

the-mikedavis:
I agree, setting read_concurrency sounds good to me for these tables. I wonder if it would worsen performance when there is queue or binding churn? And if it does would it matter anyways (since churn is already bad)?

ansd:
> Did you measure the impact?

When I start the broker as follows:
``` 
make run-broker LEAVE_PLUGINS_DISABLED=1
```
and generate load as follows:
```
java -jar target/perf-test.jar -x 8 -y 1 -s 12 -z 60 --autoack
```
I get around
```
id: test-144220-533, sending rate avg: 166958 msg/s
id: test-144220-533, receiving rate avg: 166958 msg/s
```
prior to this PR and around
```
id: test-144028-283, sending rate avg: 170046 msg/s
id: test-144028-283, receiving rate avg: 170042 msg/s
```
after this PR.

But there is quiet a high variability in the different test runs.

That read concurrency option is about contention. There are better workloads to trigger this contention, for example by using a machine with many more CPU cores (e.g. 16 or more) and many more channels. But I haven't taken the time to investigate that deeply.

From the few tests I did, performance seems to get slightly better.

> I wonder if it would worsen performance when there is queue or binding churn? And if it does would it matter anyways (since churn is already bad)?

It might likely worsen performance for queue or binding churn. I haven't taken the time to test it. I would argue it's more important to optimise RabbitMQ for high message throughput than for creation and deletion of queues :)

I leave it up to both of you whether to merge or not :)

ansd:
FWIW we set this option for the routing table in Mnesia: https://github.com/rabbitmq/rabbitmq-server/blob/5b9f98ac6a603e5d2ab8177da9e498f03fc51841/deps/rabbit/src/rabbit_table.erl#L469

PR #14543, opened 2025-09-15
Fix regression with Khepri binding args

### LINKED ISSUE #14533: Regression with Khepri and binding arguments
### Describe the bug

Khepri handles routing decision incorrectly when multiple bindings are created to a source exchange of exchange type `direct` or `fanout` when different binding arguments are used.

Note that such binding arguments are entirely ignored when RabbitMQ performs routing decisions for the `direct` or `fanout` exchange. However, there might be client apps that use binding arguments to add some metadata to the binding, for example `app-id` or `user` or `purpose` and might use this metadata as a form of reference counting in deciding when to delete `auto-delete` exchanges or just for informational/operational purposes. So, this bug only occurs for edge cases.

### Reproduction steps

On branch https://github.com/rabbitmq/rabbitmq-server/tree/binding-args-khepri-bug run
```
make -C deps/rabbit ct-bindings t=cluster:binding_args RABBITMQ_METADATA_STORE=mnesia
```
and
```
make -C deps/rabbit ct-bindings t=cluster:binding_args RABBITMQ_METADATA_STORE=khepri
```
The test succeeds with Mnesia and fails with Khepri.

### Expected behavior

Khepri routes correctly.

### Additional context

The problem is that ETS table `rabbit_khepri_index_route` cannot
differentiate between two bindings with different binding arguments, and
therefore deletes entries too early, leading to wrong routing decisions.

The solution to this bug is to include the binding arguments in the
`rabbit_khepri_index_route` projection, similar to how the binding args
are also included in the `rabbit_index_route` Mnesia table

ansd:
Fixes #14533

This PR fixes the bug for clusters created in >=4.2.
For clusters being upgraded from <4.2 to >= 4.2, the bug still exists, which may be acceptable given the edge case for the bug to be triggered. To fix the bug for upgrades <4.2, RabbitMQ needs a way to migrate projections. One option is to unregister and register the projection in the `rabbitmq_4.2.0` feature flag migration callback. This would require to block routing while this migration function runs.

### DIFF
--- a/deps/rabbit/src/rabbit_khepri.erl
+++ b/deps/rabbit/src/rabbit_khepri.erl
@@ -60,7 +60,7 @@
 %% executed. If the migration runs concurrently, whether it started before or
 %% during the execution of the Mnesia-specific anonymous function, {@link
 %% handle_fallback/1} will watch for "no exists" table exceptions from Mnesia
-%% and will retry the Mnesia functino or run the Khepri function accordingly.
+%% and will retry the Mnesia function or run the Khepri function accordingly.
 %% The Mnesia function must be idempotent because it can be executed multiple
 %% times.
 %%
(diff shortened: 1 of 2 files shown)


dumbbell:
@ansd says in an internal chat:
> The PR fixes this issue for clusters created in 4.2+. However, if you upgrade from 4.1 with Khepri enabled previously, the bug still occurs. That's because the old projection table is still being used. The question is how do you upgrade projections?
>
> One option I can think of: deregister the projection and then register again in the `rabbitmq_4.2.0` feature flag migration function. That means we need to block routing though while the migration function runs.
>
> Another option I can think of: run a migration in the `machine_version` upgrade. But it seems this API is meant to be private to the Khepri app and not meant to be hooked into by the rabbit app.
>
> Yet another option for this specific PR: Provide the fix as is for new clusters created in 4.2+. Given this is a rare edge case, we can leave it unfixed for clusters created in 4.1.

Another option is to create a new projection; in other words, rename the current one from `rabbit_khepri_index_route` to `rabbit_khepri_index_route_v2`. The node running RabbitMQ 4.2 will get the new fixed projection and an old node will continue to use the old projection. We could then unregister the old projection in the migration function of the `rabbitmq_4.2.0` feature flag. I think this is a good middle ground: the old node continues to suffer from the bug, but as you said, that’s probably ok because it is a rare use case.

@the-mikedavis: Do you have an opinion?

the-mikedavis:
Adding a v2 and eventually unregistering the v1 in `rabbitmq_4.2.0` also sounds like a good middle-ground to me :+1:

ansd:
Thank you! Just that I understand correctly: Should the v2 projection be registered upon node boot? This means it becomes available on the old nodes too in a mixed version cluster, but won't cause there any harm, right?

dumbbell:
Yes, I confirm that the newer node will initialise the new projection on boot and the ETS table will appear on all nodes, including an old one. That’s ok because the old node won’t use the new projection. The new node just need to support the new projection.

ansd:
Closing this PR in favour of https://github.com/rabbitmq/rabbitmq-server/pull/14546

PR #14545, opened 2025-09-15
Check for protected tag during user update and deletion via management API

sunfinite:
## Proposed Changes

Implements `protected` users proposed in [rabbitmq#14282](https://github.com/rabbitmq/rabbitmq-server/discussions/14282)

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

I will update relevant [documentation](https://github.com/rabbitmq/rabbitmq-website/blob/d4b76ff21082a7cbb8ca556e0f5f99eb40baecd2/docs/management/index.md?plain=1#L191) once this change is reviewed.

### DIFF
--- a/deps/rabbitmq_web_dispatch/src/rabbit_web_dispatch_access_control.erl
+++ b/deps/rabbitmq_web_dispatch/src/rabbit_web_dispatch_access_control.erl
@@ -24,7 +24,7 @@
 -export([id/2]).
 -export([not_authorised/3, halt_response/5]).
 
--export([is_admin/1, is_policymaker/1, is_monitor/1, is_mgmt_user/1]).
+-export([is_admin/1, is_policymaker/1, is_monitor/1, is_mgmt_user/1, is_protected_user/1]).
 
 -import(rabbit_misc, [pget/2]).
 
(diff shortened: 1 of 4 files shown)


deps/rabbitmq_web_dispatch/src/rabbit_web_dispatch_access_control.erl
```diff
@@ -243,6 +243,7 @@ is_policymaker(T) -> intersects(T, [administrator, policymaker]).
 is_monitor(T)     -> intersects(T, [administrator, monitoring]).
 is_mgmt_user(T)   -> intersects(T, [administrator, monitoring, policymaker,
                                     management]).
+is_protected_user(T) -> intersects(T, [protected]).
```
michaelklishin, line 246: Since you are looking up a single list element, `lists:member/2` would do. We don't have to optimize this rarely called function but `lists:member/2` is also more specific.

michaelklishin (changes requested):
This looks close to what I had in mind.

the-mikedavis:
@michaelklishin what do you think about including this in 4.2.0? To me it seems low-risk, but on the other hand I know we're getting pretty close to the 4.2 release now. I can shepherd in the backport if you think it's alright

michaelklishin:
@the-mikedavis sorry, it wasn't backported to `v4.2.x` by accident :)

PR #14546, opened 2025-09-15
Speed up fanout exchange

### LINKED ISSUE #14531: Optimise fanout exchange

The fanout exchange can be optimised. A lot of CPU time is wasted due to `ets:select/2` compiling the same match spec for every message.

For example, if a single classic queue is bound to a fanout exchange and a client publishes to that fanout exchange as fast as possible `rabbit_exchange:route()` wastes about 12% of CPU time.

Do not use `ets:select/2`.

Possible solutions (just brainstorming, haven't tried them out):
* Use another routing projection where the ETS table key is only the `SourceExchange` rather than `{SourceExchange, BindingKey}`. Downside: more memory usage for that additional ETS table
* Ignore the binding key for fanout exchanges and store the binding key as the empty string, a bit similarly as being forced in the local random exchange in https://github.com/rabbitmq/rabbitmq-server/blob/5b9f98ac6a603e5d2ab8177da9e498f03fc51841/deps/rabbit/src/rabbit_exchange_type_local_random.erl#L103-L106 Downside: backward compat issues?

### Additional context

Similar enhancement was done in https://github.com/rabbitmq/rabbitmq-server/pull/4606

Some other exchange types (in addition to the fanout exchange) suffer exactly the same performance issue:
* rabbit_exchange_type_recent_history
* rabbit_exchange_type_random
* rabbit_sharding_exchange_type_modulus_hash

### LINKED ISSUE #14533: Regression with Khepri and binding arguments

ansd:
Resolves #14531 
Resolves #14533

 ## What?
Increase end-to-end message throughput for messages routed via the fanout exchange by ~42% (see benchmark below).
In addition to the fanout exchange, a similar speed up is achieved for the following exchange types:
* modulus hash
* random
* recent history

This applies only if Khepri is enabled.

 ## How?
Use an additional routing table (projection) `rabbit_khepri_route_by_source` whose table key is the source exchange.
Looking up the destinations happens then by an ETS table key.

Prior to this commit, CPUs were busy compiling the same match spec for every incoming message.

For the bug fix of #14533, we also introduce a new projection called `rabbit_khepri_route_by_source_key`. This can be thought of a v2 of the old `rabbit_khepri_index_route` projection. The old `rabbit_khepri_index_route` projection gets deleted in the `rabbitmq_4.2.0` feature flag callback. The two new projections are registered at boot time (as discussed in https://github.com/rabbitmq/rabbitmq-server/pull/14543#issuecomment-3292865179)

 ## Benchmark
1. Start RabbitMQ:
```
make run-broker RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+S 5" \
    RABBITMQ_CONFIG_FILE="advanced.config" PLUGINS="rabbitmq_management"
```
where `advanced.config` contains:
```
[
 {rabbitmq_management_agent, [
  {disable_metrics_collector, true}
 ]}
].
```

2. Create a queue and binding:
```
deps/rabbitmq_management/bin/rabbitmqadmin declare queue queue_type=classic durable=true name=q1 && \
deps/rabbitmq_management/bin/rabbitmqadmin declare binding source=amq.fanout destination=q1
```

3. Create the load
```
java -jar target/perf-test.jar -p -e amq.fanout -u q1 -s 5 --autoack -z 60
```

Before this commit:
```
sending rate avg: 97394 msg/s
receiving rate avg: 97394 msg/s
```

After this commit:
```
sending rate avg: 138677 msg/s
receiving rate avg: 138677 msg/s
```

The CPU flamegraph shows that `rabbit_exchange:route/3` consumes the following CPU amounts:
* 13.5% before this commit
* 3.4% after this commit

 ## Downsides
Additional ETS memory usage for the new projection table.
However, the new table does not store any binding entries for the following
source exchange types:
* direct
* headers
* topic
* x-local-random
* x-jms-topic

Also, care must be taken to not add extensive amounts of bindings with the same source exchange to projection `rabbit_khepri_route_by_source`. For fanout exchanges, binding an extensive amount of queues (e.g. thousands) to the same fanout exchanges is nonsensical anyway. One edge case could be binding thousands of queues to the `random` exchange type. But even such an exotic use case could be realised efficiently by using intermediate random exchanges (i.e. exchange-to-exchange bindings) 🙂

### DIFF
--- a/deps/rabbit/src/rabbit_db_binding.erl
+++ b/deps/rabbit/src/rabbit_db_binding.erl
@@ -708,8 +708,15 @@ match_routing_key_in_mnesia(SrcName, RoutingKeys, UseIndex) ->
     end.
 
 match_routing_key_in_khepri(Src, ['_']) ->
-    ets:lookup_element(?KHEPRI_ROUTE_BY_SOURCE_PROJECTION,
-                       Src, #route_by_source.destination, []);
+    try
+        ets:lookup_element(?KHEPRI_ROUTE_BY_SOURCE_PROJECTION,
+                           Src,
+                           #route_by_source.destination,
+                           [])
+    catch
+        error:badarg ->
+            []
+    end;
 match_routing_key_in_khepri(Src, RoutingKeys) ->
     lists:foldl(
       fun(RK, Acc) ->


deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -1414,35 +1426,78 @@ register_rabbit_bindings_projection() ->
...
+    %% advertises this access pattern, e.g. as a boolean flag in the #exchange.options field.
+    Exchange = #if_data_matches{
+                  pattern = #exchange{type = '$1', _ = '_'},
+                  conditions = [{'andalso',
+                                 {'=/=', '$1', headers},
+                                 {'=/=', '$1', topic},
+                                 {'=/=', '$1', direct},
+                                 {'=/=', '$1', 'x-local-random'},
+                                 {'=/=', '$1', 'x-jms-topic'}
+                                }]},
```
ansd, line 1484: The approach in this PR is good enough in my opinion.

To implement what I suggested in the code comment, we'd need yet another migration of exchange records, which I would like to avoid for now.

Yet another approach I can think of is adding a new required callback to `rabbit_exchange_type` such that every exchange type implementation has to advertise its routing lookup/access pattern. However, it seems they can't be used in these Khepri trigger conditions. For example, writing the following trigger condition isn't supported:
```erl
rabbit_exchange:route_access_pattern(ExchangeType) =:= source
```

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -332,8 +333,10 @@ init(IsVirgin) ->
                    "up to the Raft cluster leader", [],
                    #{domain => ?RMQLOG_DOMAIN_DB}),
                 ok ?= case IsVirgin of
-                          true  -> register_projections();
-                          false -> ok
+                          true ->
+                              register_projections();
+                          false ->
+                              register_4_2_0_projections()
```
dumbbell: All projections should probably be registered everytime. Khepri won’t modifly one that is already registered (based on its name).

deps/rabbit/src/rabbit_khepri.erl
dumbbell: Yeah, this needs to always give the same result on all nodes. If we were to support function call in this pattern, we would have to extract the function and copy its bytecode in the pattern.

deps/rabbit/src/rabbit_khepri.erl
ansd: Good point :) I will try this out. Thank you!

deps/rabbit/src/rabbit_khepri.erl
ansd: Ok, let's keep the current approach then, and leave the comment in the code as a reminder for a future optimisation.

deps/rabbit/src/rabbit_db_binding.erl
```diff
@@ -707,21 +708,14 @@ match_routing_key_in_mnesia(SrcName, RoutingKeys, UseIndex) ->
...
-        MatchHead = #index_route{source_key  = {Src, '_'},
-                                 destination = '$1',
-                                 _           = '_'},
-        ets:select(?KHEPRI_INDEX_ROUTE_PROJECTION, [{MatchHead, [], ['$1']}])
-    catch
-        error:badarg ->
-            []
-    end;
+    ets:lookup_element(?KHEPRI_ROUTE_BY_SOURCE_PROJECTION,
+                       Src, #route_by_source.destination, []);
```
the-mikedavis: I think we want to keep the `try`/`catch` here, see https://github.com/rabbitmq/rabbitmq-server/pull/11667 / https://github.com/rabbitmq/rabbitmq-server/pull/11700

ansd: Thank you. Fixed.

the-mikedavis (approved):
👍

I see similar results locally for that perf-test scenario. `main`:

```
id: test-100516-125, sending rate avg: 95139 msg/s
id: test-100516-125, receiving rate avg: 95139 msg/s
```

This PR:

```
id: test-100853-180, sending rate avg: 140581 msg/s
id: test-100853-180, receiving rate avg: 140581 msg/s
```

ansd:
As always, thanks a lot to both of you @dumbbell @the-mikedavis  for your thorough reviews!

Issue #14551, opened 2025-09-16
HTTP API: `consumer_utilisation` can be excluded from queue metrics if it hasn't been emitted yet (is undefined)


chengjoey:
### Describe the bug

<img width="520" height="259" alt="Image" src="https://github.com/user-attachments/assets/9a84fd64-7882-4643-8f58-9d688e630dea" />

Our program will request the queue information from the management's /api/queues, but we sometimes find that the `consumer_utilisation` field is missing from the queue information.

### Reproduction steps

It is difficult to reproduce, please ask under what circumstances the consumer_utilisation field may be lost.

### Expected behavior

consumer_utilisation could get from /ap/queues

chengjoey:
I tried to directly curl the rabbitmq management's /api/queues interface and found that the `consumer_utilisation` indeed disappeared in some queue information.

RabbitMQ server version: 3.9.13/3.9.28

Can I add a judgment in `rabbit_amqqueue_process.erl`, if there is no `consumer_utilisation`, set it to 0?

michaelklishin:
@chengjoey thank you for the PR. The change is reasonable but I cannot help but mention that RabbitMQ 3.9 has reached EOL so long ago that it is not [even listed on the release series support page](https://www.rabbitmq.com/release-information).

michaelklishin:
@chengjoey also, using `GET /api/queues` without pagination is **highly** recommended against, it produces 70 or so metrics per queue and once you get into thousands of queues, the JSON responses begin burning a lot of CPU and network bandwidth resources.

Use [Prometheus and Grafana](https://www.rabbitmq.com/docs/prometheus) instead, at least this way every node reports its own metrics and not 70+ metrics for every queue across the entire cluster.

chengjoey:
@michaelklishin thanks!

We have a plan to upgrade our rabbitmq server version, currently we are using the classic queue, and we are working on a plan to upgrade to the Quorum queue.

Additionally, I have already passed the cla certification for `cluster_operator`, may I also sign the cla for rabbitmq-server so that #14552 can be merged into main, at least to keep the fields in the queue info from being missing without modifying our program code.

chengjoey:
CLA has been signed

https://github.com/rabbitmq/cluster-operator/pull/1954

michaelklishin:
[...]

PR #14552, opened 2025-09-16
Queue metrics: ff consumer utilization is not yet computed, report it as 0 instead of `undefined`

chengjoey:
## Proposed Changes

Fixes: #14551 

## Types of Changes

- [x] Build system and/or CI

## Checklist

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue_process.erl
+++ b/deps/rabbit/src/rabbit_amqqueue_process.erl
@@ -1156,7 +1156,10 @@ i(messages, State) ->
 i(consumers, _) ->
     rabbit_queue_consumers:count();
 i(consumer_utilisation, State) ->
-    i(consumer_capacity, State);
+  case i(consumer_capacity, State) of
+    undefined -> 0;
+    Value -> Value
+  end;
 i(consumer_capacity, #q{consumers = Consumers}) ->
     case rabbit_queue_consumers:count() of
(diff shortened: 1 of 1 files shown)


michaelklishin:
@chengjoey thank you. In order to accept this contribution we need you to [digitally] [sign our contributor CLA](https://github.com/rabbitmq/cla).

michaelklishin:
A better option would be to do this for `consumer_capacity`, which is a newer name for `consumer_utilisation`. I'll do it.

michaelklishin:
Looking at the code, I am not sure `i(consumer_capacity)` would ever return `undefined` in `main` or `v4.1.x`:

`i(consumer_utilisation)` delegates to `i(consumer_capacity)`, which uses `rabbit_queue_consumers:capacity/1`, which, per its type spec, can only return `ratio()`, that is, a float. That function ends up in `rabbit_misc:moving_avg/4` and I don't see how an `undefined` would be handled there without an exception.

In several statically typed HTTP API clients, this field is not an `Option` but rather a float (e.g. `f32` in Rust).

So the fix here is to [upgrade](https://www.rabbitmq.com/docs/upgrade) to RabbitMQ 4.1.x [using a Blue-Green deployment](https://www.rabbitmq.com/blog/2025/07/29/latest-benefits-of-rmq-and-migrating-to-qq-along-the-way).

PR #14560, opened 2025-09-17
rabbitmq-diagnostics message_size_stats

mkuratczyk:
This command displays cluster-wide message size
statistics. It's less detailed than what can be
retrieved from the Prometheus endpoint, but
it'll be available to all users, regardless of their monitoring setup, or lack thereof.

Example:
```
$ rabbitmq-diagnostics -n rabbit-1 message_size_stats
Gathering message size statistics from cluster via rabbit-1@K6L59PF0JR ...
┌────────────────────────┬───────┬────────────┐
│ Message Size           │ Count │ Percentage │
├────────────────────────┼───────┼────────────┤
│ below 100B             │ 1409  │ 19.90      │
├────────────────────────┼───────┼────────────┤
│ between 100B and 1KB   │ 2847  │ 40.21      │
├────────────────────────┼───────┼────────────┤
│ between 1KB and 10KB   │ 40    │ 0.56       │
├────────────────────────┼───────┼────────────┤
│ between 10KB and 100KB │ 2785  │ 39.33      │
├────────────────────────┼───────┼────────────┤
│ between 100KB and 1MB  │ 0     │ 0.00       │
├────────────────────────┼───────┼────────────┤
│ between 1MB and 10MB   │ 0     │ 0.00       │
├────────────────────────┼───────┼────────────┤
│ between 10MB and 50MB  │ 0     │ 0.00       │
├────────────────────────┼───────┼────────────┤
│ between 50MB and 100MB │ 0     │ 0.00       │
├────────────────────────┼───────┼────────────┤
│ above 100MB            │ 0     │ 0.00       │
└────────────────────────┴───────┴────────────┘
```


michaelklishin:
Given how many commands `rabbitmqctl` already has, I think we should only expose this new command via `rabbitmq-diagnostics`.

Issue #12597, opened 2024-10-28
With Khepri, transient exclusive queues are not deleted  in case of a partition


luos:
### Describe the bug

Hi,

In a Khepri cluster if a node becomes partitioned, classic queues which get deleted due to the partition - do not get cancelled. 

I am not sure if this is the intended behaviour - but in this case basically in a cluster no other than qq and streams are usable. 

### Reproduction steps

1. 3 node cluster, enable khepri, RabbitMQ 4.0.2
2. Create a transient exclusive queue on each node and attach a consumer
3. Create a partition on one of the nodes (full partition, but expect the same happens in case of a partial partition)
4. The node becomes partitioned
4. a. Queue gets deleted on majority partition
4. b. Queue stays `running` on minority partition 
5. Remove the partition
6. Minority node syncs with khepri, queue disappears
7. No notification to the consumer on minority node that the queue no longer exists 

### Expected behavior

Any or all of the following: :-) 

a. Client gets notified about the queue deletion
b. Queue is not deleted as the consumer was not disconnected
c. Something akin to pause_minority
d. Clarification of the intended use of transient exclusive queues in a cluster 
e. Specified behaviour in case of partial partition as well

### Additional context

- The queue process is still running, even though the queue is deleted from khepri. (which is understandable, I am just noting it here).

dumbbell:
@luos: The fix for this issue is ready and available in rabbitmq/rabbitmq-server#14573. Could you please give it a try and tell me if it is fixed for you too?

(sorry for pinging you twice here and in the question about auto-delete queues, this is to make sure the communication is consistent and I don’t forget :-)

PR #14573, opened 2025-09-19
Keep exclusive/auto-delete queues with Khepri + network partition

### LINKED ISSUE #12949: Exclusive queues can be deleted without the consumers being notified
### Describe the bug

Given a 3-node cluster with Khepri enabled, if a node is partitioned (certainly happens with the leader being partitioned, I'm not sure if it can if the follower is partitioned), exclusive queues can be deleted without their corresponding connections being notified nor terminated (so from the client's perspective, it just looks like there are no messages).

Originally reported in:
https://github.com/rabbitmq/rabbitmq-server/discussions/12829#discussion-7563564

### Reproduction steps

1. 3-node cluster with Khepri enabled, a publisher and consumer(s) using exclusive queues on a node with the Khepri leader
2. The node with the Khepri leader, all connections and queues is partitioned
3. Clients don't notice anything, since all connections and queues are available on the local node
4. A minute later (net tick timeout), the nodes realize they can't talk to one of the nodes. They all trigger the deletion of exclusive queues on the partitioned node. However, the node with the queues doesn't delete them, [because the queue processes are alive](https://github.com/rabbitmq/rabbitmq-server/blob/96ebce8c007bb8811a466bd59ab1aaee5cd5b64c/deps/rabbit/src/rabbit_amqqueue.erl#L1957). The other two nodes delete the queues in Khepri, but it has no practical effect, because they can't communicate with the node where the queues are running.
5. Once the partition heals, the partitioned node realized it should be a Khepri follower now, synchronizes the metadata changes (including the deletions), deletes the queues but doesn't notify the consumers

### Expected behavior

We discussed removing the special case for exclusive queues altogether. Currently, they are considered transient internally, but with Khepri, there's really no concept of a transient declaration, since all declarations are persisted in Khperi's log. Getting rid of the whole "transient" concept once and for all should simplify the code, documentation and in general make RabbitMQ easier to use and understand.

Either way, the queues should not be deleted or the consumers should be notified.

### Additional context

All logs from a reproduction:
[rmq-server-0.log](https://github.com/user-attachments/files/18151182/rmq-server-0.log)
[rmq-server-1.log](https://github.com/user-attachments/files/18151183/rmq-server-1.log)
[rmq-server-2.log](https://github.com/user-attachments/files/18151184/rmq-server-2.log)

* a 3-node cluster deployed to Kubernetes
* rmq-server-2 is the Khepri leader
* `perf-test -H amqp://rmq-server-2.rmq-nodes -E -t fanout -e amq.fanout -y 10 -r 1` - all connections and exclusive queues are on rmq-server-2; as expected, perf-test publishes 1 message per second and consumes 10 messages per second (fanout to 10 exclusive queues)
* Mon Dec 16 12:41:20 CET 2024: rmq-server-2 is partitioned 
* Mon Dec 16 12:42:24 CET 2024: rmq-server-2 is no longer partitioned 
* perf-test keeps running without any errors, still publishes 1 msg/s but doesn't consume any messages anymore

The logs contain some [additional debug logs](https://github.com/rabbitmq/rabbitmq-server/tree/exclusive-queue-debugging) which show that all nodes attempted to deleted the exclusive queue on a down node, but:
* server-2 didn't, because they were alive
* server-1 deleted the queues (`10 transient queues from node 'rabbit@rmq-server-2.rmq-nodes.chaos-tests' deleted in 0.066262s`) and sent notificaitons but this was not effective, since server-2 didn't get the memo
* at 12:42:26 server-2 rejoins the cluster, becomes the follower and actually deletes the queues, but doesn't notify consumers

dumbbell:
Keep exclusive/auto-delete queues with Khepri + network partition

## Why

With Mnesia, when the network partition strategy is set to `pause_minority`, nodes on the "minority side" are stopped.

Thus, the exclusive queues that were hosted by nodes on that minority side are lost:
* Consumers connected on these nodes are disconnected because the nodes are stopped.
* Queue records on the majority side are deleted from the metadata store.

This was ok with Mnesia and how this network partition handling strategy is implemented. However, it does not work with Khepri because the nodes on the "minority side" continue to run and serve clients. Therefore the cluster ends up in a weird situation:
1. The "majority side" deleted the queue records.
2. When the network partition is solved, the "minority side" gets the record deletion, but the queue processes continue to run.

This was similar for auto-delete queues.

## How

With Khepri, we stop to delete transient queue records in general, just because there is a node going down. Thanks to this, an exclusive or an auto-delete queue and its consumer(s) are not affected by a network partition: they continue to work.

However, if a node is really lost, we need to clean up dead queue records. This was already done for durable queues with both Mnesia and Khepri. But with Khepri, transient queue records persist in the store like durable queue records (unlike with Mnesia).

That's why this commit changes the clean-up function, `rabbit_amqqueue:forget_all_durable/1` into `rabbit_amqqueue:forget_all/1` which deletes all queue records of queues that were hosted on the given node, regardless if they are transient or durable.

In addition to this, the queue process will spawn a temporary process who will try to delete the underlying record indefinitely if no other processes are waiting for a reply from the queue process. That's the case for queues that are deleted because of an internal event (like the exclusive/auto-delete conditions). The queue process will exit, which will notify connections that the queue is gone.

Thanks to this, the temporary process will do its best to delete the record in case of a network partition, whether the consumers go away during or after that partition. That said, the node monitor drives some failsafe code that cleans up record if the queue process was killed before it could delete its own record.

Fixes #12949, #12597, #14527.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue_process.erl
+++ b/deps/rabbit/src/rabbit_amqqueue_process.erl
@@ -352,19 +352,64 @@ terminate_delete(EmitStats, Reason0, ReplyTo,
                                                 fun() -> emit_stats(State) end);
            true      -> ok
         end,
-        %% This try-catch block transforms throws to errors since throws are not
-        %% logged. When mnesia is removed this `try` can be removed: Khepri
-        %% returns errors as error tuples instead.
-        Reply = try rabbit_amqqueue:internal_delete(Q, ActingUser, Reason0) of
-                    ok ->
-                        {ok, Len};
-                    {error, _} = Err ->
-                        Err
-                catch
-                    {error, ReasonE} -> error(ReasonE)
-                end,
-        send_reply(ReplyTo, Reply),
+        case ReplyTo of
+            _ when ReplyTo =/= none ->
+                Reply = case delete_queue_record(Q, ActingUser, Reason0) of
+                            ok ->
+                                {ok, Len};
+                            {error, _} = Err ->
+                                Err
+                        end,
+                send_reply(ReplyTo, Reply);
+            none ->
+                %% No processes are waiting for this queue process to exit. We
+                %% can handle the deletion of the queue record differently: if
+                %% the deletion times out, we retry indefinitely.
+                %%
+                %% For instance, this allows an auto-delete queue process to
+                %% wait and retry until a network partition is resolved (or
+                %% this node stops of course). This reduces the risk of a
+                %% "leak" of a queue record in the metadata store.
+                %%
+                %% If for whatever reason the queue record is still leaked
+                %% (this process could not delete it before it was killed), the
+                %% "leaked" queue record will be cleaned up when the partition
+                %% is solved (or this node is removed from the cluster).
+                %% Indeed, when the partition is solved, all nodes are notified
+                %% with the `node_up' message from `rabbit_node_monitor'. This
+                %% calls `rabbit_amqqueue:on_node_up/1' which will delete any
+                %% transient queues.
+                %%
+                %% This infinite delete attempts loop is executed in a
+                %% separate process to let this queue process exits. This way,
+                %% connections will be notified that the queue process is
+                %% gone.
+                worker_pool:submit_async(
+                  fun() ->
+                          _ = infinite_internal_delete(Q, ActingUser, Reason0)
+                  end),
+                ok
+        end,
         BQS1
+    end.
+
+infinite_internal_delete(Q, ActingUser, Reason) ->
+    case delete_queue_record(Q, ActingUser, Reason) of
+        {error, timeout} ->
+            _ = rabbit_khepri:fence(infinity),
+            infinite_internal_delete(Q, ActingUser, Reason);
+        Ret ->
+            Ret
+    end.
+
+delete_queue_record(Q, ActingUser, Reason) ->
+    %% This try-catch block transforms throws to errors since throws are not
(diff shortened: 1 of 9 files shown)


dumbbell:
There are two test flakes in the new tests that I’m looking at. Otherwise, the patch looks ready for further testing.

deps/rabbit/src/rabbit_amqqueue_process.erl
```diff
@@ -352,21 +352,65 @@ terminate_delete(EmitStats, Reason0, ReplyTo,
...
+                  fun() ->
+                          _ = infinite_internal_delete(Q, ActingUser, Reason0)
+                  end),
+                ok
+        end,
         BQS1
     end.
 
+infinite_internal_delete(Q, ActingUser, Reason) ->
+    case delete_queue_record(Q, ActingUser, Reason) of
```
kjnilsson, line 397: If the node is partitioned with a khepri leader on it this code could grow the khepri log infinitely.

dumbbell: I see what you mean. Then, I need to explore @lhoguin’s idea of waiting for the `node_up` message.

dumbbell: In fact, using a Khepri fence after the first delete attempt should be enough: the call waits for all updates to be applied locally. I just pushed that change.

mkuratczyk (approved):
Everything seems to work as expected. Thanks!

PR #14589, opened 2025-09-23
Quorum queues: fix resend issues after network partition

kjnilsson:
A quorum queue client that send a message during a network partition
that later caused a distribution disconnection would in some cases
never resend the lost message, even if kept in the pending buffer.
Subsequent sends would be accepted by the state machine but would
never be enqueued as there would be a missing sequence.

In the case of publishers that use pre-settled sends the pending
messages would have also been incorrectly removed from the
pending map.

To fix we removed timer resend aapproach and instead have the leader
send leader_change messages on node up to prompt any queue clients
to resend their pending buffer.


michaelklishin (approved):
The Selenium suite failures are unrelated and affect (or affected, I've merged a Selenium-related PR earlier today) other branches.

PR #14609, opened 2025-09-25
Introduce "plugins_data_dir" which is preserved during khepri migration

### LINKED ISSUE #11304: Allow Khepri post-migration step to not delete certain files or dirs

As the last step of migrating to Khepri (ie enabling the `khepri_db` feature flag) all Mnesia related files are deleted (https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_khepri.erl#L1469-L1480) However the heuristic to find Mnesia related files is too eager, it deletes everything in the data dir recursively except for those which are known to be not Mnesia (ie msg_stores, coordination, quorum, stream and rabbitmq_metadata dirs are not deleted).

This is an issue for community plugins which want to migrate away from Mnesia into a new subdir. (They might not even use Khepri, or use Khepri but a different store id.) As an example (and the motivation of this ticket) the delayed message store is planned to store its leveldb data in a `dmx` subdir (https://github.com/rabbitmq/rabbitmq-delayed-message-exchange/pull/253/files#diff-64860364dca47a85f9e9fc932a30c53a07673f78e50ebc2a622e64bf78e91ce2R41-R42). After the delayed message plugin nicely migrated its messages to a new storage backend, the post-migration step wipes the newly created `dmx` dir. There is no clean callback that is called after this step that the plugin can implement to eg move the migrated data from a tmp dir to its final location.

I don't have a strong idea for a solution. One option is that plugins can register a list of additional files and dirs (maybe an optional callback for the migration behaviour) that are not deleted in the post-migration step.

Alternative could be an optional migration callback that is executed after `rabbit_khepri:khepri_db_migration_post_enable/0` that the plugins can implement to move the migrated data from a tmp dir to its final location.

A workaround could be that the community plugin stores its data under one of the dirs which are not deleted, but this is a hack. The plugin data does not belong there.

### Additional context

_No response_

gomoripeti:
## Proposed Changes

This is mainly for community plugins which want to migrate away from Mnesia into a new subdir. (They might not even use Khepri, or use Khepri but a different store id.)

Fixes #11304

If accepted, I shall update docs for community plugin developers somewhere on https://www.rabbitmq.com/plugin-development

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments


michaelklishin (changes requested):
The idea is reasonable. The name `plugins_data_dir` suggests that it is a directory used by RabbitMQ but it's not the case.

We need a better name. `user_plugins_dir` would make it immediately obvious that the directly is NOT maintained by RabbitMQ.

gomoripeti:
thanks for looking at it and the name suggestions. Let me note that "plugins_dir" is the dir where plugin code is, while this directory is for data (from user code). what about `user_data_dir`? Or the longer ones `user_plugins_data_dir`, `additional_plugins_data_dir` or `community_plugins_data_dir`?

michaelklishin:
@gomoripeti yup, after re-reading #11304 I have figured the difference and the intent. https://github.com/rabbitmq/rabbitmq-server/pull/14622 uses `rabbit.user_provided_plugins_data_dir`, so the "plugins data" part is preserved.

This contribution will ship in `4.2.0`. Thank you.

PR #14613, opened 2025-09-25
Continuous Membership Reconciliation for streams

SimonUnge:
## Proposed Changes
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
This PR implements automatic membership reconciliation for RabbitMQ streams, a mirror of the existing quorum queue membership reconciliation feature, as discussed [in this thread](https://github.com/rabbitmq/rabbitmq-server/discussions/14246 ) created by @tschuettig                                                                                                                                                                                                                                                                                                           
                           
In its current form, it behaves verbatim like the feature for Quorum Queues.

In the discussion, @kjnilsson brought up that "member changes for streams isn't as safe to do as for e.g. quorum queues which means that we're probably not quite ready for any such automation." but I wanted us to have a starting point to work from. It is an opt in feature and uses the same steps you'd manually do using the CLI stream commands. It has the same conditions for growing (and shrinking) as quorum queues - perhaps this in an area where we could add more checks, e.g that all replicas are healthy etc?

If this is the route to go, we could create shared code for QQ and Stream auto reconciliation, but I wanted to keep it separate in this PR to be able to clearly discuss the feature.
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
### Core Implementation                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
**rabbit_stream_periodic_membership_reconciliation.erl** - Main gen_server that performs periodic membership reconciliation for streams                                                                                                                                                                                                                                                                                                                                                              
**rabbit_stream_event_subscriber.erl** - Event subscriber that triggers reconciliation on cluster events (node up/down, policy changes)                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
### Configuration                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
Added stream-specific configuration schema entries in rabbit.schema:                                                                                                                                                                                                                                                                                                                                                                                                                                 
```
stream.continuous_membership_reconciliation.enabled                                                                                                                                                                                                                                                                                                                                                                                                                                                
stream.continuous_membership_reconciliation.auto_remove                                                                                                                                                                                                                                                                                                                                                                                                                                            
stream.continuous_membership_reconciliation.interval                                                                                                                                                                                                                                                                                                                                                                                                                                               
stream.continuous_membership_reconciliation.trigger_interval                                                                                                                                                                                                                                                                                                                                                                                                                                       
stream.continuous_membership_reconciliation.target_group_size    
```                                                                                                                                                                                                                                                                                                                                                                                                                                  
## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -2689,6 +2689,29 @@ end}.
 
 
 %%
+%% Stream membership reconciliation
+%%
+
+{mapping, "stream.continuous_membership_reconciliation.enabled", "rabbit.stream_membership_reconciliation_enabled", [
+    {datatype, {enum, [true, false]}}]}.
+
+{mapping, "stream.continuous_membership_reconciliation.auto_remove", "rabbit.stream_membership_reconciliation_auto_remove", [
+    {datatype, {enum, [true, false]}}]}.
+
+{mapping, "stream.continuous_membership_reconciliation.interval", "rabbit.stream_membership_reconciliation_interval", [
+    {datatype, integer}, {validators, ["non_negative_integer"]}
+]}.
+
+{mapping, "stream.continuous_membership_reconciliation.trigger_interval", "rabbit.stream_membership_reconciliation_trigger_interval", [
+    {datatype, integer}, {validators, ["non_negative_integer"]}
+]}.
+
+{mapping, "stream.continuous_membership_reconciliation.target_group_size", "rabbit.stream_membership_reconciliation_target_group_size", [
+    {datatype, integer}, {validators, ["non_negative_integer"]}
+]}.
+
+
+%%
 %% Runtime parameters
 %%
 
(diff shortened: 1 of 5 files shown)


LoisSotoLopez:
When considering the shrinking of members the proposed code takes into account the number of running nodes. I wonder if it would be interesting to also take into account the target group size.

Would it be interesting for customers to be able to reduce from e.g. 7 to 5 the target group size to reduce the number of members of a stream on a 7 nodes cluster?

SimonUnge:
> When considering the shrinking of members the proposed code takes into account the number of running nodes. I wonder if it would be interesting to also take into account the target group size.

Thats a good question. I personally think it is too scary for automation to take that decision, which node(s) should it remove, how would that decision be made? My guts says this is something we should leave to a human to take action on, unless we can come up with a safe and predictable choice. 

> Would it be interesting for customers to be able to reduce from e.g. 7 to 5 the target group size to reduce the number of members of a stream on a 7 nodes cluster?

PR #14614, opened 2025-09-26
Ensures pending counter in rabbit_shovel_status is always an integer

markusgust:
## Proposed Changes

We observed that when a shovel connection for AMQP091 run into blocked state, pending messages stack up in a list of messages, which leads to the response of `rabbitmq_shovel_status:status()` can get very large and can cause memory leaks with many shovels (or one shovel using `no-ack`). It also breaks the management UI for shovel status.

steps to reproduce:
- create a shovel from cluster-a to cluster-b. 
- simulate memory watermark on b.
- shovel a message, navigate to shovel status. should now break with 500 internal error.

This PR ensures that the number of pending messages is an integer for all shovel protocols.

I have tried verifying the state of shovels where connection gets blocked for AMQP1.0 as well, but I haven't been able to as I am getting a different error, so I'll continue to investigate that. 

## Types of Changes

- [X] Bug fix (non-breaking change which fixes issue )
- [ ] Build system and/or CI

## Checklist

- [x ] I have added tests that prove my fix is effective or that my feature works

## Further Comments


michaelklishin:
The core team has no objection to this `shovel` interface (Erlang behavior) change.

If we consider this a bug fix, it can go into `4.2.0` after the code freeze/RC1 release next week, too.

deps/rabbitmq_shovel/src/rabbit_local_shovel.erl
```diff
@@ -437,6 +438,11 @@ add_routing(Msg0, Dest) ->
 status(_) ->
     running.
 
+pending_count(#{source := #{current := #{unacked_message_q := UAMQ}}}) ->
```
dcorbacho: Pending messages in AMQP0.9.1 are counted in the destination side, thus in local shovels they should be dest -> unconfirmed

deps/rabbitmq_shovel/src/rabbit_amqp10_shovel.erl
```diff
@@ -317,6 +318,10 @@ status(_) ->
     %% Destination not yet connected
     ignore.
 
+pending_count(#{dest := Dest}) ->
```
dcorbacho, line 321: Pending messages in AMQP0.9.1 are counted in the destination side, thus in AMQP1.0 shovels they should be dest -> unacked

gomoripeti: my understanding is that in case of AMQP 1.0 the dest -> unacked field is only used in on-confirm ack-mode. But there can be pending messages in case of other ack-modes too when there is no more link credit to send the message to the dest. In this case the whole message is buffered in the shovel process. My understanding is that the pending metric only counts the number of buffered ie unsent messages. Unacked additionally counts messages that were sent but not yet acked by the dest.

There is also the metric `remaining_unacked`, but I'm not sure it is counted correctly in case of on-confirm. It is decremented when the message is sent (the same way as for other ack-modes) and not when the message is accepted/rejected by the dest.

deps/rabbitmq_shovel/src/rabbit_local_shovel.erl
gomoripeti: if I see it correctly there is no flow control between a local shovel and a queue so there can be no pending aka buffered messages for local shovels  (according to my above definition of pending).

dest -> unconfirmed is only non-zero in case of on-confirm ack-mode. source -> unacked_message_q can be non-empty for on-confirm and on-publish.

gomoripeti (commented):
to sum up my comments I think
- if the pending metric should be the count of messages buffered in the shovel this PR is correct for AMQP 0.9.1 and 1.0 but should always be zero for local shovels
- if the pending metric should be the count of messages not yet acked to the source queue this PR is correct for local shovels but needs to be updated for AMQP 0.9.1 and 1.0

I think exposing the count of buffered messages is more relevant as those can take up considerable memory and less visible in source/dest connection and queue metrics.

dcorbacho:
Thanks for the clarification @gomoripeti
If pending messages means the number of messages stored (whole) in the shovel, then the changes for AMQP1.0 and AMQP0.9.1 are correct. Local shovels then must always return 0, as the uamq just stores messages ids and it is not the equivalent to pending messages in other protocols.

Regarding to the `remaining_unacked` counter, it is correct for all three protocols. It's always decremented when a message is forwarded (regardless of it being later acked/nacked) and it is used to decide when to shutdown an auto-delete shovel (i.e. uses the 'delete-after' feature).

markusgust:
thank you @dcorbacho for your help and feedback, based on this I added some changes to what the local shovel returns.

PR #14622, opened 2025-09-26
Introduce a configurable directory for 3rd party plugins that's not deleted during the Mnesia => Khepri migration

michaelklishin:
This is https://github.com/rabbitmq/rabbitmq-server/pull/14609/files by @gomoripeti with a different name for the new setting (intentionally kept specific to `advanced.config`, at least for now).

During the Mnesia => Khepri migration, all directories [under the node's data directory] except for a set of whitelisted ones will be deleted. This allows the users who install 3rd party plugins (e.g. community plugins) and keep them under the node's data directory to avoid plugin deletion.

Fixes https://github.com/rabbitmq/rabbitmq-server/issues/11304.


gomoripeti:
thank you for taking care of this

PR #14624, opened 2025-09-26
Allow configuring whether queue types are enabled

the-mikedavis:
This change adds config options for controlling the `is_enabled/0` queue type callback for classic queues, quorum queues and stream queues. This is useful for operators to be able to control whether users can declare these queues - i.e. to prevent users from declaring queue types which cannot be supported operationally.

Setting

    queue_types.stream.enabled = false

In a config file for example prevents stream queues from being declared.

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -2629,6 +2629,18 @@ end}.
     {datatype, {enum, [true, false]}}
 ]}.
 
+%% Enable queue types
+
+{mapping, "queue_types.classic.enabled", "rabbit.classic_queues_enabled", [
+    {datatype, {enum, [true, false]}}
+]}.
+{mapping, "queue_types.stream.enabled", "rabbit.stream_queues_enabled", [
+    {datatype, {enum, [true, false]}}
+]}.
+{mapping, "queue_types.quorum.enabled", "rabbit.quorum_queues_enabled", [
+    {datatype, {enum, [true, false]}}
+]}.
+
 %%
 %% Backing queue version
 %%
--- a/deps/rabbit/src/rabbit_classic_queue.erl
+++ b/deps/rabbit/src/rabbit_classic_queue.erl
@@ -126,7 +126,7 @@ validate_policy(Args) ->
     end.
 
 -spec is_enabled() -> boolean().
-is_enabled() -> true.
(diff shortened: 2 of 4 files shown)


the-mikedavis:
I didn't add a test case for this since it's a very straightforward change. This can be tested manually by creating a config file `config.conf` with `queue_types.stream.enabled = false`, running the broker (`make run-broker RABBITMQ_CONFIG_FILE=config.conf`) and trying to declare a stream queue through either the management UI or with `perf-test --stream-queue`.

I also didn't add a config option for the MQTT QoS0 queue type since I think that queue type is more of an implementation detail and not worth controlling other than with the feature flag.

mkuratczyk:
This PR prevents streams from being declared in some, but not all, ways. If you just run stream-perf-test, it declares a stream without problems, despite `queue_types.stream.enabled = false`.

Moreover, trying to disable classic queues seems to prevent a node from booting at all. Just trying `make run-broker` with `queue_types.classic.enabled = false` leads to:
```
error:{badmatch,{error,queue_type_feature_flag_is_not_enabled}}

    rabbit:insert_default_data/0, line 1236
    rabbit_boot_steps:-run_step/2-lc$^0/1-0-/2, line 53
    rabbit_boot_steps:run_step/2, line 60
    rabbit_boot_steps:-run_boot_steps/1-lc$^0/1-0-/1, line 24
    rabbit_boot_steps:run_boot_steps/1, line 25
    rabbit:start/2, line 1009
    application_master:start_it_old/4, line 299
```

the-mikedavis:
For the streams protocol it looks like we'll need to add a check of `rabbit_queue_type:is_enabled/1` in `rabbit_stream_manager:do_create_stream/4` so it behaves more like `rabbit_amqqueue:declare/7`.

For the failure to boot it looks like this check fails when adding the default vhost: https://github.com/rabbitmq/rabbitmq-server/blob/a09383d1e5707a7d969c5b7dbbf036715b563f6f/deps/rabbit/src/rabbit_vhost.erl#L212-L219

So we can't add vhosts where the default queue type is not enabled. I think it might be more reasonable to warn in that situation since you might always explicitly declare your queue types - IMO `is_enabled/1` should only stop attempts to declare a queue and not block vhost creation.

I will send some follow-ups. Thanks @mkuratczyk!

PR #14638, opened 2025-09-29
Minor: Fix Mgmt UI local shovel delete-after

gomoripeti:
## Proposed Changes

Context: Management UI > Shovel Management Tab > "Add a new shovel" section.

There was no input field to set the number for delete-after "After num messages" parameter for local shovels as opposed to AMQP 1.0 shovels.

The help text was also missing for local shovel auto-delete. (The `?` showed an empty popup when Local protocol is selected for Source)

The solution is just a copy past from AMQP 1.0

**EDIT**: now the bellow works

Setting delete-after to a number when creating a local shovel still **doesn't work** as the value is sent in the PUT json as a string instead of a number. Doing the same with an AMQP 1.0 works fine sending a json number. Where the conversion happens on the frontend is beyond my understanding of the Mgmt javascript. I would need some help to address this part.

```
{
  "component": "shovel",
  "vhost": "/",
  "name": "s2",
  "value": {
    "src-uri": "amqp://",
    "dest-uri": "amqp://",
    "ack-mode": "on-confirm",
    "src-protocol": "local",
    "dest-protocol": "local",
    "src-queue": "cq1",
    "src-delete-after": "3", <--- string instead of number
    "dest-queue": "cq2",
    "dest-add-forward-headers": false
  }
}
```

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

### DIFF
--- a/deps/rabbitmq_shovel_management/priv/www/js/shovel.js
+++ b/deps/rabbitmq_shovel_management/priv/www/js/shovel.js
@@ -75,7 +75,8 @@ dispatcher_add(function(sammy) {
             //copy the correct delete-after value
             if (this.params['src-delete-after-selector'] == 'never') {
                 this.params['src-delete-after'] = 'never';
-            } else if (this.params['src-delete-after-selector'] == 'number') {
+            } else if (this.params['src-delete-after-selector'] == 'number' ||
+                       this.params['src-delete-after-selector'] == 'local-number') {
                 num_keys.push('src-delete-after');
             }
 


LoisSotoLopez:
As far as I understood the js code - I'm definitely not an expert - when  the user select an option for a value that should be considered numeric, the name of the field where you input that number has to get pushed into the array of numeric fields.

gomoripeti:
thank you Lois, that was it

(iirc the option value must be unique because otherwise some selector mixes it up with the amqp10 number field in https://github.com/rabbitmq/rabbitmq-server/pull/14638/files#diff-ca4603ff511325d1b70af1db91e4e29d6e7b594cd743ed820377ba2b95aaade1R142)

PR #14641, opened 2025-09-29
Optionally return HTTP auth backend denial reason to client

ansd:
This PR contains two commits.

# Commit 1:

 ## What?
Support logging (at INFO level) the reason provided by the HTTP auth backend why authentication or autorisation was denied.

 ## Why?
* Security and compliance often require detailed logs of why access was denied
* Operational debugging: Operators see immediately in RabbitMQ logs why authentication/authorization failed without checking the HTTP backend logs

 ## How?
The HTTP body returned by the HTTP auth server is allowed to be `deny <Reason>` where `<Reason>` is any text that will be logged by RabbitMQ at INFO level.

# Commit 2:

 ## What?

If the new config setting `authorization_failure_disclosure`
for an authz backend is set to `true`, (`false` by default), RabbitMQ
will return the reason why access was denied to the client.
For now, only the HTTP auth backend supports this new config setting.

 ## Why?

This helps debugging and troubleshooting directly in the client.
Some users might not have access to the RabbitMQ logs, for other
users it's cumbersome to correlate authz denial in the client with
logs on the broker.

For example, some customers would like to pass the reason why
authorization was denied from their custom HTTP auth backend via
RabbitMQ back to the client.

 ## How?

Authz backends can now return `{false, Reason}` as an alternative to
just `false` if access is denied.

For security reasons, the additional denial reason by the authz backend
will be returned to the client only if the operator opted in by setting
`authorization_failure_disclosure` to `true`.

Note that `authorization_failure_disclosure` applies only to
already authenticated clients when they try to access resources (e.g. vhosts,
exchanges, queues, topics). For security reasons, no detailed denial reason is
returned to the client if **authentication** fails.

Also note that `authorization_failure_disclosure` is set separately per
auth backend instead of being set globally for all auth backends. This
more fine granular configurability helps for use cases where the broker
should reveal the authz denial reason for only a specific auth backend.

### DIFF
--- a/deps/rabbitmq_auth_backend_http/src/rabbit_auth_backend_http.erl
+++ b/deps/rabbitmq_auth_backend_http/src/rabbit_auth_backend_http.erl
@@ -25,6 +25,8 @@
 
 -define(SUCCESSFUL_RESPONSE_CODES, [200, 201]).
 
+-define(APP, rabbitmq_auth_backend_http).
+
 %%--------------------------------------------------------------------
 
 description() ->
@@ -37,7 +39,7 @@ user_login_authentication(Username, AuthProps) ->
     Path = p(user_path),
     Query = q([{username, Username}] ++ extract_other_credentials(AuthProps)),
     case http_req(Path, Query) of
-        {error, _} = Err  ->
+        {error, _} = Err ->
             Err;
         "deny " ++ Reason ->
             ?LOG_INFO("HTTP authentication denied for user '~ts': ~ts",
@@ -53,7 +55,9 @@ user_login_authentication(Username, AuthProps) ->
                     {ok, #auth_user{
                             username = Username,
                             tags = Tags,
-                            impl = fun() -> proplists:delete(username, AuthProps) end}}
+                            impl = fun() -> proplists:delete(username, AuthProps) end}};
+                Other ->
+                    {error, {bad_response, Other}}
             end
     end.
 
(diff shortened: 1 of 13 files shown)


deps/rabbitmq_auth_backend_http/src/rabbit_auth_backend_http.erl
```diff
@@ -34,16 +36,29 @@ description() ->
...
+    Query = q([{username, Username}] ++ extract_other_credentials(AuthProps)),
+    case http_req(Path, Query) of
+        {error, _} = Err ->
+            Err;
+        "deny " ++ Reason ->
+            ?LOG_INFO("HTTP authentication denied for user '~ts': ~ts",
+                      [Username, Reason]),
+            {refused, "Denied by the backing HTTP service", []};
+        Body ->
+            case string:lowercase(Body) of
```
dumbbell, line 49: Isn’t converting the string to lowercase a breaking change for tags?

ansd: There should be no breaking change in this PR.
I actually convert here to lowercase just for backward compat since that's done on `main` branch too in https://github.com/rabbitmq/rabbitmq-server/blob/5a3b62b3417b73b0bf1adad39440579785f07221/deps/rabbitmq_auth_backend_http/src/rabbit_auth_backend_http.erl#L243

ansd: If an HTTP auth backend returns `deny Exchange Declarations Disallowed Globally`, then the AMQP client should receive exactly this reason `Exchange Declarations Disallowed Globally`, i.e. capital letters in the custom reason should be preserved.

dumbbell: I see, thank you!

PR #14660, opened 2025-10-01
amqp_client: Emit 'connection.blocked' in direct connections

### LINKED ISSUE #14657: Block direct shovel publishes during alarms

Currently direct connections (`amqp_direct_connection`) do not notice when publishers are blocked due to alarms like disk or memory usage. For an example:

1. `mkdir /tmp/small-disk`
1. `sudo mount -t tmpfs -o size=3000M small-disk /tmp/small-disk`
1. `TEST_TMPDIR=/tmp/small-disk make run-broker`
1. `sbin/rabbitmqctl set_disk_free_limit 1000M`
1. Publish until the disk alarm is triggered (`perf-test -qq -u qq -x 1 -y 0`)
1. `sbin/rabbitmqctl set_parameter shovel my-shovel '{"src-protocol": "amqp091", "src-uri": "amqp://", "src-queue": "qq", "dest-protocol": "amqp091", "dest-uri": "amqp://", "dest-queue": "qq-copied"}'`

The shovel can successfully move the messages from `qq` to `qq-copied` despite publishers being blocked. If the disk free limit is set low enough, moving the messages in a queue like this can fully exhaust disk space. It also probably works to shovel in messages from another instance for the same effect. As long as the destination connection is direct, publishing will not be blocked by alarms.

Direct connections can register themselves with `rabbit_alarm` and emit a `#'connection.blocked'{}` when an alarm fires. `rabbit_amqp091_shovel` then avoids sending messages to the destination (since #6224).

_No response_

### Additional context

_No response_

the-mikedavis:
Fixes #14657

Previously direct 0-9-1 connections did not notice when memory or disk alarms were set. This could allow an 0-9-1 shovel where the destination is a direct connection to completely overload a broker which is already in alarm. With this change, direct connections register the connection process with `rabbit_alarm` and emit `connection.blocked` and
`connection.unblocked` to the blocked handler if one is registered. `rabbit_amqp091_shovel` already respects the `connection.blocked`, so the destination will not receive any messages.

### DIFF
--- a/deps/amqp_client/src/amqp_gen_connection.erl
+++ b/deps/amqp_client/src/amqp_gen_connection.erl
@@ -31,6 +31,7 @@
                 server_properties,
                 %% connection.block, connection.unblock handler
                 block_handler,
+                blocked_by = sets:new([{version, 2}]),
                 closing = false %% #closing{} | false
                }).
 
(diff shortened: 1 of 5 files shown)


ansd:
Just in case it helps, the following shows how to easily trigger memory and/or disk alarms in a test case:
https://github.com/rabbitmq/rabbitmq-server/blob/2303570d9082f32e42098ec82ce0ad866c40b14c/deps/rabbit/test/amqp_client_SUITE.erl#L3200-L3234

the-mikedavis:
Thanks for the pointers @ansd & @gomoripeti! I moved the existing tests for alarms and 0-9-1 shovels to a new suite and added a group there that uses a direct connection

michaelklishin (approved):
Ugh, this is a tough behavior to test :(

PR #14665, opened 2025-10-02
clustering_recovery_SUITE: Skip tests that require RabbitMQ 4.2.0 in mixed-version testing

dumbbell:
## Why

The `*_queue_after_partition_recovery_1` testcases rely on the fact that the queue process retries to delete its queue record in its terminate function.

RabbitMQ 4.1.x and before don't have that. Thus, depending on the timing of the election of the Khepri leader after a network partition, the test might fail.

This is a follow-up commit to #14573.


dumbbell:
This will be backported to 4.2.x as part of #14654.

PR #14672, opened 2025-10-03
Use tick-related timeout to repair leader record

acogoluegnes:
A quorum queue tries to repair its record in a tick handler. This can happen during a network partition and the metadata store may itself be unavailable, making the update likely to time out.

The default metadata store timeout is usually higher than the tick interval, so the tick handler may be stuck during several ticks. The record takes some time to be updated (timeout + tick interval, 30 + 5 seconds by default), significantly longer than it takes the metadata store to trigger an election and recover.

Client applications may rely on the quorum queue topology to connect to an appropriate node, so making the system reflect the actual topology faster is important to them.

This commit makes the record update operations use a timeout 1-second lower than the tick interval. The tick handler process should finish earlier in case of metadata datastore unavailability and it should not take more than a couple of ticks once the datastore is available to update the record.


deps/rabbit/src/rabbit_db_queue.erl
```diff
@@ -637,9 +638,22 @@ get_many_in_ets(Table, Names) ->
 %% @private
 
 update(QName, Fun) ->
+    update(QName, Fun, #{}).
+
+-spec update(QName, UpdateFun, Options) -> Ret when
+      QName :: rabbit_amqqueue:name(),
+      Queue :: amqqueue:amqqueue(),
+      UpdateFun :: fun((Queue) -> Queue),
+      Options :: khepri:query_options(),
```
dumbbell: The same `Options` variable is passed to both a query and a put operation. The put operation doesn’t accept query options, so this is incorrect.

Because you are interested in the `timeout` option, you could set the spec to `Options :: #{timeout => timeout()}` instead.

deps/rabbit/src/rabbit_db_queue.erl
```diff
@@ -637,9 +638,22 @@ get_many_in_ets(Table, Names) ->
 %% @private
 
 update(QName, Fun) ->
+    update(QName, Fun, #{}).
+
+-spec update(QName, UpdateFun, Options) -> Ret when
+      QName :: rabbit_amqqueue:name(),
+      Queue :: amqqueue:amqqueue(),
+      UpdateFun :: fun((Queue) -> Queue),
```
dumbbell: Even if Dialyzer doesn’t pay attention to that yet, this `fun()` likely doesn’t return its argument unchanged.

The correct spec would be e.g.:
```erlang
UpdateFun :: fun((Queue) -> NewQueue),
NewQueue :: amqqueue:amqqueue(),
```

kjnilsson: why would you ever want to signal return of an unchanged element in the type?, just don't return it if it isn't to change

dumbbell: Because it was written somewhere that re-using the type alias (`Queue`) does not only means that it is of the same type, it also means it is the same value. This kind of verification is not implemented by Dialyzer. But the spec of the type spec is built like this.

I discovered this only a few years ago and I have several mistakes like this in older code because I also assumed that reusing the type name just meant the same type was repeated, regardless of the actual value. I don’t remember where I read that.

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -303,7 +303,14 @@ do_internal_declare(Q0, false) ->
             'not_found' | amqqueue:amqqueue().
 
 update(Name, Fun) ->
-    rabbit_db_queue:update(Name, Fun).
+    update(Name, Fun, #{}).
+
+-spec update
+        (name(), fun((amqqueue:amqqueue()) -> amqqueue:amqqueue()),
+        #{timeout => timeout()}) -> 'not_found' | amqqueue:amqqueue().
+
```
kjnilsson, line 310: why is the spec disconnected from the function by a newline? This makes it harder to read and realise they "go together"

dumbbell: I see this comment was left unanswered but still marked as resolved and the patch was merged. Was this discussed out-of-band?

PR #14700, opened 2025-10-06
rabbit_vhost: Allow vhosts with not-enabled default queue types

the-mikedavis:
The prior validation for vhosts' default queue types ensures that the default queue type can be declared. This seems slightly heavy-handed since users might never rely on default-queue-type behavior. Instead the creation of the vhost should succeed but warn the user that declarations relying on the default queue type will fail - the `is_enabled/0` callback of `rabbit_queue_type` should be limited to block queue declarations, and not block the creation of a vhost.

This is a follow-up to https://github.com/rabbitmq/rabbitmq-server/pull/14624

### DIFF
--- a/deps/rabbit/src/rabbit_vhost.erl
+++ b/deps/rabbit/src/rabbit_vhost.erl
@@ -215,7 +215,12 @@ do_add(Name, Metadata0, ActingUser) ->
                         true ->
                             ok;
                         false ->
-                            throw({error, queue_type_feature_flag_is_not_enabled})
+                            ?LOG_WARNING("Default queue type '~ts' of virtual "
+                                         "host '~ts' is not enabled. All "
+                                         "declarations relying on the default "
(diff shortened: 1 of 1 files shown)


the-mikedavis:
An alternative to this change could be to make no changes at all. The problem being solved here is: if you set `queue_types.classic.enabled = false` in config then boot-up fails from declaring the default vhost with the default queue type. If you want to really want to prevent classic queue creation you would also want to update the default queue type as well as this config option. So we could make no changes instead, although it would be a bit of a footgun.

michaelklishin:
@the-mikedavis I'm leaning towards the latter because updating the DQT is not entirely trivial per https://www.rabbitmq.com/docs/vhosts#default-queue-type.

I have thought of a command that would update (via a merge) a virtual host's metadata, including the DQT, but I don't think we have something like that.

So at the moment, the virtual host declaration must fail early.

PR #14725, opened 2025-10-14
Grafana dashboard const 60s rate interval replaced to function $__rate_interval

AlexeyMAGician:
Grafana dashboard work incorrect if prometheus scrape interval more or equal 60s.
Const 60s rate interval replaced to function $__rate_interval

[x] I have read the `CONTRIBUTING.md` document


mkuratczyk:
thanks!

PR #14740, opened 2025-10-16
Use deliver v2 in stream test utils

acogoluegnes:
Deliver v2 contains the last committed ID (first offset in the last
committed chunk), which can be useful to calculate the offset lag in
some tests.

This requires to use the exchange_command_versions command when opening
the stream connection.


acogoluegnes:
@Mergifyio backport v4.1.x

PR #14765, opened 2025-10-21
rabbit_vhost: Set default description and tags in vhost metadata

dcorbacho:


dcorbacho:
@Mergifyio backport v4.1.x

PR #14767, opened 2025-10-21
unit_access_control_authn_authz_context_propagation_SUITE: Declare test auth backend module

dumbbell:
## Why

rabbit_access_control now ensures all configured auth backend modules are provided by `rabbit` or an enabled plugin. This is not the case for this testsuite's own auth backend.

## How

Declare `rabbit_auth_backend_context_propagation_mock` as a test auth backend module.


michaelklishin:
@dumbbell I like the idea.

FTR, the relevant core part this change relies on is [rabbit_access_control.erl#L158](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_access_control.erl#L158).

PR #14795, opened 2025-10-23
Handle multiple alarms consistently in protocol readers

the-mikedavis:
The MQTT, WebMQTT and STOMP reader modules tracked whether they were blocked by any alarm, but they should instead keep a set of active alarms. Only tracking the `Conserve` part of the alarm notification from `rabbit_alarm` doesn't work correctly if a cluster enters and exits multiple alarms. For example if a cluster had both memory and disk alarms active and then cleared memory, these readers would unblock while they should remain blocked until both alarms are cleared.

Keeping a set of `rabbit_alarm:resource_alarm_source()`s matches the AMQP 0-9-1 and 1.0 readers. This change also updates the 0-9-1 reader to use map-based `sets` for the sake of consistency.


michaelklishin:
Excellent catch. Thank you, @the-mikedavis.

PR #14796, opened 2025-10-23
rabbit_logger_exchange_h: Do not re-enter itself

dumbbell:
## Why

Publishing a log message to an exchange might trigger other messages to be logged. This caused two issues:
1. the exchange logger re-entering itself in an infinite loop
2. if the message is logged from a gen_server-like process, like a Ra server, that is involved in the publishing code path, the process might call itself, leading to a blocked situation

## How

The first issue is fixed with a variable stored in the process dictionary by the `log/2` function. This way, the function can check if it is called from itself because the first incantation stored a variable there.

The second issue is fixed by publishing the message asynchronously from a separate process. This is ok because we don't care if the publish was successful or not. We re-use the process that was started initially to declare the exchange.

Fixes #14069.

### DIFF
--- a/deps/rabbit/src/rabbit_logger_exchange_h.erl
+++ b/deps/rabbit/src/rabbit_logger_exchange_h.erl
@@ -180,7 +180,13 @@ wait_for_initial_pass(N) ->
     end.
 
 setup_proc(
-  #{config := #{exchange := Exchange}} = Config) ->
+  #{id := Id,
+    config := #{exchange := Exchange}} = Config) ->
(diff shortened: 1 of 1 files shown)


michaelklishin (approved):

michaelklishin:
FTR, this was backported to `v4.1.x` for `4.1.8` because it a relevant change for those upgrading to `4.1` (let's not get into why some still upgrade to `4.1` when the same set of series can upgrade directly to `4.2`).

PR #14814, opened 2025-10-24
Allow configuring osiris data_dir in Cuttlefish config

the-mikedavis:
This is the same as the `raft.data_dir` option but for Osiris' data directory. Configuring this in Cuttlefish is nicer than the existing `$RABBITMQ_STREAM_DIR` environment variable way of changing the dir.


michaelklishin:
If we treat https://github.com/rabbitmq/rabbitmq-server/pull/15011 as a [usability] bug, then it's only fair to treat this contribution the same way.

PR #14815, opened 2025-10-24
Allow configuring per-mount-point per-queue-type disk alarms

the-mikedavis:
This is an extension of the free disk space alarm which allows configuring additional mount points to monitor and which queue type(s) to block when they are near full. For example with a config like so:

```ini
stream.data_dir = /mnt/data/streams
# Directory where the file system is mounted.
disk_free_limits.stream.mount_point = /mnt/data/streams
# Alarm threshold: if free space falls under this absolute
# limit then an alarm fires per queue type.
disk_free_limits.stream.absolute = 2GB
# Queue types to block when the threshold is breached.
disk_free_limits.stream.queue_types = stream
```

Publishers to streams would be blocked once the free space of `/data/stream-data` falls under 2GB. Publishers to classic or quorum queues could continue though.

The motivation of this feature is that you may want to use separate disks for different queue types. For example for higher throughput you may want to use volume(s) with better throughput and/or IOPS for streaming but use standard disks for queue data. Also, alarms are currently fairly aggressive by blocking all publishing. Ideally you should be able to continue using queues when the space you have allocated for streams fills up, or vice versa.

This is a different approach than https://github.com/rabbitmq/rabbitmq-server/pull/14086. Instead of measuring disk usage under a directory like [`du(1)`](https://www.man7.org/linux/man-pages/man1/du.1.html), `rabbit_disk_monitor` is updated to measure free space of all mounts at once with [`disksup:get_disk_info/0`](https://www.erlang.org/doc/apps/os_mon/disksup#get_disk_info/0). Under the hood this performs the same [`df(1)`](https://www.man7.org/linux/man-pages/man1/df.1.html) check as `rabbit_disk_monitor` had been doing previously - measuring mount-point free space is much cheaper than measuring directory disk footprint. Monitoring mount points is also quite flexible: you can use multiple disks on one mount point with RAID-0 striping or split up a single disk with partitions.

This is a draft - it needs tests and currently only AMQP 0-9-1 is updated to perform selecting blocking. All other protocols currently block for any alarm.

Some of the commits in this branch are refactors that could be cherry-picked out. https://github.com/rabbitmq/rabbitmq-server/pull/14814 is pretty trivial and the refactors to use maps instead of `dict` in `rabbit_alarm` and use `disksup` instead of the custom `df` code in `rabbit_disk_alarm` are not strictly related to the feature here.

Discussed in https://github.com/rabbitmq/rabbitmq-server/discussions/14590


deps/amqp_client/src/amqp_gen_connection.erl
```diff
@@ -214,18 +215,30 @@ handle_cast({register_blocked_handler, HandlerPid},
...
+    case should_block(State1) of
         true ->
             handle_method(#'connection.unblocked'{}, State1);
-        false when WasNotBlocked ->
+        false when not WasBlocked ->
+            handle_method(#'connection.blocked'{}, State1);
+        false ->
+            {noreply, State1}
+    end;
+handle_cast({channel_published_to_queue_type, _ChPid, QT},
```
the-mikedavis: This feature might need a feature flag. Here for direct connections if old client code is used on a newer server then it would error after publishing since it isn't expecting this cast. I think it would be unlikely to happen in practice but the mixed-version test suite will probably run into this.

samuelmasse:
What would the config setup be for having a main disk that contains quorum and classic queues and a secondary disk that contains streams. Would we specify the same mount point for quorum and classic with each defining queue_types as quorum and classic respectively? Would that result in a common alarm for both or two alarms looking at the same thing.

the-mikedavis:
Ah yeah, in that scenario you could have a config like so:

```ini
disk_free_limits.streaming.mount_point = /mnt/data/streams
disk_free_limits.streaming.absolute = 2GB
disk_free_limits.streaming.queue_types = stream

disk_free_limits.messaging.mount_point = /mnt/data/queues
disk_free_limits.messaging.absolute = 2GB
disk_free_limits.messaging.queue_types = classic,quorum
```

And if `/mnt/data/queues` went under its configured limit it would set two alarms (disk for classic and disk for quorum queue types) but wouldn't affect streams.

samuelmasse:
Ah I see thanks! So if I understand correctly the name here `disk_free_limits.[name].mount_point` can be any name we want to set it to, it wouldn't have to be the name of a queue type. So I could set `disk_free_limits.bob.mount_point` for example.

Taking that thought further, what would the process of adding that "bob" disk alarm to an existing broker look like. If node A thinks the "bob" alarm exists but node B doesn't can there be issues that comes from the disagreement? When node A restarts with the new configuration do all nodes now know of the "bob" alarm or just node A until all other nodes also restart.

Also after I added my new "bob" disk alarm, what ways do I have as a user to then monitor the "bob" alarm to see if it's currently alarming, what value it is configured to and how close it's getting to the alarm point. For MQ's use case currently we are getting this information from the `/api/nodes` endpoint using the `disk_free_limit`, `disk_free` and `disk_free_alarm` values. Are we thinking about adding an alarm name map to the output of this API to give those values for each disk alarm. So I would maybe access `disk_free_map.bob.disk_free` to know the status of my "bob" disk alarm.

Lastly for the RabbitMQ console we currently have a column named "Disk space" that displays the information of the as of now only disk alarm. When I add this "bob" disk alarm would we want to dynamically add a new column to that table named something like "Disk space (bob)". In that case would we also support defining the ordering of those columns. For example if we consider that it's most relevant to display the disk alarm of quorum queues on the left, then streams, then classic and at the end the disk alarm for non queue storage, would we be able to define that order manually in some console config.

the-mikedavis:
The `[name]` part is only used to name the disk. Different members of the cluster could use different names if they wanted: alarms are set by queue-type rather than by disk name. So if you added a node to a cluster and it set an alarm, the other nodes would know to block those queue types regardless of the name of the disk.

Yeah before this change is finished we should really expose the configured disks in the API (and maybe prometheus as well?) and the UI. For the ordering part maybe we can use the order they are listed in the config file? There's an ETS table holding the info of the last-known available bytes and configured limit which can be queried cheaply for the metrics. And we should extend the CLI commands so that the limits can be updated dynamically.

the-mikedavis:
Looks like we'll have to hold off on using `disksup` in `rabbit_disk_monitor` as it currently doesn't handle Unicode correctly https://github.com/erlang/otp/pull/10721. `unicode_SUITE` fails because of that change.

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
```diff
@@ -865,6 +869,12 @@ get_data(exchange_names, _, _) ->
                         Label = <<"vhost=\"", VHost/binary, "\",exchange=\"", Name/binary, "\",type=\"", (atom_to_binary(EType))/binary, "\"">>,
                         [{Label, 1}|Acc]
                 end, [], rabbit_exchange:list());
+get_data(mount_metrics, _, _) ->
+    [{<<"disk=", Name/binary>>, Available, Limit}
```
lukebakken:
```suggestion
    [{<<"disk=\"", Name/binary, "\"">>, Available, Limit}
```

lukebakken (commented):
All other Prometheus label values are in double-quotes, so it seems like `disk` should be as well.

lukebakken (commented):
Here's a summary of the review performed by @lukebakken and the 🧞:

## Commits added during review

- f534680fb8 - `rabbit_disk_monitor`: Document NaN alarm behavior in `alarmed_mounts/1` and `alarmed_queue_types/1`
- 82da3af0fb - `unit_disk_monitor_SUITE`: Test that duplicate mount paths are allowed
- b5587f6b17 - `rabbit.schema`: Validate that `disk_free_limits` entries have all required fields; test added
- ddb65ff3bb - `prometheus`: Fix mount metric label to use quoted value (`disk="name"` not `disk=name`)
- f1c68778e7 - `set_disk_free_limit_command_test`: Add tests for the `mount` subcommand
- c051d42f58 - `rabbit_alarm`: Clarify log message for per-queue-type disk alarms
- f1379d179f - `rabbit_prometheus_http_SUITE`: Test mount metric label format
- a23840ac92 - `rabbit_disk_monitor`: Handle multiple disks in `resolve_data_dir/0`
- b8a3c72117 - `amqp_client_SUITE`: Test per-queue-type disk alarm blocking for direct connections
- 66f298e441 - `amqp_client`: Per-channel `queue_types_published` tracking for direct connections

## Issues resolved

**`amqp_gen_connection.erl`**: The original flat set approach meant a direct connection that once published to a queue type would always be considered as publishing to that type, even after channels closed. This has been fixed: the server channel pid is now threaded through the channel startup path as a tagged tuple `{direct, ServerPid} | {network, WriterPid}`, allowing `amqp_gen_connection` to track `queue_types_published` per server channel pid and clean up correctly when a channel closes. The full `amqp_client_SUITE` (157 tests) passes with this change.

**Duplicate queue types**: If the same queue type appears in two `disk_free_limits` entries, `alarmed_queue_types/1` deduplicates via `sets:union` - no crash, no double-alarming. A connection is blocked if any mount carrying that queue type is alarmed, and unblocked only when all such mounts clear. This is correct and intentional behavior.

## Issues to discuss

**STOMP and STOMP-over-WebSockets**: These protocols are architecturally different from all others: they are built on top of `amqp_client` rather than implementing the broker protocol directly. `rabbit_stomp_processor` uses `amqp_connection:start(#amqp_params_direct{})`, so publishes go through `amqp_gen_connection` → `rabbit_channel`. This creates a layering problem:

- `amqp_gen_connection` is registered with `rabbit_alarm` (via `rabbit_direct`) and now correctly implements selective per-queue-type blocking
- `rabbit_stomp_reader` is **also** independently registered with `rabbit_alarm` and has its own socket throttling via `control_throttle`, which uses `not sets:is_empty(BlockedBy)` - blocking on any alarm regardless of queue type
- `rabbit_web_stomp_handler` has a `conserve_resources` boolean field that appears to be dead code - nothing in the codebase sends `{conserve_resources, Conserve}` to it

The correct fix is to remove `rabbit_stomp_reader`'s direct `rabbit_alarm` registration and instead register it as the block handler on `amqp_gen_connection` after login, letting `amqp_gen_connection` drive socket throttling via `connection.blocked` / `connection.unblocked`. This is pre-existing behavior not introduced by this PR - STOMP was already blocking on all alarms before. Is this worth addressing in this PR or should it be a follow-up?

## Notes

- The PR description says "currently only AMQP 0-9-1 is updated to perform selective blocking" - this appears to be outdated. Stream, MQTT, MQTT-over-WebSockets, and AMQP 1.0 all have selective blocking implemented in the branch.

PR #14825, opened 2025-10-27
Switch from pg_local to pg

mkuratczyk:
pg_local is based on the pg2 module, which got removed
from Erlang/OTP years ago. It was replaced by the more
efficient pg module, so let's use it directly.

We use node-local scopes and single-item groups so that
(de)registration is fast, but we can list all connections/channels
(by listing all groups in the scope).

Initially we wanted to remove pg_local altogether, but we decided to keep it
just for rabbit_direct, which uses potentially remote PIDs, so it can't use pg.


michaelklishin:
To add a bit of context here.

Process groups can be used in many ways but specifically in RabbitMQ they were used for tracking of exclusive queues on an AMQP 0-9-1 connection, so that when a connection terminates, all exclusive queues (if any) on it would be terminated very shortly after.

RabbitMQ Stream Protocol does not allow clients to declare exclusive streams, so it likely can avoid using process groups entirely. And so can most protocol readers.

deps/rabbit/src/rabbit_volatile_queue.erl
```diff
@@ -208,7 +208,7 @@ local_call(Pid, Request) ->
 
 is_local(Pid) ->
     rabbit_amqp_session:is_local(Pid) orelse
-    pg_local:in_group(rabbit_channels, Pid).
+    pg:get_local_members(rabbit:pg_scope_amqp091_channel(), Pid) =/= [].
```
ansd: Should we follow the same approach as done for `rabbit_amqp_session:is_local(Pid)` and introduce a function `rabbit_channel:is_local/1`?
Currently, this code line leaks a lot of details of how channels are tracked into the volatile queue type.

deps/rabbit/src/rabbit_direct.erl
```diff
@@ -186,7 +190,7 @@ connect1(User = #user{username = Username}, VHost, Pid, Infos) ->
             AuthzContext = proplists:get_value(variable_map, Infos, #{}),
             try rabbit_access_control:check_vhost_access(User, VHost,
                                                {ip, PeerHost}, AuthzContext) of
-                ok -> ok = pg_local:join(rabbit_direct, Pid),
+                ok -> ets:insert(?TABLE, {Pid}),
```
ansd: It's unfortunate that we can't use pg here. Does this Pid need to be monitored? In other words, is disconnect/2 always guaranteed to be called? Just want to make sure we won't create a leak here.

michaelklishin: `is_disconnect/2` is seemingly only run when `amqp_direct_connection:terminate/2` runs. Which is not the case in certain failure scenarios, e.g. on the node which is the client fails.

We can introduce an equivalent of the metrics GC process which runs periodically and sweeps what seems stale (unused).

ansd: Alternatively, we don't delete the pg_local module and just keep using it only in `rabbit_direct`. This sounds better to me than manual ETS tables with periodically sweeping them.

This whole PR here still provides benefits by removing pg_local usages almost everywhere. We can mark pg_local as deprecated via code so that no new code starts using it.

michaelklishin: I agree, we can keep that module for just `rabbit_direct` and use the standard `pg` as much as we can.

mkuratczyk:
PR review feedback has been incorporated: I've re-introduced pg_local just for rabbit_direct and added `rabbit_channel:is_local/1`

ansd:
I added a deprecation directive to deprecate the entire `pg_local` module. This way, new code won't accidentally use this old module.

ansd (approved):

PR #14827, opened 2025-10-27
khepri_core_ff: Fix 'rabbitmq_4.2.0' feature flag definition

dumbbell:
## Why

The definition specified an `enable` callback, but it should simply define a dependency.


dumbbell:
Please disregard this pull request, I misunderstood the callback.

PR #14864, opened 2025-10-30
Update RabbitMQ series versions in discussion templates

the-mikedavis:
Since v4.2.0 is now released we can add 4.2.x to the discussion templates


michaelklishin:
Thank you!

PR #14873, opened 2025-10-31
Shovel bugfix: handle cluster alarms in AMQP10 and local shovels

dcorbacho:
AMQP10 needs to handle the `remote_income_window_exceeded` error when an alarm is triggered.

Local shovels need to stop publishing and store the pending deliveries when an alarm is triggered. When all alarms are cleared, it must publish the pending deliveries.

### DIFF
--- a/deps/rabbitmq_shovel/src/rabbit_local_shovel.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_local_shovel.erl
@@ -24,6 +24,7 @@
 
 -export([
          boot_step/0,
+         conserve_resources/3,
          parse/2,
          connect_source/1,
          connect_dest/1,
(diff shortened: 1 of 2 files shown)


michaelklishin (changes requested):
The test failures are reproducible.

deps/rabbitmq_shovel/src/rabbit_local_shovel.erl
```diff
@@ -425,8 +457,11 @@ add_routing(Msg0, Dest) ->
...
-status(_) ->
-    running.
+status(State) ->
+    case is_blocked(State) of
+        true -> blocked;
+        false -> running
+    end.
 
 pending_count(_State) ->
     0.
```
gomoripeti, line 469: now that there can be pending messages in local shovels too, pending_count should be the length of `pending_delivery`

PR #14884, opened 2025-11-03
rabbit_stream_coordinator: Handle delete_stream idempotently

### LINKED ISSUE #14852: Stream queues become 'stuck' in a partial deletion
### Describe the bug

When deleting a stream, the stream coordinator may delete the underlying stream but the deletion from `rabbit_db_queue` could fail since they are independent Raft clusters. If this happens and the stream coordinator fully deletes the stream, the stream queue then becomes sort of 'stuck' as subsequent calls to `rabbit_stream_queue:delete/4` time out: when the stream coordinator doesn't know about a stream in a `{delete_stream, StreamId, #{}}` command it does not reply to the caller.

### Reproduction steps

This is probably really hard to reproduce in practice. In a shell though we can 'fake' this state pretty easily:

1. `make run-broker`
1. `stream-perf-test --time 1` - create the stream
1. `[SQ] = rabbit_db_queue:get_all().`
1. `rabbit_stream_coordinator:process_command({delete_stream, maps:get(name, amqqueue:get_type_state(SQ)), #{}}).` - simulate partial failure by only deleting from the stream coordinator and not the metadata store.

After this, the stream can't be deleted via AMQP, the stream protocol, UI, etc..

### Expected behavior

I think it's reasonable for the stream coordinator to reply `ok` when prompted with a delete stream for an unknown stream ID. So the command would be idempotent.

### Additional context

The stream coordinator replies after it performs the deletion. If there is no stream then we hit the clause [here](https://github.com/rabbitmq/rabbitmq-server/blob/520119f33f1fc94fe0d3e13f48944624c443515c/deps/rabbit/src/rabbit_stream_coordinator.erl#L1720-L1721) which results in no reply. So calls to delete a stream which do not exist (according to the coordinator) will time out. We could add a clause for the `delete_stream` command against an `undefined` stream which would result in an `ok` reply. It looks like this may take some refactoring though. Then [`rabbit_stream_coordinator:delete/2`](https://github.com/rabbitmq/rabbitmq-server/blob/520119f33f1fc94fe0d3e13f48944624c443515c/deps/rabbit/src/rabbit_stream_coordinator.erl#L191-L204) would continue to delete from the metadata store.

the-mikedavis:
Within `rabbit_stream_coordinator:delete_stream/2`, the stream could be successfully deleted from the stream coordinator but the update to the metadata store (`rabbit_amqqueue:internal_delete/2`) could fail separately. If this happens then subsequent attempts to delete the stream also fail because the coordinator did not reply to `delete_stream` requests for unknown stream IDs.

This change has the stream coordinator reply `ok` when it receives a `delete_stream` command and the stream is `undefined`. This would let the `delete_stream/2` continue and remove the lingering state from the metadata store.

Fixes #14852

### DIFF
--- a/deps/rabbit/src/rabbit_stream_coordinator.erl
+++ b/deps/rabbit/src/rabbit_stream_coordinator.erl
@@ -585,7 +585,7 @@ apply(#{index := _Idx, machine_version := MachineVersion} = Meta0,
                            State0#?MODULE{streams = Streams0#{StreamId => Stream},
                                           monitors = Monitors}, Reply, Effects)
             end;
-        Reply ->
+        {reply, Reply} ->
             return(Meta, State0, Reply, [])
     end;
 apply(Meta, {sac, SacCommand}, #?MODULE{single_active_consumer = SacState0,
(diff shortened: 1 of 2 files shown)


the-mikedavis:
The change turned out to be smaller than I expected if we take advantage of the existing `filter_command/3` helper.

Does this need a new machine version? The only difference between old and new version is adding the reply effect, and reply effects are dropped by followers and during recovery IIRC.

PR #14885, opened 2025-11-03
Prometheus: stream HTTP responses (a double digit % reduction of both CPU and memory footprint)

the-mikedavis:
`prometheus_text_format:format/1` produces a binary of the format for the entire registry. For clusters with many resources, this can lead to large replies from `/metrics/[:registry]` especially for large registries like `per-object`. Instead of formatting the response and then sending it, we can stream the response by taking advantage of the new `format_into/3` callback (which needs to be added upstream to the `prometheus` dep). This uses `cowboy_req:stream_body/3` to stream the iodata as `prometheus` works through the registry.

This should hopefully be a nice memory improvement. The other benefit is that results are sent eagerly. For a stress-testing example,

1. `make run-broker`
2. `rabbitmqctl import_definitions path/to/100k-classic-queues.json`
3. `curl -s localhost:15692/metrics/per-object`

Before this change `curl` would wait for around 8 seconds and then the entire response would arrive. With this change the results start streaming in immediately.

Discussed in #14865

### DIFF
--- a/deps/rabbitmq_prometheus/src/rabbit_prometheus_handler.erl
+++ b/deps/rabbitmq_prometheus/src/rabbit_prometheus_handler.erl
@@ -14,7 +14,6 @@
 
 -define(SCRAPE_DURATION, telemetry_scrape_duration_seconds).
 -define(SCRAPE_SIZE, telemetry_scrape_size_bytes).
--define(SCRAPE_ENCODED_SIZE, telemetry_scrape_encoded_size_bytes).
 
 -define(AUTH_REALM, "Basic realm=\"RabbitMQ Prometheus\"").
 
@@ -58,14 +57,9 @@ setup_metrics(Registry) ->
                   {help, "Scrape size, not encoded"},
                   {labels, ["registry", "content_type"]},
                   {registry, Registry}],
-    ScrapeEncodedSize = [{name, ?SCRAPE_ENCODED_SIZE},
-                         {help, "Scrape size, encoded"},
-                         {labels, ["registry", "content_type", "encoding"]},
-                         {registry, Registry}],
 
     prometheus_summary:declare(ScrapeDuration),
-    prometheus_summary:declare(ScrapeSize),
-    prometheus_summary:declare(ScrapeEncodedSize).
+    prometheus_summary:declare(ScrapeSize).
 
 %% ===================================================================
 %% Private functions
@@ -100,47 +94,23 @@ format_metrics(Request, Registry) ->
     %% Formatting registries produces large binaries. Fullsweep eagerly to
     %% evict the large binaries faster and make GC cheaper.
     process_flag(fullsweep_after, 0),
-    AcceptEncoding = cowboy_req:header(<<"accept-encoding">>, Request, undefined),
     ContentType = prometheus_text_format:content_type(),
-    Encoding = accept_encoding_header:negotiate(AcceptEncoding, [<<"identity">>,
-                                                                 <<"gzip">>]),
-    Headers = #{<<"content-type">> => ContentType,
-                <<"content-encoding">> => Encoding},
-    case Encoding of
-        <<"gzip">> ->
-            Scrape = prometheus_summary:observe_duration(
-                       Registry,
-                       ?SCRAPE_DURATION,
-                       [Registry, ContentType],
-                       fun () -> prometheus_text_format:format(Registry) end),
-            prometheus_summary:observe(Registry,
-                                       ?SCRAPE_SIZE,
-                                       [Registry, ContentType],
-                                       iolist_size(Scrape)),
-            Encoded = zlib:gzip(Scrape),
-            prometheus_summary:observe(telemetry_registry(),
-                                       ?SCRAPE_ENCODED_SIZE,
-                                       [Registry, ContentType, Encoding],
-                                       iolist_size(Encoded)),
-            cowboy_req:reply(200, Headers, Encoded, Request);
-        <<"identity">> ->
-            Req = cowboy_req:stream_reply(200, Headers, Request),
-            Fmt = fun(Size, Data) ->
-                cowboy_req:stream_body(Data, nofin, Req),
-                Size + byte_size(Data)
-            end,
-            Size = prometheus_summary:observe_duration(
-                     Registry,
-                     ?SCRAPE_DURATION,
-                     [Registry, ContentType],
-                     fun () -> prometheus_text_format:format_into(Registry, Fmt, 0) end),
-            cowboy_req:stream_body(<<>>, fin, Req),
-            prometheus_summary:observe(Registry,
-                                       ?SCRAPE_SIZE,
-                                       [Registry, ContentType],
-                                       Size),
-            Req
-    end.
+    Req = cowboy_req:stream_reply(200, #{<<"content-type">> => ContentType}, Request),
+    Fmt = fun(Size, Data) ->
+        cowboy_req:stream_body(Data, nofin, Req),
+        Size + byte_size(Data)
+    end,
+    Size = prometheus_summary:observe_duration(
+             Registry,
+             ?SCRAPE_DURATION,
+             [Registry, ContentType],
+             fun () -> prometheus_text_format:format_into(Registry, Fmt, 0) end),
+    cowboy_req:stream_body(<<>>, fin, Req),
+    prometheus_summary:observe(Registry,
+                               ?SCRAPE_SIZE,
+                               [Registry, ContentType],
+                               Size),
+    Req.
 
 validate_registry(undefined, auto) ->
     {true, default};
--- a/deps/rabbitmq_prometheus/Makefile
+++ b/deps/rabbitmq_prometheus/Makefile
@@ -9,7 +9,7 @@ endef
 PROJECT := rabbitmq_prometheus
 PROJECT_DESCRIPTION = Prometheus metrics for RabbitMQ
 PROJECT_MOD := rabbit_prometheus_app
-DEPS = accept cowboy rabbit rabbitmq_management_agent prometheus rabbitmq_web_dispatch
+DEPS = cowboy rabbit rabbitmq_management_agent prometheus rabbitmq_web_dispatch
 BUILD_DEPS = amqp_client rabbit_common rabbitmq_management
 TEST_DEPS = rabbitmq_ct_helpers rabbitmq_ct_client_helpers rabbitmq_stream
 
--- a/deps/rabbitmq_prometheus/test/rabbit_prometheus_http_SUITE.erl
+++ b/deps/rabbitmq_prometheus/test/rabbit_prometheus_http_SUITE.erl
@@ -373,7 +373,7 @@ content_type_test(Config) ->
 
 encoding_test(Config) ->
     {Headers, Body} = http_get(Config, [{"accept-encoding", "deflate"}], 200),
-    ?assertMatch("identity", proplists:get_value("content-encoding", Headers)),
+    ?assertMatch(undefined, proplists:get_value("content-encoding", Headers)),
     ?assertEqual(match, re:run(Body, "^# TYPE", [{capture, none}, multiline])).
 
 gzip_encoding_test(Config) ->
--- a/rabbitmq-components.mk
+++ b/rabbitmq-components.mk
@@ -39,7 +39,6 @@ endif
 # We do that in this file, which is included by all projects, to ensure
 # all projects use the same versions. It avoids conflicts.
 
-dep_accept = hex 0.3.5
 dep_cowboy = hex 2.14.1
 dep_cowlib = hex 2.16.0
 dep_credentials_obfuscation = hex 3.5.0


the-mikedavis:
Ah welp, when measured this doesn't look as promising as I thought. I have three EC2 instances, one acting as the scraper and the other two "galactica" and "kestrel" running single-instance brokers with the [100k-classic-queues.json](https://github.com/rabbitmq/sample-configs/blob/main/queues/100k-classic-queues.json) definition import. The scraping node runs this script to scrape from each node every 2 seconds:

```bash
#! /usr/bin/env bash

N=600
SLEEP=2
for i in $(seq 1 $N)
do
  echo "Sleeping ${SLEEP}s... ($i / $N)"
  sleep $SLEEP
  echo "Ask for metrics from $1... ($i / $N)"
  curl -s "http://$1:15692/metrics/per-object" --output /dev/null &
done

wait
```

I swapped which node was running which branch, but we can see that this branch consistently has _more_ EC2 instance-wide memory usage rather than less! Galactica:

galactica

Kestrel:

kestrel

In the first test (01:03 - 01:23) Galactica runs `main` and in the second (02:00 - 02:20) Kestrel runs `main`.

So it looks like this branch is worse for memory usage as-is. I will have to do a bit more digging. Seems like passing the iodata to the Cowboy process might be creating more garbage than writing the data to the ram_file port. We might be able to buffer some of the iodata in the callback or restructure things in the prometheus dep to improve memory usage.

lhoguin:
How big does the counter get? You may have increased the number of messages to Cowboy drastically and each message has a cost (data gets processed and buffered at various steps of the sending process).

the-mikedavis:
It's actually not quite as big as I thought, I see 147_161_056 bytes sent in 557 calls to `cowboy_req:stream_body/3`. Looking at the formatting code, per metric-family the prelude (`# TYPE` and `# HELP`) is formatted with an iolist and then the metrics themselves are all built with one binary taking advantage of the VM's binary append optimization. Let me try making an iolist out of the prelude and metrics binary so we cut the number of `cowboy_req:stream_body/3`s in half. Also the binaries must be pretty large, I wonder if `process_flag(fullsweep_after, 0)` could help with the garbage collection here.

the-mikedavis:
Setting `fullsweep_after` does seem to help. Reducing the number of calls to `cowboy_req:stream_body/3` doesn't seem to make a difference. Looking an observer_cli I think what's happening here is that the streaming version just has more outstanding requests at once (`cowboy_stream_h:request_process/3`) and each has a fairly large size in its process heap (~200-350 MB). And it has more outstanding requests because it's slower, and the script scrapes on an interval rather than waiting for the full response.

```
$ time curl -v http://<streaming>:15692/metrics/per-object --output /dev/null
real	0m7.490s
user	0m0.012s
sys	0m0.056s
$ time curl -v http://<ram_file>:15692/metrics/per-object --output /dev/null
real	0m5.790s
user	0m0.009s
sys	0m0.046s
```

the-mikedavis:
Ok! I spent some time looking at allocation in `prometheus_text_format:format/1` and with optimizations there, this change to stream results is looking like a clear improvement now.

Tracking host usage with Prometheus' `node_exporter` this time, `galactica` runs this branch and `kestrel` runs `main`:

#### `kestrel` / baseline (CPU)

grafana-kestrel-cpu

#### `kestrel` / baseline (RAM)

grafana-kestrel-mem

#### `galactica` / this change (CPU)

grafana-galactica-cpu

#### `galactica` / this branch (MEM)

grafana-galactica-mem

---

This uses the same setup as above to scrape per-object metrics for single-instance brokers with `100k-classic-queues.json` imported every two seconds. This is using two `m7g.xlarge` EC2 instances (4vCPU ARM, 16 GB RAM) running RabbitMQ via `make` on Erlang/OTP 27. We see `main` pinned at around 95% CPU usage and hovering around 9 GB of peak memory usage. With this change the CPU usage hovers around 60-65% instead and around 6.5-7.5 GB peak memory usage.

NOTE! The garbage improvements to the `prometheus` dep actually make it slightly more desirable to not stream the response. CPU instead hovers around 57-61% CPU with similar memory usage. My recommendation is to stream the response anyways since applications which scrape RabbitMQ will be able to work gradually rather than handle the entire response at once.

michaelklishin:
So now we have sound double digit % improvements for both CPU and memory footprint. Awesome!

the-mikedavis:
The peak memory footprint is really at the edge of "double digits" if I'm being honest, it's right around 10% - these instances really have more like 15 GB memory 😅. I thought we would see great peak memory usage improvements here but it's CPU savings instead actually. Reducing the work the GC needs to do seems to pay off.

Looking at the `tprof` output (see https://github.com/prometheus-erl/prometheus.erl/pull/196) there is more we could do if we were really motivated to optimize these endpoints. We pay a surprisingly high price for the conversions we do with `prometheus_model_helpers` - every record/tuple we allocate with those helpers adds up, up to 12%, the second highest factor in this `tprof` output. At the cost of replacing the prometheus dep entirely we could format directly into binaries or iodata and avoid the intermediate allocations. This would be a fairly large change though, and wouldn't benefit the other downstream `prometheus` dependencies like this does.

mkuratczyk:
Thanks for looking into this, the improvements are already significant. Your observations about prometheus.erl concur what I saw when switching to ra_counters. I even implemented a simple prometheus-format exporter and it was already better. However, I ultimately didn't use it since we would lose some functionality (eg. `global_labels`, that seshat is unaware of and, while I'm likely the only user of `global_labels`, I'd feel bad if I removed them :) ). I think prometheus.erl should have "fast track" functions - if we know that what we are exporting is already correct, skip all the validations and conversions.

lhoguin:
You can stream gzip responses too by using `cowboy_compress_h` instead of doing the gzip manually.

the-mikedavis:
Ah that's handy. Looks like that stream handler is enabled by default: https://github.com/rabbitmq/rabbitmq-server/blob/34c6ef4e2c0ccd838288f3ff7032dc67922ab430/deps/rabbitmq_web_dispatch/src/rabbit_web_dispatch_sup.erl#L118-L125

(Unless you explicitly configure `prometheus.tcp.compress = false` or `prometheus.ssl.compress = false`.) I bet we can just delete the manual gzipping here then.

EDIT: yep I can see compression working correctly with `curl --compressed`. We can just rely on `cowboy_compress_h` to perform the gzipping during streaming :+1:

the-mikedavis:
I'll try experimenting with changing up the formatter when I get the chance. I think we might need to change or introduce a new callback for collectors that would let us build an accumulator. Also TIL about global_labels, that does seem nice at least for performance testing!

lhoguin:
Looks good but there's a test error:

```
2025-11-07T19:21:12.9538760Z - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
2025-11-07T19:21:12.9539375Z rabbit_prometheus_http_SUITE:encoding_test failed on line 376
2025-11-07T19:21:12.9542629Z ##[error]Reason: {assertMatch,[{module,...},{...}|...]}
2025-11-07T19:21:12.9543801Z - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
```

the-mikedavis:
Yep, we'll need to get those prometheus.erl changes merged+published upstream before I can mark this as properly ready-for-review. Without the allocation changes, this change performs worse than baseline :/

For the accept-encoding stuff, we could just delete the line: https://github.com/rabbitmq/rabbitmq-server/blob/b7d6bf9bd4ef65cca5ce94e9f68cfb6ccc26ef4b/deps/rabbitmq_prometheus/test/rabbit_prometheus_http_SUITE.erl#L376

That's what fails if we totally omit sending the `content-encoding` header. Previously we were always sending `content-encoding: identity` but I don't think that response header is required in any Prometheus docs I've read. The closest I've seen is https://prometheus.io/docs/instrumenting/content_negotiation/ which only talks about what SHOULD be allowed for `accept-encoding`. (Maybe it would be nice to relax that and allow requesting `zstd` too, since it's supposed to be quite efficient and was added in Erlang/OTP 28)

lhoguin:
The test is wrong. `identity` content coding doesn't actually exist (see paragraph below example at https://www.rfc-editor.org/rfc/rfc9110.html#section-8.4 ) so we shouldn't send it. The test can be modified to check the absence of `content-encoding` header.

I've assigned Nelson to review the prometheus.erl change, wait and see for now.

lhoguin:
We can likely drop the `accept` dependency as a result.

the-mikedavis:
Aha, great catch! Yeah, looks like that was the only use of `accept` so it can also be deleted along with this change.

deps/rabbitmq_prometheus/src/rabbit_prometheus_handler.erl
```diff
@@ -59,7 +59,7 @@ setup_metrics(Registry) ->
                   {labels, ["registry", "content_type"]},
                   {registry, Registry}],
     ScrapeEncodedSize = [{name, ?SCRAPE_ENCODED_SIZE},
-                         {help, "Scrape size, encoded"},
+                         {help, "Scrape size, encoded (deprecated)"},
```
the-mikedavis: Now that we don't eagerly compress with `zlib:gzip/1` we don't know how large compressed responses actually are so we can't record this metric. I doubt anyone was relying on / measuring this but I suppose it's a breaking change, eh? Should we remove this rather than never recording it?

mkuratczyk: Yeah, I think you can just delete it

the-mikedavis:
`prometheus` 6.1.0 (and 6.1.1 with a fix 😅) were published. I tested again with the same script and also with `tprof` and `time curl ...`. This time I used the `100k-classic-queues.json` script for both the per-request and stress test scenarios.

* Per-request total allocation: according to `tprof`, main allocates `174_103_304` words while this branch allocates `132_271_433` words (24% reduction).
    * `tprof` does not know about the port driver's memory usage, so the total savings are probably much more significant.
* Per-request time: `main` takes 6.73 seconds and this branch takes 5.01 seconds (25.6% reduction).
    * Results also stream gradually on this branch. On main, Rabbit waits for most of this time and then responds all-at-once.
* Stress-test CPU usage: `main` hovers around 88%, this branch around 63% (28% reduction).
* Stress-test total memory usage: `main` hovers around 8.5 GB, this branch around 6.5 GB (23.5% reduction).

<details><summary>tprof comparison...</summary>

this branch:

```
****** Process <0.301071.0>  --  99.95% of total *** 
FUNCTION                                                                                  CALLS      WORDS    PER CALL  [    %]
... everything smaller omitted for brevity ...
prometheus_text_format:render_labels/1                                                  2108195     864032        0.41  [ 0.65]
prometheus_model_helpers:label_pair/1                                                    429393    1717572        4.00  [ 1.30]
prometheus_rabbitmq_core_metrics_collector:'-emit_queue_info/3-fun-0-'/3                 100000    2500000       25.00  [ 1.89]
prometheus_model_helpers:counter_metric/2                                                301325    3615900       12.00  [ 2.73]
prometheus_rabbitmq_core_metrics_collector:'-collect_metrics/2-lc$^1/1-0-'/2            2100143    4000076        1.90  [ 3.02]
prometheus_text_format:'-render_labels/1-fun-0-'/2                                       321434    4178762       13.00  [ 3.16]
prometheus_model_helpers:'-metrics_from_tuples/2-lc$^0/1-0-'/2                          2108453    4216300        2.00  [ 3.19]
lists:'-filter/2-lc$^0/1-0-'/2                                                          2208458    4416304        2.00  [ 3.34]
erlang:integer_to_binary/1                                                              2006931    6020818        3.00  [ 4.55]
prometheus_rabbitmq_core_metrics_collector:label/1                                      2000038   10000020        5.00  [ 7.56]
prometheus_rabbitmq_core_metrics_collector:'-collect_metrics/2-lc$^0/1-1-'/2            2100143   10500190        5.00  [ 7.94]
prometheus_text_format:render_value/2                                                   2108200   10541000        5.00  [ 7.97]
prometheus_text_format:render_series/4                                                  2108200   10541663        5.00  [ 7.97]
prometheus_model_helpers:gauge_metric/2                                                 1806812   21681744       12.00  [16.39]
ets:match_object/2                                                                           19   34806092  1831899.58  [26.31]
                                                                                                 132271433              [100.0]
```

main:

```
****** Process <0.301075.0>  --  99.99% of total *** 
FUNCTION                                                                                   CALLS      WORDS    PER CALL  [    %]
... everything smaller omitted for brevity ...
prometheus_text_format:render_labels/1                                                   2108195     864032        0.41  [ 0.50]
prometheus_model_helpers:label_pair/1                                                     429393    1717572        4.00  [ 0.99]
prometheus_rabbitmq_core_metrics_collector:'-emit_queue_info/3-fun-0-'/3                  100000    2500000       25.00  [ 1.44]
prometheus_model_helpers:counter_metric/2                                                 301325    3615900       12.00  [ 2.08]
prometheus_rabbitmq_core_metrics_collector:'-collect_metrics/2-lc$^1/1-0-'/2             2100143    4000076        1.90  [ 2.30]
prometheus_text_format:'-render_labels/1-fun-0-'/2                                        321434    4178762       13.00  [ 2.40]
prometheus_model_helpers:'-metrics_from_tuples/2-lc$^0/1-0-'/2                           2108454    4216300        2.00  [ 2.42]
lists:'-filter/2-lc$^0/1-0-'/2                                                           2208459    4416304        2.00  [ 2.54]
erlang:integer_to_binary/1                                                               2006892    6020698        3.00  [ 3.46]
prometheus_rabbitmq_core_metrics_collector:label/1                                       2000038   10000020        5.00  [ 5.74]
prometheus_rabbitmq_core_metrics_collector:'-collect_metrics/2-lc$^0/1-1-'/2             2100143   10500190        5.00  [ 6.03]
prometheus_text_format:'-emit_mf_metrics/2-fun-0-'/3                                     2108150   10541413        5.00  [ 6.05]
prometheus_model_helpers:gauge_metric/2                                                  1806812   21681744       12.00  [12.45]
prometheus_text_format:has_special_char/1                                               21697535   22147190        1.02  [12.72]
prometheus_text_format:render_series/3                                                   2108200   29713405       14.09  [17.07]
ets:match_object/2                                                                            19   34806094  1831899.68  [19.99]
                                                                                                  174103304              [100.0]
```

---

</details>

Stress test graphs:

| | main | this branch |
|---|---|---|
| CPU | grafana-kestrel-cpu | grafana-galactica-cpu |
| mem | grafana-kestrel-mem | grafana-galactica-mem |

lhoguin:
Looks good, thanks! @mkuratczyk probably wants to test this.

michaelklishin:
@mkuratczyk @lhoguin @the-mikedavis I assume we want to backport this to `v4.2.x` because it is a meaningful optimization of a monitoring endpoint. Any objections?

PR #14888, opened 2025-11-03
Optimization: make stream read-ahead limit configurable

the-mikedavis:
In some consumption patterns, increasing the read-ahead limit can increase consumer throughput where currently the consumer is slower than the producer and falls behind (https://github.com/rabbitmq/rabbitmq-server/discussions/14877). This change adds Cuttlefish config for the option in https://github.com/rabbitmq/osiris/pull/198:

```ini
stream.read_ahead_limit = 8kiB
```

---

EDIT: this configuration was updated in https://github.com/rabbitmq/rabbitmq-server/pull/14894 to change the key name to `stream.read_ahead`. To configure this on v4.2.1+ or v4.3.x+, use:

```ini
stream.read_ahead = 8kiB
```


michaelklishin:
@the-mikedavis we cannot accept a PR that uses an Osiris fork but this optimization can go into mainline Osiris.

the-mikedavis:
Ah yeah, that part of the change is meant to be dropped / replaced when/if this can be added to a ~hex~ git-published version of Osiris

PR #14894, opened 2025-11-04
Merge `stream.read_ahead` and `stream.read_ahead_limit` options

the-mikedavis:
Read-ahead can be controlled by a single option, with zero acting the same as `false`.

This is the companion change for https://github.com/rabbitmq/osiris/pull/199 - see extra context there. This replaces the config option in https://github.com/rabbitmq/rabbitmq-server/pull/14888.

```ini
# prevent any read-ahead
stream.read_ahead = false
# tune read-ahead limit
stream.read_ahead = 8kiB
```


michaelklishin:
@the-mikedavis https://github.com/rabbitmq/osiris/pull/199 was merged and included into the `1.10.2` release.

PR #14897, opened 2025-11-05
Registry: make it possible to define a custom entity class via module attributes

dcorbacho:
`rabbit_registry` classes are currently static, hard-coded in rabbit registry. Dinamically searching for new classes using Erlang attributes allows for plugins to add new classes and make use of the registry.
i.e. shovel protocols can benefit of this feature


michaelklishin:
Backporting to `v4.2.x` because this code path cannot affect anyone except for custom plugins, and the affected custom plugins must have found a way to work around this long existing `rabbit_registry` limitation one way or another.

So, no one should hit this code path for a while because this is a new low-level extension point.

Per discussion with @dcorbacho.

PR #14902, opened 2025-11-05
rabbit_db: Eliminate the `delete_queue` Khepri transaction

dumbbell:
… by using `keep_while` conditions on bindings and auto-delete exchanges.

## Why

The `delete_queue` transaction's anonymous function has to be extracted by Horus, like any Khepri transaction. This is an expensive operation, but Horus uses caching to avoid most work after the first extraction.

Unfortunately, even with this caching, this transaction is still very expensive when there are massive simultaneous deletions of queues. For instance when the queues' lifetime is linked to that of many clients, and all these clients get disconnected at once.

## How

This patch removes the transaction entirely. Instead, it uses `keep_while` conditions on bindings and auto-delete exchanges to let Khepri handle the deletion of semantically related tree nodes. RabbitMQ just has to make a simle "delete this queue" command.

To maintain backward compatibility, we introduce the `replaced_delete_queue_transaction` feature flag. Its `enable` callback will take care of rewriting all bindings and auto-delete exchanges record in the store to add the new `keep_while` conditions.

The binding deletion transaction is also simplified because it benefits from the `keep_while` conditions. We may be able to replace this transaction is the future as well in the same manner.

### DIFF
--- a/deps/rabbit/src/rabbit_core_ff.erl
+++ b/deps/rabbit/src/rabbit_core_ff.erl
@@ -232,3 +232,15 @@
       stability     => stable,
       depends_on    => ['rabbitmq_4.3.0']
      }}).
+
+-rabbit_feature_flag(
+   {tie_binding_to_dest_with_keep_while_cond,
+    #{desc          =>
+      "Use keep_while condition to tie a binding record lifetime to its "
+      "destination record in Khepri",
+      stability     => stable,
+      depends_on    => ['rabbitmq_4.3.0'],
+      callbacks     => #{enable =>
+                         {rabbit_db_queue,
+                          tie_binding_to_dest_with_keep_while_cond_enable}}
+     }}).
--- a/deps/rabbit/src/rabbit_db_queue.erl
+++ b/deps/rabbit/src/rabbit_db_queue.erl
@@ -8,6 +8,7 @@
 -module(rabbit_db_queue).
 
 -include_lib("khepri/include/khepri.hrl").
+-include_lib("rabbit_common/include/logging.hrl").
 -include_lib("rabbit_common/include/rabbit.hrl").
 -include("amqqueue.hrl").
 
@@ -62,6 +63,9 @@
          update_in_khepri_tx/2,
          get_in_khepri_tx/1
         ]).
+
+%% Used for the `tie_binding_to_dest_with_keep_while_cond' feature flag.
+-export([tie_binding_to_dest_with_keep_while_cond_enable/1]).
 
 %% For testing
 -export([clear/0]).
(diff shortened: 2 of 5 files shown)


the-mikedavis:
I was looking into eliminating the transaction for queue deletion a while back but I didn't get the changes over the finish line. I just pushed up the branch I was working on: https://github.com/rabbitmq/rabbitmq-server/tree/md/khepri-q-delete-command. It's quite old but maybe it has some interesting leads for this branch. I also looked into using stored procedures where we use transactions now - wouldn't help us with the BEAM file changes in OTP 28 but it avoided some performance impacts. The code-server work for transactions is expensive but also the size of transaction commands can be quite large (the compiled module binary + env) and we could do a lot less work with commands + keep-while conditions. But, if I am remembering correctly, we would need some sort of migration for existing Khepri DBs to transition from transactions to keep-while and regular commands so that they could add the keep-while conditions.

dumbbell:
> But, if I am remembering correctly, we would need some sort of migration for existing Khepri DBs to transition from transactions to keep-while and regular commands so that they could add the keep-while conditions.

Yes, exactly. This patch is incomplete on purpose (the needed feature flag is a simple hard-coded `true` for now): I just want to validate the solution first, because updating all bindings and auto-delete exchanges is non-trivial.

Thank you very much for sharing your work! I will take a look at it :-)

dumbbell:
This pull request is ready for review as it now takes care of the upgrade of cluster as well with the `replaced_delete_queue_transaction` feature flag.

@the-mikedavis: I looked at your branch and incorporated your changes in `rabbit_db_binding:delete/2`. I used a different approach to support mixed-version clusters, based on a feature flag. Could you please take a look at the result?

deps/rabbit/src/rabbit_core_ff.erl
```diff
@@ -232,3 +232,12 @@
       stability     => stable,
       depends_on    => ['rabbitmq_4.3.0']
      }}).
+
+-rabbit_feature_flag(
+   {replaced_delete_queue_transaction,
```
the-mikedavis: Minor naming nit: I think feature flags usually are not in the past tense. I think `replace_delete_queue_transaction` sounds good for this

deps/rabbit/src/rabbit_db_queue.erl
```diff
@@ -978,6 +1002,71 @@ list_with_possible_retry(Fun) ->
...
     end.
 
+replaced_delete_queue_transaction_enable(#{feature_name := FeatureName}) ->
+    ?LOG_DEBUG(
+       "Feature flag `~s`: send transaction to add `keep_while` conditions",
+       [FeatureName],
+       #{domain => ?RMQLOG_DOMAIN_DB}),
+    Ret = rabbit_khepri:transaction(
+            fun() ->
+                    %% Auto-detele exchange.
```
the-mikedavis:
```suggestion
                    %% Auto-delete exchange.
```

dumbbell: Fixed, thanks!

deps/rabbit/src/rabbit_core_ff.erl
dumbbell: I used "replaced" as an adjective here, but it may not be correct english. Anyway, it’s unclear and the feature flag name should probably reflect what’s new, not what existed before.

What about "tie_binding_to_dest_with_keep_while_cond"? Almost a novel, but hopefully clearer.

the-mikedavis: I don't mind longer names, that seems clearer to me. I was in favor of [_Qyrad The Taker Of Life_](https://github.com/rabbitmq/khepri/discussions/60) way back when, maybe we can use that here :smile:

dumbbell: Hahaha :-)

I went with my suggestion for now. I’m marking this as resolved. Feel free to reopen it!

dumbbell:
@the-mikedavis: If you don’t have any more comments, are you ok if I mark this branch as ready for review and merge it?

dumbbell:
Thank you! This is the case when many MQTT clients are disconnected simultaneously for instance. This might be what you are referring to.

PR #14925, opened 2025-11-10
Plugins: add a way to mark plugins as bundling all deps

michaelklishin:
A plugin distributed as an .ez file might choose to include all dependencies into its "fat .ez".

For such plugins, the `ensure_dependencies/1` validation does not make as much sense, so let's introduce a way to mark such plugins using their .app file metadata, and skip the dependency list validation for them.


michaelklishin:
Closing because this change and follow-up changes are still not enough without revisiting the "one dependency, one app, one code path directory" assumption the `rabbit_plugins` and `rabbit` modules have.

PR #14961, opened 2025-11-15
rabbit_fifo_dlx_worker: Info-log when pending messages become routable

the-mikedavis:
The at-least-once dead lettering feature for QQs can cause a pile-up of dead-lettered messages in a QQ when the dead-letter routing topology leads to no viable dead-letter queues. When this happens the dead-letter worker logs at warning level. When you adjust the routing topology so that pending messages can be delivered we should log once at info level as well. This way you can confirm in the logs that you've successfully fixed your routing topology. Otherwise you would only be able to know by seeing the message stats change for dead-lettered messages in the source QQ and available messages in the destination dead-letter queue.

Discussed in https://github.com/rabbitmq/rabbitmq-server/discussions/14494
There's more we could do to make unroutable messages more visible, but this is some nice low-hanging fruit.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo_dlx_worker.erl
+++ b/deps/rabbit/src/rabbit_fifo_dlx_worker.erl
@@ -304,7 +304,8 @@ lookup_dlx(#state{exchange_ref = DLXRef} = State0) ->
             State = log_missing_dlx_once(State0),
             {not_found, State};
         {ok, X} ->
-            {X, State0}
+            State = clear_log_missing_dlx_once(State0),
+            {X, State}
     end.
 
 -spec forward(mc:state(), non_neg_integer(), rabbit_amqqueue:name(),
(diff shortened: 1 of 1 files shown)


deps/rabbit/src/rabbit_fifo_dlx_worker.erl
```diff
@@ -513,7 +515,7 @@ redeliver0(#pending{delivery = Msg0,
             State1 = log_cycles(Cycles, DLRKeys, State0),
             case RouteToQs of
                 [] ->
-                    State1;
+                    log_no_route_once(State1);
```
the-mikedavis, line 518: If you delete your DLX and then try to forward a dead-letter message, you'll get the warning about the DLX not existing. If you then create the DLX but don't yet have a binding that makes the route to the DLQ viable, we previously weren't logging the no-route warning. I think the old code was assuming that, because we're redelivering in this function, we would've already warned about no-route.

ansd: good spot, thanks!

PR #14966, opened 2025-11-18
rabbitmq_tracing: explicitly specify UTF-8 encoding for served trace files

### LINKED ISSUE #13952: RABBITMQ_TRACING: Open tracing log file with charset='UTF-8' in browser

We are using `rabbitmq_tracing` plugin and faced an issue - when opening log.file from `admin->tracing->trace log files`  we see bad enconding for cyrillic  symbols cause no charset is set by default.

Maybe you should consider adding js function for opening file in new window:  
in tracing.js add ( https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_tracing/priv/www/js/tracing.js )
```js
function loadTextFile(node, file) {
  fetch('/api/trace-files/node/' + node + '/' + file)
    .then(response => response.text())
    .then(text => {
      const win = window.open();
      win.document.write('<pre style="word-wrap: break-word; white-space: pre-wrap;" >' + text + '</pre>');
    });
}
```

and in traces.ejs ( https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_tracing/priv/www/js/tmpl/traces.ejs#L94 )  

```js
 <td><a onclick=loadTextFile(<%= node.name %>, <%= file.name %>) ><%= file.name %></a></td>
```

_No response_

### Additional context

_No response_

bas0N:
Set charset=utf-8 in Content-Type header to ensure proper display of Cyrillic and other Unicode characters in browsers.

Fixes #13952

## Proposed Changes

This PR fixes incorrect encoding display for Cyrillic and other Unicode characters when viewing trace log files through the RabbitMQ management UI. The issue was that browsers were misinterpreting UTF-8 encoded content as Latin-1 when no charset was specified in the `Content-Type` header.

**Why this change:**

- Trace files are written in UTF-8 by `rabbit_tracing_consumer` (using `io_lib:format` and `rabbit_json:encode`)
- Without an explicit charset parameter, browsers default to Latin-1/ISO-8859-1, causing Cyrillic characters to display as garbled text (e.g., `Привет мир` → `ÐŸÑ€Ð¸Ð²ÐµÑ‚ Ð¼Ð¸Ñ€`)
- Adding `charset=utf-8` to the `Content-Type` header ensures browsers correctly interpret the file content

**Solution chosen:**
The original issue suggested a client-side JavaScript solution, but we implemented a server-side fix instead because:

- Server-side solution provides a proper URL in the browser (the JS solution resulted in `about:blank`)
- More reliable and standard approach (charset should be specified in HTTP headers)
- Simpler and cleaner - no client-side workarounds needed
- Follows HTTP best practices (RFC 7231)

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #13952)
- [ ] Build system and/or CI

## Checklist

## Further Comments

**Changes made:**

- Modified `deps/rabbitmq_tracing/src/rabbit_tracing_wm_file.erl` to set `Content-Type: text/plain; charset=utf-8` header in `serve/2`
- Added test `trace_file_content_type_test` in `deps/rabbitmq_tracing/test/rabbit_tracing_SUITE.erl` to verify the Content-Type header includes charset

**Why this is safe:**

- Content negotiation remains unchanged (still accepts `text/plain`)
- Charset is optional metadata per RFC 7231 - adding it doesn't break compatibility
- Trace files are consistently written as UTF-8 by the plugin
- Similar pattern exists elsewhere in codebase (e.g., `rabbit_mgmt_oauth_bootstrap.erl` sets `text/javascript; charset=utf-8`)

**Testing:**

- Added test case that verifies Content-Type header contains both `text/plain` and `charset=utf-8`
- Manually tested with Cyrillic characters - displays correctly in browsers
- Existing tests continue to pass

Before the fix:
Before the fix

After the fix:
After the fix

### DIFF
--- a/deps/rabbitmq_tracing/src/rabbit_tracing_wm_file.erl
+++ b/deps/rabbitmq_tracing/src/rabbit_tracing_wm_file.erl
@@ -7,7 +7,7 @@
 -module(rabbit_tracing_wm_file).
 
 -export([init/2, resource_exists/2, serve/2, content_types_provided/2,
-         is_authorized/2, allowed_methods/2, delete_resource/2]).
+         charsets_provided/2, is_authorized/2, allowed_methods/2, delete_resource/2]).
 -export([serve/1]).
 
 -include_lib("rabbitmq_management_agent/include/rabbit_mgmt_records.hrl").
@@ -19,6 +19,9 @@ init(Req, _State) ->
 
 content_types_provided(ReqData, Context) ->
    {[{<<"text/plain">>, serve}], ReqData, Context}.
+
+charsets_provided(ReqData, Context) ->
+    {[<<"utf-8">>], ReqData, Context}.
 
 allowed_methods(ReqData, Context) ->
     {[<<"HEAD">>, <<"GET">>, <<"DELETE">>], ReqData, Context}.
@@ -34,10 +37,7 @@ serve(ReqData, Context) ->
     Content = rabbit_tracing_util:apply_on_node(ReqData, Context,
                                                 rabbit_tracing_wm_file,
                                                 serve, [Name]),
-    ReqWithCharset = cowboy_req:set_resp_header(<<"content-type">>,
-                                                 <<"text/plain; charset=utf-8">>,
-                                                 ReqData),
-    {Content, ReqWithCharset, Context}.
+    {Content, ReqData, Context}.
 
 serve(Name) ->
     Path = rabbit_tracing_files:full_path(Name),


deps/rabbitmq_tracing/test/rabbit_tracing_SUITE.erl
```diff
@@ -128,6 +129,35 @@ tracing_validation_test(Config) ->
...
+    http_put(Config, "/traces/%2f/test-charset", Args, ?CREATED),
+
+    Ch = rabbit_ct_client_helpers:open_channel(Config),
+    amqp_channel:cast(Ch, #'basic.publish'{ exchange    = <<"amq.topic">>,
+                                            routing_key = <<"key">> },
+                      #amqp_msg{props   = #'P_basic'{},
+                                payload = <<"Test message">>}),
+    rabbit_ct_client_helpers:close_channel(Ch),
+
+    timer:sleep(100),
```
michaelklishin: This makes the test a flake. Please use `await_condition/1` from `rabbit_ct_helpers` instead.

deps/rabbitmq_tracing/test/rabbit_tracing_SUITE.erl
```diff
@@ -128,6 +129,35 @@ tracing_validation_test(Config) ->
...
+    Args = #{format  => <<"text">>,
+             pattern => <<"#">>},
+    http_put(Config, "/traces/%2f/test-charset", Args, ?CREATED),
+
+    Ch = rabbit_ct_client_helpers:open_channel(Config),
+    amqp_channel:cast(Ch, #'basic.publish'{ exchange    = <<"amq.topic">>,
+                                            routing_key = <<"key">> },
+                      #amqp_msg{props   = #'P_basic'{},
+                                payload = <<"Test message">>}),
+    rabbit_ct_client_helpers:close_channel(Ch),
```
michaelklishin: You are using a managed (by RabbitMQ CT helpers) channel. You don't have to close these explicitly, although there's no downside to doing that.

deps/rabbitmq_tracing/test/rabbit_tracing_SUITE.erl
```diff
@@ -128,6 +129,35 @@ tracing_validation_test(Config) ->
...
+
+    timer:sleep(100),
+
+    http_delete(Config, "/traces/%2f/test-charset", ?NO_CONTENT),
+    {ok, {{_HTTP, 200, _}, Headers, _Body}} =
```
michaelklishin: I understand this `rabbitmq_tracing` is very rarely updated and is years behind `rabbitmq_management` in terms of HTTP API test helpers but this very module uses `http_get/5`, why not use it in/for this assertion?

deps/rabbitmq_tracing/src/rabbit_tracing_wm_file.erl
```diff
@@ -34,7 +34,10 @@ serve(ReqData, Context) ->
     Content = rabbit_tracing_util:apply_on_node(ReqData, Context,
                                                 rabbit_tracing_wm_file,
                                                 serve, [Name]),
-    {Content, ReqData, Context}.
+    ReqWithCharset = cowboy_req:set_resp_header(<<"content-type">>,
+                                                 <<"text/plain; charset=utf-8">>,
```
lhoguin: You need to implement the `charsets_provided` callback instead as charsets are negotiated.

(CloudFlare is currently failing so I can't immediately provide a link to the documentation.)

deps/rabbitmq_tracing/test/rabbit_tracing_SUITE.erl
bas0N: Good point! Replaced `timer:sleep(100)` with `rabbit_ct_helpers:await_condition/1`

deps/rabbitmq_tracing/test/rabbit_tracing_SUITE.erl
bas0N: Removed the `close_channel` calls

deps/rabbitmq_tracing/test/rabbit_tracing_SUITE.erl
bas0N: Created `http_get_headers/5` helper function that follows the same pattern as `http_get/5` and `http_get_raw/5`it calls `req` internally, validates the status code with `assert_code`, and returns headers

deps/rabbitmq_tracing/src/rabbit_tracing_wm_file.erl
bas0N: Implemented `charsets_provided/2` as requested. Found the docs now that Cloudflare is back up ;) Cowboy handles charset negotiation automatically, so charset=utf-8 is added to the text/plain responses

michaelklishin:
Thank you for contributing, @bas0N.

PR #14967, opened 2025-11-18
support amqp10 message without body

thorhalvor:
## Proposed Changes

Currently RabbitMQ-server does not support an AMQP 1.0 message with empty/undefined body.
My understanding of the AMQP 1.0 protocol is that this is perfectly valid case.
Many systems sends a message with properties only. They _could_ set an empty array as Body, but that need changes in sometimes hundreds of client-applications.

This change will accept if Body is 'undefined'.

I used AmqpNetLite library and published a message where BodySection is not set

        var message = new Message
        {
            // BodySection = <notset>
            Properties = new Properties()
        };
and the error occured. 

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

### DIFF
--- a/deps/rabbit/src/mc_amqp.erl
+++ b/deps/rabbit/src/mc_amqp.erl
@@ -635,10 +635,7 @@ msg_body_decoded([#'v1_0.footer'{content = FC} | Rem], Msg) ->
     msg_body_decoded(Rem, Msg#msg_body_decoded{footer = FC}).
 
 msg_body_encoded([], _Payload, Msg) ->
-    %% handle empty body
-    Msg;
-msg_body_encoded(undefined, _Payload, Msg) ->
-    %% treat undefined body the same as [].
+    %% handle empty or undefined body
     Msg;
 msg_body_encoded([#'v1_0.header'{} = H | Rem], Payload, Msg) ->
     msg_body_encoded(Rem, Payload, Msg#msg_body_encoded{header = H});


kjnilsson:
@thorhalvor please can you refer to the section of the spec where you got this interpretation? I am not sure it is correct if you look at §3.2 it reads as if only the body _is_ required.

Altogether a message consistsof the following sections:
• Zero or one header.
• Zero or one delivery-annotations.
• Zero or one message-annotations.
• Zero or one properties.
• Zero or one application-properties.
• The body consists of **one** of the following three choices: **one or more** data sections, **one or more** amqp-sequence
sections, or a **single** amqp-value section.
• Zero or one footer.

kjnilsson:
I think this PR requires further discussions (with @ansd ) and some research. I read the spec as if the body is required but it isn't completely unambiguous and allowing any non empty bare message sounds like a reasonable request.

thorhalvor:
When re-reading spec I more or less agree with you @kjnilsson. 
However, I still think this is something that should be added into the core

- if not, rabbitmq-server will have a breaking-change from 3.x to 4.x.
- sending a message without body "sounds fair enough". without faking a empty array
- I have tested sending the same message missing BodySection to Azure ServiceBus. **it is accepted, no errors**
- looks like the empty array is returned from amqp10_framing:decode_bin. if it is decided to be a protocol-error, maybe it should be stopped here? in msg_body_encoded its more a folllow-up error

Awaiting response after you and the team have discussed further. 

tnx!

kjnilsson:
> * I have tested sending the same message missing BodySection to Azure ServiceBus. **it as accepted, no errors**

This is a good data point - thanks for testing this.

michaelklishin:
The empty body case is valid in both AMQP 1.0 and AMQP 0-9-1. I wouldn't say that it is a common case but if we do not support empty message bodies for AMQP 1.0, then we should at least consider it.

If Azure accepts empty body messages, that's even better. Deviating a bit from the spec is something that protocol implementations almost inevitably have to consider at some point. Open source project maintainers rarely have the time and the patience to try to change the AMQP 1.0 spec over something like this.

ansd:
The AMQP spec is very clear that at least one application-data (body) section is required. That section may denote an empty body though. Therefore, RabbitMQ 4.x requires a body section.
I documented the breaking change and made RabbitMQ provide a better error message in https://github.com/rabbitmq/rabbitmq-server/pull/15048 if a client sends a message without a body section.

> if not, rabbitmq-server will have a breaking-change from 3.x to 4.x.

Yes, RabbitMQ 4.0 was the perfect place to include this breaking change.

> sending a message without body "sounds fair enough"

The AMQP spec prohibits this.

> I have tested sending the same message missing BodySection to Azure ServiceBus. it is accepted, no errors

Just because other brokers accept non-compliant AMQP doesn't mean RabbitMQ should accept non-compliant AMQP.

PR #14979, opened 2025-11-20
4.3: Additional cleanup work in and around CQs

lhoguin:
This branch continues cleaning up CQs.

Intent in this branch WAS to add at least `delivery-count` but there was too much cleanup remaining. A future branch will implement `delivery-count`.


michaelklishin:
Yay, the file handle cache is no more!

PR #15020, opened 2025-11-26
4.3: Implement AMQP-1.0 delivery-count for CQs

lhoguin:
Please tell me if there are other tests that need to be updated.

### DIFF
--- a/deps/rabbit/test/amqp_client_SUITE.erl
+++ b/deps/rabbit/test/amqp_client_SUITE.erl
@@ -388,23 +388,26 @@ delivery_count(QType, Config) ->
                    first_acquirer := true},
                  amqp10_msg:headers(Msg1)),
 
-    %% delivery-failed=true in modified coutcome must increment the delivery-count
     %% delivery-failed=false in modified coutcome must not increment the delivery-count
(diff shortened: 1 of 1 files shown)


deps/rabbit/src/rabbit_amqqueue_process.erl
```diff
@@ -1527,13 +1532,21 @@ handle_cast({deliver,
...
+handle_cast({reject, false, AckTags, ChPid}, State) ->
+    handle_cast({discard, AckTags, ChPid}, State);
+
+handle_cast({complete, AckTags, ChPid}, State) ->
     noreply(ack(AckTags, ChPid, State));
 
-handle_cast({reject, true,  AckTags, ChPid}, State) ->
-    noreply(requeue(AckTags, ChPid, State));
+handle_cast({requeue, AckTags, ChPid}, State) ->
+    noreply(requeue(AckTags, true, ChPid, State));
```
ansd: This boolean looks wrong to me because the `released` outcome must not increment the delivery-count.

ansd: I pushed a test case which currently fails due to this issue.

ansd:
The `rejected` outcome must increment the delivery-count, which isn't done in this PR.

PR #15021, opened 2025-11-26
Make it possible to shrink only a subset of quorum queues by providing a naming pattern

Ayanda-D:
## Proposed Changes

Hello Rabbit Team! 👋 

These changes extend the QQ force shrink operations to use regx for queue names on shrink (globally or per vhost). There cases where in the event of Quorum Queues being in a bad state (e.g no leader or bad multiple partitions), only a set need be shrunk on emergency disaster recovery operations, to prevent and rescue total message loss before queue is either re-grown or permanently deleted and re-declared. The extension (to the existing operations) allows the following:

```
ctl eval 'rabbit_quorum_queue:force_all_queues_shrink_member_to_current_member( QQMatchSpec ).'

ctl eval 'rabbit_quorum_queue:force_vhost_queues_shrink_member_to_current_member( VHost, QQMatchSpec ).'
```

Default will match all queues as existing behaviour. This change will help rescue only a specific set of queues without necessarily shrinking everything (which can take a longtime on heavily loaded brokers).  We are really keen on having this available for recovery procedures to help prevent message loss. Please take a look, thanks!

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -2108,25 +2108,34 @@ force_shrink_member_to_current_member(VHost, Name) ->
     end.
 
 force_vhost_queues_shrink_member_to_current_member(VHost) when is_binary(VHost) ->
-    force_vhost_queues_shrink_member_to_current_member(VHost, <<".*">>).
+    ?LOG_WARNING("Shrinking all quorum queues in vhost '~ts' to a single node: ~ts",
+        [VHost, node()]),
+    ListQQFun = fun() -> rabbit_amqqueue:list(VHost) end,
+    force_all_queues_shrink_member_to_current_member(ListQQFun, _MatchFun = fun(_) -> true end).
 
 force_vhost_queues_shrink_member_to_current_member(VHost, QueueSpec)
   when is_binary(VHost), is_binary(QueueSpec) ->
     ?LOG_WARNING("Shrinking all quorum queues matching '~ts' in vhost '~ts' to a single node: ~ts",
         [QueueSpec, VHost, node()]),
-    ListQQs = fun() -> rabbit_amqqueue:list(VHost) end,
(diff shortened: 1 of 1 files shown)


michaelklishin:
Some members of the core team cannot be convinced that "mass shrinking" of replicated queues is a good idea.

This PR like can be reimplemented as a custom plugin, both the CLI part and the QQ-related functions that run on RabbitMQ nodes.

Ayanda-D:
@michaelklishin hello! 👋 i'm not sure if the purpose of these changes has been understood? Our current problem is the "mass shrinking" which we cant avoid when using these default helper shrink functions provided by rabbitmq.

> Some members of the core team cannot be convinced that "mass shrinking" of replicated queues is a good idea.

This change is actually helping reduce "mass shrinking", by adding extra clauses to reduce the number of queues shrunk (matching a smaller set, instead of just shrinking "everything"). Without these changes, the shrink all will shrink everything across the broker/cluster with no option to reduce the QQ blast radius for the user.

> This PR like can be reimplemented as a custom plugin

We're happy to implement things via plugins. But our current problem is the "mass shrinking" which we cant avoid from the broker. Here, we are providing a means to reduce "mass shrinking" of everything on the broker (current shrink all ends up shrinking even queues which dont need to be shrunk when recovering from disaster).

michaelklishin (approved):

michaelklishin:
One final note: together #13873 #15166 with #grow these public API extensions allow a custom plugin to

1. Shrink a subset of queues (maybe even just one queue) down to 1 replica
2. Trigger a leader election for those queues (that are virtually guaranteed to succeed with one replica)
3. Re-grow the number of replicas to 3 (or 5, or another desired odd number) in a single operation

Exposing these new options to `rabbitmq-queues shrink` is also possible if needed, and reaching consensus around a CLI extension should be lightning fast compared to how long this PR and #13873 #15166 have taken.

PR #15025, opened 2025-11-27
rabbit_khepri: Fix topic binding deletion leak

### LINKED ISSUE #15024: Memory utilisation growing with high-churn of routing-keys
### Describe the bug

Hello,

We have recently upgraded our RabbitMQ cluster from RabbitMQ 3.7.5 (Erlang 20.3.4) to RabbitMQ 4.1.1 (Erlang 27.3.4).

Since the upgrade, we have noticed that RabbitMQ's memory utilisation is continually growing.

Upon investigation, we have noted:

- 'rabbitmq-diagnostics memory_breakdown' indicates that 'metadata_store_ets' consumes most of the memory: 1.1684 gb (59.11%)
- 'rabbitmq-diagnostics observer' indicates the 'rabbit_khepri_topic_trie' ETS table consumes 1.0855 GB

My suspicion is that 'rabbit_khepri_topic_trie' is continually growing larger due to one of our sub-systems utilising a high-churn of routing-keys on a particular exchange.

This suspicion seems to be validated by running:

- rabbitmqctl eval 'ets:tab2list(rabbit_khepri_topic_trie).'

which indicates that most of the rows in this table are relating to exchange of the sub-system in question.

We have not made any changes to this sub-system other than upgrading RabbitMQ. Prior to the upgrade, RabbitMQ memory consumption was not continually growing.

I note that RabbitMQ 4.1.1 is using Khepri for metadata storage instead of Mnesia used in 3.7.5.

It seems to me that Khepri metadata storage is not coping with the high-churn of routing-keys like Mnesia did.

To give an idea of routing-key churn:

We have ~2500 routing keys currently applied to the exchange, but the size of the 'rabbit_khepri_topic_trie' table is '4,335,432' rows.

### Reproduction steps

1. Create high-churn of different routing-keys on an exchange.
2. Observe rabbit_khepri_topic_trie ETS table continually grow.

### Expected behavior

I would expect memory utilisation to correlate with the number of routing-keys currently applied to an exchange, rather than grow indefinitely.

### Additional context

_No response_

mkuratczyk:
## Why

We use a Khepri projection to compute a graph for bindings that have a topic exchange as their source. This allows more efficient queries during routing. This graph is not stored in Khepri, only in the projection ETS table.

When a binding is deleted, we need to clean up the graph. However, the pattern used to match the trie edges to delete was incorrect, leading to "orphaned" trie edges. The accumulation of these leftovers caused a memory leak.

## How

The pattern was fixed to correctly match the appropriate trie edges.

However, this fix alone is effective for new deployments of RabbitMQ only, when the projection function is registered for the first time. We also need to handle the update of already registered projections in existing clusters.

To achieve that, first, we renamed the projection from `rabbit_khepri_topic_trie` to `rabbit_khepri_topic_trie_v2` to distinguish the bad one and the good one. Any updated RabbitMQ nodes in an existing cluster will use this new projection. Other existing out-of-date nodes will continue to use the old projection. Because both projections continue to exist, the cluster will still be affected by the memory leak.

Then, each node will verify on startup if all other cluster members support the new projection. If that is the case, they will unregister the old projection. Therefore, once all nodes in a cluster are up-to-date and use the new projection, the old one will go away and the leaked memory will be reclaimed.

This startup check could have been made simpler with a feature flag. We decided to go with a custom check in case a user would try to upgrade from a 4.1.x release that has the fix to a 4.2.x release that does not for instance. A feature flag would have prevented that upgrade path.

Fixes #15024.

### DIFF
--- a/deps/rabbit/src/rabbit_db_topic_exchange.erl
+++ b/deps/rabbit/src/rabbit_db_topic_exchange.erl
@@ -26,7 +26,7 @@
 -define(MNESIA_NODE_TABLE, rabbit_topic_trie_node).
 -define(MNESIA_EDGE_TABLE, rabbit_topic_trie_edge).
 -define(MNESIA_BINDING_TABLE, rabbit_topic_trie_binding).
--define(KHEPRI_PROJECTION, rabbit_khepri_topic_trie).
+-define(KHEPRI_PROJECTION, rabbit_khepri_topic_trie_v2).
 
 -type match_result() :: [rabbit_types:binding_destination() |
                          {rabbit_amqqueue:name(), rabbit_types:binding_key()}].
--- a/deps/rabbit/src/rabbit_khepri.erl
+++ b/deps/rabbit/src/rabbit_khepri.erl
@@ -175,6 +175,9 @@
          get_feature_state/0, get_feature_state/1,
          handle_fallback/1]).
 
+%% Called remotely to handle unregistration of old projections.
+-export([supports_rabbit_khepri_topic_trie_v2/0]).
+
 -ifdef(TEST).
 -export([register_projections/0,
          force_metadata_store/1,
@@ -1541,7 +1544,7 @@ projection_fun_for_sets(MapFun) ->
     end.
 
 register_rabbit_topic_graph_projection() ->
-    Name = rabbit_khepri_topic_trie,
+    Name = rabbit_khepri_topic_trie_v2,
     %% This projection calls some external functions which are disallowed by
     %% Horus because they interact with global or random state. We explicitly
     %% allow them here for performance reasons.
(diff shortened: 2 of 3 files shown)


ghost:
Thanks very much for this.

I note the 'backport-v4.2.x' label. Will this also be back-ported for v4.1.x?

dumbbell:
This will only fix the problem for new deployments, not existing ones. The reason is that the projection function is stored in Khepri the first time it is registered. This is mandatory to ensure the behaviour is consistent across a cluster.

The current code in `rabbit_khepri` doesn’t try to manage updates of this function unfortunately.

michaelklishin:
@dumbbell @mkuratczyk so, should we merge this as a first step and then see what can be done to work around the "persistent" nature of projections in Khepri?

dumbbell:
I think we should have a way forward for upgrades first.

mkuratczyk:
@Mergifyio backport v4.1.x

PR #15028, opened 2025-11-27
Prometheus endpoint: faster scraping

mkuratczyk:
With these changes, scraping /metrics/per-object
on a system with thousands of queues is 20-30% faster.


lukebakken (approved):
🚀

PR #15074, opened 2025-12-05
Add /users/$user/queues endpoint

lhoguin:
It can be used to list queues by owner.

Let's see if that's good enough before documenting the API/adding to the builtin rabbitmqadmin (should it be added there?).

### DIFF
--- a/deps/rabbitmq_management/priv/www/api/index.html
+++ b/deps/rabbitmq_management/priv/www/api/index.html
@@ -929,6 +929,14 @@ or:
         <td></td>
         <td></td>
         <td></td>
+        <td class="path">/api/users/<i>user</i>/queues</td>
+        <td>A list of all queues owned by a given user.</td>
+      </tr>
+      <tr>
+        <td>X</td>
+        <td></td>
+        <td></td>
+        <td></td>
         <td class="path">/api/users/<i>user</i>/topic-permissions</td>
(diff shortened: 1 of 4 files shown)


ansd:
Awaiting user feedback

lhoguin:
Should we document it?

ansd:
Yes, let's include the new endpoint in HTTP API reference: https://rawcdn.githack.com/rabbitmq/rabbitmq-server/v4.2.1/deps/rabbitmq_management/priv/www/api/index.html

ansd:
> adding to the builtin rabbitmqadmin (should it be added there?)

I don't think that's necessary for now.

michaelklishin:
`rabbitmqadmin` v1 is no longer maintained. We should not add anything to it.

You can submit a PR for `rabbitmqadmin` v2 but let's ship this PR first :)

michaelklishin:
We can expose it as `queues list_declared_by --username` or something like that in [`rabbitmqadmin` v2](https://github.com/rabbitmq/rabbitmqadmin-ng).

I would avoid the word "owned" because that term is pretty much only describes exclusive queue ownership in our docs, and this endpoint filters the queues by their declaring user.

michaelklishin:
https://github.com/rabbitmq/rabbitmq-website/pull/2418

Issue #1443, opened 2017-12-07
Consider an option to dead-letter rejected publishes


hairyhum:
Since version 3.7.0 RabbtitMQ queues have an option to reject publishes if a queue reaches it's limit.
Currently rejected messages are not dead-lettered, unlike messages dropped from the queue head.
Dead-lettering them will help if publisher confirms are not enabled or if the client does not know what to do with a rejected message.

WGH-:
> Dead-lettering them will help if publisher confirms are not enabled or if the client does not know what to do with a rejected message.

To elaborate, publisher confirms don't give the information which queue overflowed. So you can't re-publish the message to that exact queue, only re-publish the message as a whole, which would lead to duplicates in other queues.

When message is dead-lettered, however, it includes the exact queue where they were dropped, so you can republish it back to exact same queue using the default direct exchange.

This trick can be used to implement sort of publisher flow control: once it notices own messages being dead-lettered, it can slow down producing until all dead-lettered message are successfully republished.

You can employ this trick with `drop-head`, but since messages are dropped from the front, you'll obviously end up with reordered messages. 

Moreover, using it for flow control is complicated by the fact that one publisher may get its own messages back even though it's some other publisher who overflowed the queue and needs to pause.

michaelklishin:
@hairyhum @kjnilsson @dcorbacho thoughts?

ansd:
> To elaborate, publisher confirms don't give the information which queue overflowed.

We could include the name of the queue that rejected the message (and even the reason why the message was rejected) in the `error` field of the `Rejected` outcome.

Done in  https://github.com/rabbitmq/rabbitmq-server/pull/15075

PR #15075, opened 2025-12-05
Provide queue name and reason to publisher for rejected messages

ansd:
## What?
1. This commit adds the name of the queue that rejected a message as well as the reason why the message was rejected in the Rejected outcome for AMQP 1.0.
2. For MQTT 5.0 publishers, the reason is provided in the PUBACK packet.

 ## Why?
It may be helpful for publishers to know which queue out of multilple target queues rejected the message. One such use case is described in https://github.com/rabbitmq/rabbitmq-server/issues/1443 It may also be helpful to know for publishers whether the given target queue rejected the message because its maximum queue length was reached or because the target queue happens to be unavailable.

 ## How?
RabbitMQ will include the UTF-8 encoded queue name (rather than an AMQP address) as well as the reason in the `info` field of the Rejected outcome's `error` field:
* `queue: <queue name>`
* `reason: maxlen | unavailable`

For MQTT 5.0, if the queue length limit is exceeded, RabbitMQ will return the reason code for `Quota exceeded` to the publisher.

Docs PR: https://github.com/rabbitmq/rabbitmq-website/pull/2413

### DIFF
--- a/deps/rabbit/src/rabbit_amqp_session.erl
+++ b/deps/rabbit/src/rabbit_amqp_session.erl
@@ -373,7 +373,10 @@
 
           %% Queue actions that we will process later such that we can confirm and reject
           %% delivery IDs in ranges to reduce the number of DISPOSITION frames sent to the client.
-          stashed_rejected = [] :: [{rejected, rabbit_amqqueue:name(), [delivery_number(),...]}],
+          stashed_rejected = [] :: [{rejected,
+                                     rabbit_amqqueue:name(),
+                                     rabbit_queue_type:reject_reason(),
+                                     [delivery_number(),...]}],
           stashed_settled = [] :: [{settled, rabbit_amqqueue:name(), [delivery_number(),...]}],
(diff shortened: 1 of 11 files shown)


michaelklishin:
For STOMP, we can add these values via `x-*` headers on the [`NACK` frame](https://stomp.github.io/stomp-specification-1.1.html#NACK).

AMQP 0-9-1 would require a protocol extension similarly to how `basic.nack` emerged because `basic.reject` was lacking the necessary fields.

ansd:
> For STOMP, we can add these values via x-* headers on the [NACK frame](https://stomp.github.io/stomp-specification-1.1.html#NACK).

The `NACK` frame is the wrong frame because the NACK frame is sent from client to server. "`NACK` is the opposite of `ACK`. It is used to tell the server that the client did not consume the message."

Instead, the target queue and reason why RabbitMQ rejected a message would need to be included in the STOMP `ERROR` frame. That's something we can do in future as part of Native STOMP.

PR #15081, opened 2025-12-05
rabbit_quorum_queue: Shrink batches of QQs in parallel

the-mikedavis:
Shrinking a member node off of a QQ can be parallelized. The operation involves

* removing the node from the QQ's cluster membership (appending a command to the log and committing it) with `ra:remove_member/3`
* updating the metadata store to remove the member from the QQ type state with `rabbit_amqqueue:update/2`
* deleting the queue data from the node with `ra:force_delete_server/2` if the node can be reached

All of these operations are I/O bound. Updating the cluster membership and metadata store involves appending commands to those logs and replicating them. Writing commands to Ra synchronously in serial is fairly slow - sending many commands in parallel is much more efficient. By parallelizing these steps we can write larger chunks of commands to WAL(s).

`ra:force_delete_server/2` benefits from parallelizing if the node being shrunk off is no longer reachable, for example in some hardware failures. The underlying `rpc:call/4` will attempt to auto-connect to the node and this can take some time to time out. By parallelizing this, each `rpc:call/4` reuses the same underlying distribution entry and all calls fail together once the connection fails to establish.

Discussed in https://github.com/rabbitmq/rabbitmq-server/discussions/15057

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -1492,39 +1492,70 @@ delete_member(Q, Node) when ?amqqueue_is_quorum(Q) ->
       {ok, pos_integer()} | {error, pos_integer(), term()}}].
 shrink_all(Node) ->
     ?LOG_INFO("Asked to remove all quorum queue replicas from node ~ts", [Node]),
-    [begin
-         QName = amqqueue:get_name(Q),
-         ?LOG_INFO("~ts: removing member (replica) on node ~w",
-                         [rabbit_misc:rs(QName), Node]),
-         Size = length(get_nodes(Q)),
-         case delete_member(Q, Node) of
-             ok ->
-                 {QName, {ok, Size-1}};
-             {error, cluster_change_not_permitted} ->
-                 %% this could be timing related and due to a new leader just being
-                 %% elected but it's noop command not been committed yet.
-                 %% lets sleep and retry once
-                 ?LOG_INFO("~ts: failed to remove member (replica) on node ~w "
-                                 "as cluster change is not permitted. "
-                                 "retrying once in 500ms",
-                                 [rabbit_misc:rs(QName), Node]),
-                 timer:sleep(500),
-                 case delete_member(Q, Node) of
-                     ok ->
-                         {QName, {ok, Size-1}};
-                     {error, Err} ->
-                         ?LOG_WARNING("~ts: failed to remove member (replica) on node ~w, error: ~w",
-                                            [rabbit_misc:rs(QName), Node, Err]),
-                         {QName, {error, Size, Err}}
-                 end;
-             {error, Err} ->
-                 ?LOG_WARNING("~ts: failed to remove member (replica) on node ~w, error: ~w",
-                                    [rabbit_misc:rs(QName), Node, Err]),
-                 {QName, {error, Size, Err}}
-         end
-     end || Q <- rabbit_amqqueue:list(),
-            amqqueue:get_type(Q) == ?MODULE,
-            lists:member(Node, get_nodes(Q))].
+    %% This operation is bound by I/O so this default is set high.
+    Size = application:get_env(rabbit, quorum_queue_shrink_batch_size, 64),
+    Chunks = ra_lib:lists_chunk(Size, [Q || Q <- rabbit_amqqueue:list(),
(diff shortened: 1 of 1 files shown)


the-mikedavis:
With this change and the default `64` set here (just a sensible-seeming constant) I see my test in https://github.com/rabbitmq/rabbitmq-server/discussions/15057 of shrinking from 1000 QQs go from taking ~2hrs to taking 1min52sec.

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1513,40 +1513,53 @@ delete_member(Q, Node) when ?amqqueue_is_quorum(Q) ->
...
+    Size = application:get_env(rabbit, quorum_queue_shrink_batch_size, 64),
+    Chunks = ra_lib:lists_chunk(Size, [Q || Q <- rabbit_amqqueue:list(),
+                                       amqqueue:get_type(Q) == ?MODULE,
+                                       lists:member(Node, get_nodes(Q))]),
+    lists:append([begin
+                      Running = [spawn(fun() ->
+                                               Res = shrink(Node, Q),
+                                               Parent ! {self(), Res}
+                                       end) || Q <- Chunk],
+                      [receive
```
michaelklishin: We should introduce a timeout here. All timeout defaults are wrong but something like 10s or 15s won't suffer from false positives much but avoids blocking the parent process.

the-mikedavis: Yeah, I wonder if we should also `spawn_monitor` and look for `'DOWN'`s as well. I haven't seen this code crash before but it wouldn't hurt to catch exits and avoid this list comprehension getting stuck

michaelklishin: That would be even better. Both are just typical defensive techniques.

kjnilsson:
This looks fine to me, at least for now.

It would be quite possible to get much higher throughput on this and use command pipelining instead of spawning a bunch of processes just to exercise the WAL more. We'd need to add that as an option to the Ra API however.

the-mikedavis:
Ah yeah, with pipelining we could use the WAL much more efficiently. That shouldn't be too bad to add to Ra - just a new function in `ra` that would use `ra_server_proc:cast_command/3`, right? Once mnesia is gone we could use Khepri async commands for the metadata store updates so both of those parts could be done with pipelining.

I'm actually more worried about the `ra:force_delete_server/2` part since that step can take a while (7 seconds) if the connection to the node times out. An easy way around that would be adding a function in `rabbit_quorum_queue` to call `ra:force_delete_server/2` on all queues after the membership and metadata store parts are done. Then it would just be one RPC call which could time out.

In the meantime making this parallel seems like an easy improvement since we can continue using the `delete_member/2` helper. But in the long run we should definitely use pipelining instead :+1:

michaelklishin:
@kjnilsson do you have any more feedback on the updated version?

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1481,39 +1481,70 @@ delete_member(Q, Node) when ?amqqueue_is_quorum(Q) ->
...
-     end || Q <- rabbit_amqqueue:list(),
-            amqqueue:get_type(Q) == ?MODULE,
-            lists:member(Node, get_nodes(Q))].
+    %% This operation is bound by I/O so this default is set high.
+    Size = application:get_env(rabbit, quorum_queue_shrink_batch_size, 64),
+    Chunks = ra_lib:lists_chunk(Size, [Q || Q <- rabbit_amqqueue:list(),
+                                       amqqueue:get_type(Q) == ?MODULE,
+                                       lists:member(Node, get_nodes(Q))]),
+    Parent = self(),
+    lists:flatten([begin
```
kjnilsson, line 1501: could you use `ra_lib:partition_parallel/2|3` here?

the-mikedavis: Not directly: `shrink/2` returns the current or updated size of the cluster and that's used in the output of the `rabbitmq-queues shrink` command. With `ra_lib:partition_parallel/2` we need to return a boolean so we can't add the size info. Implementation-wise though, this looks nearly the same.

michaelklishin:
@kjnilsson if we want to get this into `4.3.0`, we need to start finalizing it now.

kjnilsson (approved):

PR #15093, opened 2025-12-09
Remove RAM node type

lhoguin:
This removes access to the functionality while leaving the code that will be removed along Mnesia. Removing the code now would be a much larger and riskier endeavor.

Recommended upgrade steps: 1. Remove or convert RAM nodes. 2. Perform the upgrade as usual.


lhoguin:
Test failure is some unrelated Docker thing that started happening today.

PR #15105, opened 2025-12-10
Introduce rabbit_data_coercion:to_boolean/1

michaelklishin:


michaelklishin:
The failure is specific to Erlang/OTP 28.3 that's shipped today.

michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/pull/15108 should address the 28.3-specific Dialyzer failure.

PR #15123, opened 2025-12-12
Add config to adjust MQTT whether disconnect when publish or subscribe non-authorized topic

iamdrq:
## Proposed Changes

Thanks all works.

In MQTT spec, not clear descipe  client whether disconnect when publish or subscribe non-authorized topic,

So implement has diffrent action:
```
HiveMQ: 

subscribe non-authorized topic:
MQTT3.1 → disconnect 
MQTT3.1.1 → SUBACK failure no disconnect
MQTT5 → reason codes no disconnect

publish non-authorized topic:
disconnect client (but MQTT5 can send reason code then disconnect)
```
```
EMQX:

subscribe non-authorized topic:
SUBACK deny (MQTT5 not authorized, MQTT3 fail) no disconnect

publish non-authorized topic:
Reject publish, no disconnect by default
```
```
Rabbitmq MQTT:

subscribe non-authorized topic:
disconnect

publish non-authorized topic:
disconnect
```

In AMQP spec, disconnect client when touch non-authorized topic, so Rabbitmq MQTT keep this action now.

In MQTT spec,this is not clear. 
Someone hope touch non-authorized topic disconnect and someone hope not, https://github.com/rabbitmq/rabbitmq-server/discussions/12902
So let's add config to adjust Rabbitmq MQTT whether disconnect when `publish` or `subscribe` non-authorized topic,
give control to the user.

rabbitmq.conf 
```
mqtt.ignore_unauthorized=false|true (false defualt)
```
mqtt.ignore_unauthorized=false
```
disconnect client when publish or subscribe non-authorized topic, keep Rabbitmq MQTT action now
```
mqtt.ignore_unauthorized=true
```
subscribe: 
reply SUBACK(ok,non-authorized) and send retained messages for the successfully subscribed topics
and keep connection

publish: 
MQTT v5 + QoS1: reply with PUBACK including an error reason code and keep connection.
MQTT v3 or QoS0 : drop silently and keep connection.
```

## Types of Changes

## Checklist

### DIFF
--- a/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
+++ b/deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
@@ -451,7 +451,7 @@ process_request(?SUBSCRIBE,
                 State0 = #state{cfg = #cfg{proto_ver = ProtoVer,
                                            binding_args_v2 = BindingArgsV2}}) ->
     ?LOG_DEBUG("Received a SUBSCRIBE with subscription(s) ~p", [Subscriptions]),
-    DisconnectOnUnauthorized = application:get_env(rabbitmq_mqtt, disconnect_on_unauthorized, true),
+    {ok, DisconnectOnUnauthorized} = application:get_env(?APP_NAME, disconnect_on_unauthorized),
     {ResultRev, RetainedRev, State1} =
     lists:foldl(
       fun(_Subscription, {[{error, _} = E | _] = L, R, S}) when DisconnectOnUnauthorized =:= true ->
@@ -515,7 +515,7 @@ process_request(?SUBSCRIBE,
             %% If disconnect_on_unauthorized is false, do not disconnect the client,
             %% send retained messages for the topics to which the client could successfully subscribe.
             %% Otherwise, disconnect the client, treat the subscription failure.
-            case application:get_env(rabbitmq_mqtt, disconnect_on_unauthorized, true) of
+            case DisconnectOnUnauthorized of
               false ->
                 State = send_retained_messages(lists:reverse(RetainedRev), State1),
                 {ok, State};
@@ -2280,32 +2280,19 @@ publish_to_queues_with_checks(
             end;
         {error, access_refused} ->
             %%  If disconnect_on_unauthorized is false,
-            %%  MQTT v5 and QoS1 reply with PUBACK including an error reason code and keep connection,
-            %%  MQTT v3 and QoS1 reply with PUBACK no error reason code and keep connection,
-            %%  QoS0 drop silently and keep connection.
+            %%  QoS1 reply with PUBACK and keep connection,
+            %%  QoS0,WILL_MSG drop silently and keep connection.
             %%  Otherwise, disconnect.
-            case application:get_env(rabbitmq_mqtt, disconnect_on_unauthorized, true) of
+            {ok, DisconnectOnUnauthorized} = application:get_env(?APP_NAME, disconnect_on_unauthorized),
+            case DisconnectOnUnauthorized of
                 false ->
                     case Msg#mqtt_msg.qos of
                         ?QOS_1 ->
-                            case State#state.cfg#cfg.proto_ver of
-                                ?MQTT_PROTO_V5 ->
-                                    Reply = #mqtt_packet{
-                                        fixed = #mqtt_packet_fixed{type = ?PUBACK},
-                                        variable = #mqtt_packet_puback{
-                                            packet_id = Msg#mqtt_msg.packet_id,
-                                            reason_code = ?RC_NOT_AUTHORIZED
-                                        }
-                                    },
-                                    _ = send(Reply, State);
+                            case Msg#mqtt_msg.packet_id of
+                                ?WILL_MSG_QOS_1_CORRELATION ->
+                                    ok;
                                 _ ->
-                                    Reply = #mqtt_packet{
-                                        fixed = #mqtt_packet_fixed{type = ?PUBACK},
-                                        variable = #mqtt_packet_puback{
-                                            packet_id = Msg#mqtt_msg.packet_id
-                                        }
-                                    },
-                                    _ = send(Reply, State)
+                                    send_puback(Msg#mqtt_msg.packet_id, ?RC_NOT_AUTHORIZED, State)
                             end;
                         _ ->
                             ok
--- a/deps/rabbitmq_mqtt/test/disconnect_on_unauthorized_SUITE.erl
+++ b/deps/rabbitmq_mqtt/test/disconnect_on_unauthorized_SUITE.erl
@@ -33,7 +33,8 @@ groups() ->
 test_cases() ->
     [
         publish_unauthorized_no_disconnect,
-        subscribe_unauthorized_no_disconnect
+        subscribe_unauthorized_no_disconnect,
+        unsubscribe_unauthorized_no_disconnect
     ].
 
 init_per_suite(Config0) ->
@@ -57,7 +58,7 @@ init_per_suite(Config0) ->
 
     Config2 = rabbit_ct_helpers:set_config(Config1, [{mqtt_user, User}, {mqtt_pass, Password}]),
 
-    rabbit_ct_broker_helpers:set_permissions(Config2, User, <<"/">>, <<".*">>, <<"">>, <<".*">>),
+    rabbit_ct_broker_helpers:set_permissions(Config2, User, <<"/">>, <<".*">>, <<".*">>, <<".*">>),
     ok = rabbit_ct_broker_helpers:rpc(Config2, 0,
         rabbit_auth_backend_internal, set_topic_permissions,
         [?config(mqtt_user, Config2), <<"/">>,
@@ -132,6 +133,52 @@ subscribe_unauthorized_no_disconnect(Config) ->
             ?assertEqual(?RC_NOT_AUTHORIZED, ReasonCode);
         v4 ->
             ?assertEqual(?RC_Failure, ReasonCode)
+    end,
+
+    timer:sleep(300),
+    %% Client still connected
+    ?assert(is_process_alive(C)),
+
+    ok = emqtt:disconnect(C).
+
+unsubscribe_unauthorized_no_disconnect(Config) ->
+    ok = rabbit_ct_broker_helpers:rpc(Config, 0,
+        rabbit_auth_backend_internal, set_topic_permissions,
+        [?config(mqtt_user, Config), <<"/">>,
+            <<"amq.topic">>, <<"^(topic1|topic3)$">>, <<"^(topic1|topic3)$">>, <<"acting-user">>]),
+
+    C = util:connect(
+        <<"sub_client">>,
+        Config,
+        [{username, ?config(mqtt_user, Config)},
+            {password, ?config(mqtt_pass, Config)}]),
+
+    {ok, _, _} =
+        emqtt:subscribe(
+            C,
+            {<<"topic3">>, qos0}
+        ),
+
+    ok = rabbit_ct_broker_helpers:rpc(Config, 0,
+        rabbit_auth_backend_internal, set_topic_permissions,
+        [?config(mqtt_user, Config), <<"/">>,
+            <<"amq.topic">>, <<"^topic1$">>, <<"^topic1$">>, <<"acting-user">>]),
+
+    timer:sleep(1000),
+
+    case ?config(mqtt_version, Config) of
+        v5 ->
+            {ok, _, [?RC_NOT_AUTHORIZED]} =
+                emqtt:unsubscribe(
+                    C,
+                    <<"topic3">>
+                );
+        v4 ->
+            {ok, _, _} =
+                emqtt:unsubscribe(
+                    C,
+                    <<"topic3">>
+                )
     end,
 
     timer:sleep(300),
--- a/deps/rabbitmq_mqtt/Makefile
+++ b/deps/rabbitmq_mqtt/Makefile
@@ -96,7 +96,7 @@ define ct_master.erl
 	halt(0)
 endef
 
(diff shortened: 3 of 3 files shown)


deps/rabbitmq_mqtt/priv/schema/rabbitmq_mqtt.schema
```diff
@@ -15,6 +15,14 @@
 {mapping, "mqtt.allow_anonymous", "rabbitmq_mqtt.allow_anonymous",
     [{datatype, {enum, [true, false]}}]}.
 
+%  {rabbitmq_mqtt,
+%   [%% Setting whether disconnect when client publish,subscribe to unauthorized topic.
```
michaelklishin: This is confusing, please remove this inline comment that starts after the beginning of a list literal.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -120,6 +120,17 @@
 
 -opaque state() :: #state{}.
 
+%% -------------------------
+%% Config helper
+%% -------------------------
+%% Read configuration from application env (rabbitmq.conf -> mqtt.ignore_unauthorized)
```
michaelklishin: Does this comment block really contribute anything? Was it added by an LLM?

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -120,6 +120,17 @@
...
+ignore_unauthorized() ->
+    case application:get_env(rabbitmq_mqtt, ignore_unauthorized) of
```
michaelklishin: `application:get_env/3` has been around for years.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -509,6 +521,17 @@ process_request(?SUBSCRIBE,
                                        reason_codes = lists:reverse(ReasonCodesRev)}},
     _ = send(Reply, State1),
     case hd(ResultRev) of
+        {error, access_refused} ->
+            %% ignore_unauthorized setting:
+            %%  - true: not disconnect client,send retained messages for the successfully subscribed topics.
+            %%  - false: disconnect client,treat subscribe failure as fatal and disconnect.
```
michaelklishin: "disconnect the client", "treat the subscription failure"

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -509,6 +521,17 @@ process_request(?SUBSCRIBE,
                                        reason_codes = lists:reverse(ReasonCodesRev)}},
     _ = send(Reply, State1),
     case hd(ResultRev) of
+        {error, access_refused} ->
+            %% ignore_unauthorized setting:
+            %%  - true: not disconnect client,send retained messages for the successfully subscribed topics.
```
michaelklishin: "do not disconnect the client", "for the topics to which the client could successfully subscribe"

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -2266,7 +2289,30 @@ publish_to_queues_with_checks(
                     Error
             end;
         {error, access_refused} ->
-            {error, access_refused, State}
+            %%  ignore_unauthorized setting:
+            %%  - true:
```
michaelklishin: While I'd agree that our comment format is quite varied, we definitely do not use nested lists for describing three possible scenarios each of which is explained with just one line.

This is just weird.

deps/rabbitmq_mqtt/priv/schema/rabbitmq_mqtt.schema
```diff
@@ -15,6 +15,14 @@
...
+%%
+%% {ignore_unauthorized, false},
+
+{mapping, "mqtt.ignore_unauthorized", "rabbitmq_mqtt.ignore_unauthorized",
```
michaelklishin: I have a problem with this name. Ignore authorized what specifically?

This could be `mqtt.ignore_authorization_failures` but that would not explain when exactly and the name can be misleading.

`mqtt.maintain_connection_on_authorization_failures` is slightly more specific and is the best suggestion I have.

deps/rabbitmq_mqtt/priv/schema/rabbitmq_mqtt.schema
iamdrq: Thank you very much for guiding my code and showing me how I can better contribute to this PR. I have already modified my code according to your suggestions.

deps/rabbitmq_mqtt/priv/schema/rabbitmq_mqtt.schema
ansd: I find `maintain_connection_on_authorization_failures` is too long.

Let's name it `disconnect_on_unauthorized` (defaulting to `true` since that's today's behaviour).

ansd (changes requested):
The rationale of this PR makes sense because the MQTT 5.0 spec states:
> Errors other than Malformed Packet and Protocol Errors cannot be anticipated by the sender because the receiver might have constraints which it has not communicated to the sender. A receiving Client or Server might encounter a transient error, such as a shortage of memory, that prevents successful processing of an individual MQTT Control Packet. Acknowledgment packets PUBACK, PUBREC, PUBREL, PUBCOMP, SUBACK, UNSUBACK with a Reason Code of 0x80 or greater indicate that the received packet, identified by a Packet Identifier, was in error. There are no consequences for other Sessions or other Packets flowing on the same Session.

EMQX also has a config option to toggle this disconnect on authz failure behaviour.

I find `maintain_connection_on_authorization_failures` is too long.

Let's name it `disconnect_on_unauthorized` (defaulting to `true` since that's today's behaviour).

@iamdrq tests are missing. Please add tests.

michaelklishin:
`mqtt.disconnect_on_unauthorized` sounds reasonable.

deps/rabbitmq_mqtt/priv/schema/rabbitmq_mqtt.schema
iamdrq: Thanks suggestions, I named it `disconnect_on_unauthorized` (defaulting to `true`)

michaelklishin:
@iamdrq this was not forgotten but chances are, it will now be reviewed after Jan 2.

iamdrq:
@michaelklishin Thanks for the update! That’s totally fine, I appreciate you letting me know.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -2266,7 +2279,41 @@ publish_to_queues_with_checks(
                     Error
             end;
         {error, access_refused} ->
-            {error, access_refused, State}
+            %%  If disconnect_on_unauthorized is false,
+            %%  MQTT v5 and QoS1 reply with PUBACK including an error reason code and keep connection,
+            %%  MQTT v3 and QoS1 reply with PUBACK no error reason code and keep connection,
+            %%  QoS0 drop silently and keep connection.
+            %%  Otherwise, disconnect.
+            case application:get_env(rabbitmq_mqtt, disconnect_on_unauthorized, true) of
```
ansd: You can use here the `?APP_NAME` macro.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -509,6 +511,17 @@ process_request(?SUBSCRIBE,
                                        reason_codes = lists:reverse(ReasonCodesRev)}},
     _ = send(Reply, State1),
     case hd(ResultRev) of
+        {error, access_refused} ->
+            %% If disconnect_on_unauthorized is false, do not disconnect the client,
+            %% send retained messages for the topics to which the client could successfully subscribe.
+            %% Otherwise, disconnect the client, treat the subscription failure.
+            case application:get_env(rabbitmq_mqtt, disconnect_on_unauthorized, true) of
```
ansd: You already configured the default of `true` in the `Makefile`. So, you can omit the default `true` in this line. It's better to configure the defaults for configuration in just one place instead at multiple places throughout the code base.

ansd: You can use here the `?APP_NAME` macro.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
ansd: You already configured the default of true in the Makefile. So, you can omit the default true in this line. It's better to configure the defaults for configuration in just one place instead at multiple places throughout the code base.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -451,10 +451,12 @@ process_request(?SUBSCRIBE,
                 State0 = #state{cfg = #cfg{proto_ver = ProtoVer,
                                            binding_args_v2 = BindingArgsV2}}) ->
     ?LOG_DEBUG("Received a SUBSCRIBE with subscription(s) ~p", [Subscriptions]),
+    DisconnectOnUnauthorized = application:get_env(rabbitmq_mqtt, disconnect_on_unauthorized, true),
```
ansd: You already configured the default of true in the Makefile. So, you can omit the default true in this line. It's better to configure the defaults for configuration in just one place instead at multiple places throughout the code base.

ansd: You can use here the ?APP_NAME macro.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -2266,7 +2279,41 @@ publish_to_queues_with_checks(
...
+                                    _ = send(Reply, State);
+                                _ ->
+                                    Reply = #mqtt_packet{
+                                        fixed = #mqtt_packet_fixed{type = ?PUBACK},
+                                        variable = #mqtt_packet_puback{
+                                            packet_id = Msg#mqtt_msg.packet_id
+                                        }
+                                    },
+                                    _ = send(Reply, State)
+                            end;
```
ansd: These two separate branches are not necessary. Module `rabbit_mqtt_packet` encodes the packet correctly without the reason code for MQTT v4 and v3.

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -2266,7 +2279,41 @@ publish_to_queues_with_checks(
...
+            case application:get_env(rabbitmq_mqtt, disconnect_on_unauthorized, true) of
+                false ->
+                    case Msg#mqtt_msg.qos of
+                        ?QOS_1 ->
+                            case State#state.cfg#cfg.proto_ver of
+                                ?MQTT_PROTO_V5 ->
+                                    Reply = #mqtt_packet{
+                                        fixed = #mqtt_packet_fixed{type = ?PUBACK},
+                                        variable = #mqtt_packet_puback{
+                                            packet_id = Msg#mqtt_msg.packet_id,
```
ansd: This function is also called for will messages, in which case you include a packet ID of `0`, which isn't allowed according to the spec. So you need extra logic for the will message. Probably we don't want to send any PUBACK to the client if this is the will message?
See test case `publish_permission_will_message` and macro `WILL_MSG_QOS_1_CORRELATION`

deps/rabbitmq_mqtt/test/disconnect_on_unauthorized_SUITE.erl
```diff
@@ -0,0 +1,141 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
+%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%
+%% Copyright (c) 2007-2025 Broadcom. All Rights Reserved. The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries. All rights reserved.
+
+
+-module(disconnect_on_unauthorized_SUITE).
```
ansd, line 8: You probably need to add this suite to the Makefile.

deps/rabbitmq_mqtt/test/disconnect_on_unauthorized_SUITE.erl
```diff
@@ -0,0 +1,141 @@
...
+groups() ->
+    [
+        {v4, [], test_cases()},
+        {v5, [], test_cases()}
+    ].
+
+test_cases() ->
+    [
+        publish_unauthorized_no_disconnect,
+        subscribe_unauthorized_no_disconnect
```
ansd: Can you also add a test case for `UNSUBSCRIBE`, please? We have the same expectation: The network connection should stay open if the client tries to unsubscribe from a topic filter it doesn't have authorization for. (It seems this is already implemented prior to your PR.)

deps/rabbitmq_mqtt/test/disconnect_on_unauthorized_SUITE.erl
ansd: Let's add it to `PARALLEL_CT_SET_1_A` in  https://github.com/rabbitmq/rabbitmq-server/blob/dd9633976d2b550d6ee21e95647bfde2f5be8cc7/deps/rabbitmq_mqtt/Makefile#L98-L102

ansd:
Thank you @iamdrq. I don't have permission to push to your branch. So, I created a new PR in https://github.com/rabbitmq/rabbitmq-server/pull/15201 and cherry-picked your commits and did some small refactor on top.

PR #15136, opened 2025-12-15
CQ: Use open_eventually in recovery

lhoguin:
It's possible for Windows to be so stubborn that the file can't be opened after open_eventually has given up, which causes a crash and force the queue to be restarted, then trying to open that same file during recovery still errors out.

Using open_eventually there may or may not help.

Also use open_eventually in queue store's get_read_fd as a crash log has shown it can happen there as well.

#15134 

Note that this isn't a "real fix" but just more workarounds for some annoying Windows behavior.


lhoguin:
This can be backported to 4.2.x and 4.1.x if we still produce those.

PR #15149, opened 2025-12-16
 Always explicitly use rabbit_framing_amqp_0_9_1

lhoguin:
This commit removes the 'protocol' field from rabbit_channel's
'#conf' record; rabbit_reader's '#connection' record (defined
in rabbit.hrl); and rabbit_writer's '#wstate' record.

The 'protocol' field in the '#content' record is still present
but no longer used. It can be removed in the RabbitMQ version
that follows the next LTS, as then all we will need is some
compatibility code when reading older messages from disk (and
not when sending messages across clusters), making its removal
less dangerous than now.

As amqp_client does a few RPC calls we are leaving functions
with a 'Protocol' argument that is ignored. They can be removed
in a few releases when we no longer have to worry about older
amqp_client versions connecting, similarly to '#content'.

### DIFF
--- a/deps/amqp_client/src/amqp_channel.erl
+++ b/deps/amqp_client/src/amqp_channel.erl
@@ -927,7 +927,7 @@ is_connection_method(Method) ->
     ?PROTOCOL:lookup_class_name(ClassId) == connection.
 
 server_misbehaved(#amqp_error{} = AmqpError, State = #state{number = Number}) ->
-    case rabbit_binary_generator:map_exception(Number, AmqpError, ?PROTOCOL) of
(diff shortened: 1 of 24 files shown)


lhoguin:
This removes `protocol` from being used in the `#content` record, allowing for a future mostly painless removal. It also removes it from the state of a few processes and the `#connection` record.

The PR is functionally ready but a test never finishes with these changes. I will need to fix it before it can be merged.

ansd:
This is a follow up of https://github.com/rabbitmq/rabbitmq-server/pull/15088

ansd (approved):
Thank you!

PR #15166, opened 2025-12-19
#13873 by @Ayanda-D, polished and rebased

michaelklishin:
This is #13873 by @Ayanda-D with clarifying comments and one CLI command fix from me.

Due to over 600 commits rebased, this branch has accumulated dozens of commits that are only recorded due to conflict resolution. The actually meaningful part of the diff should be very comparable to #13873.


michaelklishin:
I would not even try to backport this via Mergify => removed the tag.

PR #15171, opened 2025-12-20
DQT: be more defensive, improve unit and property-based test coverage

michaelklishin:
References #11541, #12109, #12821, #13837


michaelklishin:
https://github.com/michaelklishin/rabbitmq-3.13-dqt-bug-reproduction explains the broader context. This is the a fix per se but rather an improvement in test coverage and an extra couple of layers of protection against the issue explained in the repo.

The repo also mentions a durable workaround.

PR #15189, opened 2025-12-29
Introduce rabbitmq-stream equivalents of 'rabbitmq-queues {grow,shrink}'

michaelklishin:
I will separately clarify why these commands were not added, it is not necessarily something overlooked.

Anyhow, here's a PR, let's see if it should be merged :)


michaelklishin:
Merging per discussion with @kjnilsson.

[Stream replica management](https://www.rabbitmq.com/docs/streams#member-management) has room for improvement but maybe those concerns are mostly theoretical.

PR #15205, opened 2026-01-03
Add a 'resource-counts' endpoint to Management Plugin

AaronKaa:
## Proposed Changes

I have a need for a lightweight endpoint that supplies resource counts for a frontend project. The need for me is that in building a front end client I couldn't find a simple way to get these values without calling many other endpoints that provide way too much data. 

A previous iteration I made was to update api/overview and added switches so it could only supply this data but it was messy.

Im happy to work this further with feedback etc. 

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments


michaelklishin:
Thank you for taking the time to contribute but we won't accept this PR.

We highly recommend potential contributors

1. First start a discussion clarifying whether their contribution has a chance of being accepted
2. Sign the [CLA](https://github.com/rabbitmq/cla) around the time they submit a PR

`GET /api/overview` already returns

```
  "queue_totals": {},
  "object_totals": {
    "channels": 0,
    "consumers": 0,
    "exchanges": 45,
    "queues": 0,
    "connections": 0
  },
```

and this is what the management UI uses on the overview page.

You haven't explained why your "frontend project" needs something beyond that.

RabbitMQ does use "resources" internally but it's not a term we use a whole lot. `GET /api/overview` uses "objects", definitions-related docs and commands use "entities" in some places.

If I had to introduce such "counters" or "counts" endpoint with per-virtual host filtering,
I would use something like this

 * `GET /api/objects/counts`
 * `GET /api/objects/counts/{vhost}`
 
and include every object type that exported [definitions](https://www.rabbitmq.com/docs/definitions) include plus `streams`, since they do not have a separate definitions document key and are grouped with `queues` but semantically do form an object category of their own.

Returning a node count in `GET /api/objects/counts/{vhost}` feels weird and unnecessary. On top of that, there are already `GET /api/nodes` that is effectively guaranteed to produce a small response: you likely won't have hundreds or thousands of cluster nodes.

So, please start a discussion and make your case there.

AaronKaa:
Thanks @michaelklishin thats exactly the kind of feedback I was looking for, I have just signed the CLA - where would I start a discussion ?

Also I appreciate you mentioning adding other exported definitions, this was on my mind.

michaelklishin:
@AaronKaa in this repository's [Discussions](https://github.com/rabbitmq/rabbitmq-server/discussions).

Here is a [discussion from a regular contributor](https://github.com/rabbitmq/rabbitmq-server/discussions/15098) at AWS, for example.

PR #15208, opened 2026-01-05
Remove looking glass from rabbit_fifo_int_SUITE

ansd:
Looking glass got removed from RabbitMQ in a previous version.

This commit fixes the following error:
```
{lg,{"no such file or directory","lg.app"}}
```


ansd:
Same unrelated failing test as in https://github.com/rabbitmq/rabbitmq-server/pull/15206#issuecomment-3708105166

PR #15214, opened 2026-01-06
rabbit_khepri: Optimize cleanup in topic binding projection

dumbbell:
## Why

Before this patch, the projection function that managed the topic bindings graph only kept track of "topic word -> topic word" = "this node ID". Reading and updating this graph was efficient.

However, deleting a node in this graph was expensive because it used `ets:match/3` to determine if an edge was pointing to nothing and could be reclaimed. `ets:match()` does a full scan of the table. On my laptop, this scan took 20 ms with 100k topic bindings, thus >30 minutes to delete all of them.

## How

The new projection function tracks the number of children a target node has as well. This way, it knows that if this counter reaches 0, the edge can be reclaimed.

The same test with 100k topic bindings takes 3.5 seconds to delete them.

### DIFF
--- a/deps/rabbit/src/rabbit_db_topic_exchange.erl
+++ b/deps/rabbit/src/rabbit_db_topic_exchange.erl
@@ -26,7 +26,7 @@
 -define(MNESIA_NODE_TABLE, rabbit_topic_trie_node).
 -define(MNESIA_EDGE_TABLE, rabbit_topic_trie_edge).
 -define(MNESIA_BINDING_TABLE, rabbit_topic_trie_binding).
--define(KHEPRI_PROJECTION, rabbit_khepri_topic_trie_v2).
+-define(KHEPRI_PROJECTION, rabbit_khepri_topic_trie_v3).
 
 -type match_result() :: [rabbit_types:binding_destination() |
                          {rabbit_amqqueue:name(), rabbit_types:binding_key()}].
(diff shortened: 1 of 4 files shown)


dumbbell:
This not ready for review or testing at all, my understanding of the problem is incorrect.

mkuratczyk:
I thnik we need to bump Horus or remove this dep_horus:
https://github.com/rabbitmq/rabbitmq-server/blob/optimize-topic-bindings-deletion/deps/rabbit/Makefile#L136

As it is, Horus 0.3.1 is used on this branch, causing a known/fixed issue.

dumbbell:
Good catch, I forgot about this old temporary change.

dumbbell:
> I thnik we need to bump Horus or remove this dep_horus: https://github.com/rabbitmq/rabbitmq-server/blob/optimize-topic-bindings-deletion/deps/rabbit/Makefile#L136
> 
> As it is, Horus 0.3.1 is used on this branch, causing a known/fixed issue.

This was fixed in #15230.

mkuratczyk (approved):
As discussed, I can still make this behave poorly as we need one more change (fix for exchange serials) but this part is good and the serials fix is coming soon. thanks!

dumbbell:
The mentionned fix to exchange serials is now submitted serapately in #15236.

PR #15225, opened 2026-01-08
Add committed offset in stream stats

acogoluegnes:
The committed offset (offset of the last entry in the stream) is now
available in Osiris. It is used instead of the "committed chunk ID"
where appropriate: to calculate the offset lag and the number of
available messages.

It is more accurate to use the committed offset for the offset lag than
to use the committed chunk ID.
Before: offset lag = committed chunk ID - ID of the last dispatched
chunk
Now: offset lag = committed offset - offset of the last message in the
last dispatched chunk

This change uses the rabbitmq_4.3.0 feature flag to
enable the calculation of the committed offset in Osiris only when all
the nodes have been migrated. The streams need to be restarted as the
feature is enabled when starting the writer and the replicas.

This PR bumps Osiris to v1.11.0.

References https://github.com/rabbitmq/osiris/pull/203

### DIFF
--- a/deps/rabbit/src/rabbit_stream_queue.erl
+++ b/deps/rabbit/src/rabbit_stream_queue.erl
@@ -796,32 +796,32 @@ i(messages_unacknowledged, Q) when ?is_amqqueue(Q) ->
         [] ->
             0
     end;
-i(committed_offset, Q) ->
+i(committed_offset = F, Q) ->
     %% TODO should it be on a metrics table?
     %% The queue could be removed between the list() and this call
     %% to retrieve the overview. Let's default to '' if it's gone.
     Key = {osiris_writer, amqqueue:get_name(Q)},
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_core_ff.erl
```diff
@@ -190,6 +190,14 @@
...
      }}).
 
+-rabbit_feature_flag(
+   {stream_committed_offset,
+    #{desc          => "Make stream committed offset counter available",
+      doc_url       => "https://github.com/rabbitmq/rabbitmq-server/pull/15225",
+      stability     => stable,
+      depends_on    => [stream_queue]
+     }}).
+
```
ansd: If this PR is for 4.3 and not meant to be backported, should we remove this feature flag and instead make use of the already existing feature flag `rabbitmq_4.3.0`?

PR #14119, opened 2025-06-24
Make queue commands generic - i.e. make them work with any queue type

ikavgo:
Continuing Queues as Plugins Saga with generalizing queues commands.

I also tried to define types for #resource record fields, but unsuccessfully - the r1/2/3 helper types we have confuse dialyzer.

PR #15226, opened 2026-01-08
Make a minimal set of `rabbitmq-queues` CLI commands generic

ansd:
## What?

This commit makes the following `rabbitmq-queues` CLI commands queue
type generic such that they do not only work against quorum queues, but
also against other (replicated or Ra based) queue types:
```
rabbitmq-queues add_member
rabbitmq-queues delete_member
rabbitmq-queues quorum_status
rabbitmq-queues grow
rabbitmq-queues shrink
```

This PR is a subset of what was aimed for in https://github.com/rabbitmq/rabbitmq-server/pull/14119.

 ## Why?

Plugins may introduce other Ra based queue types which should
also support above CLI commands. Instead of creating new queue type
specific CLI commands, it's more user friendly if above existing CLI
commands work with the new queue types.

 ## How?

The following queue based CLI commands already work against multiple
queue types:
```
rabbitmqctl list_queues
rabbitmqctl purge_queue
rabbitmqctl delete_queue
rabbitmqctl list_unresponsive_queues
rabbitmq-queues rebalance
rabbitmq-queues check_if_node_is_quorum_critical
```

The following CLI commands stay quorum queue specific:
```
rabbitmq-queues check_if_new_quorum_queue_replicas_have_finished_initial_sync
rabbitmq-queues peek
rabbitmq-queues reclaim_quorum_memory
rabbitmq-diagnostics check_for_quorum_queues_without_an_elected_leader
rabbitmq-queues force_checkpoint
```
Command `rabbitmq-queues force_checkpoint` will be a no-op starting in 4.3 thanks to Ra v3.

The following CLI commands are made queue type generic:
```
rabbitmq-queues add_member
rabbitmq-queues delete_member
rabbitmq-queues quorum_status
rabbitmq-queues grow
rabbitmq-queues shrink
```
Adding a member, deleting a member, and getting the status for a queue
are the most essential commands. Commands `grow` and `shrink` are only
wrappers for `add_member` and `delete_member`. Arguably, a simple bash
script is sufficient to call `add_member` or `delete_member` multiple
times. But for consistency and convenience this commit decides to make
`grow` and `shrink` also generic.

These CLI commands call from now on into `rabbit_queue_type_ra` instead of
calling into `rabbit_quorum_queue`.

This commit also ensures that the CLI commands are mixed version
compatible. In a mixed version 4.3 / 4.2 cluster, both a 4.3 CLI
and a 4.2 CLI should be able to execute for example
`rabbitmq-queues quorum_status` against either 4.3 or 4.2 nodes.

Making the following quorum queue specific HTTP API endpoints queue type generic
is out of scope in this commit:
```
/queues/quorum/:vhost/:queue/status
/queues/quorum/:vhost/:queue/replicas/add
/queues/quorum/:vhost/:queue/replicas/delete
/queues/quorum/replicas/on/:node/grow
/queues/quorum/replicas/on/:node/shrink
```
Unfortunately, they contain `quorum` in the HTTP path and also contain
quorum queue specific configuration variables, such as "restriction" `quorum_queue_replica_operations`.
It's up for discussion whether these same HTTP endpoints should be made
queue type generic or whether new HTTP endpoints should be introduced for new
queue types.

### DIFF
--- a/deps/rabbit/src/rabbit_queue_type_ra.erl
+++ b/deps/rabbit/src/rabbit_queue_type_ra.erl
@@ -33,31 +33,30 @@
 -spec status(rabbit_types:vhost(), rabbit_misc:resource_name()) ->
     [[{binary(), term()}]] | {error, term()}.
 status(VHost, Name) ->
-    QName = rabbit_misc:queue_resource(VHost, Name),
-    case rabbit_amqqueue:lookup(QName) of
-        {ok, Q} ->
-            Mod = amqqueue:get_type(Q),
-            case is_ra_based(Mod) of
-                true ->
-                    Mod:status(Q);
-                false ->
-                    {error, {unsupported, Mod}}
-            end;
-        {error, not_found} = Err ->
-            Err
-    end.
+    with_ra_queue(VHost, Name,
+                  fun(Mod, Q) -> Mod:status(Q) end).
 
 -spec add_member(rabbit_types:vhost(), rabbit_misc:resource_name(),
                  node(), ra_membership(), timeout()) ->
     ok | {error, term()}.
 add_member(VHost, Name, Node, Membership, Timeout) ->
+    with_ra_queue(VHost, Name,
+                  fun(Mod, Q) -> Mod:add_member(Q, Node, Membership, Timeout) end).
+
+-spec delete_member(rabbit_types:vhost(), rabbit_misc:resource_name(), node()) ->
+    ok | {error, term()}.
+delete_member(VHost, Name, Node) ->
+    with_ra_queue(VHost, Name,
+                  fun(Mod, Q) -> Mod:delete_member(Q, Node) end).
+
+with_ra_queue(VHost, Name, Fun) ->
     QName = rabbit_misc:queue_resource(VHost, Name),
     case rabbit_amqqueue:lookup(QName) of
         {ok, Q} ->
             Mod = amqqueue:get_type(Q),
             case is_ra_based(Mod) of
                 true ->
-                    Mod:add_member(Q, Node, Membership, Timeout);
+                    Fun(Mod, Q);
                 false ->
                     {error, {unsupported, Mod}}
             end;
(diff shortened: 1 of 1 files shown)


michaelklishin (approved):

deps/rabbit/src/rabbit_queue_type_ra.erl
```diff
@@ -0,0 +1,185 @@
...
+
+-callback add_member(amqqueue:amqqueue(), node(), ra_membership(), timeout()) ->
+    ok | {error, term()}.
+
+-callback delete_member(amqqueue:amqqueue(), node()) ->
+    ok | {error, term()}.
+
+-spec status(rabbit_types:vhost(), rabbit_misc:resource_name()) ->
+    [[{binary(), term()}]] | {error, term()}.
+status(VHost, Name) ->
```
Ayanda-D, line 35: @ansd cool change - just an addition which can further minimilize the code - `status/2`, `add_member/5` and `delete_member/3` are doing mostly the same things except the specific ra operation. what do you think we extract that to a single common function (seeing making things even _more_ generic is what we're going for 🙂), for example:

```erlang
status(VHost, Name) ->
    some_common_function(VHost, Name, fun(Mod, Q) -> Mod:status(Q) end);
    
add_member(VHost, Name, Node, Membership, Timeout) ->
    some_common_function(VHost, Name,
        fun(Mod, Q, Node, Membership, Timeout) -> Mod:add_member(Q, Node, Membership, Timeout) end);

delete_member(VHost, Name, Node) ->
    some_common_function(VHost, Name, fun(Mod, Q) -> Mod:delete_member(Q, Node) end);

some_common_function(VHost, Name, RaQFun) ->
    QName = rabbit_misc:queue_resource(VHost, Name),
    case rabbit_amqqueue:lookup(QName) of
        {ok, Q} ->
            Mod = amqqueue:get_type(Q),
            case is_ra_based(Mod) of
                true ->
                    RaQFun(Mod, Q);
                false ->
                    {error, {unsupported, Mod}}
            end;
        {error, not_found} = Err ->
            Err
    end.

File: deps/rabbit/src/rabbit_queue_type_ra.erl [hunk 2]
```diff
@@ -0,0 +1,185 @@
...
+                    {error, {unsupported, Mod}}
+            end;
+        {error, not_found} = Err ->
+            Err
+    end.
+
+%% For each Ra-based queue matching VHostSpec and QueueSpec, add a member on Node.
+-spec add_members(binary(), binary(), node(), all | even, ra_membership()) ->
+    [{rabbit_amqqueue:name(), {ok, pos_integer()} | {error, pos_integer(), term()}}].
+add_members(VHostSpec, QueueSpec, Node, Strategy, Membership) ->
```
Ayanda-D, line 70: i reckon we can do the same for `add_members/5` and `delete_members/4` - which are doing mostly the same things, we can reduce code footprint by passing `Mod:add_member/4` and `Mod:delete_member/2` to a common function for both, example:

```erlang

add_members(VHostSpec, QueueSpec, Node, Strategy, Membership) ->
     another_common_function(..., fun(Mod, Q) -> Mod:add_member(Q, Node, Membership, RA_MEMBERS_TIMEOUT) end).
     
delete_members(VHostSpec, QueueSpec, Node, Strategy) ->
     another_common_function(..., fun(Mod, Q) -> Mod:delete_member(Q, Node) end).
     
another_common_function(VHostSpec, QueueSpec, Strategy, RaQsFun) ->
    case lists:member(Node, rabbit_nodes:list_running()) of
        true ->
            [begin
                 Mod = amqqueue:get_type(Q),
                 QName = amqqueue:get_name(Q),
                 QNodes = amqqueue:get_nodes(Q),
                 Size = length(QNodes),
                 {ok, RaName} = rabbit_queue_type_util:qname_to_internal_name(QName),
                 Res = case all_members_stable(RaName, QNodes) of
                           true ->
                               case RaQsFun(Mod, Q) of
                                   ok ->
                                       {ok, Size + 1};
                                   {error, Reason} ->
                                       {error, Size, Reason}
                               end;
                           false ->
                               {error, Size, {error, non_stable_members}}
                       end,
                 {QName, Res}
             end
             || Q <- rabbit_amqqueue:list(),
                not lists:member(Node, amqqueue:get_nodes(Q)),
                matches_strategy(Strategy, amqqueue:get_nodes(Q)),
                is_ra_based(amqqueue:get_type(Q)),
                is_match(amqqueue:get_vhost(Q), VHostSpec),
                is_match(get_resource_name(amqqueue:get_name(Q)), QueueSpec)];
        false ->
            {error, {node_not_running, Node}}
    end.

ansd: Thanks for your review. You're right that this removes code duplication. Whether this code is simpler to understand is not clear though. But let's go with what you suggested.

deps/rabbit/src/rabbit_queue_type_ra.erl
ansd: Thanks for your review. You're right that this removes code duplication. Whether this code is simpler to understand is not clear though. But let's go with what you suggested.

PR #15255, opened 2026-01-13
rabbit_db_{exchange,vhost}: be more defensive

michaelklishin:
when a path query returns an empty result.

During concurrent deletes in rabbitmq_exchange_federation's exchange_SUITE, the lack of handling for these cases produces test flakes.

Note that rabbit_db_queue has a catch-all (_) clause which makes that module immune to this problem.


michaelklishin:
I keep hitting these with federation test suites run in a loop, so let's merge it.

PR #15258, opened 2026-01-14
Prevent federation links from restarting during node shutdown

michaelklishin:
or plugin shutdown, for that matter.

With this guardrail in place, nodes with hundreds or thousands of federation links will avoid potentially significant shutdown delays that have to do with
links being restarted while the node as a whole is preparing to shut down.

This state is node-local, as is the shutdown state, so this will not prevent links migrating between nodes (under `mirrored_supervisor`) from starting.

Per discussion with @dcorbacho @ansd.

Note that this PR cannot be backported exactly to `v4.1.x` and earlier branches. The federation plugin split in `main` first shipped in `4.2.0`.

### DIFF
--- a/deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
+++ b/deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
@@ -311,14 +311,20 @@ record_binding(B = #binding{destination = Dest},
                   end,
     {DoIt, State#state{bindings = maps:put(key(B), Set, Bs)}}.
 
+-spec forget_binding(rabbit_types:binding(), #state{}) -> {boolean(), #state{}}.
 forget_binding(B = #binding{destination = Dest},
                State = #state{bindings = Bs}) ->
-    Dests = sets:del_element(Dest, maps:get(key(B), Bs)),
-    {DoIt, Bs1} = case sets:size(Dests) of
-                      0 -> {true,  maps:remove(key(B), Bs)};
-                      _ -> {false, maps:put(key(B), Dests, Bs)}
-                  end,
-    {DoIt, State#state{bindings = Bs1}}.
+    case maps:find(key(B), Bs) of
+        error ->
+            {false, State};
+        {ok, Dests0} ->
+            Dests = sets:del_element(Dest, Dests0),
+            {DoIt, Bs1} = case sets:size(Dests) of
+                              0 -> {true,  maps:remove(key(B), Bs)};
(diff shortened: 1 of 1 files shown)


deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
```diff
@@ -66,6 +66,14 @@ start_link(Args) ->
     gen_server2:start_link(?MODULE, Args, [{timeout, infinity}]).
 
 init({Upstream, XName}) ->
+    case rabbit_federation_app_state:is_shutting_down() of
+        true ->
+            ignore;
```
ansd, line 72: Should we add a log message at DEBUG or INFO level here?
This way, it's obvious to understand that a federation link (re)start was taking place, but that we intentionally bailed out since we are shutting down.
I think an additional log message won't pollute the logs.

deps/rabbitmq_queue_federation/src/rabbit_federation_queue_link.erl
```diff
@@ -54,6 +54,14 @@ q(QName) ->
 %%----------------------------------------------------------------------------
 
 init({Upstream, Queue}) when ?is_amqqueue(Queue) ->
+    case rabbit_federation_app_state:is_shutting_down() of
+        true ->
+            ignore;
```
ansd, line 60: Same question as above: Should we add a log statement here?

deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
MirahImage: I think at least a debug level log would be helpful. I could imagine normal log level might be noisy in the customer environments with thousands of federation links where the issue this is solving comes up.

deps/rabbitmq_queue_federation/src/rabbit_federation_queue_link.erl
michaelklishin: I will.

deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
michaelklishin: No objections.

michaelklishin:
I have re-created the branch to make GitHub pick up on commit a8347b24eefa228dc01b1dade1b0c5a104fed417.

PR #15267, opened 2026-01-15
Revert "rabbit_db_{exchange,vhost}: be more defensive (backport #15255)"

michaelklishin:
Reverts rabbitmq/rabbitmq-server#15262


michaelklishin:
We need to investigate the value returned from the metadata store more carefully. This was just a flaky exchange federation test fix, not a functional problem I have seen in a real world deployment.

The original PR hasn't shipped in any release, so I will remove all milestones.

PR #15270, opened 2026-01-15
Stop federation supervision hierachies before stopping `rabbit`

ansd:
## What?

Federation links started in the federation plugins are put
under the `rabbit` app supervision tree (unfortunately).

This commit ensures that the entire federation supervision hierarchies
(including all federation links) are stopped **before** stopping app
`rabbit` when stopping RabbittMQ.

 ## Why?

Previously, we've seen cases where hundreds of federation links are
stopped during the shutdown procedure in app `rabbit` leading to
federation link restarts happening in parallel to vhosts being stopped.
In one case, the shutdown of app `rabbit` even got stuck (with a federation
error being the last lines being logged).

The cleaner approach is to gracefully stop all federation
links, i.e. the entire supervision hierarchy under
`rabbit_exchange_federation_sup` and `rabbit_queue_federation_sup`
when stopping the federation apps, i.e. **before** proceeding to stop
app `rabbit`.

 ## How?

The boot step cleanup steps for the federation plugins are skipped when
stopping RabbitMQ.

Hence, this commit ensures that the supervisors are stopped in the
stop/1 application callback.

This commit does something similar to https://github.com/rabbitmq/rabbitmq-server/pull/14054
but uses a simpler approach.

### DIFF
--- a/deps/rabbitmq_exchange_federation/src/rabbit_exchange_federation_app.erl
+++ b/deps/rabbitmq_exchange_federation/src/rabbit_exchange_federation_app.erl
@@ -42,6 +42,7 @@ prep_stop(State) ->
     State.
 
 stop(_State) ->
+    ok = rabbit_exchange_federation_sup:stop(),
     ets:delete(?FEDERATION_ETS, rabbitmq_exchange_federation),
     rabbit_federation_pg:stop_scope(?FEDERATION_PG_SCOPE),
     ok.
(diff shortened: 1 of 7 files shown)


ansd:
To do some local testing, you can use for example the following commands:
```bash
make run-broker
./sbin/rabbitmq-plugins enable rabbitmq_exchange_federation
./sbin/rabbitmqctl set_parameter federation-upstream origin '{"uri":"amqp://localhost:5672"}'
./sbin/rabbitmqctl set_policy exchange-federation "^amq.direct" '{"federation-upstream-set":"all"}' --priority 10 --apply-to exchanges
./sbin/rabbitmqctl stop
```

ansd:
My commits were cherry-picked to https://github.com/rabbitmq/rabbitmq-server/pull/15271. I will therefore review https://github.com/rabbitmq/rabbitmq-server/pull/15271 soon.

ansd:
I close this PR in favor of https://github.com/rabbitmq/rabbitmq-server/pull/15271

PR #15271, opened 2026-01-15
Federation: parallel shutdown; disconnect links before stopping

michaelklishin:
Technical design pair: @ansd.

## Proposed Changes

This PR makes the most expensive part of federation link shutdown — closing AMQP 0-9-1 connections to the upstream — parallel, by notifying links in the `prep_stop` shutdown callback.

This yields very significant efficiency gains with hundreds or thousands of links, all without changing the supervision tree structure.

## Why Not Use `simple_one_for_one`?

Indeed the `simple_one_for_one` OTP supervisor restart strategy would shut down all child processes concurrently for us. But this would require changing the child identity (key)
to an Erlang PID, which would require intrusive and painful to test changes (such as
an ETS table that maps their PIDS to the current identities and the other way around).

## Throttling to Avoid Overwhelming the Upstream

To avoid overwhelming the upstream schema data store (which could be a 7-9 node cluster on 3.x with Mnesia), we limit the degree of parallelism and add batching with configurable throttling delays into the process.

The entire link shutdown process is now capped at 180 seconds (by default), and should not meaningfully exceed that time period even on nodes with many thousands of links.

By default we close up to 128 links per batch, with a 50 ms delay, and a 180 second hard cap (timeout) for the entire link termination operation.

## Data Safety Considerations

Federation uses publisher confirms by default, and most users never change it, therefore
aggressive connection closures are safe and acceptable.

In addition, the user can set `resource-cleanup-mode` to `never` to make sure that the
upstream resources (e.g. internal queues used by exchange federation) are never deleted
by the links running in the downstream cluster.

### Microbenchmarks (Supervisor Child Process Termination)

Below are some microbenchmarks that measure everything beyond the actual
AMQP 0-9-1 connection termination part on an 8 core aarch64 CPU from 2022:

```
┌───────┬──────────┬────────────┬─────────┐
│ Links │ Parallel │ Sequential │ Speedup │
├───────┼──────────┼────────────┼─────────┤
│ 1,000 │ 16ms     │ 6,401ms    │ ~400x   │
├───────┼──────────┼────────────┼─────────┤
│ 5,000 │ 30ms     │ 32,689ms   │ ~1,000x │
└───────┴──────────┴────────────┴─────────┘
```

### Worst Case Scenario Calculations

If we consider the worst case scenario where every link connection hits its timeout,
1K links would take about 83 minutes to start for the sequential (status quo) version and 5.6 seconds (see below) with these changes.

### Real World Federation Links with Outgoing Connections

With the throttling delay of 0, the time it takes to shut down N links to a remote upstream
cluster look like this:

```
┌───────┬─────────────────┐
│ Links │      Time       │
├───────┼─────────────────┤
│ 10    │ 100ms           │
├───────┼─────────────────┤
│ 50    │ 580ms           │
├───────┼─────────────────┤
│ 100   │ 1,067ms         │
├───────┼─────────────────┤
│ 1,000 │ 5,579ms         │
└───────┴─────────────────┘
```

### [Maintenance Mode](https://www.rabbitmq.com/docs/upgrade#maintenance-mode) Integration

Maintenance mode integration of these changes needs to be done with care: since maintenance mode stops all client connection listeners, we run the risk of stopping the listeners
before this part of the federation shutdown has a chance to do its job as designed.

For that reason, we have to special case the federation plugins in the core and first trigger their termination, then stop the listeners.

When the node is revived (the maintenance mode is rolled back), all links are restarted.

### DIFF
--- a/deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
+++ b/deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
@@ -45,7 +45,7 @@
                 unacked,
                 internal_exchange_timer,
                 internal_exchange_interval,
-                link_state = running}).
+                link_state = starting}).
 
 %%----------------------------------------------------------------------------
 
@@ -573,7 +573,8 @@ go(S0 = {not_started, {Upstream, UParams, DownXName}}) ->
               ?LOG_INFO("Federation link for ~ts (upstream: ~ts) will perform internal exchange checks "
                                          "every ~b seconds", [rabbit_misc:rs(DownXName), UName, round(Interval / 1000)]),
               TRef = erlang:send_after(Interval, self(), check_internal_exchange),
-              {noreply, State#state{internal_exchange_timer = TRef}}
+              {noreply, State#state{internal_exchange_timer = TRef,
+                                    link_state = running}}
       end, Upstream, UParams, DownXName, S0).
 
 log_link_startup_attempt(#upstream{name = Name, channel_use_mode = ChMode}, DownXName) ->
--- a/deps/rabbitmq_queue_federation/src/rabbit_federation_queue_link.erl
+++ b/deps/rabbitmq_queue_federation/src/rabbit_federation_queue_link.erl
@@ -27,7 +27,7 @@
 
 -record(not_started, {queue, run, upstream, upstream_params}).
 -record(state, {queue, run, conn, ch, dconn, dch, upstream, upstream_params,
-                unacked, link_state = running}).
+                unacked, link_state = starting}).
 
 start_link(Args) ->
     gen_server2:start_link(?MODULE, Args, [{timeout, infinity}]).
@@ -302,7 +302,8 @@ go(S0 = #not_started{run             = Run,
                                dch             = DCh,
                                upstream        = Upstream,
                                upstream_params = UParams,
-                               unacked         = Unacked}}
+                               unacked         = Unacked,
+                               link_state      = running}}
       end, Upstream, UParams, QName, S0).
 
 check_upstream_suitable(Conn) ->


deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
```diff
@@ -265,9 +316,38 @@ join(Name) ->
...
+
 x(XName) ->
     pg:get_members(?FEDERATION_PG_SCOPE, pgname({rabbit_federation_exchange, XName})).
 
+-spec disconnect_all() -> ok.
+disconnect_all() ->
+    Pids = all_local(),
+    case Pids of
+        [] -> ok;
+        _  -> ?LOG_DEBUG("Exchange federation: disconnecting ~b local link(s) for shutdown",
```
ansd: Can we log the number of links to be disconnected at INFO level?

deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
```diff
@@ -265,9 +316,38 @@ join(Name) ->
...
+    end,
+    [gen_server2:cast(Pid, disconnect_for_shutdown) || Pid <- Pids],
+    ok.
+
+-spec reconnect_all() -> ok.
+reconnect_all() ->
+    Pids = all_local(),
+    case Pids of
+        [] -> ok;
+        _  -> ?LOG_DEBUG("Exchange federation: reconnecting ~b local link(s)",
```
ansd: Same here, I suggest INFO level.

deps/rabbitmq_federation_common/src/rabbit_federation_pg.erl
```diff
@@ -20,29 +19,92 @@ stop_scope(Scope) ->
...
+    end,
+    AllMembers = lists:usort(lists:flatmap(GetMembers, Groups)),
+    case AllMembers of
+        [] -> ok;
+        _  -> do_terminate_all(AllMembers, Timeout, BatchSize, ThrottleDelay)
+    end.
+
+do_terminate_all(Members, Timeout, BatchSize, ThrottleDelay) ->
+    Count = length(Members),
+    ?LOG_DEBUG("Federation: initiating paced shutdown for ~b link(s) "
```
ansd: Same here, I suggest INFO level.

deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
```diff
@@ -126,6 +127,49 @@ handle_cast({enqueue, Serial, Cmd},
...
+                        catch amqp_channel:call(Ch, #'queue.delete'{queue = Queue}),
+                        catch amqp_channel:call(Ch, #'exchange.delete'{exchange = IntExchange})
+                    catch
+                        _:_ -> ok
+                    end
+            end,
+            rabbit_federation_link_util:ensure_connection_closed_async(Conn, Timeout)
+    end,
+    {noreply, State#state{downstream_connection = undefined,
+                          connection = undefined}};
```
ansd: I'm not sure whether some other code paths rely on the state's `channel` and `downstream_channel` to be set even if `connection = undefined` or `downstream_connection = undefined`. Maybe it's cleaner to add a new field to the state record called `link_state` or `link_lifecycle` (or a similar name) and set it from `running` to `closing`. (A bit similar to https://github.com/rabbitmq/rabbitmq-server/blob/7083ba39041d605988588c2d70566fa8c3c7f68d/deps/rabbit/include/rabbit_amqp_reader.hrl#L53-L55 just with fewer different states).

michaelklishin:
Some numbers on the effectiveness of these changes:

image

PR #15292, opened 2026-01-17
Speed up tests

ansd:
```
rabbit_ct_broker_helpers:stop_broker/2
rabbit_ct_broker_helpers:start_broker/2
```

is about 6 seconds faster compared to

```
rabbit_ct_broker_helpers:stop_node/2
rabbit_ct_broker_helpers:start_node/2
```

This commit uses the faster versions in many tests where for example we only want to trigger a leader change or a queue becoming unavailable.

In some test cases, we actually want the entire Erlang node to be down. Those tests continue to use `rabbit_ct_broker_helpers:stop_node/2`.


michaelklishin:
Thank you, @ansd!

PR #15294, opened 2026-01-18
Add test enabling all feature flags with exchange logger

ansd:
Add a test enabling all stable feature flags with exchange logging enabled.

This is a more general test case for https://github.com/rabbitmq/rabbitmq-server/discussions/11652 that should catch all future feature flag issues when `rabbit_ff_controller` publishes a message.

Relates
https://github.com/rabbitmq/rabbitmq-server/discussions/11652 https://github.com/rabbitmq/rabbitmq-server/discussions/14069 https://github.com/rabbitmq/rabbitmq-server/pull/14796

### DIFF
--- a/deps/rabbit/test/logging_SUITE.erl
+++ b/deps/rabbit/test/logging_SUITE.erl
@@ -15,6 +15,9 @@
 -include_lib("amqp_client/include/amqp_client.hrl").
 -include_lib("rabbitmq_ct_helpers/include/rabbit_assert.hrl").
 
+-import(rabbit_ct_broker_helpers,
+        [rpc/4]).
+
 -export([suite/0,
          all/0,
          groups/0,
@@ -54,6 +57,7 @@
          logging_to_exchange_works/1,
          update_log_exchange_config/1,
          use_exchange_logger_when_enabling_khepri_db/1,
+         use_exchange_logger_when_enabling_all_feature_flags/1,
 
          logging_to_syslog_works/1]).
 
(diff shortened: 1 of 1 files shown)


ansd:
This test should be backported to `v4.2.x`, `v4.1.x`, and `v4.0.x`.
This test is already part of VMware Tanzu RabbitMQ `v3.13.x`.

gomoripeti:
thank you
(we only test with log exchange after a release is out and only on info level, so https://github.com/rabbitmq/rabbitmq-server/discussions/14069 was only caught on a customer's cluster. great to now have a test case to catch anything like this early)

ansd:
> we only test with log exchange after a release is out and only on info level

Thank you. Enabling feature flag `message_containers_deaths_v2` on the OSS `v3.13.x` branch with the log exchange at **info** level also gets stuck. Did you catch this during your manual testing?

michaelklishin:
Note that backporting https://github.com/rabbitmq/rabbitmq-server/pull/14796 makes the new test pass.

So that fix will be available to paying RabbitMQ customers in the next round of releases, `4.0.18` and `3.13.12` (but not in this repo).

https://github.com/rabbitmq/rabbitmq-server/pull/14796 was also backported to `v4.1.x` for `4.1.8`.

gomoripeti:
> > we only test with log exchange after a release is out and only on info level
> 
> Thank you. Enabling feature flag `message_containers_deaths_v2` on the OSS `v3.13.x` branch with the log exchange at **info** level also gets stuck. Did you catch this during your manual testing?

I _think_ my colleagues might have noticed it, but we already noted the workaround that we had to disable the log exchange when enabling FFs after an upgrade to 3.13.x so that stopped any further investigation.

ansd:
> I think my colleagues might have noticed it, but we already noted the workaround that we had to disable the log exchange when enabling FFs after an upgrade to 3.13.x so that stopped any further investigation.

Thank you for letting us know. Yes, disabling the log exchange temporarily is a valid workaround (as long as some other form of logging, e.g. to stdout or to a file is configured).

PR #15297, opened 2026-01-19
logging_SUITE: Fix style issues

dumbbell:
This corresponds to the comments I would have made to #15294, but this pull request was merged before I had a chance to review it.

* Break long lines. This applies to a few long lines that were already there before #15294.
* Do not `-import()` functions as this makes the code more difficult to read and maintain.

While here, fix the copyright line: the initial year was incorrect and the line was also too long.


michaelklishin:
@mergifyio backport v4.2.x v4.1.x

PR #15320, opened 2026-01-21
rabbit_logger_exchange_h: Always declare exchange, even if exchange logging is disabled

dumbbell:
## Why

When the exchange logger handler is added, the broker is not ready to declare the exchange.

The previous implementation used a temporary process that entered a loop to wait for the broker to be ready and declare the exchange. It was possible that definitions were imported before the exchange was declared, leading to errors if bindings referenced it.

There was another problem: the exported definitions were dependent on the exchange logger configuration:
* The default vhost under which the logger exchange is declared is configurable.
* The logger exchange is only declared if the exchange logger handler is enabled in the configuration.

Because of this, the import of definitions could fail if the configuration of the node where the definitions are imported does not enable the exchange logger handler.

## How

This patch changes two things:

1. The logger exchange is always declared, regardless if exchange-based logging is enabled or not. This also allows to get rid of the polling in the setup process.

2. The logger exchange is included in the exported definitions.

This way, the exported definitions don't depend on any configuration on the nodes they are imported in: the exported definitions are self-contained and know about this logger exchange.

Part 1 could be enough to fix the problem on RabbitMQ version that include this patch. Part 2 allows to fix the problem when definitions are exported from a version with this patch and are imported on a version without.

### DIFF
--- a/deps/rabbit/src/rabbit_definitions.erl
+++ b/deps/rabbit/src/rabbit_definitions.erl
@@ -1021,10 +1021,10 @@ args(L)  -> rabbit_misc:to_amqp_table(L).
 
 list_exchanges() ->
     %% Exclude internal exchanges, they are not meant to be declared or used by
-    %% applications, except the exchange used by `rabbit_logger_exchange_h'
-    %% because its vhost depends on the configurable default vhost: we don't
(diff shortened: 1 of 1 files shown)


michaelklishin:
I agree with the idea of unconditionally declaring the `amq.rabbitmq.log` exchange. It's the most straightforward solution for definition imports that include queues and bindings for that exchange.

PR #15353, opened 2026-01-27
Fix stream SAC coordinator deadlock when deactivating consumer disconnects

acogoluegnes:
When a consumer in {connected, deactivating} state had its node
disconnect, it would transition to {disconnected, deactivating}.
This state blocked rebalancing because is_active/1 returned true,
leaving the group with no active consumer indefinitely.

Changes:
- handle_group_after_connection_node_disconnected now transitions
  {connected, deactivating} to {disconnected, waiting} and triggers
  rebalancing, since the deactivation handshake cannot complete
- handle_group_connection_presumed_down applies the same fix as
  defense in depth for {disconnected, deactivating} consumers
- handle_connection_node_disconnected updated to accumulate effects
  from group processing
- bump stream coordinator machine version to 6 and handle backward
  compatibility

Added test cases covering:
- Node disconnect during deactivation (simple and super streams)
- Connection down during deactivation
- Multiple consumers from same connection with one deactivating
- Reconnection after disconnect during deactivation
- Presume down with deactivating consumer (simple and super streams)

References rabbitmq/rabbitmq-stream-dotnet-client#447

### DIFF
--- a/deps/rabbit/src/rabbit_stream_coordinator.erl
+++ b/deps/rabbit/src/rabbit_stream_coordinator.erl
@@ -86,8 +86,8 @@
 -include_lib("kernel/include/logger.hrl").
 
 -define(REPLICA_FRESHNESS_LIMIT_MS, 10 * 1000). %% 10s
--define(V2_OR_MORE(Vsn), Vsn >= 2).
--define(V5_OR_MORE(Vsn), Vsn >= 5).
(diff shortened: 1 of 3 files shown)


kjnilsson:
NB this cannot be backported to 4.1.x as the next series 4.2 is already released.

PR #15356, opened 2026-01-28
Shovel: infra for indexed [by source] definition access, definition key extraction

michaelklishin:
This PR has two parts, both preparatory work
that was deemed necessary to have during
the 4.2.0 development cycle:

 * Definition key extraction functions
 * A new Khepri projection for indexing shovels by source

plus is a drive-by `rabbit_data_coercion:to_binary/1` fix.

## Shovels Indexed by Source

A new module, rabbit_shovel_index, provides
an ETS-backed index of shovels by their source
(a queue name or an exchange name).

This enables fast definition lookups without a full runtime parameter projection scan.

## Definition Key Access

`rabbit_shovel_definition` is a new module that extracts source info (type, queue/exchange, virutal host, protocol)
from a shovel [runtime parameter] definition.

Important: make sure to use the
Horus-compatible function versions in
Khepri projections, not all functions
in this module are Horus serialization-safe.

## rabbit_data_coercion

As a drive-by change, thsi fixes `rabbit_data_coercion:to_binary/1` which now fails with a function_clause if it
runs into a value that it cannot parse,
instead of acting as an identity function
(returning the input).

It should be a safe change to make, plus there is a non-trivial number of new tests
for `rabbit_data_coercion`.


michaelklishin:
Merged per discussion with @dcorbacho.

PR #15376, opened 2026-01-30
Bump Ra to 2.17.2

michaelklishin:
Ra [`2.17.2` release notes](https://github.com/rabbitmq/ra/releases/tag/v2.17.2).


michaelklishin:
Not setting the milestone because `4.3.0` is expected to ship with Ra 3.0, which is currently in the final stages of development, testing and integration.

PR #15384, opened 2026-02-02
Management UI: use direct path for HTTP API documentation link

majialoong:
## Proposed Changes

This change updates the HTTP API documentation link in the management UI footer from `api/` to `api/index.html`.

Previously, clicking the "HTTP API" link would trigger a server-side redirect from `/api` to `/api/index.html`. By using the direct path, we avoid this unnecessary redirect and slightly improve the user experience.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/tmpl/layout.ejs
+++ b/deps/rabbitmq_management/priv/www/js/tmpl/layout.ejs
@@ -43,7 +43,7 @@
 <div id="main"></div>
 <div id="footer">
   <ul>
-    <li><a rel="noopener noreferrer" href="api/" target="_blank">HTTP API</a></li>
+    <li><a rel="noopener noreferrer" href="api/index.html" target="_blank">HTTP API</a></li>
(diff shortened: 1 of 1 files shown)


michaelklishin:
I'm sorry but this is not an improvement we expect from the community.

A server redirect from `/api` to `/api/index.html` is invisible to the user, likely takes tiny fractions of a second, and on top of that, the `/api` endpoint (the original home of the [HTTP API docs](https://www.rabbitmq.com/docs/http-api-reference)) is simply not used particular often.

majialoong:
> I'm sorry but this is not an improvement we expect from the community.
> 
> A server redirect from `/api` to `/api/index.html` is invisible to the user, likely takes tiny fractions of a second, and on top of that, the `/api` endpoint (the original home of the [HTTP API docs](https://www.rabbitmq.com/docs/http-api-reference)) is simply not used particular often.

@michaelklishin Thanks for the explanation and for taking a look.

PR #15387, opened 2026-02-03
Differentiate AMQP 0.9.1 `basic.reject` and `basic.nack` for delivery-count

ansd:
## What?

AMQP 0-9-1 `basic.reject` now increments the AMQP 1.0 delivery-count header, while `basic.nack` does not.

Docs PR: https://github.com/rabbitmq/rabbitmq-website/pull/2444

 ## Why?

The AMQP 1.0 delivery-count header tracks failed delivery attempts, enabling poison message handling. Quorum queues discard messages when delivery-count exceeds the delivery limit (default 20).

AMQP 0-9-1 clients need a way to signal whether message processing actually failed (increment delivery-count) versus the consumer being temporarily unable to process messages (no increment).

This will become particularly important with the upcoming QQ v8 changes introduced in RabbitMQ 4.3 which allow for unlimited returns (see https://github.com/rabbitmq/rabbitmq-server/pull/13885)

AMQP 0-9-1 `basic.reject requeue=false` maps now nicely to the AMQP 1.0 rejected outcome since both increment the delivery-count.

 ## How?

The new mapping from AMQP 0.9.1 method to `rabbit_queue_type:settle/5` operation is as follows:
* basic.reject requeue=false -> discard (delivery-count incremented)
* basic.reject requeue=true  -> modify with delivery-failed=true (delivery-count incremented)
* basic.nack requeue=false   -> modify with undeliverable-here=true (delivery-count unchanged)
* basic.nack requeue=true    -> requeue (delivery-count unchanged)

### DIFF
--- a/deps/rabbit/src/rabbit_channel.erl
+++ b/deps/rabbit/src/rabbit_channel.erl
@@ -1238,7 +1238,7 @@ handle_method(#'basic.nack'{delivery_tag = DeliveryTag,
              true ->
                  requeue;
              false ->
-                 {modify, false, true, #{}}
+                 settle(false, true)
          end,
     reject(DeliveryTag, Op, Multiple, State);
 
@@ -1551,7 +1551,7 @@ handle_method(#'basic.reject'{delivery_tag = DeliveryTag, requeue = Requeue},
               _, State) ->
     Op = case Requeue of
              true ->
-                 {modify, true, false, #{}};
+                 settle(true, false);
              false ->
                  discard
          end,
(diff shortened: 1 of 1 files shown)


deps/rabbit/src/rabbit_channel.erl
```diff
@@ -1234,7 +1234,13 @@ handle_method(#'basic.publish'{exchange    = ExchangeNameBin,
 handle_method(#'basic.nack'{delivery_tag = DeliveryTag,
                             multiple     = Multiple,
                             requeue      = Requeue}, _, State) ->
-    reject(DeliveryTag, Requeue, Multiple, State);
+    Op = case Requeue of
+             true ->
+                 requeue;
+             false ->
+                 {modify, false, true, #{}}
```
michaelklishin: Should we add a comment as to what those values are? A record sounds like an overkill.

ansd: I pushed a commit which makes it more obvious what those values are.

PR #15388, opened 2026-02-03
AWS peer discovery: filter EC2 instances by state

lukebakken:
## Summary

Adds instance state filtering to the AWS peer discovery plugin to prevent dead or dying EC2 instances from being included in cluster membership during node joins. This addresses a race condition where a new node joining the cluster could sync dead nodes from existing members before a removal command completes.

## Problem

When using AWS peer discovery with AutoScaling or EC2 tags, the `ec2:DescribeInstances` API returns instances in **all states**, including `stopping`, `stopped`, `shutting-down`, and `terminated`. During instance replacement, this can cause a race condition:

1. Old instance enters `stopping` or `shutting-down` state
2. New instance starts and calls `list_nodes()`
3. `DescribeInstances` returns both old (dying) and new (running) instances
4. New node joins cluster with dead node in membership
5. Removal command issued later fails because dead node was already re-added

This race condition was observed where a node removal failed because the new node's Mnesia schema sync re-added the dead node 318ms before the removal command was issued.

## Solution

Adds a new `aws_ec2_instance_states` configuration option that filters EC2 instances by state during peer discovery. The default value is `["running", "pending"]`, which excludes dead or dying instances while including operational and starting instances.

### Configuration

**Via `rabbitmq.conf`:**
```erlang
cluster_formation.aws.ec2_instance_states.1 = running
cluster_formation.aws.ec2_instance_states.2 = pending
```

**Via environment variable:**
```bash
AWS_EC2_INSTANCE_STATES="running,pending"
```

**Via `advanced.config`:**
```erlang
[{rabbit, [
   {cluster_formation, [
       {peer_discovery_aws, [
           {aws_ec2_instance_states, [running, pending]}
       ]}
   ]}
]}].
```

### Implementation Details

- Adds `Filter.N.Name=instance-state-name` with multiple values to `ec2:DescribeInstances` API calls
- Applies to both AutoScaling mode (`get_hostname_by_instance_ids/2`) and tag-based mode (`get_hostname_by_tags/1`)
- Validates configured states against AWS EC2 valid states, filtering out invalid values with a warning
- Normalizes both atom and string inputs for consistent handling
- Filter format verified against AWS EC2 API specification and tested with live AWS API

### Testing

- Unit tests for filter building logic
- Integration tests with mocked EC2 API calls for both discovery modes
- Validation tests for both string and atom inputs
- Configuration schema tests for `rabbitmq.conf` parsing

### Backward Compatibility

- Default behavior changes to filter by `running` or `pending` state (safer default)
- Users can configure empty list `[]` to disable filtering (restore old behavior)
- Existing configurations without this setting get the safe default

### DIFF
--- a/deps/rabbitmq_peer_discovery_aws/src/rabbit_peer_discovery_aws.erl
+++ b/deps/rabbitmq_peer_discovery_aws/src/rabbit_peer_discovery_aws.erl
@@ -30,6 +30,8 @@
 
 -define(BACKEND_CONFIG_KEY, peer_discovery_aws).
 
+-define(VALID_EC2_INSTANCE_STATES, ["pending", "running", "shutting-down", "terminated", "stopping", "stopped"]).
+
 -define(CONFIG_MAPPING,
          #{
           aws_autoscaling                    => #peer_discovery_config_entry_meta{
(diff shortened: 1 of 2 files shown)


lukebakken:
Thank you @michaelklishin

PR #15390, opened 2026-02-03
rabbit_osiris_metrics: Use committed_offset counter

the-mikedavis:
This change uses the committed offset as the last offset in the calculation of available offsets rather than the last-written offset. Stream readers can't access tail data until it becomes committed so this is a more accurate count for "ready" messages.

This is meant to extend https://github.com/rabbitmq/rabbitmq-server/pull/15225 so it's for v4.3.x only.

### DIFF
--- a/deps/rabbit/src/rabbit_osiris_metrics.erl
+++ b/deps/rabbit/src/rabbit_osiris_metrics.erl
@@ -57,10 +57,10 @@ handle_cast(_Request, State) ->
 handle_info(tick, #state{timeout = Timeout} = State) ->
     Data = osiris_counters:overview(),
     _ = maps:map(
-      fun ({osiris_writer, QName}, #{offset := Offs,
+      fun ({osiris_writer, QName}, #{committed_offset := COffs,
                                      first_offset := FstOffs}) ->
-              COffs = Offs + 1 - FstOffs,
-              rabbit_core_metrics:queue_stats(QName, COffs, 0, COffs, 0),
+              Offsets = COffs + 1 - FstOffs,
+              rabbit_core_metrics:queue_stats(QName, Offsets, 0, Offsets, 0),
               Infos = try
                           %% TODO complete stats!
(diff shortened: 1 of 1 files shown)


deps/rabbit/src/rabbit_osiris_metrics.erl
```diff
@@ -57,10 +57,10 @@ handle_cast(_Request, State) ->
 handle_info(tick, #state{timeout = Timeout} = State) ->
     Data = osiris_counters:overview(),
     _ = maps:map(
-      fun ({osiris_writer, QName}, #{offset := Offs,
+      fun ({osiris_writer, QName}, #{committed_offset := COffs,
```
the-mikedavis, line 60: Do we need to check the feature flag and use it to decide whether we should use `offset` or `committed_offset`?

michaelklishin: If those values can come from other nodes, we do. If they are entirely local state, maybe not.

acogoluegnes: It should not crash as `committed_offset` is always available, but the stats will be slightly inaccurate before the changes in #15225 are enforced with the 4.3 feature flag. `committed_offset` is the offset of the last committed chunk pre-15225, so it is behind by the number of messages in the chunk compared to the actual last offset.

This should fix itself when the upgrade is done and after the streams restart, so I guess it is reasonable considering it is just stats.

WDYT @kjnilsson ?

kjnilsson: yeah it is mostly informational at this point so shouldn't matter, let's not bother the ff subsystem if we don't have to.

the-mikedavis:
I looked into the tests and I think this makes some cases where we check message count flakier. I didn't see a specific test failing reliably - it made a few cases more likely to fail. I will try to give this another look at some point but there's no reason to rush it into 4.3.0

PR #15392, opened 2026-02-04
Khepri: Mark `khepri_db` as required

dumbbell:
## Why

This is the next step in Khepri adoption. Khepri will become the only supported metadata store in RabbitMQ 4.3.0.

## How

The `khepri_db` feature flag is marked as required. It is a "soft" required feature flag which means that it will be automatically enabled if a cluster upgraded to RabbitMQ 4.3+ is still running Mnesia.

As of this commit, the Mnesia-related code is still there. It will be deleted separately. We will just keep the minimum to allow the migration from Mnesia to Khepri.


mkuratczyk:
I have found issues that we need to resolve first. To be continued

dumbbell:
There are a lot of changes to make to the testsuites and the GitHub Actions workflow as well.

PR #15394, opened 2026-02-04
rabbitmq_cli: change tmpdir to rabbitmq_cli/logs for background node

dcorbacho:
This allows CI to collect logs as for any other plugin
Also all test changes are contained inside the plugin directory


deps/rabbit_common/mk/rabbitmq-run.mk
```diff
@@ -14,7 +14,9 @@ exec_verbose = $(exec_verbose_$(V))
 ifeq ($(PLATFORM),msys2)
 TEST_TMPDIR ?= $(TEMP)/rabbitmq-test-instances
 else
+ifndef TMPDIR
 TMPDIR ?= /tmp
+endif
```
lhoguin: My bad, ifndef isn't necessary because ?= already sets the value only if it's not already set. So the change in this file is not needed.

dcorbacho: Mine too! I didn't realised either when I found TMPDIR, looking backwards from RABBITMQ_LOG_BASE which cannot be overridden.

PR #15398, opened 2026-02-04
Management UI: Add autocomplete, autofocus and label linking to input at login page.

simi:
## Proposed Changes

The only purpose of the login page is to put login as password thus first action is to get username focus. This PR adds auto focus to this input. Also as a side effect updates the autocomplete attributes and adds ids to inputs to link labels properly to input, which makes label clickable to focus the input also (just minor tweak).

UI should be visually 100% the same.

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/tmpl/login.ejs
+++ b/deps/rabbitmq_management/priv/www/js/tmpl/login.ejs
@@ -5,12 +5,12 @@
     <div id="login-status"></div>
     <table class="form">
       <tr>
-        <th><label>Username:</label></th>
-        <td><input type="text" name="username"/><span class="mand">*</span></td>
+        <th><label for="username">Username:</label></th>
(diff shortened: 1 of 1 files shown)


michaelklishin:
@simi thank you for considering a contribution.

We can only accept it if you sign [our CLA](https://github.com/rabbitmq/cla), which is done online and should not take long.

simi:
@michaelklishin done!

simi:
the selenium error is related to PR from fork

> Error: google-github-actions/auth failed with: the GitHub Action workflow must specify exactly one of "workload_identity_provider" or "credentials_json"! If you are specifying input values via GitHub secrets, ensure the secret is being injected into the environment. By default, secrets are not passed to workflows triggered from forks, including Dependabot.

the other failure seems not related (maybe flaky test?)

michaelklishin:
Yup, out of the thousands tests in our suites, some are flaky.

Selenium tests require access to a secret.

PR #15400, opened 2026-02-05
rabbit_quorum_queue:stat/2 use find_leader/1

mkuratczyk:
This should deflake:
make -C deps/rabbit ct-dynamic_qq t=cluster_size_3:takeover_on_failure

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -1168,17 +1168,16 @@ stat(Q) when ?is_amqqueue(Q) ->
 -spec stat(amqqueue:amqqueue(), non_neg_integer()) -> {'ok', non_neg_integer(), non_neg_integer()}.
 
 stat(Q, Timeout) when ?is_amqqueue(Q) ->
-    Leader = amqqueue:get_pid(Q),
     try
-        case rabbit_fifo_client:stat(Leader, Timeout) of
-          {ok, _, _} = Success -> Success;
-          {error, _}           -> {ok, 0, 0};
-          {timeout, _}         -> {ok, 0, 0}
+        maybe
+            Leader ?= find_leader(Q),
+            {ok, _, _} = Result ?= rabbit_fifo_client:stat(Leader, Timeout),
+            Result
+        else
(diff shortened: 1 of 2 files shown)


michaelklishin:
@mergifyio backport v4.2.x

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1168,17 +1168,16 @@ stat(Q) when ?is_amqqueue(Q) ->
...
 
 stat(Q, Timeout) when ?is_amqqueue(Q) ->
-    Leader = amqqueue:get_pid(Q),
     try
-        case rabbit_fifo_client:stat(Leader, Timeout) of
-          {ok, _, _} = Success -> Success;
-          {error, _}           -> {ok, 0, 0};
-          {timeout, _}         -> {ok, 0, 0}
+        maybe
+            Leader ?= find_leader(Q),
```
ansd, line 1173: This doesn't look correct to me.
If `find_leader(Q)` returns `undefined`, function `stat/2` should probably return `{ok, 0, 0}`?
What do you think @mkuratczyk @michaelklishin @kjnilsson ?

kjnilsson: you're right, ill fix it in my current pending changes.

kjnilsson: oh actually it is in main, we probably need a separate PR for the `{_, _} = Leader` change

ansd: Actually `rabbit_fifo_client:stat(undefined, Timeout)` will probably crash. The exception will be caught and `{ok, 0, 0}` will be returned.

So, either use Karl's suggestion (better) or just use `=` instead of `?=` in this line.

mkuratczyk: thanks for catching this. solved in https://github.com/rabbitmq/rabbitmq-server/pull/15423

ansd: Thank you!

PR #15411, opened 2026-02-06
Fix message store crash when file deleted during concurrent GC

lukebakken:
During high load operations (such as maintenance drain with heavy queue leadership transfers), the message store can crash with `{case_clause, {false, []}}` in `write_action/3` at line 1121.

This occurs when:
1. A message exists in the index with `ref_count=0` referencing file N
2. File N is marked for deletion (`valid_total_size=0`)
3. The garbage collector asynchronously deletes file N from `FileSummaryEts`
4. A write operation attempts to look up file N
5. `ets:lookup(FileSummaryEts, N)` returns `[]`
6. No case clause matches `{false, []}`

The crash cascades to queue processes attempting to write messages.

## Solution

Add two new case clauses to `write_action/3` to handle files already deleted by GC:

- `{false, []}` - Delete stale index entry and write fresh copy to current file
- `{false_if_increment, []}` - Ignore write when client is dying

This follows the same pattern as existing clauses that handle files being deleted or locked. Since `ref_count=0`, the old copy is orphaned and should be safe to discard.

### DIFF
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -1122,7 +1122,20 @@ write_action({Mask, #msg_location { ref_count = 0, file = File,
         {_Mask, [#file_summary {}]} ->
             ok = index_update_ref_counter(IndexEts, MsgId, +1), %% Effectively set to 1.
             State1 = adjust_valid_total_size(File, TotalSize, State),
-            {confirm, File, State1}
+            {confirm, File, State1};
+        {false, []} ->
+            %% Handle case where GC has deleted the file from the summary table
+            %% before we could look it up. This can occur during high load when
+            %% the GC asynchronously deletes files while write operations are
+            %% in progress. Since ref_count=0, the old copy is orphaned and
+            %% safe to discard. Delete stale index entry and write fresh copy
+            %% to current file.
+            ok = index_delete(IndexEts, MsgId),
+            {write, State};
+        {false_if_increment, []} ->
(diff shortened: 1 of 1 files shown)


lukebakken:
I appreciate the review @lhoguin - thanks!

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1122,7 +1122,20 @@ write_action({Mask, #msg_location { ref_count = 0, file = File,
...
-            {confirm, File, State1}
+            {confirm, File, State1};
+        {false, []} ->
+            %% Handle case where GC has deleted the file from the summary table
+            %% before we could look it up. This can occur during high load when
+            %% the GC asynchronously deletes files while write operations are
+            %% in progress. Since ref_count=0, the old copy is orphaned and
+            %% safe to discard. Delete stale index entry and write fresh copy
+            %% to current file.
+            ok = index_delete(MsgId, State),
```
michaelklishin: `index_delete/2` accepts an ETS table handle as the first argument. There are examples [in the same function](https://github.com/amazon-mq/upstream-to-rabbitmq-server/blob/e1fb5b7cfe3ca5256dc7b4d082861035c96379ba/deps/rabbit/src/rabbit_msg_store.erl#L1110-L1115).

So `index_delete(MsgId, State)` will fail with another exception.

lukebakken: 😮‍💨 🤦‍♂️

PR #15412, opened 2026-02-06
Show oldest stream message timestamp in management UI

the-mikedavis:
This change shows the `first_timestamp` counter's value as a detail in the management UI as a formatted timestamp. This is somewhat useful for attaching offset listeners by timestamp - you know the oldest timestamp in the stream you can start from.

<details><summary>Example...</summary>

oldest-message

</details>

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/formatters.js
+++ b/deps/rabbitmq_management/priv/www/js/formatters.js
@@ -92,10 +92,16 @@ function fmt_date0(d) {
 }
 
 function fmt_timestamp(ts) {
+    if (ts == undefined) {
+        return UNKNOWN_REPR;
+    }
     return fmt_date(new Date(ts));
 }
 
 function fmt_timestamp_mini(ts) {
+    if (ts == undefined) {
+        return UNKNOWN_REPR;
+    }
     return fmt_date_mini(new Date(ts));
 }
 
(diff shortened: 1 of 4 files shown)


deps/rabbitmq_management/priv/www/js/tmpl/stream-queue-stats.ejs
```diff
@@ -28,6 +28,10 @@
         <th>Segments</th>
         <td><%= fmt_string(queue.segments) %></td>
       </tr>
+      <tr>
+        <th>Oldest message</th>
```
the-mikedavis, line 32: "Oldest message" ? "First timestamp" ? I'm not attached to the phrasing here

deps/rabbitmq_management/priv/www/js/formatters.js
```diff
@@ -92,10 +92,16 @@ function fmt_date0(d) {
...
 function fmt_timestamp(ts) {
+    if (ts == undefined) {
+        return UNKNOWN_REPR;
+    }
     return fmt_date(new Date(ts));
 }
 
 function fmt_timestamp_mini(ts) {
+    if (ts == undefined) {
+        return UNKNOWN_REPR;
```
the-mikedavis, line 103: These changes show the dimmed "?" when the page is waiting for data to arrive. Otherwise these fmts would show "NaN-NaN-NaN"

deps/rabbitmq_management/priv/www/js/tmpl/stream-queue-stats.ejs
michaelklishin: "Oldest" is unambiguous while "first" can be interpreted as the head or the tail, so I'd use "oldest" personally.

acogoluegnes (approved):

PR #15434, opened 2026-02-10
Incorporate mgt UI changes required by stream browser

MarcialRosales:
## Proposed Changes

These changes are required by two PRs: one in rabbitmq-server that adds more UI support for superstream  and another one in tanzu-rabbitmq that adds the stream browser.

- Refactor get-message section to a template
- Add mechanism to register PostProcessor(s) and call them from postprocess function 
- Add subscribe-notify mechanism for vhost changed, refresh and tab activated events
- Allow ui extensions to declare an array of javascript and css files or just one javascript/css file
- Export function (is_authorized/5) required by some new endpoints in Tanzu

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/global.js
+++ b/deps/rabbitmq_management/priv/www/js/global.js
@@ -1131,8 +1131,12 @@ function unregisterPostProcessor(name) {
 } 
 function invokeRegisteredPostProcessors() {
   for (const [name, processorFun] of current_postprocessors) {
-    console.log(`Calling postprocessor ${name}`);
-    processorFun();  
+    console.debug(`Calling postprocessor ${name}`);
+    try {
+      processorFun();
+    } catch (err) {
+      console.error(`PostProcessor ${name} failed due to ${err}`);
+    }
   }
 }
 


deps/rabbitmq_management/priv/www/js/global.js
```diff
@@ -1110,3 +1113,63 @@ var BINARY_STATISTICS = {
...
+  }
+}
+function notifyOnVhostChange(newVhost) {
+  for (const [_name, listener] of applicationListeners) {
+    listener.onVhostChange(newVhost);  
+  }
+}
+function notifyActivatedTab(tab) {
+  for (const [_name, listener] of applicationListeners) {
+    listener.onTabActivated(tab);
```
michaelklishin: This uses `onTabActivated` but `ApplicationListener` above only provides a function named `onTabDeactivated`.

deps/rabbitmq_management/priv/www/js/main.js
```diff
@@ -235,19 +238,83 @@ function update_vhosts() {
...
+                        javascript_files.push(extension.javascript[j]);
+                    }
+                } else {
+                    javascript_files.push(extension.javascript);
+                }
+            } 
+            if (extension.hasOwnProperty('css')) {
+                dynamic_css_load(extension.css);                
+            }
             extension_count++;
```
michaelklishin: In this version, a CSS-only extension will prevent `dispatcher_modules.length == extension_count` from evaluating to true, as far as I can tell.

We only have one such plugin (themes) but still.

deps/rabbitmq_management/priv/www/js/main.js
```diff
@@ -1341,9 +1416,11 @@ function with_req(method, path, body, fun) {
             if (ix != -1) {
                 outstanding_reqs.splice(ix, 1);
             }
-            if (check_bad_response(req, true)) {
+            if (check_bad_response(req, full_page_404)) {
```
michaelklishin: This will both display a generic 404 message plus run the callback. Looks like `check_bad_response` needs to be extended to suppress the standard pop-up if we want to run some code instead (and that code can always display a pop-up if needed).

So, is the idea to allow both "handles" to execute or just the `full_page_404` one?

deps/rabbitmq_management/priv/www/js/global.js
MarcialRosales: Well spotted! Thanks. The stream browser does not use yet this callback but it will do it very soon to do some clean-up. I initially had both events: onTabActivated and onTabDeactivated. But then realized that I would only need one, `onTabActivated`. Commit coming.

deps/rabbitmq_management/priv/www/js/main.js
MarcialRosales: umm. i am not seeing what you suggest... An extension may have zero or many javascripts and it is handled on the first `if` statement. And an extension may have zero or many css files and that is handled on the second `if` statement. And regardless whether the extension has or not js or css file, the `extension_count` is incremented by one. Thus, in theory, `dispatcher_modules.length == extension_count` will always evaluate to true...  

```
          if (extension.hasOwnProperty('javascript')) {
                // Collect JavaScript files for sequential loading
                if (Array.isArray(extension.javascript)) {
                    for (var j = 0; j < extension.javascript.length; j++) {
                        javascript_files.push(extension.javascript[j]);
                    }
                } else {
                    javascript_files.push(extension.javascript);
                }
            } 
            if (extension.hasOwnProperty('css')) {
                dynamic_css_load(extension.css);                
            }
            extension_count++;
```

deps/rabbitmq_management/priv/www/js/main.js
MarcialRosales: You are right but it turns out that the http error message for 404 is not "not_found" but "Not Found" therefore, it never runs the `show_popup` branch. It only runs the branch on line 1423. 
The reason for this is because most of the existing management endpoints returns 404 along with the error message "not_found". Whereas the streambrowser's endpoint returns 404 via the cowboy `resource_exists` function/callback. So, cowboy returns `Not Found`. 
But you are right that the `check_bad_response` should be extended so that it can be configured to either do full_page, or show_popup , or call an on404 callback.

MarcialRosales: i have pushed a change where 404 is entirely handled within `check_bad_response` function.

deps/rabbitmq_management/priv/www/js/main.js
michaelklishin: Alright. @MarcialRosales do you think we can move `extension_count++` to the `if (extension.hasOwnProperty('javascript'))` block?

MarcialRosales: I changed the code so that we call `start_app()` only when all javascript(s) have been loaded. Regardless whether the javascript is a dispatcher module (i.e. one that has a `dispatcher_add` function call).

deps/rabbitmq_management/priv/www/js/main.js
```diff
@@ -1446,17 +1523,26 @@ function check_bad_response(req, full_page_404) {
...
+        if (    error == 'bad_request' || 
+                error == 'not_found'  || 
+                reason == 'Not Found' || 
+                error == 'not_authorised' || 
+                error == 'not_authorized') {
             if ((req.status == 401 || req.status == 403) && oauth.enabled) {
-              initiate_logout(oauth, reason);
+                initiate_logout(oauth, reason);
+            } else if (on404fun && (typeof on404fun === 'function') && req.status == 404) {
+                on404fun(JSON.parse(req.responseText));
```
michaelklishin: Can we reuse the `response` value above?

deps/rabbitmq_management/priv/www/js/global.js
```diff
@@ -1110,3 +1110,63 @@ var BINARY_STATISTICS = {
...
```
michaelklishin: This does not handle exceptions, so the entire sequence would stop if one listener throws. This could be acceptable, I trust your judgement :)

deps/rabbitmq_management/priv/www/js/main.js
MarcialRosales: Absolutely! And do not need to be sorry. You are making me and the code better with your comments.

deps/rabbitmq_management/priv/www/js/global.js
MarcialRosales: This is something that I am currently doing it in the streambrowser code, i.e.to follow a more defensive programming style. I am looking forward to migrating to typescript to delegate part of it to the language.
I have pushed the changes.

PR #15457, opened 2026-02-12
4.3: Mark amqp_address_v1 as denied_by_default

lhoguin:


deps/rabbitmq_shovel/src/rabbit_amqp10_shovel.erl
```diff
@@ -517,3 +539,18 @@ validate_amqp10_map(Name, Terms0) ->
...
+decl_queue(Sess, Addr, false) ->
+    case re:run(Addr, "(^/queues/)(.*)", [{capture, [2], binary}]) of
+        nomatch ->
+            ok;
+        {match, [QName]} ->
+            {ok, LinkPair} = rabbitmq_amqp_client:attach_management_link_pair_sync(Sess, <<"mgmt link pair">>),
+            Args = #{},
+            %% Best effort to declare queue, if something fails the next connection
+            %% steps will detect the failure
+            _ = rabbitmq_amqp_client:declare_queue(LinkPair, QName, #{arguments => Args}),
```
ansd: Unquoting the queue name is missing.
I suggest calling `rabbitmq_amqp_address:to_map(Addr)` instead of this regex parsing.

ansd (changes requested):

deps/rabbitmq_shovel/src/rabbit_amqp10_shovel.erl
lhoguin: Done.

PR #15459, opened 2026-02-12
4.3: Mark transient_nonexcl_queues as denied_by_default

lhoguin:


deps/amqp_client/src/amqp_rpc_server.erl
```diff
@@ -71,7 +71,7 @@ stop(Pid) ->
 init([Connection, Q, Fun]) ->
     {ok, Channel} = amqp_connection:open_channel(
                         Connection, {amqp_direct_consumer, [self()]}),
-    amqp_channel:call(Channel, #'queue.declare'{queue = Q}),
+    amqp_channel:call(Channel, #'queue.declare'{queue = Q, durable = true}),
```
lhoguin, line 74: Possible breaking change.

michaelklishin: Yup, the default in earlier versions would be `default = false`. But that's OK, that RPC Erlang server is unlikely to be used even by many Erlang users.

deps/rabbitmq_shovel/src/rabbit_local_shovel.erl
```diff
@@ -747,7 +747,9 @@ check_fun(QName, VHost, User) ->
 src_decl_exchange(SrcX, SrcXKey, VHost, User) ->
     Methods = [#'queue.bind'{routing_key = SrcXKey,
                              exchange    = SrcX},
-               #'queue.declare'{exclusive = true}],
+               %% Local can't be exclusive as it doesn't provide
+               %% a connection pid
+               #'queue.declare'{durable = true}],
```
michaelklishin: Then we need to use a stable name here, e.g. the shovel name prefixed with `shovel.` or something like that.

PR #15480, opened 2026-02-16
rabbit_db: Fix handling of async m2k puts

dumbbell:
## Why

Several `rabbit_db_*_m2k_converter` modules use async puts to Khepri to parallelize writes and make them faster. This is handled by the `rabbit_db_m2k_converter:with_correlation_id/2` function.

Ra will send back a Ra event message for each async write to reply with the result of the write. The problem is that the code behind `with_correlation_id/2` has a logic issue: it only consumes 1 Ra event every 64 async writes. It will consume all of them at the end of the overall, but meeawhile, messages will accumulate. If there are a lot of Mnesia records to copy, this can fill the process mailbox and slow it down significantly.

## How

The logic is changed to regularily consume all messages in the mailbox with a timeout of 0. Only at the end, remaining messages are consumed with an infinity timeout because we know how many messages we are expecting (the actual timeout is handled by the write itself).


mkuratczyk (approved):
thanks! I see a significant speed up

PR #15483, opened 2026-02-16
Filter internal runtime parameters from definitions export

lukebakken:
`rabbitmqctl export_definitions` could possibly include internal runtime parameters (`exchange-delete-in-progress` and `imported_definition_hash_value`) in the exported definitions. These are implementation details that shouldn't appear in exported definitions.

During testing of RabbitMQ (killing nodes) with in-progress exchange operations, I've seen `exchange-delete-in-progress` remain in the parameters even though it should be cleaned up. My guess is that it has something to do with timing.

This adds a blocklist of internal runtime parameter components to `rabbit_definitions:list_runtime_parameters/0`. The filter is in `rabbit_definitions` rather than `rabbit_runtime_parameters:list/0` because the lower-level function is used by other callers that may legitimately need to see all runtime parameters.

Since the list of internal runtime parameters rarely changes, I think that a hard-coded list is fine for this feature.

The `imported_definition_hash_value` component name is moved from a local macro in `rabbit_definitions_hashing` to a shared macro in `rabbit.hrl` alongside the existing `EXCHANGE_DELETE_IN_PROGRESS_COMPONENT`.

### DIFF
--- a/deps/rabbit/src/rabbit_definitions.erl
+++ b/deps/rabbit/src/rabbit_definitions.erl
@@ -1112,7 +1112,7 @@ user_definition(User) ->
 %%   - rabbit_definitions_hashing (imported_definition_hash_value)
 -define(INTERNAL_RUNTIME_PARAMETER_COMPONENTS, [
     ?EXCHANGE_DELETE_IN_PROGRESS_COMPONENT,
-    <<"imported_definition_hash_value">>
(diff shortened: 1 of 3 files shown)


the-mikedavis (approved):

lukebakken:
Thank you @michaelklishin

PR #15498, opened 2026-02-17
`rabbit_msg_store`: terminate GC with exit signal during shutdown

lukebakken:
## Problem

When `rabbit_msg_store` shuts down, its `terminate/2` callback calls `rabbit_msg_store_gc:stop/1`, which sends a `gen_server2:call(stop, infinity)` to the GC process. If the GC is blocked mid-`handle_cast` on disk I/O (for example, during compaction under disk pressure), the call blocks indefinitely. `terminate/2` never reaches the code that writes the recovery files (`file_summary.ets`, `msg_store_index.ets`, `clean.dot`). Eventually the msg_store child's supervisor shutdown timeout expires (`msg_store_shutdown_timeout`, default 600s) and the supervisor kills the msg_store process. With no recovery files on disk, the next startup rebuilds indices from scratch by scanning every segment file, which is slow and expensive for large stores.

We hit this in production on a broker under PerfTest load. The persistent message store for the `/` vhost logged "Stopping message store" and then nothing for 10 minutes until the supervisor killed it with reason `killed`. The GC process was blocked on disk I/O while disk free space was hovering near the 2 GiB limit. On restart, the store logged "rebuilding indices from scratch" despite the shutdown having been initiated gracefully via `rabbitmqctl stop`.

## Fix

Replace the synchronous `rabbit_msg_store_gc:stop/1` call in `terminate/2` with a new `stop_gc/2` helper that monitors the GC, sends `exit(GcPid, shutdown)`, and waits for the `'DOWN'` message.

`rabbit_msg_store_gc` does not trap exits, so an exit signal terminates it immediately even if it is running inside a `handle_cast` callback on disk I/O. Exit signals are processed by the scheduler, not the user-level receive loop, so the signal preempts the blocked callback. The GC's `terminate/2` is a no-op, so bypassing it has no side effect.

The wait for `'DOWN'` is bounded by `max(msg_store_shutdown_timeout - 60_000, 5_000)` so that `terminate/2` stays within the msg_store child's own supervisor shutdown timeout and leaves at least 60s for the remaining steps (syncing the current file, writing the file summary, tearing down ETS, writing recovery terms). If the shutdown signal does not produce a `'DOWN'` in time, fall back to `exit(GCPid, kill)`. `kill` cannot be trapped, so the inner `receive` has no timeout. A warning is logged on the fallback path so the case can be tracked in production logs.

Killing the GC mid-operation is safe with respect to message data:

- `compact_file` copies messages before updating the index, and the original data remains on disk until truncation. The code comments confirm: "it's OK if we crash at any point before we update the index because the old data is still there until we truncate."
- `truncate_file` only removes data that has already been compacted to earlier offsets.
- `delete_file` only deletes files with zero valid messages, enforced by assertions before the delete.

The unclean recovery path (`build_index/3`) rebuilds everything from the actual segment files on disk using `scan_file_for_valid_messages`, so any inconsistency between the file summary and the on-disk state is handled. In the common case (GC killed before it modified the file summary ETS), the recovery files are fully consistent and the next startup recovers cleanly without a rebuild.

## Tests

Two test cases in `backing_queue_SUITE` exercise the fix:

- `msg_store_gc_stuck_suspended` suspends the GC with `sys:suspend`, terminates the persistent store via the supervisor, and verifies that the store recovers cleanly (`successfully_recovered_state` returns `true`) with all messages intact. This covers the case where the GC is blocked in its receive loop.
- `msg_store_gc_stuck_mid_callback` mocks `compact_file/2` to block indefinitely inside the `handle_cast` callback, sends a compact cast to the GC to put it in the blocking callback, then terminates the store via the supervisor. This covers the scenario that motivated the PR: a GC stuck on disk I/O inside a callback.

Both tests confirm that `exit(GcPid, shutdown)` terminates the GC process regardless of what Erlang code it is running, that `terminate/2` completes and writes the recovery files, and that the next startup recovers cleanly.

### DIFF
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -10,6 +10,7 @@
 -behaviour(gen_server2).
 
 -export([start_link/5, successfully_recovered_state/1,
+         gc_pid/1,
          client_init/3, client_terminate/1, client_delete_and_terminate/1,
          client_pre_hibernate/1, client_ref/1,
          write/4, write_flow/4, read/2, read_many/2, contains/2, remove/2]).
@@ -398,6 +399,11 @@ start_link(VHost, Type, Dir, ClientRefs, StartupFunState) when is_atom(Type) ->
 
 successfully_recovered_state(Server) ->
     gen_server2:call(Server, successfully_recovered_state, infinity).
+
+-spec gc_pid(server()) -> pid().
+
+gc_pid(Server) ->
+    gen_server2:call(Server, gc_pid, infinity).
 
 -spec client_init(server(), client_ref(), maybe_msg_id_fun()) -> client_msstate().
 
@@ -806,6 +812,7 @@ init([VHost, Type, BaseDir, ClientRefs, StartupFunState]) ->
 prioritise_call(Msg, _From, _Len, _State) ->
     case Msg of
         successfully_recovered_state                        -> 7;
+        gc_pid                                              -> 7;
         {new_client_state, _Ref, _Pid, _MODC}               -> 7;
         _                                                   -> 0
     end.
@@ -825,6 +832,9 @@ prioritise_info(Msg, _Len, _State) ->
 
 handle_call(successfully_recovered_state, _From, State) ->
     reply(State #msstate.successfully_recovered, State);
+
+handle_call(gc_pid, _From, State) ->
+    reply(State #msstate.gc_pid, State);
 
 handle_call({new_client_state, CRef, CPid, MsgOnDiskFun}, _From,
             State = #msstate { dir                = Dir,
@@ -971,9 +981,13 @@ terminate(Reason, State = #msstate { index_ets           = IndexEts,
         _ -> {" with reason ~0p", [Reason]}
     end,
     ?LOG_INFO("Stopping message store for directory '~ts'" ++ ExtraLog, [Dir|ExtraLogArgs]),
-    %% stop the gc first, otherwise it could be working and we pull
-    %% out the ets tables from under it.
-    ok = rabbit_msg_store_gc:stop(GCPid),
+    %% Terminate the GC first, otherwise it could still be running and we
+    %% pull the ETS tables out from under it. The GC does not trap exits,
+    %% so an exit signal terminates it immediately even if it is stuck
+    %% mid-callback on disk I/O. Bound the wait so terminate stays within
+    %% the msg_store child's own supervisor shutdown timeout, and fall
+    %% back to kill if the shutdown signal does not take effect in time.
+    stop_gc(GCPid),
     State3 = case CurHdl of
                  undefined -> State;
                  _         -> State2 = internal_sync(State),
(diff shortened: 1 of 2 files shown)


lukebakken:
Leaving as draft until I can re-reproduce the issue without this fix, then verify this fix, in a "real" environment. Early reviews are welcome, of course 😸

michaelklishin:
@lhoguin can you please take a quick look? Thank you.

lukebakken:
No hurry because I still am working on this PR and testing it.

lhoguin:
Perhaps just change `rabbit_msg_store_gc` to do `exit(GcPid, shutdown)` instead of sending a message. It'll stop faster for everyone.

themavik (commented):
terminate/2 now funnels GC shutdown through stop_gc/2: rabbit_msg_store_gc:stop/2 honors a timeout derived from msg_store_shutdown_timeout, then kills the stuck pid so recovery files still flush. backing_queue_SUITE:msg_store_gc_stop_timeout with sys:suspend on the GC exercises the timeout branch.

lukebakken:
Hi Loïc, the genie and I looked into this and I want to make sure I understand your suggestion correctly.

`rabbit_msg_store_gc` runs as a `gen_server2`, which sets `trap_exit = true`. We dug into the OTP signal handling code (`erts/emulator/beam/erl_proc_sig_queue.c`, OTP-26.2.5.11, around line 4266) and confirmed that when a process has `F_TRAP_EXIT` set, any exit signal except `kill` is converted to a message in the process mailbox:

```c
if ((op != ERTS_SIG_Q_OP_EXIT || reason != am_kill)
    && (c_p->flags & F_TRAP_EXIT)) {
    convert_prepared_sig_to_msg(c_p, sig, xsigd->message, next_nm_sig);
```

So `exit(GcPid, shutdown)` doesn't immediately terminate the GC - it just puts a message in its mailbox, the same as the `stop` call does. If the GC is blocked mid-callback on disk I/O, the signal sits in the mailbox until the callback returns, which is exactly the scenario we're trying to fix.

Only `exit(GcPid, kill)` bypasses `trap_exit` and causes immediate termination, which is what the PR already falls back to after the timeout expires.

Am I missing something in your suggestion?

---

**Addendum:** The analysis above is incorrect. It was produced by Claude (an AI assistant) and I posted it without verifying the load-bearing claim. `rabbit_msg_store_gc` does **not** set `trap_exit`:

- `deps/rabbit/src/rabbit_msg_store_gc.erl` contains no `process_flag` calls; `init/1` does not set `trap_exit`.
- `deps/rabbit_common/src/gen_server2.erl` has exactly one `process_flag(trap_exit, true)` call, and it is inside the spawned middleman process in `do_multi_call/4` — not on the `init_it/6` path used by normal `gen_server2` processes.

So @lhoguin is right: `exit(GcPid, shutdown)` would terminate the GC immediately regardless of what it is doing inside a callback. Apologies for the noise — I'll reconsider the approach in this PR.

lhoguin:
> Am I missing something in your suggestion?

I don't think it sets `trap_exit`? The only reference to `trap_exit` is in a temporary process to do multi-calls.

lukebakken:
Thanks again @lhoguin for taking a look.

lhoguin (approved):

lhoguin:
Thank you!!

PR #15499, opened 2026-02-18
Added configuration options to disable Prometheus metrics and endpoints

kenliao94:
## Proposed Changes

This PR adds configuration options to the rabbitmq_prometheus plugin that allow operators to:

Disable specific Prometheus endpoints - New configuration flags to disable the /metrics/per-object, /metrics/detailed, and /metrics/memory-breakdown endpoints individually:

```
disable_per_object_endpoint (default: true)
disable_detailed_endpoint (default: false)
disable_memory_breakdown_endpoint (default: false)
```

Disable individual metrics - A new disabled_metrics configuration option that accepts a list of metric names to exclude from scrape responses. The implementation normalizes metric names by stripping rabbitmq_, rabbitmq_detailed_, and rabbitmq_cluster_ prefixes, allowing users to specify metrics in any format.

Why this is useful:
- Reduces scrape payload size when certain metrics are not needed
- Improves scrape performance by skipping unnecessary metric collection
- Allows operators to disable expensive endpoints (like per-object) in large deployments
- Provides fine-grained control over which metrics are exposed

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

### DIFF
--- a/deps/rabbitmq_prometheus/priv/schema/rabbitmq_prometheus.schema
+++ b/deps/rabbitmq_prometheus/priv/schema/rabbitmq_prometheus.schema
@@ -181,3 +181,23 @@ end}.
 %% Authentication options ========================================================
 {mapping, "prometheus.authentication.enabled", "rabbitmq_prometheus.authentication.enabled",
     [{datatype, boolean}]}.
+
+%% Endpoint disable options ========================================================
+{mapping, "prometheus.disable_memory_breakdown_endpoint", "rabbitmq_prometheus.disable_memory_breakdown_endpoint",
+    [{datatype, {enum, [true, false]}}, {default, false}]}.
+
+{mapping, "prometheus.disable_per_object_endpoint", "rabbitmq_prometheus.disable_per_object_endpoint",
+    [{datatype, {enum, [true, false]}}, {default, true}]}.
+
+{mapping, "prometheus.disable_detailed_endpoint", "rabbitmq_prometheus.disable_detailed_endpoint",
+    [{datatype, {enum, [true, false]}}, {default, false}]}.
+
+%% Metric blocklist options ========================================================
+{mapping, "prometheus.disabled_metrics.$metric", "rabbitmq_prometheus.disabled_metrics",
+    [{datatype, atom}]}.
+
+{translation, "rabbitmq_prometheus.disabled_metrics",
+fun(Conf) ->
+    Settings = cuttlefish_variable:filter_by_prefix("prometheus.disabled_metrics", Conf),
+    [V || {_, V} <- Settings]
+end}.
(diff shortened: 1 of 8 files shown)


lukebakken (changes requested):
Thank you for this contribution. After reviewing the changes, I have the following feedback:

**Breaking default for `disable_per_object_endpoint`**

The schema sets `{default, true}` for `disable_per_object_endpoint`, which silently disables `/metrics/per-object` for all existing deployments. This endpoint is documented as always accessible regardless of `return_per_object_metrics`, so changing its availability by default is a breaking change that will affect users who scrape it directly. Please change the default to `false` so that existing behavior is preserved and operators must explicitly opt in to disabling it.

**`normalize_metric_name/1` was duplicated in the test suite**

The test suite defined its own local copy of `normalize_metric_name/1` rather than testing the production function. This has been addressed in a follow-up commit by extracting the normalization logic into a new `rabbit_prometheus_util` module, which is now called from both the collector and the test suite.

**Performance**

`normalize_disabled_metrics/1` is called on every scrape, re-reading and re-normalizing the disabled metrics list from app env each time. This is minor when the list is short, but worth addressing in a follow-up.

**Documentation**

The website documentation needs to be updated to describe the new configuration options and their defaults. The relevant sections are in `docs/prometheus/index.md` in the rabbitmq-website repository:

- https://github.com/rabbitmq/rabbitmq-website/blob/main/docs/prometheus/index.md#L516-L580

Each endpoint section (`/metrics/memory-breakdown`, `/metrics/per-object`, `/metrics/detailed`) should note the corresponding disable option, and a new section should cover `prometheus.disabled_metrics`.

themavik (commented):
rabbit_prometheus_dispatcher gates register_collectors for per-object/detailed/memory-breakdown when disable_* envs are set; prometheus_rabbitmq_core_metrics_collector:mf/4 filters names via rabbit_prometheus_util:normalize_disabled_metrics/1. nit: default disable_per_object_endpoint=true changes defaults—worth a release note.

lukebakken:
@themavik stop spamming repositories in the `rabbitmq` org with your useless summary comments.

deps/rabbitmq_prometheus/priv/schema/rabbitmq_prometheus.schema
```diff
@@ -184,13 +184,13 @@ end}.
 
 %% Endpoint disable options ========================================================
 {mapping, "prometheus.disable_memory_breakdown_endpoint", "rabbitmq_prometheus.disable_memory_breakdown_endpoint",
-    [{datatype, {enum, [true, false]}}, {default, false}]}.
+    [{datatype, {enum, [true, false]}}]}.
```
michaelklishin, line 187: With Cuttlefish `3.9.x`, you can use the `boolean` type.

PR #15500, opened 2026-02-18
Fix logs dir rm command for rabbitmq_cli

deadtrickster:
As of now it is not possible to just run make -C deps/rabbitmq_cli tests twice in a row. It starts nodes, then complains `logs` is a directory and exits, leaving orphaned erlang nodes.


deadtrickster:
cc @dcorbacho

PR #15515, opened 2026-02-19
Make the list of ignored node DB directories extensible

michaelklishin:
## Proposed Changes

This is the most minimalistic way of making the list of (sub)directories that
are ignored when determining whether a node should start as a blank one (a node with an unintialized data directory).

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

Per discussion with @deadtrickster.


michaelklishin:
This is minimalistic but won't work for additional plugin packs installed on an open source RabbitMQ node.

We need to investigate using `rabbit_registry` instead.

PR #15516, opened 2026-02-19
Make the list of ignored node DB directories extensible (take 2, `rabbit_registry`-based)

michaelklishin:
## Proposed Changes

This is the most "proper" way to allow plugins extend the list of (sub)directories that
are ignored when determining whether a node should start as a blank one (a node with an unintialized data directory.

With this change, a plugin can define a boot step that updates the registry and runs before the database initialization boot step does.

## Further Comments

Note that some tests might call `is_virgin_node/0` directly, bypassing the boot step
subsystem entirely. For those cases, we handle the missing table exception and return
an empty list of additional subdirectories.

Supersedes https://github.com/rabbitmq/rabbitmq-server/pull/15515.

### DIFF
--- a/deps/rabbit/src/rabbit_mnesia.erl
+++ b/deps/rabbit/src/rabbit_mnesia.erl
@@ -1083,9 +1083,13 @@ core_ignored_filenames() ->
 
 -spec additional_ignored_filenames() -> [file:filename_all()].
 additional_ignored_filenames() ->
-    lists:flatmap(
-      fun({_Type, Module}) -> Module:ignored_data_dir_names() end,
-      rabbit_registry:lookup_all(node_data_dir_filter)).
+    try
+        lists:flatmap(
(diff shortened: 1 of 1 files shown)


michaelklishin:
@ansd this is not a new generic mechanism, `rabbit_registry` has been around since 2011 or so, originally introduced for pluggable runtime parameter validation (Shovel, Federation).

The registry is not Mnesia-specific.

Hardcoding a couple of additional strings is also an option that won't take anything away from OSS installations (beyond the ability for some theoretical plugins to reuse the same subdirectories).

ansd:
> The registry is not Mnesia-specific.

I meant that having to list these files in https://github.com/rabbitmq/rabbitmq-server/blob/8cb4fe232368dcba5a64c309cfdadae60fc547b6/deps/rabbit/src/rabbit_mnesia.erl#L1053-L1077 is Mnesia specific.

PR #15524, opened 2026-02-20
Add missing QUEUE_TYPE for rabbit_mqtt_qos0_queue

MarcialRosales:
Add `rabbit_mqtt_qos0_queue` queue type.

Use `classic-queue-get-message` template when `get_message` action is enabled and not template was configured.
This code keeps backward compatibility with how the get-message UI used to work. Originally, the queue page (queue.ejs) contained the get-message HTML inline. It has since been moved into a separate template that is selected via the queue type’s QueueType.tmpl.get_message attribute.
Some plugins outside rabbitmq-server were written for the old behaviour: they enable the get-message action but never set a get-message template, because the UI was always present. The RabbitMQ Delayed Message plugin is one example: it enables the get-message action but does not declare get_message on its queue type. With this fallback, such queue types still get a template (the classic-queue get-message template) and the UI continues to work without requiring those plugins to be updated.

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/global.js
+++ b/deps/rabbitmq_management/priv/www/js/global.js
@@ -998,6 +998,21 @@ QUEUE_TYPE["stream"] = {
     }
 };
 
+QUEUE_TYPE["rabbit_mqtt_qos0_queue"] = {
+    label: "MQTT QoS0",
+    params: {},
+    actions: {
+        get_message: false,
+        purge: false
+    },
+    tmpl: {
+        "list"   : "classic-queue-list",
+        "stats"  : "classic-queue-stats",
+        "node_details" : "classic-queue-node-details"
+    }
+};
+
+
 // here I'll shortcut for now and let it be like that
 // other queue types can inject themlves where they want.
 // since the 'sections' object will likely keep key insertion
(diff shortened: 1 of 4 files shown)


deps/rabbitmq_management/priv/www/js/global.js
```diff
@@ -998,6 +998,26 @@ QUEUE_TYPE["stream"] = {
...
 };
 
+QUEUE_TYPE["rabbit_mqtt_qos0_queue"] = {
+    label: "MQTT QoS0",
+    params: {},
+    policy_apply_to: "rabbit_mqtt_qos0_queue",
+    actions: {
+        get_message: true,
+        purge: true
+    },
```
ansd: Note that the MQTT QoS 0 queue type does not support policies. Neither can messages be received from this queue type via the Management UI nor can this queue type be purged.

deadtrickster:
I think you can just remove that "if" for DQ. ping me once this pr is merged and I will update DQ's stuff (copy paste that classic queue template).

deps/rabbitmq_management/priv/www/js/global.js
MarcialRosales: Thanks @ansd ! I will update its queue_type accordingly.

MarcialRosales:
@deadtrickster I am going to leave the default mechanism that defaults to "classic-queue-get-message" if the queue type has the `get_message` action enabled. It does not cause any harm, i think. The outer if-statement that checks whether the action `get_message` is enabled is still required because not all queue types support get_message.

PR #15529, opened 2026-02-21
Fix #12681: Classic config peer discovery still might select two differe

### LINKED ISSUE #12681: Peer discovery backend might fail to join the whole cluster
### Describe the bug

Sometimes in CI the testcase `peer_discovery_classic_config_SUITE:successfully_discovery` fails to cluster all 7 nodes.
Six nodes join the same seed, and one other node selects itself as a seed and does not join the rest of the cluster.

The peer discovery seed selection needs to be reviewed

### Reproduction steps

1. It's a CI flake

### Expected behavior

The 7 nodes join the cluster

### Additional context

_No response_

danielalanbates:
Fixes #12681

## Summary
This PR fixes: Classic config peer discovery still might select two different seed nodes

## Changes
```
deps/rabbit/src/rabbit_peer_discovery.erl | 11 ++++++++++-
 1 file changed, 10 insertions(+), 1 deletion(-)
```

## Testing
Please review the changes carefully. The fix was verified against the existing test suite.

---
*This PR was created with the assistance of Claude Sonnet 4.6 by Anthropic | effort: low. Happy to make any adjustments!*

### DIFF
--- a/deps/rabbit/src/rabbit_peer_discovery.erl
+++ b/deps/rabbit/src/rabbit_peer_discovery.erl
@@ -601,9 +601,18 @@ do_query_node_props(Nodes, FromNode) when Nodes =/= [] ->
 
     %% TODO: Replace with `rabbit_nodes:list_members/0' when the oldest
     %% supported version has it.
+    %%
+    %% We use ?ERPC_CALL_TIMEOUT here so that nodes which are still
+    %% initialising their Erlang distribution get a chance to respond before
+    %% being excluded from the sorted list. Without a retry window, a node
+    %% that starts slightly later than its peers can end up excluded from all
+    %% other nodes' queries, while it can still reach those peers itself.
+    %% This asymmetry causes two different seeds to be selected.
     MembersPerNode = [try
                           {ok,
-                           erpc_call(Node, rabbit_nodes, all, [], FromNode)}
+                           erpc_call(
+                             Node, rabbit_nodes, all, [], FromNode,
+                             ?ERPC_CALL_TIMEOUT)}
                       catch
                           Class:Reason ->
                               {Class, Reason}


michaelklishin:
Oh, that's a good find.

This looks good to me but we need to go through the [CLA process](https://github.com/rabbitmq/cla) before I can accept them. I'll continue reviewing, though.

danielalanbates:
Thank you for the review! Glad it helps. 🙏

michaelklishin:
So apparently with `Timeout` set to `0`, if a peer's Erlang distribution isn't registered with `epmd` yet, the `erpc:call/6` fails with an `{erpc, noconnection}` and that peer is immediately skipped by the peer discovery.

Relevant `erpc` source links:

 * [`erpc.erl#L280-L286`](https://github.com/erlang/otp/blob/OTP-27.3.4.8/lib/kernel/src/erpc.erl#L280-L286)
 * [`erpc.erl#L1371-L1372`](https://github.com/erlang/otp/blob/OTP-27.3.4.8/lib/kernel/src/erpc.erl#L1371-L1372)
 
The description in the PR is not super precise but fair enough, I believe I understand the scenario.

danielalanbates:
Thank you for the review! Glad it helps. 🙏

lukebakken:
@michaelklishin see the discussion I stared in `#core-team` with regard to this, thanks.

michaelklishin:
I'll wait for @lukebakken to review this PR (it's a part of a research study, not sure if it'd be OK for me to reveal any details).

lukebakken:
Unfortunately I can't use the required software to record my review 😢

michaelklishin:
I've dug in again to see whether this timeout would have any side effects.

And indeed, for a 5, 7 or 9 cluster, this can add tens of seconds of delays in pathological cases, a scenario that we have seen become a problem with certain deployment tools, specifically for the first node.

So while the analysis is correct, the likelihood of this scenario has to be weighed against a more likely scenario of the first booting node being blocked for long enough to cause a different kind of a problem during formation of a brand new cluster.

So I conclude that this PR is a net negative, even though the subtle scenario described is indeed possible, even if very unlikely.

Even with a much shorter 3s timeout, the result can end up being a low double digit second delay per node.

I'm open to revisiting this one day if the unlikely "two seeds" scenario starts materializing often enough (we haven't see it in a while after the latest round of peer discovery refactoring), so thank you for taking some time to contribute.

michaelklishin:
A relevant change https://github.com/rabbitmq/rabbitmq-server/commit/c402032a78f56534190ba4ed5910963b727d10fc

PR #15542, opened 2026-02-23
Remove Mnesia support (including partition handling strategies)

dumbbell:
## Why

Khepri is now the default and only metadata store supported by RabbitMQ.

## How

Almost all Mnesia-related code is deleted. There is some code left: it is only used to implement the migration from Mnesia to Khepri.

This clean-up includes the network partition handling strategies because they were specific to our use of Mnesia.


deps/rabbit/test/rabbit_db_queue_SUITE.erl
michaelklishin, line 1: Lines  345 and 347 in this file need updating for the changes to `rabbit_db_queue:delete/2`, whose second argument is now `OnlyDurable :: boolean()`, so these tests do not fail but this is an issue of semantics:

```erl
Deletions = rabbit_db_queue:delete(QName, normal),
```

```erl
?assertEqual(ok, rabbit_db_queue:delete(QName, normal)),
```

deps/rabbitmq_ct_helpers/src/clustering_utils.erl
michaelklishin, line 1: [Two lines](https://github.com/rabbitmq/rabbitmq-server/blob/70c76e4bd6034f111f202c24b02177a7e2f0e904/deps/rabbitmq_ct_helpers/src/clustering_utils.erl#L96-L97) still reference `rabbit_mnesia` functions.

michaelklishin (changes requested):
A comment on a file that were not touched and therefore I cannot comment on them or their lines directly:

`per_user_connection_channel_limit_partitions_SUITE.erl:187` still references `rabbit_mnesia`

deps/rabbit/src/rabbit_health_check.erl
michaelklishin, line 1: `rabbit_health_check.erl:60-61` can be simplified because the condition `rabbit_khepri:is_enabled(RemoteNode)` will now always evaluate to `true`.

deps/rabbit/src/rabbit_mnesia.erl
michaelklishin, line 1: On lines 31-33, the `cluster_status/0` exports seem unused.

deps/rabbit/src/rabbit_db.erl
michaelklishin, line 1: [This comment on line 199](https://github.com/rabbitmq/rabbitmq-server/blob/70c76e4bd6034f111f202c24b02177a7e2f0e904/deps/rabbit/src/rabbit_db.erl#L199) is a copy-paste artifact.

deps/rabbit/src/rabbit_db_maintenance.erl
michaelklishin, line 1: `table_definitions/0` was exclusive to Mnesia.

deps/rabbit/src/rabbit_mnesia.erl
michaelklishin: So is `is_running/0` on line 45.

deps/rabbit/src/rabbit_db_queue.erl
michaelklishin, line 1: 1. Lines 43-44: `rabbit_db_queue.erl` has a TODO for "when Mnesia is removed"
2. Lines 55-56: suggests that a few functions used by `rabbit_amqqueue` can be removed "once Mnesia and transient entities are removed". We should update the comment
3. Lines 61-63: same thing as 2 but with another function, `set_dirty/1` (called in `rabbit_quorum_queue.erl:828`

deps/rabbit/src/rabbit_db_exchange.erl
michaelklishin, line 1: An [unaddressed TODO](https://github.com/rabbitmq/rabbitmq-server/blob/70c76e4bd6034f111f202c24b02177a7e2f0e904/deps/rabbit/src/rabbit_db_exchange.erl#L518) that suggest moving a callback to `rabbit_exchange` once Mnesia is removed.

deps/rabbit/src/rabbit_amqqueue_process.erl
michaelklishin, line 1: Two TODOs for "when Mnesia can be removed" on lines 332-333, then lines 407-408.

dumbbell:
Thank you @michaelklishin for the review and fixes!

I squashed all of them into the main commit, except the following chunks:
* In `rabbit_db_cluster:join()`, we still need to enable Khepri on a remote cluster before joining, in case it’s running an older version of RabbitMQ.
* The Mnesia retry timeout and limit parameters are still used during the migration from Mnesia to Khepri.

PR #15548, opened 2026-02-24
QQ DLX: use a pipeline command for checkout

mkuratczyk:
Before this change, with two leader elections in short succession, we could end up in a situation where no DLX worker is running.

single_dlx_worker test appeared to be flaky, but it actually triggered a real race condition:

A node that briefly was a leader, started a DLX worker which registered itself as a consumer via ra:process_command.

When that node loses leadership a moment later, process_command on the now-follower Ra server redirects the checkout command to the actual leader, where it gets committed to the Raft log. This overwrites the new/real leader's worker registration, terminating the leader's worker.

The result is that the leader has no DLX worker: the old leader's worker is stopped to allow the new leader to take over, but the new leader's worker exits due to a stale checkout.

This PR should address this failure:
https://github.com/rabbitmq/rabbitmq-server/actions/runs/22343794026/job/64653112641?pr=15502

### DIFF
--- a/deps/rabbit/src/rabbit_fifo_dlx_client.erl
+++ b/deps/rabbit/src/rabbit_fifo_dlx_client.erl
@@ -15,6 +15,7 @@
 -record(state,{
           queue_resource :: rabbit_types:r(queue),
           leader :: ra:server_id(),
+          checkout_corr :: undefined | reference(),
           last_msg_id :: non_neg_integer() | -1
          }).
 -type state() :: #state{}.
@@ -35,39 +36,32 @@ settle(MsgIds, #state{leader = Leader} = State)
     {ok, State}.
 
 -spec checkout(rabbit_amqqueue:name(), ra:server_id(), non_neg_integer()) ->
-    {ok, state()} | {error, non_local_leader | ra_command_failed}.
+    state().
 checkout(QResource, Leader, NumUnsettled) ->
     Cmd = rabbit_fifo_dlx:make_checkout(self(), NumUnsettled),
-    State = #state{queue_resource = QResource,
-                   leader = Leader,
-                   last_msg_id = -1},
-    process_command(Cmd, State, 5).
+    Corr = make_ref(),
+    ra:pipeline_command(Leader, Cmd, Corr, normal),
+    #state{queue_resource = QResource,
+           leader = Leader,
+           checkout_corr = Corr,
+           last_msg_id = -1}.
 
-process_command(_Cmd, _State, 0) ->
-    {error, ra_command_failed};
-process_command(Cmd, #state{leader = Leader} = State, Tries) ->
-    case ra:process_command(Leader, Cmd, 60_000) of
-        {ok, ok, Leader} ->
-            {ok, State#state{leader = Leader}};
-        {ok, ok, NonLocalLeader} ->
-            ?LOG_WARNING("Failed to process command ~tp on quorum queue leader ~tp because actual leader is ~tp.",
-                               [Cmd, Leader, NonLocalLeader]),
-            {error, non_local_leader};
-        Err ->
-            ?LOG_WARNING("Failed to process command ~tp on quorum queue leader ~tp: ~tp~n"
-                               "Trying ~b more time(s)...",
-                               [Cmd, Leader, Err, Tries]),
-            process_command(Cmd, State, Tries - 1)
-    end.
-
--spec handle_ra_event(pid(), term(), state()) ->
-    {ok, state(), actions()}.
+-spec handle_ra_event(term(), term(), state()) ->
+    {ok, state(), actions()} | {reject, state()} | ignore.
 handle_ra_event(Leader, {dlx_delivery, _} = Del,
                 #state{leader = _Leader} = State) when node(Leader) == node() ->
     handle_delivery(Del, State);
-handle_ra_event(From, Evt, State) ->
+handle_ra_event(_From, {applied, [{Corr, ok}]},
+                #state{checkout_corr = Corr} = State)
+  when is_reference(Corr) ->
+    {ok, State#state{checkout_corr = undefined}, []};
+handle_ra_event(_From, {rejected, {not_leader, _Leader, Corr}},
+                #state{checkout_corr = Corr} = State)
+  when is_reference(Corr) ->
+    {reject, State};
+handle_ra_event(From, Evt, _State) ->
     ?LOG_DEBUG("Ignoring ra event ~tp from ~tp", [Evt, From]),
-    {ok, State, []}.
+    ignore.
 
 handle_delivery({dlx_delivery, [{FstId, _} | _] = IdMsgs},
                 #state{queue_resource = QRes,
--- a/deps/rabbit/src/rabbit_fifo_dlx_client.erl
+++ b/deps/rabbit/src/rabbit_fifo_dlx_client.erl
@@ -8,14 +8,12 @@
 
 -include_lib("kernel/include/logger.hrl").
 
-
--export([checkout/3, settle/2, handle_ra_event/3,
+-export([checkout/3, settle/2, handle_registration/1, handle_ra_event/3,
          overview/1]).
 
 -record(state,{
           queue_resource :: rabbit_types:r(queue),
           leader :: ra:server_id(),
-          checkout_corr :: undefined | reference(),
           last_msg_id :: non_neg_integer() | -1
          }).
 -type state() :: #state{}.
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_fifo_dlx_worker.erl
```diff
@@ -569,6 +580,12 @@ cancel_timer(#state{timer = TRef} = State)
 cancel_timer(State) ->
     State.
 
+queue_event_to_dlx_client({ServerId, Evt}, DlxState)
+  when is_tuple(ServerId), tuple_size(ServerId) =:= 2 ->
```
kjnilsson: could be replaced with a `{_, _}` in the function clause.

deps/rabbit/src/rabbit_fifo_dlx_worker.erl
```diff
@@ -569,6 +580,12 @@ cancel_timer(#state{timer = TRef} = State)
 cancel_timer(State) ->
     State.
 
+queue_event_to_dlx_client({ServerId, Evt}, DlxState)
```
kjnilsson: rename `maybe_handle_dlx_event`?

deps/rabbit/src/rabbit_fifo_dlx_client.erl
```diff
@@ -35,39 +36,32 @@ settle(MsgIds, #state{leader = Leader} = State)
...
-    {ok, state()} | {error, non_local_leader | ra_command_failed}.
+    state().
 checkout(QResource, Leader, NumUnsettled) ->
     Cmd = rabbit_fifo_dlx:make_checkout(self(), NumUnsettled),
-    State = #state{queue_resource = QResource,
-                   leader = Leader,
-                   last_msg_id = -1},
-    process_command(Cmd, State, 5).
+    Corr = make_ref(),
+    ra:pipeline_command(Leader, Cmd, Corr),
```
kjnilsson: this should be `ra:pipeline_command(Leader, Cmd, Corr, normal).` to ensure the priority is not sent as `low`

kjnilsson (changes requested):

ansd:
Relevant logs of the flake:
```
2026-02-24 09:06:19.413770+00:00 [debug] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/': vote granted for term 3 votes 2
2026-02-24 09:06:19.414048+00:00 [debug] <0.2377.0> started rabbit_fifo_dlx_worker <0.2600.0> for queue 'single_dlx_worker_source' in vhost '/'
2026-02-24 09:06:19.414096+00:00 [notice] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/': candidate -> leader in term: 3 machine version: 7, last applied 5
2026-02-24 09:06:19.414388+00:00 [debug] <0.2602.0> queue 'single_dlx_worker_source' in vhost '/': updating leader record to current node rmq-ct-cluster_size_3-1-28000@localhost
2026-02-24 09:06:19.414279+00:00 [info] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/': leader saw request_vote_rpc from {'%2F_single_dlx_worker_source','rmq-ct-cluster_size_3-3-28144@localhost'} for term 4 abdicates term: 3!
2026-02-24 09:06:19.417479+00:00 [notice] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/': leader -> follower in term: 4 machine version: 7, last applied 5
2026-02-24 09:06:19.417533+00:00 [debug] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/': is not new, setting election timeout.
2026-02-24 09:06:19.417740+00:00 [info] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/': declining vote for {'%2F_single_dlx_worker_source','rmq-ct-cluster_size_3-3-28144@localhost'} for term 4, candidate last log {index, term} was: {5,2}  last log entry {index, term} is: {{6,3}}
2026-02-24 09:06:19.417824+00:00 [debug] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/': leader call - leader not known. Command will be forwarded once leader is known.
2026-02-24 09:06:19.418190+00:00 [debug] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/' declining pre-vote to {'%2F_single_dlx_worker_source','rmq-ct-cluster_size_3-2-28072@localhost'} for term 3, current term 4
2026-02-24 09:06:19.428043+00:00 [debug] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/': resetting last index to 5 from 6 in term 4
2026-02-24 09:06:19.428157+00:00 [info] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/': detected a new leader {'%2F_single_dlx_worker_source','rmq-ct-cluster_size_3-3-28144@localhost'} in term 4
2026-02-24 09:06:19.428280+00:00 [debug] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/': mem table overwriting detected whilst staging entries, opening new mem table
2026-02-24 09:06:19.436299+00:00 [debug] <0.2377.0> queue 'single_dlx_worker_source' in vhost '/': enabling ra cluster changes in 4, index 6
2026-02-24 09:06:19.436411+00:00 [debug] <0.2377.0> Terminating <31028.1894.0> since <31122.2516.0> becomes active rabbit_fifo_dlx_worker
2026-02-24 09:06:19.437003+00:00 [debug] <0.2377.0> Terminating <31122.2516.0> since <0.2600.0> becomes active rabbit_fifo_dlx_worker
2026-02-24 09:06:19.437107+00:00 [warning] <0.2600.0> Failed to process command {dlx,{checkout,<0.2600.0>,2}} on quorum queue leader {'%2F_single_dlx_worker_source',
2026-02-24 09:06:19.437107+00:00 [warning] <0.2600.0>                                                                                 'rmq-ct-cluster_size_3-1-28000@localhost'} because actual leader is {'%2F_single_dlx_worker_source',
2026-02-24 09:06:19.437107+00:00 [warning] <0.2600.0>                                                                                                                                                      'rmq-ct-cluster_size_3-3-28144@localhost'}.

```

deps/rabbit/src/rabbit_fifo_dlx_worker.erl
```diff
@@ -146,26 +146,37 @@ handle_cast({dlx_event, _LeaderPid, lookup_topology},
...
+                    State1 = State0#state{queue_type_state = QTypeState1},
+                    State = handle_queue_actions(Actions, State1),
+                    {noreply, State};
+                {eol, Actions} ->
+                    State = handle_queue_actions(Actions, State0),
+                    remove_queue(QRef, State);
+                {protocol_error, _Type, _Reason, _Args} ->
+                    {noreply, State0}
+            end
+    end;
```
ansd, line 183: Can we simplify this?
Conceptually after the worker sent its initial `{dlx, #checkout{}}` message with a correlation, it's in a separate state `registering`. The only message it should receive in this state is a reply with that given correlation. Maybe we can make this more apparent by introducing a `rabbit_fifo_dlx_client:handle_registration` function? If successful, the state changes to `registered`. Only thereafter will the worker handle other messages including dlx deliveries from its source queue and publisher confirms from target queues.

mkuratczyk: I've refactored the code. Please take a look if that's what you had in mind

ansd: Thank you. The code is much simpler now.

> The only message it should receive in this state is a reply with that given correlation.

@kjnilsson gave me a hint that I was wrong on this one. Apparently, the `applied` notifications in Ra (including that the dlx checkout command was applied) are delivered **after** Ra effects (which may already include dlx deliveries). So, we have 3 options I suppose:
1. Stash any early (pre-registartion) dlx deliveries that the dlx worker proc receives in its state, or
2. Send the dlx worker checkout command for pre evaluation to the aux machine (similar as done in https://github.com/rabbitmq/rabbitmq-server/blob/5388ef9e3c10043ccb5caf03ef751db2020e5520/deps/rabbit/src/rabbit_fifo_client.erl#L981-L982) The leader will then only append this Ra command to its log if the command was sent from a local dlx worker, or
3. Modify Ra to have an option so that ra:process_command() won't redirect to the leader.

What do you think? Which approach is the simplest and/or safest?

themavik (commented):
checkout/3 now uses ra:pipeline_command with a correlation ref and handle_registration/1 instead of the synchronous process_command retry loop; rabbit_fifo_dlx_worker keeps checkout_corr until RA confirms checkout. Returning {shutdown, queue_not_found} when rabbit_amqqueue:lookup misses beats a badmatch.

ansd:
Closing in favor of https://github.com/rabbitmq/rabbitmq-server/pull/16203

PR #15566, opened 2026-02-26
python_SUITE: Wait for bindings to be active before publishing messages

dumbbell:
## Why

Sometimes, the message was published too soon. This led to a consumer waiting forever and to the testsuite timing out.

## How

We use `rabbitmqctl list_bindings` in a loop to determine when the bindings are active.


dcorbacho:
@Mergifyio backport v4.1.x

dcorbacho:
@Mergifyio backport v4.0.x

PR #15595, opened 2026-02-28
CQs: fix a `read_rom_q_tail/4` crash when `tune_read/2` rounds past `end_seq_id`

michaelklishin:
`tune_read/2` may round up the read range to a segment boundary, advancing `start_seq_id` past `end_seq_id`. Guard against this in both branches of `read_from_q_tail;4`.

This scenario was detected by a sporadic failure in `classic_queue_prop_SUITE` which now passes 200 times in a row.

### DIFF
--- a/deps/rabbit/test/classic_queue_prop_SUITE.erl
+++ b/deps/rabbit/test/classic_queue_prop_SUITE.erl
@@ -1132,8 +1132,8 @@ do_reg_recovery_tune_read_crash(Config) ->
     St2 = next_state(St1, Res2, {call, undefined, cmd_channel_open, [St1]}),
 
     Res3 = cmd_channel_publish_many(St2, Res2, 511, 260, 1, true, undefined),
-    true = postcondition(St2, {call, undefined, cmd_channel_publish_many, [St2, Res3, 511, 260, 1, true, undefined]}, Res3),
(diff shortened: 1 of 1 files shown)


lhoguin:
Checking.

lhoguin:
Could you provide me with the crash log for this? Both to confirm the fix and to write a regression test case.

michaelklishin:
@lhoguin I will dig in, they were logs from last Friday.

michaelklishin:
Below a standard CT `.zip` file straight from Actions. It should include node data directories.

The failure can be seen in `ct_run.rabbit_shard4@localhost.2026-02-28_08.34.49/deps.rabbit.classic_queue_prop_SUITE.logs/run.2026-02-28_08.36.08/classic_queue_prop_suite.classic_queue_v2.html`.

[CT logs (rabbit parallel-ct-set-1 OTP-28 ).zip](https://github.com/user-attachments/files/25688978/CT.logs.rabbit.parallel-ct-set-1.OTP-28.zip)

lhoguin:
Thank you. I can reproduce.

lhoguin:
I have amended and pushed the change.

I added a regression test based on the CT logs and tweaked the code, notably removing the check in the second clause because it is already covered by `case QTailCount - QHead1Len` where we know nothing remains in q_tail and we don't use `QTailSeqId1` in that case.

PR #15603, opened 2026-03-02
rabbitmq_shovel: detach link when closing amqp10 connections

dcorbacho:
When using an autodelete shovel, acks to source might still be in-flight when the connections start to close. Let's give them a short buffer to finish sending acks and then close.

`shovel_dynamic_SUITE:autodelete_classic_on_publish_with_rejections/1` often flakes in CI because of this

### DIFF
--- a/deps/rabbitmq_shovel/src/rabbit_amqp10_shovel.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_amqp10_shovel.erl
@@ -421,7 +421,7 @@ close_source(#{source := #{current := #{conn := Conn,
 close_source(_Config) -> ok.
 
 connection_close(Conn, Sess, LinkRef) ->
-    _ = amqp10_client:detach_link(LinkRef),
+    catch amqp10_client:detach_link(LinkRef),
     _ = amqp10_client:end_session(Sess),
     _ = amqp10_client:close_connection(Conn).
 


deps/rabbitmq_shovel/src/rabbit_amqp10_shovel.erl
```diff
@@ -391,15 +392,18 @@ handle_dest(_Msg, _State) ->
...
-    _ = amqp10_client:close_connection(Conn),
+    %% Wait for pending acks/nacks to be delivered
+    _ = timer:apply_after(10_000, ?MODULE, connection_close, [Conn, Sess]),
     ok;
 close_source(_Config) -> ok.
 
+connection_close(Conn, Sess) ->
+    _ = amqp10_client:end_session(Sess),
+    _ = amqp10_client:close_connection(Conn).
+
```
ansd, line 427: The link should be detached first.

ansd: Detaching the link will cause acks to be delivered to the quorum queue before the quorum queue removes the consumer:
https://github.com/rabbitmq/rabbitmq-server/blob/a07a30f55af995e7fd395604a68a6ea0d223175c/deps/rabbit/src/rabbit_fifo_client.erl#L482

michaelklishin:

PR #15619, opened 2026-03-03
Replace Khepri topic routing projection with trie + ordered_set (v4)

### LINKED ISSUE #15588: Performance for subscriptions with the same topic filter
Start RabbitMQ on `main` branch:
```
make run-broker TEST_TMPDIR="$HOME/scratch/rabbit/test"  RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+S 6 +JPperf true" RABBITMQ_CONFIG_FILE="$HOME/scratch/rabbit/mqtt-rabbitmq.conf" ENABLED_PLUGINS="rabbitmq_management rabbitmq_mqtt rabbitmq_prometheus" FULL=1
```
with `mqtt-rabbitmq.conf`:
```
mqtt.tcp_listen_options.sndbuf = 2048
mqtt.tcp_listen_options.recbuf = 2048
mqtt.tcp_listen_options.buffer = 2048
mqtt.tcp_listen_options.backlog = 512
mqtt.tcp_listen_options.nodelay = true
vm_memory_high_watermark.relative = 0.95
vm_memory_high_watermark_paging_ratio = 0.95
disk_free_limit.absolute = 200MB
loopback_users = none
log.file.level = warning
```

Connect 20k MQTT clients in parallel. Each client has only 1 subscription, but with the **same** topic filter:
```
./omq --uri mqtt://localhost:1883 mqtt -x 0 -y 20000 --parallel-consumer-start --consume-from "mytopic"
```

Problem:
Clients time out when creating the subscriptions.

RabbitMQ logs lots of:
```
[error] <0.47157.0> Failed to add binding between exchange 'amq.topic' in vhost '/' and queue 'mqtt-subscription-omq-consumer-12543qos0' in vhost '/' for topic filter mytopic: timeout
```

The CPU flame graph shows how the Khepri Ra server proc is busy copying to/from ETS:

Image

If you run only with 3 clients connecting to the same topic filter as follows
```
./omq --uri mqtt://localhost:1883 mqtt -x 0 -y 3 --parallel-consumer-start --consume-from "mytopic"
```
then you see all 3 different bindings in the same ETS row:
```
1> rp(ets:tab2list(rabbit_khepri_topic_trie_v3)).
[{topic_trie_edge_v2,{trie_edge,{resource,<<"/">>,exchange,
                                          <<"amq.topic">>},
                                #Ref<0.770359334.1341915139.10425>,bindings},
                     {bindings,#{{binding,{resource,<<"/">>,exchange,
                                                    <<"amq.topic">>},
                                          <<"mytopic">>,
                                          {resource,<<"/">>,queue,
                                                    <<"mqtt-subscription-omq-consumer-1qos0">>},
                                          [{<<"x-binding-key">>,longstr,<<"mytopic">>},
                                           {<<"x-mqtt-subscription-opts">>,table,
                                            [{<<"no-local">>,bool,false},
                                             {<<"qos">>,unsignedbyte,0},
                                             {<<"retain-as-published">>,bool,false},
                                             {<<"retain-handling">>,unsignedbyte,0}]}]} =>
                                     [],
                                 {binding,{resource,<<"/">>,exchange,<<"amq.topic">>},
                                          <<"mytopic">>,
                                          {resource,<<"/">>,queue,
                                                    <<"mqtt-subscription-omq-consumer-2qos0">>},
                                          [{<<"x-binding-key">>,longstr,<<"mytopic">>},
                                           {<<"x-mqtt-subscription-opts">>,table,
                                            [{<<"no-local">>,bool,false},
                                             {<<"qos">>,unsignedbyte,0},
                                             {<<"retain-as-published">>,bool,false},
                                             {<<"retain-handling">>,unsignedbyte,0}]}]} =>
                                     [],
                                 {binding,{resource,<<"/">>,exchange,<<"amq.topic">>},
                                          <<"mytopic">>,
                                          {resource,<<"/">>,queue,
                                                    <<"mqtt-subscription-omq-consumer-3qos0">>},
                                          [{<<"x-binding-key">>,longstr,<<"mytopic">>},
                                           {<<"x-mqtt-subscription-opts">>,table,
                                            [{<<"no-local">>,bool,false},
                                             {<<"qos">>,unsignedbyte,0},
                                             {<<"retain-as-published">>,bool,false},
                                             {<<"retain-handling">>,unsignedbyte,0}]}]} =>
                                     []}},
                     undefined},
 {topic_trie_edge_v2,{trie_edge,{resource,<<"/">>,exchange,
                                          <<"amq.topic">>},
                                root,<<"mytopic">>},
                     #Ref<0.770359334.1341915139.10425>,1}]
ok
```

I don't suggest this is the correct solution, but in contrast, with Mnesia on `v4.2.x` these are all in separate ETS rows:
```
(rabbit@nuc)2> ets:tab2list(rabbit_topic_trie_edge).
[{topic_trie_edge,{trie_edge,{resource,<<"/">>,exchange,
                                       <<"amq.topic">>},
                             root,"mytopic"},
                  <<111,200,14,198,26,220,215,250,84,12,77,239,121,191,95,
                    159>>}]

(rabbit@nuc)3> ets:tab2list(rabbit_topic_trie_node).
[{topic_trie_node,{trie_node,{resource,<<"/">>,exchange,
                                       <<"amq.topic">>},
                             root},
                  1,0},
 {topic_trie_node,{trie_node,{resource,<<"/">>,exchange,
                                       <<"amq.topic">>},
                             <<111,200,14,198,26,220,215,250,84,12,77,239,121,191,95,
                               159>>},
                  0,3}]

(rabbit@nuc)4> ets:tab2list(rabbit_topic_trie_binding).
[{topic_trie_binding,{trie_binding,{resource,<<"/">>,
                                             exchange,<<"amq.topic">>},
                                   <<111,200,14,198,26,220,215,250,84,12,77,239,121,191,95,
                                     159>>,
                                   {resource,<<"/">>,queue,
                                             <<"mqtt-subscription-omq-consumer-1qos0">>},
                                   [{<<"x-binding-key">>,longstr,<<"mytopic">>},
                                    {<<"x-mqtt-subscription-opts">>,table,
                                     [{<<"no-local">>,bool,false},
                                      {<<"qos">>,unsignedbyte,0},
                                      {<<"retain-as-published">>,bool,false},
                                      {<<"retain-handling">>,unsignedbyte,0}]}]},
                     const},
 {topic_trie_binding,{trie_binding,{resource,<<"/">>,
                                             exchange,<<"amq.topic">>},
                                   <<111,200,14,198,26,220,215,250,84,12,77,239,121,191,95,
                                     159>>,
                                   {resource,<<"/">>,queue,
                                             <<"mqtt-subscription-omq-consumer-2qos0">>},
                                   [{<<"x-binding-key">>,longstr,<<"mytopic">>},
                                    {<<"x-mqtt-subscription-opts">>,table,
                                     [{<<"no-local">>,bool,false},
                                      {<<"qos">>,unsignedbyte,0},
                                      {<<"retain-as-published">>,bool,false},
                                      {<<"retain-handling">>,unsignedbyte,0}]}]},
                     const},
 {topic_trie_binding,{trie_binding,{resource,<<"/">>,
                                             exchange,<<"amq.topic">>},
                                   <<111,200,14,198,26,220,215,250,84,12,77,239,121,191,95,
                                     159>>,
                                   {resource,<<"/">>,queue,
                                             <<"mqtt-subscription-omq-consumer-3qos0">>},
                                   [{<<"x-binding-key">>,longstr,<<"mytopic">>},
                                    {<<"x-mqtt-subscription-opts">>,table,
                                     [{<<"no-local">>,bool,false},
                                      {<<"qos">>,unsignedbyte,0},
                                      {<<"retain-as-published">>,bool,false},
                                      {<<"retain-handling">>,unsignedbyte,0}]}]},
                     const}]
```

This means clients subscribing to the same topic in parallel will overload Khepri.

20k clients subscribing all in parallel as shown above is a lot. But even when 1k clients subscribe per second **in batches**, the problem is that the ETS table will generate more and more copy overhead per subscription being added leading to timeouts eventually.

The use cases are cloud to devices fanouts as described in https://www.rabbitmq.com/blog/2023/03/21/native-mqtt#benefit-1-large-fan-outs where many clients subscribe to the same topic.

(This is not specific to MQTT, but also affects AMQP clients subscribing to the same topic.)

ansd:
Resolves #15588.

The previous Khepri topic routing projection (v3) stored topic bindings as sets:set(#binding{}) inside trie leaf nodes. This design had a major performance drawback:

On the insertion/deletion path (in the single Khepri Ra process), every binding change required a read-modify-write of the entire sets:set(), making it O(N) in the number of bindings at that leaf. With many MQTT clients connecting concurrently (each subscribing to the same topic filter), this made the Ra process a bottleneck.

Another less severe performance issue was that the entire binding was being copied including the binding arguments containing the MQTT 5.0 subscription options such as:
```
{<<"x-mqtt-subscription-opts">>,table,
 [{<<"id">>,unsignedint,1},
  {<<"no-local">>,bool,false},
  {<<"qos">>,unsignedbyte,0},
  {<<"retain-as-published">>,bool,false},
  {<<"retain-handling">>,unsignedbyte,0}]}]
```

Replace the single ETS projection table with two purpose-built tables:

1. Trie edges table (ETS set, read_concurrency=true):
   - Row: `{{XSrc, ParentNodeId, Word}, ChildNodeId, ChildCount}`
   - XSrc = {VHost, ExchangeName} (compact 2-tuple of binaries)
   - NodeId = root | reference()
   - ChildCount tracks outgoing edges for garbage collection

2. Leaf bindings table (ETS ordered_set, read_concurrency=true):
   - Key: {NodeId, BindingKey, Dest}
   - Stored as 1-tuples: {{NodeId, BindingKey, Dest}}
   - No value column; all data is in the key to minimize copying

The trie structure preserves O(depth * 3) routing complexity regardless of the number of overall bindings or wildcard filters. At each trie level, we probe at most 3 edges (literal word, <<"*">>, <<"#">>), each via ets:lookup_element/4 which copies only the ChildNodeId (a reference).

The ordered_set for bindings provides:
- O(log N) insert and delete per binding (no read-modify-write)
- The binding key (needed for MQTT subscription identifiers and topic aliases) is part of the key, so it is returned directly during destination collection without additional lookups

Collecting destinations at a matched trie leaf uses a hybrid strategy:
- Fanout 0-2 (the common case: unicast, device + stream): up to 3 ets:next/2 probes. Each ets:next/2 call costs O(log N) because the CATree (used with read_concurrency) allocates a fresh tree traversal stack on each call.
- Fanout > 2: ets:select/2 with a partially bound key does an O(log N) seek followed by an O(F) range scan. The match spec compilation overhead amortises over the larger result set.

ets:lookup_element/4 (OTP 26+) returns a default value on miss instead of throwing badarg, and copies only the requested element on hit. This avoids both exception overhead (misses are common during trie traversal of <<"*">> and <<"#">> branches) and unnecessary data copying (we only need the ChildNodeId, not the full row).

Trie node IDs are ephemeral (the tables are rebuilt when the Khepri projection is re-registered). make_ref() is fast, globally unique within a node, and has good hash distribution for the ETS set table.

When a binding is deleted, the trie path from root to leaf is collected in a single downward walk (trie_follow_down_get_path). Empty nodes are then pruned bottom-up: a node is empty when its ChildCount is 0 and it has no bindings in the ordered_set table.

Benchmarks below were run with 500K routing operations per scenario (on the same machine, back-to-back between main (v3) and this commit.

Significant insert/delete improvements:

  Churn insert (8K bindings, 4 filters/client): ~1,120 vs ~810 ops/s (+38%)
    v3 did a read-modify-write of sets:set() per binding; v4 does
    a single ets:insert into the ordered_set plus trie edge updates.

  MQTT device insert (20K bindings): ~650 vs ~420 ops/s (+55%)
    Same mechanism as churn insert. Particularly impactful when many
    clients share the same wildcard filter (e.g. "broadcast.#"),
    since v3's sets:set() grew with each client while v4 inserts
    are O(log N) regardless.

  Same-key fanout insert (10K): ~415 vs ~290 ops/s (+43%)
    The worst case for v3: all 10K bindings share the same key,
    so each insert copies and rebuilds the growing sets:set().

Routing improvements:

  MQTT unicast (10K devices, 20K bindings): ~460K vs ~250K ops/s (+80%)
    Each route matches 1 queue among 10K unique exact keys plus
    10K queues sharing "broadcast.#". v3 stored bindings in the same
    ETS row as the trie edge, so every trie lookup copied the entire
    sets:set(). v4 separates trie edges (small rows, set table) from
    bindings (ordered_set), so the trie walk copies only references.

  Large fanout (10K queues, same key): ~3,100 vs ~1,170 ops/s (+165%)
    v3 copied a 10K-element sets:set() out of ETS in a single
    ets:lookup, then called sets:to_list/1. v4 uses ets:select/2
    with a partially bound key, which does an O(log N) seek and
    then an efficient O(F) range scan without intermediate set
    conversion.

  MQTT broadcast (10K fanout): ~0.6 vs ~0.9 ms/route (+50%)
    Same mechanism as above.

Scenarios with no significant change (within benchmark noise):

  Exact match, wildcard *, wildcard #, mixed wildcards, and many
  wildcard filters showed no clear difference. Both v3 and v4 use
  a trie walk, so routing speed is comparable when the fanout is
  small and the bottleneck is trie traversal rather than destination
  collection.

### DIFF
--- a/deps/rabbit/src/rabbit_khepri.erl
+++ b/deps/rabbit/src/rabbit_khepri.erl
@@ -1708,7 +1708,7 @@ register_rabbit_topic_trie_projection() ->
            end,
     Projection = khepri_projection:new(rabbit_khepri_topic_trie_v4, PFun, Opts),
     PathPattern = topic_binding_path_pattern(),
-    _ = unregister_old_rabbit_topic_trie_projections(),
+    unregister_old_rabbit_topic_trie_projections(),
     khepri:register_projection(?STORE_ID, PathPattern, Projection).
 
 topic_binding_path_pattern() ->
@@ -1799,9 +1799,11 @@ trie_node_has_bindings(BindingTab, NodeId) ->
         _ -> false
     end.
 
+-spec supports_rabbit_khepri_topic_trie_v2() -> boolean().
 supports_rabbit_khepri_topic_trie_v2() ->
     true.
 
+-spec supports_rabbit_khepri_topic_trie_version() -> non_neg_integer().
 supports_rabbit_khepri_topic_trie_version() ->
     4.
 
--- a/deps/rabbit/src/rabbit_db_topic_exchange.erl
+++ b/deps/rabbit/src/rabbit_db_topic_exchange.erl
@@ -46,7 +46,7 @@ match(#resource{virtual_host = VHost, name = XName} = X, RoutingKey, Opts) ->
                     []
             end;
         _ ->
-            trie_match_in_khepri(X, Words, BKeys)
+            trie_match_v3(X, Words, BKeys)
     end.
 
 -spec split_topic_key_binary(RoutingKey) -> Words when
@@ -170,69 +170,49 @@ format_dest_bkeys([{#resource{kind = queue} = Dest, BKey} | Rest], Acc) ->
 format_dest_bkeys([{Dest, _BKey} | Rest], Acc) ->
     format_dest_bkeys(Rest, [Dest | Acc]).
 
-%% --------------------------------------------------------------
-%% Internal
-%% --------------------------------------------------------------
+%% ==============================================================
+%% Old v3 Khepri topic graph.
+%% Delete these *_v3 functions when feature flag
+%% topic_binding_projection_v4 becomes required.
+%% ==============================================================
 
--spec add_matched([rabbit_types:binding_destination() |
-                   {rabbit_types:binding_destination(), BindingArgs :: list()}],
-                  ReturnBindingKeys :: boolean(),
-                  match_result()) ->
-    match_result().
-add_matched(Destinations, false, Acc) ->
-    Destinations ++ Acc;
-add_matched(DestinationsArgs, true, Acc) ->
-    lists:foldl(
-      fun({DestQ = #resource{kind = queue}, BindingArgs}, L) ->
-              case rabbit_misc:table_lookup(BindingArgs, <<"x-binding-key">>) of
-                  {longstr, BKey} ->
-                      [{DestQ, BKey} | L];
-                  _ ->
-                      [DestQ | L]
-              end;
-         ({DestX, _BindingArgs}, L) ->
-              [DestX | L]
-      end, Acc, DestinationsArgs).
-
-%% Khepri topic graph
-
-trie_match_in_khepri(X, Words, BKeys) ->
+trie_match_v3(X, Words, BKeys) ->
     try
-        trie_match_in_khepri(X, root, Words, BKeys, [])
+        trie_match_v3(X, root, Words, BKeys, [])
     catch
         error:badarg ->
             []
     end.
 
-trie_match_in_khepri(X, Node, [], BKeys, ResAcc0) ->
-    Destinations = trie_bindings_in_khepri(X, Node, BKeys),
-    ResAcc = add_matched(Destinations, BKeys, ResAcc0),
-    trie_match_part_in_khepri(
+trie_match_v3(X, Node, [], BKeys, ResAcc0) ->
+    Destinations = trie_bindings_v3(X, Node, BKeys),
+    ResAcc = add_matched_v3(Destinations, BKeys, ResAcc0),
+    trie_match_part_v3(
       X, Node, <<"#">>,
-      fun trie_match_skip_any_in_khepri/5, [], BKeys, ResAcc);
-trie_match_in_khepri(X, Node, [W | RestW] = Words, BKeys, ResAcc) ->
+      fun trie_match_skip_any_v3/5, [], BKeys, ResAcc);
+trie_match_v3(X, Node, [W | RestW] = Words, BKeys, ResAcc) ->
     lists:foldl(fun ({WArg, MatchFun, RestWArg}, Acc) ->
-                        trie_match_part_in_khepri(
+                        trie_match_part_v3(
                           X, Node, WArg, MatchFun, RestWArg, BKeys, Acc)
-                end, ResAcc, [{W, fun trie_match_in_khepri/5, RestW},
-                              {<<"*">>, fun trie_match_in_khepri/5, RestW},
+                end, ResAcc, [{W, fun trie_match_v3/5, RestW},
+                              {<<"*">>, fun trie_match_v3/5, RestW},
                               {<<"#">>,
-                               fun trie_match_skip_any_in_khepri/5, Words}]).
+                               fun trie_match_skip_any_v3/5, Words}]).
 
-trie_match_part_in_khepri(X, Node, Search, MatchFun, RestW, BKeys, ResAcc) ->
-    case trie_child_in_khepri(X, Node, Search) of
+trie_match_part_v3(X, Node, Search, MatchFun, RestW, BKeys, ResAcc) ->
+    case trie_child_v3(X, Node, Search) of
         {ok, NextNode} -> MatchFun(X, NextNode, RestW, BKeys, ResAcc);
         error          -> ResAcc
     end.
 
-trie_match_skip_any_in_khepri(X, Node, [], BKeys, ResAcc) ->
-    trie_match_in_khepri(X, Node, [], BKeys, ResAcc);
-trie_match_skip_any_in_khepri(X, Node, [_ | RestW] = Words, BKeys, ResAcc) ->
-    trie_match_skip_any_in_khepri(
+trie_match_skip_any_v3(X, Node, [], BKeys, ResAcc) ->
+    trie_match_v3(X, Node, [], BKeys, ResAcc);
+trie_match_skip_any_v3(X, Node, [_ | RestW] = Words, BKeys, ResAcc) ->
+    trie_match_skip_any_v3(
       X, Node, RestW, BKeys,
-      trie_match_in_khepri(X, Node, Words, BKeys, ResAcc)).
+      trie_match_v3(X, Node, Words, BKeys, ResAcc)).
 
-trie_child_in_khepri(X, Node, Word) ->
+trie_child_v3(X, Node, Word) ->
     case ets:lookup(
            ?KHEPRI_PROJECTION_V3,
            #trie_edge{exchange_name = X,
@@ -242,7 +222,7 @@ trie_child_in_khepri(X, Node, Word) ->
         []                                        -> error
     end.
 
-trie_bindings_in_khepri(X, Node, BKeys) ->
+trie_bindings_v3(X, Node, BKeys) ->
     case ets:lookup(
            ?KHEPRI_PROJECTION_V3,
            #trie_edge{exchange_name = X,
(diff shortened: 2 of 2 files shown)


deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -1533,93 +1534,192 @@ projection_fun_for_sets(MapFun) ->
...
+%% projection may be triggered before the binding projection's table is
+%% initialised due to Khepri's prepend-based list ordering.
+ensure_topic_binding_table(Name) ->
+    case ets:whereis(Name) of
+        undefined ->
+            ets:new(Name, [ordered_set, named_table, protected,
+                           {read_concurrency, true}]);
+        Tid ->
+            Tid
+    end.
```
ansd: @dcorbacho @dumbbell @the-mikedavis How can we register two Khepri projections that depend on each other?

dumbbell: After thinking about this a bit more, what about managing multiple ETS tables from a single projection?

In other words:
1. You register a projection with some options:
    ```
    #{tables => #{rabbit_khepri_topic_trie_v4 => EtsOptions,
                  rabbit_khepri_topic_binding_v4 => EtsOptions}}
    ```
2. Khepri continues to create the ETS tables with the specified options
3. When a projection is triggered, the callback is called with a map of table name/IDs (today, it’s called with a table ID as it’s first argument)

What do you think? In the context of this patch, you would have a single callback to implement that handles both ETS tables.

ansd: Our goal is that at the point in time the projection is triggered both ETS tables must be present.
I think your suggestion solves this. So, yes, that would be perfect, thank you.

ansd:
I mark this PR as `Ready for review` but we should **not merge** this PR until all of the following are done:
- [x] More benchmarks checking for any major regressions (e.g. via MQTT or AMQP or definition import workloads) @mkuratczyk couldn't spot regressions. @the-mikedavis in case you have still have your benchmark tools available from your previous topic exchange work, could you please also run them against this PR?
- [x] User feedback
- [x] Khepri changes discussed in https://github.com/rabbitmq/rabbitmq-server/pull/15619#discussion_r2877029123

dcorbacho:
> I mark this PR as `Ready for review` but we should **not merge** this PR until all of the following are done:
> 
> * [x]  More benchmarks checking for any major regressions (e.g. via MQTT or AMQP or definition import workloads) @mkuratczyk couldn't spot regressions. @the-mikedavis in case you have still have your benchmark tools available from your previous topic exchange work, could you please also run them against this PR?
> * [x]  User feedback
> * [x]  Khepri changes discussed in [Replace Khepri topic routing projection with trie + ordered_set (v4) #15619 (comment)](https://github.com/rabbitmq/rabbitmq-server/pull/15619#discussion_r2877029123)

I run some of the import benchmarks we have from previous Khepri migration work, and couldn't find any regression. Import times are roughly the same.

the-mikedavis:
I gave this branch a run and it looks very similar to main using the topic routing perf-test command I have from old tests:

```
perf-test -e amq.topic -t topic -qp q.%d.a.b.c.d.e.f.g -qpf 1 -qpt 64 -x 64 -y 64 -c 3000 -z 60 -s 12 -E
```

Main
```
id: test-114742-171, sending rate avg: 46403 msg/s
id: test-114742-171, receiving rate avg: 46413 msg/s
```

This branch
```
id: test-114913-857, sending rate avg: 46141 msg/s
id: test-114913-857, receiving rate avg: 46160 msg/s
```

The average send/receive throughput varies pretty widely even between runs on the same branch, but it looks pretty similar here so I don't think there's a regression.

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -179,7 +179,7 @@
 
 %% Called remotely to handle unregistration of old projections.
 -export([supports_rabbit_khepri_topic_trie_v2/0,
-         supports_rabbit_khepri_topic_trie_version/0]).
+         supported_topic_trie_version/0]).
```
dumbbell: You can’t rename this function, it is called from remote nodes.

If you want to use a new name, you have to add a new function and support calling the old name as a fallback.

ansd: Thanks for catching this one

PR #15627, opened 2026-03-04
HTTP API: refactor how QQ replica ops fetch node names

michaelklishin:
Follow-up to #15621 #15622 #15625.

### DIFF
--- a/deps/rabbitmq_management/src/rabbit_mgmt_wm_quorum_queue_replicas_grow.erl
+++ b/deps/rabbitmq_management/src/rabbit_mgmt_wm_quorum_queue_replicas_grow.erl
@@ -26,25 +26,22 @@ content_types_accepted(ReqData, Context) ->
   {[{'*', accept_content}], ReqData, Context}.
 
 resource_exists(ReqData, Context) ->
-  case rabbit_mgmt_util:id(node, ReqData) of
-    none -> {false, ReqData, Context};
-    Node ->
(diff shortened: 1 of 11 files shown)


deps/rabbitmq_management/src/rabbit_mgmt_wm_quorum_queue_replicas_grow.erl
```diff
@@ -26,25 +26,22 @@ content_types_accepted(ReqData, Context) ->
...
     fun([VHPattern, QPattern, Strategy], Body, _ReqData) ->
-      Membership = maps:get(<<"membership">>, Body, promotable),
-      rabbit_quorum_queue:grow(
-        rabbit_data_coercion:to_atom(NewReplicaNode),
-        VHPattern,
-        QPattern,
-        rabbit_data_coercion:to_atom(Strategy),
-        rabbit_data_coercion:to_atom(Membership))
+      Strat = rabbit_mgmt_nodes:safe_atom(Strategy, <<"strategy">>),
+      Membership0 = maps:get(<<"membership">>, Body, promotable),
```
the-mikedavis, line 42: Since `with_decode/4` converts the keys of the body map to existing atoms should we instead be fetching here with

```suggestion
      Membership0 = maps:get(membership, Body, promotable),
```

? I wonder if this was an existing bug in the grow API. I may just be reading this wrong though.

michaelklishin: In any case, that's a great catch, thank you!

PR #15655, opened 2026-03-06
Allow MFs (module-function pairs) in rabbit_mnesia_tables_to_khepri_db attr

LoisSotoLopez:
## Why

As discussed in rabbitmq/rabbitmq-server#15559, it would be useful to allow a module-function pair as value for the `rabbit_mnesia_tables_to_khepri_db` module attribute.

This PR implements so.

### DIFF
--- a/deps/rabbit/src/rabbit_khepri.erl
+++ b/deps/rabbit/src/rabbit_khepri.erl
@@ -2102,11 +2102,30 @@ discover_mnesia_tables_to_migrate1([], MigrationsPerApp) ->
               Acc ++ maps:get(App, MigrationsPerApp)
       end, [], Apps).
 
+%% The MFA implementation must not assume any Rabbit sub-system is ready. Also,
+%% it must be able to run even when provided by a plugin and said plugin is not
+%% running.
+mnesia_tables_from_mfa(Mod, Fun, Args) ->
+    Ret = apply(Mod, Fun, Args),
+    ?assert(is_list(Ret)),
+    ?assert(lists:all(fun(Table) -> is_atom(Table) end, Ret)),
+    Ret.
+
 do_migrate_mnesia_tables(FeatureName, Migrations) ->
-    Tables = lists:map(
+    Tables = lists:flatmap(
                fun
-                   ({Table, _Mod}) when is_atom(Table) -> Table;
-                   (Table) when is_atom(Table)         -> Table
+                   ({Table, _Mod}) when is_atom(Table) ->
+                       [Table];
+                   (Table) when is_atom(Table) ->
+                       [Table];
+                   ({{mfa, Mod, Fun, Args}, _Mod}) when is_atom(Mod),
+                                                        is_atom(Fun),
+                                                        is_list(Args) ->
+                       mnesia_tables_from_mfa(Mod, Fun, Args);
+                   ({mfa, Mod, Fun, Args}) when is_atom(Mod),
+                                                is_atom(Fun),
+                                                is_list(Args) ->
+                       mnesia_tables_from_mfa(Mod, Fun, Args)
                end,
                Migrations),
     ?LOG_NOTICE(


LoisSotoLopez:
Any suggestions/requirements to get this PR moving? :)

michaelklishin:
@dumbbell do you have an opinion?

gomoripeti:
do you think it will be possible to include a solution for this in a 4.2.x release (not just 4.3.x)? it would allow a continuous (3.13..4.1 w mnesia) -> 4.2 (khepri) -> 4.3 (khepri) upgrade path without disabling/enabling a plugin.

michaelklishin:
@gomoripeti we are not aware of any plugins that need to be disabled for 3.13 to upgrade to 4.2 and then 4.3. What plugin is that?

michaelklishin:
Looks like the plugin is the message deduplication one #15559, a tier 2 plugin and not a tier 1 (that ships with RabbitMQ).

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -2103,10 +2103,16 @@ discover_mnesia_tables_to_migrate1([], MigrationsPerApp) ->
       end, [], Apps).
 
 do_migrate_mnesia_tables(FeatureName, Migrations) ->
-    Tables = lists:map(
+    Tables = lists:flatmap(
                fun
-                   ({Table, _Mod}) when is_atom(Table) -> Table;
-                   (Table) when is_atom(Table)         -> Table
+                   ({{mf, Mod, Fun}, _Mod}) when is_atom(Mod), is_atom(Fun) ->
+                       apply(Mod, Fun, []);
```
dumbbell: What about accepting a list of arguments?
```erlang
{mfa, Mod, Fun, Args}
```

This allows the user to use one module+function for several purposes.

dumbbell: Could you please add some assertions to verify the return value of the MFA?

Like:
```erlang
?assert(is_list(Ret)),
?assert(lists:all(fun erlang:is_atom/1, Ret)),
```

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -2103,10 +2103,16 @@ discover_mnesia_tables_to_migrate1([], MigrationsPerApp) ->
       end, [], Apps).
 
 do_migrate_mnesia_tables(FeatureName, Migrations) ->
-    Tables = lists:map(
+    Tables = lists:flatmap(
                fun
-                   ({Table, _Mod}) when is_atom(Table) -> Table;
-                   (Table) when is_atom(Table)         -> Table
+                   ({{mf, Mod, Fun}, _Mod}) when is_atom(Mod), is_atom(Fun) ->
```
dumbbell: Could you please move the handling of MFA after the handling of `Table`? `{Table, Mod}` and `Table` will be way more common, let’s put them first.

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -2103,10 +2103,22 @@ discover_mnesia_tables_to_migrate1([], MigrationsPerApp) ->
...
-                   (Table) when is_atom(Table)         -> Table
+                   ({Table, _Mod}) when is_atom(Table) ->
+                       [Table];
+                   (Table) when is_atom(Table) ->
+                       [Table];
+                   ({{mfa, Mod, Fun, Args}, _Mod}) when is_atom(Mod), is_atom(Fun), is_list(Args) ->
+                       Ret = apply(Mod, Fun, Args),
+                       ?assert(is_list(Ret)),
+                       ?assert(lists:all(fun(Table) -> is_atom(Table) end, Ret)),
+                       Ret;
```
dumbbell: After seeing this, could you please a function (with type spec) out of this to avoid code duplication?

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -2102,11 +2102,23 @@ discover_mnesia_tables_to_migrate1([], MigrationsPerApp) ->
...
-    Tables = lists:map(
+    Tables = lists:flatmap(
                fun
-                   ({Table, _Mod}) when is_atom(Table) -> Table;
-                   (Table) when is_atom(Table)         -> Table
+                   ({Table, _Mod}) when is_atom(Table) ->
+                       [Table];
+                   (Table) when is_atom(Table) ->
+                       [Table];
+                   ({{mfa, Mod, Fun, Args}, _Mod}) when is_atom(Mod), is_atom(Fun), is_list(Args) ->
```
dumbbell: A cosmetic comment: could you please wrap this long line to fit in the 80-column limit?

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -2102,11 +2102,23 @@ discover_mnesia_tables_to_migrate1([], MigrationsPerApp) ->
...
+                       mnesia_tables_from_mfa(Mod, Fun, Args);
+                   ({mfa, Mod, Fun, Args}) when is_atom(Mod), is_atom(Fun), is_list(Args) ->
```
dumbbell: Likewise for this one :-)

dumbbell (changes requested):

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -2102,11 +2102,23 @@ discover_mnesia_tables_to_migrate1([], MigrationsPerApp) ->
               Acc ++ maps:get(App, MigrationsPerApp)
       end, [], Apps).
 
+mnesia_tables_from_mfa(Mod, Fun, Args) ->
+    Ret = apply(Mod, Fun, Args),
```
dumbbell, line 2109: Could you please add a comment here and in the commit message about the fact that the called function must not assume that any RabbitMQ subsystems are ready? And that the function must work without the plugin running if it is part of a plugin?

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -2102,11 +2102,30 @@ discover_mnesia_tables_to_migrate1([], MigrationsPerApp) ->
               Acc ++ maps:get(App, MigrationsPerApp)
       end, [], Apps).
 
+% The plugin-provided MFA must not assume any Rabbit sub-system is ready. Also,
```
dumbbell: This statement is also true if we use this MFA feature in a module belonging to RabbitMQ core, this is only true for plugins.

Also please use `%%` for comments.

PR #15667, opened 2026-03-07
rabbit_db: complete m2k rework

mkuratczyk:
TODO: review the tests to make sure such issues are caught

Follow-up to commit 3b8718b385
(https://github.com/rabbitmq/rabbitmq-server/pull/15392)

Plugin converters still used to old API,so upgrades were failing.

Additionally, fix test upgrades flakes (node not starting)


dumbbell:
I reviewed and approved the patch out of GitHub.

Thank you for the fix!

PR #15674, opened 2026-03-09
Add peek function to amqp10 parser

MarcialRosales:
## Proposed Changes

This PR adds a peek function to the amqp10 parser so that it is possible to know, before parsing the entire coming Amqp 10 section, its total size and the sections's type/name. This PR was generated on the back of the stream browser plugin's PR. 
The goal is to avoid having to parse an entire AMQP 1.0 section which can be a large AMQP data section or a metada section with large number of entries which the stream browser does not need to stream to the client, or it could be a section whose size exceeds the maximum size the stream browser should stream to the client (the management ui).

The PR adds another unrelated change which is a change to stream_test_util.erl module, in particular, a function for compressed subbatches. This change was required to pass tests which tested compressed sub-batches.

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/amqp10_common/src/amqp10_binary_parser.erl
+++ b/deps/amqp10_common/src/amqp10_binary_parser.erl
@@ -16,7 +16,9 @@
 -include("amqp10_framing.hrl").
 
 -export([parse/1,
-         parse_many/2]).
+         parse_many/2,
+         peek/1,
+         peek_value_size/1]).
 
 %% §1.6
 -define(CODE_ULONG, 16#80).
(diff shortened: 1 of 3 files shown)


deps/amqp10_common/src/amqp10_binary_parser.erl
```diff
@@ -338,3 +340,82 @@ pm_compound(UnitSize, Bin, O, B) ->
...
+    <<_:B1/binary, Rest1/binary>> = Rest,
+    B2 = peek_value_size(Rest1),
+    {section_descriptor_to_atom(Descriptor), 1 + B1 + B2};
+peek(<<Type, _/binary>>) ->
+    throw({not_described_type, Type}).
+
+section_descriptor_to_atom(Descriptor) ->
+    case Descriptor of
+        {ulong, 16#70} -> 'v1_0.header';
+        {ulong, 16#71} -> 'v1_0.delivery_annotations';
```
ansd: If this function stays, can we use the macros defined above, e.g. `DESCRIPTOR_CODE_APPLICATION_PROPERTIES` for `16#71`.

deps/amqp10_common/src/amqp10_binary_parser.erl
```diff
@@ -338,3 +340,82 @@ pm_compound(UnitSize, Bin, O, B) ->
...
+        {symbol, <<"amqp:header:list">>} -> 'v1_0.header';
+        {symbol, <<"amqp:delivery-annotations:map">>} -> 'v1_0.delivery_annotations';
+        {symbol, <<"amqp:message-annotations:map">>} -> 'v1_0.message_annotations';
+        {symbol, <<"amqp:properties:list">>} -> 'v1_0.properties';
+        {symbol, <<"amqp:application-properties:map">>} -> 'v1_0.application_properties';
+        {symbol, <<"amqp:data:binary">>} -> 'v1_0.data';
+        {symbol, <<"amqp:amqp-sequence:list">>} -> 'v1_0.amqp_sequence';
+        {symbol, <<"amqp:amqp-value:*">>} -> 'v1_0.amqp_value';
+        _ -> undefined
+    end.
```
ansd: This function returns atoms such as `'v1_0.header'`, `v1_0.delivery_annotations'`. Maybe this module should not return these atoms. These atoms/records are defined separately in separate header files already.
Rather than having section_descriptor_to_atom/1 in amqp10_binary_parser, the peek/1 function could return the raw parsed descriptor (e.g. {ulong, 16#75}) and let the caller decide what to do with it. The stream browser plugin can then either use this function as is or call element(1, amqp10_framing0:record_for(Descriptor)) itself to get the atom (if that's acceptable for performance):
  • The caller (the stream browser) already knows it's dealing with AMQP message sections and already depends on amqp10_framing. It can trivially do the mapping itself
  • It avoids duplicating the descriptor table that amqp10_framing0 already maintains as the single source of truth
  • The footer section (16#78) is missing from section_descriptor_to_atom but is present in amqp10_framing0:record_for/1 -- this kind of drift is what happens with duplicated mappings

MarcialRosales: Thank you very much @ansd for reviewing and these suggestions. I am going to follow your recommendation to use the existing mappings.

ansd:
Thank you @MarcialRosales !

PR #15683, opened 2026-03-10
Update release notes for 4.3.0

lhoguin:
This includes my own changes feel free to use it as basis for larger release notes update.

I removed duplicated text as well, likely an error from a while back.


michaelklishin:
Merged manually and backported to `v4.3.x` with a few edits.

PR #15689, opened 2026-03-10
Allow filtering `/metrics/detailed` queue metrics by queue name

ansd:
The `/metrics/detailed` endpoint now accepts an optional `queue` query parameter whose value is a regular expression. When provided, only queues whose names match the regex are included in the response. This applies to all queue-related metric families (queue_coarse_metrics, queue_metrics, queue_consumer_count, queue_delivery_metrics, queue_exchange_metrics, stream_consumer_metrics, ra_metrics) as well as the queue_info metric.

The filter can be combined with the existing `vhost` filter.

Examples:
```
/metrics/detailed?family=queue_coarse_metrics&queue=^quorum-
/metrics/detailed?family=queue_metrics&vhost=prod&queue=^orders-
```

This is useful for deployments with a large number of queues where only a subset needs to be monitored. For example, a cluster with thousands of classic queues but only a handful of quorum queues following a naming convention can now scrape metrics for just those quorum queues, significantly reducing the metrics payload size, Prometheus scrape time, and storage costs.

Only one `queue` parameter is accepted per request. Providing multiple values returns HTTP 400. An invalid regex also returns HTTP 400 with a descriptive error message.

Docs PR: https://github.com/rabbitmq/rabbitmq-website/pull/2467

### DIFF
--- a/deps/rabbitmq_prometheus/src/rabbit_prometheus_handler.erl
+++ b/deps/rabbitmq_prometheus/src/rabbit_prometheus_handler.erl
@@ -183,6 +183,8 @@ parse_metric_families(_) ->
 
 parse_queue_filter(undefined) ->
     false;
+parse_queue_filter(<<>>) ->
+    false;
 parse_queue_filter(B) when byte_size(B) > ?MAX_QUEUE_FILTER_LENGTH ->
     {error, <<"'queue' regex too long">>};
 parse_queue_filter(B) when is_binary(B) ->
--- a/deps/rabbitmq_prometheus/test/rabbit_prometheus_http_SUITE.erl
+++ b/deps/rabbitmq_prometheus/test/rabbit_prometheus_http_SUITE.erl
@@ -86,7 +86,8 @@ groups() ->
                                      queue_filter_queue_exchange_metrics_test,
                                      queue_filter_raft_metrics_test,
                                      queue_filter_combined_with_vhost_test,
-                                     queue_filter_multiple_patterns_rejected_test
+                                     queue_filter_multiple_patterns_rejected_test,
+                                     queue_filter_empty_is_ignored_test
         ]},
        {special_chars, [], [core_metrics_special_chars]},
        {authentication, [], [basic_auth]}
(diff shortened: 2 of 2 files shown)


deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
```diff
@@ -925,3 +952,32 @@ vhosts_filter_from_pdict() ->
...
+queue_filter_from_pdict() ->
+    case get(prometheus_queue_filter) of
+        undefined -> false;
+        MP -> MP
+    end.
+
+matches_queue_filter(_QName, false) ->
+    true;
+matches_queue_filter(QName, MP) ->
+    re:run(QName, MP, [{capture, none}]) =:= match.
```
michaelklishin: @ansd this matches queue names against a regex that comes from a user input: a classic ReDoS attack target.

`re:run/2` should use the [`match_limit` and `match_limit_depth` (previously `match_limit_recursion`) options](https://www.erlang.org/doc/apps/stdlib/re.html#t:option/0) for protection.

Those options express computational complexity and default to 10M. We can go
with 50K for both values, that'd be conservative and should not produce any false positives for practical inputs (say, up to 255 characters).

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_raft_metrics_collector.erl
```diff
@@ -96,15 +96,23 @@ collect_detailed_metrics(Prefix, Callback) ->
...
+    collect_all_matching_metrics(Prefix, Callback, FilterFun).
+
+build_filter_fun(VHostFilterFun, undefined) ->
+    VHostFilterFun;
+build_filter_fun(VHostFilterFun, QueueFilter) ->
+    fun(#{queue := Q} = Labels) ->
+            VHostFilterFun(Labels)
+                andalso re:run(Q, QueueFilter, [{capture, none}]) =:= match;
+       (_) ->
+            false
```
michaelklishin, line 112: Should this stop the filtering or delegate to the "underlying" filter like so?

```erl
build_filter_fun(VHostFilterFun, QueueFilter) ->
    fun(#{queue := Q} = Labels) ->
            VHostFilterFun(Labels)
                andalso re:run(Q, QueueFilter, [{capture, none}]) =:= match;
       (Labels) ->
            VHostFilterFun(Labels)
    end.
```

michaelklishin:
One other open ended question: should `?queue=` mean "match everything" like it currently does or should such empty filters be ignored entirely?

I personally think that an empty filter is as good as no filter provided at all, so it should be ignored to safe some CPU cycles.

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_raft_metrics_collector.erl
ansd: The current code returns `false` for any entry without a queue label (like Khepri/coordination Raft metrics), which means: "if you filter by queue name, you only get queue-related Raft metrics."
I think the current behavior is the right one. If a user explicitly asks for `?queue=my-queue`, they're saying "I only care about queue metrics."

ansd:
> One other open ended question: should ?queue= mean "match everything" like it currently does or should such empty filters be ignored entirely? I personally think that an empty filter is as good as no filter provided at all, so it should be ignored to safe some CPU cycles.

I agree. I pushed a commit.

deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_core_metrics_collector.erl
ansd: I pushed a commit

PR #15692, opened 2026-03-11
Eliminate a flake in `rabbit_fifo_dlx_integration_SUITE`

michaelklishin:
Killing a Ra member transfers leadership most of the time but not all of the time: the same node can be re-elected.

If the purpose of the test is to transfer leadership and observe a certain DLX worker behavior, why not transfer leadership to a specific node instead?

### DIFF
--- a/deps/rabbit/test/rabbit_fifo_dlx_integration_SUITE.erl
+++ b/deps/rabbit/test/rabbit_fifo_dlx_integration_SUITE.erl
@@ -948,17 +948,18 @@ single_dlx_worker(Config) ->
          0,
          length(rpc(Config, Server1, supervisor, which_children, [rabbit_fifo_dlx_sup], 1000)))),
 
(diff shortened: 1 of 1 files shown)


ansd:
If I remember correctly, the purpose was first stopping a node gracefully `rabbit_ct_broker_helpers:stop_broker/2` and then killing another one. Should we use `rabbit_ct_broker_helpers:kill_node/2` instead of `true = rpc(Config, Leader0, erlang, exit, [Pid, kill]),`?

The same test case also flakes for another reason: https://github.com/rabbitmq/rabbitmq-server/pull/15548#issuecomment-3958045171

michaelklishin:
@ansd if the idea in the test is to specifically kill a member, then `kill_node/2` makes sense and should be safer flakes-wise.

Otherwise transferring leadership cleanly makes more sense to me.

I have pushed a change that adopts `kill_node/2`.

PR #15694, opened 2026-03-11
Remove file_descriptor_limit alarm

lhoguin:
This alarm was effectively dead code since the removal of the file_handle_cache module. This commit removes the alarm along with the concept of "local-only alarms" (which could easily be restored from Git) as well as a Prometheus metric that shows this file_descriptor_limit alarm was in effect.

### DIFF
--- a/deps/rabbitmq_management/test/rabbit_mgmt_http_health_checks_SUITE.erl
+++ b/deps/rabbitmq_management/test/rabbit_mgmt_http_health_checks_SUITE.erl
@@ -34,6 +34,7 @@ groups() ->
      {cluster_size_5, [], [is_quorum_critical_test]},
      {single_node, [], [
                         alarms_test,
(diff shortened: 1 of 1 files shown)


michaelklishin:
We should reintroduce `check_local_alarms` as a no-op for backwards compatibility, similar to what we did for `node_health_check` in the past.

I'm on it.

michaelklishin:
Selenium failures are due to an earlier `main` branch change https://github.com/rabbitmq/rabbitmq-server/pull/15716.

lhoguin:
> We should reintroduce `check_local_alarms` as a no-op for backwards compatibility, similar to what we did for `node_health_check` in the past.
> 
> I'm on it.

I left them because I was under the impression that some cluster-wide alarms would be listed when they originated from the node we were querying. Don't know if useful and happy either way, just FYI.

PR #15706, opened 2026-03-11
rabbit_stream_coordinator: Make new_stream command idempotent

the-mikedavis:
This is similar to https://github.com/rabbitmq/rabbitmq-server/pull/14884 but for `new_stream`.

Although its unlikely from the calling code, the stream coordinator can end up with two `{new_stream, StreamId, #{}}` commands in its log for the same `StreamId`. When handling the second `new_stream` it will then warning-log that the stream can't be updated, like so:

```log
2026-03-10 17:31:38.897507+00:00 [warning] <0.14468.0> rabbit_stream_coordinator failed to update stream:
2026-03-10 17:31:38.897507+00:00 [warning] <0.14468.0> function_clause
2026-03-10 17:31:38.897507+00:00 [warning] <0.14468.0> [{rabbit_stream_coordinator,update_stream0,[#{index => 51,system_time => 1773163898892,reply_mode => await_consensus,machine_version => 6,...},{new_stream,[95|...],#{}},{stream,[...],...}],[{file,[114|...]},{line,1397}]},{rabbit_stream_coordinator,update_stream,3,[{file,[...]},{line,...}]},{rabbit_stream_coordinator,apply,3,[{file,...},{...}]},{ra_server,apply_with,2,[{...}|...]},{ra_log,fold,5,[...]},{ra_server,apply_to,5,...},{ra_server,evaluate_commit_index_follower,...},{ra_server,...}]
```

`update_stream0/3` only has a function clause for when there is no stream (i.e. undefined, <https://github.com/rabbitmq/rabbitmq-server/blob/0b951b8b288e4926ff198f38855729b38e876dc8/deps/rabbit/src/rabbit_stream_coordinator.erl#L1397-L1420>). I think it's reasonable to have the new_stream command become idempotent and effectively skip it if it is in the log twice.

I haven't been able to reproduce the scenario where this happened. Next time I will dump Khepri and the coordinator's logs to see what sequence of commands lead to this.

### DIFF
--- a/deps/rabbit/src/rabbit_stream_coordinator.erl
+++ b/deps/rabbit/src/rabbit_stream_coordinator.erl
@@ -1411,6 +1411,17 @@ filter_command(_Meta, {delete_stream, _StreamId, #{}}, undefined) ->
     %% Attempting to delete a stream which does not exist. Reply 'ok' to the
     %% caller so that this action is idempotent.
     {reply, ok};
+filter_command(#{machine_version := Vsn},
+               {new_stream, _StreamId, #{}},
+               #stream{members = Members}) when ?V7_OR_MORE(Vsn) ->
+    MaybeLeader = [Pid || _Node := #member{state = {running, _, Pid},
+                                           role = {writer, _}} <- Members],
+    case MaybeLeader of
+        [LeaderPid] ->
+            {reply, {ok, LeaderPid}};
+        [] ->
+            {reply, '$ra_no_reply'}
+    end;
 filter_command(_, _, _) ->
     ok.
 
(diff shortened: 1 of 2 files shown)


the-mikedavis:
I don't think this needs a new machine version since it only affects the reply effects. We could base it on #15695 to take advantage of version 7 though.

acogoluegnes:
LGTM. We can indeed wait for #15695 to be merged and use the v7+ macro in a guard on the new filter_command clause.

acogoluegnes:
@the-mikedavis #15695 has been merged, can you rebase and apply this change only after v7?

the-mikedavis:
Yep this is now using the `?V7_OR_MORE` macro in the `filter_command/3` guard

PR #15712, opened 2026-03-11
HTTP API: refactor `read_complete_body/1`

michaelklishin:
To terminate earlier.


michaelklishin:
The failures are unrelated Selenium suite flakes, I will try to address at least one of them in a separate PR.

michaelklishin:
Specifically, the failing tests are:

1. `amqp10-connections/sessions-for-monitoring-user.js`: depends on stats DB initialization timming
2. `oauth-idp-initiated-with-uaa-and-prefix/landing.js`: there is a race condition with JWT token expiration

PR #15703, opened 2026-03-11
Stream protocol: fix a crash in `parse_command/1` on zero-size frames

lukebakken:
When a client sends a zero-size frame — for example, a port scanner sending `<<0,0,0,0,0,0,0,0>>` to the TLS-enabled stream listeners port, `parse_command/1` is called with an empty binary `<<>>`. No existing clause matches, causing a `function_clause` crash that terminates the stream connection process. This has been observed consistently on EC2-hosted brokers.

Add a catch-all clause that returns `{unknown, Data}` for any binary that is not a valid request or response frame. `{unknown, binary()}` is already a valid `command()` type and is handled gracefully by all three frame handlers in `rabbit_stream_reader`: pre-auth closes the connection, post-auth sends a protocol-level close frame with `RESPONSE_CODE_UNKNOWN_FRAME`, and post-close ignores it.

The fix is accompanied by a test that reproduces the exact crash using the 8-byte payload observed in production logs.

PR #15713, opened 2026-03-12
Suppress spurious warnings from port scanner connections

lukebakken:
Follow-up to #15703, which fixed a crash when port scanners connect to the stream port. That fix revealed two remaining sources of log noise:

**1. Unrecognised pre-auth data logged at warning level**

`handle_frame_pre_auth/4` had a catch-all that logged `{unknown, Data}` at warning level. Since this is expected from port scanners and other non-RabbitMQ clients, a specific `{unknown, _}` clause is added that logs at debug level instead. The existing warning-level catch-all remains for any other unexpected pre-auth commands.

A specific `{unknown, _}` clause is also added to `handle_frame_post_auth/4` with a clearer log message (post-auth unrecognised data is more noteworthy and stays at warning). `send_close_and_increment/2` is extracted to eliminate the resulting duplication.

**2. Spurious "invalid transition" warnings from partial frames**

When incoming data does not contain a complete frame (e.g. a port scanner sends fewer bytes than a full frame header), no commands are parsed and `connection_step` remains unchanged. The five pre-`open` state handlers (`tcp_connected`, `peer_properties_exchanged`, `authenticating`, `tuning`, `tuned`) treated this as an invalid transition and logged a warning.

A guard is added to each handler that returns `keep_state` when `NextConnectionStep` equals the current state name, correctly waiting for more data. The existing negotiation timeout will still close connections that make no progress.

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_reader.erl
@@ -240,20 +240,28 @@ tcp_connected(info, Msg, StateData) ->
                 fun(NextConnectionStep,
                     #statem_data{transport = Transport,
                                  connection = #stream_connection{socket = S}} =
-                        StatemData,
+                    StatemData,
                     NewConnection,
                     NewConnectionState) ->
-                   if NextConnectionStep =:= peer_properties_exchanged ->
-                          {next_state, peer_properties_exchanged,
-                           StatemData#statem_data{connection = NewConnection,
-                                                  connection_state =
-                                                      NewConnectionState}};
-                      true ->
-                          invalid_transition(Transport,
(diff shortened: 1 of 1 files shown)


deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -240,20 +240,28 @@ tcp_connected(info, Msg, StateData) ->
                 fun(NextConnectionStep,
                     #statem_data{transport = Transport,
                                  connection = #stream_connection{socket = S}} =
-                        StatemData,
```
lukebakken, line 243: FYI I re-formatted these functions using `=` in neovim.

PR #15716, opened 2026-03-12
Selenium suite fixes (Mar 11, 2026)

michaelklishin:
1. Use AMQP 1.0 v2 addresses as v1 are disabled by default in `main`
2. There was a race condition in one of the tests between the assertions and the JWT token lifetime (expiration)


michaelklishin:
This cannot possibly affect upgrade failures. A ton of jobs fail on Actions at the moment :(

PR #15728, opened 2026-03-12
Management UI: avoid a confusing error message right after the password is changed

michaelklishin:
When the user changes their password using
the management UI, the request that updates
the credentials will race with an auto-refresh
request, and likely run into a 401 error.

Before the UI is modernized, this is the least
hacky option available to us.

References #15047.

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/dispatcher.js
+++ b/deps/rabbitmq_management/priv/www/js/dispatcher.js
@@ -248,12 +248,13 @@ dispatcher_add(function(sammy) {
             if (isOwnUser) {
                 pause_auto_refresh();
                 window.own_creds_changed_at = Date.now();
-                setTimeout(resume_auto_refresh, 15000);
+                window.own_creds_resume_timer = setTimeout(resume_auto_refresh, 15000);
             }
             if (sync_put(this, '/users/:username')) {
                 go_to('#/users');
             } else {
                 if (isOwnUser) {
+                    clearTimeout(window.own_creds_resume_timer);
                     resume_auto_refresh();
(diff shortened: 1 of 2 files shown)


deadtrickster:
I like this approach - pause refreshes and have guarding timers.

Minor thing:
I tried it and get some JS errors:

```
___ViewO.push("<div class=\"updatable\">\n");
___ViewO.push("    ");  if (users.items.length > 0) { ; ___ViewO.push("\n");
```

items is undefined here. 

The orange bottom banner says this: 
```
TypeError: Cannot read properties of undefined (reading 'length') TypeError: Cannot read properties of undefined (reading 'length') at Array.process (eval at compile (http://localhost:15672/js/ejs-1.0.min.js:1:6654), :6:41) at EJS.render (http://localhost:15672/js/ejs-1.0.min.js:1:1885) at format (http://localhost:15672/js/main.js:1358:21) at http://localhost:15672/js/main.js:632:24 at with_reqs (http://localhost:15672/js/main.js:1340:9) at http://localhost:15672/js/main.js:1336:17 at req.onreadystatechange (http://localhost:15672/js/main.js:1428:21)
```

I checked the sources in the console, they look up-to-date.

If yours doesnt do these errors, probably something still cached by my browser.

michaelklishin:
@deadtrickster so in order to keep the callback chain executing, we need to pass in a dummy response to the `with_req`, `with_reqs`, `with_update` chain and that multiplies the hacks.

michaelklishin:
The more time I spend on this change, the more I realize that I like the "generate your seed user's password at deployment time" workaround better.

A modernized UI should give us more options.

PR #15730, opened 2026-03-13
An attempt to fix 401 Not Authorized after self password change

ikavgo:
References https://github.com/rabbitmq/rabbitmq-server/discussions/15047.

@michaelklishin I don't have any JS or 401 errors with this approach.


michaelklishin:
@ikavgo I will take a look. Thank you for not giving up as quickly as I have.

PR #15731, opened 2026-03-13
Add function to clean up and re-evaluate stream SAC group

acogoluegnes:
The function is meant to be used by an operator as a last resort to clean up a consumer group from "ghost" consumers and unblock the group.
At this point the SAC coordinator should work as expected but an upgrade from a version with known issues can end up with blocked groups. So this function should help unblock the groups manually.

It can be called like this:

```shell
rabbitmqctl eval 'rabbit_stream_coordinator:evaluate_sac_group(<<"/">>,
                                                               <<"stream-0">>,
                                                               <<"app1">>).'
```

The actual call is hidden behind the 4.3.0 feature flag. This ensures all the nodes have been upgraded and are using the latest machine version.

References rabbitmq/rabbitmq-stream-dotnet-client#447

### DIFF
--- a/deps/rabbit/src/rabbit_stream_sac_coordinator.erl
+++ b/deps/rabbit/src/rabbit_stream_sac_coordinator.erl
@@ -24,7 +24,8 @@
                      #command_activate_consumer{} |
                      #command_connection_reconnected{} |
                      #command_purge_nodes{} |
-                     #command_update_conf{}.
+                     #command_update_conf{} |
+                     #command_evaluate_group{}.
 
 -opaque state() :: #?MODULE{}.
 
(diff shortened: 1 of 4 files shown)


deps/rabbit/src/rabbit_stream_sac_coordinator.erl
```diff
@@ -145,13 +147,78 @@ activate_consumer(VH, Stream, Name) ->
...
+    lists:filter(fun(Pid) -> not is_pid_alive(Pid) end, Pids).
+
+is_pid_alive(Pid) when node(Pid) =:= node() ->
+    erlang:is_process_alive(Pid);
+is_pid_alive(Pid) ->
+    try
+        erpc:call(node(Pid), erlang, is_process_alive, [Pid], 5000)
+    catch
+        _:_ ->
+            false
```
kjnilsson, line 218: Not sure it is safe to just say false here.

PR #15737, opened 2026-03-16
Update setup_beam to @v1 everywhere applicable

lhoguin:
Setup_beam v1.23 now includes the ci.erlang.mk problem matchers so we no longer need to install those before running tests.


lhoguin:
This does not need to be backported as v4.2.x does not have the problem matchers.

PR #15743, opened 2026-03-17
Optimise annotation expansion for mc_amqp

kjnilsson:


ansd (approved):

michaelklishin:
There are some test gaps but they are not new, so can be addressed with a follow-up PR when the pre-`4.3.0` release dust settles.

PR #15745, opened 2026-03-17
`epmd` monitor: do not assume that the port is guaranteed to be discovered by the time this code runs

michaelklishin:
Suggested by @NeilFraser in the website repo.

This addresses a log exception.


michaelklishin:
@mergifyio backport `v4.2.x`

michaelklishin:
Backported manually for `4.2.6`.

PR #15755, opened 2026-03-18
Add OSS changes for `rabbitmq_jms` plugin

ansd:
This commit adds OSS changes required for the new VMware Tanzu RabbitMQ `rabbitmq_jms` plugin.

The deleted tests in this commit moved to `~/tanzu-rabbitmq/deps/rabbitmq_jms/test/`.


deps/rabbit/src/rabbit_quorum_memory_manager.erl
```diff
@@ -1,67 +1,57 @@
-%% This Source Code Form is subject to the terms of the Mozilla Public
```
MirahImage, line 1: This change seems unintentional

PR #15758, opened 2026-03-18
Mark several old feature flags as required

dumbbell:
## Why

User have to upgrade to RabbitMQ 4.2.x first before jumping to 4.3.x+.

In particular, `rabbit_exchange_type_local_random` and `detailed_queues_endpoint` were introduced a long time ago.


deps/rabbit/src/rabbit_core_ff.erl
```diff
@@ -207,14 +207,14 @@
...
-      stability     => stable,
+      stability     => required,
       depends_on    => ['rabbitmq_4.0.0']
      }}).
 
 -rabbit_feature_flag(
    {'rabbitmq_4.2.0',
     #{desc          => "Allows rolling upgrades to 4.2.x",
-      stability     => stable,
+      stability     => ,
```
ansd: typo ?

dumbbell: Looks like it :-)

PR #11415, opened 2024-06-07
cuttlefish tls schema for amqp_client

SimonUnge:
## Proposed Changes

With OTP 26, would be nice to config some amqp client settings easier, for example when using the federation plugin.

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
If you're unsure about any of them, don't hesitate to ask on the mailing list.
We're here to help!
This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
- [x] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [x] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

## Further Comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution
you did and what alternatives you considered, etc.

PR #15767, opened 2026-03-18
Re-submit cuttlefish tls schema for `amqp_client`

lukebakken:
Follow-up PR to the following:
* Original PR - https://github.com/rabbitmq/rabbitmq-server/pull/11415
* PR that reverted the original - https://github.com/rabbitmq/rabbitmq-server/pull/11531

Having cuttlefish schemas for `amqp_client` and `amqp10_client` still seems useful, so I looked into why this was reverted. Since bazel is no longer used, that reason no longer exists. In my local env, all of these test runs succeed just fine:

```
make -C deps/rabbit ct-config_schema
make -C deps/amqp10_client tests
make -C deps/amqp_client tests
```

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -2940,6 +2940,218 @@ fun(Conf) ->
     end
 end}.
 
+%% ----------------------------------------------------------------------------
+%%  AMQP client 1.0 TLS options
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you, @lukebakken. Re-submitted with some follow-up changes from me https://github.com/rabbitmq/rabbitmq-server/pull/15772.

PR #15782, opened 2026-03-19
Terminate when ra_systems_sup exits

the-mikedavis:
The goal of this change is to shut down Rabbit when the `ra_systems_sup` supervisor exits. If disk is completely exhausted then the `ra_systems_sup` supervisor can exit from the repeated `enospc` errors. If we do not also crash Rabbit, Rabbit continues on but Khepri is unavailable, so no incoming connections can successfully log in. Plus there are other effects like `rabbit_vhost_process` deleting a vhost because Khepri does not say that it exists.

Connects to https://github.com/rabbitmq/ra/issues/585

### DIFF
--- a/deps/rabbit/src/rabbit_ra_systems.erl
+++ b/deps/rabbit/src/rabbit_ra_systems.erl
@@ -7,6 +7,8 @@
 
 -module(rabbit_ra_systems).
 
+-behaviour(gen_server).
+
 -include_lib("kernel/include/logger.hrl").
 
 -include_lib("rabbit_common/include/logging.hrl").
@@ -19,6 +21,21 @@
          ensure_ra_system_stopped/1,
          ensure_started/0,
          ensure_stopped/0]).
+
+-export([start_link/0,
+         init/1,
+         handle_call/3,
+         handle_cast/2,
+         handle_info/2,
+         terminate/2,
+         code_change/3]).
+-export([stop/0]).
+
+-rabbit_boot_step({rabbit_ra_systems_monitor,
+                   [{description, "monitor process to shut down when Ra systems exit"},
+                    {mfa,         {rabbit_sup, start_child, [?MODULE]}},
+                    {requires,    [kernel_ready]},
+                    {cleanup,     {?MODULE, stop, []}}]}).
 
 -type ra_system_name() :: atom().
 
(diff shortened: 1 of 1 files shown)


deps/rabbit/src/rabbit_ra_systems.erl
the-mikedavis, line 1: Using an extra gen_server just to crash the Rabbit app seems a bit hacky to me, though it works. I'm sure there must be a better way to crash Rabbit when ra_systems_sup exits.

the-mikedavis:
Here is a full log from a `make run-broker` on the current main (4b59c196417b2f6f858db5fcfcb82027268d23d1) run where ra_systems_sup exits from repeated enospc. It's easier to reproduce with a config like so:

```ini
raft.data_dir = /path/to/small/tmpfs/mount
# Tune down WAL file size. If there are a few large WAL
# files then recovery might take long enough that it avoids
# hitting max restart intensity.
raft.wal_max_size_bytes = 67108864
```

[enospc-crash-log.txt](https://github.com/user-attachments/files/26120119/enspoc-crash-log.txt)

dumbbell:
Thanks, I will review it next week when I get back from holidays.

deps/rabbit/src/rabbit_ra_systems.erl
```diff
@@ -187,3 +203,34 @@ ensure_ra_system_stopped(RaSystem) ->
...
+stop() ->
+    case supervisor:terminate_child(rabbit_sup, ?MODULE) of
+        ok -> ok = supervisor:delete_child(rabbit_sup, ?MODULE);
+        {error, not_found} -> ok
+    end.
+
+init(_) ->
+    process_flag(trap_exit, true),
+    Pid = whereis(ra_systems_sup),
+    true = link(Pid),
```
dumbbell, line 227: Don’t you need to unlink in the `terminate/2` callback? Otherwise, stopping `rabbit` will kill `ra_systems_sup`. I know that `rabbit` messes with its dependencies so it might not be a problem right now. But if one day, we fix `rabbit`, it will be one.

deps/rabbit/src/rabbit_ra_systems.erl
```diff
@@ -192,3 +209,38 @@ ensure_ra_system_stopped(RaSystem) ->
...
+    {ok, Pid, hibernate}.
+
+handle_call(_Request, _From, State) -> {noreply, State}.
```
dumbbell, line 230: This will leave the caller waiting for a reply, until timeout if it set one. I think you can reply with an `ok` just to avoid that. What do you think?

deps/rabbit/src/rabbit_ra_systems.erl
```diff
@@ -192,3 +209,38 @@ ensure_ra_system_stopped(RaSystem) ->
...
+
+handle_cast(_Msg, State) -> {noreply, State}.
+
+handle_info({'EXIT', RaSystemsSup, Reason} = E, RaSystemsSup) ->
+    ?LOG_ERROR(
+        ?MODULE_STRING ": Ra system supervisor exited with reason ~tp~n",
+        [Reason],
+        #{domain => ?RMQLOG_DOMAIN_GLOBAL}),
+    exit(E);
```
dumbbell, line 239: What about returning `{stop, Reason, State}`? Would it have a different behaviour compared to exiting?

the-mikedavis: Yeah actually, Rabbit doesn't exit unless we use `exit/1` here. With `{stop, Reason, State}` Rabbit keeps going. I'm not really sure why. This is the part that seems pretty hacky to me :sweat_smile:

the-mikedavis:
I think a monitor might be a bit cleaner here but I need to experiment with the exiting again. I don't fully understand why `{stop, Reason, State}` in the monitoring server is not sufficient but exiting is.

I looked at doing something in `ra` directly as well (https://github.com/rabbitmq/ra/issues/585#issuecomment-4653854642) but I don't think we can do it from `ra` alone because of the way that the supervision trees work.

the-mikedavis:
Closing in favor of https://github.com/rabbitmq/rabbitmq-server/pull/17156 - I think that's the more natural way to handle this

PR #15794, opened 2026-03-23
(selenium) Capture why mock-auth-backend-http fails to start in some jobs in CI

MarcialRosales:
## Proposed Changes

There is a suite which fails to start the mock http container in tanzu. This commit captures the container output.
It also skipping building a mocha-image if not required.

## Types of Changes

- [x] Build system and/or CI

### DIFF
--- a/selenium/README.md
+++ b/selenium/README.md
@@ -220,6 +220,15 @@ following command:
 MOCHA_DOCKER_FILE=\location\of\my\Dockerfile ./run-suites.sh
 ```
 
+When you run the tests locally, you need to run them with the environment variable 
+`MOCHA_DOCKER_FILE=Dockerfile.local` so that the scripts that build the docker image
+`mocha-test` which internally downloads npm packages. To download these packages, 
+you need to be in VPN via a gateway and have .npmrc file in the selenium folder 
+with a api key downloaded from Artifactory. There is an internal confluence page 
+that explains how do it.
+
+Without this docker file and .npmrc file, it is impossible to build mocha-test image.
+
 ## Issues deploying UAA 
 
 If you are not able to successfully run any test case that depends on UAA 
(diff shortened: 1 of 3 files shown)


michaelklishin:
@MarcialRosales is this PR still a draft or just was never marked as ready?

selenium/bin/components/mock-auth-backend-http
```diff
@@ -23,14 +23,22 @@ start_mock-auth-backend-http() {
...
     --publish 8888:1080 \
     --env MOCKSERVER_INITIALIZATION_JSON_PATH="/config/defaultExpectations.json" \
     -v ${AUTH_BACKEND_HTTP_DIR}:/config \
-    ${MOCKSERVER_DOCKER_IMAGE}
+    ${MOCKSERVER_DOCKER_IMAGE} 2>&1 | tee -a "${logfile}"
+
+  if [[ ${PIPESTATUS[0]} -ne 0 ]]; then
+    print "docker run failed; see ${logfile}"
+    return 1
+  fi
```
Zerpet, line 41: I'm not sure this change will yield the result you expect. Since you are running `docker run` with `--detach` flag, the output of `docker run` will be a container SHA, but it won't capture the container logs.

If you want to capture the container logs on a failure, I would recommend to run `docker logs <container-name>` when the container fails to start.

An idea to determine "did mocha-server start?" is to poll the exposed HTTP port in 8888, and ensure it passes consistently for a number of seconds.

MarcialRosales: I use `docker logs <container-name>` but if the container failed, `docker logs ` is not giving me nothing.. I am already using the technique you suggest. see ` wait_for_url $AUTH_BACKEND_HTTP_BASEURL/ready` .. 
The purpose of this PR was to actually find out why in CI, sometimes, the container fails to start ..

Zerpet: That makes sense. You will only capture anything meaningful if `docker run` itself fails. If that's the intended outcome, I'm good with this PR.

PR #15819, opened 2026-03-24
Add rabbit_exchange_type:recover/2 callback

the-mikedavis:
This callback is similar to the rabbit_queue_type version. It's meant to be a hook for exchanges like the consistent hash exchange to recover, initialize state, etc. during node boot-up. Previously the consistent hash exchange extra state was recovered with a boot step. The new callback is optional for the sake of backwards compatibility.

### DIFF
--- a/deps/rabbitmq_consistent_hash_exchange/src/rabbit_exchange_type_consistent_hash.erl
+++ b/deps/rabbitmq_consistent_hash_exchange/src/rabbit_exchange_type_consistent_hash.erl
@@ -16,8 +16,8 @@
 -export([description/0, serialise_events/0, route/3]).
 -export([validate/1, validate_binding/2,
          create/2, delete/2, policy_changed/2,
-         add_binding/3, remove_bindings/3, assert_args_equivalence/2]).
--export([init/0]).
+         add_binding/3, remove_bindings/3, assert_args_equivalence/2,
+         recover/2]).
 -export([info/1, info/2]).
 -export([ring_state/2]).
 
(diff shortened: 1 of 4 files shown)


michaelklishin (approved):

deps/rabbitmq_consistent_hash_exchange/src/rabbit_exchange_type_consistent_hash.erl
```diff
@@ -116,18 +105,10 @@ maybe_initialise_hash_ring_state(#exchange{name = Name}) ->
...
-    %% starting with RabbitMQ 3.8.4
-    Xs = list_exchanges(),
-    ?LOG_DEBUG("Consistent hashing exchange: have ~b durable exchanges to recover", [length(Xs)]),
+recover(_VHost, Xs) ->
     %% TODO: reset storage if this is the first node in the cluster
-    [recover_exchange_and_bindings(X) || X <- lists:usort(Xs)].
-
-list_exchanges() ->
-    Pattern = #exchange{durable = true, type = 'x-consistent-hash', _ = '_'},
-    rabbit_db_exchange:match(Pattern).
```
the-mikedavis, line 130: There is a slight behavior change here since `rabbit_db_exchange:recover/1` ignores transient exchanges and recovers all of them. I think the logic here to only recover durable ones may have been a bug before this change? (At least with Khepri where transient exchanges are not really a thing.)

michaelklishin: In the Khepri age we can assume that everything that's in the data store is durable.

michaelklishin:
Given that the `x-modulus-hash` exchange was moved to `deps/rabbit` and some relevant tests were reworked, I find it easier to reimplement this branch than doing a rebase.

I'll submit a new PR for @the-mikedavis to review.

PR #15835, opened 2026-03-25
Clarify 'reclaim_quorum_memory' command description and banner

the-mikedavis:
The aux effect triggered by `rabbit_quorum_queue:reclaim_memory/2` no longer triggers a WAL rollover, just a `erlang:garbage_collect/0` run. This change updates the description and banner of the `reclaim_quorum_memory` CLI command to remove mentions of WAL rollover.


the-mikedavis:
Looks like this has been true since back before 4.2 so we could backport this as far as 4.2.x: https://github.com/rabbitmq/rabbitmq-server/blob/v4.2.x/deps/rabbit/src/rabbit_fifo.erl#L1162-L1163

PR #15860, opened 2026-03-27
Add config for stream initial cluster size, default to 3

the-mikedavis:
This matches the `quorum_queues.initial_cluster_size` config option which also defaults to three. Previously streams defaulted to `length(rabbit_nodes:list_members())` - starting a replica on all members of the cluster:

```
% ENABLED_PLUGINS=rabbitmq_stream NODES=5 make start-cluster
% stream-perf-test --producers 1 --consumers 0 --time 1
% sbin/rabbitmq-streams stream_status -n rabbit-1 stream
Status of stream stream on node rabbit-1@mango2 ...
┌─────────┬─────────────────┬───────┬─────────┬──────────────────┬────────────────────┬──────────────┬─────────┬──────────┐
│ role    │ node            │ epoch │ offset  │ committed_offset │ committed_chunk_id │ first_offset │ readers │ segments │
├─────────┼─────────────────┼───────┼─────────┼──────────────────┼────────────────────┼──────────────┼─────────┼──────────┤
│ writer  │ rabbit-1@mango2 │ 1     │ 1087375 │ 1087375          │ 1086181            │ 0            │ 4       │ 1        │
├─────────┼─────────────────┼───────┼─────────┼──────────────────┼────────────────────┼──────────────┼─────────┼──────────┤
│ replica │ rabbit-2@mango2 │ 1     │ 1087375 │ 1087375          │ 1086181            │ 0            │ 0       │ 1        │
├─────────┼─────────────────┼───────┼─────────┼──────────────────┼────────────────────┼──────────────┼─────────┼──────────┤
│ replica │ rabbit-3@mango2 │ 1     │ 1087375 │ 1087375          │ 1086181            │ 0            │ 0       │ 1        │
├─────────┼─────────────────┼───────┼─────────┼──────────────────┼────────────────────┼──────────────┼─────────┼──────────┤
│ replica │ rabbit-4@mango2 │ 1     │ 1087375 │ 1087375          │ 1086181            │ 0            │ 0       │ 1        │
├─────────┼─────────────────┼───────┼─────────┼──────────────────┼────────────────────┼──────────────┼─────────┼──────────┤
│ replica │ rabbit-5@mango2 │ 1     │ 1087375 │ 1087375          │ 1086181            │ 0            │ 0       │ 1        │
└─────────┴─────────────────┴───────┴─────────┴──────────────────┴────────────────────┴──────────────┴─────────┴──────────┘
```

I think it makes more sense to default to a subset of 3 nodes like quorum queues.

### DIFF
--- a/deps/rabbit/docs/rabbitmq.conf.example
+++ b/deps/rabbit/docs/rabbitmq.conf.example
@@ -481,6 +481,11 @@
 ## at declaration time.
 # quorum_queue.initial_cluster_size = 3
 
+## Sets the initial stream queue replica count for newly declared stream queues.
+## This value can be overridden using the 'x-initial-cluster-size' queue argument
+## at declaration time, or by setting 'initial-cluster-size' in a policy.
+# stream.initial_cluster_size = 3
+
 ## Sets the maximum number of unconfirmed messages a channel can send
 ## before publisher flow control is triggered.
 ## The current default is configured to provide good performance and stability
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -2912,6 +2912,11 @@ end}.
     {validators, ["is_supported_information_unit"]}
 ]}.
 
+{mapping, "stream.initial_cluster_size", "rabbit.stream_cluster_size", [
+  {datatype, integer},
+  {validators, ["non_zero_positive_integer"]}
+]}.
+
 {mapping, "cluster_tags.$tag", "rabbit.cluster_tags", [
     {datatype, [binary]}
 ]}.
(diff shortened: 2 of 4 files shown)


the-mikedavis:
I wasn't sure if the old behavior of adding a replica on all nodes was intentional or not. We could remove the default value of 3 here and add replicas on all nodes when the config option is not set in order to avoid the behavior change.

michaelklishin (approved):
FTR, the current implementation of `rabbit_stream_queue:initial_cluster_size/1` has been like this since Oct 2020 (commit 53b78579, https://github.com/rabbitmq/rabbitmq-server/pull/2484) and then it was adapted to `rabbit_nodes` changes in Jan 2023 in d6563719.

It does not seem to have been intentional: quorum queues had the same default back in Oct 2020.

But then QQs have switched to 3 replicas by default and streams did not.

kjnilsson:
It would have been a good idea to discuss this one with the original designers of streams. That streams span all nodes by default is a deliberate design decision. Streams need to span all nodes as a default as otherwise you cannot consume from streams on nodes that do not have a member. Hence the sensible default is all nodes.

Please revert this PR.

PR #15872, opened 2026-03-28
Revert "Split make config into per-plugin files templated when plugins enabled (backport #15855) (backport #15856)"

michaelklishin:
Reverts rabbitmq/rabbitmq-server#15857.

With it, Tanzu RabbitMQ nodes cannot start with enabled plugins no matter what I try.

### DIFF
--- a/deps/rabbit_common/mk/rabbitmq-run.mk
+++ b/deps/rabbit_common/mk/rabbitmq-run.mk
@@ -166,113 +166,43 @@ endif
 # Run a full RabbitMQ.
 # --------------------------------------------------------------------
 
-# Shell function that returns true when a plugin is enabled.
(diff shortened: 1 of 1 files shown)


michaelklishin:
@MirahImage @mkuratczyk my digging arrived at the `peer` module migration commits that were not backported to `v4.2.x`.

With this change

```
gmake run-broker ENABLED_PLUGINS="rabbitmq_management rabbitmq_shovel rabbitmq_shovel_management rabbitmq_federation rabbitmq_federation_management rabbitmq_stream rabbitmq_stream_management"
```

would start without any plugins enabled (`0`) in the logs.

I will revert this change as non-essential as it blocks a Tanzu RabbitMQ upstream merge.

PR #15883, opened 2026-03-29
Handle delivery-count mismatch between session and QQ

ansd:
Prior to this commit, the following issue could happen: Delivery effects and credit_reply effect and applied notifications get lost from the QQ leader to the session on a non-member node got, e.g. due to network issues. However, the credit command got applied in the machine successfully nevertheless.

When the network recovers, rabbit_fifo_client receives a leader_change notification and re-sends the credit command. The QQ applies the 2nd credit command again (idempotently), but given that the QQ already assigned the messages to the consumer and there are no further messages in the ready queue, the QQ will reply only with a credit_reply effect. This credit_reply now has a delivery-count that **leads** the delivery-count of the session proc.

This commit takes the following fix:
The assertion that the delivery-count from the queue must be in sync with the delivery-count in the session is removed. Instead the session sets the delivery-count from the queue. This is inline with the AMQP spec which writes:
"The delivery-count is initialized by the sender when a link endpoint is created, and is incremented whenever a message is sent. Only the sender MAY independently modify this field. The receiver's value is calculated based on the last known value from the sender" [2.6.7]

The consequence of this fix is that it can happen that the session grants subsequently too many credits to the sending queue. The spec writes for a different use case the following:
"If the link-credit is reduced by the receiver when transfers are in-flight, the receiver MAY either handle the excess messages normally or detach the link with a transfer-limit-exceeded error code." So, the client should either handle the excess messages as well or detach the link. Both is less worse than the previous crash in rabbit_amqp_session.

In the newly added test case, if m3 is omitted from being sent, the client won't receive m1 and m2. That's because rabbit_fifo_client will only get missing deliveries (m1 and m2), if it receives a new out of order delivery (m3). So, rabbit_fifo_client will only get missing deliveries eventually. Maybe a better fix would be that the QQ includes in its leader_change notification also the last sent sequence number for the consumers. This way rabbit_fifo_client can check itself whether it missed any deliveries, and then pro-actively gets missing deliveries when handling the leader_change notification rather than waiting for the next out of order deliveries to arrive.

This commit also fixes another bug where the missing deliveries are fetched from the QQ out of order.
https://github.com/rabbitmq/rabbitmq-server/pull/14386 attempted to fix this, but unfortunately the implementation in that PR was wrong.

The following new test case reproduces the `delivery_count_mismatch` crash in rabbit_amqp_session (without the fixes in this commit applied):
```
make -C deps/rabbit ct-amqp_client t=cluster_size_3:lost_deliveries_quorum_queue FULL=1
```

### DIFF
--- a/deps/rabbit/src/rabbit_amqp_session.erl
+++ b/deps/rabbit/src/rabbit_amqp_session.erl
@@ -1715,9 +1715,7 @@ handle_credit_reply0(#credit_reply{ctag = Ctag,
                                              echo = CEcho,
                                              properties = OldProps
                                             } = CFC0,
-                        queue_flow_ctl = #queue_flow_ctl{
-                                            delivery_count = QDeliveryCount
-                                           } = QFC0,
+                        queue_flow_ctl = QFC0,
                         stashed_credit_req = StashedCreditReq
                        } = Link0,
                      #state{cfg = #cfg{writer_pid = Writer,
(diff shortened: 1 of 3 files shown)


mkuratczyk:
I tested this with the same workload that triggered the issue before and I cannot reproduce it. It was pretty hard to trigger so I can't be certain but I think this PR fixes the issue.

PR #15886, opened 2026-03-30
Add a metric for blocked at-least-once dead-lettering QQs

the-mikedavis:
This is a quick stab at continuing https://github.com/rabbitmq/rabbitmq-server/discussions/14494. We can expose a metric for the `map_size/1` of `#state.logged` in the QQ DLX worker. Then you can gather and report on that metric to quickly find QQs with messages stuck in the internal dead letter queue.


the-mikedavis:
Alternatively we could add some counters for the DLX worker in general. We could expose a gauge for the number of pending messages, have counters for the number of successful deliveries, etc..

PR #15889, opened 2026-03-30
Trust store: log some details of rejected certificates

michaelklishin:
Otherwise human cluster operators wouldn't have much to work with when investigating rejected client
connections (what certificates were rejected and why).

References #15854.

### DIFF
--- a/deps/rabbitmq_trust_store/src/rabbit_trust_store.erl
+++ b/deps/rabbitmq_trust_store/src/rabbit_trust_store.erl
@@ -121,14 +121,16 @@ whitelisted(#'OTPCertificate'{}=C, {bad_cert, unknown_ca}, continue) ->
         true ->
             {valid, confirmed};
         false ->
-            {fail, "CA not known AND certificate not whitelisted"}
+            log_certificate_rejected("CA is not known and certificate is not whitelisted", C),
+            {fail, "CA is not known and certificate is not whitelisted"}
     end;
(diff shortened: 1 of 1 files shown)


deps/rabbitmq_trust_store/src/rabbit_trust_store.erl
```diff
@@ -370,3 +372,28 @@ clean_deleted_providers(Providers) ->
...
+    try
+        DER = public_key:pkix_encode('OTPCertificate', C, otp),
+        Fingerprint = format_fingerprint(crypto:hash(sha256, DER)),
+        ?LOG_WARNING("Trust store: ~ts "
+                     "(fingerprint: ~ts, "
+                     "issuer: ~ts, "
+                     "validity: ~ts)",
+                     [Reason,
+                      Fingerprint,
+                      rabbit_cert_info:issuer(DER),
```
mgrafl, line 386: Thanks a lot for adding the log entry. 
This will work nicely if the certificate has a reasonable CA (case 1). But for self-signed certificates (case 2), the issuer might be bogus. Would the certificate "subject" be useful here or does that go against RabbitMQ's policy about logging such details?

michaelklishin: @mgrafl I have assumed that the fingerprint is the best possible identifier and removed the subject because the logged lines ended up being very long.

PR #15913, opened 2026-04-01
Upgrade Ra to 3.1.3 and Osiris to 1.13.1

kjnilsson:
The only notable change is the gen_batch_server upgrade to 0.10.0


michaelklishin (changes requested):
This does not update `release-notes/4.3.0.md` which lists what dependencies were changed (for our own needs, if anything).

PR #16039, opened 2026-04-13
Set acquired-count across QQ dead-lettering

ansd:
## What?

Similar to passing on the delivery-count across dead-lettering from a QQ, also pass on the acquired-count.

 ## Why?

Arguably, it would be unexpected for a client to receive a message with a delivery-count (number of **failed** delivery attempts) to be higher than the acquired-count (number of all - both failed and non-failed - delivery attempts).

Do we want this change at all?


kjnilsson:
I think we should do this.

kjnilsson (approved):

PR #16046, opened 2026-04-13
Multiple channel interceptors

Ayanda-D:
## Proposed Changes

Hi folks 👋

The following changes seek to add support for multiple channel interceptors within the broker. Currently, when developing channel interceptor type plugins, development is restricted to single interceptor per AMQP operation, per installation. When use-cases grow and need for multiple interceptors for the same AMQP operation arises, this becomes a problem (and combining multiple use-cases within the same channel interceptor module makes things even more complex).

In these changes, we propose supporting multiple channel interceptors for the same AMQP operation, based on configurable priority. Interceptors with the same priority level for the same AMQP operation will not be allowed, as per current behaviour in channels. For example, if 3 interceptors need to intercept `queue.declare`, these would be configured as following in **rabbitmq.conf**:

```
channel_interceptor.priority.example_interceptor_1 = 1
channel_interceptor.priority.example_interceptor_2 = 2
channel_interceptor.priority.example_interceptor_3 = 3
```

These are executed in priority order (starting with the lowest) in the channel. If 2 interceptors for same AMQP operation have the same priority level, registry will fails as current RabbitMQ channel interceptor behaviour. The following wouldnt work if both intercept the same AMQP operation (also priorities default to `0` if not configured, same restriction applies if multiple interceptors are enabled without any config in **rabbitmq.conf**):

```
channel_interceptor.priority.example_interceptor_1  = 0
channel_interceptor.priority.example_interceptor_2 = 0
```

In addition, these changes also introduce a new `rabbitmqctl list_channel_interceptors` command to get visibility on active interceptors (as number of interceptors grow, we'll need this to keep track of whats active and execution order from the priority levels)

```
$> rabbitmqctl list_channel_interceptors
Listing channel interceptors ...
  name                    applies_to                    priority
  dummy_interceptor_1     basic.publish, queue.declare  0
  dummy_interceptor_2     basic.publish, queue.declare  1
```

**NOTE**: I've moved `dummy_interceptor` module from `rabbit/test/` to `rabbit/src/` to make available for the cli (elixir) tests - we could optionally keep it in the test DIR and do dynamic code loading over RPC of compiled beam/object code - but it feels like too much complexity for running simple cli tests. Placing in `rabbit/src/` much simpler.

Please take a look, we are currently stuck and restricted on only having a single channel-interceptors, whilst uses-cases grow. This change would significantly help and improve extensibility of the broker for plugin developers (and also mean less need to propose changes to the broker core, which is always the goal 😇) - we'd appreciate this a lot. Thanks!

## Types of Changes

- [ ] Build system and/or CI

## Checklist


michaelklishin:
@Ayanda-D thanks but for the umpteenth time, the core team suggests that you start a discussion first to gauge the interest in a certain change instead of showing up straight with a PR.

michaelklishin:
> This change would significantly help and improve extensibility of the broker for plugin developers

Sorry but I'm a little bit less enthusiastic about the impact of this change.

This is a lot of complexity for a change that 99% of RabbitMQ users will not need, and it was not previously discussed.

Ayanda-D:
@michaelklishin sorry for not starting off with discussion - the change seemed straightforward and not too complex, as it's mainly altering the restriction `rabbit_channel_interceptor:check_no_overlap/1` currently imposes on conflicting channel interceptors and making it priority based. The actual AMQP interception code `rabbit_channel_interceptor:intercept_in/3` remains unchanged (this already supports looping through all registered interceptor in `rabbit_registry`). The rest of the changes are more config management, cli tools for maintenance, and tests. Its only initialization checks that we're relaxing a bit. (having this in the core also saves us the continual custom patching per release in future to enable these features - which can be a chicken & egg situation on whats merged upstream and what we backport on our end).

michaelklishin:
@Ayanda-D my argument that this is a lot of complexity around the hot code path, and that 99% of RabbitMQ users will never need this feature, still stands.

What can justify a feature with such an unfavourable complexity/reward ratio?

Ayanda-D:
@michaelklishin I guess the justification is - we still find ourselves stuck with the problem of "the broker only allows 1 channel interceptor per AMQP operation", despite all the major advancements in its capabilities over the years. To date, we still can't intercept AMQP operations like `queue.declare` in 2 or more different plugins. Channel interceptors are one of the "cool" features that give us the (community) flexibility and capabilities to stretch the broker beyond feature limits. And we dont want to have this limitation.

In legacy plugins, this was a known problem. For example, the broker doesnt/wouldnt allow enabling the plugins listed below because they both intercept [basic.publish](https://github.com/rabbitmq/rabbitmq-routing-node-stamp/blob/main/src/rabbit_routing_node_stamp_interceptor.erl#L46) - which would and still does [crash](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_channel_interceptor.erl#L55) from the conflicting interceptors.

- https://github.com/rabbitmq/rabbitmq-routing-node-stamp 
- https://github.com/rabbitmq/rabbitmq-message-timestamp

I also recall facing the same problem in a custom plugin which I had to build once upon a time (for some CQ use-case) - https://github.com/Ayanda-D/rabbitmq-message-persister#limitations 

With this change, we will no longer have this limitation. We will now be able to configure the broker to accept more than 1 interceptor for the same AMQP primitive, like `basic.publish`, by simply configuring the order/priority in which they are executed in the config file. And the channel will execute them in a loop.

For example purposes (I'm aware some of these are deprecated), for these 3 plugins which I've mentioned^^^ which all intercept `basic.publish` - we would be able to use all 3 by setting priority or order of execution (lowest being executed first).

```
channel_interceptor.priority.rabbit_timestamp_interceptor = 1
channel_interceptor.priority.rabbit_routing_node_stamp_interceptor = 2
channel_interceptor.priority.rabbit_message_persist_interceptor = 3 
```

Using all 3 is currently not possible and the broker will [crash](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_channel_interceptor.erl#L55) when more than 1 of these^ plugins are enabled.

AMQP is very broad and rich - we really feel channel interceptors should likewise be broad enough, capable and flexible enough to also support the vast number of AMQP operations.

(**NOTE**: This is also an opt-in feature, only applicable when used/configured. Hot code path remains the same, & when configured, will execute each interceptor in one after the other).

We'd really appreciate having this available in future releases (& avoid custom patching releases to have this 😊).

michaelklishin:
Yes, it's an opt-in feature. The performance hit is I'm sure equivalent to statistical noise (I don't see any numbers shared, though).

I'm afraid it's the core team who will be on the hook maintaining this and any other contributed feature, forever, and if it does not work out, removing it will be a couple of orders of magnitude harder than adding it was.

michaelklishin:
## Benchmarks

### Workloads

To get one measureable counterargument, I've tested several workloads:

1. Perftest: 1 publisher without publisher confirms 1 consumer in automatic ack mode, three no-op interceptors on `basic.publish`, `queue.declare`, (consumer-sent) `basic.ack`
2. Perftest: 1 publisher publisher confirms and a batch size of 250, 1 consumer in manual ack mode with, three no-op interceptors on `basic.publish`, `queue.declare`, (consumer-sent) `basic.ack`
3. Java client (custom, not via Perftest) with heavy channel churn
4. Rust client (custom, bunny-rs) with heavy classic queue churn (many `queue.declare`s and `queue.delete`s) and interceptors declared in a way that would involve some actual sorting work (`[2, 3, 1]`, not `[1, 2, 3]`)
5. A variant of 1 with two no-op interceptors and two real (node stamp, routing timestamp) interceptors
6. A variant of 2 with two no-op interceptors and two real (node stamp, routing timestamp) interceptors

When I say "no-op" I mean this:

```erl
-module(bench_server_16046).
-behaviour(rabbit_channel_interceptor).
-export([init/1, description/0, intercept/3, applies_to/0]).

init(_Ch)           -> undefined.
description()       -> [{description, <<"noop bp">>}].
%% return unchanged values
intercept(M, C, _S) -> {M, C}.
applies_to()        -> ['basic.publish'].
```

Classic queues were used to avoid the Raft cluster formation and replication costs of QQs,
something that would hide the true cost of any additional channel process work.

The interceptors in scenarios 1 and 2 we no-op to see the overhead

### Per-Message Cost (with no-op `basic.publish` Interceptors) with Publisher Confirms

The throughput effects are close to zero, significantly under 1%.

The list of interceptors coming from a channel field holds very well. Likely thanks to the JIT but I wasn't curious enough to begin comparing flame graphs.

### Per-Message Cost (`basic.publish` Interceptors) without Publisher Confirms

In this workload the channel processes are supposed to be pushed to their limits by a fire-and-forget publisher. However, that's not what I observed: the median was very comparable at 203K messages/second but with more variance. That variance can potentially come from the PerfTest JVM GC pauses.

### Channel Churn (`channel.open` Interceptors)

The third one with heavy channel churn, the extra cost of opening a channel seems to be 10-12 microseconds. And it can be reduced further by avoiding a `application:get_env/3` call in `priority/1`.

### Classic Queue Churn (`queue.declare`, `queue.delete` Interceptors)

The fourth one suggests that the drop with three interceptors is about 2%, and this is on a workload with extreme churn of over 60K+ queues declared and deleted a second.
The kind of churn we would never recommend as a baseline scenario.

### Per-Message Cost (with the node stamp, timestamp `basic.publish` Interceptors) with Publisher Confirms

Here I observe a 10% drop in throughput. It has to do with the interceptors on this path
doing some actual work and mutating the message.

But at least this workload confirms that the benchmarks were not broken or so naive that they'd catch a regression.


## Results

### Per-Message Cost

The is no cost per message published if channel interceptors are not configured, and
it's effectively zero if a few `basic.publish` interceptors are but publisher confirms serve as a natural publisher flow control. 

A much more surprising outcome is a sub-1% throughput drop with three no-op `basic.publish` interceptors and the fastest publisher

### Channel Churn

The cost of 10-12 microseconds per channel start (`init/1`) and a 2% drop with extreme classic queue churn with a very fast client.

### Classic Queue Churn

A drop of 2% on a very extreme, unrealistic (except for short bursts) workload.

## Conclusion on My Hot Path Argument Above

So my "on the hot path" argument was not completely correct: the hot path is touched but the true cost comes down to what the interceptor actually does. No-op interceptors
do not incur a throughput drop.

Which makes the case for this PR much more interesting.

michaelklishin:
@Ayanda-D now I'm trying to reach consensus around this PR. A lot of people are on holidays, so this can take a while.

michaelklishin:
@Ayanda-D the rest of the team would like to understand what kind of plugins or features would require intercepting a `queue.declare` frame in three different places.

We understand that there are probably limits to what you can say publicly. You know where to find the core team.

Ayanda-D:
@michaelklishin regarding your feedback:

Thanks for doing some rigorous testing. Yes the main "hot" `rabbit_channel_interceptor:intercept_in/3` path (which is executed in `rabbit_channel` for every inbound AMQP method) hasn't been touched in this PR. So all messaging traffic should still be the same (the same cost remains the same - enabling more plugins, and as you also mention, "what the interceptor actually does"). Hence we didn't consider this a concern at all. 

The main work is done on channel initialization, in `rabbit_channel_interceptor:init/1` on `channel.open` - so only extremely high channel churn cases (as your test results also show this), as there extra "small" work done by the plugin. Likewise, we also didn't consider this a concern as its only part of initialization.

Ayanda-D:
@michaelklishin regarding the request from the team:

> @Ayanda-D the rest of the team would like to understand what kind of plugins or features would require intercepting a queue.declare frame in three different places.

Unfortunately no, we cannot disclose what we are doing at this point on any platforms (without of course the necessary permission, paperwork being signed, etc).

Nonetheless, I don't think we need to go to such extents of confidentiality. 

Consider the [rabbitmq-sharding](https://github.com/rabbitmq/rabbitmq-server/tree/main/deps/rabbitmq_sharding) plugin. This is a core plugin which has an interceptor `rabbit_sharding_interceptor` which intercepts the following AMQP operations:

```erlang
applies_to() ->
    ['basic.consume', 'basic.get', 'queue.delete', 'queue.declare',
     'queue.bind', 'queue.unbind', 'queue.purge'].

```

So by simply enabling the sharding plugin:

```
rabbitmq-plugins enable rabbitmq_sharding
```

No other interceptor plugin can be used on the broker to intercept all those seven AMQP methods:
`'queue.declare', ....` 

Without this feature, RabbitMQ users cannot use the core **rabbitmq-sharding** plugin along with any other custom interceptor plugin for these seven AMQP methods.

And to confirm, I just tested (Elixir client) - enabling the **rabbitmq-sharding** and enabling anything else intercepting e.g. `queue.declare` crashes all `channel.open` requests (as expected).

```
iex(8)> {:ok, conn} = AMQP.Connection.open
{:ok, %AMQP.Connection{pid: #PID<0.355.0>}}
iex(9)> AMQP.Channel.open conn
11:40:57.319 [error] GenServer #PID<0.355.0> terminating
** (stop) :socket_closed_unexpectedly
Last message: :socket_closed
State: {:state, :amqp_network_connection, {:state, #Port<0.12>, "client 127.0.0.1:63682 -> 127.0.0.1:5672", 10, #PID<0.358.0>, 131072, #PID<0.354.0>, :undefined, false}, #PID<0.357.0>, {:amqp_params_network, "guest", {:plaintext, "guest"}, "/", ~c"localhost", 5672, 0, 0, 10, 50000, :none, [&:amqp_auth_mechanisms.plain/3, &:amqp_auth_mechanisms.amqplain/3], [], []}, 2047, [{"capabilities", :table, [{"publisher_confirms", :bool, true}, {"exchange_exchange_bindings", :bool, true}, {"basic.nack", :bool, true}, {"consumer_cancel_notify", :bool, true}, {"connection.blocked", :bool, true}, {"consumer_priorities", :bool, true}, {"authentication_failure_close", :bool, true}, {"per_consumer_qos", :bool, true}, {"direct_reply_to", :bool, true}]}, {"cluster_name", :longstr, "rabbit@ayandahost"}, {"copyright", :longstr, "Copyright (c) 2007-2026 Broadcom Inc and/or its subsidiaries"}, {"information", :longstr, "Licensed under the MPL 2.0. Website: https://rabbitmq.com"}, {"platform", :longstr, "Erlang/OTP 28.4.1"}, {"product", :longstr, "RabbitMQ"}, {"version", :longstr, "4.2.3"}], :none, false}
** (exit) exited in: :gen_server.call(#PID<0.364.0>, :open, 70000)
    ** (EXIT) shutdown
    (stdlib 5.2.3.3) gen_server.erl:419: :gen_server.call/3
    (amqp_client 4.0.3) amqp_gen_connection.erl:84: :amqp_gen_connection.open_channel/3
    (amqp 4.1.0) lib/amqp/channel.ex:73: AMQP.Channel.do_open_channel/2
    iex:9: (file)

```

So just using the core **rabbitmq-sharding** plugin completely blocks off any other use of any channel interceptor plugin intercepting any of:

```
['basic.consume', 'basic.get', 'queue.delete', 'queue.declare', 'queue.bind', 'queue.unbind', 'queue.purge']

```

deps/rabbit/src/rabbit_channel_interceptor.erl
```diff
@@ -37,29 +37,60 @@ added_to_rabbit_registry(_Type, _ModuleName) ->
...
+-spec set_priorities([{module(), integer()}]) -> ok.
+%% Merge the given interceptor priorities into the current configuration,
+%% keeping priorities of interceptors that are not mentioned, and refresh
+%% all channels so the new ordering takes effect.
+set_priorities(NewPriorities) ->
+    Current = application:get_env(rabbit, channel_interceptor_priorities, []),
+    Merged = lists:foldl(fun({Mod, P}, Acc) ->
+                             lists:keystore(Mod, 1, Acc, {Mod, P})
+                         end, Current, NewPriorities),
+    ok = application:set_env(rabbit, channel_interceptor_priorities, Merged),
```
lukebakken: The merged list is committed before anything validates it, and `rabbit_channel:refresh_interceptors/0` logs per-channel failures and returns `ok` regardless. So if the merge puts two interceptors that handle the same operation at the same priority, `init/1` raises in every channel process on this node, all of them die with their connections, every subsequent `channel.open` fails the same way (`rabbit_channel.erl:482`), and the command still exits 0 because `DefaultOutput` maps `:ok` to success.

Running the overlap check against the merged list before `application:set_env/3` and returning an error would keep that from happening. Also worth saying in `description/0` that the change does not survive a restart.

deps/rabbit/src/rabbit_channel_interceptor.erl
```diff
@@ -37,29 +37,60 @@ added_to_rabbit_registry(_Type, _ModuleName) ->
...
+                    sets:new(),
+                    Sets),
     ok.
 
+priority(Mod) ->
+    Priorities = application:get_env(rabbit, channel_interceptor_priorities, []),
+    case lists:keyfind(Mod, 1, Priorities) of
+        {Mod, P} -> P;
+        false     -> 0
+    end.
```
lukebakken, line 172: Avoiding the `application:get_env/3` in `priority/1` was already raised in the benchmark comment above. There is a correctness reason for it as well: one `init/1` reads the env once per sort comparison and again once per module in `check_no_overlap/1`, so a concurrent `set_priorities/1` can leave a single call working from two versions of the configuration. Two interceptors that share a priority in the committed list can then be read as having different ones and pass the check.

Reading it once fixes both and gives `set_priorities/1` something to validate against:

```suggestion
init(Ch) ->
    Mods = [M || {_, M} <- rabbit_registry:lookup_all(channel_interceptor)],
    Priorities = application:get_env(rabbit, channel_interceptor_priorities, []),
    Sorted = lists:sort(fun(A, B) -> priority(A, Priorities) =< priority(B, Priorities) end,
                        Mods),
    check_no_overlap(Sorted, Priorities),
    [{Mod, Mod:init(Ch)} || Mod <- Sorted].

%% Reject any two interceptors that share the same priority and handle the same
%% AMQP operation. Interceptors with different priorities may overlap freely.
check_no_overlap(Mods, Priorities) ->
    ByPriority = lists:foldl(fun(Mod, Acc) ->
                                 P = priority(Mod, Priorities),
                                 maps:update_with(P, fun(Ms) -> [Mod | Ms] end, [Mod], Acc)
                             end, #{}, Mods),
    maps:foreach(fun(_Priority, Group) ->
                     check_no_overlap1([sets:from_list(Mod:applies_to()) || Mod <- Group])
                 end, ByPriority).

check_no_overlap1(Sets) ->
    _ = lists:foldl(fun(Set, Union) ->
                        Is = sets:intersection(Set, Union),
                        case sets:size(Is) of
                            0 -> ok;
                            _ ->
                                internal_error("Interceptor: more than one module handles ~tp", [Is])
                        end,
                        sets:union(Set, Union)
                    end,
                    sets:new(),
                    Sets),
    ok.

priority(Mod) ->
    priority(Mod, application:get_env(rabbit, channel_interceptor_priorities, [])).

priority(Mod, Priorities) ->
    case lists:keyfind(Mod, 1, Priorities) of
        {Mod, P} -> P;
        false -> 0
    end.
```

deps/rabbit/test/dummy_interceptor.erl
```diff
@@ -6,7 +6,7 @@
 -include_lib("rabbit_common/include/rabbit_framing.hrl").
 
 
--compile(export_all).
+-export([init/1, description/0, intercept/3, applies_to/0]).
```
lukebakken, line 9: This puts the module in `rabbit.app`'s module list on every installation (it is there in `deps/rabbit/ebin/rabbit.app` after a build), where `intercept/3` empties the payload of every `basic.publish` and any plugin or console user can register it.

Since 78bf0f6ede moved it while clearing dialyzer warnings, is there a way to get that without shipping it? The sharding plugin you cite above registers `rabbit_sharding_interceptor` from a boot step, so adding `rabbitmq_sharding` to `ENABLED_PLUGINS` in `deps/rabbitmq_cli/Makefile` would give the CLI tests a real interceptor to list and `dummy_interceptor` could stay in `test/`. The `assert result == []` case in `list_channel_interceptors_command_test.exs` would need to change with it.

deps/rabbit/src/rabbit_channel_interceptor.erl
```diff
@@ -37,29 +37,60 @@ added_to_rabbit_registry(_Type, _ModuleName) ->
 removed_from_rabbit_registry(_Type) ->
     rabbit_channel:refresh_interceptors().
 
+list() ->
+    Mods = [M || {_, M} <- rabbit_registry:lookup_all(channel_interceptor)],
+    [[{name, Mod}, {applies_to, Mod:applies_to()}, {priority, priority(Mod)}] || Mod <- Mods].
```
lukebakken: `call_module/4` guards `undef` because "Mod might be unloaded at any point". `list/0` calls `Mod:applies_to()` unguarded, so `list_channel_interceptors` racing a plugin being disabled returns a bare `{badrpc, {'EXIT', {undef, ...}}}`.

deps/rabbit/src/rabbit_channel_interceptor.erl
```diff
@@ -37,29 +37,60 @@ added_to_rabbit_registry(_Type, _ModuleName) ->
...
-                      end,
-                    sets:union(Set, Union)
-                end,
-                sets:new(),
-                Sets),
+                        Is = sets:intersection(Set, Union),
+                        case sets:size(Is) of
+                            0 -> ok;
+                            _ ->
+                                internal_error("Interceptor: more than one module handles ~tp", [Is])
```
lukebakken: Now that overlap is scoped per priority, the fix is always "give one of these two a different priority", but the message names neither the modules nor the priority they share.

lukebakken (commented):
> [!NOTE]
> This review was drafted by Claude (Anthropic's Claude Code) under @lukebakken's direction. The findings and code suggestions are Claude's; each was checked against this branch, and the suggested patches were compiled and format-checked.

A pass over the implementation as written, separate from the question of whether the feature should land. The one I would call blocking is `set_priorities/1`: it commits the merged configuration before validating it, so a conflicting priority takes down every channel on the node while `rabbitmqctl` exits 0. The rest are smaller; four have committable suggestions.

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -2705,6 +2705,30 @@ end}.
     end
 }.
 
+%%
+%% Channel interceptors
+%%
+
+%% Priority levels for channel interceptor modules. Lower values run first, default is 0.
```
lukebakken, line 2712: Lower-wins is the opposite of the two priorities operators already know: consumers are served by `priority_queue:highest/1` and `rabbit_policy:priority_comparator/2` is `>=`, so both run higher first. `channel_interceptor.order.$name` would not carry the wrong intuition.

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -2705,6 +2705,30 @@ end}.
...
+{mapping, "channel_interceptor.priority.$name", "rabbit.channel_interceptor_priorities", [
+    {datatype, integer}]}.
+
+{translation, "rabbit.channel_interceptor_priorities",
+ fun(Conf) ->
+         case cuttlefish_variable:filter_by_prefix("channel_interceptor.priority", Conf) of
+             [] ->
+                 cuttlefish:unset();
+             L ->
+                 [{list_to_atom(Name), Priority}
```
lukebakken, line 2726: A name that matches no registered interceptor is turned into an atom and then never looked at again, so a misspelled module keeps the default 0 and the configuration looks applied. The set of valid names is not knowable at parse time, but a warning at boot for entries that match nothing registered would make this findable.

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/set_channel_interceptor_priorities_command.ex
```diff
@@ -0,0 +1,85 @@
...
+  use RabbitMQ.CLI.Core.AcceptsDefaultSwitchesAndTimeout
+  use RabbitMQ.CLI.Core.MergesNoDefaults
+
+  def validate([], _) do
+    {:validation_failure, :not_enough_args}
+  end
+
+  def validate(args, _) when rem(length(args), 2) != 0 do
+    {:error, "Interceptors and priorities must be provided as <interceptor> <priority> pairs"}
+  end
```
lukebakken, line 21: `{:error, msg}` from `validate/2` reaches the `format_error/3` catch-all, which exits 69 (`EX_UNAVAILABLE`, "service unavailable" for a typo) and prints no usage. 33 call sites use `{:validation_failure, {:bad_argument, msg}}`, which exits 65 and prints usage. (`set_log_level` has the same problem, so this is not new here.)

```suggestion
  def validate(args, _) when rem(length(args), 2) != 0 do
    msg = "Interceptors and priorities must be provided as <interceptor> <priority> pairs"
    {:validation_failure, {:bad_argument, msg}}
  end
```

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/set_channel_interceptor_priorities_command.ex
```diff
@@ -0,0 +1,85 @@
...
+
+  def validate(args, _) do
+    priorities = args |> Enum.drop(1) |> Enum.take_every(2)
+
+    case Enum.find(priorities, fn p -> not integer_string?(p) end) do
+      nil ->
+        :ok
+
+      invalid ->
+        {:error, "Priority \"#{invalid}\" is not a valid integer"}
```
lukebakken: Same here.

```suggestion
        {:validation_failure, {:bad_argument, "Priority \"#{invalid}\" is not a valid integer"}}
```

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/set_channel_interceptor_priorities_command.ex
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  use RabbitMQ.CLI.Core.RequiresRabbitAppRunning
+
+  def run(args, %{node: node_name, timeout: timeout}) do
+    priorities =
+      args
+      |> Enum.chunk_every(2)
+      |> Enum.map(fn [interceptor, priority] ->
+        {String.to_atom(interceptor), String.to_integer(priority)}
```
lukebakken, line 43: The name is never checked against anything, so `set_channel_interceptor_priorities dumy_interceptor 7` reports success and leaves the interceptor the operator meant to reorder untouched. `enable_feature_flag` has the same shape and returns `{:error, :unsupported}` for a name the node does not know; `rabbit_registry:lookup_all(channel_interceptor)` would give `set_priorities/1` the same ability.

deps/rabbit/test/channel_interceptor_SUITE.erl
```diff
@@ -303,6 +306,98 @@ conflicting_interceptors_close_direct_connections_gracefully1(Config) ->
...
+                                  dummy_interceptor_priority_1),
+    ok = rabbit_registry:register(channel_interceptor,
+                                  <<"dummy interceptor priority 1 conflict">>,
+                                  dummy_interceptor_priority_1_conflict),
+    try
+        %% Initialising interceptors must fail: two interceptors with the same
+        %% priority handle the same AMQP operation.
+        rabbit_channel_interceptor:init(self())
+    catch
+        exit:{amqp_error, internal_error, _, _} -> ok
```
lukebakken, line 362: Nothing here fails if the exit is not raised: drop the overlap check entirely and this case still returns `passed`.

```suggestion
    try
        %% Initialising interceptors must fail: two interceptors with the same
        %% priority handle the same AMQP operation.
        ?assertExit({amqp_error, internal_error, _, _},
                    rabbit_channel_interceptor:init(self()))
```

deps/rabbit/priv/schema/rabbit.schema
Ayanda-D: WIP, I am still looking into this suggestion.

michaelklishin:
@Ayanda-D thanks, that explains a lot around Bloomberg engineering team communication (or lack of, rather). I understand how things work at companies of that scale.

This now comes down to my [in]ability to convince a few more core team members that this is a worthy and largely de-risked change for plugin combination compatibility.

Assuming that I do get such buy-in, this PR will be polished a bit, something I have already volunteered to do by doing the benchmarking, I guess 😅

michaelklishin:
@lukebakken thank you for the review.

@lukebakken @the-mikedavis have you folks seen any need for this change in your custom plugins?

lukebakken:
> @lukebakken @the-mikedavis have you folks seen any need for this change in your custom plugins?

TBH, no, but @Ayanda-D's comment [here](https://github.com/rabbitmq/rabbitmq-server/pull/16046#issuecomment-5279990887) does do a good job of explaining why this change is useful.

michaelklishin:
@Ayanda-D I finally have good news beyond the benchmarks. I think we have enough consensus internally to accept these changes and backport them to `v4.3.x`.

I will review this branch again with the additional changes and the benchmarks data we have.

michaelklishin:
@Ayanda-D @lukebakken I have added a few changes in https://github.com/rabbitmq/rabbitmq-server/pull/17182, most notably the `rabbitmq.conf` key naming scheme that follows the naming conventions we have around collections.

It's not clear to me from the comments if something else may be coming from @Ayanda-D, so my PR is a draft for now, and I will rebase it on anything that lands to this branch, or an explicit confirmation that nothing will and we can move to #17182.

Thanks.

Ayanda-D:
@michaelklishin @lukebakken I was concerned on the naming scheme, but renaming the key from `priority` to `priorities` from this commit https://github.com/rabbitmq/rabbitmq-server/pull/17182/changes/9d124dfe470b2c068d5a52317c9b837dc688b811 sounds good. thanks.

Then on a different note, @lukebakken, we keep seeing "force pushed" updates from your side on the "entire PR". Which affects the entire commit history of this work. This is making it very very difficult for us (Bloomberg to be specific) to track what specific changes you or your organization have added, on top of what Bloomberg has contributed. We'd like to keep it clear what commits/updates are from Bloomberg, and which are added. (thank you for all the review comments, which we've addressed).

JFYI, each of our contributions/commits to open source projects are part of the bigger Bloomberg's OSS family of projects which are tracked - https://www.bloomberg.com/company/values/tech-at-bloomberg/open-source/projects/ (so the reflected contributions/commits/commit history/etc are important for this department at Bloomberg).

I'll restore the commit history as before, rebased to latest `main`, and @michaelklishin please rebase https://github.com/rabbitmq/rabbitmq-server/pull/17182 again, thanks. If there's any extra commits to be added, we'd really appreciate it if these are pushed in isolation, like https://github.com/rabbitmq/rabbitmq-server/pull/17182/changes/9d124dfe470b2c068d5a52317c9b837dc688b811 and not "force" to affect the entire commit history. Thank you! 🙏

lukebakken:
> Then on a different note, @lukebakken, we keep seeing "force pushed" updates from your side on the "entire PR".

This is keeping your PR branch up-to-date with `main` is all. You're the first person to bring this up as an issue. Please keep your PR branches up-to-date with `main` on your own going forward. Depending on how @michaelklishin merges your PR, the commits may be rewritten anyway, so take that into account as well.

Ayanda-D:
On my personal contributions, definitely I don't mind these things and folks are free to update as they like. This is just Bloomberg OSS work, and not my own. Thats all. I'll try keep branch up to date with main. Thanks.

Regarding:

> @Ayanda-D thanks, that explains a lot around Bloomberg engineering team communication (or lack of, rather). I understand how things work at companies of that scale.

Thats for @willhoy to look into :) hopefully we can setup something for the comms 👍 thanks

michaelklishin:
@Ayanda-D I will rebase #17812. We do not rewrite commits or tweak authorship, my changes land on top of the original branch.

Note that when a commit is backported, its parent commit changes and thus the commit SHA. We cherry-pick with `-x` to preserve the original commit SHA for backporting history tracing with only the repo (without access to the backporting infrastructure logs and other data).

michaelklishin:
@Ayanda-D if you want to keep track of the Bloomberg-contributed changes specifically, I suggest filtering for contributor emails. Unfortunately SHAs of git commits can and will change when backporting or (well considered, rare) commit amendments can potentially be used.

michaelklishin:
Moved to https://github.com/rabbitmq/rabbitmq-server/pull/17182.

PR #16048, opened 2026-04-13
AMQP 0-9-1: refactoring

michaelklishin:


deps/rabbit/src/rabbit_amqp_reader.erl
```diff
@@ -944,24 +956,36 @@ set_credential0(Cred,
             handle_exception(State, 0, Error)
     end.
 
+%% erlang:send_after/3 accepts at most 2^32-1 milliseconds (~49.7 days).
```
ansd: IIUC this is no longer the case since OTP 18:
```
	      The API of BIF timers has also been extended. Timeout values
	      are for example no longer limited to 32-bit integers. For
	      more information see the documentation of
	      erlang:start_timer/4, erlang:send_after/4,
	      erlang:cancel_timer/2, and erlang:read_timer/2.
```
https://www.erlang.org/patches/otp-18.0

ansd (changes requested):

deps/rabbit/src/rabbit_amqp_reader.erl
michaelklishin: Great. I will back out the timer "overflow" bits.

deps/rabbit/src/rabbit_amqp_reader.erl
```diff
@@ -962,6 +959,13 @@ maybe_start_credential_expiry_timer(User) ->
             end
     end.
 
+cancel_credential_expiry_timer(undefined) -> ok;
+cancel_credential_expiry_timer(Ref) ->
+    _ = erlang:cancel_timer(Ref),
```
ansd, line 964: Why did you remove the `{info, false}` option?

deps/rabbit/src/rabbit_amqp_reader.erl
```diff
@@ -962,6 +959,13 @@ maybe_start_credential_expiry_timer(User) ->
...
+    receive credential_expired -> ok
```
ansd, line 965: Why do you need to receive/flush the `credential_expired ` message? The timer got already cancelled at this point.

If a credential expires a few milliseconds before or after the client refreshes the token, it's fine to close the connection. It's the client's responsibility to refresh the OAuth 2.0 token in time, at least several seconds, or ideally even minutes before the currently active token expires.

ansd (changes requested):

deps/rabbit/src/rabbit_amqp_reader.erl
ansd: Thank you.

deps/rabbit/src/rabbit_amqp_reader.erl
michaelklishin: Because this way when `erlang:cancel_timer/1` returns, we know that the timer was cancelled.

michaelklishin: I will back out this part, it's not particularly important to have.

michaelklishin:
Will submit a new version.

michaelklishin:
A much narrower in scope version that does not try to unify anything and therefore does not touch the Stream Protocol reader, for example.

https://github.com/rabbitmq/rabbitmq-server/pull/16092

ansd:
Thank you @michaelklishin !

PR #16055, opened 2026-04-14
Stream Protocol reader: simplify chunk selector

michaelklishin:
It only ever deals with two possible values.


deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -4139,8 +4139,10 @@ send_close_and_increment(Transport, S) ->
     increase_protocol_counter(?UNKNOWN_FRAME).
 
 get_chunk_selector(Properties) ->
-    binary_to_atom(maps:get(<<"chunk_selector">>, Properties,
-                            <<"user_data">>)).
+    case maps:get(<<"chunk_selector">>, Properties, <<"user_data">>) of
+        <<"all">> -> all;
+        _         -> user_data
```
kjnilsson, line 4144: redundant whitespace.

PR #16061, opened 2026-04-14
AMQP 1.0 binary parser: more specific, stricter fun head for some collection types

michaelklishin:
Plus a few more property-based tests.

### DIFF
--- a/deps/amqp10_common/src/amqp10_binary_parser.erl
+++ b/deps/amqp10_common/src/amqp10_binary_parser.erl
@@ -136,6 +136,9 @@ parse_array1(Count, <<?DESCRIBED, Rest/binary>>) ->
                        end, List),
     % this format cannot represent an empty array of described types
     {array, {described, Descriptor, Type}, Values};
+parse_array1(Count, <<Type, ArrayBin/binary>>)
+  when Count > byte_size(ArrayBin) ->
+    exit({failed_to_parse_array_count_exceeds_input, Type, Count, byte_size(ArrayBin)});
 parse_array1(Count, <<Type, ArrayBin/binary>>) ->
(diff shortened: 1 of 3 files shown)


deps/amqp10_common/src/amqp10_binary_parser.erl
```diff
@@ -136,6 +136,9 @@ parse_array1(Count, <<?DESCRIBED, Rest/binary>>) ->
                        end, List),
     % this format cannot represent an empty array of described types
     {array, {described, Descriptor, Type}, Values};
+parse_array1(Count, <<Type, ArrayBin/binary>>)
+  when Count > byte_size(ArrayBin) ->
+    exit({failed_to_parse_array_count_exceeds_input, Type, Count, byte_size(ArrayBin)});
```
ansd, line 141: This fix assumes that every element in an AMQP 1.0 array requires at least 1 byte of payload data. Because of this assumption, it enforces that the number of elements (Count) can never be greater than the remaining bytes in the payload (byte_size(ArrayBin)).
  However, this violates the AMQP 1.0 specification. The protocol explicitly defines several "zero-width" primitive types, such as null (0x40), true (0x41), false (0x42), and list0 (0x45).
  When an array is constructed using one of these zero-width types, the elements consume exactly 0 bytes in the data payload. For example, a perfectly valid array containing 3 null elements will have a Count of 3 but a payload size of 0 bytes.
  The proposed guard clause (3 > 0) would incorrectly reject this valid array, breaking AMQP 1.0 protocol compliance for any client that sends arrays of zero-width types.

ansd: fixed in https://github.com/rabbitmq/rabbitmq-server/pull/16174

PR #16077, opened 2026-04-14
`mc_amqpl`: stricter validation of parsed values

michaelklishin:


deps/rabbit/src/mc_amqpl.erl
```diff
@@ -421,22 +423,32 @@ convert_to(mc_amqp, #content{payload_fragments_rev = PFR} = Content, Env) ->
                  %% `type' doesn't have a direct equivalent so adding as
                  %% a message annotation here
                  MAC = map_add(symbol, <<"x-basic-type">>, utf8, Type, MAC0),
-                 #'v1_0.message_annotations'{content = MAC};
-             Section ->
-                 Section
+                 #'v1_0.message_annotations'{content = MAC}
          end,
+    ReversedPFR = lists:reverse(PFR),
```
ansd, line 428: `PFR` stands for Payload Fragments Reversed.
So `ReversedPFR` stands for `Reversed Payload Fragments Reversed`.
This is double negation. Can you rename this variable to just `PayloadFragements` or in short `PF`?

ansd (changes requested):

deps/rabbit/src/mc_amqpl.erl
```diff
@@ -856,11 +868,23 @@ is_internal_header(_) ->
...
+                _         -> undefined
+            catch
+                _:_ -> undefined
+            end;
         _ ->
             undefined
     end.
 
+is_amqp10_section(T) when is_tuple(T), tuple_size(T) >= 2, is_atom(element(1, T)) ->
+    case atom_to_binary(element(1, T)) of
```
ansd, line 882: Let's avoid here `atom_to_binary/1` since this may be expensive to call on the hot path. Instead can we match directly on the relevant AMQP 1.0 message section records?

ansd:
As decided by @michaelklishin and @kjnilsson and myself, let's close this PR and instead remove entirely parsing of the following headers in 4.4:
```
-define(AMQP10_PROPERTIES_HEADER, <<"x-amqp-1.0-properties">>).
-define(AMQP10_APP_PROPERTIES_HEADER, <<"x-amqp-1.0-app-properties">>).
-define(AMQP10_MESSAGE_ANNOTATIONS_HEADER, <<"x-amqp-1.0-message-annotations">>).
-define(AMQP10_FOOTER, <<"x-amqp-1.0-footer">>).
```
These headers are parsed only for backwards compat for the case where a message was sent via AMQP 1.0 to a <= 3.13 node.

PR #16085, opened 2026-04-14
AMQP 0-9-1: apply `configure` checks to passive queue, exchange declarations

michaelklishin:
Just like we do for regular declarations.

### DIFF
--- a/deps/rabbit/src/rabbit_amqp_management.erl
+++ b/deps/rabbit/src/rabbit_amqp_management.erl
@@ -78,11 +78,13 @@ handle_http_req(<<"GET">>,
                 _Query,
                 null,
                 Vhost,
-                _User,
+                User,
                 _ConnPid,
-                PermCaches) ->
+                {PermCache0, TopicPermCache}) ->
     QNameBin = cow_uri:urldecode(QNameBinQuoted),
     QName = queue_resource(Vhost, QNameBin),
+    PermCache = check_resource_access(QName, configure, User, PermCache0),
+    PermCaches = {PermCache, TopicPermCache},
     case rabbit_amqqueue:with(
            QName,
            fun(Q) ->
--- a/deps/rabbit/src/rabbit_channel.erl
+++ b/deps/rabbit/src/rabbit_channel.erl
@@ -2461,9 +2461,10 @@ handle_method(#'queue.declare'{queue       = QueueNameBin,
 handle_method(#'queue.declare'{queue   = QueueNameBin,
                                nowait  = NoWait,
                                passive = true},
-              ConnPid, _AuthzContext, _CollectorPid, VHostPath, _User) ->
+              ConnPid, AuthzContext, _CollectorPid, VHostPath, User) ->
     StrippedQueueNameBin = strip_cr_lf(QueueNameBin),
     QueueName = rabbit_misc:queue_resource(VHostPath, StrippedQueueNameBin),
+    check_configure_permitted(QueueName, User, AuthzContext),
     Fun = fun (Q0) ->
               QStat = maybe_stat(NoWait, Q0),
               {QStat, Q0}
@@ -2569,9 +2570,10 @@ handle_method(#'exchange.declare'{exchange    = XNameBin,
                                             AutoDelete, Internal, Args);
 handle_method(#'exchange.declare'{exchange    = ExchangeNameBin,
                                   passive     = true},
-              _ConnPid, _AuthzContext, _CollectorPid, VHostPath, _User) ->
+              _ConnPid, AuthzContext, _CollectorPid, VHostPath, User) ->
     ExchangeName = rabbit_misc:r(VHostPath, exchange, strip_cr_lf(ExchangeNameBin)),
     check_not_default_exchange(ExchangeName),
+    check_configure_permitted(ExchangeName, User, AuthzContext),
     _ = rabbit_exchange:lookup_or_die(ExchangeName).
 
 handle_deliver(CTag, Ack, Msgs, State) ->
--- a/deps/rabbit/test/amqp_auth_SUITE.erl
+++ b/deps/rabbit/test/amqp_auth_SUITE.erl
@@ -96,7 +96,12 @@ groups() ->
        bind_to_topic_exchange,
        unbind_queue_source,
        unbind_queue_target,
-       unbind_from_topic_exchange
+       unbind_from_topic_exchange,
+
+       %% Passive topology operations
+       passive_queue_declare,
+       passive_exchange_declare,
+       get_queue
       ]
      }
     ].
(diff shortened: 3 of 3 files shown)


m1schka-bdr:
FYI: this is a breaking change within a patch version (4.2.5 to 4.2.6). we rely on queues being defined and declared externally, the prod user in our case do not have the configure permission (for at least the last 5 years). we automatically install patch versions while locking minors and this almost broke our production system

Kujoen:
@michaelklishin sorry to bother here, but wasn't changing this behavior a breaking change? If I understand correctly, a user without configure permissions on a resource could perform a passive declare on it? (e.g check if it exists). Reason i'm here is because we were consuming from queues with users without config permissions using spring annotations:

`@RabbitListener(queues = ["..."])`

And I believe this change was breaking for us when upgrading from 4.25 -> 4.3 (expecting nothing to break).

Best,
Andreas F.

Kujoen:
@m1schka-bdr amazing timing 😄

Kujoen:
@m1schka-bdr In our case it did not just "almost" break prod 😭

michaelklishin:
@m1schka-bdr @Kujoen take a moment to think about what else that change might be, that we cannot yet reveal publicly. Yes, sometimes the only way to fix something is to change a behavior, and whether it's OK to ship such fixes in a patch release entirely depends on who you ask.

Again, take a some time to consider under what circumstances and as what kind of an issue  the original behavior could have been flagged, and whether SemVer purity is a good argument to apply in those contexts. And whether the team might have been under pressure to address the pre-existing behavior quickly and ship it without waiting months.

michaelklishin:
That's a breaking change for a behavior that was never intentional. "Our system depended on passive declares requiring no permissions" is a dubious excuse.

ansd:
@Kujoen @m1schka-bdr can you confirm that your users which do a "passive declare" have "read" or "write" permissions at the very least (given that they don't have "configure" permissions)?

Kujoen:
@ansd Yeah they have read and write permissions. They did not have configure permissions as the queues are declared by another service akin to @m1schka-bdr

Kujoen:
@ansd Technically, now being forced to give them configure permissions, these users can now delete their own queues?

kjnilsson:
> @ansd Technically, now being forced to give them configure permissions, these users can now delete their own queues?

Correct. We totally understand that is undesirable and are currently discussing a change to this behavour.

m1schka-bdr:
agree to @Kujoen. our technical users just have read permissions for queues ( [Principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) ). in our case Spring boot internally does a passive declare to check the existence of the queue before consuming but our technical users should not have the possibility to alter its config

michaelklishin:
@m1schka-bdr so you apply the principle of least privilege by requiring that RabbitMQ perform no checks like it did before the change? That's an interesting approach to security, "no permission checks, no security issues".

Anyhow, the simplest option is to simply revert the change and ignore it on all future security scans as "yes, this should be enforced but our community relies on that not being the case". For the name of security!

michaelklishin:
By the way, since we effectively are forced to do an early disclosure: reverting this will re-enable resource enumeration via passive declares. This requires an authenticated connection but still, there is a certain security cost to this convenience.

michaelklishin:
@kjnilsson @ansd have suggested a tweak to the permission required **specifically for passive declares**: instead of requiring `configure` specifically (just like for regular `queue.declare` operations), we can check for *any* permission, which would prevent resource enumeration but also allow any user that has any permission, whether it's `read` or `write` semantically, to passively declare the queues they intend to use.

I am working on a PR.

Kujoen:
@michaelklishin I think the more effective argument here is why is springs RabbitMQListener implementation enforcing a passive declare on a purely consuming connection (which isn't your fault or problem). If this wasn't the case, I'm sure this wouldn't really have caused issues. I think we are just a bit bummed out because it meant indeed increasing the user permissions to include destructive deletes, which was the only reasonable quick fix, as the alternative was going away from spring-rabbit and implementing the consumers using the java client.

Also I totally get you not wanting to wait a full major release, even though this security patch does break previous (allbeit unintentional) behavior.

Edit:
I just saw your response, thank you very much for the tweak. Also in general thank you for your work, we weren't trying to attack you or the work you do 👍

michaelklishin:
@Kujoen I cannot answer that question. I do have a direct communication channel with the Spring team but I'm afraid we cannot promise anything around Spring changes.

What our team can do is to tweak RabbitMQ, which is what I'm focusing on.

m1schka-bdr:
> @michaelklishin I think the more effective argument here is why is springs RabbitMQListener implementation enforcing a passive declare on a purely consuming connection (which isn't your fault or problem). If this wasn't the case, I'm sure this wouldn't really have caused issues. I think we are just a bit bummed out because it meant indeed increasing the user permissions to include destructive deletes, which was the only reasonable quick fix, as the alternative was going away from spring-rabbit and implementing the consumers using the java client.
> 
> Also I totally get you not wanting to wait a full major release, even though this security patch does break previous (allbeit unintentional) behavior.
> 
> Edit: I just saw your response, thank you very much for the tweak. Also in general thank you for your work, we weren't trying to attack you or the work you do 👍

I could not have put it here in better words. We're not attacking or blaming anyone here. I'm also not the guy developing those Spring applications, I'm on the infra side.

I'm also pro security, don't get me wrong. If you want to (keep) enforce the configure permission on passive declare, do it. Then it's the way it is and the spring library needs to deal with consumers only consuming and not passively declaring the queue. I think it's their way of "create if not exist" or "check if it exists", which is fine, except that until now this didn't require the configure permission when the queue was already there and an exception was thrown when it's not there because the user had no permission (which is IMO also a valid way, because the consumer is breaking and signaling that there is something wrong with the config or it's assertion that the queue must be there before starting to consume)

michaelklishin:
Voilà https://github.com/rabbitmq/rabbitmq-server/pull/16272.

Kujoen:
@michaelklishin Merci!

bou3108:
I would like to deeply thank you for your reactivity and caring of the community requests.

The project I'm working on has the same issues than @Kujoen  and @m1schka-bdr and I couldn't say better than them :
- we chose to rely on a 'no configure permissions policy' because we didn't understand what was the design behind that. We just wanted to avoid users to delete their own queues (they have only read permissions on their queues, and write on a specific exchange)
- we perfectly understand security issues and the pressure that comes with fixing them as quickly as possible, and we would have change our configuration if needed
- but the tweak you suggested sounds that a very acceptable compromise, and very handy for us
- in general, working with RabbitMQ is a pleasure as the maintainers do an outstanding job, in terms of features and quickness (and the A+++ quality of the documentation).

So thank you for your commitment in general and the solution you brought up in this particular case !

michaelklishin:
Thank you for the kind words, @bou3108!

PR #16142, opened 2026-04-15
Revert CQ shared store: Delete from index on remove or roll over (#13959)

### LINKED ISSUE #16141: Classic queue message store GC cannot keep pace under high throughput after 0278980ba0
## Summary

Commit 0278980ba0 (PR #13959, "CQ shared store: Delete from index on remove or roll over") introduced a regression in the classic queue message store GC that causes unbounded disk growth under sustained publish load when a slow-consumer queue shares the same vhost as high-throughput queues.

The regression is present in the current `main` branch. Reverting 0278980ba0 restores stable disk behavior.

## Root Cause

PR #13959 changed `scan_and_vacuum_message_file` in `delete_file` to an eager index cleanup mechanism. As a side effect, messages removed from non-current files now produce `not_found` index lookups during `scan_and_vacuum_message_file` instead of `previously_valid` ones. This was noted in the PR review by @gomoripeti:

> if I see it correctly compaction might become slower with this change (as during `scan_file_for_valid_messages` before this change there were ref-count=0 index entries which resulted in a `previously_valid` status while now `not_found` entries result in `invalid` status, and causing a scan_next_byte scanning mode)

Under high throughput with many queues, the GC compaction rate drops far enough that it cannot keep pace with the publish rate. Files accumulate faster than they are reclaimed, and disk usage can grow without bound. The GC stall also causes consumer latency spikes and broker unresponsiveness on established TCP connections.

## Reproduction

Three concurrent workloads on a single RabbitMQ node (m7g.large, 196 GB EBS):

- **main-workload**: 100 classic queues, 100 producers + 100 consumers, 120 KB messages, 5 msg/s per producer (500 msg/s aggregate), consumers acking immediately
- **slow-ack-publisher**: 1 producer, 3 msg/s to `slow-ack-queue`, 120 KB messages
- **slow-ack-consumer**: consumer on `slow-ack-queue` holding acks for 1-30 minutes (up to 1000 messages in flight simultaneously)

All queues in the same vhost with `queue-version: 2` policy.

Reproduction scripts: https://github.com/lukebakken/rmq-gc-lag

## Observed Behavior

**With 0278980ba0 present (`main`):**

Disk declined ~3.1 GB in 6 minutes during the baseline phase (200 msg/s), then briefly recovered when the spike phase began (500 msg/s), then resumed declining. Over ~100 minutes of monitoring, disk fell from 185.4 GB to ~169 GB - a loss of ~16 GB. Ready messages grew from 0 to 3500-4200 as the broker fell behind on delivery, and unacked messages accumulated to 3500+ across multiple reconnect cycles.

The GC stall also causes consumer latency spikes. The broker stopped sending data on an established TCP connection long enough to trigger a client-side socket read timeout:

```
[AMQP Connection 10.0.1.90:5672] ERROR - An unexpected connection driver error occurred
java.net.SocketTimeoutException: Read timed out
```

Consumer latency at time of socket read timeout:
```
min/median/75th/95th/99th/max consumer latency:
64886 / 1,511,958 / 6,254,378 / 46,742,920 / 54,532,926 / 568,205,100 µs
```

(median 1.5s, 95th 46s, 99th 54s, max 568s)

Grafana dashboard showing continuous disk decline on unpatched main

**With 0278980ba0 reverted (branch `lukebakken/cq-gc`):**

Disk stable in a 0.5 GB oscillation band (184.96-185.47 GB) throughout three consecutive 20-minute monitoring windows (60 minutes total) under the same workload at 500 msg/s with ~1000 unacked messages. Ready messages held at 0 throughout. No latency spikes, no broker unresponsiveness.

Grafana dashboard showing stable disk on patched lukebakken/cq-gc branch

**With v3.13.7 (pre-regression, pre-refactor):**

Disk stable throughout a 23-minute run under the same workload. Note: v3.13.7 predates the major `rabbit_msg_store` refactor that introduced the shared store architecture, so this data point establishes a pre-regression baseline but is not directly comparable to `main`.

Grafana dashboard showing stable disk on v3.13.7

## Fix

Revert 0278980ba0. Three independent improvements from that commit can be retained safely:

- `compact_file/2` early-exit guard (file already deleted)
- `prioritise_cast/3` in `rabbit_msg_store_gc` (delete requests before compaction)
- `index_update_fields` assertion relaxed (`true=` to `_=`)

See branch `lukebakken/cq-gc` on https://github.com/lukebakken/rmq-rabbitmq-server for the revert with retained improvements.

## Workaround

Move queues with long consumer timeouts to a dedicated vhost. This gives them a separate message store instance whose unacked messages do not pin files in the shared store. Confirmed effective: disk stable throughout a 40-minute run with the same workload after vhost isolation.

lukebakken:
## Summary

Reverts 0278980ba0 ("CQ shared store: Delete from index on remove or roll over", PR #13959), which introduced a regression in the classic queue message store GC. Three independent improvements from that commit are retained.

Fixes #16141.

## Problem

PR #13959 replaced the `scan_and_vacuum_message_file` call in `delete_file` with an eager index cleanup mechanism (`current_file_removes`). As a side effect, messages removed from non-current files now produce `not_found` index lookups during `scan_and_vacuum_message_file` instead of `previously_valid` ones. This was noted in the PR review by @gomoripeti but not addressed before merge.

Under high throughput with many queues, the byte-by-byte `scan_next_byte` scanning mode triggered by `not_found` entries causes GC compaction to fall far enough behind the publish rate that disk usage can grow without bound. The stall also causes consumer latency spikes and broker unresponsiveness on established TCP connections.

Reproduction: 100 classic queues at 500 msg/s (120 KB messages) with a slow-ack consumer queue in the same vhost (acks held 1-30 min, up to 1000 in flight). On an m7g.large with a 196 GB EBS volume, disk fell from 185.4 GB to ~169 GB in ~100 minutes. Consumer latency reached a median of 1.5s and a max of 568s.

Reproduction scripts: https://github.com/lukebakken/rmq-gc-lag

## Changes

Two commits:

1. Revert 0278980ba0 in full.
2. Restore three independent improvements from that commit that are unrelated to the broken `current_file_removes` mechanism:
   - Relax `index_update_fields` assertion (`true=` to `_=`) so a missing key does not crash the process
   - Add `prioritise_cast/3` to `rabbit_msg_store_gc` so delete requests are processed before compaction requests, avoiding unnecessary compaction of files already pending deletion
   - `compact_file/2` early-exit guard (file already deleted) was already present after the revert

## Testing

Ran the reproduction workload against this branch for 60 minutes (three consecutive 20-minute monitoring windows) at 500 msg/s with ~1000 unacked messages. Disk held stable in a 0.5 GB oscillation band (184.96-185.47 GB). Ready messages held at 0 throughout. No latency spikes.

For comparison, the same workload against unpatched `main` lost ~16 GB of disk in ~100 minutes with ready messages growing to 3500-4200.

### DIFF
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -80,8 +80,6 @@
           current_file_handle,
           %% current write file offset
           current_file_offset,
-          %% messages that were potentially removed from the current write file
-          current_file_removes = [],
           %% TRef for our interval timer
           sync_timer_ref,
           %% files that had removes
@@ -1166,11 +1164,7 @@ write_message(MsgId, Msg, CRef,
               end, CRef, State1)
     end.
 
-remove_message(MsgId, CRef,
-               State = #msstate{
-                        index_ets = IndexEts,
-                        current_file = CurrentFile,
-                        current_file_removes = Removes }) ->
+remove_message(MsgId, CRef, State = #msstate{ index_ets = IndexEts }) ->
     case should_mask_action(CRef, MsgId, State) of
         {true, _Location} ->
             State;
@@ -1182,32 +1176,22 @@ remove_message(MsgId, CRef,
             %% ets:lookup(FileSummaryEts, File),
             State;
         {_Mask, #msg_location { ref_count = RefCount, file = File,
-                                total_size = TotalSize } = Entry}
+                                total_size = TotalSize }}
           when RefCount > 0 ->
             %% only update field, otherwise bad interaction with
             %% concurrent GC
+            Dec = fun () -> index_update_ref_counter(IndexEts, MsgId, -1) end,
             case RefCount of
                 %% Don't remove from cur_file_cache_ets here because
                 %% there may be further writes in the mailbox for the
-                %% same msg. We will remove 0 ref_counts when rolling
-                %% over to the next write file.
-                1 when File =:= CurrentFile ->
-                    index_update_ref_counter(IndexEts, MsgId, -1),
-                    State1 = State#msstate{current_file_removes =
-                        [Entry#msg_location{ref_count=0}|Removes]},
-                    delete_file_if_empty(
-                      File, gc_candidate(File,
-                        adjust_valid_total_size(
-                              File, -TotalSize, State1)));
-                1 ->
-                    index_delete(IndexEts, MsgId),
-                    delete_file_if_empty(
-                      File, gc_candidate(File,
-                        adjust_valid_total_size(
-                              File, -TotalSize, State)));
-                _ ->
-                    index_update_ref_counter(IndexEts, MsgId, -1),
-                    gc_candidate(File, State)
+                %% same msg.
+                1 -> ok = Dec(),
+                     delete_file_if_empty(
+                       File, gc_candidate(File,
+                         adjust_valid_total_size(
+                               File, -TotalSize, State)));
+                _ -> ok = Dec(),
+                     gc_candidate(File, State)
             end
     end.
 
@@ -1269,9 +1253,7 @@ flush_or_roll_to_new_file(
                      cur_file_cache_ets  = CurFileCacheEts,
                      file_size_limit     = FileSizeLimit })
   when Offset >= FileSizeLimit ->
-    %% Cleanup the index of messages that were removed before rolling over.
-    State0 = cleanup_index_on_roll_over(State),
-    State1 = internal_sync(State0),
+    State1 = internal_sync(State),
     ok = writer_close(CurHdl),
     NextFile = CurFile + 1,
     {ok, NextHdl} = writer_open(Dir, NextFile),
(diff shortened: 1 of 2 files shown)


michaelklishin:
@lukebakken Loïc will be back next week. We will have a day or two to decide before the final `4.3.0` cut-off/freeze.

lukebakken:
Well this is something that shouldn't be rushed!

lhoguin (approved):
Perhaps I went a bit too far on this. Deletion was too expensive so the PR fixed that, at the cost of making some compactions more expensive. A plain revert would mean we'd just go back to that state, but since we keep some changes I'm OK with merging this now and seeing how it fares overall in the wild.

lhoguin:
In particular perhaps the prioritisation of delete is good enough to avoid the bigger issues of deletes.

lhoguin:
At some point we should rework the file format to avoid these issues entirely. We need the file itself to contain information about messages existing, or holes existing, and what size they are. The in-memory index should only ever refer to existing messages and act more like a cache over what is in the files, to be used by readers.

michaelklishin:
@lukebakken so that you don't have to guess: the earliest this can ship now is `4.3.1`.

michaelklishin:
@mergifyio backport v4.3.x

(the PR will be marked as delayed for `4.3.1`)

lukebakken:
Thanks @lhoguin

PR #16185, opened 2026-04-17
Fix consumer count for quorum queues in queue.declare

kjnilsson:
Previously, `rabbit_fifo_client:stat/2` incorrectly returned the number of checked out messages (`num_checked_out`) instead of the actual number of consumers (`num_consumers`). This caused `queue.declare` operations on quorum queues to return the unacknowledged message count in the `consumer_count` field rather than the true number of active consumers.

This commit:
- Updates `rabbit_fifo_client:stat/2` to correctly extract `num_consumers` from the ra machine overview.
- Adds `queue_declare_passive_consumer_count` to `quorum_queue_SUITE` to verify that passive declares return the correct consumer count when subscribing and cancelling consumers.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo_client.erl
+++ b/deps/rabbit/src/rabbit_fifo_client.erl
@@ -579,7 +579,7 @@ stat(Leader, Timeout) ->
     %% but we use it for backwards compatibilty
     case ra:member_overview(Leader, Timeout) of
       {ok, #{machine := #{num_ready_messages := R,
-                          num_checked_out := C}}, _} ->
+                          num_consumers := C}}, _} ->
             {ok, R, C};
       {error, _} = Error ->
             Error;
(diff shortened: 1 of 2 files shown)


Gsantomaggio:
It works correctly now, tested with:

```java
 Environment environment = new AmqpEnvironmentBuilder()
                .build();
        Connection connection = environment.connectionBuilder().build();
        connection.management().queue("my_qq").quorum().queue().declare();

        for (int i = 0; i < 15; i++) {
            Consumer consumer = connection.consumerBuilder().queue("my_qq").messageHandler((context, message) -> {
                System.out.println("Received message: " + new String(message.body()));
            }).build();
        }
       Management.QueueInfo info = connection.management().queueInfo("my_qq");

       System.out.println("Consumer count " + info.consumerCount() );
```

out:

```shell
Consumer count 15
```

Gsantomaggio (approved):

PR #16196, opened 2026-04-21
Support regular expression in OAuth 2.0 scopes

MarcialRosales:
## Proposed Changes

Support regular expression in OAuth 2.0 scopes.

Docs accompanying PR:  https://github.com/rabbitmq/rabbitmq-website/pull/2512

## Types of Changes

- [ ] Build system and/or CI


michaelklishin (changes requested):
With this PR, `check_vhost_access/3`, `check_resource_access/4` and `check_topic_access/4` all go via `resolve_resource_server/1` which needs one value, `ResourceServer#resource_server.scope_pattern_syntax`, and
computes it over and over,
for every client operation and message published, hitting the app environment (an ETS read) from `get_root_resource_server/0`.

That value will be immutable for the duration of the node's run, so it should be reasonably easy to fetch/compute once and pass around in the token map.

For example, something like this:

```diff
--- a/deps/rabbitmq_auth_backend_oauth2/src/rabbit_auth_backend_oauth2.erl
+++ b/deps/rabbitmq_auth_backend_oauth2/src/rabbit_auth_backend_oauth2.erl
@@ -142,10 +142,11 @@ update_state(AuthUser, NewToken) ->
                 {ok, DecodedToken} ->
                     CurToken = AuthUser#auth_user.impl,
                     case ensure_same_username(
                             ResourceServer#resource_server.preferred_username_claims,
                             CurToken(), DecodedToken) of
                         ok ->
                             Tags = tags_from(DecodedToken),
+                            Token1 = DecodedToken#{<<"x-rmq-scope-pattern-syntax">> =>
+                                                   ResourceServer#resource_server.scope_pattern_syntax},
                             {ok, AuthUser#auth_user{tags = Tags,
-                                                    impl = fun() -> DecodedToken end}};
+                                                    impl = fun() -> Token1 end}};
                         {error, mismatch_username_after_token_refresh} ->
                             {refused,
                                 "Not allowed to change username on refreshed token", []}
@@ -208,12 +209,8 @@ with_decoded_token(DecodedToken, Fun) ->
 with_scopes_for_resource_syntax(Token, Resource, Fun) ->
-    case resolve_resource_server(Token) of
-        {error, _} ->
-            false;
-        {ResourceServer, _} ->
-            Syntax = ResourceServer#resource_server.scope_pattern_syntax,
-            Scopes = get_expanded_scopes(Token, Resource),
-            Fun(Syntax, Scopes)
-    end.
+    Syntax = maps:get(<<"x-rmq-scope-pattern-syntax">>, Token, wildcard),
+    Scopes = get_expanded_scopes(Token, Resource),
+    Fun(Syntax, Scopes).

 %% This is a helper function used with HOFs that may return errors.
 -spec auth_user_from_token(Token, ResourceServer) -> Result
@@ -221,8 +218,10 @@ with_scopes_for_resource_syntax(Token, Resource, Fun) ->
 auth_user_from_token(Token0, ResourceServer) ->
     Username = username_from(
         ResourceServer#resource_server.preferred_username_claims,
         Token0),
     Tags     = tags_from(Token0),
+    Token1   = Token0#{<<"x-rmq-scope-pattern-syntax">> =>
+                       ResourceServer#resource_server.scope_pattern_syntax},
     {ok, #auth_user{username = Username,
                     tags = Tags,
-                    impl = fun() -> Token0 end}}.
+                    impl = fun() -> Token1 end}}.
```

And yes, the downside of this proposed change is that using `rabbitmqctl eval` to override the value won't be picked up but that's the 0.1% of cases, so perhaps we should optimize the hot code path for the 99.9% instead.

MarcialRosales:
@michaelklishin  Thanks for your comments. Yes, it sounds reasonable to cache it. I doubt it, but if we ever need to use another configuration value we can cache the entire resource_server record. 
I am applying those changes.

michaelklishin (approved):
I have renamed `regexpr` to `regex` in the `rabbitmq.conf` enum,
then introduces a number of restrictions for the user-provided
regular expressions:

1. A traversal depth limit (against ReDoS)
2. Escaping of metacharacters and slashes to prevent scopes with special characters from breaking the intended behavior (intentionally or not)
3. Pattern length limit

michaelklishin:
The slash validation was rolled back as breaking in https://github.com/rabbitmq/rabbitmq-server/pull/16265.

MarcialRosales:
Thank you so much and especially for making the code more robust against possible security attacks !!

PR #16220, opened 2026-04-27
Capture streams protocol in message size histogram metrics

the-mikedavis:
Currently the streams protocol does not affect the message size histogram metrics. This change takes a look at adding them.

I suspect that this wasn't added in the original PR because of batching? With batches we know the batch size and number of messages, but we'd need to scan or even decompress a batch to know the message sizes within. This change currently skips batches but we should discuss how to handle them properly.

### DIFF
--- a/deps/rabbitmq_stream/src/rabbit_stream.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream.erl
@@ -46,6 +46,7 @@ start(_Type, _Args) ->
                                 ?PROTOCOL_COUNTERS),
     rabbit_global_counters:init(#{protocol => stream,
                                   queue_type => ?STREAM_QUEUE_TYPE}),
+    rabbit_msg_size_metrics:init(stream),
     rabbit_stream_sup:start_link().
 
 tls_host() ->
--- a/deps/rabbitmq_stream/src/rabbit_stream_utils.erl
+++ b/deps/rabbitmq_stream/src/rabbit_stream_utils.erl
@@ -72,7 +72,7 @@ write_messages(?VERSION_1 = V, ClusterLeader,
                  Message:MessageSize/binary,
                  Rest/binary>>) ->
     write_messages0(V, ClusterLeader, PublisherRef, PublisherId, InternalId,
-                    PublishingId, Message, Rest);
+                    PublishingId, MessageSize, Message, Rest);
 write_messages(?VERSION_1 = V, ClusterLeader,
                PublisherRef,
                PublisherId,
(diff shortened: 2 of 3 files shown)


the-mikedavis:
One thing we could do is capture a histogram of sizes of batches. So batches would be skipped in the the message size metrics (current behavior of this branch), but we'd have an extra histogram you could scrape to see batch sizes and maybe number of messages within each batch. What do you think?

ansd:
> I suspect that this wasn't added in the original PR because of batching?

Yes, that's right. Since batching is the default, I thought it's better to omit the individual message sizes for the stream protocol entirely. The stream protocol's main advantage is extreme high throughput, which it achieves by treating message batches as opaque blobs.
Message sizes are still tracked in the histogram correctly if you publish to a stream via AMQP.
So, I'm not sure this partial approach as done in this PR is worth it (as in whether it leads to more clarity&insights or more confusion).

> extra histogram you could scrape to see batch sizes and maybe number of messages within each batch

Yes, the stream plugin already exposes its own Prometheus metrics, so we could add another histogram there. This approach makes much more sense to me.
(Should this new histogram also include the batch sizes of non-Stream protocol publishers? If yes, it can't really live in the stream plugin.)

the-mikedavis:
> the stream plugin already exposes its own Prometheus metrics, so we could add another histogram there

This approach sounds good to me too, I'll take a look at this change 👍. I think it's reasonable for it to be specific to the streams protocol. Other protocols can't do batching though, can they? I guess I mean the client-side sub-batching concept (`osiris:batch()`) rather than the delivery of multiple messages in a frame - we can count multiple messages in one frame accurately.

Knowing how many messages are in a batch and the total batch size would be really useful for performance tuning / resource analysis. I ran some tests and batching even a few messages together on the client side very significantly cuts down on CPU cost.

ansd:
>  Other protocols can't do batching though, can they? 

RabbitMQ will batch internally even if messages arrive via other protocols.
There is a `batchable` field in the AMQP transfer frame. But we don't make use of it today.

> I think it's reasonable for it to be specific to the streams protocol.

I agree. I guess, one could argue that it might be useful to get some Prometheus metrics for batch sizes in streams independent of which protocol was used to send to a stream. (In that case this histogram would be emitted by Osiris?) But making this histogram specific to the stream protocol and the stream plugin is probably good enough for most use cases. So, if that's simplest, we can go with this option.

the-mikedavis:
Aha ok, makes sense. If we might send batches from other protocols, it might make sense to move this into Osiris. I will take another pass and see where it fits best.

the-mikedavis:
Ok, I'm thinking that this fits better into Osiris now and we can drop the changes here. We could have two histograms:

* Simple entry bytes (single messages)
* Batch bytes (compressed size)

We could also have a histogram tracking number of messages in a batch but maybe that is not so interesting. It's a little bit orthogonal to this but I think a histogram of chunk sizes would be useful, both to know chunk sizes and as a proxy for a "chunks/sec" rate. I think it would be useful to track these node-wide within Osiris, not per-stream.

What do you think?

PR #16253, opened 2026-04-28
Refactor Direct Reply-to one more time

michaelklishin:
There are no functional (user-observable) changes.

"TTB" in the tests is for "term to binary".


deps/rabbit/src/rabbit_pid_codec.erl
```diff
@@ -9,6 +9,7 @@
 
 -export([decompose/1,
          decompose_from_binary/1,
+         decompose_from_binary_safe/1,
```
kjnilsson: Do we need a new function or can we just change the old one? AFAIK only the volatile queue uses decomposition.

deps/rabbit/src/rabbit_pid_codec.erl
```diff
@@ -42,6 +50,21 @@ decompose_from_binary(Bin) ->
...
+decompose_from_binary_safe(<<?TTB_PREFIX, ?NEW_PID_EXT,
+                             ?ATOM_UTF8_EXT, Len:16/integer, Node:Len/binary,
+                             ID:32/integer, Serial:32/integer,
+                             Creation:32/integer>>) ->
+    {ok, #{node => Node, id => ID, serial => Serial, creation => Creation}};
+decompose_from_binary_safe(<<?TTB_PREFIX, ?NEW_PID_EXT,
+                             ?SMALL_ATOM_UTF8_EXT, Len:8/integer, Node:Len/binary,
+                             ID:32/integer, Serial:32/integer,
+                             Creation:32/integer>>) ->
+    {ok, #{node => Node, id => ID, serial => Serial, creation => Creation}};
```
kjnilsson: suggestion: put map fields on separate lines.

just a personal preference, helps with future diffs.

kjnilsson (changes requested):
just wondering if we need a new function for this.

deps/rabbit/src/rabbit_pid_codec.erl
michaelklishin: Yeah, it was an abundance of caution I guess. I will modify the pre-existing one.

deps/rabbit/src/rabbit_volatile_queue.erl
```diff
@@ -377,19 +377,30 @@ pid_from_name(<<?PREFIX, Bin/binary>>, CandidateNodes) ->
...
     catch error:_ -> error
     end;
 pid_from_name(_, _) ->
     error.
 
+%% Returns the integer hash following "@" in a synthetic node-name binary
+%% such as <<"reply@1234567">>. Malformed input raises and is caught by
+%% the surrounding try.
+node_hash_from_binary(NodeBin) ->
+    [_Prefix, Suffix] = binary:split(NodeBin, <<"@">>),
```
ansd, line 402: We could use a compiled pattern here as done in https://github.com/rabbitmq/rabbitmq-server/blob/b8591acdbbcb3e2afcb24a9b3a07abf657e3b12a/deps/rabbit/src/rabbit_volatile_queue.erl#L369-L378 for performance reasons since this is called per message.

PR #16258, opened 2026-04-28
Fix `auth_backend_cache` cache miss on reconnect (#16255)

lukebakken:
Closes #16255

## What

The `rabbit_auth_backend_cache` cache backend no longer hits its cache across reconnections of the same user and credentials, which on busy deployments puts significant extra pressure on the upstream authentication backend (the reporter's deployment uses `rabbit_auth_backend_http`).

## Why

Since 4.2.0, `rabbit_auth_mechanism_plain` has passed the underlying Erlang socket to `rabbit_access_control:check_user_login/2` inside `AuthProps` as `{sockOrAddr, Socket}`. The cache backend uses `{FunctionName, [Username, AuthProps]}` as its ETS lookup key, and the raw Erlang port is unique per connection, so every reconnection produces a distinct key and the cache always misses. The regression was introduced in 6d24aef9b050c48ad65cbaf6c2d97875f322c914 as part of #13818 (the `rabbit_auth_backend_internal_loopback` plugin), which needs the socket in order to check whether the connection is from localhost.

## How

Replace the `sockOrAddr` entry in `AuthProps` with a precomputed `is_loopback` boolean at every producer. The boolean is stable across reconnections of the same peer, so the cache key is stable and reconnections hit the cache. The loopback backend now reads `is_loopback` directly rather than calling `rabbit_net:is_loopback/1` on the raw socket.

Files changed:

- `rabbit_auth_mechanism_plain:build_auth_props/2` computes `rabbit_net:is_loopback/1` once and emits `{is_loopback, Bool}`
- `rabbit_auth_backend_internal_loopback:user_login_authentication/2` reads `is_loopback` directly
- `rabbit_web_dispatch_access_control` (management HTTP basic auth) uses the same shape for consistency; its `sockOrAddr` value was already an IP tuple so it did not cause cache misses, but the shape is now aligned and raw peer terms no longer appear in the ETS cache
- `rabbit_auth_backend_http:extract_other_credentials/1` filters `is_loopback` out alongside `sockOrAddr` so that neither is forwarded to external HTTP auth servers

The `sockOrAddr` key is undocumented and not referenced in any behaviour callback, README, or website content, so removing it from `AuthProps` is not a public contract change.

## Test

[...]

Issue #16255, opened 2026-04-28
`4.2.x`: `rabbitmq_auth_backend_cache` does not have the expected effect


motmot80:
# Describe the bug

(not AI generated)

# Version and Environment
- Kubernetes environment
- RabbitMQ 4.2.4 (4.2.4-management-main + operator)
- Protocol: AMQPv1
- Authn/Authz: rabbit_auth_backend_cache, rabbit_auth_backend_http

# Authn/Authz-Configuration

```
auth_backends.1 = internal
auth_backends.2 = cache
auth_cache.cached_backend = http
auth_cache.cache_ttl      = 14400000
auth_cache.cache_refusals = false
auth_http.http_method   = get
auth_http.user_path     = ...
auth_http.vhost_path    = ...
auth_http.resource_path = ...
auth_http.topic_path    = ...
```

# Description
We migrated from RMQ 3.x to RMQ 4.2.4 and noticed that our internal http authentication backend was under much more pressure during connection peaks (AMQPv1). We changed nothing but the rabbitmq version itself.

# Expected
The cache backend within the rabbitmq host caches positive client authentications and authorizations for the timespan of `auth_cache.cache_ttl`. In RabbitMQ 3.x this was the case.

# Actual
Reconnecting the same client with the same credentials does not hit the cache.
Reconnecting always results in not using the cache and forwarding the request to the http auth backend.

# Details
We are not sure (because we are not even erlang amateurs), but it seems that the AuthProps seem to be part of the lookup key. In complex network environments with NAT and service-meshes etc it's pretty normal that the client connection is not particularly consistent.

So what's the use of having an authentication cache if the cache is invalidated as soon as the client reconnects?
If this is the case - is it by design? Could it be that the AuthProps were not sent along with the AMQPv1 plugin on RabbitMQ 3.x? Is there a workaround? 

# Reproduction steps

1. Configure rabbit_auth_backend_cache (rabbit_auth_cache_ets) => rabbit_auth_backend_http
2. Connect to the rabbitmq host with AMQPv1 (SASL plain)
3. Disconnect from the rabbitmq host
4. Change client ip (or maybe just the port)
5. Reconnect to the rabbitmq host

Result: rabbit_auth_backend_cache does not hit the cache and uses rabbit_auth_backend_http

# Expected behavior

The cache backend within the rabbitmq host caches positive client authentications and authorizations for the timespan of `auth_cache.cache_ttl`.

# Additional context

### Checking the ets cache
[...]

PR #16262, opened 2026-04-29
`rabbitmq_auth_backend_cache`: a follow-up to #16258 #16255

michaelklishin:
1. Avoid using credentials in cache keys, use a salted hash instead (the 2nd issue brought up in #16255)
2. More tests (mostly property-based)

## Why Not use `credential_obfuscation`?

This may seem like a case where we could reuse `credential_obfuscation`. Unfortunately,
this is not the case, even though the change in this PR is heavily inspired by it:

1. `credential_obfuscation` is non-deterministic, every `encrypt/{1,5}` call generates a new salt value
2. `credential_obfuscation` produces reversible value (an encrypted value can be decrypted using the same library on the same node without restarts) compared to hashing, and we do not need that to obfuscate the cache key component
3. This change is entirely node-local, while `credential_obfuscation` to a certain extent must take clustering into account for decryption of values stored on other nodes, namely runtime parameters with sensitive values: shovels

## References

References #16258 #16255.

### DIFF
--- a/deps/rabbit/src/rabbit_auth_mechanism_plain.erl
+++ b/deps/rabbit/src/rabbit_auth_mechanism_plain.erl
@@ -53,7 +53,9 @@ handle_response(Response, _State) ->
 
 build_auth_props(Pass, Socket) ->
     %% Pass a precomputed boolean, not the raw socket, so that AuthProps is
-    %% stable across reconnections of the same client. See issue #16255.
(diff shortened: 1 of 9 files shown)


michaelklishin:
@lukebakken @motmot80 this addresses the 2nd part of #16255: cache keys now use salted hashes instead of the sensitive input value.

PR #16272, opened 2026-04-29
Allow users with any permission type to perform passive declarations

michaelklishin:
In `4.3.0` and `4.2.6`, the standard `configure` permission was enforced for passive queue and exchange declarations, matching the "regular" declaration case.

However, this can be a breaking change for some applications that do not have the `configure` permission on target resources.

With this change, any permission (e.g. `read`) on the target resource will be considered sufficient as passive declaration is completely non-destructive by definition.

Note that this permission check prevents arbitrary resource enumeration via passive declares.

The change applies to both `queue.declare` and `exchange.declare`.

References #16085.


ansd (approved):

kjnilsson (approved):

michaelklishin:
Thank you, @ansd.

Issue #7593, opened 2023-03-11
More predictable and firm connection_max for limiting number of connections per RabbitMQ node


SimonUnge:
Today it is possible to limit maximum allowed connections per vhost, per user, or per node.

Running rabbitMQ as a service, having limits like this is very useful, to avoid brokers running out of memory. The current back pressure mechanism for high memory usage (memory alarm) does not stop clients from creating new connections, and thus have no real effect at improving memory usage. 

The per vhost and user logic is handled by rabbitmq, and is set via runtime configuration, and
works as expected, with a nice error when you hit the limit.

For node level connection limit, rabbitmq passes along the limit value, set as config, to ranch. The config parameter is 'connection_max', and its existence is not really advertised.

This works, but has a few drawbacks.
* It is a 'soft' limit, basically max_connections + the number of acceptors.
* Hitting the limit will result in ranch not accepting anymore connections. This will not result in further connections being rejected, and thus no with a user friendly error message, as the kernel option allows queueing incoming connections. Instead, the attempt to create a connection will hang until a timeout is reached.

Suggestion: Similar to how rabbitmq can be configured to limit connections per vhost or per user, add the option to also
limit number of connections per node, handled within rabbitmq, to give a nice informative reason to the client of why 
a connection request is not accepted.

SimonUnge:
@michaelklishin Would like to hear your guys thoughts about a limit like this?

michaelklishin:
I don't have any objections but why not use the same `connection_max` setting name?

SimonUnge:
@michaelklishin Yes, my first draft implementation did use the same configuration setting. The issue with this, though, is that we then would hit the limit in rabbitmq, and ranch, at the same time. Which would start ranch to queue up the connections, so called 'sleepers', until the 10 acceptors are used, and subsequent connections would then hit the same timeout issue - i.e hanging connection attempt until client timeout triggers. 

Perhaps had a config, say, `node_connection_max` and leave `connection_max` as is (default to infinity) and put `node_connection_max` in the documentation and kinda leave connection_max hidden as is?

michaelklishin:
@SimonUnge why not change the implementation but keep the key? If `node_connection_max` is for the node then what is `connection_max`?

It's a good question whether this should apply to all connections regardless of the protocol.

SimonUnge:
[...]

PR #16299, opened 2026-05-04
Add amqp10_connection_max to limit AMQP 1.0 connections per node

sauravonwww:
Add a new configuration option amqp10_connection_max that allows operators to set the maximum number of AMQP 1.0 client connections per node via rabbitmq.conf. This mirrors the existing connection_max setting which only applies to AMQP 0-9-1 connections.

The limit is enforced in rabbit_amqp_reader during the connection open handshake, after authentication but before the connection transitions to the running state. When the limit is reached, the broker closes the connection with amqp:resource-limit-exceeded.

The connection count is obtained from rabbit_amqp1_0:list_local() which tracks AMQP 1.0 connections via pg groups.

Changes:
- rabbit.schema: new mapping and translation for amqp10_connection_max
- rabbit_amqp_reader.erl: check_node_amqp10_connection_limit/0
- rabbitmq.conf.example: documented new config option
- rabbit.snippets: schema validation test entries
- per_node_limit_SUITE.erl: integration test for AMQP 1.0 limit

## Proposed Changes

## Types of Changes

- [ ] Build system and/or CI

## Checklist


michaelklishin:
A much better option would be to apply the existing limit to both AMQP 1.0 and AMQP 0-9-1 connections.

sauravonwww:
@michaelklishin  Thank you for the feedback! I understand the preference for simplicity — applying the existing `connection_max` to both protocols.                                                                                     
 
However, I'd like to share the reasoning behind a protocol-specific limit proposal: 

AMQP 1.0 connections have a significantly different resource profile compared to AMQP 0-9-1. In our benchmarking, AMQP 1.0 connections are roughly 50% more memory-efficient than AMQP 0-9-1. If we apply a single `connection_max` to both protocols, operators running primarily AMQP 1.0 workloads would be artificially constrained to a limit sized for the heavier AMQP 0-9-1 footprint. Conversely, a limit raised to accommodate AMQP 1.0's efficiency would be unsafe for AMQP 0-9-1 connections. The same argument extends to other protocols — MQTT connections have yet another resource profile. A per-protocol limit gives operators the flexibility to tune each protocol independently.

michaelklishin:
@sauravonwww there's already a per-node connection limit, a per-virtual host limit, a per-user limit. 

These limits is a connection leak protection mechanism, not a resource management one.

the-mikedavis:
I think there's some merit in having a node-wide connection limit per-protocol. For MQTT for example you'd want a different limit set than 0-9-1. (Also curious what you think about this specifically with MQTT @ansd)

michaelklishin:
I'm not against per-protocol limits in general but not sure how they will interact with the other, non-protocol-aware limits.

At least the two core protocols can be treated as a group.

the-mikedavis:
I had the 🧞 throw together a script to test connections between 0-9-1 and 1.0 and they do seem to impose roughly the same memory cost in my testing. I did notice a difference with an idle channel vs. a session though: sessions are noticeably lighter. But without channels or sessions, here's a run of that script on a few dummy single-instance brokers (m7g.xlarge):

Screenshot 2026-05-05 at 12 10 41

So a limit applying to either / both seems appropriate for these. We should think about the right approach for MQTT though.

Script: https://gist.github.com/the-mikedavis/45f1a00790bcc99ee9bb2c321c08920b

ansd:
Thank you @the-mikedavis !

> connections between 0-9-1 and 1.0 and they do seem to impose roughly the same memory cost in my testing. 

Yes, your memory benchmark results match mine in https://www.rabbitmq.com/blog/2024/08/21/amqp-benchmarks#summary-many-connections 

(As a side note: Last week, I was testing the new `hibernate_after` for supervisors - https://www.erlang.org/docs/28/apps/stdlib/supervisor.html#module-supervisor-flags. There are some tiny improvements if we run such artificial workloads with many connections and idle sessions/channels. I decided to not add this setting, but on a second thought maybe we can add this new setting.)

> At least the two core protocols can be treated as a group.

I think that's reasonable. (Note that operators can additionally limit the maximum number of sessions per AMQP 1.0 connection and maximum number of links per AMQP 1.0 session.)

> We should think about the right approach for MQTT though.

I also see that @SimonUnge introduced `ranch_connection_max` in https://github.com/rabbitmq/rabbitmq-server/commit/b42e99acfee56a0f1870b3fd5c47438eb83aa6cc . 

Is this sufficient for MQTT? If not, I don't have any objections introducing additional separate `connection_max` limits for applications `rabbitmq_mqtt`, `rabbitmq_stomp`,  and `rabbitmq_stream`. I suppose it makes sense to allow operators to configure less strict MQTT connection limits given that those connections are more lightweight.

the-mikedavis:
I will have to run some tests with `ranch_connection_max`. From the discussion in https://github.com/rabbitmq/rabbitmq-server/issues/7593 I'm not sure what behavior clients would see. I think the explicit errors and closing we do in the protocol readers would be clearer to clients.

michaelklishin:
@the-mikedavis @sauravonwww please take a look at https://github.com/rabbitmq/rabbitmq-server/issues/16347.

michaelklishin:
@the-mikedavis I remember working on that change with Simon U. The behavior we want is a clear and nearly immediate error, not a connection put into an acceptor queue (à la the kernel's inbound TCP connection queue).

PR #16318, opened 2026-05-06
Move from thoas to stdlib json

mkuratczyk:
jose was bumped to the latest version,
because the stdlib json module support
was added in 1.11.11.


michaelklishin (approved):
I just want to be fair to the Thoas maintainer and contributors: Thoas has served us well.

But it's hard to beat the ubiquity of the standard library, in particular if they are very comparable in terms of efficiency.

michaelklishin:
Backporting to `v4.3.x` and `v4.2.x` to reduce the delta. There should be no observable functional changes, including HTTP API efficiency.

ansd:
For the record, @mkuratczyk explained that this PR requires OTP 27. Hence, we do not backport this to `4.3.x`.

PR #16330, opened 2026-05-06
Deprecated features: prevent local configuration from being overridden by cluster peers

michaelklishin:
Much of this PR description is my earlier findings from #16324 with a few clarifying edits.

## Upgrading to `4.3.0`: Problem Definition

In 4.2.x, `transient_nonexcl_queues` is permitted by default, so its underlying feature flag stays disabled across the cluster.

During an upgrade to `4.3.0`, the deprecated feature phase changes to `denied-by-default`. Without an explicit override in the local `rabbitmq.conf` in place, every node figures
"this FF should be enable" (the deprecated feature should be denied) and that enabled state is replicated cluster-wide.

The deployment tooling adds an exclusion to `rabbitmq.conf` and restarts the node. But the FF state is already "enabled" on other cluster nodes, so after the restart, the state on the reconfigured `4.3.0` node is synced back, and the registry keeps the synced value.

In other words, once any node in the cluster computes the underlying [the deprecated feature] feature flag as enabled, the metadata sync sets it that way on every node, and node-specific configuration is silently ignored on subsequent restarts.

Re-creating the cluster and starting fresh with the config in place works because no node ever computed "enabled" for the underlying FF in the first place, and there's nothing to sync.

## An Available Workaround

Pre-configuring the `4.2.x` nodes to explicitly enable the deprecated features
that `4.3.0` disables by default and restarting all nodes so that they pick up
their updated configs should be enough of a workaround, and I will add it
to the release notes.

## A Low Risk Solution

It would be great to eliminate this less-than-obvious behavior.

An oversimplified solution is therefore:

1. Add a way for a node to determine whether a specific deprecated feature was explicitly configured locally
2. When the deprecated feature FFs syncs, if the corresponding deprecated feature is explicitly listed as permitted locally, use the local value and ignore the value received from the peer

## Notes for Reviewers

**Important**: note that you can only reproduce the problematic behavior
with a multi-node cluster.

I have edited the multi-line comments in tests several times and it's
never abundantly clear. I guess it's one of those issues where you need
to have the entire context, otherwise a single comment won't tell you
that much.

### DIFF
--- a/deps/rabbit/src/rabbit_deprecated_features.erl
+++ b/deps/rabbit/src/rabbit_deprecated_features.erl
@@ -119,6 +119,8 @@
          get_warning/1]).
 -export([extend_properties/2,
          should_be_permitted/2,
+         is_permit_explicitly_configured/1,
+         is_feature_in_use/2,
          enable_underlying_feature_flag_cb/1,
          list/1]).
 
@@ -510,7 +512,8 @@ should_be_permitted(FeatureName, FeatureProps) ->
         denied_by_default ->
             is_permitted_in_configuration(FeatureName, false);
         Phase ->
-            case is_permitted_in_configuration(FeatureName, false) of
+            Permitted = is_permitted_in_configuration(FeatureName, false),
+            case Permitted andalso should_log_warning(FeatureName) of
                 true ->
                     ?LOG_WARNING(
                        "Deprecated features: `~ts`: ~ts feature, it "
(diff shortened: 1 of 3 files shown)


dumbbell (commented):
This approach based on a local state that could be different on another node will lead to diferent feature flag states between members of a cluster. This is not possible.

I’m working on another approach at a higher level in the subsystem.

michaelklishin:
@dumbbell alright, I will close this PR then but keep the branch around just in case.

PR #16336, opened 2026-05-07
HTTP API: format empty proplists as empty JSON objects in a few more places

michaelklishin:
For the umpteenth time, we have to avoid the well known scenario where an empty proplist ends up being serialized as an empty JSON array where an empty JSON object is expected by the HTTP API clients.

For example, this can happen when a channel info endpoint is called before the connection had a chance to emit its first stats (metric samples).

Found by a few integration tests in the [HTTP API client for Zig](https://github.com/michaelklishin/rabbitmq-http-api-zig).


michaelklishin:
@mergifyio backport v4.3.x v4.2.x

michaelklishin:
Backported manually to `v4.3.x` for `4.3.1`.

michaelklishin:
Backported manually to `v4.2.x` for `4.2.7`.

PR #16341, opened 2026-05-07
Implement todo in stream reader

acogoluegnes:


deps/rabbitmq_stream/src/rabbit_stream_reader.erl
```diff
@@ -4241,3 +4233,55 @@ advertised_port_fun(tcp) ->
     port;
 advertised_port_fun(ssl) ->
     tls_port.
+
+check_node_connection_limit(undefined) ->
+    ok;
+check_node_connection_limit(RanchRef) ->
+    case application:get_env(rabbit, connection_max, infinity) of
```
the-mikedavis: There's some overlap with the discussion in https://github.com/rabbitmq/rabbitmq-server/pull/16299 here. We might want to have a protocol-specific connection max setting for streams. We could always make that change later though (i.e. have another config value that takes precedence over `connection_max`.

I'll re-run my connection script to add stream protocol connections and give it another run. I'm curious how an idle stream connection fares vs 0-9-1 and 1.0 connections.

the-mikedavis (approved):

deps/rabbitmq_stream/src/rabbit_stream_reader.erl
ansd: Good point. Should this line read from its own app? i.e. as follows:
```erl
application:get_env(rabbitmq_stream, connection_max, infinity)
```
This way we achieve protocol-specific connection max settings, right?

the-mikedavis: The results of the test are not so interesting. Idle connections seem like roughly the same impact between all three:

Screenshot 2026-05-07 at 12 16 48

michaelklishin:
Note a follow-up PR https://github.com/rabbitmq/rabbitmq-server/pull/16355.

PR #16342, opened 2026-05-07
Friendly TLS mismatch errors

mkuratczyk:
When someone mistakenly tried to connect without TLS
to a TLS-enabled listener or with TLS to a non-TLS
listener, we logged all kinds of confusing errors
(bad headers, unexpected messages, frame too large, etc).

Now we just log a user-friendly message


lhoguin (approved):
LGTM. I was wondering about log_level vs log_alert and log_alert is deprecated in favor of log_level so all is good.

PR #16343, opened 2026-05-07
Modernize for 4.4

mkuratczyk:
* require OTP27
* since we require OTP27, remove some code that was only required for older OTP version
* remove some compatibility code, mostly with 3.13
* don't use long-deprecated `string` module functions
* replace `catch` with `try ... catch`


deps/rabbit/src/rabbit_amqp1_0.erl
```diff
@@ -9,29 +9,6 @@
...
-emit_connection_info_local(Items, Ref, AggregatorPid) ->
-    ConnectionPids = list_local(),
-    rabbit_control_misc:emitting_map_with_exit_handler(
-      AggregatorPid,
-      Ref,
-      fun(Pid) ->
-              rabbit_amqp_reader:info(Pid, Items)
-      end,
-      ConnectionPids).
-
```
ansd, line 34: Should we now delete this module entirely?

ansd: Pushed a commit to delete this module.

ansd (approved):

PR #16367, opened 2026-05-09
Add `mqtt.max_connections` per-node connection limit

### LINKED ISSUE #16347: Improve connection limits, introduce per-protocol connection limits
This is an umbrella issue for a number of closely related discussions on [the limits RabbitMQ supports](https://www.rabbitmq.com/docs/limits) or should support with @the-mikedavis @ansd @acogoluegnes.

## Problem Definition

The VMware, Inc. part of the core team proposes reviewing our limits to eventually (likely before `4.4.0`) converge on the following limits design/structure/separation:

1. Per-user and per-virtual host apply to all protocols
2. `connection_max` covers core protocols (AMQP 1.0, AMQP 0-9-1)
3. `{plugin}.connection_max` covers each plugin protocol, including their Protocol-over-WebSockets variants, so `mqtt.connections_max` will cover both MQTT and MQTT-over-WebSockets

Note that most of this design is already in place but given the number of limits and protocols RabbitMQ supports, we need to review that nothing is missing, in particular for the per-user and per-virtual host limits.

## HTTP API

If we put the WebSockets-based protocols aside, the HTTP API is not really covered by the above limits and given the request-oriented nature of HTTP (even with persistent connections, even with HTTP/2, HTTP/3), the protective measure here is a Ranch-level connection limit introduced in https://github.com/rabbitmq/rabbitmq-server/pull/7753, https://github.com/rabbitmq/rabbitmq-server/issues/7593.

## Tanzu RabbitMQ Protocols

`connections_max` should cover AMQP 1.0-over-Websockets since its a WebSocket variant of the core protocols.

## Remaining Items as a TODO List

 * [x] Rename `stream.connection_max` to `stream.max_connections` (MK)
 * [x] Implement `mqtt.max_connections` (https://github.com/rabbitmq/rabbitmq-server/pull/16367)
 * [x] Implement `stomp.max_connections` (https://github.com/rabbitmq/rabbitmq-server/pull/16368)
 * [x] Ship [Cuttlefish `3.7.0`](https://github.com/Kyorai/cuttlefish/releases/tag/v3.7.0)
 * [x] Bump Cuttlefish in RabbitMQ to `3.7.0` (https://github.com/rabbitmq/rabbitmq-server/pull/16389)
 * [x] Alias `connection_max` in the core broker as `max_connections` (https://github.com/rabbitmq/rabbitmq-server/pull/16414)
 * [x] Investigate an optimal solution for the HTTP API (https://github.com/rabbitmq/rabbitmq-server/pull/16407)

lukebakken:
Closes #16347 (partial - adds `mqtt.max_connections`).

## What this PR does

Adds `mqtt.max_connections`, a per-node connection limit for the MQTT plugin. When the limit is reached, the broker sends a CONNACK with return code `not_authorized` (MQTT 3.x/4) or reason code `quota_exceeded` (MQTT 5) and closes the connection.

## How it works

The limit is checked in `rabbit_mqtt_processor:process_connect/5` as the first step in the CONNECT packet handler, before authentication. It does not operate at the Ranch transport layer, so the TCP listen queue is unaffected and Ranch continues accepting connections normally.

The active connection count uses `ets:info(persistent_term:get(?PG_SCOPE), size)`. The MQTT plugin creates a node-local PG scope in `rabbit_mqtt_sup`; each connection joins it via `pg:join/3` in `register_client_id/4`, using `{VHost, ClientId}` as the group key. Since the MQTT spec requires unique client IDs per vhost, which RabbitMQ enforces, each group has exactly one member. The PG scope ETS table therefore has one row per active connection node-wide, covering all listeners and transports (plain TCP, TLS, and Web MQTT). `ets:info/2` reads this count in O(1) - important given that MQTT nodes can host a very large number of connections.

The check runs before `register_client_id/4`, so the current connection is not yet counted; the limit comparison is `>= max_connections`.

## Configuration

```ini
mqtt.max_connections = 1000
```

The default is `infinity` (no limit). The setting accepts a non-negative integer or `infinity`.

## Testing

Adds `node_connection_limit` to the `limit` group in `auth_SUITE`. The test sets the limit to 0 via RPC so the first CONNECT attempt is rejected, covering both MQTT v4 and v5 via the existing `expected_connection_limit_error/1` helper.

### DIFF
--- a/deps/rabbitmq_web_mqtt/test/web_mqtt_command_SUITE.erl
+++ b/deps/rabbitmq_web_mqtt/test/web_mqtt_command_SUITE.erl
@@ -28,7 +28,8 @@ groups() ->
     [
      {unit, [], [merge_defaults]},
      {v5, [], [run,
-               user_property]}
+               user_property,
+               node_connection_limit_cross_transport]}
     ].
 
 suite() ->
@@ -171,6 +172,39 @@ user_property(BaseConfig) ->
            {user_property, UserProp}]],
          'Elixir.Enum':to_list(?COMMAND:run([<<"client_id">>, <<"user_property">>], Opts)))),
     ok = emqtt:disconnect(C).
+
+node_connection_limit_cross_transport(BaseConfig) ->
+    %% Wait connections from other tests to drain (`pg` cleanup is asynchronous).
+    rabbit_ct_helpers:eventually(
+      ?_assertEqual(
+         0,
+         length(rabbit_ct_broker_helpers:rpc(
+                  BaseConfig, 0, rabbit_mqtt, local_connection_pids, []))),
+      500, 60),
+    Config = [{websocket, true} | BaseConfig],
+    ok = rabbit_ct_broker_helpers:rpc(
+           BaseConfig, 0, application, set_env,
+           [rabbitmq_mqtt, max_connections, 2]),
+    try
+        C1 = connect(<<"plain-mqtt-client">>, BaseConfig, [{ack_timeout, 1}]),
+        C2 = connect(<<"web-mqtt-client">>, Config, [{ack_timeout, 1}]),
+        {ok, C3} = emqtt:start_link(
+                     [{host, "localhost"},
+                      {port, rabbit_ct_broker_helpers:get_node_config(
+                               Config, 0, tcp_port_web_mqtt)},
+                      {proto_ver, v5},
+                      {clientid, <<"web-mqtt-client-2">>},
+                      {ws_path, "/ws"},
+                      {ack_timeout, 1}]),
+        unlink(C3),
+        ?assertMatch({error, {quota_exceeded, _}}, emqtt:ws_connect(C3)),
+        ok = emqtt:disconnect(C1),
+        ok = emqtt:disconnect(C2)
+    after
+        ok = rabbit_ct_broker_helpers:rpc(
+               BaseConfig, 0, application, unset_env,
+               [rabbitmq_mqtt, max_connections])
+    end.
 
 start_amqp_connection(Type, Node, Port) ->
     amqp_connection:start(amqp_params(Type, Node, Port)).
--- a/deps/rabbitmq_mqtt/priv/schema/rabbitmq_mqtt.schema
+++ b/deps/rabbitmq_mqtt/priv/schema/rabbitmq_mqtt.schema
@@ -160,6 +160,19 @@ end}.
     {datatype, integer}
 ]}.
 
+{mapping, "mqtt.max_connections", "rabbitmq_mqtt.max_connections",
+    [{datatype, [{atom, infinity}, integer]}, {validators, ["non_negative_integer"]}]}.
+
+{translation, "rabbitmq_mqtt.max_connections",
+fun(Conf) ->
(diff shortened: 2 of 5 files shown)


ansd (commented):
If `mqtt.max_connections` is set, does the limit as proposed in your PR work correctly in any of the following cases:
* Port to Virtual Host Mapping: https://www.rabbitmq.com/docs/mqtt#port-to-virtual-host-mapping
* Dual stack (e.g. `0.0.0.0` for IPv4 and `::` for IPv6)
* A mix of MQTT and MQTTS connections

Because Ranch creates independent connection supervisors for every unique protocol, IP family, and port combination, `ranch:info(RanchRef)` only counts connections for that specific listener. In the scenarios above, the limit will multiply rather than act as a global node limit.

Additionally, the MQTT implementation in RabbitMQ is designed to scale to over 1 million concurrent connections per node. The connection limit check should be done as efficiently as possible.

To summarise, just following the same pattern for MQTT as done for Stream connections may not be sufficient. MQTT works differently because:
1. Connections can be mapped to vhosts by port.
2. MQTT use cases typically need to handle an order of magnitude more connections than Stream use cases.
 
My idea is to use the following to query the current number of MQTT connections on a given node:
```erl
ets:info(persistent_term:get(?PG_SCOPE), size).
```
This approximation should be good enough since the MQTT spec defines:
> Each Client connecting to the Server has a unique ClientID.

RabbitMQ enforces this. This solution is more correct globally and more efficient. What do you think?

ansd (approved):
Thank you @lukebakken !
Your new approach is simpler, more efficient, and more correct.

michaelklishin:
The test coverage of this feature is very limited, I will add a few more basic tests.

michaelklishin (approved):

deps/rabbitmq_web_mqtt/test/web_mqtt_command_SUITE.erl
```diff
@@ -172,6 +173,39 @@ user_property(BaseConfig) ->
...
+                      {ack_timeout, 1}]),
+        unlink(C3),
+        ?assertMatch({error, {quota_exceeded, _}}, emqtt:ws_connect(C3)),
+        ok = emqtt:disconnect(C1),
+        ok = emqtt:disconnect(C2)
+    after
+        ok = rabbit_ct_broker_helpers:rpc(
+               BaseConfig, 0, application, unset_env,
+               [rabbitmq_mqtt, max_connections])
+    end.
```
ansd, line 207: `web_mqtt_command_SUITE` is meant for testing CLI commands.
The new test case added is unrelated to CLI commands and should probably be moved somewhere else.

michaelklishin: I will submit a follow-up.

michaelklishin: https://github.com/rabbitmq/rabbitmq-server/pull/16405

ansd: Thank you

PR #16368, opened 2026-05-10
Add `stomp.max_connections` per-node connection limit

### LINKED ISSUE #16347: Improve connection limits, introduce per-protocol connection limits

lukebakken:
Closes #16347 (partial - adds `stomp.max_connections`).

## What this PR does

Adds `stomp.max_connections`, a per-node connection limit for the plain STOMP plugin, matching the existing `stream.max_connections` and `mqtt.max_connections` settings. When the limit is reached, the broker sends a STOMP ERROR frame and closes the connection.

## How it works

The enforcement follows the same application-level pattern used by `stream.max_connections` and `rabbit.connection_max` - it does not operate at the Ranch transport layer, so the TCP listen queue is unaffected and Ranch continues accepting connections normally.

`rabbit_stomp_processor:initial_state/2` is kept as a one-line wrapper delegating to a new `initial_state/3` that accepts the Ranch listener reference, so `rabbit_web_stomp_handler` (Web STOMP) requires no change and Web STOMP connections are never subject to the plain STOMP plugin limit.

Inside `process_connect/3`, `check_node_connection_limit/1` is the first check in the `maybe` chain. It calls `ranch:info(RanchRef)` to get `active_connections` for the specific listener and compares it against `application:get_env(rabbitmq_stomp, max_connections, infinity)`.

The limit is per-listener: TCP and TLS listeners are enforced independently.

## Configuration

```ini
stomp.max_connections = 1000
```

The default is `infinity` (no limit). The setting accepts a non-negative integer or `infinity`.

## Testing

Adds `node_connection_limit` to `connections_SUITE`. The test sets the limit to 0 via RPC, sends a CONNECT frame over a raw TCP socket, and asserts that an ERROR frame is received.

### DIFF
--- a/deps/rabbit/priv/schema/rabbit.schema
+++ b/deps/rabbit/priv/schema/rabbit.schema
@@ -3264,8 +3264,9 @@ fun(IpStr) ->
 end}.
 
 {validator, "non_negative_integer", "number should be greater or equal to zero",
-fun(Int) when is_integer(Int) ->
-    Int >= 0
+fun(infinity) -> true;
+   (Int) when is_integer(Int) -> Int >= 0;
+   (_) -> false
 end}.
 
 {validator, "non_zero_positive_integer", "number should be greater or equal to one",
--- a/deps/rabbitmq_stomp/include/rabbit_stomp.hrl
+++ b/deps/rabbitmq_stomp/include/rabbit_stomp.hrl
@@ -5,6 +5,10 @@
 %% Copyright (c) 2007-2026 Broadcom. All Rights Reserved. The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries. All rights reserved.
 %%
 
+-define(APP_NAME, rabbitmq_stomp).
+
+-define(PG_SCOPE, pg_scope_rabbitmq_stomp_connections).
+
 -record(stomp_configuration, {default_login,
                               default_passcode,
                               force_default_creds = false,
--- a/deps/rabbitmq_stomp/priv/schema/rabbitmq_stomp.schema
+++ b/deps/rabbitmq_stomp/priv/schema/rabbitmq_stomp.schema
@@ -157,6 +157,19 @@ end}.
     {datatype, integer}
 ]}.
 
+{mapping, "stomp.max_connections", "rabbitmq_stomp.max_connections",
+    [{datatype, [{atom, infinity}, integer]}, {validators, ["non_negative_integer"]}]}.
+
+{translation, "rabbitmq_stomp.max_connections",
+fun(Conf) ->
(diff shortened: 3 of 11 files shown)


michaelklishin:
I am extending this to be very similar to the final design of https://github.com/rabbitmq/rabbitmq-server/pull/16367.

michaelklishin:
Give that the delta in `main` and `v4.3.x` for STOMP is substantial, this will have to be an alternative implementation for `v4.3.x`, `v4.2.x` rather than a backport or an adapted backport.

PR #16378, opened 2026-05-11
Fix groovy-all dependency resolution

ansd:
Fix groovy-all dependency resolution in java_SUITE_data pom.xml Groovy 3.0+ packages `groovy-all` as a `pom` rather than a `jar`. This commit adds `<type>pom</type>` to the `groovy-all` plugin dependency in `deps/rabbitmq_mqtt/test/java_SUITE_data/pom.xml` to fix the resolution error introduced by the recent dependency version bump to 3.0.25 in https://github.com/rabbitmq/rabbitmq-server/pull/16363


michaelklishin:
@mergifyio backport v4.3.x v4.2.x

PR #16395, opened 2026-05-12
QQ: allow delayed-retry policies

### LINKED ISSUE #16394: Can't create policy with delayed settings from UI
### Describe the bug

I try to correct a new policy with the `delayed-retry` settings but it fails

Image

### Reproduction steps

1. Create new policy
2. Put a random name and pattern
3. Choose `Quorum queues`
4. Click on the `delayed-retry-type` and input `all`
5. Click on the `delayed-retry-min` and input any number value

### Expected behavior

The policy is created

### Additional context

_No response_

mkuratczyk:
We missed that in the original implementation,
even though the documentation says delayed-retry-* can be set through a policy. Now they can be indeed.

Fixes https://github.com/rabbitmq/rabbitmq-server/issues/16394

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -187,7 +187,7 @@
 
 -rabbit_boot_step(
    {?MODULE,
-    [{description, "QQ target group size policies. "
+    [{description, "QQ policies. "
       "target-group-size controls the targeted number of "
       "member nodes for the queue. If set, RabbitMQ will try to "
       "grow the queue members to the target size. "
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -198,18 +198,78 @@
...
      {requires, [rabbit_registry]},
      {enables, recovery}]}).
 
 validate_policy(Args) ->
-    Count = proplists:get_value(<<"target-group-size">>, Args, none),
-    case is_integer(Count) andalso Count > 0 of
-        true -> ok;
-        false -> {error, "~tp is not a valid qq target count value", [Count]}
+    case lists:foldl(fun ({Key, Value}, ok) -> validate_policy0(Key, Value);
+                         (_, Error)         -> Error
```
kjnilsson: redundant whitespace

kjnilsson (approved):
just a bit of formatting inconsistencies.

PR #16422, opened 2026-05-13
Prevent false-positive vhost deletion

mkuratczyk:
This issue came up during chaos testing. It's very unlikely in real life, but better safe than sorry.

## Why

In the unlikely scenario where Khepri Ra member restarts, the vhost projection can be empty / out-of-sync during recovery. Every 5 seconds, each vhost checks whether it should still exists (in case a vhost was deleted while the node was down) and it the checks happened while the projection is missing/empty/out-of-sync, the vhost concluded it was deleted and would shut down and delete its data (including CQ messages). In other words, a transient Khepri issue could lead to a vhost going down forever and CQ data being permanently deleted.

Repro:
```
$ make run-broker
> ra:stop_server(coordination, {rabbitmq_metadata, node()}), timer:sleep(6000), ra:restart_server(coordination, {rabbitmq_metadata, node()}).
```

Look for `Deleting message store directory for vhost...` in the log file.

## How

To prevent this issue, this PR adds a fallback check - if the vhost doesn't exist in the projection, we check with Khepri directly. If Khepri doesn't respond, we don't shutdown/delete the vhost (the check happens every 5s, so if it's a legitimate deletion, it will go away a bit later).

### DIFF
--- a/deps/rabbit/src/rabbit_db_vhost.erl
+++ b/deps/rabbit/src/rabbit_db_vhost.erl
@@ -65,9 +65,9 @@ create_or_get(VHostName, Limits, Metadata)
     case rabbit_khepri:create(Path, VHost) of
         ok ->
             {new, VHost};
-        {error, {khepri, mismatching_node,
-                 #{node_path := Path,
-                   node_props := #{data := ExistingVHost}}}} ->
+        {error, ?khepri_error(mismatching_node,
+                              #{node_path := Path,
+                                node_props := #{data := ExistingVHost}})} ->
             {existing, ExistingVHost};
         Error ->
             throw(Error)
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_db_vhost.erl
```diff
@@ -202,6 +203,32 @@ exists(VHostName) when is_binary(VHostName) ->
...
+%% not exist, or `{error, Reason}' if Khepri is unavailable.
+%%
+%% @private
+
+exists_uncached(VHostName) when is_binary(VHostName) ->
+    Path = khepri_vhost_path(VHostName),
+    case rabbit_khepri:get(Path) of
+        {ok, _} ->
+            true;
+        {error, {khepri, node_not_found, _}} ->
```
dumbbell: Could you please use the `?khepri_error()` macro exposed by Khepri?

Like:
```erlang
{error, ?khepri_error(node_not_found, _)} ->
```

dumbbell (changes requested):

deps/rabbit/src/rabbit_db_vhost.erl
mkuratczyk: I made this change for all error handing in that file for consistency

dumbbell (approved):
Thank you!

PR #16450, opened 2026-05-17
Fix blocked consumer status not returned by the HTTP API

razvanphp:
## Proposed Changes

While developing our OCPP plugin we got into a situation where the messages were not notified to the queue after delivery, so after exactly 200 messages, nothing was delivered, even tho there was a consumer connected.

https://github.com/rabbitmq/rabbitmq-server/blob/b5059a8492decf6de2acc2f665283725a70b286c/deps/rabbit/src/rabbit_queue_consumers.erl#L26

This PR adds the visibility so a consumer is correctly shown as blocked when this is the case. I also had to fix some caching layer conditions for this to work.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

Hope this will save someone's time, we wasted two days figuring out what was happening.

Thank you!
R

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue_process.erl
+++ b/deps/rabbit/src/rabbit_amqqueue_process.erl
@@ -440,7 +440,10 @@ code_change(_OldVsn, State, _Extra) ->
 %%----------------------------------------------------------------------------
 
 maybe_notify_decorators(false, State) -> State;
-maybe_notify_decorators(true,  State) -> notify_decorators(State), State.
+maybe_notify_decorators(true,  State) -> 
+    notify_decorators(State),
+    notify_consumers_state_changed(State),
+    State.
 
 notify_decorators_if_became_empty(WasEmpty, State) ->
     case (not WasEmpty) andalso is_empty(State) of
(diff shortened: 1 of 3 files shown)


razvanphp:
Fixed the tests, but should we add a new unit test case for blocked?

michaelklishin:
It depends on how straightforward and useful that new test would be. It's a balancing act.

We'll get to all outstanding PRs when `4.3.1` and `4.2.7` (and other patch releases) ship, in a couple of days or so.

michaelklishin (changes requested):
For workloads where consumers significantly outnumber publishers and consumer (channel) prefetch is low (10 in my tests), this
change introduces a 22-23% throughput (consumer delivery rate) drop.

I will share the `PerfTest` flags used in a bit but so far this behavior can be reproduced fairly easily.

michaelklishin:
The `PerfTest` flags used were these:

```shell
-u perf-common-C3 -p -x 1 -y 20 -q 10 -A 0 -s 100 -z 60 -i 5
```

michaelklishin:
`notify_consumers_state_changed/1` results in O(n) update complexity (where `n` is the number of consumers).

The consumer awareness of functions such as `block_consumer/2` and `unblock/2` is lost in this implementation, which forces an O(n) iteration.

michaelklishin:
I have a version that avoids new O(n) operations on the "consumer switching" path, will re-run all benchmarks and submit a PR in a bit.

michaelklishin:
Moved to #16532.

PR #16449, opened 2026-05-16
Add peer_cert_serial_number and fix validity

razvanphp:
## Proposed Changes

Fix the peer cert end date rendered as a tuple
X.509 encodes dates two different ways depending on whether the year fits in two digits:
- `UTCTime` — YYMMDDHHMMSSZ, used for years 1950–2049.
- `GeneralizedTime` — YYYYMMDDHHMMSSZ, used for 2050 and later.

Screenshot 2026-05-16 at 23 45 28

## Types of Changes

What types of changes does your code introduce to this project?

- [X] Bug fix (non-breaking change which fixes issue #NNNN)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
This is simply a reminder of what we are going to look for before merging your code._

- [ ] **Mandatory**: I (or my employer/client) have have signed the CA (see https://github.com/rabbitmq/cla)
- [ ] I have read the `CONTRIBUTING.md` document
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

## Further Comments

We run a fairly large mTLS fleet of IoT devices and the serial number is a must to be able to check the exact cert a device is using or if it rotated it's cert correctly. I'm sure it will come in handy for other users as well.

Screenshot 2026-05-17 at 00 30 15

PR #16460, opened 2026-05-18
Display peer (client) certificate serial number in management UI and select CLI commands

michaelklishin:
This is #16449 by @razvanphp with a man page fix and some (unit) tests for the introduced helper.


michaelklishin:
I will manually test the CLI commands just in case, their test coverage is not particularly extensive.

razvanphp:
Thank you very much for your time!

PR #16479, opened 2026-05-19
Reduce boot time

the-mikedavis:
This PR has two changes to cut cold boot time of a virgin node. (And I love cutting this time since it helps anyone running `make run-broker` iterate faster :slightly_smiling_face:.) On my machine this change reduces the cold boot time from 1827ms to 1489ms (saving 338ms, 18.5%) according to the startup banner.

1. Eagerly load modules in parallel with `code:ensure_modules_loaded/1` when scanning module attributes. Saves 250 ms on my machine.
2. Avoid loading `jose` when the auth mechanism is not OAuth2. Saves 88ms on my machine.

The commits have more details on each change.

### DIFF
--- a/deps/oauth2_client/Makefile
+++ b/deps/oauth2_client/Makefile
@@ -1,8 +1,8 @@
 PROJECT = oauth2_client
 PROJECT_DESCRIPTION = OAuth2 client from the RabbitMQ Project
 
-BUILD_DEPS = rabbit
-DEPS = rabbit_common jose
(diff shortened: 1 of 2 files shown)


michaelklishin:
This should also help all tests that start nodes, which is the majority of tests.

mkuratczyk:
For me it's roughly 1600ms -> 1450ms. Always happy to see this kind of improvements

michaelklishin (approved):

PR #16480, opened 2026-05-19
Reduce test time of config_schema_SUITEs

the-mikedavis:
There are two easy improvements we can make to cut the time it takes to run a `config_schema_SUITE` in any plugin. These tests currently take around 10 seconds on my machine but we can get them all down to around a second.

1. Reuse schemas for each snippet. Previously each snippet re-loaded all schemas. We can reuse the full schema for each snippet. `config_schema_SUITE` took linear time per snippet, now we can test in constant time.
2. Avoid spawning a broker. This was originally necessary with the way that schema generation worked, but since https://github.com/rabbitmq/rabbitmq-server/commit/cbfb11efe629ce919796adc6f6b614343739b9cf it hasn't been. We need to set the CWD for some tests which assume the CWD for paths for certificates, for example, but we can avoid spawning a broker at all.

### DIFF
--- a/deps/rabbit/test/config_schema_SUITE.erl
+++ b/deps/rabbit/test/config_schema_SUITE.erl
@@ -28,27 +28,14 @@ end_per_suite(Config) ->
     rabbit_ct_helpers:run_teardown_steps(Config).
 
 init_per_testcase(Testcase, Config) ->
-    rabbit_ct_helpers:testcase_started(Config, Testcase),
-    Config1 = rabbit_ct_helpers:set_config(Config, [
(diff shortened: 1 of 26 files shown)


michaelklishin:
@the-mikedavis thank you!

We will get to all pending PRs once `4.3.1` and `4.2.7` ship.

michaelklishin:
I can confirm that this speeds up `config_schema_SUITE`s from about 50% of the original execution time to whopping 34-35% of the time it takes in `main`.

Good job, @the-mikedavis!

Issue #16340, opened 2026-05-07
Stream Coordinator Performance and Availability Issue during Node Drain


kjnilsson:
## Summary

A user experienced a stream coordinator performance and availability issue during cluster maintenance. When placing a node into maintenance mode (`rabbitmq-upgrade drain`), stream leadership did not transfer as expected. Upon subsequently shutting down the drained node, the stream coordinator became unresponsive, causing streams to report as "down", and client applications to experience timeouts and failures. 

## Environment Details
- **Cluster Size**: 3 nodes
- **Workload**: ~2,000 streams (with ~1,500 being part of super streams, mostly 4 partitions each).
- **Connections**: ~20,000 active connections.
- **Node Specs**: Virtual Machines, 6 cores (~3600 Mhz), ~24 GB RAM.

## Steps to Reproduce
1. Establish a cluster with a significant number of streams, super streams, and active client connections.
2. Identify a node (`node1`) that is currently the stream leader for several streams.
3. Execute `rabbitmq-upgrade drain` on `node1`.
4. Observe that stream leadership does not transfer to other nodes.
5. Stop `node1`.

## Expected Behavior
- Executing the `drain` command should gracefully move the stream coordinator and stream leaders to non-maintenance nodes in an ordered manner.
- The stream coordinator should continue to function responsively during node maintenance and shutdown.

## Actual Behavior / Symptoms
1. **No Leadership Change on Drain**: Stream leaders remain on `node1` after `rabbitmq-upgrade drain` is executed. Re-election only occurs upon actual node shutdown.
2. **Coordinator Timeout**: After stopping `node1`, running `rabbitmq-streams coordinator_status` begins to return timeouts:
   ```
   %Protocol.UndefinedError{protocol: String.Chars, value: {:rabbit_stream_coordinator, :rabbit@node2}, description: ""}
   ```
3. **Followers Fall Behind**: `coordinator_status` reveals that followers significantly lag behind the leader's `LastLog Index`.
4. **Client Application Impact**: Thousands of clients attempting to reconnect and publish experience crashes and errors due to the coordinator being unreachable:
   ```
   AMQP close-reason, initiated by Peer, code=506, text='RESOURCE_ERROR - failed to deliver message: {error_fetching_locally_known_coordinator_members, {timeout, {rabbit_stream_coordinator, rabbit@node...'
   ```
[...]

Issue #3369, opened 2021-09-06
`rabbitmq-upgrade drain` exits with exit code 69 when it shouldn't


gerhard:
We keep seeing `FailedPreStopHook` when 3.9 LRE nodes get rolled, and we were able to track it down to the `rabbitmq-upgrade drain` command, specifically:

```
Exec lifecycle hook ([/bin/bash -c if [ ! -z "$(cat /etc/pod-info/skipPreStopChecks)" ]; then exit 0; fi; rabbitmq-upgrade await_online_quorum_plus_one -t 604800; rabbitmq-upgrade await_online_synchronized_mirror -t 604800; rabbitmq-up │
│ grade drain -t 604800]) for Container "rabbitmq" in Pod "rabbitmq-server-2_lre-3-9(338b4baf-0dea-4260-94a3-11a4dd2c613a)" failed - error: command '/bin/bash -c if [ ! -z "$(cat /etc/pod-info/skipPreStopChecks)" ]; then exit 0; fi; rabbitmq-upgrade await_on │
│ line_quorum_plus_one -t 604800; rabbitmq-upgrade await_online_synchronized_mirror -t 604800; rabbitmq-upgrade drain -t 604800' exited with 69:                                                                                                                   │
│ 15:16:59.517 [warn]  This node is being put into maintenance (drain) mode                                                                                                                                                                                        │
│                                                                                                                                                                                                                                                                  │
│ 15:16:59.521 [warn]  Suspended all listeners and will no longer accept client connections                                                                                                                                                                        │
│ Error:                                                                                                                                                                                                                                                           │
│ {:noproc, {:gen_server, :call, [#PID<11735.5319.0>, {:shutdown, 'Node was put into maintenance mode'}, :infinity]}}
```

We are not sure why these are happening, but even running the `drain` on a completely healthy node results in the same error code, even though the failure is different:

```
rabbitmq@rabbitmq-server-2:/$ rabbitmq-upgrade drain ; echo $?
Will put node rabbit@rabbitmq-server-2.rabbitmq-nodes.lre-3-9 into maintenance mode. The node will no longer serve any client traffic!

15:35:00.320 [warn]  This node is being put into maintenance (drain) mode

15:35:00.323 [warn]  Suspended all listeners and will no longer accept client connections
Error:
{:channel_termination_timeout, {:gen_server, :call, [#PID<11702.1403.0>, {:shutdown, 'Node was put into maintenance mode'}, :infinity]}}
69
```

[...]

PR #16515, opened 2026-05-22
Stream coordinator: v8 optimisations and bug fixes

kjnilsson:
## Overview

The stream coordinator is a single Ra (Raft) state machine: every command runs on all replicas and blocks the machine while it executes. Several `apply/3` paths scaled with the _total_ number of streams and single-active-consumer (SAC) groups in the cluster rather than with the work an event actually implies. Because a node going down or rejoining delivers one command per affected connection/member, these O(total) paths compounded into roughly O(connections × groups) — CPU-starving the state machine and stalling unrelated commands (such as electing new stream leaders) for many minutes.
Separately, the leader executedstopping/deleting members) in away that let a single mass event spawn an unbounded number of worker processes and flood the state machine with feedback commands.

This PR bumps the coordinator mnd, gated on v8:

- makes the hot `apply/3` paths proportional to the footprint of what actually changed;
- reworks how the leader executes actions so a mass event cannot overwhelm it (state-machine-owned retry backoff + bounded worker concurrency);
- restructures the module so older machine versions run a frozen copy of the previous implementation;
- fixes a set of writer-liveness gaps, removes a wall-clock non-determinism, and improves leadership transfer during maintenance.

All behavioural changes are gated on machine version 8. Machine versions up to 7 are now served by a frozen, verbatim copy of the previous implementation
(`rabbit_stream_coordinator_v7`er runs at v8. During a rollingupgrade the effective version stays at the cluster minimum, so mixed old/new nodes run the frozen v7 code and stay in agreement; the new behaviour activates only once the whole cluster is at v8. Log replay is unaffected.

### Performance / scalability

- **Connection DOWN / node-discoved from O(connections ×groups) to proportional to affected groups only
- **`connection_reconnected`** ups) to O(connection footprint)
- **Node reconnection (`nodeup`)** — stream side improved from O(total streams) to only streams with members on the returning node; SAC side improved from O(total groups) similarly

Benchmarks show dramatic improvected` improves from 69.7 ms to2.8 ms at 50k groups; `nodeup` stream side from 44 ms to 3.4 ms.

Representative `apply/3` latencies per event, before → after:

| Operation (per event)              | Groups / streams | Before   | After   | Speed-up  |
| ---------------------------------- | ---------------: | -------: | ------: | --------: |
| SAC `connection_reconnected`       |           10,000 |    13 ms |  0.5 ms |      ~26× |
| SAC `connection_reconnected`       |           50,000 |  69.7 ms |  2.8 ms |      ~25× |
| Stream `nodeup` (apply)            |           10,000 |   7.7 ms | 0.34 ms |      ~23× |
| Stream `nodeup` (apply)            |           50,000 |    44 ms |  3.4 ms |      ~13× |
| SAC `handle_node_reconnected`      |           10,000 |  2.33 ms | 0.48 µs |    ~4,900× |
| SAC `handle_node_reconnected`      |           50,000 |  11.5 ms | 0.42 µs |   ~27,000× |

(The `handle_node_reconnected` path drops to microseconds because at v8 the connection is kept in `pids_groups` across a disconnect, so reconnection avoids the full group scan/merge entirely.)

Notes:
- All figures are single-command apply latency (the state machine blocks for this duration), so the reduction directly translates to less time the coordinator is unavailable to unrelated commands like leader elections.
- The handle_node_reconnected speed-ups are so large because the old path was O(total groups) and the new one is essentially O(1) for the reconnect (units cross from ms to µs) — I'd keep the "~4,900×/~27,000×" framing since the raw ratio looks almost too good otherwise.

### Action execution under mass events (v8-gated)

A node hosting members for many streams going down previously made the leader (a) spawn an unbounded number of action worker processes at once and (b) sleep 5–10s inside each worker before retrying a failed action, holding the slot for the whole backoff and firing a thundering herd of feedback commands back into the single state machine.

- **Retry backoff moved into the state machine** — workers no longer sleep. A failed member action parks the member as `{sleeping, RetryAt}` and is re-driven by a single, coalesced `retry_reconcile` timer ordered by next-retry time (a `gb_trees`); the timer is armed dynamically to the earliest retry (no polling). Workers return immediately, freeing their slot. Backoff durations derive from command metadata, keeping `apply/3`
deterministic.
- **Bounded action concurrency** — the leader runs at most N action workers at once (default 64, overridable via the `stream_coordinator_max_concurrent_actions` application environment key). Excess actions are queued in the (leader-local, non-replicated) aux
state and started FIFO as runniued action whose stream wasdeleted while it waited is dropped rather than started, so it cannot orphan an osiris member. The pending queue is rebuilt from the state machine on a leader change, so losing it is safe.

### Module structure

Following the `rabbit_fifo` / `rabbit_fifo_v7` convention, the previous implementation is frozen verbatim as `rabbit_stream_coordinator_v7` and `which_module/1` routes machine
versions ≤ 7 to it. The live `ris therefore v8-only: itsper-version `?V*_OR_MORE` guards and now-unreachable pre-v8 clauses have been removed, leaving a single current code path (the `machine_version/3` conversion chain is kept,
since upgrades are applied on tmigrated across the boundarystays tuple-compatible, remote rpc/erpc targets resolve on old-release nodes that only have the `rabbit_stream_coordinator` module, and the transient aux record is version-tagged and upgraded in place after a machine version change (Ra does not re-initialise aux state).

### Writer liveness fixes (v8-gated)

- Level-triggered writer recovery prevents streams from settling leaderless
- Stale start latches cleared t
- `restart_stream` no longer resurrects deleting members
- Disconnected replica resume on `nodeup`
- Fixed `delete_member` action shape

### Determinism

The consumer timestamp (and the action retry backoff) are sourced from command metadata instead of wall-clock reads, ensuring byte-deterministic replay.

### Maintenance

Leadership transfer for local sdrain prevents leaderless
streams.

### Testing

Coordinator unit suite (35 casete (73 cases) pass; `dialyze`and `xref` are clean. New coverkoff park/reconcile cycle, thebounded-concurrency queue and drop, and the cross-versionaux-record upgrade.


kjnilsson:
[rabbitmq-server-windows-4.2.0+beta.4.773.gcaadad0.zip](https://github.com/user-attachments/files/28140809/rabbitmq-server-windows-4.2.0%2Bbeta.4.773.gcaadad0.zip)

windows release artefact for user testing.

kjnilsson:
https://github.com/rabbitmq/rabbitmq-server/issues/16340

This is the primary issue. This PR may not address everything.

mkuratczyk (approved):
I've played around with the current state of this PR and can confirm significant speed ups and

acogoluegnes (approved):

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1632,7 +1632,40 @@ drain(TransferCandidates) ->
...
+                  {ok, PreferredNode} ->
+                      {_, Ref} = spawn_monitor(fun() ->
+                          rabbit_stream_coordinator:restart_stream(Q, #{preferred_leader_node => PreferredNode})
+                      end),
+                      Ref;
+                  undefined ->
+                      undefined
+              end
+          end || Q <- Queues],
+         [receive {'DOWN', Ref, process, _, _} -> ok end || Ref <- Refs, Ref =/= undefined],
```
lukebakken: This `receive` has no `after` clause and no `demonitor`. The queue-type drain runs inside `safe_maintenance_step`, which catches exceptions but not blocking, so a stalled `restart_stream` worker stalls the whole drain (the drain-must-not-hang contract from #3369). Normally each worker is bounded by `process_command`'s per-server `cmd_timeout` retries, but if the local coordinator is down, `ensure_coordinator_started` calls `global:set_lock/1`, which retries with `infinity` and can block with no upper bound. Suggest bounding it: `after Timeout -> erlang:demonitor(Ref, [flush]), ok`.

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1632,7 +1632,40 @@ drain(TransferCandidates) ->
...
+         timer:sleep(1000)
```
lukebakken, line 1668: `timer:sleep(1000)` runs for every chunk, including the last one and chunks where no transfer happened (all `Refs` were `undefined`). Consider sleeping only between chunks that actually spawned workers.

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1632,7 +1632,40 @@ drain(TransferCandidates) ->
...
+     end || Queues <- QueuesChunked],
+    ok.
+
+list_local_leaders() ->
+    [Q || Q <- rabbit_amqqueue:list(),
```
lukebakken: `rabbit_amqqueue:list/0` does `ets:tab2list` over all queues of every type. `rabbit_amqqueue:list_by_type(rabbit_stream_queue)` uses `ets:match_object` scoped to streams, materialising far fewer records on clusters with many classic/quorum queues, and makes the `?amqqueue_type_is(Q, ?MODULE)` guard redundant.

lukebakken (commented):
A few notes on the new stream drain leadership-transfer path.

deps/rabbit/src/rabbit_stream_queue.erl
kjnilsson: how would you only sleep between chunks that spawns workers without massively overcomplicating the code?

deps/rabbit/src/rabbit_stream_queue.erl
kjnilsson: will fix, all those list functions are terrible really. been on my list to address at some point.

PR #16532, opened 2026-05-26
By @razvanphp: Fix blocked consumer status not returned by the HTTP API

michaelklishin:
This is #16450 by @razvanphp with an optimization that avoids an O(n) consumer state change that results in a 20-25% consumer throughput drop on the workloads where consumers with low prefetch significantly outnumber publishers.

Below are the `PerfTest` parameters that simulate the specific workload where I could spot a regression:

```
--queue pr-16450-workload-3 \
--predeclared \
--producers 1 \
--consumers 20 \
--qos 10 \
--multi-ack-every 0 \
--size 100 \
--time 180 \
--interval 5
```


michaelklishin:
Something very interesting that I have noticed about this PR: while the original goal was a consumer status reporting bug fix, and my follow-up was to eliminate an O(n) penalty on consumer state updates, I also observe something unexpected.

The workload involves a very fast publisher and N competing consumers, so it's a fairly specialized workload, not one of the most common ones.

The latency and throughput vary within, say, 2-7% both ways depending on what else is going on on this host [1]. However, the end-to-end latency reported by PerfTest is x5-10 times lower (yes, five to ten times, this is not a typo).

I have run the benchmarks twice now (30 and 60 minutes each, without interruptions), used 4 workflows, compared flame graph data (no new spikes/long lines/outliers) and this behavior seems to be reproducible.

So something that I do not fully understand right now improves the publisher/consumer operation balance of the queue leader process when there are multiple consumers. Or, put it differently, it improves consumer bias in that workflow.

1. `PerfTest` is JVM-based and thus cannot avoid the stop-the-world GC, so this variability is not entirely unusual

michaelklishin:
Ah, with a non-flooding publisher the difference drops to zero. So yeah, better behavior under peak ingress rate. Not bad for a bug fix :)

razvanphp:
Thank you for your time spent on this! 🙂

PR #16556, opened 2026-05-28
Verify Prometheus endpoint OAuth2 authentication

MarcialRosales:
Add Selenium tests that confirm the /metrics endpoint correctly accepts OAuth2 Bearer tokens from clients with the monitoring tag and rejects tokens from clients that only have the management tag.

Also configure two new Keycloak service account clients in the test realm and add the rabbitmq.tag:monitoring client scope.

Fix npm install in Dockerfile by adding --ignore-scripts to avoid the chromedriver ESM/CJS incompatibility with proxy-agent on Node.js 22.

## Types of Changes

- [x] Build system and/or CI


michaelklishin:
Damn, I have only now noticed a requested review from @Zerpet. Sorry. FWIW all tests did pass and from what I understand this work is relevant to reducing the Selenium test suite flakiness we have seen recently.

PR #16557, opened 2026-05-28
Add validation for consumer timeout arguments

mkuratczyk:
Validate 'x-consumer-timeout' and 'x-consumer-disconnected-timeout' when declaring a queue.


michaelklishin:
The newly added tests may seem a bit silly but we did not have any direct coverage for that key, so I figured I should add them as part of the review process.

PR #16560, opened 2026-05-28
Fix prodkeycloak-proxy command invocation

fallintoplace:
## Summary
- restore the `gen-httpd-conf` invocation in `prodkeycloak-proxy`

## Why
- the current line merges the script path and first argument into the command name
- that makes Bash try to execute a non-existent path before the container starts
- this restores the known-good form of the command

## Validation
- `bash -n selenium/bin/components/prodkeycloak-proxy`
- `git diff --check`

### DIFF
--- a/selenium/bin/components/prodkeycloak-proxy
+++ b/selenium/bin/components/prodkeycloak-proxy
@@ -31,7 +31,7 @@ start_prodkeycloak-proxy() {
   MOUNT_HTTPD_CONFIG_DIR=$CONF_DIR/httpd
 
   mkdir -p $MOUNT_HTTPD_CONFIG_DIR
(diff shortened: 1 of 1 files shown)


michaelklishin:
@fallintoplace I appreciate your time spent on this but even though you did provide some "validation" steps, you did not explain why this change is necessary in the first place.

Why should we accept this PR?

michaelklishin:
So apparently this is not valid shell script for `bash` and it comes from a larger refactoring in 8ab8d8c46f. It happens to work but would not pass any lint (like `shellcheck`).

michaelklishin:
The affected long lived branches are `main, `v4.3.x`, `v4.2.x`.

PR #16570, opened 2026-05-30
Selenium: improve logging in attempts to understand/fix a CI failure

michaelklishin:


michaelklishin:
Merged because the Selenium-Prometheus job in question does not run for PRs.

PR #16583, opened 2026-06-01
QQ: v9

kjnilsson:
Summary

This branch lays the groundwork for RabbitMQ's next quorum queue release (rabbit_fifo v9) and adds four related features:

- Deferred message tokens over AMQP 1.0 — consumers can park a message under a client-chosen token (via the MODIFIED outcome's x-opt-deferral-token + x-opt-delivery-time annotations) and later pull it back on demand by listing tokens in a FLOW frame's rabbitmq:deferred-tokens property, instead of waiting on a delayed-retry timer. Support is advertised via the rabbitmq:deferred-tokens link capability and documented in deps/rabbit/DEFERRED.md
- Per-node ingress byte tracking — rabbit_fifo now tracks enqueue bytes per node in the replicated state machine, exposes cumulative totals via overview/1, and computes smoothed per-node ingress rates (via ra_li leaky integrators) in aux state.
- **Message delays** — `rabbit_quorum_queue:deliver/3` stamps a `dt` annotation on a message when the client's `x-opt-delivery-time` is in the future; `rabbit_fifo`'s enqueue path then routes it straight into the delayed bucket instead of the ready queue, so it only becomes deliverable once its delivery time has passed, mirroring how redelivery already works. Priorities are ignored for delayed messages, which are enqueued onto the returns queue for performance and simplicity.

Details

- rabbit_fifo version is bumped to 9, with the prior v8 implementation preserved as rabbit_fifo_v8 (new module) so that v7 -> v8 conversion logic can be split out cleanly; v8 -> v9 conversion is currently a no-op
- rabbit_queue_type gains a callback, implemented for quorum queues via rabbit_fifo_client
- A deferral token is only hsets an explicitx-opt-delivery-time — the delayed-retry path never creates a deferred entry, keeping tokens purely client-driven
- Aux state version bumped to aux_v5 with an upgrade clause for rolling restarts;
ingress decay time is config


mkuratczyk:
Review of the work in progress. The interactions with message priorities were discussed  in the past, but flagging them in case we want to reconsider.

1. There was a state conversion missing: I've pushed a commit that fixes it.
2. Followers log an unhandled message: `[debug] <0.4627.0> queue 'omq-0' in vhost '/': follower received unhandled msg: machine_timeout`
3. Ripe messages (delayed messages that reached their delivery time) are effectively treated as the highest priority - if I have a backlog of ripe messages, consumers will be receiving them exclusively until that backlog is drained. All other messages, including those published with a high priority, are queued behind ripe messages, regardless of their priority
4. ~Messages of different priorities scheduled for the same time are delivered ignoring their priorities. However, if the broker is then restarted, the priorities become effective (because at the time of recovery, the messages are no longer delayed so normal priority rules apply)~ (I must have botched the test - can't reproduce this)

kjnilsson:
> Review of the work in progress. The interactions with message priorities were discussed in the past, but flagging them in case we want to reconsider.
> 
> 1. There was a state conversion missing: I've pushed a commit that fixes it.

Thank you.

> 2. Followers log an unhandled message: `[debug] <0.4627.0> queue 'omq-0' in vhost '/': follower received unhandled msg: machine_timeout`

This is harmless and should only happen after a leader change.

> 3. Ripe messages (delayed messages that reached their delivery time) are effectively treated as the highest priority - if I have a backlog of ripe messages, consumers will be receiving them exclusively until that backlog is drained. All other messages, including those published with a high priority, are queued behind ripe messages, regardless of their priority

Yes correct, QQs treat delayed messages as "returns" i.e. a buffer that sits in front of the priority queue. The reason for this is partly not to regress on non functional properties, in particular, snapshotting performance as taking the priority into account could enqueue messages out of raft index order, nullifying some important optimisations in `rabbit_fifo:live_indexes/1`.

I suggest we keep the current behaviour and make sure it is documented. It does meant that you can't practically combine priorities with delays but I think this is preferrable to regress performance wise. We can always revisit this behaviour in a future version of QQs should someone come up with a convincing need for this.

> 4. Messages of different priorities scheduled for the same time are delivered ignoring their priorities. However, if the broker is then restarted, the priorities become effective (because at the time of recovery, the messages are no longer delayed so normal priority rules apply)

This should really not be the case as that would be non deterministic.

lukebakken (commented):
> [!NOTE]
> This review was drafted by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who provided significant guidance throughout and reviewed it before posting. The findings were verified against the qq-vNext code.

Reviewed the deferred-message, message-delay, and per-node ingress work. Notes are inline; most are confirm-intended questions. The two I would prioritise are the `ingress_bytes_by_node` leak (`all_nodes/1`) and the past-TTL claim delivery (`assign_claimed/6`).

One finding cannot be anchored inline because it is on a line the PR does not change. In `deps/rabbit/src/rabbit_fifo.erl`, `query_single_active_consumer/1` (the `is_tuple(MaybeV7)` clause) is, after the versioned_query change, the only query still on the `is_tuple -> rabbit_fifo_v7` fallback. It works during a v8 to v9 upgrade only because v7's `active_consumer/1` crashes on the v8 `#consumer` shape, raising the `function_clause` that reaches `versioned_query`; if those shapes ever converge it would return wrong data instead. A `#?STATE{} -> disabled` clause like the sibling queries would make it raise directly.

Companion PR with two fixes found during this review (the untracked/stateless enqueue path ignoring `x-opt-delivery-time`, plus a documentation comment): #17288.

deps/rabbit/src/rabbit_amqp_session.erl
```diff
@@ -3361,32 +3378,78 @@ handle_outgoing_link_flow_control(
...
+                                  T
+                          end,
             Link = Link0#outgoing_link{
                      stashed_credit_req = #credit_req{
                                              delivery_count = DeliveryCountRcv,
                                              credit = LinkCreditRcv,
                                              drain = Drain,
-                                             echo = Echo}},
+                                             echo = Echo,
+                                             tokens = PrevTokens ++ parse_deferred_tokens(FlowProps)}},
```
lukebakken, line 3407: Each FLOW is capped at 256 tokens, but stashed tokens accumulate across FLOWs here (`PrevTokens ++ ...`) with no aggregate cap, so a client pipelining FLOWs within one credit round-trip can build an oversized single `assign_deferred`. Worth an aggregate cap, or is the round-trip window a sufficient bound?

deps/rabbit/src/rabbit_amqp_session.erl
```diff
@@ -1989,17 +1988,26 @@ pop_credit_req(
...
   S0 = #state{cfg = #cfg{max_queue_credit = MaxQueueCredit},
               outgoing_links = OutgoingLinks,
               queue_states = QStates0}) ->
     LinkCreditSnd = amqp10_util:link_credit_snd(
                       DeliveryCountRcv, LinkCreditRcv, CDeliveryCount),
     CappedCredit = cap_credit(LinkCreditSnd, MaxQueueCredit),
-    {ok, QStates, Actions} = rabbit_queue_type:credit(
-                               QName, Ctag, QDeliveryCount,
-                               CappedCredit, Drain, QStates0),
+    %% see handle_outgoing_link_flow_control/2 for why the claim goes first
```
lukebakken: This claim-then-credit sequence is duplicated near-verbatim in `handle_outgoing_link_flow_control/2`, kept in sync by comment only. Might be worth extracting a small helper so the ordering invariant lives in one place.

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -523,35 +540,44 @@ apply_(#{system_time := Ts} = Meta, #delayed_cmd{op = {retry, Mode}},
...
-        #consumer{status = Status} when Status =/= up ->
-            {State0, {error, consumer_not_active}, []};
-        #consumer{credit = Credit} ->
-            {Msgs, NotFound, Delayed} = take_deferred(Tokens, Delayed0),
-            NumMsgs = length(Msgs),
-            case NumMsgs > Credit of
-                true ->
-                    Err = {insufficient_credit, Credit, NumMsgs},
-                    {State0, {error, Err}, []};
+apply_(Meta, #delayed_cmd{op = {assign_deferred, ConsumerKey, Tokens}},
```
lukebakken, line 547: Claiming for a consumer that is in the map but not `up` moves its tokens off the shared deferred map onto it, and `deliver_claimed` only serves `up` consumers, so the message is unclaimable by any other consumer until this one recovers or the delivery time passes. It self-heals, but it drops the old `{error, consumer_not_active}` signal to the client. Intended?

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -696,9 +731,12 @@ apply_(Meta, {nodeup, Node}, #?STATE{consumers = Cons0,
 apply_(_, {nodedown, _Node}, State) ->
     {State, ok};
 apply_(Meta, #purge_nodes{nodes = Nodes}, State0) ->
-    {State, Effects} = lists:foldl(fun(Node, {S, E}) ->
+    {State1, Effects} = lists:foldl(fun(Node, {S, E}) ->
                                            purge_node(Meta, Node, S, E)
                                    end, {State0, []}, Nodes),
+    State = State1#?STATE{
+        ingress_bytes_by_node =
+            maps:without(Nodes, State1#?STATE.ingress_bytes_by_node)},
```
lukebakken, line 746: `ingress_bytes_by_node` is pruned only here via `#purge_nodes`, and the stale-node set is `all_nodes/1` (see `rabbit_quorum_queue:handle_tick`), which reports only nodes with a live consumer/enqueuer/waiting entry. A node that published and then lost its enqueuer drops out of `all_nodes`, so once it leaves the cluster it never appears in `Stale` and its entry leaks here permanently. Seeding `all_nodes/1` with `maps:keys(ingress_bytes_by_node)` would let this `maps:without` reach it. Low severity (keys are cluster nodes), but unbounded across repeated permanent removals.

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -2511,8 +2664,16 @@ evaluate_limit0(Index, BeforeState,
                 Effects0) ->
     case is_over_limit(State0) of
         true when Strategy == drop_head ->
-            {State, Effects} = drop_head(State0, Effects0),
-            evaluate_limit0(Index, BeforeState, State, Effects);
+            case drop_head(State0, Effects0) of
+                {State0, Effects} ->
```
lukebakken, line 2686: This match is cheap only because `drop_head`'s no-op branch returns the identical `State0` term (pointer equality). A future refactor that rebuilt the record on that no-op path would silently turn this into a full-state deep compare on every over-limit command. An explicit `no_progress` return would be robust against that.

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -2655,81 +2818,214 @@ take_delayed_for_retry(N, Ts, #delayed{tree = Tree0,
...
-              next_msg_id = NextMsgId0,
-              credit = Credit0,
-              delivery_count = DelCnt0,
-              cfg = Cfg} = Con0 = maps:get(ConsumerKey, Consumers0),
+%% Take up to Credit of the messages this consumer has claimed, oldest first
+%% within each token, dropping any claimed key whose message has since left
+%% the delayed tree.
+take_claimed_up_to(Credit, #consumer{deferred_claims = Claims0} = Con,
+                   #?STATE{delayed = #delayed{tree = Tree0} = Delayed} = State) ->
+    {Msgs, Claims, Tree} = drain_claims(maps:iterator(Claims0, ordered),
```
lukebakken, line 3030: When a consumer claims several tokens at once, `maps:iterator(_, ordered)` delivers them in token-binary order, not enqueue order (oldest-first only within a single token). Deterministic across replicas, so not a divergence; just confirming the cross-token order is meant to be unspecified.

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -2655,81 +2818,214 @@ take_delayed_for_retry(N, Ts, #delayed{tree = Tree0,
...
+            drain_claimed_keys(Rem, N - 1, gb_trees:delete(Key, Tree0),
+                               [Msg | Acc]);
+        none ->
+            %% the message was promoted by its delivery time, retried or
+            %% purged, so drop the stale key
+            drain_claimed_keys(Rem, N, Tree0, Acc)
+    end.
+
+%% Check a batch of claimed messages out to their consumer.
+assign_claimed(#{system_time := Ts} = Meta, ConsumerKey,
```
lukebakken, line 3077: `assign_claimed/6` checks messages out with no `msg_is_expired/2` gate, unlike the normal path (`checkout_one` -> `expire_msgs`, and `expire_shallow` on promotion). If a message's TTL is shorter than its deferral delay, a claim delivers it past its expiry instead of dead-lettering it as `expired`. Intended (an explicit claim overrides TTL), or should expired claims be dead-lettered?

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1348,7 +1350,8 @@ deliver0(QName, Correlation, Msg, QState0) ->
 
 deliver(QSs, Msg0, Options) ->
     Correlation = maps:get(correlation, Options, undefined),
-    Msg = mc:prepare(store, Msg0),
+    Msg1 = annotate_delivery_time(Msg0),
```
lukebakken, line 1359: This runs `mc:x_header` per message for all quorum-queue traffic, not just delayed messages. Assuming that is the intended trade (parse once here so replicas read a cheap annotation) - flagging in case the per-publish cost on the common path is worth avoiding.

kjnilsson: yes that is the trade, but honestly - how would you even know it isn't a delayed message wihtout checking the x_header?

deps/rabbit/src/rabbit_fifo.erl
kjnilsson: Added a comment. The strictly correctly order would have been the order of tokens received by assing deferred. It is not at this point worth bothering to do a large refactory. Most assign deferrals will be single token / single message I am sure.

deps/rabbit/src/rabbit_fifo.erl
kjnilsson: I am sure if it changed in the future some AI would catch it. Let's keep it simple for now.

deps/rabbit/src/rabbit_fifo.erl
kjnilsson: Well let's hope that no users are unwise enough to routinely change the cluster membership just to, e.g. run upgrades....  ;)

Fixed...

deps/rabbit/src/rabbit_fifo.erl
kjnilsson: intended, it would be odd if ttl removed messages that are explicitly claimed.

deps/rabbit/src/rabbit_fifo.erl
kjnilsson: this is fine yes

deps/rabbit/src/rabbit_amqp_session.erl
kjnilsson: done

deps/rabbit/src/rabbit_fifo.erl
kjnilsson: Addressed something adjacent to this issue which I think is the actual issue. I.e. returning claimes when other consumer messages would have been returned also (consumer lock timeout etc).

PR #16584, opened 2026-06-02
Minor refactoring of two HTTP API areas

michaelklishin:


michaelklishin:
Reverting in favor of another PR by @lhoguin.

PR #16590, opened 2026-06-02
Fix a `trie_binding_to_dest_with_keep_while_cond`-specific `function_clause` on data-less Khepri tree nodes

michaelklishin:
The callback in question walked exchanges and bindings with `?KHEPRI_WILDCARD_STAR`, which also matches intermediate tree nodes that carry no payload, e.g. orphan `exchanges/{name}` nodes left behind by stale child bindings.

Adding `#if_has_data{}` as a condition to
skip them follows the convention used in `rabbit_db_exchange` and `rabbit_db_binding`.

References #14902 #15954.
Closes #16587.

Note that `v4.2.x` is not affected.


michaelklishin:
Converting to a draft to

1. Fix the test failures
2. Add `rabbitmqctl set_topic_permissions` validation that we suspect is one of the ways to hit #16587

deps/rabbit/src/rabbit_db_queue.erl
```diff
@@ -1040,12 +1041,13 @@ tie_binding_to_dest_with_keep_while_cond_enable(
...
                        [FeatureName],
                        #{domain => ?RMQLOG_DOMAIN_DB}),
+                    %% Same `#if_has_data{}' guard as above.
                     BindingPattern = rabbit_db_binding:khepri_route_path(
                                        ?KHEPRI_WILDCARD_STAR,
                                        ?KHEPRI_WILDCARD_STAR,
                                        ?KHEPRI_WILDCARD_STAR,
                                        ?KHEPRI_WILDCARD_STAR,
-                                       ?KHEPRI_WILDCARD_STAR),
+                                       #if_has_data{}),
```
dumbbell, line 1050: Did you identify situations where a binding tree node can exist without data?

michaelklishin:
I have pushed some extra CLI-level validations that would prevent one identified scenario in #16587 much earlier and even without the rest of this fix.

deps/rabbit/src/rabbit_db_queue.erl
michaelklishin: @dumbbell after a fair amount of digging I have arrived to the following conclusion: the only scenarios that seem theorically possible would involve custom plugins invoking code paths that regular clients would not be able to reproduce

michaelklishin:
I could not find any additional scenarios that seem feasible with just clients and CLI operations that do not involve `rabbitmqctl eval`.

Given that this is a matter of upgrade safety for `4.3.x`, we have meaningful tests and CI is green, I'd like to merge it for `4.3.2`.

dumbbell (changes requested):
Could you please squash commits 2 and 3 into the first one?

michaelklishin:
The forced push was a squash of the first 3 commits out of 4 on this branch.

PR #16596, opened 2026-06-03
Await stream sync on await_online_quorum_plus_one

mkuratczyk:
When calling `rabbitmq-upgrade await_online_quorum_plus_one`, wait for the stream replicas to not only be running, but also to be sufficiently up to date (max 10s behind right now). Otherwise we may restart the next node in a rolling restart process and end up with a degraded stream.

Additionally, report the replication lag in `rabbitmq-streams stream_status`

### DIFF
--- a/deps/rabbit/src/rabbit_stream_coordinator.erl
+++ b/deps/rabbit/src/rabbit_stream_coordinator.erl
@@ -56,7 +56,9 @@
          ra_local_query/1]).
 
 -export([log_overview/1,
-         key_metrics_rpc/1]).
+         key_metrics_rpc/1,
+         is_replica_fresh/2,
+         is_replication_stale/2]).
 
 %% for SAC coordinator
 -export([sac_state/1,
@@ -214,11 +216,7 @@ add_replica(Q, Node) when ?is_amqqueue(Q) ->
     Pid = amqqueue:get_pid(Q),
     try
         ReplState0 = osiris_writer:query_replication_state(Pid),
-        {{_, InitTs}, ReplState} = maps:take(node(Pid), ReplState0),
-        {MaxTs, MinTs} = maps:fold(fun (_, {_, Ts}, {Max, Min}) ->
-                                           {max(Ts, Max), min(Ts, Min)}
-                                   end, {InitTs, InitTs}, ReplState),
-        case (MaxTs - MinTs) > ?REPLICA_FRESHNESS_LIMIT_MS of
+        case is_replication_stale(node(Pid), ReplState0) of
             true ->
                 {error, {disallowed, out_of_sync_replica}};
             false ->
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -926,6 +934,29 @@ status(Vhost, QueueName) ->
...
     {Key, maps:get(Key, Cnt, undefined)}.
 
+format_sync_status(false, _) ->
+    <<"unknown">>;
+format_sync_status(true, {IsFresh, TimeLag, OffsetLag}) ->
+    IsFreshStr = format_boolean(IsFresh),
+    TimeLagStr =
+        case TimeLag of
+            undefined -> <<"unknown">>;
+            _         -> list_to_binary(integer_to_list(TimeLag) ++ "ms")
```
kjnilsson: redundant whitespace

deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -214,11 +216,7 @@ add_replica(Q, Node) when ?is_amqqueue(Q) ->
     Pid = amqqueue:get_pid(Q),
     try
         ReplState0 = osiris_writer:query_replication_state(Pid),
-        {{_, InitTs}, ReplState} = maps:take(node(Pid), ReplState0),
-        {MaxTs, MinTs} = maps:fold(fun (_, {_, Ts}, {Max, Min}) ->
-                                           {max(Ts, Max), min(Ts, Min)}
-                                   end, {InitTs, InitTs}, ReplState),
-        case (MaxTs - MinTs) > ?REPLICA_FRESHNESS_LIMIT_MS of
+        case is_replication_stale(node(Pid), ReplState0) of
```
the-mikedavis, line 219: I'm not sure how impactful it is but we could track this replication lag stuff as gauges, to avoid the extra call to the writer. I can update https://github.com/rabbitmq/osiris/pull/197 if we want to explore doing that

the-mikedavis:
One snag with the chunk timestamp check is that there could be a spurious big gap if a stream is not active for an extended period, and then a message is published. It could appear like replication is not caught-up even though there is only one chunk to replicate. It depends on when the writer is queried.

I don't see a simple solution though. Timestamps: a gap between publishing bursts makes replication look unhealthy. Offsets: tiny chunks look like replication lag compared to large chunks. Bytes: same issue as offsets (indirectly because syscall overhead stacks up, smaller reads are more expensive), and bytes are more expensive to track. CMQs in <4.x exposed this roughly as `messages - sync_messages`, but that's imperfect for the same reasons as offsets and bytes. Smaller chunks replicate slower, so there isn't a stable way to know if, or by how much, a member is behind, in a way that's portable between workloads.

This is creeping outside of the scope of this PR. But is there appetite for identifying replication freshness/healthiness? I think with the imperfect measurements we have available to us the best we can do is try to set reasonable thresholds and expose something that can be measured as progress, so that health checks know whether replication is stuck somehow or will eventually catch up.

Then we can have a classic progress bar 😄

deps/rabbit/src/rabbit_stream_coordinator.erl
mkuratczyk: sounds great

mkuratczyk:
> This is creeping outside of the scope of this PR. But is there appetite for identifying replication freshness/healthiness? I think with the imperfect measurements we have available to us the best we can do is try to set reasonable thresholds and expose something that can be measured as progress, so that health checks know whether replication is stuck somehow or will eventually catch up.

I think this would be valuable. I added the lag to the `stream_status` in this PR but it did feel like this should be more generic and observable through monitoring as well.

Also important to note, that while testing this PR, it felt like the replication catch-up is just slow (I have no scientific data to support this claim at this point, it just took significantly longer than I'd expect). Quite likely, there's something we can do around read aheads and/or network settings to make it faster. I'm fairly sure users would prefer fast catch-up with imperfect observability than a slow catch-up they can stare at. ;) Of course, ideally it could be faster and observable.

michaelklishin:
@mkuratczyk @the-mikedavis is this PR ready for one last review round or do you want to address something else?

the-mikedavis:
We could block this on https://github.com/rabbitmq/osiris/pull/197 to avoid the writer call. I don't think it's super impactful though (and actually, making the writer do more work makes it more efficient, from my measurements), so I think it's fine as a follow-up goal. Looks like I have to fix the CI for that one too

PR #16604, opened 2026-06-04
Test flake chase: address a hard to reproduce flake (Jun 3, 2026) (round 2)

michaelklishin:
One more speculative flake fix (meaning I could not reproduce the failure even in dozens of runs before addressing it).


michaelklishin:
Not scheduled to a technical glitch at submission time, re-submitted with a different branch name https://github.com/rabbitmq/rabbitmq-server/pull/16605.

PR #16598, opened 2026-06-03
Safely handle channel creation errors on initiating connections and top-level supervisors

Ayanda-D:
## Proposed Changes

Hello! 

We've been noticing that channel creation can fail with `{error, Reason}` - for example when channel-interceptors clash on the AMQP operation they intend to intercept (i.e. expected behaviour to fail). The calling processes and supervision tree currently don't seem to handle these errors/failures gracefully, and fail with loud `badmatch` exceptions (which are not the most useful for operators). As a result, calling components (connections + supervisors) crash ungracefully. These changes seek to add safe handling of `{error, Reason}` channel creation errors, to a channel's top/adjacent supervisor, as well as the initiating connection types - both network connections and direct connections. There's also `shutdown_supervisor/1` which ensures clean termination of the related supervisor before an error is returned. 

### CURRENT CRASHES/UNHANDLED CONNECTION TERMINATIONS

Example crash logs (from conflicting channel-interceptor modules/plugins) are as follows:

<details>
<summary>CRASH-1: rabbit_reader</summary>

```
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>   crasher:
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>     initial call: rabbit_reader:init/3
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>     pid: <0.786.0>
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>     registered_name: []
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>     exception error: no match of right hand side value
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>                      {error,
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>                       {'EXIT',
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>                        {{badmatch,
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>                          {error,
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>                           {{amqp_error,internal_error,
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>                             "Interceptor: more than one module handles #{'queue.declare' => []}",
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>                             none},
2026-06-02 15:40:09.715490+01:00 [error] <0.786.0>                            {child,undefined,channel,
[...]

PR #16610, opened 2026-06-04
By @AyandaDube: Safely handle channel creation errors on initiating connections and top-level supervisors

michaelklishin:
This is #16598 with one tweak from me: in one unlikely (shutdown timeout) case, we can log without crashing the caller.

That's in line with the goal of #16598.


michaelklishin:
With these changes, certain tests that depended on connection termination due to an exception now fail. The solution is to explicitly handle returned errors in their callers, and in the direct client also explicitly close the connection (on the RabbitMQ end).

I am working on the necessary updates.

michaelklishin:
For the above reason, I don't think we should backport these to `v4.3.x` as it is a subtle change that can potentially cause confusion.

michaelklishin:
While updating the direct client behavior I have also noticed that the direct client did not enforce the same set of limits as the TCP-based "network" connections.

Now it does, as a drive-by change.

PR #16639, opened 2026-06-09
Management UI: add 'Delayed' column to the queue list

Guierf22:
## What

Add a 'Delayed' column to the management UI queue list page, in the Messages column group, positioned between Unacked and In Memory.

## Why

RabbitMQ 4.3 introduced a native retry policy for quorum queues, which added a `messages_delayed` counter visible on the queue detail page. However, the main queue list page did not include this counter, making it necessary to open each queue individually to check it.

The 'In Memory' column (`messages_ram`) follows the same pattern — it only applies to classic queues but is included in the list, hidden by default. This change applies the same treatment to 'Delayed'.

## Changes

- `deps/rabbitmq_management/priv/www/js/global.js`: add `msgs-delayed` entry to the Messages column group (hidden by default)
- `deps/rabbitmq_management/priv/www/js/tmpl/queues.ejs`: add the column header and data cell
- `selenium/test/queuesAndStreams/list.js`: update the column picker assertion to include the new entry

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/global.js
+++ b/deps/rabbitmq_management/priv/www/js/global.js
@@ -84,6 +84,7 @@ var ALL_COLUMNS =
                    ['state',                'State',                  true]],
       'Messages': [['msgs-ready',      'Ready',          true],
                    ['msgs-unacked',    'Unacknowledged', true],
+                   ['msgs-delayed',    'Delayed',        false],
                    ['msgs-ram',        'In memory',      false],
                    ['msgs-persistent', 'Persistent',     false],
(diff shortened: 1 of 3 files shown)


michaelklishin:
@Guierf22 thank you but we will not accept this PR for now. Eventually this metric will be available via Prometheus but in the management UI, it has to be reconciled with the Delayed Queue type.

Per discussion with @MirahImage.

michaelklishin:
Some on our team are interested in merging this PR as it is.

ansd (approved):

PR #16669, opened 2026-06-12
Implement AMQP sole conn spec

acogoluegnes:
This commit implements the "AMQP Enforcing Connection Uniqueness
Version 1.0" specification extension. It introduces cluster-wide
tracking of AMQP 1.0 client connections to ensure only a single
connection can be established for a given `container-id` per vhost.

Implementation Details:
* Introduces `rabbit_sole_conn`, a `gen_server` that manages
  a dedicated Khepri (Raft-based) store to track active connection
  leases across the cluster.
* Updates `rabbit_amqp_reader` to parse the `sole-connection-for-container`
  capability and coordinate with the sole connection registry during
  the initial AMQP 1.0 `open` handshake.
* Enforces two conflict resolution policies defined by the spec:
  - `refuse-connection`: Rejects the new connection with an `invalid-field`
    condition if an existing connection holds the lease.
  - `close-existing`: Terminates the existing connection with a
    `resource-locked` condition and grants the lease to the new connection.
* Applies a `weak` detection policy, enforcing uniqueness only when
  explicitly requested by the client via connection properties.

Security and Semantics:
* Connection leases are scoped strictly to the `vhost` and `container-id`.
* Enforces user ownership: a lease can only be taken over or replaced
  (`close-existing`) by the same authenticated user. Attempts by different
  users to hijack an active `container-id` are unconditionally refused.

Operability:
* Adds `rabbitmqctl sole_connection_status` to inspect the Raft consensus
  state of the sole connection registry.
* Adds `rabbitmqctl sole_connection_force_delete` to allow operators to
  manually evict stale leases if necessary.
* Introduces operator commands `wipe` and `start` for low-level recovery
  and bootstrapping of the Khepri store.

Configuration Parameters:
* Expose the three configuration parameters of `rabbit_sole_conn`
  in `rabbitmq.conf` under the `sole_connection` prefix.
* `sole_connection.tick_interval` -> `rabbit.sole_connection_tick_interval`
* `sole_connection.leader_wait_retry_timeout` -> `rabbit.sole_connection_leader_wait_retry_timeout`
* `sole_connection.default_timeout` -> `rabbit.sole_connection_default_timeout`


deps/rabbit/src/rabbit_amqp_sole_conn.erl
```diff
@@ -0,0 +1,982 @@
...
+                            try
+                                erpc:cast(New, ?MODULE, ensure_running, [])
+                            catch
+                                Class:Reason ->
+                                    ?LOG_WARNING("~ts: Failed to cast ensure_running to node ~w. "
+                                                 "Error: ~p:~p",
+                                                 [?MODULE, New, Class, Reason])
+                            end
+                    end,
+                    %% Calculate nodes to add
```
dumbbell: You mean "delete" here

deps/rabbit/src/rabbit_amqp_sole_conn.erl
```diff
@@ -0,0 +1,982 @@
...
+                {error, _} = Err ->
+                    Err
+            catch
+                error:{erpc, timeout} ->
+                    {error, timeout};
+                error:{erpc, RpcReason} ->
+                    {error, RpcReason}
+            end;
+        pang ->
+            %% Node is offline, we can ask it to reset itself.
```
dumbbell: "we can't"

deps/rabbit/src/rabbit_amqp_sole_conn.erl
```diff
@@ -0,0 +1,982 @@
...
+    {ok, #state{resizer_pid = undefined}}.
+
+handle_info(cluster_tick, State = #state{resizer_pid = ResizerPid}) ->
+    ?LOG_DEBUG("sole_conn cluster tick"),
+    erlang:send_after(tick_interval(), self(), cluster_tick),
+    case is_leader() of
+        true when ResizerPid =:= undefined ->
+            ?LOG_DEBUG("leader, spawning resizing process"),
+            %% We are the leader and no resize is currently running. Start one.
+            {Pid, _MonitorRef} = spawn_monitor(fun maybe_resize_cluster/0),
```
dumbbell: I’m uncomfortable with the start of a process outside of the supersivison tree of `rabbit`.  We already do that for the process tha reconciliates vhost processes and it’s nightmare in CI with the short-lived RabbitMQ nodes, because the process continues to run after `rabbit` is stopped, crashing all over the place, leading to many test flakes.

deps/rabbit/src/rabbit_amqp_sole_conn.erl
```diff
@@ -0,0 +1,982 @@
...
+            status0()
+    end.
+
+status0() ->
+    case members() of
+        {ok, Members} ->
+            [begin
+                 %% Securely call ra:key_metrics/1 on the remote node
+                 MetricsResult = try
+                                     erpc:call(N, ra, key_metrics, [ServerId], ?RPC_TIMEOUT)
```
dumbbell: You can use `erpc:multicall()` which performs calls in parallel. If there are unresponsive nodes, this is way faster.

deps/rabbit/src/rabbit_amqp_sole_conn.erl
```diff
@@ -0,0 +1,982 @@
...
+                    timer:sleep(?RESIZE_POLL_INTERVAL),
+                    wait_for_resize_to_settle(Node, Deadline)
+            end;
+        false ->
+            ok
+    end.
+
+is_resizing_remote(Node) ->
+    try erpc:call(Node, ?MODULE, is_resizing, [], ?ALIVENESS_RPC_TIMEOUT) of
+        Result -> Result
```
dumbbell: No need to use the `try ... of` variant here.

deps/rabbit/src/rabbit_amqp_sole_conn.erl
```diff
@@ -0,0 +1,982 @@
...
+check_conn(#conn{pid = Pid}) ->
+    Node = node(Pid),
+    case Node =:= node() of
+        true ->
+            is_process_alive(Pid);
+        false ->
+            try erpc:call(Node, erlang, is_process_alive, [Pid],
+                          ?ALIVENESS_RPC_TIMEOUT) of
+                Result ->
+                    Result
```
dumbbell: Likewise, you can use a simple `try` block instead.

deps/rabbit/src/rabbit_sole_conn.erl
```diff
@@ -0,0 +1,982 @@
...
+    ?RABBITMQ_KHEPRI_VHOST_PATH(VHost, [amqp10_sole_conn, ContainerId]).
+
+kill_connection_sproc_path() ->
+    ?RABBITMQ_KHEPRI_ROOT_PATH([amqp10_sole_conn, kill_connection]).
+
+kill_connection_sproc_trigger_pattern() ->
+    ?RABBITMQ_KHEPRI_VHOST_PATH(?KHEPRI_WILDCARD_STAR_STAR,
+                                [amqp10_sole_conn,
+                                 ?KHEPRI_WILDCARD_STAR_STAR]).
+
```
dumbbell, line 1092: I’m not sure you need `?KHEPRI_WILDCARD_STAR_STAR` (which is recursive). Why not `?KHEPRI_WILDCARD_STAR`?

dumbbell (changes requested):
The Khepri part looks good overall. I just have minor comments for now.

deps/rabbit/src/rabbit_sole_conn.erl
acogoluegnes: Maybe a mistake with completion, I switched both patterns `?KHEPRI_WILDCARD_STAR`.

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/amqp_sole_conn_status_command.ex
```diff
@@ -0,0 +1,55 @@
+## This Source Code Form is subject to the terms of the Mozilla Public
+## License, v. 2.0. If a copy of the MPL was not distributed with this
+## file, You can obtain one at https://mozilla.org/MPL/2.0/.
+##
+## Copyright (c) 2007-2026 Broadcom. All Rights Reserved. The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries.  All rights reserved.
+
+defmodule RabbitMQ.CLI.Ctl.Commands.AmqpSoleConnStatusCommand do
```
ansd: Should we change the CLI command name from
```
rabbitmqctl amqp_sole_conn_status
```
to
```
rabbitmqctl sole_connection_status
```
?

Maybe we would like to use the same CLI command if we make use of this feature for protocols other than AMQP in the future?

ansd: Same for the other CLI command `force_delete`.

deps/rabbit/src/rabbit_amqp_sole_conn.erl
acogoluegnes: I added a supervisor between the `rabbit_sup` supervisor and the gen_server in sole conn. This avoids taking down other workers in case the sole conn gen_server crashes.

I also linked the resizing process to the gen_server process (with `proc_lib:spawn_link/1`), this way it is stopped when the gen_server is terminated.

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/amqp_sole_conn_status_command.ex
acogoluegnes: I used the `amqp_` prefix because the feature is from the "AMQP Enforcing Connection Uniqueness" spec extension, so linked to AMQP.

I would have thought that we would use another prefix for other protocols, or even another name ("sole conn" is used in the spec extension, it is not a universal term or concept), because it may reflect another feature in another protocol.

What would the plan for the future if we were using `rabbitmqctl sole_connection_status`? Adding a flag for the protocol, e.g. `rabbitmqctl sole_connection_status --protocol amqp` and `rabbitmqctl sole_connection_status --protocol mqtt`?

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -417,6 +417,18 @@ end}.
 {mapping, "metadata_store.khepri.snapshot_interval", "rabbit.khepri_snapshot_interval",
     [{datatype, {integer, positive}}]}.
 
+{mapping, "amqp10_sole_conn.tick_interval", "rabbit.amqp10_sole_conn_tick_interval",
```
ansd: Can you rename all these config options so that they are prefixed with `amqp_` instead of `amqp10_` ?

deps/rabbit/src/rabbit_amqp_sole_conn.erl
```diff
@@ -0,0 +1,1005 @@
...
+    gen_server:start_link({local, ?GEN_SERVER_NAME}, ?MODULE, [], []).
+
+init([]) ->
+    %% must trap exits for spawn_link and terminate/2 to work correctly
+    process_flag(trap_exit, true),
+    erlang:send_after(tick_interval(), self(), cluster_tick),
+    {ok, #state{resizer_pid = undefined}}.
+
+handle_info(cluster_tick, State = #state{resizer_pid = ResizerPid}) ->
+    ?LOG_DEBUG("sole_conn cluster tick"),
```
ansd: This debug message every 30 seconds feel a bit useless. Should we remove it?

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/amqp_sole_conn_force_delete_command.ex
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def output({:error, :sole_conn_not_started_or_available}, _opts) do
+    {:error, RabbitMQ.CLI.Core.ExitCodes.exit_software(),
+     "Cannot force-delete the AMQP 1.0 sole connection lease as the enforcement feature is not started or unavailable"}
+  end
+
+  def output({:error, :sole_conn_feature_flag_not_enabled}, _opts) do
+    {:error, RabbitMQ.CLI.Core.ExitCodes.exit_software(),
+     "AMQP 1.0 sole connection enforcement requires all nodes in the cluster to run RabbitMQ 4.4.0 or later"}
```
ansd:
```
rabbit_feature_flags:is_enabled('rabbitmq_4.4.0')
```
is mapped to
```
{error, sole_conn_feature_flag_not_enabled}
```
which here gets again mapped back to
```
  def output({:error, :sole_conn_feature_flag_not_enabled}, _opts) do
    {:error, RabbitMQ.CLI.Core.ExitCodes.exit_software(),
     "AMQP 1.0 sole connection enforcement requires all nodes in the cluster to run RabbitMQ 4.4.0 or later"}
  end
```
Any reason to not directly return `{error, {feature_flag_disabled, 'rabbitmq_4.4.0'}}` in `rabbit_amqp_sole_conn:status/0` ?

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/amqp_sole_conn_status_command.ex
ansd: > I would have thought that we would use another prefix for other protocols, or even another name ("sole conn" is used in the spec extension, it is not a universal term or concept), because it may reflect another feature in another protocol.

"sole conn" means the "only/single connection". This concept is also applicable to other protocols, see "Session taken over" in https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.html. By decoupling the feature (and CLI commands) from the protocol, we could re-use the feature (and CLI commands) for other protocols in the future (if we ever wanted).

deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/amqp_sole_conn_force_delete_command.ex
acogoluegnes: Updated the sole conn module to return `{error, {feature_flag_disabled, 'rabbitmq_4.4.0'}}`. The CLI commands extract the name of the feature flag to display it in the error message (they don't hardcode the FF name anymore).

ansd (approved):

michaelklishin:
@acogoluegnes @ansd would it be too late to rename the `rabbitmq.conf` keys to something like

`amqp10.sole_connection.{key}`?

We don't have a standard prefix like that for AMQP 1.0 but we do have `mqtt.*` and `stomp.*` keys. That is, prefixes named after protocols.

So I suggest that we organize new AMQP 1.0 settings under `amqp10.*` (with as many levels of grouping/nesting as we want).

acogoluegnes:
@michaelklishin We removed the protocol prefix (I updated the PR description, it was outdated, but the code is correct). We use only the `sole_connection` prefix, as the feature could be used for other protocols. So it is no longer specific to AMQP 1.0, so using an `amqp10.` prefix would not be relevant.

deps/rabbit/src/rabbit_sole_conn.erl
```diff
@@ -0,0 +1,1101 @@
...
+    Opts = default_create_options(ConnPid),
+    Payload = #conn{pid = ConnPid,
+                    username = Username},
+    case khepri_adv:create(get_store_id(), Path, Payload, Opts) of
+        {ok, _} ->
+            %% no node yet, accept
+            %% node will clean itself when the connection is closed
+            ok;
+        {error, {khepri, mismatching_node, #{node_props := #{data := ExistingConn}}}} ->
```
dumbbell: Could you please use the `?khepri_error()` instead of hard-coding the tuple format?

```erlang
{error, ?khepri_error(mismatching_node, #{node_props := #{data := ExistingConn}})}
```

deps/rabbit/src/rabbit_sole_conn.erl
```diff
@@ -0,0 +1,1101 @@
...
+    end;
+do_acquire(close_existing = Plcy, VHost, ContainerId, Username, ConnPid) ->
+    Path = conn_path(VHost, ContainerId),
+    Opts = default_create_options(ConnPid),
+    Payload = #conn{pid = ConnPid,
+                    username = Username},
+    case khepri_adv:create(get_store_id(), Path, Payload, Opts) of
+        {ok, _} ->
+            ok;
+        {error, {khepri, mismatching_node, #{node_props := #{data := ExistingConn}}}} ->
```
dumbbell: Same comment as above

dumbbell (changes requested):
A couple minor comments

deps/rabbit/src/rabbit_sole_conn.erl
```diff
@@ -0,0 +1,1101 @@
...
+
+refuse_connection_error() ->
+    %% the error field of close MUST have an error with the condition field
+    %% of error being invalid-field and the info field of error having
+    %% the symbol key invalid-field taking the symbol value container-id.
+    %% [sole conn 3.2.1]
+    amqp_error(
+      ?V_1_0_AMQP_ERROR_INVALID_FIELD,
+      <<"the container-id is already in use">>,
+      {?V_1_0_AMQP_ERROR_INVALID_FIELD, {symbol, <<"container-id">>}}).
```
ansd: Using macro `V_1_0_AMQP_ERROR_INVALID_FIELD` is wrong since this should be `{symbol, <<"invalid-field">>}` instead of `{symbol, <<"amqp:invalid-field">>}`.

ansd (changes requested):

deps/rabbit/src/rabbit_sole_conn.erl
```diff
@@ -0,0 +1,1101 @@
...
+                        [] ->
+                            ok;
+                        [Old | _] when length(RabbitNodes) > 0 ->
+                            %% This should be rare, as forget_cluster_node shrinks
+                            %% this cluster as well
+                            ?LOG_INFO("~ts: RabbitMQ node ~w was formally removed from the cluster, "
+                                      "evicting it from the sole_conn Khepri cluster",
+                                      [?MODULE, Old]),
+                            _ = evict_node(Old),
+                            ok;
```
dumbbell, line 215: Same as above: here, you only evict one of the former cluster members. Is it on purpose? I mean, do you rely on the fact that it is evaluated on a regular basis?

deps/rabbit/src/rabbit_sole_conn.erl
```diff
@@ -0,0 +1,1101 @@
...
+                    AddableNodes = [N || N <- RabbitNodes, lists:member(N, Present)],
+                    %% Calculate nodes to add
+                    case AddableNodes -- MemberNodes of
+                        [] ->
+                            ok;
+                        [New | _] ->
+                            ?LOG_INFO("~ts: Expanding sole_conn Khepri cluster to "
+                                      "running node ~w", [?MODULE, New]),
+                            try
+                                erpc:cast(New, ?MODULE, ensure_running, [])
```
dumbbell, line 196: Here, you only consider one new node among all possible new nodes. Is it on purpose? I mean, do you rely on the fact that it is evaluated on a regular basis?

deps/rabbit/src/rabbit_sole_conn.erl
```diff
@@ -0,0 +1,1101 @@
...
+%% --------------------------------------------------------------
+%% Lifecycle and store management
+%% --------------------------------------------------------------
+
+recover() ->
+    LocalServerId = {get_store_id(), node()},
+    %% We ask RA to passively check the disk and restart the Khepri state machine
+    ?LOG_DEBUG("Trying to restart local sole_conn RA server on ~p", [node()]),
+
+    case ra:restart_server(get_ra_system(), LocalServerId) of
```
dumbbell: Why do you bypass Khepri here to restart the store? Khepri does more than restarting the state machine, so you can’t really do that here.

dumbbell (changes requested):

deps/rabbit/src/rabbit_sole_conn.erl
acogoluegnes: This is the same strategy as the stream coordinator. It was originally to spread the addition/removal of nodes over time, because it triggers a restart of each stream IIRC. The resizing is triggered every 30 seconds, so the Khepri instance should span the whole cluster at the end.

deps/rabbit/src/rabbit_sole_conn.erl
acogoluegnes: See answer in other comment.

deps/rabbit/src/rabbit_sole_conn.erl
acogoluegnes: Updated to use `ra_directory:uid_of/2` to detect the present of local files. Then use `khepri:start/2` accordingly.

dumbbell (approved):

PR #16672, opened 2026-06-12
Replace rabbit_fifo_dlx_sup with two-level supervisor hierarchy

### LINKED ISSUE #16652: `at_least_once` DLX quorum queues permanently stranded during membership churn
## Describe the bug

On a 3-node cluster with many `at_least_once` dead-lettering quorum queues, quorum-queue membership churn (a rolling node replacement: a node leaving, a replacement rejoining, and abrupt node terminations) can permanently strand a subset of the `at_least_once` DLX queues with a member stuck `noproc` and no elected leader. Recovery requires a manual `ra:restart_server`. Plain quorum queues (no DLX, or `at_most_once`) recover normally via election.

Observed on 4.2.4 (Erlang 27.3.4.2). The relevant code paths are unchanged on `main`.

## Mechanism (from logs and source)

1. During membership churn, many `at_least_once` DLX queues transition leadership. On entering leader state, `rabbit_fifo_dlx:state_enter(leader, ..., at_least_once, ...)` calls `start_worker`, which does `{ok, Pid} = supervisor:start_child(rabbit_fifo_dlx_sup, [QRef])`.
2. The `rabbit_fifo_dlx_worker` instances fail to register their `{dlx,{checkout}}` command against the flapping leader and exit `{error, ra_command_failed}`, en masse (observed about 80 worker exits in a single second).
3. `rabbit_fifo_dlx_sup` (`simple_one_for_one, intensity => 100, period => 1`) becomes unavailable on every node. On one node it logged an explicit `reached_max_restart_intensity` give-up; on another it stopped servicing calls and subsequent `start_child` returned `noproc`.
4. With the DLX worker supervisor unavailable cluster-wide, each `at_least_once` DLX queue entering leader state crashes its `ra_server` at the `{ok, Pid} = supervisor:start_child(...)` line: `exit:{noproc,{gen_server,call,[rabbit_fifo_dlx_sup,{start_child,...}]}}`. This is the intended behavior per the comment in `start_worker/1` ("if starting the rabbit_fifo_dlx_worker fails, let the Ra server process crash in which case another Ra node will become leader"), but the failure is correlated across all nodes, so there is no healthy failover target.
5. The crashing `ra_server` trips its per-queue `ra_server_sup` (`intensity => 2, period => 5`), which after 2 crashes in 5 seconds permanently abandons the member: `noproc`, no auto-restart. The queue stays stranded until `ra:restart_server` is run manually.

In our incident this stranded 52 `at_least_once` DLX queues; `ra_server_sup` logged 500+ max-restart-intensity give-ups per surviving node, while `rabbit_fifo_dlx_sup` itself gave up at most once per node.

## Why existing mitigations do not cover this

- PR #7677 raised `rabbit_fifo_dlx_sup` to `intensity 100, period 1` ("do not terminate the sup if a low number of children terminate") and added "last checkout wins" and "do not restart the DLX worker if the leader is non-local". Under DLX churn the 100/1 budget is still exceeded, or the supervisor is otherwise left unavailable.
- PR #14401 (in 4.2.0, present in our 4.2.4 build) batches leadership transfers during graceful drain. Our churn came from node rejoin and abrupt termination, which do not go through that path, so it did not help.
- PR #16203 (merged 2026-04-23, backported to v4.2.x/v4.3.x, **not present in our 4.2.4 build dated 2026-05-12**) switches the DLX checkout command from `ra:process_command/3` to `ra:pipeline_command/4` with a selective receive, preventing Ra from redirecting the checkout to a new leader during failover. This reduces the rate of `{error, ra_command_failed}` worker crashes (workers exit with `non_local_leader` instead of retrying 5 times), making the mass-crash less likely. However, the `checkout0` retry loop can still timeout and return `ra_command_failed` under sustained churn, so the structural problem remains: if enough workers crash simultaneously from any cause, `rabbit_fifo_dlx_sup` still dies and the same cascade occurs.
- Discussion #14780 reports a closely related symptom (a rolling restart leaving quorum queues without an elected leader); the underlying cause was not pinned there.

## Reproduction steps

Tested on a 3-node cluster running RabbitMQ 4.2 (Erlang 27.3.4.2).

Setup: a quorum queue with `dead-letter-strategy: at-least-once` and `overflow: reject-publish` (required for `at-least-once` to activate), plus a dead-letter exchange and target queue.

Step 1 below uses `unregister(rabbit_fifo_dlx_sup)` to simulate the state that results from the supervisor reaching max restart intensity. This was observed on a cluster with ~564 `at_least_once` DLX queues with membership churn (80+ workers crashing with `{error, ra_command_failed}` in a 1-second window, exhausting the supervisor's `intensity 100/period 1` budget). The `unregister` is a shortcut to reach that end state deterministically without requiring the conditions that kill the supervisor in the wild.

```bash
rabbitmqctl eval '
  %% 1. Unregister rabbit_fifo_dlx_sup so start_child returns noproc.
  %%    Simulates the supervisor being unavailable after reaching max
  %%    restart intensity. The process stays alive so rabbit_sup
  %%    (one_for_all, intensity 0) does not notice and the node stays up.
  unregister(rabbit_fifo_dlx_sup).
'

rabbitmqctl eval '
  %% 2. Transfer leadership away then back to trigger state_enter(leader).
  QName = rabbit_misc:r(<<"vhost">>, queue, <<"my_queue">>),
  {ok, Q} = rabbit_amqqueue:lookup(QName),
  Members = amqqueue:get_nodes(Q),
  OtherNodes = [N || N <- Members, N =/= node()],
  rabbit_quorum_queue:transfer_leadership(Q, hd(OtherNodes)),
  timer:sleep(2000),
  {ok, Q2} = rabbit_amqqueue:lookup(QName),
  rabbit_quorum_queue:transfer_leadership(Q2, node()).
'

# Wait for ra_server_sup (intensity 2/period 5) to give up
sleep 10

rabbitmqctl eval '
  %% 3. Verify the queue is stranded.
  {ok, RaName} = rabbit_queue_type_util:qname_to_internal_name(
    rabbit_misc:r(<<"vhost">>, queue, <<"my_queue">>)),
  whereis(RaName).  %% returns undefined = stranded
'
```

The ra_server crashes 3 times in ~1.1 seconds (each crash-loop iteration elects itself candidate, enters leader state, calls `start_worker`, hits `{noproc, start_child}`, crashes), exhausting `ra_server_sup` intensity 2/period 5. The member is permanently abandoned.

Log output:

```
State machine 'vhost_my_queue' terminating
Reason for termination = exit:{noproc,
                                   [rabbit_fifo_dlx_sup,
                                    {start_child,[{resource,<<"vhost">>,queue,<<"my_queue">>}]},
                                    infinity]}}
...
reason: reached_max_restart_intensity
```

## Expected behavior

A transient inability to start DLX workers during membership churn should not permanently strand quorum-queue members.

## Suggested fix: double-layer supervisor (`_sup_sup` pattern)

The root structural problem is that `rabbit_fifo_dlx_sup` is a single, globally-registered `simple_one_for_one` supervisor that hosts ALL DLX workers for ALL `at_least_once` queues on the node. A burst of worker exits that exceeds its intensity kills the supervisor, taking all remaining healthy workers with it and leaving no viable `start_child` target for any queue on the node.

The `_sup_sup` pattern already used by Ra (`ra_server_sup_sup` / `ra_server_sup`) and elsewhere in RabbitMQ (`rabbit_channel_sup_sup`, `rabbit_amqqueue_sup_sup`) eliminates this:

```
rabbit_fifo_dlx_sup_sup          (simple_one_for_one, children are supervisors, restart => temporary)
  +-- rabbit_fifo_dlx_worker_sup [queue A]  (one_for_one, intensity N/period M)
  |     +-- rabbit_fifo_dlx_worker [queue A]
  +-- rabbit_fifo_dlx_worker_sup [queue B]
        +-- rabbit_fifo_dlx_worker [queue B]
```

Why this works:

- Each DLX worker is isolated behind its own per-worker supervisor.
- A burst of worker crashes kills individual per-worker supervisors, not the top-level `sup_sup`.
- `restart => temporary` on the child spec means OTP's supervisor never restarts a dead per-worker supervisor and never increments the `sup_sup` intensity counter (OTP `supervisor.erl` `do_restart/3` for temporary children deletes the child and returns without calling `add_restart/1`). The `sup_sup` is structurally immune to `reached_max_restart_intensity` regardless of the number or rate of child crashes.
- `supervisor:start_child(rabbit_fifo_dlx_sup_sup, ...)` always works because the `sup_sup` is always alive. When leadership settles and `state_enter(leader)` fires again, `start_worker` can create a fresh per-worker supervisor + worker. The queue recovers on the next leadership attempt rather than being permanently stranded.

Changes required:

1. `rabbit_fifo_dlx_sup.erl` becomes `rabbit_fifo_dlx_sup_sup.erl`: `type => supervisor`, `restart => temporary`, child start is `{rabbit_fifo_dlx_worker_sup, start_link, []}`.
2. New `rabbit_fifo_dlx_worker_sup.erl`: trivial `one_for_one` supervisor hosting one `rabbit_fifo_dlx_worker`.
3. `rabbit_fifo_dlx:start_worker/1`: target changes from `rabbit_fifo_dlx_sup` to `rabbit_fifo_dlx_sup_sup`.
4. `rabbit_fifo_dlx:ensure_worker_terminated/1`: terminate the per-worker supervisor (e.g. via the worker Pid's parent, similar to how `ra_server_sup_sup:stop_server` uses `ra_directory:where_is_parent`).

## Additional context

See the attached `rabbitmq-dlx-artifacts.md` for full log evidence: the `rabbit_fifo_dlx_worker` exits with `{error, ra_command_failed}`, the `rabbit_fifo_dlx_sup` `child_terminated` and `reached_max_restart_intensity` reports, the `ra_server` "State machine ... terminating" dumps with `exit:{noproc,{gen_server,call,[rabbit_fifo_dlx_sup,{start_child,...}]}}`, the `ra_server_sup` give-up reports, and the clean reproduction log output.

[rabbitmq-dlx-artifacts.md](https://github.com/user-attachments/files/28805014/rabbitmq-dlx-artifacts.md)

lukebakken:
## Summary

Fixes #16652.

A burst of DLX worker crashes can trip the flat `rabbit_fifo_dlx_sup` supervisor's restart intensity (100/1s), rendering it permanently unavailable. Once unavailable on all nodes, every `at_least_once` DLX quorum queue that enters leader state crashes its Ra server on `{noproc, start_child}`. Each Ra server then trips `ra_server_sup` (intensity 2/period 5), permanently stranding the queue with no automatic recovery.

## Changes

Replace the flat `simple_one_for_one` supervisor with a two-level hierarchy:

- `rabbit_fifo_dlx_sup_sup` - top-level `simple_one_for_one` whose children are per-queue worker supervisors (`restart => temporary`).
- `rabbit_fifo_dlx_worker_sup` - per-queue `one_for_one` with `intensity => 0`. A worker crash terminates only its own supervisor, isolating failures to a single queue without affecting the top-level supervisor or any other queue.

With `restart => temporary` children in the sup_sup, crashes never increment the top-level intensity counter (OTP `supervisor.erl` skips `add_restart` for temporary children in `do_restart`). The sup_sup itself can never reach max restart intensity regardless of how many workers crash simultaneously.

Additional fixes included in this PR:

- `ensure_worker_started` now checks live supervisor state via `find_worker_sup/1` before starting a new worker, preventing duplicate worker_sups from concurrent `state_enter` and `{dlx, setup}` aux calls racing.
- `state_enter(eol, ...)` now terminates the DLX worker on queue deletion, preventing orphaned worker_sups.

## How to reproduce

See the reproduction steps in #16652. The simplest deterministic trigger:

1. `erlang:unregister(rabbit_fifo_dlx_sup)` on all nodes (simulates supervisor unavailability)
2. Transfer leadership of an `at_least_once` DLX quorum queue
3. Observe the queue becomes permanently stranded (`noproc`)

With the fix applied, step 1 is no longer possible - individual worker crashes are isolated to their own `rabbit_fifo_dlx_worker_sup` and cannot affect the top-level supervisor or other queues.

## Test

- `rabbit_fifo_dlx_SUITE` - 5/5 pass
- `rabbit_fifo_dlx_integration_SUITE` - 19/19 pass (multiple clean runs)

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -1113,8 +1113,7 @@ state_enter(leader,
                     enqueuers = Enqs,
                     waiting_consumers = WaitingConsumers,
                     cfg = #cfg{resource = QRes,
-                               dead_letter_handler = DLH},
-                    dlx = DlxState}) ->
+                               dead_letter_handler = DLH}}) ->
     % return effects to monitor all current consumers and enqueuers
     Pids = lists:usort(maps:keys(Enqs)
                        ++ [P || ?CONSUMER_PID(P) <- maps:values(Cons)]
@@ -1124,17 +1123,11 @@ state_enter(leader,
     NodeMons = lists:usort([{monitor, node, node(P)} || P <- Pids]),
     NotifyDecs = notify_decorators_startup(QRes),
     Effects = Mons ++ Nots ++ NodeMons ++ [NotifyDecs],
-
-    case DLH of
-        at_least_once ->
-            ensure_worker_started(QRes, DlxState);
-        _ ->
-            ok
-    end,
-    Effects;
+    maybe_add_dlx_effect(DLH, Effects);
 state_enter(eol, #?STATE{enqueuers = Enqs,
                          consumers = Cons0,
-                         waiting_consumers = WaitingConsumers0}) ->
+                         waiting_consumers = WaitingConsumers0} = State) ->
+    ok = maybe_terminate_worker(State),
     Custs = maps:fold(fun(_K, ?CONSUMER_PID(P) = V, S) ->
                               S#{P => V}
                       end, #{}, Cons0),
@@ -1145,15 +1138,8 @@ state_enter(eol, #?STATE{enqueuers = Enqs,
     [{send_msg, P, eol, ra_event}
      || P <- maps:keys(maps:merge(Enqs, AllConsumers))] ++
     [{aux, eol}];
-state_enter(_, #?STATE{cfg = #cfg{dead_letter_handler = DLH,
-                                  resource = _QRes},
-                       dlx = DlxState}) ->
-    case DLH of
-        at_least_once ->
-            ensure_worker_terminated(DlxState);
-        _ ->
-            ok
-    end,
+state_enter(_, State) ->
+    ok = maybe_terminate_worker(State),
     %% catch all as not handling all states
     [].
 
@@ -1506,15 +1492,7 @@ handle_aux(_RaState, _, force_checkpoint,
                                    ReclaimableBytes, true),
     {no_reply, Aux#?AUX{last_checkpoint = Check}, RaAux, Effects};
 handle_aux(leader, _, {dlx, setup}, Aux, RaAux) ->
-    #?STATE{dlx = DlxState,
-            cfg = #cfg{dead_letter_handler = DLH,
-                       resource = QRes}} = ra_aux:machine_state(RaAux),
-    case DLH of
-        at_least_once ->
-            ensure_worker_started(QRes, DlxState);
-        _ ->
-            ok
-    end,
+    ok = maybe_start_dlx_worker(RaAux),
     {no_reply, Aux, RaAux};
 handle_aux(_, _, {dlx, teardown, Pid}, Aux, RaAux) ->
     terminate_dlx_worker(Pid),
@@ -1524,7 +1502,6 @@ handle_aux(_, _, Unhandled, Aux, RaAux) ->
     ?LOG_DEBUG("~ts: rabbit_fifo: unhandled aux command ~P",
                [rabbit_misc:rs(QR), Unhandled, 10]),
     {no_reply, Aux, RaAux}.
-
 
 eval_gc(RaAux, MacState,
         #?AUX{gc = #aux_gc{last_raft_idx = LastGcIdx} = Gc} = AuxState) ->
@@ -3989,16 +3966,7 @@ dlx_apply(_, {dlx, {checkout, ConsumerPid, Prefetch}},
                 discards = Discards0,
                 msg_bytes = Bytes,
                 msg_bytes_checkout = BytesCheckout} = State0) ->
-    %% Since we allow only a single consumer, the new consumer replaces the old consumer.
-    case ConsumerPid of
-        OldConsumerPid ->
-            ok;
-        _ ->
-            ?LOG_DEBUG("Terminating ~p since ~p becomes active rabbit_fifo_dlx_worker",
-                       [OldConsumerPid, ConsumerPid]),
-            %% turn into aux command
-            ensure_worker_terminated(State0)
-    end,
+    ok = ensure_worker_terminated(ConsumerPid, OldConsumerPid, State0),
     %% All checked out messages to the old consumer need to be returned to the discards queue
     %% such that these messages will be re-delivered to the new consumer.
     %% When inserting back into the discards queue, we respect the original order in which messages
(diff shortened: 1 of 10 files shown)


deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -4024,41 +3992,79 @@ dlx_apply(_, Cmd, DLH, State) ->
 %% down: 90 bytes
 %% enqueue overhead 210
 
-ensure_worker_started(QRef, #?DLX{consumer = undefined}) ->
-    start_worker(QRef);
-ensure_worker_started(QRef, #?DLX{consumer = #dlx_consumer{pid = Pid}}) ->
+maybe_add_dlx_effect(at_least_once, Effects) ->
+    Effects ++ [{aux, {dlx, setup}}];
```
the-mikedavis: We can probably make this a prepend?

deps/rabbit/src/rabbit_fifo.erl
```diff
@@ -4024,41 +3992,79 @@ dlx_apply(_, Cmd, DLH, State) ->
...
 
 %% Ensure that starting the rabbit_fifo_dlx_worker succeeds.
 %% Therefore, do not use an effect.
 %% Also therefore, if starting the rabbit_fifo_dlx_worker fails, let the
 %% Ra server process crash in which case another Ra node will become leader.
-start_worker(QRef) ->
-    {ok, Pid} = supervisor:start_child(rabbit_fifo_dlx_sup, [QRef]),
+start_dlx_worker(QRef) ->
+    {ok, SupPid} = supervisor:start_child(rabbit_fifo_dlx_sup_sup, [QRef]),
+    [{_, Pid, worker, _}] = supervisor:which_children(SupPid),
```
the-mikedavis: Looks like we do this in multiple modules. Maybe we can extract a helper in `rabbit_fifo_dlx_sup_sup`?

deps/rabbit/src/rabbit_fifo_dlx_worker.erl
```diff
@@ -100,12 +100,16 @@
...
--spec init(rabbit_amqqueue:name()) ->
+-spec init({rabbit_amqqueue:name(), pid()}) ->
     {ok, undefined, {continue, rabbit_amqqueue:name()}}.
-init(QRef) ->
+init({QRef, SupPid}) ->
+    %% Stored in the process dictionary so that terminate_dlx_worker/2 can
+    %% retrieve it via process_info(WorkerPid, dictionary) without blocking
+    %% on a gen_server call. Same pattern as rabbit_misc:store_proc_name/1.
+    put(sup_pid, SupPid),
     {ok, undefined, {continue, QRef}}.
```
the-mikedavis: This makes me wonder if we should be storing info in an ETS table instead? If the goal is easy querying then a table should be ideal. Process dictionary stuff is always squirrely.

the-mikedavis (commented):
A few first-pass comments...

deps/rabbit/src/rabbit_fifo_dlx_worker.erl
lukebakken: I moved the termination logic into `rabbit_fifo_dlx_worker:terminate_worker/1` (the module that owns the process dictionary key) and introduced a `?DLX_WORKER_SUP_PID_KEY` macro so there's one definition for the contract.

The AI and I considered ETS but the tradeoff doesn't seem worth it here: `ra_directory` justifies ETS because it's a full registry (5 fields per entry, queried from many code paths). Our case is a single pid mapping read only during termination. The process dictionary entry auto-cleans on process death (no table lifecycle, no cleanup-on-crash path), and the pattern matches `rabbit_misc:store_proc_name/1` which is read externally via `process_info` by `rabbitmq_top`.

michaelklishin:
@lukebakken this potentially breaks a Tanzu RabbitMQ-specific queue type, so that will take some time to confirm and find a reasonable mitigation.

deps/rabbit/src/rabbit_fifo_dlx_worker_sup.erl
```diff
@@ -0,0 +1,30 @@
...
+init([QRef]) ->
+    DlxWorkerSupPid = self(),
+    SupFlags = #{strategy => one_for_one,
+                 intensity => 0,
+                 period => 1},
+    ChildSpec = #{id => rabbit_fifo_dlx_worker,
+                  start => {rabbit_fifo_dlx_worker, start_link, [QRef, DlxWorkerSupPid]},
+                  type => worker,
+                  restart => transient,
+                  shutdown => 5000},
```
the-mikedavis, line 31: For this mid-layer supervisor I think we want to leverage a somewhat recent OTP feature to add `auto_shutdown => any_significant` to the `SupFlags` map, and mark `significant => true` to the `ChildSpec` map. This way we make it impossible for this midlayer supervisor to survive a worker termination, if the termination is graceful (`normal` exit, which never actually happens, or `{shutdown, _}`, which does happen).

lukebakken: Got it!

Issue #8495, opened 2023-06-07
reject-publish ignored by Dead-Letter quorum queues


leadenmoth:
### Describe the bug

RabbitMQ 3.11.11, Erlang 25.3, deployed in AKS via cluster-operator 2.2.0

Setting `max-length: <int>` and `overflow: reject-publish` policy on a quorum **dead letter** queue leads to queue continuously filling beyond max-length (I stopped the test at 25000 messages with max-length:10). According to the docs, 

> When a quorum queue reaches the `max-length` limit and `reject-publish` is configured it notifies each publishing channel who from thereon will reject all messages back to the client.

My understanding is that pushing a dead message from main queue to dead letter exchange and then to bound dead letter queue is done entirely on the server side?.. If so, my best guess is that whatever does that publishing (some internal publisher?..) ignores the reject-publish notification. Setting `drop-head` (well, unsetting `reject-publish` and letting it default to `drop-head`) instead works perfectly.

### Reproduction steps

1. Set up normal and DL exchanges, set up normal and DL queues of type `quorum`, set queue-exchange bindings, set `x-dead-letter-exchange`, `x-dead-letter-routing-key` and `x-delivery-limit` on normal queue
2. Set up a `max-length` and `reject-publish` policy on the DL queue
3. Publish significantly more messages than `max-length` to normal queue
4. Consume and reject messages from normal queue continuously until `x-delivery-limit` is reached and the messages start moving to DL queue
5. Observe that all messages are moved to DL queue, regardless of `max-length`
...

### Expected behavior

Messages to-be-dead-lettered above the max-length limit should be discarded when `reject-publish` is set

### Additional context

[...]

PR #16673, opened 2026-06-12
rabbit_fifo: Add `reject_publish_dlx` overflow strategy

the-mikedavis:
This operates just like the classic queue overflow strategy. If enqueueing a message would overflow the queue, the message is run through the dead lettering route and also the enqueue is rejected.

Prior to this change, QQs would fall back to drop-head when `reject-publish-dlx` was specified.

This is based on the next QQ version PR: #16583.

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -203,8 +203,7 @@
               delayed_op/0]).
 
 -spec init(config()) -> state().
-init(#{name := Name,
-       queue_resource := Resource} = Conf) ->
+init(#{name := Name, queue_resource := Resource} = Conf) ->
     update_config(Conf, #?STATE{cfg = #cfg{name = Name,
                                            resource = Resource}}).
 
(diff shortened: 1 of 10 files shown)


the-mikedavis:
I couldn't quite dig up the history for why this wasn't implemented, historically. I think it was because it doesn't stem publishers off when the queue is full like the `reject_publish` strategy? So publishers might extend the log with messages that won't end up in the queue and bump up the disk footprint. But with compaction I think this cost might be more acceptable?

lukebakken:
Is this PR related to https://github.com/rabbitmq/rabbitmq-server/issues/8495?

the-mikedavis:
Not quite, but the concepts overlap in a way where it's not at all obvious. #8495 is about how `reject-publish` is ignored on the dead-letter _target_ queue, not about the source queue's capabilities (i.e. queue being dead-lettered _from_). This PR covers the source QQ and makes the QQ both reject and dead-letter a new publish which would otherwise overflow the queue.

PR #16689, opened 2026-06-15
Add configuration setting for max super stream partitions

acogoluegnes:
Introduce the `stream.max_super_stream_partitions` setting to allow cluster administrators to set an explicit cap on partition counts during super stream topology deployment.

Other changes included in this commit:
- Moved partition limits checking to the core stream manager logic, ensuring it is enforced consistently regardless of the incoming protocol.
- Optimized string token processing in the management handler and CLI by relying exclusively on native binary BIFs, significantly reducing the temporary memory overhead associated with parsing lengthy payloads.
- Implemented an early fail-fast validation structure within the management API endpoint to gracefully intercept and terminate requests before allocating memory blocks for sub-stream structures.

### DIFF
--- a/deps/rabbitmq_stream/priv/schema/rabbitmq_stream.schema
+++ b/deps/rabbitmq_stream/priv/schema/rabbitmq_stream.schema
@@ -144,3 +144,16 @@ fun(Conf) ->
         _         -> cuttlefish:invalid("should be a non-negative integer")
     end
 end}.
+
+{mapping, "stream.max_super_stream_partitions", "rabbitmq_stream.max_super_stream_partitions",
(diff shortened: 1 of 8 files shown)


the-mikedavis (approved):
Should we have a recommended default maximum? From what I've seen streams are less resource intensive when you use fewer, higher-throughput streams, since higher-throughput and especially batching drives larger chunks, and that decreases fixed per-chunk CPU costs.

Since the maximum here causes a validation failure maybe it's better to set it higher but in documentation we recommend using fewer?

michaelklishin (approved):

acogoluegnes:
@the-mikedavis We kept the 1,000-limit from the HTTP API as the default value, but this is a safeguard. I would not expect super streams to have more than a few dozens partitions.

PR #16705, opened 2026-06-17
UI: encrypt sensitive values that pass through local storage

MarcialRosales:
## Proposed Changes

Implement feature requests:
- RMQ-3752 (encrypted http basic credentials)
- RMQ-2544 (move all cookie management to server-side so that we can fully secure them)

The consequences of implementing the second change are the following:
- Removal of the loggedIn cookie: The loggedIn cookie has been completely removed from both the backend API and the client-side JavaScript. The Management UI no longer relies on cookies for session tracking.
- Client-side session enforcement via localStorage: The Management UI enforces session timeouts proactively. Before dispatching any HTTP request, the client verifies that the current browser time has not exceeded the session_expiry timestamp stored in localStorage during login.
- Server-side TTL for encrypted credentials: When RabbitMQ is configured to use encrypted basic credentials (credential_encryption_secret), the backend embeds an expiry timestamp inside the encrypted token payload. The server now actively enforces it during token decryption. This ensures that even if a user manipulates their localStorage timestamp, the stateless backend will securely reject the expired token with a 401 Unauthorized ("Session expired") response.

docs pr: https://github.com/rabbitmq/rabbitmq-website/pull/2543

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/rabbitmq_management/priv/schema/rabbitmq_management.schema
+++ b/deps/rabbitmq_management/priv/schema/rabbitmq_management.schema
@@ -504,6 +504,13 @@ end}.
 {mapping, "management.disable_basic_auth", "rabbitmq_management.disable_basic_auth",
     [{datatype, boolean}]}.
 
+%% When set, the management UI stores an AES-256-GCM encrypted bearer token
+%% (prefixed "rmqe.") instead of a plaintext base64(username:password).
+%% The value must be identical on every node in the cluster.
+{mapping, "management.credential_encryption_secret",
+          "rabbitmq_management.credential_encryption_secret",
+    [{datatype, [tagged_binary, binary]}]}.
+
 %% Management only
 
 {mapping, "management.disable_stats", "rabbitmq_management.disable_management_stats", [
--- a/.gitignore
+++ b/.gitignore
@@ -127,6 +127,9 @@ rabbitmq-server-*.tar.xz
 rabbitmq-server-*.zip
 secondary_dist/
 
+# RabbitMQ javascript files generated from EJS templates 
+deps/*/priv/www/js/*-ejs.js 
+
 # Trace tools output.
 *-ttb
 *.ti
--- a/.gitignore
+++ b/.gitignore
@@ -127,8 +127,8 @@ rabbitmq-server-*.tar.xz
 rabbitmq-server-*.zip
 secondary_dist/
 
-# RabbitMQ javascript files generated from EJS templates 
-deps/*/priv/www/js/*-ejs.js 
+# RabbitMQ javascript files generated from EJS templates
+deps/*/priv/www/js/*-ejs.js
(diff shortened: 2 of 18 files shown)


deps/rabbitmq_management/priv/schema/rabbitmq_management.schema
```diff
@@ -490,6 +490,13 @@ end}.
 {mapping, "management.disable_basic_auth", "rabbitmq_management.disable_basic_auth",
     [{datatype, boolean}]}.
 
+%% When set, the management UI stores an AES-256-GCM encrypted bearer token
+%% (prefixed "rmqe.") instead of a plaintext base64(username:password).
+%% The value must be identical on every node in the cluster.
+{mapping, "management.credentials_encryption_secret",
```
michaelklishin: Sorry to nitpick but technically `credential_encryption_secret` (in singular rather than plural) is correct.

Sorry :)

deps/rabbitmq_management/priv/schema/rabbitmq_management.schema
```diff
@@ -490,6 +490,13 @@ end}.
...
+          "rabbitmq_management.credentials_encryption_secret",
+    [{datatype, [tagged_binary, binary]}]}.
```
michaelklishin, line 512: Excellent catch, we do need to use `tagged_binary` here for `encrypted:{value}` support 👍

michaelklishin (changes requested):

deps/rabbitmq_management/priv/schema/rabbitmq_management.schema
michaelklishin: To give a decent example: we don't say "cars parking" even though a parking lot can and often does accommodate multiple vehicles.

We don't say "passwords policy" either, it's "password policy". I hope it makes at least some sense.

MarcialRosales: Nothing to be sorry about . Thank you for those corrections ! they are very welcome !

michaelklishin (changes requested):
This PR has the same regression we have fixed recently:
for OAuth 2-enabled deployments, `oauth_completeLogin` calls `set_token_auth` in a way that seemingly hardcodes the 8 hour timeout.

`selenium/test/basic-auth/rabbitmq.encrypted-credentials.conf:1`
still has `management.credentials_encryption_secret` ("credentials" in plural) vs. the rename singular Cuttlefish schema key.
This part I will correct myself since it's clear what should be done.

michaelklishin:
@MarcialRosales there was a code path (a user password change) that needed to adopt the same new function.

The rest of the changes were purely cosmetic.

michaelklishin (approved):

PR #16724, opened 2026-06-18
Add ETag/conditional request support to the definitions endpoint

carlhoerberg:
## Proposed Changes

Polling GET /api/definitions regenerates the entire payload on every request, even when nothing has changed. Add an ETag so that an up-to-date client receives a 304 Not Modified and the definitions are never built.

The validator is derived without building the response, from:

 * the product version, so an upgrade invalidates cached definitions
 * the requested scope (whole cluster vs a single virtual host)
 * the negotiated media type (JSON vs BERT)
 * a cheap metadata store version token

The store version token is the local Ra server's `{term, last_applied}` pair, exposed by the new `rabbit_khepri:metadata_store_version/0`. The applied index advances on every metadata write, so the token never fails to change when definitions change. The term is included so the token also changes across a store reset or member re-initialisation, which can restart the index from a lower value (a plain index could otherwise collide with an older state at the same index). Unrelated metadata writes may change the token too, which only costs an extra full response. When the Ra server cannot be queried cheaply (for example early during boot) `metadata_store_version/0` returns `unavailable` and `generate_etag/2` returns `undefined`, so the full response is served as before.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_khepri.erl
+++ b/deps/rabbit/src/rabbit_khepri.erl
@@ -131,6 +131,9 @@
          status/0,
          cli_cluster_status/0]).
 
+%% Cheap, monotonic store version used to derive HTTP cache validators.
+-export([metadata_store_version/0]).
+
 %% "Proxy" functions to Khepri query/update API.
 -export([is_empty/0,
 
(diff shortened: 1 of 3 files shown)


deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -1115,6 +1118,34 @@ get_ra_key_metrics(Node) ->
...
+%% may also change for unrelated metadata writes (false positives, which only
+%% cost an extra full response). `unavailable' is returned when Khepri is not
+%% the active store or the local Ra server cannot be queried cheaply.
+
+metadata_store_version() ->
+    case get_feature_state() of
+        enabled ->
+            ServerId = {?RA_CLUSTER_NAME, node()},
+            try ra:key_metrics(ServerId) of
+                #{last_applied := LastApplied} ->
```
gomoripeti: I wonder if raft term should also be included? raft index restarts on a leader election so "term=1, last_applied=10" is an older state then "term=2, last_applied=10"

lukebakken (commented):
I pointed the 🧞 at this. No big issues here, just a couple items.

---

## Worth a look (not blockers)

1. Strong ETag is shared across content-codings, and `cowboy_compress_h` is enabled by default on the management listener. RFC 7232 wants a strong validator to be unique per representation including content-coding. In practice the handler emits `Vary: accept-encoding` plus `Cache-Control: no-cache`, so a compliant cache keys on encoding and revalidates, but it is technically a strong-ETag inconsistency. Was this considered?

2. The token uses `last_applied` (consensus) while the body is read from the local projection, which is eventually consistent on a follower. There is a narrow window right after a write where `last_applied` has advanced but the local projection has not, so a first uncached read on a follower could cache a body for the previous state under the new ETag and serve 304s until the next write. Sub-millisecond and follower-only, so minor, but it is a behavior change from the old "always re-read" endpoint.

## Nit

In `generate_etag/2`, `not_found` (an unknown virtual host) maps to the `cluster` scope, while `to_json/2` answers the same request with a 400. This is harmless (a 304 is only returned when the client already holds the exact ETag, which it can only have obtained from a real response), but a one-line comment noting the intent would help future code readers.

michaelklishin:
Merged with minor fixes in https://github.com/rabbitmq/rabbitmq-server/pull/17121.

PR #16725, opened 2026-06-18
Rewrite the definitions export as one fold-based pass that uses HTTP response streaming

carlhoerberg:
## Proposed Changes

The HTTP definitions export had two problems at scale (tens of thousands of queues, hundreds of thousands of bindings):

 * the cluster-wide and per-virtual-host exports used different, heavy gathering code (the rabbit_mgmt_wm_* endpoints, rabbit_mgmt_format per item, a whitelist pass and rabbit_mgmt_format:prepare_for_encoding/1 over every item)
 * the whole document was built in memory before being sent, spiking the management process's memory and delaying the first byte until the entire body was assembled

Replace both endpoints with a single, scope-parameterised path (export/3, Scope :: all | {vhost, Name}):

 * Everything but queues and bindings is gathered as maps via rabbit_definitions' canonical builders (the same ones behind `rabbitmqctl export_definitions`), filtered to the virtual host when scoped.
 * Queues and bindings -- the categories that grow without bound -- are streamed via the new rabbit_definitions:fold_queues/3 and fold_bindings/3, which fold straight from the store (rabbit_db_queue:fold/2 over the queue projection and rabbit_db_binding:fold/2) without ever materialising the full lists.
 * The JSON response is streamed (cowboy_req:stream_reply/3 + stream_body/3) in batches, so peak memory is one batch rather than the whole document and the first bytes are sent immediately. BERT and HEAD keep a buffered fallback.

Because the export now uses rabbit_definitions throughout, the HTTP export produces the same document as `rabbitmqctl export_definitions`:

 * queues now carry their `type`
 * internal exchanges are no longer included (they are recreated by whatever declares them), so the exported `exchanges` array can be smaller
 * a virtual host is exported as `name`, `metadata` and `limits`; its description and tags live inside `metadata` rather than being duplicated at the top level

The single-virtual-host export still strips the redundant per-item `vhost` field. definitions_vhost_metadata_test is updated to the canonical shape.

New core helpers (additive): rabbit_db_queue:fold/2, rabbit_definitions:fold_queues/3 and rabbit_definitions:fold_bindings/3.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_db_queue.erl
+++ b/deps/rabbit/src/rabbit_db_queue.erl
@@ -22,6 +22,7 @@
          get_targets/1,
          get_all/0,
          get_all/1,
+         fold/2,
          get_all_by_type/1,
          get_all_by_type_and_vhost/2,
          get_all_by_type_and_node/3,
@@ -98,6 +99,29 @@ get_all() ->
               catch
                   error:badarg ->
                       []
+              end
+      end).
+
+-spec fold(Fun, Acc) -> Acc when
+      Fun :: fun((Queue :: amqqueue:amqqueue(), Acc) -> Acc),
+      Acc :: term().
+
+%% @doc Folds over all queue records without materialising them into a list,
+%% making it more memory-efficient than get_all/0 followed by a fold over the
+%% returned list. Useful when iterating over very large numbers of queues.
+%%
+%% @returns the fold accumulator.
+%%
+%% @private
+
+fold(Fun, Acc) ->
+    list_with_possible_retry(
+      fun() ->
+              try
+                  ets:foldl(Fun, Acc, ?KHEPRI_PROJECTION)
+              catch
+                  error:badarg ->
+                      Acc
               end
       end).
 
--- a/deps/rabbit/src/rabbit_definitions.erl
+++ b/deps/rabbit/src/rabbit_definitions.erl
@@ -48,7 +48,7 @@
          should_skip_if_unchanged/0
         ]).
 
--export([all_definitions/0]).
+-export([all_definitions/0, fold_queues/3, fold_bindings/3]).
 -export([
   list_users/0, list_vhosts/0, list_permissions/0, list_topic_permissions/0,
   list_runtime_parameters/0, list_global_runtime_parameters/0, list_policies/0,
@@ -295,6 +295,47 @@ all_definitions() ->
         bindings          => Bs,
         exchanges         => Xs
     }.
+
+-spec fold_queues(Scope, Fun, Acc) -> Acc when
+      Scope :: all | {vhost, vhost:name()},
+      Fun :: fun((Queue :: map(), Acc) -> Acc),
+      Acc :: term().
+%% @doc Folds over exported queues as definition maps, optionally restricted to
+%% a single virtual host, without materialising the full queue list. Exclusive
+%% queues are skipped, as they cannot be restored. See also list_queues/0.
+fold_queues(Scope, Fun, Acc) ->
+    rabbit_db_queue:fold(
+      fun(Q, Acc0) ->
+              case amqqueue:get_exclusive_owner(Q) of
+                  none ->
+                      #resource{virtual_host = VHost} = amqqueue:get_name(Q),
+                      case Scope of
+                          all            -> Fun(queue_definition(Q), Acc0);
+                          {vhost, VHost} -> Fun(queue_definition(Q), Acc0);
+                          _              -> Acc0
+                      end;
+                  _ ->
+                      Acc0
+              end
+      end, Acc).
+
+-spec fold_bindings(Scope, Fun, Acc) -> Acc when
+      Scope :: all | {vhost, vhost:name()},
+      Fun :: fun((Binding :: map(), Acc) -> Acc),
+      Acc :: term().
+%% @doc Folds over exported (explicit) bindings as definition maps, optionally
+%% restricted to a single virtual host, without materialising the full binding
+%% list, which matters on data sets with very many bindings. See also
+%% list_bindings/0.
+fold_bindings(Scope, Fun, Acc) ->
+    rabbit_db_binding:fold(
+      fun(#binding{source = #resource{virtual_host = VHost}} = Binding, Acc0) ->
+              case Scope of
+                  all            -> Fun(binding_definition(Binding), Acc0);
+                  {vhost, VHost} -> Fun(binding_definition(Binding), Acc0);
+                  _              -> Acc0
+              end
+      end, Acc).
 
 -spec has_configured_definitions_to_load() -> boolean().
 has_configured_definitions_to_load() ->
--- a/deps/rabbitmq_management/src/rabbit_mgmt_wm_definitions.erl
+++ b/deps/rabbitmq_management/src/rabbit_mgmt_wm_definitions.erl
@@ -12,8 +12,6 @@
 -export([accept_multipart/2]).
 -export([variances/2]).
 
--import(rabbit_misc, [pget/2]).
-
 -include("rabbit_mgmt.hrl").
 -include_lib("rabbitmq_management_agent/include/rabbit_mgmt_records.hrl").
 -include_lib("amqp_client/include/amqp_client.hrl").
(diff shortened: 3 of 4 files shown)


lhoguin (approved):
Sorry wrong PR!!

lhoguin:
I cannot cancel the PR approve, sorry about that. I plan to review this tomorrow. Thanks!!

lukebakken (commented):
This review was performed by an AI with guidance from @lukebakken.

## Review: fold-based, streamed definitions export

The AI and I reviewed the full diff against `main`, traced the new fold paths through `rabbit_db_queue`, `rabbit_db_binding`, Khepri and Ra, and checked the cowboy streaming contract (including how back-pressure and `data_ack` actually work) and the affected tests. We came away with one concern we would like a maintainer to confirm before merge, one behavioral change worth a deliberate decision, and several smaller items. None of this detracts from the overall direction, which we think is the right one.

## Streaming bindings can park the Khepri Ra server process under client back-pressure

`stream_export/4` folds bindings straight into the socket:

```erlang
stream_folded(<<"bindings">>, fun rabbit_definitions:fold_bindings/3, Scope, ReqData)
```

`fold_bindings/3` calls `rabbit_db_binding:fold/2`, and unlike the queue fold (which is `ets:foldl` over the local projection and runs in the caller), the binding fold goes through `rabbit_khepri:fold/3`:

```erlang
fold(Fun, Acc) ->
    {ok, Res} = rabbit_khepri:fold(Path, fun(_, #{data := Set}, Acc0) -> ... Fun(...) ... end, Acc),
```

Tracing that: `rabbit_khepri:fold` calls `khepri_machine:fold`, which calls `ra:aux_command`, which is `gen_statem:call({StoreId, node()}, {aux_command, ...})`. The query function, and therefore the fold callback passed in by `stream_folded/4`, executes inside the Khepri Ra server process via `handle_aux/5` and `perform_delayed_aux_query/3` (`QueryFun(State)`). That callback calls `cowboy_req:stream_body/3`, which parks the calling process on `receive {data_ack, Pid}`.

It is worth being precise about what `stream_body/3` blocks on, because it is not as bad as it first looks. `data_ack` is not a TCP/socket acknowledgement, and under normal conditions it does not block per chunk. It is sent by the cowboy connection process (the `cowboy_stream_h` stream handler), not by the request-handler process where the export code runs. In the `normal` state the connection process acks immediately on receipt of the `{data, Pid, ...}` message:

```erlang
info(StreamID, Data0={data, Pid, _, _}, State0=#state{stream_body_status=Status}) ->
    State = case Status of
        normal   -> Pid ! {data_ack, self()}, State0;        %% acked on receipt, not on socket write
        blocking -> State0#state{stream_body_pid=Pid, stream_body_status=blocked};
        blocked  -> State0
    end,
```

So with a fast-reading client there is no socket round-trip per chunk and the Ra process would not stall meaningfully. The concern is conditional on back-pressure: when cowboy's send buffer fills (a slow or non-reading client), the `connection_buffer_full` / `stream_buffer_full` alarm turns on, `stream_body_status` moves `normal -> blocking -> blocked`, and the ack for the in-flight chunk is withheld. The producing process then stays parked in `receive {data_ack, Pid}` until the alarm turns off:

```erlang
info(StreamID, {alarm, Name, off}, State=#state{stream_body_pid=Pid, stream_body_status=Status})
        when Name =:= connection_buffer_full; Name =:= stream_buffer_full ->
    _ = case Status of
        blocked -> Pid ! {data_ack, self()};
        ...
```

For the bindings stream specifically, that parked process is the Khepri Ra server, blocked inside `handle_aux` for as long as the back-pressure persists, and the aux query timeout defaults to `infinity`. Khepri is the cluster-wide metadata store on the hot path for every queue, exchange, binding, virtual host, policy and permission change. If our reading of the call path is correct, a slow or stalled client pulling the definitions export could therefore delay metadata-store operations cluster-wide. That would partly work against the goal of the PR, which is to take load off the management process rather than place it on a more critical shared one. We want to flag this clearly rather than overstate it: it is conditional on back-pressure, not incurred on every request.

The queue stream does not have this problem because the queue fold is `ets:foldl` over the local projection and runs in the caller. The buffered path (BERT, HEAD) also avoids it: it folds into a list with no socket I/O inside the callback and returns, so the Ra process is busy only for the in-memory fold, not for the duration of a client download.

Recommendations, in order of preference:

1. Stream bindings from the local ETS projection the way queues are streamed. `rabbit_db_binding` already reads `?KHEPRI_BINDINGS_PROJECTION` (`rabbit_khepri_binding`) via `ets:tab2list` in `get_all/0`; add an `ets:foldl`-based `fold/2` mirroring `rabbit_db_queue:fold/2`, so the callback runs in the caller, not the Ra server. This keeps memory bounded and keeps socket I/O out of Khepri.
2. If the store-backed fold must be kept, do not perform socket I/O inside it. Fold into an accumulator (or chunk-sized buffers) and stream outside the fold. That loses the "never materialise" property, so option 1 is better.

We would value a maintainer's confirmation of the call path here, and ideally @dumbbell's read given the Khepri internals involved.

## Behavioral change: bindings are no longer filtered against the exported destination set

Old `export_binding/2`:

```erlang
Src =/= <<"">> andalso
  ( (DestType =:= queue andalso lists:member({Dest, VHost}, Qs))
    orelse (DestType =:= exchange andalso Dest =/= <<"">>) )
```

It dropped bindings whose destination queue was not in the exported queue set, most importantly bindings to exclusive queues, which the export deliberately skips. The new path streams every explicit binding from `binding_definition/1` with no such cross-check.

Result: the new export can include bindings whose destination is an exclusive (or otherwise non-exported) queue. On re-import, `rabbit_definitions:add_binding_int/4` calls `rabbit_binding:add/2` and throws on `{error, _}`, so a binding to a destination that does not exist will fail the import rather than be silently skipped.

The default-exchange guard (`Src =/= <<"">>`) is effectively preserved, since implicit bindings are not explicit routes and so are not produced by `binding_definition/1`. The exclusive-queue case is the gap. We note that `rabbit_definitions:all_definitions/0` (the path behind `rabbitmqctl export_definitions`) appears to have the same gap, so this may be a pre-existing inconsistency that the HTTP path is now inheriting rather than something new to this PR. Either way, it seems worth a deliberate decision on whether the export should filter these out, and a dedicated test: declare an exclusive queue with a binding to it, export, and assert the binding is absent.

## Smaller issues

1. `description` is duplicated then effectively redundant in the single-vhost export. `small_parts({vhost, _})` sets both `metadata => vhost:get_metadata(VHost)` (which already contains `description`) and a separate top-level `description => vhost:get_description(VHost)`. The cluster-wide vhost shape (`vhost_definition/1`) carries only `name`/`limits`/`metadata`. The updated `definitions_vhost_metadata_test` documents that description should live inside metadata, yet the single-vhost top level still emits it redundantly. Minor inconsistency worth removing for true parity.

2. `original_cluster_name` is absent from the single-vhost export. The old `vhost_definitions/2` did not include it either, so this is preserved. Just confirming it is intentional.

3. `stream_value/2` proplist vs empty-map ambiguity. The comment notes that non-list values like `limits` (a proplist) are encoded whole by `rabbit_json`, which is correct for a non-empty proplist. An empty `limits` is `[]`, however, which matches the `stream_value([], _)` clause and is emitted as `[]` (a JSON array), whereas a populated `limits` proplist encodes as a JSON object. So an empty vhost `limits` serializes as `[]` and a non-empty one as `{...}`, an inconsistent type for the same field. This is pre-existing rather than introduced here: the old path also ran the whole proplist through `rabbit_json:encode`, and an empty `[]` encodes as `[]` there too. We mention it only because the rewrite is a natural opportunity to normalise it, and it is worth confirming importers tolerate `[]`. This affects only the single-vhost top-level `limits`; the per-vhost `limits` inside the `vhosts` array (cluster export) is a map element and goes through the array path.

4. Stale doc comments on the exported folds. `rabbit_db_queue:fold/2` is documented `@private` but is now exported and called cross-module from `rabbit_definitions` (a different app). Similarly `rabbit_db_binding:fold/2` still says "Just used by prometheus_rabbitmq_core_metrics_collector". Update both.

5. Test coverage is thin for the headline feature. The only adjusted test is `definitions_vhost_metadata_test`. There is no test that exercises the streamed GET path with more than `?STREAM_BATCH_SIZE` (2000) queues/bindings to catch batch-boundary comma bugs, verifies BERT and HEAD still work through the new `buffered/2` fallback, or round-trips a large export through import. The comma-placement logic in `encode_batch` / `stream_folded` is fiddly and entirely new, so a batch-boundary test (for example 2001 queues) is worth adding.

6. `list_with_possible_retry` reuse with `fold`. The retry wrapper treats `[]` as "possibly empty due to a record-version mismatch, retry". `rabbit_db_queue:fold/2` reuses this wrapper, and the accumulator flows through it: if a caller ever passes a non-list accumulator, the `[] ->` clause cannot match and the retry is silently skipped. Current callers pass list or tuple accumulators, so this is correct today; it is just a sharp edge worth a short comment so a future caller does not assume the retry applies to all accumulator shapes.

## What looks correct

- Removal of the `pget` import and the `retain_whitelisted`, `strip_vhost`, and `export_*` helpers leaves no dangling references (grep-verified).
- The `{stop, ReqData, Context}` return correctly signals cowboy_rest that the response was already sent (`set_resp_body/2` handles `{stop, ...}` via `terminate/2`), and `rabbit_mgmt_util:reply/3` already passes `{stop, _, _}` through.
- The content-type header is set by cowboy_rest before the provide callback runs (`set_content_type/2` precedes `set_resp_body/2`), so `stream_reply(200, #{}, ...)` inherits it. The comment is accurate.
- BERT and HEAD are correctly diverted to the buffered path: HEAD must not stream a body and BERT cannot be chunked.
- JSON structure: `{`, comma-joined small parts, `,"queues":[...]`, `,"bindings":[...]`, `}`. The comma logic is consistent because `small_parts` is provably non-empty (metadata keys are always present).

## Summary

To be clear, we think this is a worthwhile change and the implementation is solid. Our two main points are the binding fold running inside the Khepri Ra server process under back-pressure, and the exclusive-queue binding filter that the previous code applied. We would value a maintainer's read on the first in particular, since the cluster-wide impact depends on details of the Khepri and Ra internals that you know far better than we do.

michaelklishin (changes requested):
This streaming code is executed in the (Khepri's) Ra machine process (`ra_server_proc`) in the new `fold/2` function.

Which means that the entire metadata database is blocked waiting for an HTTP API socket I/O.

michaelklishin:
The core idea yields a very meaningful improvement: a 10x to 18x improvement in peak memory footprint [1], about eight times lower peak CPU use and a 7 to 10x faster HTTP response transfer time.

However, we must fix the Ra server process blocking because 200-300 ms metadata store blocking is very feasible even with this significant reduction in total response transfer time.

1. attributed to the definitions export operation on a file with 10K queues and 20K bindings

michaelklishin:
The Ra process blocking and a few much less important bits were addressed in https://github.com/rabbitmq/rabbitmq-server/pull/16734.

michaelklishin:
The improvement particularly matters for clusters with a lot of bindings because the root cause was a quadratic operation in the binding filtering step. That code had been around since 2010-2011.

PR #16734, opened 2026-06-18
Make definition export over the HTTP API more efficient by combining incremental encoding with HTTP response streaming

michaelklishin:
This is #16725 without a number of issues, most notably its new `fold/2` implementation for bindings that was executed in the Khepri's Ra machine process, potentially blocking the metadata store for up to hundreds of milliseconds every time a definition export was requested over the HTTP API.

The improvements are meaningful: up to 10x or 18x reduction in transfer time and peak resource (CPU, memory) footprint.

The improvement particularly matters for clusters with a lot of bindings because the root cause was a quadratic operation in the binding filtering step. That code had been around since 2010-2011.

### DIFF
--- a/deps/rabbit/src/rabbit_db_binding.erl
+++ b/deps/rabbit/src/rabbit_db_binding.erl
@@ -507,7 +507,8 @@ get_all_for_destination(Dst) ->
 %% and folding over the returned binding list.
 %%
 %% Just used by prometheus_rabbitmq_core_metrics_collector to iterate over the
-%% bindings.
+%% bindings. Runs the callback inside the Khepri server process; for callbacks
+%% that may block (e.g. streaming), use `fold_projection/2` instead.
 %%
 %% @returns the fold accumulator.
 %%
--- a/deps/rabbit/src/rabbit_db_binding.erl
+++ b/deps/rabbit/src/rabbit_db_binding.erl
@@ -20,7 +20,8 @@
          get_all/3,
          get_all_for_source/1,
          get_all_for_destination/1,
-         fold/2]).
+         fold/2,
+         fold_projection/2]).
 
 %% Routing. These functions are in the hot code path
 -export([match/2, match_routing_key/3]).
@@ -377,6 +378,28 @@ get_all() ->
     catch
         error:badarg ->
             []
+    end.
+
+-spec fold_projection(Fun, Acc) -> Acc when
+      Fun :: fun((Binding :: rabbit_types:binding(), Acc) -> Acc),
+      Acc :: any().
+%% @doc Folds over all explicit binding records from the local projection,
+%% without building an intermediate list.
+%%
+%% Unlike `fold/2`, this runs in the calling process rather than the Khepri
+%% server process, so the callback may block (e.g. on streaming backpressure)
+%% without stalling the metadata store.
+%%
+%% @private
+
+fold_projection(Fun, Acc) ->
+    try
+        ets:foldl(
+          fun(#route{binding = Binding}, Acc0) -> Fun(Binding, Acc0) end,
+          Acc, ?KHEPRI_BINDINGS_PROJECTION)
+    catch
+        error:badarg ->
+            Acc
     end.
 
 -spec get_all(VHostName) -> [Binding] when
--- a/deps/rabbit/src/rabbit_db_queue.erl
+++ b/deps/rabbit/src/rabbit_db_queue.erl
@@ -22,6 +22,7 @@
          get_targets/1,
          get_all/0,
          get_all/1,
+         fold/2,
          get_all_by_type/1,
          get_all_by_type_and_vhost/2,
          get_all_by_type_and_node/3,
@@ -98,6 +99,29 @@ get_all() ->
               catch
                   error:badarg ->
                       []
+              end
+      end).
+
+-spec fold(Fun, Acc) -> Acc when
+      Fun :: fun((Queue :: amqqueue:amqqueue(), Acc) -> Acc),
+      Acc :: term().
+
+%% @doc Folds over all queue records without materialising them into a list,
+%% making it more memory-efficient than get_all/0 followed by a fold over the
+%% returned list. Useful when iterating over very large numbers of queues.
+%%
+%% @returns the fold accumulator.
+%%
+%% @private
+
+fold(Fun, Acc) ->
+    list_with_possible_retry(
+      fun() ->
+              try
+                  ets:foldl(Fun, Acc, ?KHEPRI_PROJECTION)
+              catch
+                  error:badarg ->
+                      Acc
               end
       end).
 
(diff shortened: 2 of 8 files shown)


lukebakken (commented):
This review was performed by an AI with guidance from @lukebakken.

## Review: incremental encoding plus HTTP response streaming for definition export

Thank you for picking up the points from #16725 so quickly. We re-reviewed the full diff against `main`, with particular attention to the binding fold (our main concern on the previous PR) and to anything new the refactor introduces. The short version: the two substantive issues we raised are addressed well, the smaller items are mostly handled, and we found one new behavioral change worth a conscious decision plus a couple of minor observations.

## Previously raised points, as addressed here

1. Binding fold no longer runs in the Khepri Ra server process. The new `rabbit_db_binding:fold_projection/2` folds over the local `?KHEPRI_BINDINGS_PROJECTION` ETS table with `ets:foldl`, so the streaming callback (and its `cowboy_req:stream_body/3` back-pressure wait) runs in the calling process. `fold_bindings/3` now calls `fold_projection/2` rather than `fold/2`. This resolves the metadata-store stall we were most concerned about. The doc comment on `fold_projection/2` calls out the distinction from `fold/2` explicitly, which is helpful.

2. Exclusive-queue bindings. The commit "Definitions export: align what queue sets CLI and HTTP routes export" makes this a deliberate decision: the HTTP export now matches `rabbit_definitions:all_definitions/0` (the CLI `rabbitmqctl export_definitions`) rather than re-deriving its own filtered set. That is a reasonable resolution. One thing still worth confirming: `fold_bindings/3` continues to emit explicit bindings whose destination is an exclusive queue (exclusive queues are skipped by `fold_queues/3`), and `rabbit_definitions:add_binding_int/4` throws on `{error, _}` from `rabbit_binding:add/2`. So an exported-then-reimported document that contains such a binding would still fail import. If the intent is parity with the CLI export, this is pre-existing and consistent; we only flag it so it is a known, accepted property rather than a surprise.

3. Empty `limits` now encodes as a JSON object. `vhost_definition/1` and the single-vhost `small_parts/1` both wrap `vhost:get_limits/1` in `maps:from_list/1`, so an empty `limits` serializes as `{}` rather than `[]`. The new `definitions_large_streamed_test` and the updated `definitions_vhost_metadata_test` both assert `is_map(limits)`, which locks the behaviour in. Nicely done.

4. Stale `@private` on the queue fold. The `list_with_possible_retry/1` wrapper was removed from `rabbit_db_queue:fold/2` with a clear comment ("the callback may have side effects and the accumulator is arbitrary, so the fold must never be replayed"), which is exactly the replay concern we raised, addressed at the root.

5. Batch-boundary test coverage. `definitions_large_streamed_test` exercises 2200 queues and bindings (across the 2000-element `?STREAM_BATCH_SIZE` boundary) and asserts each item appears exactly once, that queues carry `type`, and that the scoped export strips the per-item `vhost`. This is the test we hoped for and covers the fiddly comma logic in `encode_batch`/`stream_folded`.

The `description` duplication in the single-vhost top level (our smaller item 1) is still present: `small_parts({vhost, _})` sets both `metadata` (which contains `description`) and a separate top-level `description`. This is harmless and matches prior behaviour, so we are not asking for a change, just noting it remained.

## New observation: `list_queues/0` changed its data source, which also affects the CLI export

This is the one item we would ask you to look at, because it is a behavioral change beyond the HTTP path.

Previously:

```erlang
list_queues() ->
    %% exclude exclusive queues, they cannot be restored
    [queue_definition(Q) || Q <- lists:filter(fun(Q0) ->
                                                amqqueue:get_exclusive_owner(Q0) =:= none
                                              end,
                                              rabbit_amqqueue:list())].
```

Now:

```erlang
list_queues() ->
    lists:reverse(fold_queues(all, fun(Q, Acc) -> [Q | Acc] end, [])).
```

`fold_queues/3` folds the raw `?KHEPRI_PROJECTION` ETS table via `rabbit_db_queue:fold/2`. The old path went through `rabbit_amqqueue:list/0`, which applies an additional filter:

```erlang
list() ->
    All = rabbit_db_queue:get_all(),
    NodesRunning = rabbit_nodes:list_running(),
    lists:filter(fun (Q) ->
                         Pid = amqqueue:get_pid(Q),
                         St = amqqueue:get_state(Q),
                         St =/= stopped orelse is_local_to_node_set(Pid, NodesRunning)
                 end, All).
```

So `list_queues/0` previously dropped queues in state `stopped` whose pid is not local to a running node, and now it does not. Because `list_queues/0` backs `all_definitions/0`, this changes what `rabbitmqctl export_definitions` exports, not only the HTTP API. For a definitions export, including all durable queue records regardless of where they currently run is arguably the more correct behaviour (you generally want to back up every declared queue, not just the ones whose owning node is up right now). But it is a change in observable output, so we wanted to surface it rather than let it ride in silently. If it is intended, a line in the PR description or a test asserting the new inclusiveness would make it explicit; if not, `fold_queues/3` could filter on the same running-node condition for the `all` scope.

## Minor observations

- `rabbit_db_binding:fold/2` (the original Khepri-server-process fold) is retained and still used by `prometheus_rabbitmq_core_metrics_collector` for `exchange_bindings`. That collector runs the fold inside the Ra server process on every scrape. It is out of scope for this PR and the per-scrape cost is bounded (it only counts, with no I/O in the callback), but now that `fold_projection/2` exists, that call site looks like a candidate for the same treatment in a future change. Just noting the asymmetry; no action needed here.
- The `fold/2` doc comment in `rabbit_db_binding` still reads "Just used by prometheus_rabbitmq_core_metrics_collector". With `fold_projection/2` now alongside it, a one-line note pointing readers to `fold_projection/2` for the streaming/back-pressure case would help the next person choose correctly.

## What we verified

- The streaming JSON structure is well-formed: `{`, comma-joined small parts, `,"queues":[...]`, `,"bindings":[...]`, `}`, with the leading comma safe because `small_parts/1` always contains metadata keys.
- HEAD now short-circuits before gathering any definitions and returns an empty body with `cache-control: no-cache`, which is correct and avoids the wasted work.
- BERT remains on the buffered path (it cannot be chunked incrementally); GET+JSON streams.
- Removal of the `pget` import and the `retain_whitelisted`, `strip_vhost`, and `export_*` helpers leaves no dangling references.
- `fold_queues/3` applies both the scope filter and the exclusive-queue skip, and `fold_bindings/3` applies the scope filter, inside the fold itself, so the buffered and streamed paths produce the same set.

## Summary

This addresses the substantive concerns from #16725, especially the Khepri Ra process back-pressure issue, and adds the batch-boundary test we hoped to see. The one item worth a deliberate call is the `list_queues/0` data-source change, since it alters the CLI export's output for stopped/non-local queues.

michaelklishin:
@lukebakken thank you for the reviews but can you please tone down the verbosity of those reports? Ask for a "terse, factual description of the findings" or something like that.

michaelklishin:
Yes, including stopped durable queues is a positive side-efect. Let's keep it.

They are not exclusive queues, restoring them in a new cluster makes semantic sense.

lukebakken:
@michaelklishin I just noticed your comment! I'll keep things less verbose.

PR #16735, opened 2026-06-19
Consider HTTP API (UI) prefix when setting cookie path #16665

michaelklishin:
This is a follow-up to #16665 that was discovered as part of #16734 by the Selenium suite run.


michaelklishin:
@MarcialRosales this is an FYI. The change fixed a Selenium suite and a real gap in #16665.
Sorry if you might have wasted any time looking at those failures post-#16665.

PR #16739, opened 2026-06-19
QQ AZ awareness

mkuratczyk:
Implements AZ-aware membership placement for quorum queues.

Until now, if there's more nodes than QQ members requested, we put one member locally (where the queue is declared) and the other members would be assigned randomly to other nodes, preferring nodes running at the time of declaration.

With this change, you can tag nodes using the existing node_tags feature and specify the tag used as the az information.
For example:
```
node_tags.az = az1
```

AZ awareness can be enabled:
* on the cluster level, using: `rabbit.quorum_queue_member_placement_tag=az`
* on the queue argument level: `member-placement-tag=az`
* using a policy/operator policy: `x-member-placement-tag=az`

When either of these is in effect, we'll make sure that members are evenly spread between nodes with different tag values.

For example, with 7 nodes, 3 tagged with `az = az1`, 2 with `az = az2` and 2 with `az = az3`, all 3-member quorum queues will have one member in az1, one in az2 and one in az3. The distribution of members between nodes within the same AZ remains as before: up to the limit (default of 1000 queues), we make it perfectly even, and if there's more queues - we start to assign nodes randomly (still taking AZ into consideration to be clear).

### DIFF
--- a/deps/rabbit/include/rabbit_khepri.hrl
+++ b/deps/rabbit/include/rabbit_khepri.hrl
@@ -24,6 +24,9 @@
 -define(RABBITMQ_KHEPRI_MAINTENANCE_PATH(Node),
         ?RABBITMQ_KHEPRI_ROOT_PATH([node_maintenance, Node])).
 
+-define(RABBITMQ_KHEPRI_NODE_METADATA_PATH(Node),
+        ?RABBITMQ_KHEPRI_ROOT_PATH([node_metadata, Node])).
+
 -define(RABBITMQ_KHEPRI_GLOBAL_RUNTIME_PARAM_PATH(Key),
         ?RABBITMQ_KHEPRI_ROOT_PATH([runtime_params, Key])).
 
@@ -62,3 +65,8 @@
 
 -define(RABBITMQ_KHEPRI_QUEUE_PATH(VHost, Name),
         ?RABBITMQ_KHEPRI_VHOST_PATH(VHost, [queues, Name])).
+
+-record(node_metadata, {
+          node,
+          metadata = #{}
+        }).
--- a/deps/rabbit/include/rabbit_khepri.hrl
+++ b/deps/rabbit/include/rabbit_khepri.hrl
@@ -24,8 +24,13 @@
 -define(RABBITMQ_KHEPRI_MAINTENANCE_PATH(Node),
         ?RABBITMQ_KHEPRI_ROOT_PATH([node_maintenance, Node])).
 
+-define(RABBITMQ_KHEPRI_NODE_PATH(Node),
+        ?RABBITMQ_KHEPRI_NODE_PATH(Node, [])).
+-define(RABBITMQ_KHEPRI_NODE_PATH(Node, Rest),
+        ?RABBITMQ_KHEPRI_ROOT_PATH([nodes, Node | Rest])).
+
 -define(RABBITMQ_KHEPRI_NODE_METADATA_PATH(Node),
-        ?RABBITMQ_KHEPRI_ROOT_PATH([node_metadata, Node])).
+        ?RABBITMQ_KHEPRI_NODE_PATH(Node, [node_metadata])).
 
 -define(RABBITMQ_KHEPRI_GLOBAL_RUNTIME_PARAM_PATH(Key),
         ?RABBITMQ_KHEPRI_ROOT_PATH([runtime_params, Key])).
--- a/deps/rabbit/include/rabbit_khepri.hrl
+++ b/deps/rabbit/include/rabbit_khepri.hrl
@@ -70,8 +70,3 @@
 
 -define(RABBITMQ_KHEPRI_QUEUE_PATH(VHost, Name),
         ?RABBITMQ_KHEPRI_VHOST_PATH(VHost, [queues, Name])).
-
--record(node_metadata, {
-          node,
-          metadata = #{}
-        }).
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -982,7 +982,8 @@ declare_args() ->
      {<<"x-stream-max-segment-size-bytes">>, fun check_non_neg_int_arg/2},
      {<<"x-stream-filter-size-bytes">>, fun check_non_neg_int_arg/2},
      {<<"x-initial-cluster-size">>, fun check_initial_cluster_size_arg/2},
-     {<<"x-queue-leader-locator">>, fun check_queue_leader_locator_arg/2}].
+     {<<"x-queue-leader-locator">>, fun check_queue_leader_locator_arg/2},
+     {<<"x-member-placement-tag">>, fun check_member_placement_tag_arg/2}].
 
 consume_args() -> [{<<"x-priority">>,              fun check_int_arg/2},
                    {<<"x-consumer-timeout">>,      fun check_non_neg_int_arg/2},
@@ -1181,6 +1182,17 @@ check_queue_leader_locator_arg(Val, _Args) when is_binary(Val) ->
     end;
 check_queue_leader_locator_arg(_Val, _Args) ->
     {error, invalid_queue_locator_arg}.
+
+check_member_placement_tag_arg({longstr, Val}, _Args) when byte_size(Val) > 0 ->
+    ok;
+check_member_placement_tag_arg({longstr, _}, _Args) ->
+    {error, invalid_member_placement_tag_arg};
+check_member_placement_tag_arg({Type, _}, _Args) ->
+    {error, {unacceptable_type, Type}};
+check_member_placement_tag_arg(Val, _Args) when is_binary(Val), byte_size(Val) > 0 ->
+    ok;
+check_member_placement_tag_arg(_Val, _Args) ->
+    {error, invalid_member_placement_tag_arg}.
 
 check_stream_offset_arg(Val, _Args) ->
     case rabbit_stream_queue:parse_offset_arg(Val) of
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -1189,7 +1189,7 @@ check_member_placement_tag_arg({longstr, _}, _Args) ->
     {error, invalid_member_placement_tag_arg};
 check_member_placement_tag_arg({Type, _}, _Args) ->
     {error, {unacceptable_type, Type}};
-check_member_placement_tag_arg(Val, _Args) when is_binary(Val), byte_size(Val) > 0 ->
+check_member_placement_tag_arg(Val, _Args) when is_binary(Val) andalso byte_size(Val) > 0 ->
     ok;
 check_member_placement_tag_arg(_Val, _Args) ->
     {error, invalid_member_placement_tag_arg}.
--- a/deps/rabbit/src/rabbit_db.erl
+++ b/deps/rabbit/src/rabbit_db.erl
@@ -74,6 +74,9 @@ init() ->
                "DB: initialization successful",
                #{domain => ?RMQLOG_DOMAIN_DB}),
 
+            LocalTags = application:get_env(rabbit, node_tags, []),
+            rabbit_db_node_metadata:set(node(), #{node_tags => LocalTags}),
+
             init_finished(),
 
             ok;
--- a/deps/rabbit/src/rabbit_db.erl
+++ b/deps/rabbit/src/rabbit_db.erl
@@ -75,7 +75,7 @@ init() ->
                #{domain => ?RMQLOG_DOMAIN_DB}),
 
             LocalTags = application:get_env(rabbit, node_tags, []),
-            rabbit_db_node_metadata:set(node(), #{node_tags => LocalTags}),
+            _ = rabbit_db_node_metadata:set(node(), #{node_tags => LocalTags}),
 
             init_finished(),
 
--- a/deps/rabbit/src/rabbit_db_cluster.erl
+++ b/deps/rabbit/src/rabbit_db_cluster.erl
@@ -249,6 +249,7 @@ forget_member_using_khepri(_Node, true) ->
        #{domain => ?RMQLOG_DOMAIN_DB}),
     {error, not_supported};
 forget_member_using_khepri(Node, false = _RemoveWhenOffline) ->
+    _ = rabbit_db_node_metadata:delete(Node),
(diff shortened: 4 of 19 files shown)


deps/rabbit/src/rabbit_member_placement.erl
```diff
@@ -0,0 +1,23 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
+%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%
+%% Copyright (c) 2007-2026 Broadcom. All Rights Reserved. The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries. All rights reserved.
+%%
+
+-module(rabbit_member_placement).
```
michaelklishin, line 8: Is the name intentionally generic or should it be `rabbit_quorum_placement`, `rabbit_quorum_queue_member_placement` (or something like that, specific to QQs).

We already have quite a few `rabbit_quorum_*` modules.

mkuratczyk: Yes, I assume at least delayed queues will use this placement strategy as well

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1500,59 +1514,73 @@ status(#resource{virtual_host = Vhost, name = QueueName}) ->
...
+                            {<<"Last Applied">>, <<>>},
+                            {<<"Commit Index">>, <<>>},
+                            {<<"Snapshot Index">>, <<>>},
+                            {<<"Term">>, <<>>},
+                            {<<"Machine Version">>, <<>>}
+                           ]
+                   end,
+         case TagKey of
+             undefined -> BaseRow;
+             _         -> AZ = case maps:get(N, NodeToAZ, undefined) of
```
kjnilsson: redundant whitespace

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1500,59 +1514,73 @@ status(#resource{virtual_host = Vhost, name = QueueName}) ->
...
+                                   undefined -> <<>>;
+                                   Val       -> Val
```
kjnilsson: redundant whitespace

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1182,6 +1183,17 @@ check_queue_leader_locator_arg(Val, _Args) when is_binary(Val) ->
 check_queue_leader_locator_arg(_Val, _Args) ->
     {error, invalid_queue_locator_arg}.
 
+check_member_placement_tag_arg({longstr, Val}, _Args) when byte_size(Val) > 0 ->
+    ok;
+check_member_placement_tag_arg({longstr, _}, _Args) ->
+    {error, invalid_member_placement_tag_arg};
+check_member_placement_tag_arg({Type, _}, _Args) ->
+    {error, {unacceptable_type, Type}};
+check_member_placement_tag_arg(Val, _Args) when is_binary(Val), byte_size(Val) > 0 ->
```
kjnilsson: remind me does the comma mean `and` or `or`?

deps/rabbit/src/rabbit_member_placement_az.erl
```diff
@@ -0,0 +1,190 @@
...
+    MinAZs = [AZ || AZ := C <- AZCounts, C =:= MinCount],
+    InMinAZs = [N || N <- CandidateNodes,
+                     lists:member(maps:get(N, NodeToAZ, undefined), MinAZs)],
+    case InMinAZs of
+        [] ->
+            %% All nodes from the minimum AZs are already selected; use any remaining node.
+            CandidateNodes;
+        _ ->
+            case [N || N <- InMinAZs, lists:member(N, RunningNodes)] of
+                []      -> InMinAZs;
```
kjnilsson: redundant whitespace

deps/rabbit/src/rabbit_member_placement_az.erl
```diff
@@ -0,0 +1,190 @@
...
+                Running -> Running
+            end
+    end.
+
+pick_balanced(Pool, Counters) ->
+    hd(lists:sort(fun(A, B) -> maps:get(A, Counters, 0) =< maps:get(B, Counters, 0) end, Pool)).
+
+%% Running nodes first so that AZ-aware selection prefers them within each AZ.
+running_first(Nodes, RunningNodes) ->
+    [N || N <- Nodes,     lists:member(N, RunningNodes)]
```
kjnilsson: whitespace

deps/rabbit/src/rabbit_member_placement_az.erl
```diff
@@ -0,0 +1,190 @@
...
+       member_nodes := [node()],
+       member_azs   := [binary() | undefined],
+       missing_azs  := [binary()]}].
+queues_not_fully_az_covered(undefined) ->
+    [];
+queues_not_fully_az_covered(TagKey) ->
+    Nodes = rabbit_nodes:list_members(),
+    NodeToAZ = ?MODULE:node_tags_for_nodes(Nodes, TagKey),
+    AllAZs = lists:usort([AZ || _ := AZ <- NodeToAZ, AZ =/= undefined]),
+    [#{queue        => amqqueue:get_name(Q),
```
kjnilsson: whitespace

deps/rabbit/src/rabbit_member_placement_az.erl
```diff
@@ -0,0 +1,190 @@
...
+       member_nodes => MemberNodes,
+       member_azs   => MemberAZs,
+       missing_azs  => MissingAZs}
+     || Q          <- rabbit_amqqueue:list(),
```
kjnilsson: whitespace

deps/rabbit/src/rabbit_queue_location.erl
```diff
@@ -48,14 +52,19 @@ select_leader_and_followers(Q, Size)
   when (?is_amqqueue_v2(Q)) andalso is_integer(Size) ->
     LeaderLocator = leader_locator(Q),
     QueueType = amqqueue:get_type(Q),
-    do_select_leader_and_followers(Size, QueueType, LeaderLocator).
+    PlacementTagKey = case QueueType of
+        rabbit_quorum_queue -> placement_tag_key(Q);
+        _                   -> undefined
```
kjnilsson: whitepspace

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -1434,6 +1437,18 @@ register_rabbit_user_permissions_projection() ->
     KeyPos = #user_permission.user_vhost,
     register_simple_projection(Name, PathPattern, KeyPos, false).
 
+register_rabbit_node_metadata_projection() ->
+    Name = rabbit_khepri_node_metadata,
+    PathPattern = rabbit_db_node_metadata:khepri_node_metadata_path(
+                    _WildcardNode = ?KHEPRI_WILDCARD_STAR),
```
kjnilsson: why the assignment here?

kjnilsson (commented):
mostly formatting

deps/rabbit/src/rabbit_amqqueue.erl
mkuratczyk: it's `andalso` but point taken - I replaced with an explicit one

deps/rabbit/src/rabbit_khepri.erl
mkuratczyk: I'm mostly following the pattern here. In this case we pass `?KHEPRI_WILDCARD_STAR` just once, but there's for example:
```
    PathPattern = rabbit_db_binding:khepri_route_path(
                    _VHost = ?KHEPRI_WILDCARD_STAR,
                    _ExchangeName = ?KHEPRI_WILDCARD_STAR,
                    _Kind = ?KHEPRI_WILDCARD_STAR,
                    _DstName = ?KHEPRI_WILDCARD_STAR,
                    _RoutingKey = ?KHEPRI_WILDCARD_STAR),
```
and in this case it's much more useful.

deps/rabbit/src/rabbit_db_cluster.erl
```diff
@@ -249,6 +249,7 @@ forget_member_using_khepri(_Node, true) ->
        #{domain => ?RMQLOG_DOMAIN_DB}),
     {error, not_supported};
 forget_member_using_khepri(Node, false = _RemoveWhenOffline) ->
+    _ = rabbit_db_node_metadata:delete(Node),
```
dumbbell, line 252: (Not a review feedback)

Hmm, perhaps it could be interesting to have triggers and `keep_while` conditions based on cluster membership.

deps/rabbit/src/rabbit_db_node_metadata.erl
```diff
@@ -0,0 +1,102 @@
...
+         khepri_node_metadata_path/1
+        ]).
+
+-define(KHEPRI_PROJECTION, rabbit_khepri_node_metadata).
+
+%% -------------------------------------------------------------------
+%% set().
+%% -------------------------------------------------------------------
+
+-spec set(node(), map()) -> boolean().
```
dumbbell: Returning a boolean to report success or failure is not common in Erlang. Could you please either keep the return value of Khepri, or simplify it but still keeping a `ok | {error, _}` format?

deps/rabbit/src/rabbit_db_node_metadata.erl
```diff
@@ -0,0 +1,102 @@
...
+            Metadata;
+        _ ->
+            #{}
+    end.
+
+%% -------------------------------------------------------------------
+%% delete().
+%% -------------------------------------------------------------------
+
+-spec delete(node()) -> boolean().
```
dumbbell: Likewise, a boolean is not a usual Erlang pattern.

deps/rabbit/src/rabbit_khepri.erl
dumbbell: Yes, I don’t know who started to document arguments like this, but I find it useful too.

In this case, I would name the argument `_Nodename`. It doesn’t need to say "Wildcard".

deps/rabbit/src/rabbit_khepri.erl
```diff
@@ -1434,6 +1437,18 @@ register_rabbit_user_permissions_projection() ->
...
+    Fun = fun([rabbitmq, node_metadata, _Node], #node_metadata{} = Record) ->
+                  Record
+          end,
```
dumbbell: If you don’t do any computation and don’t use the tree node path, please use the `copy` projection "fun". It will be faster and avoids the need to extract and store a function.

In fact, look at other projections calling `register_simple_projection/4`. I think it corresponds to your use case.

deps/rabbit/src/rabbit_db_node_metadata.erl
```diff
@@ -0,0 +1,102 @@
...
+%% @doc Returns the node metadata map for the given node using a consistent query.
+get_consistent(Node) ->
+    Path = khepri_node_metadata_path(Node),
+    Options = #{favor => consistency},
+    case rabbit_khepri:get(Path, Options) of
+        {ok, #node_metadata{metadata = Metadata}} ->
+            Metadata;
+        _ ->
+            #{}
+    end.
```
dumbbell: This function seems unused.

deps/rabbit/src/rabbit_db.erl
```diff
@@ -74,6 +74,9 @@ init() ->
                "DB: initialization successful",
                #{domain => ?RMQLOG_DOMAIN_DB}),
 
+            LocalTags = application:get_env(rabbit, node_tags, []),
+            rabbit_db_node_metadata:set(node(), #{node_tags => LocalTags}),
```
dumbbell: You should ignore the return value explicitly with `_ = ...`.

deps/rabbit/include/rabbit_khepri.hrl
```diff
@@ -24,6 +24,9 @@
 -define(RABBITMQ_KHEPRI_MAINTENANCE_PATH(Node),
         ?RABBITMQ_KHEPRI_ROOT_PATH([node_maintenance, Node])).
 
+-define(RABBITMQ_KHEPRI_NODE_METADATA_PATH(Node),
+        ?RABBITMQ_KHEPRI_ROOT_PATH([node_metadata, Node])).
```
dumbbell: The choice of `node_maintenance` was poor. Perhaps we should start an actual branch for everything related to nodes.

What do you think of the following branch?

```
ROOT
`-- 'node'
    `-- Nodename
         `-- 'metadata'
```

There would be two macros:

```erlang
-define(RABBITMQ_KHEPRI_NODE_PATH(Node),
        ?RABBITMQ_KHEPRI_NODE_PATH(Node, [])).
-define(RABBITMQ_KHEPRI_NODE_PATH(Node, Rest),
        ?RABBITMQ_KHEPRI_ROOT_PATH([nodes, Node | Rest])).

-define(RABBITMQ_KHEPRI_NODE_METADATA_PATH(Node),
        ?RABBITMQ_KHEPRI_NODE_PATH(Node, [node_metadata])).
```

In the future, we could move the maintenance underneath.

I also hesitate between the terms "node" and "member". Not sure.

dumbbell (changes requested):

deps/rabbit/include/rabbit_khepri.hrl
```diff
@@ -62,3 +70,8 @@
 
 -define(RABBITMQ_KHEPRI_QUEUE_PATH(VHost, Name),
         ?RABBITMQ_KHEPRI_VHOST_PATH(VHost, [queues, Name])).
+
+-record(node_metadata, {
+          node,
+          metadata = #{}
+        }).
```
dumbbell: I don’t think this record belongs to `rabbit_khepri.hrl` because it "sits" above the datastore semantically. If there is no appropriate headers, you can create `deps/rabbit/include/node_metadata.hrl`.

dumbbell: Also about this record:
* Could you please add type specs?
* Could you please define a type for the metadata map? You could also add the `node_tags` key in the map spec.

dumbbell: In fact, this record is only used in `rabbit_db_node_metadata` if I understand correctly. In this case, please define it in the module directly.

dumbbell (changes requested):

dumbbell (approved):

PR #16757, opened 2026-06-22
CLI: Add leadership transfer commands to `rabbitmq-queues` and `rabbitmq-streams`

the-mikedavis:
This is a minor QoL improvement / fix to add CLI commands to `rabbitmq-queues` and `rabbitmq-streams`, `transfer_leadership <queue> <node> [--vhost=<vhost>]`. Currently this can only be done with an eval command. It's not useful often, but it's good for tricky scenarios like handling https://github.com/rabbitmq/rabbitmq-server/discussions/14137. I think it might be helpful in the future for not-yet-seen failure/recovery scenarios.

### DIFF
--- a/deps/rabbit/docs/rabbitmq-queues.8
+++ b/deps/rabbit/docs/rabbitmq-queues.8
@@ -163,6 +163,14 @@ Example:
 .Sp
 .Dl rabbitmq-queues delete_member --vhost Qo a-vhost Qc Qo a-queue Qc Qo rabbit@decomissioned-node Qc
 .\" ------------------------------------
+.It Cm transfer_leadership Ar queue Ar node Fl -vhost Ar virtual-host
+.Pp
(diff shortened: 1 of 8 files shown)


michaelklishin:
For technical operations, having such commands is overdue.

The primary non-`eval` alternative is putting the node hosting the current leader into maintenance mode with `rabbitmq-upgrade drain` and then reviving (`rabbitmq-upgrade revive`). Which is a complete overkill for one or a handful of queues.

the-mikedavis:
I suspect it might also be nice for some manual 'heat management'. Like if you had two very active streams on the same node of a 3-node cluster. If the writers are colocated then their network egress for replication adds up and might approach limits.

michaelklishin:
@the-mikedavis I have pushed a couple of small tweaks that improve error handling.

michaelklishin (approved):

PR #16759, opened 2026-06-22
Fix SigV4 SignatureDoesNotMatch on body-less requests (e.g. S3 GET)

sdewhitt:
## Proposed Changes

This fixes a SigV4 signing bug that causes **every empty-body AWS request to fail with `SignatureDoesNotMatch` (HTTP 403)**. The most visible symptom is that fetching an object via S3 (`GetObject`) fails 100% of the time.

Root cause:

1. The SigV4 signer includes `content-length` in the signed-header set for every request. For an empty body its value is `"0"`, and it becomes part of the canonical/signed headers the signature is computed over.
2. The body-less request clause in `perform_request_with_creds/7` calls `httpc:request/4` with an empty options list. For a request with no body, Erlang/OTP `httpc` **strips** the `content-length` header before transmitting.

Because `content-length` is signed but not sent, the headers the service receives no longer match the signed-header set, and AWS rejects the request with `SignatureDoesNotMatch`.

Body-bearing requests (e.g. Secrets Manager `GetSecretValue`, any POST/PUT) are unaffected, because httpc preserves `content-length` when a body is present -- which is why this bug is invisible for most of the API surface and only bites empty-body GETs like S3.

The fix passes `{headers_as_is, true}` to httpc on the body-less clause so the signed headers -- including `content-length: 0` -- are transmitted verbatim and match the signature. The body-bearing clause is unchanged.

```erlang
-    Response = httpc:request(Method, {URI, Headers}, Options1, []),
+    Response = httpc:request(Method, {URI, Headers}, Options1, [{headers_as_is, true}]),
```

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #NNNN, no existing issue)
- [ ] Build system and/or CI

## Checklist

## Further Comments

This is a one-line change, but here is the reasoning and the alternatives considered:

- **Why `headers_as_is` rather than not signing `content-length`?** An alternative fix is to omit `content-length` from the signed-header set for empty bodies in `rabbitmq_aws_sign`, so the signed set matches what httpc actually sends. That also works, but it is a larger change in the signer and changes signing behavior; `{headers_as_is, true}` is the minimal, localized fix that makes the transport honor what was already signed.

- **Verification.** Tested live against two services using the same credentials: with the change, an S3 `GetObject` succeeds; without it, the same S3 GET fails with `SignatureDoesNotMatch` while Secrets Manager `GetSecretValue` succeeds -- confirming the failure is specific to the empty-body path and not a credential/region issue.

### DIFF
--- a/deps/rabbitmq_aws/src/rabbitmq_aws.erl
+++ b/deps/rabbitmq_aws/src/rabbitmq_aws.erl
@@ -514,7 +514,7 @@ perform_request_with_creds(State, Service, Method, Headers, Path, Body, Options,
 %% @end
 perform_request_with_creds(State, Method, URI, Headers, undefined, "", Options0) ->
     Options1 = ensure_timeout(Options0),
-    Response = httpc:request(Method, {URI, Headers}, Options1, []),
+    Response = httpc:request(Method, {URI, Headers}, Options1, [{headers_as_is, true}]),
     {format_response(Response), State};
 perform_request_with_creds(State, Method, URI, Headers, ContentType, Body, Options0) ->
     Options1 = ensure_timeout(Options0),
--- a/deps/rabbitmq_aws/test/rabbitmq_aws_tests.erl
+++ b/deps/rabbitmq_aws/test/rabbitmq_aws_tests.erl
@@ -195,7 +195,7 @@ gen_server_call_test_() ->
                             {"https://ec2.us-east-1.amazonaws.com/?Action=DescribeTags&Version=2015-10-01",
                                 _Headers},
                             _Options,
-                            []
+                            [{headers_as_is, true}]
                         ) ->
                             {ok, {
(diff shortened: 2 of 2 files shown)


lukebakken (approved):

michaelklishin (approved):

the-mikedavis (commented):
(Non-blocking follow-up. Just me thinking out loud really 😄.) I see now, httpc does not set content-length for get/head/options/trace/delete methods. I think this is a good fix but overall I wonder if we should just not be signing content-length with SigV4 for requests which don't have bodies (get/head/...). SigV4 is meant to protect against a scenario where an adversary MITM's you, and the signature header can only be reused to make basically the exact same request. The content hash of the body should protect stronger than content-length anyways. In the ad-hoc AWS API implementation in rabbitmq-stream-s3 we don't sign content-length at all.

PR #16775, opened 2026-06-23
Use CSPRNG for generating password salts

ansd:
What:
Use a Cryptographically Secure Pseudo-Random Number Generator (CSPRNG) for generating internal user password salts instead of a standard PRNG.

Why:
The `rand` module in Erlang is a PRNG and is not cryptographically secure. While a salt primarily needs to be globally unique rather than strictly unpredictable, modern security standards (such as NIST) recommend using a CSPRNG for all cryptographic material, including salts.


michaelklishin (approved):
Good find.

A mandatory mention of backwards compatibility: this new function produces a value of the same length/width (4 bytes) and the salt value is embedded into the user record in the internal database, so this change is entirely backwards compatible.

PR #16791, opened 2026-06-24
rabbitmq-plugins list: omit plugin table when node is unreachable

shashank-singh-2999:
When `rabbitmq-plugins list` cannot contact the target node, the plugin table is of limited value: running status (*) is unavailable, and E/e markers may reflect the local fallback path rather than the target node's actual enabled plugins file.

This change removes the plugin table from the output in the node-down case and keeps only the legend line, which already reports "[failed to contact <node> - status not shown]".

The generic `format_output` clause matching `%{status: status, ...}` is replaced by two explicit clauses — one for `:node_down` and one for `:running` — since those are the only two values produced by `list_command.ex`.


michaelklishin:
Thank you, @shashank-singh-2999.

michaelklishin (approved):

PR #16803, opened 2026-06-25
Update Erlang.mk

lhoguin:


michaelklishin:
@lhoguin thank you. I did find two more `ebin` references and added an `-f` to one `rm`.

`gmake run-broker`, `gmake start-cluster`, running tests are all functional, so I do not have any reasons to not merge this but feel free to tweak the Make changes as I am fairly Make illiterate.

michaelklishin (approved):

lhoguin:
@michaelklishin Changes look good, sorry I forgot to check if we were doing anything special in rabbit. What about the added `-f` to `rm` though, did you run into some issue? Asking because this will be overwritten by the next Erlang.mk update otherwise.

michaelklishin:
@lhoguin no, just a general assumption that we don't want to depend on the permissions of a marker file. We can revert it if you are sure that's an exceptionally rare scenario.

Issue #12815, opened 2024-11-26
Occasional `badarith` errors when collecting metrics


mkuratczyk:
### Describe the bug

I don't have the exact repro steps. I think this is a race condition when a queue is deleted while the metrics are collected and the queue is found by the collector but then returns empty values (`''`) instead of the expected numbers. I've seen this occur in 3 different places:

```
crasher:
  initial call: rabbit_mgmt_metrics_collector:init/1
  pid: <0.699.0>
  registered_name: connection_coarse_metrics_metrics_collector
  exception error: an error occurred when evaluating an arithmetic expression
    in operator  -/2
       called as '' - 19495760921
    in call from rabbit_mgmt_metrics_collector:difference/2 (rabbit_mgmt_metrics_collector.erl, line 649)
    in call from rabbit_mgmt_metrics_collector:aggregate_entry/4 (rabbit_mgmt_metrics_collector.erl, line 234)
    in call from lists:foldl/3 (lists.erl, line 1594)
    in call from ets:do_foldl/4 (ets.erl, line 656)
    in call from ets:foldl/3 (ets.erl, line 645)
    in call from rabbit_mgmt_metrics_collector:aggregate_metrics/2 (rabbit_mgmt_metrics_collector.erl, line 172)
    in call from rabbit_mgmt_metrics_collector:handle_info/2 (rabbit_mgmt_metrics_collector.erl, line 106)
```

```
crasher:
  initial call: rabbit_mgmt_metrics_collector:init/1
  pid: <0.7180.0>
  registered_name: connection_coarse_metrics_metrics_collector
  exception error: an error occurred when evaluating an arithmetic expression
    in operator  +/2
       called as 19340950184 + ''
    in call from rabbit_mgmt_metrics_collector:sum_entry/2 (rabbit_mgmt_metrics_collector.erl, line 631)
    in call from rabbit_mgmt_metrics_collector:'-insert_entry_op/4-fun-0-'/2 (rabbit_mgmt_metrics_collector.erl, line 598)
    in call from maps:update_with/4 (maps.erl, line 296)
    in call from rabbit_mgmt_metrics_collector:insert_entry_op/4 (rabbit_mgmt_metrics_collector.erl, line 597)
    in call from lists:foldl/3 (lists.erl, line 1594)
    in call from rabbit_mgmt_metrics_collector:aggregate_entry/4 (rabbit_mgmt_metrics_collector.erl, line 235)
[...]

PR #16832, opened 2026-06-29
Management agent: tolerate non-integer values in metrics aggregation

MugemaneBertin2001:
## Root cause

`rabbit_mgmt_metrics_collector` reads rows from the `core_metrics` ETS tables on every scrape and feeds them through `sum_entry/2` and `difference/2`, which add or subtract tuples element-wise. Each field is expected to be an integer.

Under connection or queue churn, a field can briefly read as the empty atom `''` — the sentinel `rabbit_mgmt_format` uses to represent "null". Performing arithmetic on it raises `badarith` inside the gen_server and kills the collector for that table, losing any in-progress aggregation state. This is the exact symptom captured in #12815:

```
exception error: an error occurred when evaluating an arithmetic expression
    in operator  -/2
       called as '' - 19495760921
    in call from rabbit_mgmt_metrics_collector:difference/2
    in call from rabbit_mgmt_metrics_collector:aggregate_entry/4
    in call from lists:foldl/3
    in call from ets:foldl/3
    in call from rabbit_mgmt_metrics_collector:aggregate_metrics/2
```

## Change

Wrap `sum_entry/2` and `difference/2` so each input tuple passes through `normalise/1` first:

- Fast path: if every element is an integer the tuple is reused as-is — one `tuple_size` plus up to 8 `element/2` checks, no allocation
- Slow path: a non-integer field is coerced to `0` and a rebuilt tuple is fed to the original arithmetic

A single bad scrape now produces an imperfect delta instead of a crash, and the next scrape (every `collect_statistics_interval`, default 5s) recovers a correct value.

## Test plan

- `deps/rabbitmq_management_agent/test/metrics_SUITE.erl`:
  - `collector_survives_non_integer_metric` — end-to-end: injects a synthetic Pid into `connection_coarse_metrics`, runs one scrape via `force_collect` to seed `old_aggr_stats`, mutates the row so one field is `''`, runs a second scrape, asserts the same collector Pid is still alive. Without the patch this assertion exits with `badarith`
  - `sum_entry_and_difference_unit_checks` — pins down the contract: integer math is unchanged; `''` on either operand coerces to `0`
- `erlc -W` is clean on the changed source and SUITE

## Reviewer FAQ

- **Why not skip the entry entirely instead of coercing to 0?** Skipping would silently drop the contribution of the well-formed siblings in the same tuple (e.g. `send_oct` would also be lost when only `recv_oct` is the sentinel). Coercing the bad field to `0` preserves the rest of the row and the next interval recovers a correct value
- **Why expose `sum_entry/2` and `difference/2`?** Only so the unit test can pin the contract. They are pure helpers with no state and no side effects
- **What does the dialyzer change do?** The existing `-dialyzer({no_match, …})` suppression moved to the renamed `_safe` clauses where the exhaustive-pattern warning still applies. The outer wrappers are single-clause + guard and don't need a suppression
- **Performance impact?** The hot path adds `tuple_size/1` + up to 8 `element/2` calls per arithmetic operation. Negligible compared to the surrounding ETS work; runs once per row per `collect_statistics_interval`

### DIFF
--- a/deps/rabbitmq_management_agent/src/rabbit_mgmt_metrics_collector.erl
+++ b/deps/rabbitmq_management_agent/src/rabbit_mgmt_metrics_collector.erl
@@ -22,6 +22,7 @@
          code_change/3]).
 -export([index_table/2]).
 -export([reset_all/0]).
+-export([sum_entry/2, difference/2]).
 
 -import(rabbit_mgmt_data, [lookup_element/3]).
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
@MugemaneBertin2001 thank you for contributing. May I ask you to [request a CLA and sign it](https://github.com/rabbitmq/cla) if you haven't already?

It won't take long, you need to email us and we will respond with a digital signature request.

Thank you.

deps/rabbitmq_management_agent/src/rabbit_mgmt_metrics_collector.erl
```diff
@@ -625,42 +626,72 @@ get_difference(Id, Stats, #state{old_aggr_stats = OldStats}) ->
...
+%% non-integer fields to 0 so the aggregation proceeds; the next
+%% scrape recovers a correct value.
+sum_entry(A, B) when is_tuple(A), is_tuple(B), tuple_size(A) =:= tuple_size(B) ->
+    sum_entry_safe(normalise(A), normalise(B)).
+
+-dialyzer({no_match, sum_entry_safe/2}).
+sum_entry_safe({A0}, {B0}) ->
     {B0 + A0};
-sum_entry({A0, A1}, {B0, B1}) ->
+sum_entry_safe({A0, A1}, {B0, B1}) ->
```
michaelklishin, line 642: Are you sure that this is node-local code so that you can just rename this function? Or can other nodes potentially use it, including in mixed version clusters?

MugemaneBertin2001:
Hey @michaelklishin — the CLA is already sorted. I emailed the request in on June 29 and Aitor Alberto Perez Cedres from Broadcom confirmed receipt the same day. Happy to forward that confirmation over if it helps you tick it off.

deps/rabbitmq_management_agent/src/rabbit_mgmt_metrics_collector.erl
MugemaneBertin2001: Good catch to check, but `sum_entry/2` and `difference/2` aren't actually renamed — same names, same arities, same tuple shapes going in and out. What I renamed is the internal clause set (now `sum_entry_safe`/`difference_safe`, both un-exported); the public functions are now just thin wrappers that normalise args first.

Also grepped every caller of `rabbit_mgmt_metrics_collector:*` — nothing outside this module ever calls `sum_entry` or `difference`. The two functions were module-local before this PR (not even exported), so no cross-node RPC path exists. I only added the `-export` so `metrics_SUITE` can drive the regression tests directly.

So on integer inputs — i.e. everything a real cluster produces — behaviour is identical to before. Mixed-version clusters aren't affected.

michaelklishin:
Thank you. This is treating the symptoms but leaving the root cause as it is.

Moved to https://github.com/rabbitmq/rabbitmq-server/pull/16856.

PR #16833, opened 2026-06-29
CQ: fix slash accumulation (enametoolong)

mkuratczyk:
Previously, each purge appended another '/' to the index dir path. With enough purges (thousands),
the path was so long, that the queue failed with enametoolong.

Fixes https://github.com/rabbitmq/rabbitmq-server/discussions/15338


michaelklishin (approved):
Great find, @mkuratczyk!

PR #16842, opened 2026-06-30
rabbitmq-plugins: fix remote node file path handling, offline mode remote node validation and  is-enabled false positives

shashank-singh-2999:
## Motivation

Several bugs in `rabbitmq-plugins` commands were uncovered when testing against remote nodes and offline mode:

 * `enable`, `disable`, and `set` wrote the `enabled_plugins` file using the **local** node's file path and then sent that path to the remote node over RPC. When the remote node stores its `enabled_plugins` file in a different location — the RPC call failed.

 * When an RPC call to a remote node failed mid-operation, the commands silently fell back to writing the local node's enabled_plugins file. This left the system in an inconsistent state: the local node would pick up the new plugin list on its next restart, but the intended remote node would not. The operator had no indication that the change had been applied to the wrong node. All reads and writes for remote node targets now go exclusively through RPC; there is no silent local fallback.

 * `rabbitmq-plugins is_enabled --offline` incorrectly reported every plugin present in the plugins directory as enabled, even if it was not listed in the `enabled_plugins` file. It was calling an internal helper that enumerated all plugins on disk rather than reading the file.

 * `--offline` mode accepted a remote node target (`-n rabbit@remotehost`) without error. Because offline mode reads and writes only local files, using it against a remote node silently operated on the wrong node's environment and discarded the `-n` flag entirely.

### `plugins_helpers.ex`

 * Remote node file path handling is corrected. `set_enabled_plugins/2` now calls `write_enabled_plugins_on_remote/3` for remote targets, which fetches the remote node's own `enabled_plugins` path via RPC before writing, rather than assuming the local path is valid on the remote.

 * `is_enabled --offline` now calls `read_enabled/1` to obtain the list of explicitly enabled plugins and then expands transitive dependencies via `rabbit_plugins:dependencies/3`, matching the semantics of online mode. The old `list_names/1` helper (which returned all plugins in the directory) has been removed.

 * `validate_node_and_mode/2` is introduced and enforces: `--offline` mode is rejected when the target node is remote. Best-effort mode requires the node to be running when the target is remote; for the local node it falls back to offline gracefully.

### `list_command.ex`

 * When the node cannot be reached, the command now distinguishes between a stopped local node and an unreachable remote node. For a stopped local node, it returns an `:offline` status with full plugin metadata read from disk; plugins show their enabled (`E`) and implicit (`e`) markers but not the running (`*`) marker. For an unreachable remote node, it returns a `:node_down` status with an empty plugin list, since all metadata would require RPC calls to retrieve.

### `plugins.ex` (formatter)

 * Introduces the `:offline` status for `rabbitmq-plugins list` output, used when the local node is not running. The existing `:node_down` status is now reserved exclusively for unreachable remote nodes. Because all RPCs fail when a remote node is down, the `:node_down` path produces an empty plugin list; the `:offline` path retains full plugin metadata read from disk.

### `enable_command.ex`, `disable_command.ex`, `set_command.ex`

 * `validate_execution_environment/2` in each command delegates to the new `validate_node_and_mode/2`, giving consistent validation across all three mutating plugin commands.

### DIFF
--- a/deps/rabbitmq_cli/lib/rabbitmq/cli/formatters/plugins.ex
+++ b/deps/rabbitmq_cli/lib/rabbitmq/cli/formatters/plugins.ex
@@ -9,17 +9,10 @@ defmodule RabbitMQ.CLI.Formatters.Plugins do
   @behaviour RabbitMQ.CLI.FormatterBehaviour
 
   def format_output(
-        %{status: :node_down, format: format},
+        %{status: status, format: format, plugins: plugins},
         options
       ) do
-    legend(:node_down, format, options)
-  end
-
-  def format_output(
-        %{status: :running, format: format, plugins: plugins},
-        options
-      ) do
-    legend(:running, format, options) ++ format_plugins(plugins, format)
+    legend(status, format, options) ++ format_plugins(plugins, format)
   end
 
   def format_output(%{enabled: enabled, mode: _} = output, options) do
--- a/deps/rabbitmq_cli/lib/rabbitmq/cli/formatters/plugins.ex
+++ b/deps/rabbitmq_cli/lib/rabbitmq/cli/formatters/plugins.ex
@@ -195,6 +195,10 @@ defmodule RabbitMQ.CLI.Formatters.Plugins do
     "[failed to contact #{node} - status not shown]"
   end
 
+  defp status_message(:offline, node) do
+    "#{node} is stopped; * (running) status is not available"
+  end
+
   defp applying(%{mode: :offline, set: set_plugins}, _) do
     set_plugins_message =
       case length(set_plugins) do
--- a/deps/rabbitmq_cli/lib/rabbitmq/cli/plugins/commands/directories_command.ex
+++ b/deps/rabbitmq_cli/lib/rabbitmq/cli/plugins/commands/directories_command.ex
@@ -44,6 +44,7 @@ defmodule RabbitMQ.CLI.Plugins.Commands.DirectoriesCommand do
     Validators.chain(
       [
         &require_rabbit_and_plugins/2,
+        &PluginHelpers.validate_node_and_mode/2,
         &PluginHelpers.enabled_plugins_file/2,
         &plugins_dir/2
       ],
--- a/deps/rabbitmq_cli/lib/rabbitmq/cli/plugins/commands/disable_command.ex
+++ b/deps/rabbitmq_cli/lib/rabbitmq/cli/plugins/commands/disable_command.ex
@@ -49,7 +49,6 @@ defmodule RabbitMQ.CLI.Plugins.Commands.DisableCommand do
         ]
       else
         [
-          &require_rabbit_and_plugins/2,
           &PluginHelpers.validate_node_and_mode/2
         ]
       end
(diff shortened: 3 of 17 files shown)


mkuratczyk (changes requested):
Overall looks good, thanks. I found one case where the result is not what I would expect:
```
make start-cluster
./sbin/rabbitmq-plugins list
```

for me this returns a list of plugins, all marked as disabled. However, the default node name should be `rabbit` and I have no such node running (there's `rabbit-1`, rabbit-2 and `rabbit-3`). So this should fail to connect and not print a list of plugins

lukebakken (commented):
## Review of #16842

> Reviewed by Claude (AI) under @lukebakken's guidance, including live testing against a 3-node cluster (`rabbit-1/2/3` running, default `rabbit@host` down).

The four documented bugs are genuinely fixed and verified live:

- **Remote path handling** — `enable`/`disable` against running `rabbit-1` now write the *remote* node's own `enabled_plugins` path and apply live. ✓
- **No silent local fallback** — `enable` against a down remote node is rejected; the local `enabled_plugins` file is never touched. ✓
- **`is_enabled --offline`** — now reads the file; a plugin present on disk but absent from the file is correctly reported as not enabled (incl. transitive deps). ✓
- **`--offline` + remote node** — rejected with a clear message. ✓

### 🔴 One blocker: the `:offline` status from the description is missing

The PR body says it introduces an `:offline` status (local node down) distinct from `:node_down` (unreachable remote). The code never emits `:offline` — `list_command.ex` still only produces `:node_down`/`:running`, and the formatter has no `:offline` clause.

This means @mkuratczyk's reported case still reproduces, and the formatter change amplifies it. Against a down default node:

```
$ ./sbin/rabbitmq-plugins list
 | Status: [failed to contact rabbit@... - status not shown]
[  ] rabbitmq_management        ...
   ... full table, all disabled ...
EXIT: 0
```

On `main`, the `:node_down` clause printed only the legend; the new unified clause appends the full plugin table (read from local disk) for every status. So a down default node now prints a complete "all disabled" list and exits 0.

Please either implement the `:offline` status (local-down → `:offline`, formatter handles it) or correct the description and reconsider the down-default-node output.

### Smaller items

- **No test coverage for the remote-RPC paths.** New tests use `:rabbit` (treated as *local*, so it exercises the local write path, not the new remote functions) and `:jake@thedog` (validation only). `write_enabled_plugins_on_remote/3`, `read_enabled_remote/1`, and the down-mid-operation path are untested.
- **`node_locality/1` is load-bearing and untested.** Its host comparison is a heuristic prefix match (`hosts_match?/2`); short/long-name and host edge cases have no unit tests, and these are pure functions.
- **Inconsistency:** `list` drops `require_rabbit_and_plugins` for remote nodes while `enable`/`disable`/`set` keep it. Worth making deliberate.

shashank-singh-2999 (commented):
@mkuratczyk Please re-review updated changes:-

:offline status — Implemented. The list command now distinguishes between a stopped local node (:offline) and an unreachable remote node (:node_down). When the local node is stopped, the full plugin table is still shown (read from the plugins directory on disk) but with a clear status line indicating the node is stopped and running (*) status is unavailable. When a remote node is unreachable, the plugin table is suppressed as before. Added a unit test covering the :offline output.

Test coverage — Added plugins_helpers_test.exs with unit tests.

Inconsistent validation — Dropped require_rabbit_and_plugins from the remote branch of enable/disable/set as well, consistent with the list change. These commands don't need to load local rabbit beams for remote targets since the enabled-plugins write is done via RPC on the remote node.

One caveat — In local development with start-background-broker (no configured --plugins-dir), rabbitmq-plugins list will show the plugins table without E/e enabled markers, since the enabled plugins file path is not set. This is a local testing artefact — in a real deployment the path is always configured.

PR description has been updated to reflect all changes.

mkuratczyk (approved):

PR #16858, opened 2026-07-03
OAuth 2: improve robustness of a token refresh path

michaelklishin:


deps/rabbitmq_auth_backend_oauth2/test/configurable_http_handler.erl
```diff
@@ -0,0 +1,24 @@
+%% This Source Code Form is subject to the terms of the Mozilla Public
+%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%
+%% Copyright (c) 2007-2026 Broadcom. All Rights Reserved. The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries. All rights reserved.
+%%
+-module(configurable_http_handler).
```
michaelklishin, line 7: In case you wonder why not adapt `oauth2_http_mock` or `jwks_http_handler`: the former is a request-matching mock, the latter responds with a fixed status and value.

This module allows the (test) caller control certain response parameters, so this is an, ahem, configurable mock handler.

PR #16860, opened 2026-07-03
Exit start-background-broker retry loop as soon as await_startup succeeds

shashank-singh-2999:
## What?

`start-background-broker`'s retry loop for `rabbitmqctl await_startup` now exits early as soon as an attempt succeeds, rather than always looping through all 10 attempts before a final check.

## Why?

The previous loop only used `||` to trigger a sleep on failure, so a successful early attempt would still fall through the remaining iterations (and the loop was still followed by an unconditional final `await_startup` call). This wastes time on `make` targets that start a background node, particularly in CI.


dumbbell (approved):

dumbbell:
Could you please squash both commits?

PR #16863, opened 2026-07-03
Maintenance mode: drain and revive tolerate failing housekeeping callbacks

### LINKED ISSUE #3369: `rabbitmq-upgrade drain` exits with exit code 69 when it shouldn't

MugemaneBertin2001:
Closes #3369.

## Root cause

Once `rabbit_maintenance:drain/0` has run `mark_as_being_drained/0`, the node is already in draining state at the cluster level — everything after that (`disconnect_federation_links/0`, `close_all_client_connections/0`, `transfer_leadership_of_metadata_store/1`, `rabbit_queue_type:drain/1`, etc.) is housekeeping around cleaning up in-flight state.

Historically any of those steps could raise:

- a plugin whose `drain/1` callback crashes,
- a channel that refuses to shut down within the termination timeout,
- a stuck federation link that returns `noproc`,
- a slow Ra leadership transfer that errors out.

When they did, the raise propagated out of `drain/0`, back through `rabbit_misc:rpc_call/5` as `{badrpc, {'EXIT', _}}`, and the CLI mapped it to a non-zero exit code (69 / `EX_UNAVAILABLE`) — even though the node was, in fact, drained. A subsequent call to `rabbitmq-upgrade drain` on the same node would return `0` immediately, because there was nothing left to do. Kubernetes preStop hooks and other upgrade automation treated the first failure as a real drain failure, retried or aborted the upgrade, and produced `FailedPreStopHook` events.

`revive/0` had the mirror-image problem: a failing housekeeping callback would raise before `unmark_as_being_drained/0` ran, leaving the DRAINING flag stuck and the node effectively unusable until an operator intervened.

## Fix

- Every housekeeping step in both `drain/0` and `revive/0` now runs through a local `safe_maintenance_step/2,3` helper that logs a warning (with class, reason, and stacktrace) and continues.
- The only step intentionally left unwrapped is `mark_as_being_drained/0` — if the maintenance flag can't be written, the drain hasn't actually happened and the caller must know.
- `revive/0` is now guaranteed to reach `unmark_as_being_drained/0` regardless of whether the queue-type or federation callbacks succeeded.
- The `drain/0 :: () -> ok` and `revive/0 :: () -> ok` specs are preserved: on the happy path, behaviour is unchanged; on a housekeeping failure, callers still see `ok` and can rely on the DRAINING flag reflecting the truth.

Same pattern as `rabbit_networking:close_connection/2`, which already wraps its per-connection shutdown in a `try..catch` for the same reason.

## Test

Added `drain_and_revive_tolerate_failing_callbacks/1` to `maintenance_mode_SUITE`. It:

1. Mecks `rabbit_queue_type:drain/1` and `revive/0` to `exit(injected_..._failure)`,
2. Calls `rabbit_maintenance:drain/0` via RPC and asserts it returns `ok` and the node becomes `is_being_drained_local_read = true`,
3. Calls `rabbit_maintenance:revive/0` via RPC and asserts it returns `ok` and the node becomes `is_being_drained_local_read = false`.

Without this change, step 2's RPC raises and step 3 is never reached.

The existing tests (`listener_suspension_status`, `client_connection_closure`, `quorum_queue_leadership_transfer`, `metadata_store_leadership_transfer`) continue to cover the happy path end to end.

### DIFF
--- a/deps/rabbit/src/rabbit_maintenance.erl
+++ b/deps/rabbit/src/rabbit_maintenance.erl
@@ -70,6 +70,16 @@ is_enabled() ->
 -spec drain() -> ok.
 drain() ->
     ?LOG_WARNING("This node is being put into maintenance (drain) mode"),
+    %% Marking the node as being drained is the single load-bearing step:
+    %% it is what makes the rest of the cluster stop routing new work here.
+    %% It must succeed for the drain to be meaningful, so a failure here
+    %% is intentionally propagated. Every step that follows is
+    %% housekeeping around cleaning up in-flight state; a failure in one
+    %% of them (e.g. a stuck federation link, a plugin drain callback
+    %% that crashes, a channel that refuses to shut down within the
+    %% termination timeout) must not abort the drain and must not
+    %% surface as a non-zero CLI exit code, or automation like
+    %% Kubernetes preStop hooks becomes unreliable. See GH #3369.
     mark_as_being_drained(),
     ?LOG_INFO("Marked this node as undergoing maintenance"),
     %% We intentionally do not rely on the internal event (see below) to stop federation links,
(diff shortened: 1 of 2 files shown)


michaelklishin:
@MugemaneBertin2001 the generated comments are awfully verbose, robotic and use the language that almost no human would appreciate, like that constant "load bearing" bs.

I will submit a variation with something not offensive to the human eyes (and brain) but please integrate this feedback into your future prompts.

michaelklishin:
I will also try to add some writing style clarifications to `AGENTS.md` since I know specifically what I mean by that.

michaelklishin:
Moved to https://github.com/rabbitmq/rabbitmq-server/pull/16865. Thank you, @MugemaneBertin2001.

PR #16871, opened 2026-07-04
Transfer stream leadership on node drain

### LINKED ISSUE #16340: Stream Coordinator Performance and Availability Issue during Node Drain

MugemaneBertin2001:
Closes #16340 (partially — this covers the missing per-stream leadership transfer; the coordinator starvation scenario is separate).

`rabbit_quorum_queue:drain/1` enumerates every quorum queue with a local leader and stops the local Ra leader so a peer takes over. `rabbit_stream_queue:drain/1` only transferred the stream coordinator's own leadership and left individual stream leaders in place, so re-election for each stream only kicked in when the drained node was actually stopped.

On a cluster with many streams and a lot of clients, that produced a mass reconnect at shutdown time, which starved the stream coordinator and kept streams reported as `down` for minutes — exactly what the reporter in #16340 observed.

## Change

- New `rabbit_amqqueue:list_local_stream_leaders/0`, mirrors `list_local_leaders/0`.
- `rabbit_stream_queue:drain/1` now walks that list and calls `rabbit_stream_coordinator:restart_stream/2` with a `preferred_leader_node` that is both online (not under maintenance) and already a replica of the stream.
- Regression test in `maintenance_mode_SUITE` (cluster_size_3): declare a stream, pin its leader to node A, drain A, assert `list_local_stream_leaders/0` on A becomes empty.

## Scope

This does not touch the coordinator starvation scenario reported in the second half of #16340 (thousands of reconnects contending for a single-threaded coordinator on shutdown). That's a scheduler / backpressure problem in the stream coordinator and belongs in a separate change.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -40,7 +40,8 @@
 -export([is_replicable/1, is_exclusive/1, is_not_exclusive/1, is_dead_exclusive/1]).
 -export([list_local_quorum_queues/0, list_local_quorum_queue_names/0,
          list_local_stream_queues/0, list_stream_queues_on/1,
-         list_local_leaders/0, list_local_followers/0, get_quorum_nodes/1,
+         list_local_leaders/0, list_local_stream_leaders/0,
+         list_local_followers/0, get_quorum_nodes/1,
          list_local_quorum_queues_with_name_matching/1,
          list_local_quorum_queues_with_name_matching/2]).
 -export([is_local_to_node/2, is_local_to_node_set/2]).
@@ -1329,6 +1330,13 @@ list_local_leaders() ->
     [ Q || Q <- list(),
          amqqueue:is_quorum(Q),
          amqqueue:get_state(Q) =/= crashed, amqqueue:get_leader_node(Q) =:= node()].
+
+-spec list_local_stream_leaders() -> [amqqueue:amqqueue()].
+list_local_stream_leaders() ->
+    ThisNode = node(),
+    [Q || Q <- list_local_stream_queues(),
+          is_pid(amqqueue:get_pid(Q)),
+          node(amqqueue:get_pid(Q)) =:= ThisNode].
 
 -spec list_local_followers() -> [amqqueue:amqqueue()].
 list_local_followers() ->
--- a/deps/rabbit/src/rabbit_stream_queue.erl
+++ b/deps/rabbit/src/rabbit_stream_queue.erl
@@ -1630,8 +1630,11 @@ stop(_VHost) ->
 
 drain(TransferCandidates) ->
     case whereis(rabbit_stream_coordinator) of
-        undefined -> ok;
-        _Pid -> transfer_leadership_of_stream_coordinator(TransferCandidates)
+        undefined ->
+            ok;
+        _Pid ->
+            transfer_leadership_of_stream_coordinator(TransferCandidates),
+            transfer_leadership_of_local_stream_leaders(TransferCandidates)
     end.
 
 revive() ->
@@ -1655,6 +1658,37 @@ transfer_leadership_of_stream_coordinator(TransferCandidates) ->
         Error ->
             ?LOG_WARNING("Skipping leadership transfer of stream coordinator: ~p", [Error])
     end.
+
+-spec transfer_leadership_of_local_stream_leaders([node()]) -> ok.
+transfer_leadership_of_local_stream_leaders([]) ->
+    ?LOG_WARNING("Skipping stream leadership transfer: no candidate "
+                 "(online, not under maintenance) nodes to transfer to");
+transfer_leadership_of_local_stream_leaders(TransferCandidates) ->
+    LocalLeaders = rabbit_amqqueue:list_local_stream_leaders(),
+    ?LOG_INFO("Will transfer leadership of ~b streams with current leader on this node",
+              [length(LocalLeaders)]),
+    lists:foreach(
+      fun (Q) -> transfer_leadership_of_stream(Q, TransferCandidates) end,
+      LocalLeaders),
+    ?LOG_INFO("Leadership transfer for streams hosted on this node has been initiated").
+
+-spec transfer_leadership_of_stream(amqqueue:amqqueue(), [node()]) -> ok.
+transfer_leadership_of_stream(Q, TransferCandidates) ->
+    %% A stream can only elect a new leader on a node that already hosts a replica.
+    Replicas = get_nodes(Q),
+    Options = case [N || N <- TransferCandidates, lists:member(N, Replicas)] of
+                  [] -> #{};
+                  [Preferred | _] -> #{preferred_leader_node => Preferred}
+              end,
+    QNameStr = rabbit_misc:rs(amqqueue:get_name(Q)),
+    case rabbit_stream_coordinator:restart_stream(Q, Options) of
+        {ok, NewLeader} ->
+            ?LOG_DEBUG("Stream ~ts new leader is on node ~tp", [QNameStr, NewLeader]);
+        Error ->
+            ?LOG_WARNING("Failed to transfer leadership of stream ~ts: ~tp",
+                         [QNameStr, Error])
+    end,
+    ok.
 
 queue_vm_stats_sups() ->
     {[stream_queue_procs,
--- a/deps/rabbit/src/rabbit_stream_queue.erl
+++ b/deps/rabbit/src/rabbit_stream_queue.erl
@@ -1659,6 +1659,18 @@ transfer_leadership_of_stream_coordinator(TransferCandidates) ->
             ?LOG_WARNING("Skipping leadership transfer of stream coordinator: ~p", [Error])
     end.
 
+%% Under a busy coordinator (e.g. many active publishers), each
+%% restart_stream/2 command can take up to the coordinator command timeout
+%% per coordinator member before failing. To avoid dragging drain out for
+%% minutes, cap the total wall-clock spent on per-stream transfers and let
+%% shutdown-time re-election handle any remainder.
+-define(STREAM_DRAIN_BUDGET_MS, 30000).
+%% Small pause between per-stream restart_stream commands to reduce
+%% coordinator command queue pressure when there are many local leaders.
+-define(STREAM_DRAIN_INTER_STREAM_MS, 25).
+-define(STREAM_DRAIN_RESTART_ATTEMPTS, 2).
+-define(STREAM_DRAIN_RESTART_BACKOFF_MS, 500).
+
 -spec transfer_leadership_of_local_stream_leaders([node()]) -> ok.
 transfer_leadership_of_local_stream_leaders([]) ->
     ?LOG_WARNING("Skipping stream leadership transfer: no candidate "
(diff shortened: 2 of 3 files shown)


deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1656,6 +1659,37 @@ transfer_leadership_of_stream_coordinator(TransferCandidates) ->
...
+              end,
+    QNameStr = rabbit_misc:rs(amqqueue:get_name(Q)),
+    case rabbit_stream_coordinator:restart_stream(Q, Options) of
+        {ok, NewLeader} ->
+            ?LOG_DEBUG("Stream ~ts new leader is on node ~tp", [QNameStr, NewLeader]);
+        Error ->
+            ?LOG_WARNING("Failed to transfer leadership of stream ~ts: ~tp",
+                         [QNameStr, Error])
+    end,
+    ok.
```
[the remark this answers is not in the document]
kjnilsson: @MugemaneBertin2001 I think Copilot has a point here. If ok with you we can take. Copilot's suggestion or you could make an update yourself.

kjnilsson (changes requested):
I think we need to handle the case where there are no candidates, otherwise this change looks fine. Cheers.

deps/rabbit/src/rabbit_stream_queue.erl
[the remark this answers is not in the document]
MugemaneBertin2001: Good catch — pushed 29583b2c99. `transfer_leadership_of_stream/2` now skips `restart_stream/2` when the eligible-replica list is empty and logs at info level instead. Shutdown-time re-election handles the fallback.

MugemaneBertin2001:
@kjnilsson addressed the feedback in 29583b2c99 — `transfer_leadership_of_stream/2` now skips the `restart_stream/2` call when no online, non-maintenance replica is available among the transfer candidates, and logs at info level instead. Ready for another look when you have a moment.

MugemaneBertin2001:
`bcdb0f8fba` did not fix it — the sync check on `rabbit_stream_coordinator:members/1` only proves the replicas are running as members, not that their stopped-tail matches the writer's. Both `parallel-ct-set-5` runs failed on that commit with the same log line: `Stream ... new leader is on node rmq-ct-stream_leadership_transfer-1@localhost`, epoch 3.

Root cause is in `rabbit_stream_coordinator:select_leader/2` (machine version 4+): `preferred_leader_node` is only honoured when the preferred node is tied with the current writer on the stopped tail. Right after `restart_stream` advances an epoch on the local node, the writer's tail is `{Epoch, 0}` while replicas still report `empty`, so the writer ends up alone in `Potential` and re-elects itself regardless of the hint.

Fixed in `beb58de9cd` at the drain site: retry `restart_stream/2` up to three times with a 500 ms backoff when the returned leader is the local node. Between attempts the members run under the new epoch long enough for the replicas to catch up, so the next call finds their tails tied with the writer and the `preferred_leader_node` hint wins. The test change from `bcdb0f8fba` is reverted.

mkuratczyk:
I'm testing this PR on Kubernetes. I have a 3-node cluster with `stream-perf-test  -sc 100 -x 100 -y 0 -r 1000` running.
When I perform `kubectl rollout restart sts rmq-server`, and `rmq-server-2` nodes is being drained, rmq-server-0 and rmq-server-1 log errors like this:

```
** Generic server <0.15003.0> terminating
** Last message in was {continue,
                        #{name => "__stream-036_1783429184925345148",
                          nodes =>
                           ['rabbit@rmq-server-2.rmq-nodes.foo',
                            'rabbit@rmq-server-0.rmq-nodes.foo',
                            'rabbit@rmq-server-1.rmq-nodes.foo'],
                          reference =>
                           {resource,<<"/">>,queue,<<"stream-036">>},
                          features => #{committed_offset_calculate => true},
                          leader_pid => <31794.14060.0>,
                          leader_node => 'rabbit@rmq-server-2.rmq-nodes.foo',
                          replica_nodes =>
                           ['rabbit@rmq-server-0.rmq-nodes.foo',
                            'rabbit@rmq-server-1.rmq-nodes.foo'],
                          epoch => 3,
                          event_formatter =>
                           {rabbit_stream_queue,format_osiris_event,
                            [{resource,<<"/">>,queue,<<"stream-036">>}]},
                          max_segment_size_bytes => 500000000,
                          retention => [{max_bytes,20000000000}],
                          filter_size => 16}}
** When Server state == undefined
** Reason for termination ==
** {badrpc,{'EXIT',killed}}

  crasher:
    initial call: osiris_replica:init/1
    pid: <0.15003.0>
    registered_name: []
    exception exit: {badrpc,{'EXIT',killed}}
      in function  gen_server:handle_common_reply/5 (gen_server.erl:2562)
    ancestors: [osiris_server_sup,osiris_sup,<0.131.0>]
```

mkuratczyk:
Problems can be seen with a simple `make start-cluster` as well:
```
make start-cluster
stream-perf-test  -sc 100 -x 100 -y 0 -r 1000
make restart-cluster
```

I'm seeing:
```
[warning] <0.2653.0> Coordinator timeout on server 'rabbit-3@FH7Q7R4JQ1' when processing command restart_stream
[warning] <0.2653.0> Failed to transfer leadership of stream queue 'stream-093' in vhost '/': {error, coordinator_unavailable}
```

and:
```
[warning] <0.1516.0> Error/Timeout when retrieving broker ''rabbit-2@FH7Q7R4JQ1'' metadata: {error, {erpc, timeout}}
```

After a few minutes, it's still trying to drain the first node. As soon as I stop stream-perf-test, the restart can make progress and succeeds pretty quickly.

MugemaneBertin2001:
@mkuratczyk thanks for running this on a real cluster — those two reports pointed at a genuine problem with the drain loop under load. Pushed 571f7ff225.

**On the `make start-cluster` + `stream-perf-test` case** (`Coordinator timeout on server rabbit-3…`, `{error, coordinator_unavailable}` on `stream-093`)

Root cause was on my side: the drain loop was firing `restart_stream/2` serially for every local leader with up to 3 retries. Each command can wait up to `stream_cmd_timeout` (30 s) per raft member before falling through — so a single overloaded coordinator turns into minutes of wasted wall-clock while more commands pile up behind. That's why the drain never made progress until `stream-perf-test` stopped.

`transfer_leadership_of_local_stream_leaders/1` now:
- caps total wall-clock at 30 s (remaining streams fall back to shutdown-time re-election);
- bails out on the first `{error, coordinator_unavailable}` or `{timeout, _}` from `restart_stream/2` — hammering an overloaded coordinator only prolongs the outage;
- drops retries from 3 to 2;
- inserts a 25 ms breather between successful transfers so the coordinator command queue can drain;
- logs a `transferred=N, skipped=N, remaining=N` summary so it's clear from the logs what actually happened.

**On the K8s `{badrpc,{'EXIT',killed}}` from `osiris_replica` on the surviving nodes**

That's `restart_stream/2` semantics: the coordinator sets `target = stopped` for **every** member (writer + replicas), so replicas on the other nodes are killed and restarted alongside the writer. `osiris_server_sup` brings them back, but the transient stop-restart shows up as ugly crash reports on the surviving nodes. This is inherent to the current coordinator API — I don't see a lighter "just move the writer" primitive in `rabbit_stream_coordinator`. Happy to explore adding one (or capping the number of parallel restarts) if you think that's the right direction, but it feels like a separate change.

Ready for another test cycle when you have a moment. Title updated back — the DO NOT MERGE tag was left on so the maintainer flip is intentional if you want to keep it there.

cc @kjnilsson

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1656,6 +1659,127 @@ transfer_leadership_of_stream_coordinator(TransferCandidates) ->
...
+            ?LOG_WARNING("Leadership transfer for streams hosted on this node cut short "
+                         "(transferred=~b, skipped=~b, remaining=~b); shutdown-time "
+                         "re-election will handle the remainder",
+                         [Ok, Skipped, Remaining])
+    end.
+
+transfer_leaders_until([], _Candidates, _Deadline, Acc) ->
+    Acc;
+transfer_leaders_until([Q | Rest] = Queues, Candidates, Deadline, Acc) ->
+    case erlang:monotonic_time(millisecond) >= Deadline of
```
lukebakken, line 1700: The deadline is only checked between streams. `restart_stream/2` -> `process_command/2` retries each coordinator member at `cmd_timeout()` (30s) before giving up, so one stuck stream can block ~90s on a 3-member coordinator - well past the 30s budget. Cap intent is per-stream count, not wall-clock; worth clarifying in the comment or bounding the call itself.

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1656,6 +1659,127 @@ transfer_leadership_of_stream_coordinator(TransferCandidates) ->
...
+%% Under a busy coordinator (e.g. many active publishers), each
+%% restart_stream/2 command can take up to the coordinator command timeout
+%% per coordinator member before failing. To avoid dragging drain out for
+%% minutes, cap the total wall-clock spent on per-stream transfers and let
+%% shutdown-time re-election handle any remainder.
+-define(STREAM_DRAIN_BUDGET_MS, 30000).
+%% Small pause between per-stream restart_stream commands to reduce
+%% coordinator command queue pressure when there are many local leaders.
+-define(STREAM_DRAIN_INTER_STREAM_MS, 25).
+-define(STREAM_DRAIN_RESTART_ATTEMPTS, 2).
```
lukebakken, line 1671: `beb58de9cd` set this to 3, arguing replicas need a couple of 500ms backoffs to tie the writer's tail before `preferred_leader_node` wins; `571f7ff225` dropped it to 2 (one backoff). Deliberate? Is CI stable across repeated runs of `stream_leadership_transfer` at 2?

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1656,6 +1659,127 @@ transfer_leadership_of_stream_coordinator(TransferCandidates) ->
...
+        {ok, NewLeader} when NewLeader =:= node() ->
+            %% `restart_stream` honours `preferred_leader_node` only when
+            %% the preferred node is tied with the current writer on the
+            %% stopped tail. If the writer just advanced its epoch (as
+            %% happens right after being forced onto this node), the
+            %% replicas need a moment to catch up before another attempt
+            %% can tie them and let the hint win.
+            timer:sleep(?STREAM_DRAIN_RESTART_BACKOFF_MS),
+            restart_stream_off_local(Q, QNameStr, Preferred, Attempts - 1);
+        {ok, NewLeader} ->
```
lukebakken, line 1762: Only `NewLeader =:= node()` is rejected. `select_leader/2` isn't constrained to `Preferred`, so if another replica-hosting node is draining concurrently it can win and we count it as transferred. Narrow (rolling restarts drain one at a time), but that node then shuts down and re-elects under load.

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1656,6 +1659,127 @@ transfer_leadership_of_stream_coordinator(TransferCandidates) ->
...
+    transferred | skipped | {stop, term()}.
+restart_stream_off_local(_Q, QNameStr, _Preferred, 0) ->
+    ?LOG_WARNING("Gave up transferring leadership of stream ~ts off ~tp: "
+                 "still local after retries",
+                 [QNameStr, node()]),
+    skipped;
+restart_stream_off_local(Q, QNameStr, Preferred, Attempts) ->
+    Options = #{preferred_leader_node => Preferred},
+    case rabbit_stream_coordinator:restart_stream(Q, Options) of
+        {ok, NewLeader} when NewLeader =:= node() ->
```
lukebakken, line 1753: This works around `select_leader/2` treating `preferred_leader_node` as a tie-breaker only (single longest-tail member is returned directly, `coordinator.erl:2238`). Retrying hopes replicas tie within the backoff. Longer term, coordinator-side node exclusion (like QQ drain stopping the local Ra server) would be deterministic.

deps/rabbit/src/rabbit_amqqueue.erl
```diff
@@ -1330,6 +1331,13 @@ list_local_leaders() ->
          amqqueue:is_quorum(Q),
          amqqueue:get_state(Q) =/= crashed, amqqueue:get_leader_node(Q) =:= node()].
 
+-spec list_local_stream_leaders() -> [amqqueue:amqqueue()].
+list_local_stream_leaders() ->
```
lukebakken, line 1335: This is `is_local_to_node/2` inlined (and calls `get_pid/1` twice). Could be `[Q || Q <- list_local_stream_queues(), is_local_to_node(amqqueue:get_pid(Q), node())]` - matches `list_local/1`.

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1656,6 +1659,127 @@ transfer_leadership_of_stream_coordinator(TransferCandidates) ->
...
+-spec transfer_leadership_of_local_stream_leaders([node()]) -> ok.
+transfer_leadership_of_local_stream_leaders([]) ->
+    ?LOG_WARNING("Skipping stream leadership transfer: no candidate "
+                 "(online, not under maintenance) nodes to transfer to");
+transfer_leadership_of_local_stream_leaders(TransferCandidates) ->
+    LocalLeaders = rabbit_amqqueue:list_local_stream_leaders(),
+    ?LOG_INFO("Will transfer leadership of ~b streams with current leader on this node",
+              [length(LocalLeaders)]),
+    Deadline = erlang:monotonic_time(millisecond) + ?STREAM_DRAIN_BUDGET_MS,
+    Outcome = transfer_leaders_until(LocalLeaders, TransferCandidates, Deadline,
```
lukebakken, line 1683: `remaining` is only ever set on early exit; it's `total - transferred - skipped`. Could drop it from the map and compute once at the call site.

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -1656,6 +1659,127 @@ transfer_leadership_of_stream_coordinator(TransferCandidates) ->
...
+                                                            fun (V) -> V + 1 end, Acc))
+            end
+    end.
+
+-spec transfer_leadership_of_stream(amqqueue:amqqueue(), [node()]) ->
+    transferred | skipped | {stop, term()}.
+transfer_leadership_of_stream(Q, TransferCandidates) ->
+    %% A stream can only elect a new leader on a node that already hosts a replica.
+    Replicas = get_nodes(Q),
+    QNameStr = rabbit_misc:rs(amqqueue:get_name(Q)),
```
lukebakken, line 1727: Formatted for every stream, but on the `transferred` path only used by a `?LOG_DEBUG`. Cheap to format lazily in the branches that log at info/warning.

lukebakken (commented):
A few things worth a look, mostly around the drain-time budget. Main question: does this bound drain **wall-clock**, or just the **number of streams** we attempt? See inline on the deadline check.

PR #16877, opened 2026-07-05
feat: implement native Unix Domain Socket support

Piyush0049:
- Extend listeners config to natively support UDS paths
- Bypass DNS resolution for local socket paths
- Fix ntoa usage across all core modules and plugins to prevent crashes
- Identify UDS as a secure loopback connection to permit guest login
- Add integration tests for connection lifecycles and pub/sub routing

## Proposed Changes

This PR introduces native support for Unix Domain Sockets (UDS) across the RabbitMQ core and plugin ecosystem, closing **Issue #7311**.

As noted by @michaelklishin and @jvary in #7311, the primary roadblock to implementing UDS previously was that various monitoring paths, CLI tools, and plugins implicitly assumed a `(hostname, port)` pair, causing crashes when they received a `{local, Path}` tuple.

To resolve this without breaking the ecosystem, this PR systematically audits and hardens those paths:
1. **Core Networking**: Added `{local, _}` as a recognized `loopback` connection in `rabbit_net:is_loopback/1`, allowing the default `guest` user to securely authenticate over local sockets.
2. **Monitoring & Web Server**: Replaced `inet_parse:ntoa` with the UDS-aware `rabbit_misc:ntoa` inside `rabbit_cowboy_stream_h.erl`. This ensures the Management UI and Prometheus endpoints no longer crash Cowboy when access logs are generated for a UDS client.
3. **CLI Diagnostics**: Rewrote interface formatting in the Elixir CLI codebase (`listeners.ex`) to gracefully handle non-IP interfaces, preventing `rabbitmq-diagnostics listeners` and `report` from crashing.
4. **Plugins**: Safely bypassed DNS resolution and replaced `inet:ntoa` assumptions in MQTT, Web STOMP, and HTTP Auth backends.
5. **Testing**: Added rigorous end-to-end data plane testing (`test_uds_publish_consume`) in `unix_domain_socket_SUITE.erl` to guarantee payload integrity over local sockets.

*Maintainers: As this touches several core networking and monitoring formatting paths, your review and inputs are heavily appreciated! I'm fully open to iterating on this based on your feedback.*

## Configuration Syntax

A UDS listener is configured with the same `listeners.tcp.*` (or `listeners.ssl.*`) option used for TCP listeners:

```ini
listeners.tcp.1 = unix:/var/run/rabbitmq/rabbitmq.sock:0
```

The `unix:` (or equivalent `local:`) prefix and the trailing `:0` are both required. This is a consequence of how the value is parsed by cuttlefish:

- `listeners.tcp.$name` uses the datatype union `[integer, ip, domain_socket]`, and cuttlefish tries each converter in order. A bare path cannot be positively identified as a socket, so `cuttlefish_datatypes:validate_uds/1` keys on the `unix:`/`local:` prefix to disambiguate a socket path from an IP address.
- The `domain_socket` converter reuses the same `address:port` splitting as the IP and FQDN converters, so a `:port` component must be present. A Unix domain socket has no port, so the only accepted value is `0` (mirroring the `inet` `local_address` type, where the port is always `0`).

The mandatory `:0` is a cuttlefish parsing limitation, tracked upstream at Kyorai/cuttlefish#83.

## Types of Changes

- [ ] Build system and/or CI

## Checklist


jvary:
Hello! Thanks @Piyush0049 for taking over. 
I'm surprised by how little code change was needed to squeeze the feature in. Clever tricks! 

My Erlang is a bit rusty (and even more so the RabbitMQ codebase), but LGTM.

lukebakken (changes requested):
## Review: PR #16877 - native Unix Domain Socket support

> Note: this review was produced by an AI agent with guidance and direction from @lukebakken. The findings below were verified by running code against real Unix Domain Sockets on Erlang/OTP 27, and @lukebakken has reviewed the conclusions before posting.

The AI verified the central claims by running the actual logic against real UDS sockets on Erlang/OTP 27, plus the real cuttlefish and Elixir code. The change does not currently achieve its primary goal, and it needs a different foundation before it can be merged.

## Summary

The PR is built on the premise that a UDS peer address reaches RabbitMQ as the tuple `{local, Path}`, and it adds `{local, _}` handling to formatting and loopback code on that basis. In practice, `inet:peername/1` for an accepted UDS socket returns `{ok, {local, <<>>}}`, and `rabbit_net:socket_ends/2` decomposes that as `Address = local` (a bare atom) and `Port = <<>>`. Every fix in the PR targets the `{local, Path}` tuple shape, which sits downstream of a decomposition that has already destroyed that shape. The core networking path (`socket_ends/2`) is untouched, so guest-over-UDS does not work, and the test suite masks the failure by disabling the loopback check.

## Blocking findings

### B1. Guest login over UDS does not work, so the PR's primary goal is not met

The new `is_loopback({local, _}) -> true` clause in rabbit_net.erl is effectively dead code on the connection-authentication path.

`check_user_loopback(Username, PeerHost)` is called with `PeerHost` produced by `rabbit_net:socket_ends/2` in all three connection readers: `rabbit_reader.erl:1553` (AMQP 0-9-1), `rabbit_amqp_reader.erl:869` (AMQP 1.0), and `rabbit_mqtt_processor.erl:204` (MQTT). `socket_ends/2` matches the `peername` result with `{ok, {FromAddress, FromPort}}`. For UDS, `peername` returns `{ok, {local, <<>>}}`, so `FromAddress` binds to the atom `local`. `is_loopback(local)` matches none of the `{local, _}` or IP clauses and falls through to `is_loopback(_) -> false`.

End-to-end reproduction using the real `is_loopback/1` and `socket_ends/2` logic against a live UDS socket:

```
is_loopback(AcceptedSocket) = false
socket_ends -> PeerHost=local PeerPort=<<>> Host=local Port=<<"/tmp/uds_test_real.sock">>
is_loopback(PeerHost) = false
```

With the default `loopback_users = [<<"guest">>]` (verified in deps/rabbit/ebin/rabbit.app), the check evaluates `false orelse not lists:member(<<"guest">>, [<<"guest">>])`, which is `false`, so it returns `not_allowed`. Guest is rejected over UDS, the opposite of the stated goal.

Note there are two distinct paths. The web/HTTP path (`rabbit_web_dispatch_access_control.erl:142`) calls `is_loopback(IP)` with `IP` from `cowboy_req:peer/1`, which does yield a `{local, _}` tuple, so the new clause works there. But the AMQP/MQTT connection path, which is what this PR is really about, does not. That asymmetry is why the change looks plausible but does not work as intended.

### B2. The test suite hides B1 by disabling the loopback check

`unix_domain_socket_SUITE.erl` lines 47 and 72 call `set_env(rabbit, loopback_users, [])` before connecting, which globally disables the loopback restriction, so the tests pass regardless of B1. The test that would actually exercise the feature (guest connecting over UDS with the default `loopback_users`) is exactly the one that would fail.

Additional test issues:

- `set_env(rabbit, loopback_users, [])` mutates a node-global setting and never restores it, leaking state into any later suite on the same node
- Hardcoded `/tmp/*.sock` paths ignore the `rabbit_ct_helpers` temp-dir conventions and the 108-character `sun_path` limit; there is no `end_per_testcase` cleanup, so a mid-test failure leaks both the socket file and the dynamically started listener
- `timer:sleep(500)` to wait for listener startup is racy

### B3. Config regression: the `string` catch-all turns invalid listener config into a boot crash

rabbit.schema changes `listeners.tcp.$name` and `listeners.ssl.$name` from `[integer, ip]` to `[integer, ip, string]`. Cuttlefish tries datatypes left to right and takes the first that succeeds (`cuttlefish_generator:foldm_either/3`), and `from_string(String, string)` accepts any list, so `string` becomes an unconditional catch-all.

Verified against real cuttlefish plus the real `tcp_listener_addresses/1` clauses:

- Before: `listeners.tcp.1 = localhost` was rejected with a config conversion error (`{error,{conversion,{"localhost",'IP'}}}`)
- After: it parses as the string `"localhost"`, reaches `tcp_listener_addresses("localhost")`, matches no clause (not integer, not a tuple, does not start with `/` or `.`), and crashes at boot with a `function_clause` error

So a typo in a listener value now crashes the node instead of producing a clean validation error. Cuttlefish already ships a purpose-built `domain_socket` datatype (cuttlefish_datatypes.erl:428, producing `{local, Path, Port}` with validation) that would be the idiomatic choice, though using it requires a matching `tcp_listener_addresses` clause.

### B4. Address representation is inconsistent (string at config time, binary at runtime)

`rabbit_misc:ntoa({local, Path}) -> Path` returns `Path` unchanged, but `Path` is a string when it comes from config (listener record `ip_address = {local, "..."}`) and a binary when it comes from `inet:peername`/`sockname` or `cowboy_req:peer` (OTP's `returned_non_ip_address() :: {'local', binary()}`, see otp/lib/kernel/src/inet.erl:444).

Any call of the form `list_to_binary(rabbit_misc:ntoa(X))` crashes if `X` is `{local, BinaryPath}` (verified: `list_to_binary(<<"x">>)` raises `badarg`). The call sites the PR audits happen to receive the string form today (for example rabbit_mgmt_wm_health_check_certificate_expiration.erl:113 and rabbit_mgmt_format.erl:197 read the listener record's string path), so they do not crash now, but the mixed representation is a latent trap. `rabbit_misc:ntoa/1` should normalize `{local, Path}` to a consistent type rather than passing the path through untouched.

## Non-blocking and minor

- M1 (dead code, no effect): `rabbit_mqtt_util.erl`'s `ip_address_to_binary/1` is exported and imported into `rabbit_mqtt_processor`, but has zero call sites anywhere in the tree (the unused import predates this PR). The change is harmless but does nothing.
- M2 (garbled but non-crashing output): because `socket_ends/2` yields the atom `local`, AMQP connection strings for UDS render as `local:<<>> -> local:<<"/path">>` (via `rabbit_net:connection_string/2` and `maybe_ntoab/1`, which passes atoms through). Auth-attempt metrics record the address as `<<"local">>` (rabbit_core_metrics.erl false branch, `to_binary(local)`), losing the socket path.
- M3 (latent crash if reverse DNS is enabled): with `reverse_dns_lookups = true`, `rabbit_net:rdns(local)` calls `tcp_host(local)`, which calls `inet:gethostbyaddr(local)` and errors; the result then flows into formatting. Not the default, but unhandled.
- M4 (non-idiomatic guard): the new `tcp_listener_addresses/1` guard uses `length(Path) > 0` (an O(n) traversal) where `Path =/= []` suffices, and the inline `hd(Path) =:= $/ orelse hd(Path) =:= $.` is unusual for this module. Not a bug (the `length > 0` guard does correctly protect `hd`).
- M5 (cosmetic): UDS listeners report `port: 0` in `rabbitmq-diagnostics listeners` and the HTTP API. Harmless but potentially confusing.

## What is correct in the PR

- The data plane works. `gen_tcp:connect({local, Path}, 0, [local | RABBIT_TCP_OPTS])` connects and transfers data over UDS with the exact amqp_client options, including `{nodelay, true}` (OTP silently ignores nodelay on UDS, no error). This is why the loopback-disabled test passes.
- The Elixir CLI `format_interface({:local, path})` is correct; `path` is a bound variable and it handles both string and binary paths.
- Ranch natively supports `{ip, {local, Path}}` and cleans up the socket file (ranch_tcp.erl:279).
- The listener registration path (`tcp_host/1`, `tcp_name/1`) handles `{local, StringPath}` without crashing, and the management listener-listing path works because it uses the string form.

## Suggested direction

The fix needs to live at the source of the address, not at the formatting leaves:

1. Teach `rabbit_net:socket_ends/2` (and the `is_loopback(Sock)` socket clause) to recognize `peername`/`sockname` returning `{local, Binary}` and preserve it as a `{local, Path}` value rather than decomposing it into `{Address, Port}`
2. Make the `{PeerHost, PeerPort, Host, Port}` consumers in `rabbit_reader`, `rabbit_amqp_reader`, and `rabbit_mqtt_processor` prepared for a UDS peer
3. Normalize `rabbit_misc:ntoa/1` to return a consistent type for `{local, Path}` regardless of whether the path arrived as a string or a binary
4. Prefer cuttlefish's existing `domain_socket` datatype (with a matching `tcp_listener_addresses` clause) over a `string` catch-all, to preserve config validation
5. Add a test that connects as guest over UDS with the default `loopback_users`, and restores any env it changes

I have converted this PR to draft as we begin the process of refining it. Once it's more ready, we can change the status.

michaelklishin:
It's a good question whether UDS support makes much sense for RabbitMQ.

Piyush0049:
Hi @michaelklishin and @lukebakken,

Thanks a lot for taking the time to do such a deep review, really appreciate the detailed breakdown of the edge cases.

To answer the question about whether UDS support makes sense for RabbitMQ, here are the main use cases I had in mind:

1. **Sidecar/Proxy setups:** In Kubernetes deployments, it's pretty common to have local sidecars (Envoy, HAProxy, custom agents) running alongside the broker. Using a Unix socket instead of TCP avoids exposing localhost ports entirely, which helps prevent port exhaustion and conflicts
2. **OS-level security:** UDS lets you control access through standard file permissions (`chmod`/`chown`) and user groups, giving you a solid security boundary for local clients without needing to set up TLS certificates just for local traffic
3. **Less overhead:** Skipping the TCP/IP stack (no routing, checksums, or window management) means lower CPU usage and latency for local publishers/consumers

On Luke's technical findings, you're absolutely right. Patching the formatting functions at the leaves (like `ntoa`) was the wrong approach since the core `socket_ends/2` decomposition was already stripping the `{local, Path}` tuple into a bare `local` atom before anything downstream could use it. The guest loopback auth was broken under the hood and the tests were hiding it by disabling `loopback_users`.

I understand it may not be a universal requirement, but for containerized/local-only deployments these patterns are very common, and supporting UDS cleanly would make RabbitMQ more composable in those environments. I fully agree with the suggested direction. I'll rework this from the source:
- Fix `rabbit_net:socket_ends/2` to preserve `{local, Path}` instead of decomposing it
- Update the downstream readers (`rabbit_reader`, `rabbit_amqp_reader`, `rabbit_mqtt_processor`) to handle UDS peers
- Normalize `rabbit_misc:ntoa/1` to return a consistent type
- Switch to Cuttlefish's `domain_socket` datatype instead of the `string` catch-all
- Write a proper test that validates guest login over UDS with default `loopback_users`

I'm fully committed to seeing this through, happy to invest the time needed to get this right.

michaelklishin:
@Piyush0049 sure, I understand when USD makes sense. You don't have to sell it hard.

Consider how many *RabbitMQ clients* can use UDS, though. It's a niche feature all things considered. I am not against it but I don't want to overpromise, our team may decide to not accept this PR.

Piyush0049:
@michaelklishin Apologies, definitely not trying to oversell it. Like I mentioned, I completely understand it's a niche feature and not a universal requirement for most clients. 
I'm totally fine with it if the team ultimately decides not to merge. Either way, I'm ready to work on it and contribute to fixing the architecture based on review, even if it's just for the learning experience.

Piyush0049:
Hi @lukebakken, I've pushed an update in this PR.

Here is a summary:
* Reworked `rabbit_net:socket_ends/2` to preserve the `{local, Path}` tuple instead of decomposing it, which natively fixes the `is_loopback` check so guest login works as expected.
* Swapped the string catch-all in `rabbit.schema` for the Cuttlefish `domain_socket` datatype to ensure proper config validation.
* Normalized `rabbit_misc:ntoa/1` and `ntoab/1` to guarantee consistent string and binary returns.
* Added a guard to `rabbit_net:rdns/1` to bypass lookups for local paths and prevent the reverse DNS crash.
* Rewrote the test suite to remove the global `loopback_users` override. It also now cleanly handles the 108-character `sun_path` limit dynamically and properly cleans up the listeners and socket files in `end_per_testcase`.
* Cleaned up the minor style nit and removed the unused `ip_address_to_binary/1` function.

lhoguin (approved):
LGTM

lhoguin:
Please follow the instructions around the error at https://github.com/rabbitmq/rabbitmq-server/actions/runs/29192409204/job/87321204187?pr=16877 and I will relaunch CI.

Piyush0049:
Thanks for the approval, @lhoguin! 

I have pushed a fix. I registered the new `unix_domain_socket` test suite in the `PARALLEL_CT_SET_1_A` variable inside `deps/rabbit/Makefile`.

Please do let me know if any other changes are required.

michaelklishin:
So far there does not seem to be much pushback on our team. At least conceptually.

Which means spending more time on this PR can be justified :)

Piyush0049:
> So far there does not seem to be much pushback on our team. At least conceptually.
> 
> Which means spending more time on this PR can be justified :)

I’m very glad to hear that 😊. Please do let me know if any changes are required; I’d be happy to contribute.

lukebakken:
I just rebased `Piyush0049:feature/unix-domain-sockets` on `main` and will be doing one more review with the 🧞

deps/rabbit_common/src/rabbit_net.erl
```diff
@@ -294,6 +299,7 @@ is_loopback(Sock) when is_port(Sock) ; ?IS_SSL(Sock) ->
 is_loopback({127,_,_,_})             -> true;
 is_loopback({0,0,0,0,0,0,0,1})       -> true;
 is_loopback({0,0,0,0,0,65535,AB,CD}) -> is_loopback(ipv4(AB, CD));
+is_loopback({local, _})              -> true;
```
lukebakken, line 303: This clause is dead on the raw-socket path. is_loopback(Sock) at line 292 calls peername, which on an accepted UDS socket returns {ok, {local, <<>>}}; the {ok, {Addr, _Port}} pattern then binds Addr to the atom local, so it falls through to false. Verified live: rabbit_auth_mechanism_plain:build_auth_props/2 produces {is_loopback, false} for a guest UDS connection, and internal_loopback refuses it with "the socket or address was not from loopback/localhost." A {ok, {local, _}} clause is needed in the peername case, not only the address-tuple clause.

deps/rabbit_common/src/rabbit_core_metrics.erl
```diff
@@ -424,7 +424,7 @@ update_auth_attempt(RemoteAddress, Username, Protocol, Incr) ->
         {ok, true} ->
             Addr = case inet:is_ip_address(RemoteAddress) of
                        true ->
-                           list_to_binary(inet:ntoa(RemoteAddress));
+                           list_to_binary(rabbit_misc:ntoa(RemoteAddress));
                        false ->
                            rabbit_data_coercion:to_binary(RemoteAddress)
```
lukebakken, line 434: Crashes every UDS connection when track_auth_attempt_source is enabled. RemoteAddress is {local, Path}, is_ip_address is false, so this calls to_binary({local, Path}), which has no tuple clause. Verified live: function_clause here fires from the auth success branch (rabbit_reader:1555), killing the connection right after it authenticates.

deps/rabbitmq_auth_backend_http/src/rabbit_auth_backend_http.erl
```diff
@@ -294,10 +294,4 @@ join_tags(Tags) ->
...
     rabbit_data_coercion:to_list(unknown);
 parse_peeraddr(PeerAddr) ->
-    handle_inet_ntoa_peeraddr(inet:ntoa(PeerAddr), PeerAddr).
-
--spec handle_inet_ntoa_peeraddr({'error', term()} | string(), inet:ip_address() | unknown) -> string().
-handle_inet_ntoa_peeraddr({error, einval}, PeerAddr) ->
-    rabbit_data_coercion:to_list(PeerAddr);
-handle_inet_ntoa_peeraddr(PeerAddrStr, _PeerAddr0) ->
-    PeerAddrStr.
+    rabbit_misc:ntoa(PeerAddr).
```
lukebakken: Dropping the {error, einval} fallback regresses non-IP peeraddrs. ntoa/1 handles neither the atom local (UDS peer via get_authz_data_from({socket, _})) nor a hostname binary (reverse DNS); both return {error, einval}, and quote_plus/1 then crashes with function_clause. Verified live on both inputs. Please restore an einval-to-string fallback.

deps/rabbit/src/rabbit_networking.erl
```diff
@@ -146,6 +147,10 @@ fix_ssl_options(Config) ->
 
 -spec tcp_listener_addresses(listener_config()) -> [address()].
 
+tcp_listener_addresses({local, Path, _Port}) ->
+    [{{local, Path}, 0, local}];
+tcp_listener_addresses(Path) when is_list(Path) andalso Path =/= [] andalso (hd(Path) =:= $/ orelse hd(Path) =:= $.) ->
```
lukebakken: The schema now accepts any string, but this guard only matches paths starting with / or ., so any other string (relative path, mistyped hostname) is function_clause at boot. Verified live: listeners.tcp.badstring = myapp.sock fails startup here with args ["myapp.sock"]. Prefer the typed {local, Path, Port} form and drop this string clause.

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -23,7 +23,7 @@
 ]}.
 
 {mapping, "listeners.tcp.$name", "rabbit.tcp_listeners",[
-    {datatype, [integer, ip]}
+    {datatype, [integer, ip, string, domain_socket]}
```
lukebakken: string shadows domain_socket: cuttlefish tries datatypes left to right and returns the first success, and from_string(_, string) accepts any string, so domain_socket's validation never runs. Verified: for local:/path.sock:0, the string datatype returns the raw unparsed string before domain_socket is reached. Drop string and rely on domain_socket (same on line 49).

deps/rabbit/test/unix_domain_socket_SUITE.erl
```diff
@@ -0,0 +1,134 @@
...
+    file:delete(SocketPath),
+
+    %% Dynamically start a new listener on the local unix socket
+    ok = rabbit_ct_broker_helpers:rpc(Config, rabbit_networking, start_tcp_listener, [SocketPath, 10]),
+
+    %% Let the listener start up
+    timer:sleep(500),
+
+    %% Connect using the Erlang AMQP Client over the Unix socket
+    ConnParams = #amqp_params_network{host = {local, SocketPath}, port = 0},
```
lukebakken, line 132: The CT broker runs with loopback_users=[], so guest connects over any transport regardless of loopback status; this suite would pass even with the is_loopback bug above. Set loopback_users=[<<"guest">>] and assert guest-over-UDS is accepted (and, ideally, that the internal_loopback backend accepts it too).

deps/rabbit/test/unix_domain_socket_SUITE.erl
```diff
@@ -0,0 +1,134 @@
...
+    SocketPath = ?config(uds_socket_path, Config),
+
+    %% Ensure it doesn't already exist from a crashed run
+    file:delete(SocketPath),
+
+    %% Dynamically start a new listener on the local unix socket
+    ok = rabbit_ct_broker_helpers:rpc(Config, rabbit_networking, start_tcp_listener, [SocketPath, 10]),
+
+    %% Let the listener start up
+    timer:sleep(500),
```
lukebakken: start_tcp_listener/2 returns ok synchronously once the listener child is up, so this timer:sleep(500) is unnecessary.

deps/rabbit/test/unix_domain_socket_SUITE.erl
```diff
@@ -0,0 +1,134 @@
...
+    ok = rabbit_ct_broker_helpers:rpc(Config, rabbit_networking, stop_tcp_listener, [SocketPath]),
+    file:delete(SocketPath),
+    ok.
+
+test_uds_connection_failure(Config) ->
+    SocketPath = ?config(uds_socket_path, Config),
+    file:delete(SocketPath),
+
+    ConnParams = #amqp_params_network{host = {local, SocketPath}, port = 0},
+    case amqp_connection:start(ConnParams) of
```
lukebakken: {error, _} passes on any failure. Assert the specific reason (e.g. enoent) so this cannot pass for the wrong reason.

deps/rabbit/test/unix_domain_socket_SUITE.erl
```diff
@@ -0,0 +1,134 @@
+-module(unix_domain_socket_SUITE).
```
lukebakken, line 7: Missing the standard MPL 2.0 / copyright header present on every other suite, and the file is committed as 100755 (should be 100644).

lukebakken (changes requested):
This is a review by Claude Code, guided by @lukebakken. The findings below were reproduced against a running node built from this branch (single node, default loopback_users, UDS listener plus 5672).

Requesting changes.

What works: the socket_ends/2 change correctly carries the {local, Path} tuple into the AMQP 0-9-1 reader, so on a default install guest can connect over a Unix socket and authenticate. That path is solid.

The problems are on paths that use a raw socket or a non-default configuration:

1. rabbit_net.erl: the new is_loopback({local, _}) clause is dead when called on a socket. is_loopback/1 calls peername, which on an accepted UDS socket returns {ok, {local, <<>>}}, and the existing {ok, {Addr, _Port}} clause binds Addr to the atom local. rabbit_auth_mechanism_plain calls is_loopback/1 on the raw socket, so with the internal_loopback backend a guest UDS connection is refused ("not from loopback/localhost"). Reproduced live.

2. rabbit_core_metrics.erl: with track_auth_attempt_source enabled, every UDS connection crashes in update_auth_attempt (rabbit_data_coercion:to_binary/1 has no clause for the {local, Path} tuple). The crash is on the auth success branch, so the connection dies right after authenticating. Reproduced live.

3. rabbit_auth_backend_http.erl: removing the einval fallback regresses non-IP peeraddrs. Both the UDS atom local and reverse-DNS hostname binaries make ntoa/1 return {error, einval}, which then crashes quote_plus/1. Reproduced live.

4. rabbit_networking.erl / rabbit.schema: the raw string datatype shadows the existing domain_socket datatype, and the string listener clause only accepts paths starting with / or ., so other string values crash at boot. Reproduced live: listeners.tcp.badstring = myapp.sock fails startup with function_clause in tcp_listener_addresses/1. Dropping string and relying on domain_socket addresses both.

5. unix_domain_socket_SUITE.erl: the CT broker runs loopback_users=[], so the feature's core guarantee (UDS is loopback, guest may connect) is never actually exercised. Setting loopback_users=[<<"guest">>] and asserting acceptance would have caught item 1.

Non-blocking:

6. The {local, Path} case is special-cased across roughly eight formatters rather than normalized once; items 2 and 3 are missed instances of that pattern. Normalizing at the socket_ends/2 boundary would remove the class.
7. rabbit_mqtt_processor.erl still imports the removed rabbit_mqtt_util:ip_address_to_binary/1.
8. Test nits (inline): missing MPL header and executable file mode, an unnecessary timer:sleep, and a failure test that asserts only {error, _}.

lhoguin:
@lukebakken Perhaps you can ask Claude to add tests for the issues it found and confirmed as well? Many of your inline comments would qualify ("verified live") and I don't think it'd cost much more tokens.

lukebakken:
Following up on the review: I pushed f5c6cf9c0cb026ba55c66aaf1926a7a66e4b89e5 with regression tests that reproduce the confirmed issues. They are expected to fail against the current branch and should pass once the fixes land. Written by Claude Code, guided by @lukebakken.

The new cases in unix_domain_socket_SUITE.erl:

- test_uds_is_loopback_on_raw_socket, test_uds_guest_with_loopback_users, test_uds_internal_loopback_backend - is_loopback returning false for a UDS peer (finding 1)
- test_uds_auth_attempt_source_tracking - the to_binary({local, Path}) crash with track_auth_attempt_source enabled (finding 2)
- test_uds_ntoa_local_atom - rabbit_misc:ntoa crashing on the bare atom local from a UDS peeraddr (finding 3)
- test_uds_tcp_listener_addresses - tcp_listener_addresses/1 crashing on a string path that does not start with / or . (finding 4)

The commit also adds rabbitmq_auth_backend_internal_loopback to TEST_DEPS, adds the MPL header, and fixes the file mode (755 to 644). The inline comments above still capture the requested changes; this just adds the failing tests that pin them.

Piyush0049:
Hey @lukebakken @lhoguin, I've pushed the changes after reading the review. The is_loopback(local) clause, the rabbit_core_metrics crash, the einval fallback in the HTTP auth backend, the schema/networking string issue, and the loose test assertion are all addressed. All 9 test cases in the suite pass locally including the regression tests you added. 
Let me know if anything else needs changing.

lukebakken (approved):

michaelklishin:
Moved to https://github.com/rabbitmq/rabbitmq-server/pull/16963.

PR #16881, opened 2026-07-06
stream coordinator: Fix writer liveness edge-cases

### LINKED ISSUE #16822: Stream Coordinator can leave a stream in `stopped` in a 2/3 node loss disaster
### Describe the bug

When a stream has its writer member on node A and replicas on nodes B and C, when nodes B and C are removed from the cluster in sequence (by `forget_cluster_node`), the writer on node A is terminated as a side effect of the first `delete_replica` and is not re-elected. The writer can't be healed by anything in the current stream coordinator API:

1. `evaluate_stream/3` has no branch that re-acts on a writer already in `{stopped, _, _}`
2. the only path to `{ready, NextEpoch}` is the `select_leader/2` call inside `member_stopped`'s `is_quorum` branch, and once the first `member_stopped` event has fired for the writer, no later event re-evaluates the quorum predicate
3. the `filter_command/3` clause for `delete_replica` rejects any further delete as `last_stream_member`

The log is still intact on A, but the writer can't be recovered. This sequence is actually OK as long as the writer was originally on B or C during the disaster event: then the stream can be recovered with a sequence of force-shrinks of Khepri, the Stream Coordinator (and any QQs), and then normal stream membership change commands.

### Reproduction steps

I found this behavior by simulating a disaster scenario where 2/3 nodes of a cluster are lost.

1. 3-node cluster with a stream declared.
1. Identify the node running the stream writer, call it `A`.
1. Kill nodes `B` and `C` (SIGTERM, simultaneous or staggered). Wait for `rabbit on node B/C down` on `A`. (Treat B and C as permanently lost now, all commands are against A from here on.)
1. Force-shrink Khepri: `rabbitmqctl force_standalone_khepri_boot`
1. Force-shrink the stream coordinator: `rabbitmqctl eval 'ra_server_proc:force_shrink_members_to_current_member({rabbit_stream_coordinator, node()}).'`
1. Issue `delete_replica` for `B` and then `C`. (`forget_cluster_node` also does this.) Note that the writer terminates on `A` (`osiris_writer:terminate/2 ... reason:shutdown`) when the first `delete_replica` is applied, and no `started writer ... in 2` is logged.

The condition for triggering the bug is precisely 'the stream's writer was on the surviving node at the time `delete_replica` is first issued'. Streams whose writer was on one of the dead nodes recover normally via the `{down, Pid, Reason}` -> `{down, Epoch}` path in `update_stream0/3`.

`add_replica` invoked on the stuck stream fails with:

```
[warning] queue '<Q>' ... failed to add replica on 'rabbit@<D>':
   {noproc,{gen_batch_server,call,[<OLD_WRITER_PID>,query_replication_state,5000]}}
```

where `<OLD_WRITER_PID>` is the PID that terminated at the `osiris_writer:terminate` line above.

`rabbitmq-streams restart_stream <Q>` hangs indefinitely and never returns. The command is applied to the Ra state machine, sets `target = stopped` on the single existing member (a no-op since it is already stopped), but `evaluate_stream/3` finds no actionable state transition and never emits a reply effect, so the `reply_to` Pid waits forever.

### Expected behavior

The stream coordinator should either not leave the writer member in a stopped state, or have an API to recover it. I would lean towards the former since that's how it behaves if the writer is on one of the lost nodes.

### Additional context

This was reproduced against `main` from a few weeks ago, 2bd344977625. Like QQ/Khepri/Coordinator force-shrinking, we can't guarantee that there won't be data loss in the disaster-recovery, but we can provide tools to recover availability. I think we might be able to fix it simply by re-evaluating `is_quorum` in `member_deleted`.

the-mikedavis:
This fixes https://github.com/rabbitmq/rabbitmq-server/issues/16822 and two other writer liveness edge-cases. The tail of the branch has a `proper_statem` suite for the stream coordinator and it could independently find #16822, plus two other bugs to fix. I also added a `proper_statem` for the SAC coordinator but it found no issues. Both suites balance liveness and safety properties:

| Suite | Liveness | Safety
|---|---|---
| stream coordinator | Eventually a writer reaches `running` in the most cooperative environment | <= 1 running writer allowed at any time
| SAC | Eventually a consumer becomes active if one is available in the most cooperative environment | <=1 active consumer allowed at any time


lukebakken (commented):
Reviewed the writer-liveness changes. Findings inline; two that don't anchor to a single line:

**delete/start race (see `elect_leader_if_quorum`, ~2190).** Clearing the `{starting,_}` latch on a `target=deleted` member lets `eval_replica` issue `delete_member` while a start may still be in flight, and `run_action` spawns these without per-member serialization. osiris doesn't fence it: `delete_child` on `not_found` falls through to `delete_directory` (`osiris_server_sup.erl:65-66`), and `osiris_replica:start/2` is a plain `start_child` with no epoch guard (`osiris_replica.erl:100-107`). A late start can recreate the process and directory after `member_deleted` has dropped the member. It's a concurrency race, so no deterministic repro — flagging the code path.

**Minor:** `needs_reelection/1` re-derives the live-writer check independently of `find_leader/1`; the two can drift. Could reuse the helper.

deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -2183,6 +2154,132 @@ find_leader(Members) ->
...
+                                 %% complete: member_started is fenced to the
+                                 %% stream epoch, so its reply is dropped and the
+                                 %% latch would never clear. Drop it so the member
+                                 %% is re-driven in the new epoch. A stopping latch
+                                 %% is kept, as member_stopped still matches on the
+                                 %% new role epoch.
+                                 M#member{role = {replica, NextEpoch},
+                                          preferred = false,
+                                          current = undefined};
+                             (_N, M) ->
```
lukebakken, line 2210: This catch-all bumps the role epoch but not `state`, so a `{down, E}` member ends up `role = {replica, E+1}, state = {down, E}`. The start clause at 2032 requires the state epoch to equal the stream epoch, so it never matches and the member is never re-driven (reproduced by driving the backstop with a down writer plus a stopped quorum). Since this path is newly reachable via `maybe_reelect_leader`, the state epoch needs advancing here too.

deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -2183,6 +2154,132 @@ find_leader(Members) ->
...
+                             (_N, #member{target = deleted} = M) ->
+                                 M;
+                             (_N, #member{current = {starting, _}} = M)
+                               when ?V8_OR_MORE(Vsn) ->
+                                 %% A start issued in the previous epoch can never
+                                 %% complete: member_started is fenced to the
+                                 %% stream epoch, so its reply is dropped and the
+                                 %% latch would never clear. Drop it so the member
+                                 %% is re-driven in the new epoch. A stopping latch
+                                 %% is kept, as member_stopped still matches on the
```
lukebakken, line 2205: This holds for members whose role epoch is bumped in this branch, but not for a `target = deleted` member: the clause at 2196 leaves its role at the old epoch, so its `member_stopped` is fenced out at 1610, `current` never clears, and `delete_member` never fires until a coordinator leader change. The comment overstates the guarantee.

deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -2183,6 +2154,132 @@ find_leader(Members) ->
...
+                            false
+                    end, Members),
+    case is_quorum(length(Nodes), map_size(StoppedInCurrent)) of
+        true ->
+            %% select leader
+            NewWriterNode = select_leader(Meta, StoppedInCurrent),
+            NextEpoch = Epoch + 1,
+            Members1 = maps:map(
+                         fun (N, #member{state = {stopped, E, _}} = M)
+                               when E == Epoch ->
```
lukebakken, line 2178: This clause matches before the `target = deleted` clauses below, so a deleted member that is `{stopped, Epoch, _}` gets rewritten to `{ready, NextEpoch}` while `target` stays `deleted`. Deletion still proceeds (low severity), but the intermediate state is inconsistent; reordering the deleted clauses ahead of this one avoids it.

deps/rabbit/src/rabbit_stream_coordinator.erl
```diff
@@ -90,6 +90,7 @@
 -define(V2_OR_MORE(Vsn), (Vsn >= 2)).
 -define(V5_OR_MORE(Vsn), (Vsn >= 5)).
 -define(V7_OR_MORE(Vsn), (Vsn >= 7)). %% SAC monitors no longer in monitors map
+-define(V8_OR_MORE(Vsn), (Vsn >= 8)). %% writer liveness fixes
```
lukebakken, line 93: Nit: CLAUDE.md prefers the comment on the line above rather than trailing. The `V7_OR_MORE` line above already trails, so optional.

deps/rabbit/test/rabbit_stream_coordinator_statem_SUITE.erl
```diff
@@ -0,0 +1,402 @@
...
+        {ok, P} -> put(world, deliver(W, P, Outcome));
+        none    -> ok
+    end,
+    ok.
+
+cmd_crash(Node, Reason) ->
+    W = world(),
+    case running_pid(Node, coord(W)) of
+        {ok, Pid} ->
+            R = case Reason of noconnection -> noconnection; _ -> boom end,
```
lukebakken, line 189: The generator at line 111 draws `elements([normal, noconnection])`, but `normal` is mapped to `boom` here, so a shrunk counterexample prints `normal` while the model saw `boom`. Generating `elements([boom, noconnection])` keeps the printed command and the injected reason in sync.

deps/rabbit/test/rabbit_stream_coordinator_statem_SUITE.erl
```diff
@@ -0,0 +1,402 @@
...
+
+world() -> get(world).
+coord({_Idx, Coord, _InFlight}) -> Coord.
+in_flight({_Idx, _Coord, InFlight}) -> InFlight.
+
+meta(N) ->
+    #{index => N, term => 1, machine_version => rabbit_stream_coordinator:version(), system_time => N * 2}.
+
+%% Build a pid whose node/1 is an arbitrary atom, without distribution.
+fake_pid(Node) ->
```
lukebakken, line 223: This duplicates `test_util:fake_pid/1` (already used by `rabbit_fifo_prop_SUITE`); the same copy is in `rabbit_stream_sac_coordinator_statem_SUITE:237`. Worth importing the shared helper instead.

deps/rabbit/test/rabbit_stream_sac_coordinator_statem_SUITE.erl
```diff
@@ -0,0 +1,345 @@
...
+world() -> get(world).
+
+erase_state() ->
+    _ = erase(world),
+    %% forget memoised connection pids from the previous run
+    [erase({pid, C}) || C <- ?CONNS],
+    ok.
+
+meta(Idx) ->
+    #{index => Idx, system_time => Idx * 2, machine_version => 8, term => 1}.
```
lukebakken, line 234: `machine_version` is hardcoded to `8`; the sibling `rabbit_stream_coordinator_statem_SUITE` uses `rabbit_stream_coordinator:version()`. This will silently stop tracking the current version after the next bump.

the-mikedavis:
Superseded by https://github.com/rabbitmq/rabbitmq-server/pull/16515

PR #16892, opened 2026-07-07
OAuth 2: introduce a special proxy endpoint, adopt it for the flow that benefits from it

michaelklishin:
Specifically for the flow that uses a client secret, which is a minority of deployments.

This allows us to keep the client secret value on
the server and pass it on to the IDP.

In this case RabbitMQ acts as an HTTPS client.
The trusted CA bundle can be configured via
`auth_oauth2.https.cacertfile` or system-wide
(Erlang/OTP's HTTP client should pick up
the system-wide bundle per RabbitMQ defaults).

This change tries very hard to be backwards compatible and meaningfully modernize that one
rare client secret flow we see used with only one IDP.

### DIFF
--- a/selenium/test/oauth/with-sp-initiated/token-endpoint-proxy.js
+++ b/selenium/test/oauth/with-sp-initiated/token-endpoint-proxy.js
@@ -1,11 +1,9 @@
 const assert = require('assert')
 const { buildDriver, goToHome, captureScreensFor, teardown } = require('../../utils')
 
-// Regression test for the OAuth 2 client secret disclosure (GHSA-f9f2-q3jf-wfj3).
-// The unauthenticated bootstrap script must not carry management.oauth_client_secret;
-// resource servers that need it are served through the server-side token endpoint proxy.
-// The script is fetched through the browser so it uses the browser's trust of the
-// management UI's TLS certificate.
+// The unauthenticated bootstrap script must not expose
+// management.oauth_client_secret. Fetched through the browser so it uses the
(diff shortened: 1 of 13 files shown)


MarcialRosales (commented):
The implementation looks good but do you think we may need some integration tests on the new proxy endpoint, similar to the integration tests we have in rabbit_mgmt_http_SUITE. There is some logic going on, i.e. reading configuration, making http requests, forwarding the result...

michaelklishin:
@MarcialRosales I'll see what can be done, starting with the Selenium failure.

michaelklishin:
@MarcialRosales ready for another round :)

selenium/test/oauth/with-sp-initiated/token-endpoint-proxy.js
MarcialRosales, line 1: It does not cause any harm to have this selenium test case but I think with the unit tests that verifies that we no longer return the client secrets via `auth_settings` function and after you have added the integration tests that verifies the http flow works, this particular test is not needed. With the existing selenium tests that verifies the oauth2 flows end to end, we are validating the proxy endpoint you have introduced in this PR. what do you think? Anywho, it does not cause any harm.

MarcialRosales (approved):

selenium/test/oauth/with-sp-initiated/token-endpoint-proxy.js
michaelklishin: @MarcialRosales it verifies the `bootstrap.js` bits, that's all. Let's keep it.

PR #16894, opened 2026-07-07
[mgt] Remove LOGIN_SESSION_TIMEOUT from the m Cookie

MarcialRosales:
## Proposed Changes

Remove unnecessary pseudo-cookie (entry stored within the m cookie) .
This change is part of an effort of securing cookies. All cookies that exists in the management ui are now owned by the server and fully secured with the right attributes. The only cookie left is m, controlled by the client side, which only exists for the pseudo-cookie LOGGED_IN which has a TTL that controls the lifespan of the client-side session. This cookie has no value hence it is fine that it is not fully secured because javascript created cookies cannot be fully secured. 
Also, this pseudo-cookie needs to be accessed by javascript hence it cannot be owned by the server.

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/prefs.js
+++ b/deps/rabbitmq_management/priv/www/js/prefs.js
@@ -44,7 +44,7 @@ function get_auth_scheme() {
 function clear_auth() {
     clear_local_pref(CREDENTIALS)
     clear_local_pref(AUTH_SCHEME)
-    clear_cookie_value(LOGIN_SESSION_TIMEOUT)
+    clear_local_pref(LOGIN_SESSION_TIMEOUT)
     clear_cookie_value(LOGGED_IN)
     clear_local_pref(AUTH_RESOURCE)
 }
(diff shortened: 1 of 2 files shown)


deps/rabbitmq_management/priv/www/js/prefs.js
```diff
@@ -76,26 +76,15 @@ function default_hard_session_timeout() {
 }
 
 function update_login_session_timeout(login_session_timeout) {
-    //var auth_info = get_cookie_value('auth');
-    if (get_cookie_value(LOGIN_SESSION_TIMEOUT) != undefined || !has_auth_credentials()) {
+    if (get_local_pref(LOGIN_SESSION_TIMEOUT) != undefined || !has_auth_credentials()) {
```
michaelklishin, line 81: @MarcialRosales this loses expiration of the value. Local storage values do not expire.

This means that the session expiration setting is set once and then defaults to 8 hours forever, as far as I understand.

michaelklishin (changes requested):

michaelklishin:
@MarcialRosales I have addressed my review comment by clearing the local storage value that affects session duration on every successful login, and doing that in a way that would not break with OAuth 2 that calls `set_token_auth` on every refresh.

This looks correct now w.r.t. the session length limit.

michaelklishin (approved):

MarcialRosales:
@michaelklishin Thanks! and well spotted!

PR #16942, opened 2026-07-15
Federation: be more strict when validating URIs

michaelklishin:


michaelklishin:
@mergifyio backport v4.3.x

PR #16946, opened 2026-07-15
Introduce an opt-in setting that forces uploaded definitions file to have a .json extension

MarcialRosales:
## Proposed Changes

Some deployments require strict filename extension validation for definition file uploads. When enabled, the management UI and the server both validate that the uploaded file has a .json extension.

On the client side, the management UI sets the accept=".json" attribute on the file input, which tells the browser to filter the file picker to .json files only. On the server side, the filename extension is independently validated before the file is processed.

The new configuration setting is `management.require_definition_json_extension`. It defaults to false for backwards compatibility. Regardless of its value, the server always validates that the uploaded content is valid JSON before importing it.

Docs PR: https://github.com/rabbitmq/rabbitmq-website/pull/2567

## Types of Changes

- [ ] Build system and/or CI

### DIFF
--- a/deps/rabbitmq_management/priv/schema/rabbitmq_management.schema
+++ b/deps/rabbitmq_management/priv/schema/rabbitmq_management.schema
@@ -487,6 +487,16 @@ end}.
 {mapping, "management.http.hide_allow_header", "rabbitmq_management.hide_http_allow_header",
     [{datatype, boolean}]}.
 
+%% When enabled, the management plugin only accepts definition files with a
+%% .json extension when uploaded via the management UI. Files with any other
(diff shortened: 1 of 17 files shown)


michaelklishin:
@MarcialRosales note that we also support BERT files (which are significantly more space efficient), which I guess is why this option is necessary (we cannot force `.json`).

michaelklishin:
@MarcialRosales I have renamed the key a bit and fixed a small regression in the JS part.

michaelklishin:
FTR, the BERT part is not affected by these changes.

michaelklishin (approved):

PR #16958, opened 2026-07-16
Adopt `rabbit_re` in more places

michaelklishin:
For consistency, even if some of them do not
deal with external inputs or the inputs
would require very elevated privileges.

## Backporting

I suggest that we backport this after `4.3.3` ships.

### DIFF
--- a/deps/rabbit/src/rabbit_credential_validator_password_regexp.erl
+++ b/deps/rabbit/src/rabbit_credential_validator_password_regexp.erl
@@ -34,7 +34,7 @@ validate(Username, Password) ->
 -spec validate(rabbit_types:username(), rabbit_types:password(), string()) -> 'ok' | {'error', string(), [any()]}.
 
 validate(_Username, Password, Pattern) ->
-    case re:run(rabbit_data_coercion:to_list(Password), Pattern) of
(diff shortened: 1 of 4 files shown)


lhoguin (approved):

ansd (commented):
Minor feedback and not directly related to this PR: should `rabbit_re` better live in `rabbit` instead of `rabbit_common`? I added a while back https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit_common/AGENTS.md which explains the reason

michaelklishin:
@ansd I agree, there are no reasons to keep it in `rabbit_common` (the Erlang client does not use regular expressions that I know of).

But I'd move it to `rabbit` in another PR for ease of backporting to older series.

michaelklishin:
The OCI build fails due to an expired token.

michaelklishin:
If the plan is to move the module, let's ship `4.3.3`, then merge this PR, backport and shortly after submit and backport a PR that moves the module from `rabbit_common` to `rabbit`.

michaelklishin:
Merging and backporting this now, then will submit a follow-up that moves `rabbit_re` to `deps/rabbit`.

PR #16957, opened 2026-07-15
Restore the pre-login page after an idp_initiated OAuth login

thisisnsh:
## Background

In #9820 `oauth-return-to` was added for the management UI to remember the page a user was on before the OAuth login. This worked only for the `sp_initiated` flow and not the `idp_initiated` flow. In #9820 `startWithOAuthLogin()` stores the path in `oauth-return-to` but because the identity provider posts the token to `/login` that path is overwritten to '' and UI does not have a path to return to. But also there was no logic to return to `oauth-return-to` path for IDP. It was in `oauth_redirectToHome()` for SP Flow. 

## Proposed Changes

- Do not store `oauth-return-to` in `startWithOAuthLogin()` but instead do it in `oauth_initiateLogin()` this way the path is only stored when OAuth is initiated, and not when the we are returning back from OAuth. 
  - Also helps in the case when user opens `#/exchanges` and changes the url manually to `#/queues` and hit login, the return to path is correctly set to `#/queues` and not `#/exchanges` because we do not set `oauth-return-to` when the page loaded. 
- Store `oauth-idp-pending` in IDP flow in `oauth_initiateLogin()` to know if we are coming to `start_app()` after login. If we are, we can return to `oauth-return-to` path, or if not, just keep the existing `#/` path we have today
- Clean up `oauth-return-to` which stayed set and unused for IDP flow along with `oauth-idp-pending`

## Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

- [x] **Mandatory**: I have signed the CLA (see https://github.com/rabbitmq/cla)
- [x] I have read the `CONTRIBUTING.md` document
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

## Manual Testing 
| # | Scenario | Expected | Result |
|---|---|---|---|
| 1 | Normal login — open `/`, click log in | land on home `#/` | ✅ |
[...]

PR #16961, opened 2026-07-17
By @thisisnsh: Restore the pre-login page after an idp_initiated OAuth login

michaelklishin:
This is #16957 by @thisisnsh with a small usability fix from me.

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/main.js
+++ b/deps/rabbitmq_management/priv/www/js/main.js
@@ -15,11 +15,15 @@ function format_error_response(response, reason) {
     });
 }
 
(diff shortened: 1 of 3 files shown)


michaelklishin:
Note that we can only backport this after `4.3.3` ships.

michaelklishin:
The OCI failure has to do with an expired token.

michaelklishin:
@thisisnsh congratulations on your first contribution! Well done 👍👍

thisisnsh:
> @thisisnsh congratulations on your first contribution! Well done 👍👍

Thanks 😄

PR #16976, opened 2026-07-19
Fix stream SAC coordinator ignoring deactivating consumer status

### LINKED ISSUE #16975: Stream SAC group with stuck partitions
### Describe the bug

I've been looking into a bug where a stream SAC group will end up with some partitions that are in a stuck or stranded state and no messages flow to the consumers. Restarting the consumers doesn't help as the partitions usually remain stuck or the same bug is likely to be triggered again, and  `rabbitmq-streams activate_stream_consumer` doesn't help. The only dependable way to  fix a stuck partition is to identify the specific consumer that is supposedly active on the partition (according to `rabbitmq-streams list_stream_group_consumers`) and kill that connection, or just kill all the connections on the partition. 

The bug is frequently triggered when consumers are stopping and starting around rollout of a k8s deployment consisting of a few pods all participating in the same group and resuming work on the stream by querying rabbitmq for offsets when activated. Some useful mitigations I've found are carefully controlling the deployment so only one consumer app instance is starting at a time, and using direct rabbitmq node connections instead of connecting to a load-balanced service to shorten the duration of the critical race window, but even having both of those in place is not totally effective in preventing the bug. 

The root cause appears to be a kind of collision between an in-flight consumer step-down process and a re-selection of the same deactivating consumer as active. Adding a guard clause ([rabbit_stream_sac_coordinator.erl](https://github.com/pterygota/rabbitmq-server/commit/1890a6abd5bee0358d1d7ff6a65323f19be23219#diff-b3f656ceeb3db7280c6d6c64e9c038000b5b2263d5cb2d6b4281bfb01f5cf928)) so that deactivating consumers will hit the needed effect fixes it.

I have a branch with a fix and a test ready for a PR and a demo of the bug and the fix.

### Reproduction steps

**demo of the bug and proposed fix**
[rabbitmq-sac-bug-demo](https://github.com/pterygota/rabbitmq-sac-bug-demo) (requires docker and uv)

**rabbitmq-server fork with the fix and a test** 
Branch with test and fix: [stream-sac-stuck-partition](https://github.com/pterygota/rabbitmq-server/tree/stream-sac-stuck-partition)
Branch with just the test to demonstrate bug: [test-without-fix](https://github.com/pterygota/rabbitmq-server/tree/test-without-fix)
`make ct-rabbit_stream_sac_coordinator t="tests:super_stream_partition_sac_re_notifies_stepping_down_consumer_re_selected_as_active_test"`

### Expected behavior

Stream SAC group consumers should be able to start and stop without any careful coordination between group members, and no partitions should become permanently stuck.

### Additional context

_No response_

pterygota:
## Proposed Changes

Add a check for active status to the "same active consumer" guard in the command_activate_consumer handler in rabbit_stream_sac_coordinator.erl, so that notification is sent if a deactivating consumer is reselected for activation.

Fixes #16975 

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #16975)
- [ ] Build system and/or CI

## Checklist

## Further Comments

Impacts 4.1.2+, 4.2.x, 4.3.x per the issue, any backports would be appreciated if possible.
This PR and bug investigation were done with assistance from various "AI" or LLM tools such as Claude Code, ChatGPT, and Copilot.

### DIFF
--- a/deps/rabbit/src/rabbit_stream_sac_coordinator.erl
+++ b/deps/rabbit/src/rabbit_stream_sac_coordinator.erl
@@ -386,12 +386,16 @@ apply(#command_activate_consumer{vhost = VH, stream = S, consumer_name = Name},
                         %% do we need effects or not?
                         Effects =
                             case Csr of
-                                Csr when ?SAME_CSR(Csr, ActCsr) ->
+                                Csr when ?SAME_CSR(Csr, ActCsr) andalso
+                                         ActCsr#consumer.status =:= ?CONN_ACT ->
                                     %% it is the same active consumer as before
                                     %% no need to notify it
                                     [];
                                 _ ->
                                     %% new active consumer, need to notify it
+                                    %% (also fires when the previous "active" was
+                                    %% deactivating -- its client has been told to
+                                    %% step down and must be re-notified)
                                     [notify_csr_effect(Csr, S, Name, true)]
                             end,
                         {G2, Effects}
--- a/deps/rabbit/test/rabbit_stream_sac_coordinator_SUITE.erl
+++ b/deps/rabbit/test/rabbit_stream_sac_coordinator_SUITE.erl
@@ -226,6 +226,61 @@ super_stream_partition_sac_test(_) ->
 
     ok.
 
+%% Regression test for a stuck-group state. When the arrival of the stepping-down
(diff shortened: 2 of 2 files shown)


acogoluegnes (approved):

acogoluegnes:
Thanks!

PR #16985, opened 2026-07-20
Support claims containing dots in OAuth2 additional_scopes_key

sunfinite:
## Proposed Changes

Implements changes for #16947 

OAuth2 `additional_scopes_key` currently supports nested claims separated by dots. This prevents it from accessing claims whose names contain dots such as URIs.

This PR uses `cuttlefish_variable:tokenize/1` for the value, which treats `\`. as a literal dot. The path is now tokenized once at config parse time and not per-token. Keys set directly via `advanced.config` as binaries keep the current behavior.

```
auth_oauth2.additional_scopes_key = https://example\.com/claims.roles
```

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

**Breaking case**: Claim names that end with a `\` followed by nesting work today but will be treated as a single token after this change. This does seem unlikely to affect anyone in practice, and any affected users can get the old behavior by setting the key in `advanced.config` directly.

### DIFF
--- a/deps/rabbit/docs/rabbitmq.conf.example
+++ b/deps/rabbit/docs/rabbitmq.conf.example
@@ -1319,9 +1319,11 @@
 # auth_oauth2.preferred_username_claims.1 = user_name
 # auth_oauth2.preferred_username_claims.2 = preferred_username
 
-## Configure additional fields to look for scopes in the token
+## Configure additional fields to look for scopes in the token.
+## Dots separate nested claim keys; use \. for a literal dot in a claim name.
 ##
 # auth_oauth2.additional_scopes_key = extra_scope
+# auth_oauth2.additional_scopes_key = https://example\.com/claims.roles
 
 ## Custom scope prefix instead of resource_server_id
 ##
--- a/deps/rabbitmq_auth_backend_oauth2/README.md
+++ b/deps/rabbitmq_auth_backend_oauth2/README.md
@@ -163,7 +163,7 @@ Note: if both are configured, `jwks_uri` takes precedence over `signing_keys`.
 |------------------------------------------|---------------------------------------------------------------------------
 | `auth_oauth2.resource_server_id`         | [The Resource Server ID](#resource-server-id-and-scope-prefixes)
 | `auth_oauth2.resource_server_type`       | [The Resource Server Type](#rich-authorization-request)
-| `auth_oauth2.additional_scopes_key`      | Key to fetch additional scopes from (maps to `additional_rabbitmq_scopes` in the `advanced.config` format)
+| `auth_oauth2.additional_scopes_key`      | Key to fetch additional scopes from (maps to `additional_rabbitmq_scopes` in the `advanced.config` format). Dots separate nested claim keys; use `\.` for a literal dot in a claim name, e.g. `https://example\.com/claims.roles`
 | `auth_oauth2.default_key`                | ID (name) of the default signing key
 | `auth_oauth2.signing_keys`               | Paths to signing key files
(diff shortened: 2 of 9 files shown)


lukebakken (approved):
Reviewed the change to support dot-containing claims in `additional_scopes_key`. The design is sound: the `[[binary()]]` shape correctly encodes an outer list of space-separated paths, each pre-tokenized into segment binaries, which is necessary once `\.` escaping means dots can no longer be re-split at runtime. Backward compatibility is preserved via the untouched `is_binary` clause, and there are no consumers of the field outside the plugin.

I ran the three relevant suites locally: `rabbit_oauth2_schema_SUITE` (22 ok), `unit_SUITE` (45 ok), and `config_schema_SUITE` (1 ok), all passing with the new tokenizer and escaped-dot coverage exercised.

I also pushed a single follow-up commit, 370e08df53f8b3d5ceee0cb76c82c858f8331bc7, adding a comment that documents the expected `[[binary()]]` shape of the tokenized value on the list clause of `extract_scopes_from_additional_scopes_key/2`.

MarcialRosales (commented):
Thank you for the PR!

Could we preserve backward compatibility for the original binary format of `extra_scopes_source`?

Since some legacy or direct Erlang configurations might still define this as a flat binary (e.g., `{extra_scopes_source, <<"roles">>}`), it would be ideal if the validation code and the test suite supported both the original binary format and the new list-of-lists/array-of-arrays format.

I noticed `{extra_scopes_source, <<"roles">>}` had to be changed to `{extra_scopes_source, [[<<"roles">>]]}` for the existing tests to pass. Allowing the translation layer and the test cases to accept both the legacy binary and the nested list structure would be highly appreciated.

MarcialRosales:
@lukebakken Just to make sure I understand the trade-off here: are we leaning towards introducing a breaking change so that we can support dots in scope path names?

lukebakken:
@MarcialRosales no I'm waiting for @sunfinite to address your comment. I don't see any reason for a breaking change but then again OAuth2 is a big mystery to me anyway 😆

sunfinite:
Hi @MarcialRosales , 

> Could we preserve backward compatibility for the original binary format of `extra_scopes_source`?

There are existing tests ([1](https://github.com/sunfinite/rabbitmq-server/blob/dff194df0e/deps/rabbitmq_auth_backend_oauth2/test/unit_SUITE.erl#L1019), [2](https://github.com/sunfinite/rabbitmq-server/blob/dff194df0e/deps/rabbitmq_auth_backend_oauth2/test/unit_SUITE.erl#L1056)) that verify `extra_scopes_source` as a plain `binary()` from `advanced.config`. 

>  I noticed {extra_scopes_source, <<"roles">>} had to be changed to {extra_scopes_source, [[<<"roles">>]]} for the existing tests to pass. Allowing the translation layer and the test cases to accept both the legacy binary and the nested list structure would be highly appreciated.

The tests you called out validate the cuttlefish schema output, which is always the list form after this change. To confirm, are you suggesting the parser function (`tokenize_additional_scopes_key`) itself return `binary()` instead of `[[binary()]]` when there is no `\.`? We could add a presence check on the string to return early.

I have also pushed a new test for the new `[[binary()]]` format via `advanced.config`.

sunfinite:
> leaning towards introducing a breaking change

I tried searching for docs from well-known IdPs on this. No documented non-namespaced claims contain backslashes of the form:

```
{
    "my_scope_key\\": { "claim": "rabbitmq.tag:administrator" }
 }
```

This is the case that i can think of that would break when configured with `auth_oauth2.additional_scopes_key = my_scope_key\.claim`. These users would need to set `{extra_scopes_source, <<"my_scope_key\\.claim">>}` directly in `advanced.config` to preserve the old behavior.

sunfinite:
Hi @MarcialRosales, checking in to see if you have still have any concerns on the change.

PR #16994, opened 2026-07-21
Encode AMQP 1.0 arrays of zero-octet elements as opaque values

ansd:
## What?

The AMQP 1.0 binary parser (amqp10_binary_parser) decoded an array whose element constructor has a zero-octet wire encoding (null, true, false, uint0, ulong0, list0 - wire codes 0x40-0x45) by expanding it into a full Erlang list, one term per element, via lists:duplicate/2. It now instead represents such an array as an opaque value ({as_is, TypeCode, Bin}), the same representation already used for decimal and NaN/Infinity values that RabbitMQ has no need to interpret further.

 ## Why

For this class of array, the element count is carried in a small, fixed-size wire encoding, while the previous implementation allocated one term per element when decoding it. This meant the memory needed to hold the decoded value had no relationship to the size of the input that produced it. Representing the value opaquely removes that mismatch entirely, rather than merely capping it, and keeps the parser consistent with how it already treats other value kinds it does not need to interpret.

Supersedes https://github.com/rabbitmq/rabbitmq-server/pull/16986


lhoguin (approved):
If you don't need them then yes not decoding them is the way to go.

michaelklishin (approved):
This is an elegant solution that still addresses the issue at hand. I have tried with a 130M element modification across thousands of collections and it does work as expected.

Issue #10543, opened 2024-02-14
QQs: support if-empty and if-unused deletion parameters


Zerpet:
### Is your feature request related to a problem? Please describe.

Classic Queues and Streams support "fail-safe" parameters in queue deletion requests. Quorum Queues could benefit from the same parameters, with some limitations, given the nature of QQs. The intention is to prevent accidental queue deletions when a queue has data and/or consumers.

Observe the following snippet:

<details>
<summary>Delete queues using <code>rabbitmqctl delete_queue</code></summary>

```
docker exec -t bunny rabbitmqctl list_queues name type messages consumers
Timeout: 60.0 seconds ...
Listing queues for vhost / ...
name	type	messages	consumers
c1-v1	classic	0	0
st	stream	0	0
qq	quorum	0	0
cq-v2	classic	0	0
~ % docker exec -t bunny rabbitmqctl delete_queue c1-v1 --if-empty --if-unused
Deleting queue 'c1-v1' on vhost '/' if queue is empty and if queue is unused ...
Queue was successfully deleted with 0 ready messages
~ % docker exec -t bunny rabbitmqctl delete_queue cq-v2 --if-empty --if-unused
Deleting queue 'cq-v2' on vhost '/' if queue is empty and if queue is unused ...
Queue was successfully deleted with 0 ready messages
~ % docker exec -t bunny rabbitmqctl delete_queue qq --if-empty --if-unused
Deleting queue 'qq' on vhost '/' if queue is empty and if queue is unused ...
Error:
{:amqp_error, :not_implemented, ~c"cannot delete queue 'qq' in vhost '/'. queue.delete operations with if-unused flag set are not supported by quorum queues", :none}
~ % docker exec -t bunny rabbitmqctl delete_queue st --if-empty --if-unused
Deleting queue 'st' on vhost '/' if queue is empty and if queue is unused ...
Queue was successfully deleted with 0 ready messages
```
</details>

### Describe the solution you'd like

Add support for `if-empty` and `if-unused` in queue deletion requests. Given the distributed nature of Quorum Queues, our guarantees for `if-empty` would be as follows: _when a queue deletion request arrives, check if there are ready or pending messages. If there are non-zero ready or pending messages, return a failed queue delete response. If there are zero, proceed then with the delete request._

[...]

PR #16996, opened 2026-07-21
Support if-empty and if-unused checks in quorum queue deletion

Ayanda-D:
## Proposed Changes

This change implements: https://github.com/rabbitmq/rabbitmq-server/issues/10543 for quorum queues. A `rabbit_quorum_queue:stat/1`
call is carried out to get ready message count and consumer count, for checking the queue
`if-empty` and `if-unused`, respectively, before the deletion is carried out.

We are keen to have this feature supported in QQs (for both AMQP `queue.delete` and
operator tools QQ deletions / compatibility with CQ behaviour). 

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -1081,18 +1081,61 @@ restart_server({_, _} = Ref) ->
 -spec delete(amqqueue:amqqueue(),
              boolean(), boolean(),
              rabbit_types:username()) ->
-    {ok, QLen :: non_neg_integer()} |
+    rabbit_types:ok(non_neg_integer()) |
+    rabbit_types:error(in_use | not_empty) |
     {protocol_error, Type :: atom(), Reason :: string(), Args :: term()}.
 delete(Q, IfUnused, IfEmpty, ActingUser) when ?amqqueue_is_quorum(Q) ->
-    {ok, ReadyMsgs, Consumers} = stat(Q),
-    if
-        IfEmpty andalso ReadyMsgs > 0 ->
-            {error, not_empty};
-        IfUnused andalso Consumers > 0 ->
-            {error, in_use};
-        true ->
-            do_delete(Q, ReadyMsgs, ActingUser)
+    case delete_precondition(Q, IfUnused, IfEmpty) of
+        {ok, ReadyMsgs} ->
+            do_delete(Q, ReadyMsgs, ActingUser);
+        Error ->
+            Error
     end.
+
+%% The if-unused and if-empty flags gate deletion on the queue's consumer and
+%% message counts, read from the leader via rabbit_fifo_client:stat/1. When a
+%% flag is set, any failure to read the counts is surfaced rather than defaulted
+%% to zero: a queue whose counts cannot be verified must not be deleted, as that
+%% would silently bypass the requested precondition. When no flag is set the read
+%% is best-effort and only supplies the message count reported back to the client.
+delete_precondition(Q, false, false) ->
+    ReadyMsgs = case leader_stat(Q) of
+                    {ok, R, _} ->
+                        R;
+                    _ ->
+                        0
+                end,
+    {ok, ReadyMsgs};
+delete_precondition(Q, IfUnused, IfEmpty) ->
+    QName = amqqueue:get_name(Q),
+    case leader_stat(Q) of
+        {ok, ReadyMsgs, Consumers} ->
+            if
+                IfEmpty andalso ReadyMsgs > 0 ->
+                    {error, not_empty};
+                IfUnused andalso Consumers > 0 ->
+                    {error, in_use};
+                true ->
+                    {ok, ReadyMsgs}
+            end;
+        {_, Reason} ->
+            cannot_verify_delete_flags(QName, Reason)
+    end.
+
+%% Reads the queue's ready message and consumer counts from the leader via
+%% rabbit_fifo_client:stat/1, surfacing any failure to locate or query it.
+leader_stat(Q) ->
+    case find_leader(Q) of
+        {_, _} = Leader ->
+            rabbit_fifo_client:stat(Leader);
+        undefined ->
+            {error, no_leader_available}
+    end.
+
+cannot_verify_delete_flags(QName, Reason) ->
+    {protocol_error, internal_error,
+     "cannot delete ~ts: unable to verify the if-unused/if-empty precondition (~tp)",
+     [rabbit_misc:rs(QName), Reason]}.
 
 do_delete(Q, ReadyMsgs, ActingUser) ->
     {Name, _} = amqqueue:get_pid(Q),
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -1050,6 +1050,15 @@ recover(_Vhost, Queues) ->
          %% rabbit_durable_queue
          %% So many code paths are dependent on this.
          ok = rabbit_db_queue:set_dirty(Q),
+         %% A queue left in the blocked state means a delete with an
+         %% if-unused/if-empty precondition was interrupted before it could
+         %% restore the state. The queue still exists, so make it usable again.
+         _ = case amqqueue:get_state(Q) =:= blocked of
+                 true ->
+                     set_queue_state(QName, live);
+                 false ->
+                     ok
+             end,
          case Res of
              ok ->
                  {[Q | R0], F0};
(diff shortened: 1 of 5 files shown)


deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1083,22 +1083,23 @@ restart_server({_, _} = Ref) ->
...
-    {protocol_error, not_implemented,
-     "cannot delete ~ts. queue.delete operations with if-unused flag set are not supported by quorum queues",
-     [rabbit_misc:rs(amqqueue:get_name(Q))]};
-delete(Q, _IfUnused, true, _ActingUser) when ?amqqueue_is_quorum(Q) ->
-    {protocol_error, not_implemented,
-     "cannot delete ~ts. queue.delete operations with if-empty flag set are not supported by quorum queues",
-     [rabbit_misc:rs(amqqueue:get_name(Q))]};
-delete(Q, _IfUnused, _IfEmpty, ActingUser) when ?amqqueue_is_quorum(Q) ->
+delete(Q, IfUnused, IfEmpty, ActingUser) when ?amqqueue_is_quorum(Q) ->
+    {ok, ReadyMsgs, Consumers} = stat(Q),
```
the-mikedavis: I think `stat/1` is not suitable for this, at least how it is written currently: https://github.com/rabbitmq/rabbitmq-server/blob/83d29a2675ae1912c0cda29a06281e7b725ae8e1/deps/rabbit/src/rabbit_quorum_queue.erl#L1381-L1399

Any failure to stat retuns `{ok, 0, 0}`, so if-empty and if-unused checks fail open.

`stat/1` may not be suitable altogether too since it opens TOCTOU window: the queue might add consumers or messages immediately after the check and before the deletion. Classic queues handle this atomically by checking IfUnused and IfEmpty in the queue process. I'm not sure Ra has a way to do something similar today

Ayanda-D: thanks @the-mikedavis - hardcoded result `{ok, 0, 0}` is definitely not good and not what we want here (I simply assumed result is always from `rabbit_fifo_client:stat/1`, even better calling this directly). I agree any stat result must be sourced from the queue, and TOC/TOU window could yield unanticipated results. OK this needs a bit of re-think!

Ayanda-D: And the big issue is the limitation and impact on the protocol support - AMQP `queue.delete` (client applications upgrading from CMQs to QQs no longer able to carry out these safe queue deletes) - any ideas or suggestions on how to approach/solve this would be welcome.

cc: @kjnilsson @michaelklishin @lukebakken @acogoluegnes

Ayanda-D:
We are considering an approach (without altering the current `ra`) where if a AMQP `queue.delete` with `if-empty` and `if-unused` flags set, the queue state is temporarily updated to `blocked` - during which publish/consume operations cannot be carried out. Published messages are rejected if the queue state is **blocked** (via an `is_blocked/1` check), and `basic.consume` requests are also refused. 

**PUBLISH PATH**

image


Running quick perf-test (120s, 1KB payload)

**THROUGHPUT WITHOUT PATCH**

```
id: test-123004-371, sending rate avg: 25420 msg/s
id: test-123004-371, receiving rate avg: 25421 msg/s
id: test-123004-371, consumer latency min/median/75th/95th/99th 59556/447184/471552/514183/708190 µs
```


**THROUGHPUT WITH PATCH (is_blocked/1 check on publish path)**

```
id: test-123612-197, sending rate avg: 26757 msg/s
id: test-123612-197, receiving rate avg: 26757 msg/s
id: test-123612-197, consumer latency min/median/75th/95th/99th 87939/253808/399592/419802/440555 µs
```

**THROUGHPUT WITH PATCH + AMQP queue.delete / BLOCKING REQUEST**

```
id: test-135514-450, sending rate avg: 24018 msg/s
id: test-135514-450, receiving rate avg: 23713 msg/s
id: test-135514-450, consumer latency min/median/75th/95th/99th 34151/504861/527920/558731/588801 µs
```

- The `is_blocked/1` check on the publish path has very negligible impact on throughput and latency.

- Executing an AMQP `queue.delete` with `if-empty = true` and `if-unused =true` during the PERF-TEST, published messages in that short window are rejected as illustrated below:

image

Additionally:

- Alarmed nodes will reject AMQP `queue.delete` with `if-empty` and `if-unused` flags set.
- Queue state is restored after a `?BLOCKED_STATE_RESTORE_TIMEOUT` (e.g. restoring state from partitioned nodes, which were stuck in `blocked` state)
- Low impact - only the specific queue being deleted is "blocked" on the AMQP `queue.delete` with `if-empty` and `if-unused`

Bouncing off an idea to support this (with an implementation). Normal AMQP `queue.delete` remains as is. There may be more scenarios here to take into consideration. 

(OR, if feasible - if there is an approach to support this natively in the raft library? doesnt sound feasible in current `ra`).

zanarellidev:
One thing worth checking before this merges: `check_delete_flags/3` reads counts via `rabbit_fifo_client:stat(Leader)`, which is a plain local read against whatever the leader has applied so far, not a linearizable one. A publish that's committing to the log at the exact instant this read happens can land just after the read but still land, so `with_queue_blocked` closes the "new publish after the check" window but not the "publish already in flight when the check itself runs" one.

Swapping that read for `ra:consistent_query(Leader, {rabbit_fifo, overview, []}, Timeout)` guarantees it reflects everything already committed to a majority as of the moment it's issued, which is strictly stronger than `stat/1` here. It's still not airtight on its own since a publish landing in the gap between that read and the actual delete command could still slip through. I added a short settle-and-recheck (re-read once more after a bounded sleep, refuse if anything changed) on top of the consistent read for that reason, this catches the remaining window at the cost of a small, configurable delay on conditional deletes only.

Confirmed this actually reproduces with a deterministic test (`meck`-mocking `ra:consistent_query` to return an empty overview, then a populated one) rather than relying on real timing, since the natural race window is tiny (~1/15 in a stress test). Happy to send the diff (`rabbit_fifo_client:stat` to `ra:consistent_query` and the settle-recheck) as a suggestion here if useful, since #17075 is closing.

the-mikedavis:
I'm not keen on this 'blocked' approach since we are mixing Khepri's understanding of the world and the QQ's, and both are independent systems, possibly split across a different subset of nodes in a larger cluster. Especially with projections delivering eventually-consistent results, I'm worried about a consistency result that a black-box checker like Jepsen would see as misleading: queue is deleted if-empty, message is published and publish is confirmed, queue is deleted / message is lost. I think this needs to be done within the Ra server boundary (like CQs), and I believe that providing the way to do this is the raft library's responsibility, though it will definitely take some changes in Ra. I'm trying to pin down what the changes would look like and I'll comment here with all of the details once/if I can figure it out. I'm very happy to collaborate on a Ra PR to make this possible (provided I can figure out my opinion on how it should actually work! 😅). (The Ongaro models don't cover lifecycle around deletion, so I'm having a hard time being confident about modeling the deletion path here.)

The rough direction that I'm thinking of is to piggy-back off of the existing `delete_and_terminate` throw:

```erlang
apply_with({Idx, _, {'$ra_cluster', CmdMeta, delete, ReplyType}}, {...}) ->
    ...
    throw({delete_and_terminate, State, EOLEffects ++ NotEffs ++ Effects1});
```

Appending `{append, MyDeleteAndTerinateCmd}` is not safe as it opens another TOCTOU window.

Overall my opinion is that best-effort is not appropriate for this feature and it must be atomic in order to be correct. Otherwise the check could be done equally correctly by clients via a passive declare or API lookup.

deps/rabbit/src/rabbit_quorum_queue.erl
kjnilsson: I can't recall having a single ask for these feature over the last 8 years. That said, how accurate does it _need_ to be? if it needs to be 100% accurate we will need to make changes to the QQ state machine as well as Ra to reasonably support this, if even feasible.

the-mikedavis:
I opened up https://github.com/rabbitmq/ra/issues/648 in Ra for the way I'm thinking that this could fit into Ra. I think it's a small change but I might not have considered all of the edge-cases

Ayanda-D:
Just pushing the last bits of this proposed implementation. Few notes on some of the updates and questions raised:

- per queue blocked state for a conditional AMQP.delete (only applicable to the single queue in question)
- operates with `track_qq_members_uids` feature flag - tracked member UIDs used in block/unblock to ensure any new incarnation of queues are unaffected
- conditional deletes (and blocking/unblocking) are only executed by the leader node
- if an election is in progress it polls leadership for a bounded number of attempts until election completes and restores queue state 
- unlinked timer processes are spawned on all running cluster nodes to ensure state is restored, e.g. in the event of a partition or leader node going down mid-deletion
- refused under resource alarms - conditional deletes are rejected while any node is under a resource alarm
- we consider consistency guarantees from khepri to be sufficient. updates via `rabbit_amqqueue:update/2` → `rabbit_db_queue:update/3` use a CAS loop not naive read-then-write

```erlang
  update(QName, Fun, Options) ->
      Ret1 = rabbit_khepri:adv_get(Path, Options),          % local query — may be stale
      {ok, #{... data := Q, payload_version := Vsn}} = Ret1,
      Q1 = Fun(Q),
      case rabbit_khepri:put(UpdatePath, Q1, Options) of     % put carries #if_payload_version{version = Vsn}
          ok -> Q1;
          {error, {khepri, mismatching_node, _}} -> update(QName, Fun);  % retry on version conflict
          Err -> Err
      end.
```
- all khepri writes forwarded to and committed by the khepri leader and are CAS-protected.
- if `mismatching_node` is returned due to write version conflict, then worst case is a retry, but never a wrong write. 

- the  eventual-consistency concerns on reads (`is_blocked/1` via `rabbit_amqqueue:lookup/1` → `rabbit_db_queue:get1/1`, which are a raw local projection read) are inherent to a projection-based reads - this how rabbit already does same lookups on the hot-path anyway. so we dont forsee any TOCTOU issues here.

We believe this is sufficient for the general conditional AMQP.delete's, and making use of the new metadata reliable layer for control functions (temporary per queue state change for an odd AMQP conditional delete operation, + using qq members uids). Of course whichever approach is taken we can dig further for extreme edge cases. 

That said, unless the team reckons changing raft library to support the single op - `AMQP.delete`, thats ok - please feel free to drive the initiative forward. Feasibility still quite an open ended question on the `ra` implementation end (some edge cases which we still have questions with `ra` approach).

PR #17028, opened 2026-07-24
Reduce connection log noise from AMQP 0.9.1 and 1.0 clients, shovels

mkuratczyk:
Avoid logging stack traces when connections are terminated or cannot be established.
Before these changes, even a graceful rolling restart of a cluster running shovels would lead to numerous errors / stack traces due to maintenance mode connection termination, connection refused while a node is down, etc.


michaelklishin:
I will pick this work on while Michal is away.

PR #17037, opened 2026-07-27
Implement consumer timeout for classic queues

### LINKED ISSUE #16983: Implement consumer-timeout for classic queue type

v4.3 moved the consumer timeout ("message locks") feature from `rabbit_channel` (AMQP 0-9-1 only) to the queue type ([release notes](https://www.rabbitmq.com/blog/2026/04/23/rabbitmq-4.3-release#consumer-timeouts)). Quorum queues and the JMS queue type implement consumer timeouts but not classic queues or streams. Client applications might implicitly rely on the consumer timeout feature to take the message back from a completely stuck consumer, so that other consumers may handle it. Ideally classic queue consumers could get the same improved `basic.cancel` on a timeout if supported by the client.

We can look at implementing the consumer timeout concept for the classic queue type.

It would be hacky but we could add back the concept of consumer timeout to `rabbit_channel` so that stuck consumers see the same behavior as v4.2.x and below. That feels like going backwards though.

### Additional context

I don't think the same is necessary for stream queues because of the non-destructive consumption.

the-mikedavis:
Fixes #16983.

This is a two-part change to add consumer timeout to the classic queue type. #16983 has some details on the first commit. We need to refactor `rabbit_queue_consumers` so that we use a unique identifier per checkout attempt rather than the `SeqId` (or `{Priority, SeqId}`) of the message. With different interleavings of checkout / timeout / settle operations, passing the `SeqId` as the message ID could cause a settle to affect a checkout other than the one it was intended for. This change rewrites most of `rabbit_queue_consumers` to add an incrementing integer `ack_id()` which makes channel operations like settles unambiguous across even redeliveries of the same message.

With that refactor, implementing consumer timeout then becomes a fairly straightforward change to `rabbit_amqqueue_process` - parsing the requested timeout and setting timers accordingly. We don't need changes to `rabbit_queue_type` or protocol channel/session modules, we reuse the `released` notification used by QQs.

Opening this as a draft, I think there are some threads to pull to do less work/scanning/folding in the CQ process.


the-mikedavis:
\cc @LoisSotoLopez & @gomoripeti you might be interested in giving this a look (and probably spotting my arithmetic errors :sweat_smile:). I think it's in good shape now but it could use more eyes to spot any remaining bugs. Also cc @Payal2000, please give this a look if you're interested in consumer timeout for CQs.

lukebakken (approved):
> [!NOTE]
> This review was written by Claude (Anthropic's Claude Code) under the direction of @lrbakken, who reviewed it before posting.

Approving. Reviewed the classic queue consumer timeout implementation and the follow-up fixes: exclusive and single-active consumer status is now preserved across a timeout, the message count no longer double-counts tombstoned deliveries, and parked consumers are reported inactive. `xref`, `dialyze`, and the affected test suites pass. Two adversarial review passes surfaced no remaining defects.

deps/rabbit/src/rabbit_amqqueue_process.erl
```diff
@@ -1333,18 +1380,32 @@ handle_call({basic_get, ChPid, NoAck, LimiterPid}, _From,
...
+            reply({ok, BQ:len(BQS), Msg}, State3)
     end;
 
+%% Compat, can be removed when 'rabbitmq_4.4.0' FF is required.
 handle_call({basic_consume, NoAck, ChPid, LimiterPid, LimiterActive,
              Mode, ConsumerTag, ExclusiveConsume, Args, OkMsg, ActingUser},
+            From, State) ->
+    handle_call({basic_consume, NoAck, ChPid, LimiterPid, LimiterActive,
+                 Mode, ConsumerTag, ExclusiveConsume, Args, OkMsg, ActingUser,
+                 ?DEFAULT_CONSUMER_TIMEOUT},
```
LoisSotoLopez, line 1403: Couldn't we read the configured consumer timeout here? I understand during an upgrade, the queues still in 4.3 don't understand the message reporting the timeout, but we could have `rabbit_queue_type:get_consumer_timeout` exported and called from here.

deps/rabbit/src/rabbit_amqqueue_process.erl
```diff
@@ -1492,6 +1553,31 @@ new_single_active_consumer_after_basic_cancel(ChPid, ConsumerTag, CurrentSingleA
         _                    -> CurrentSingleActiveConsumer
     end.
 
+%% A consumer timeout parks the consumer rather than removing it: it is
+%% moved into blocked_consumers and resumes once it settles the released
```
LoisSotoLopez, line 1557: So, parked consumers can only be un-parked when they settle the deliveries, right? But     
  AFAIU when the consumer is parked the channel sends a basic.cancel to the client. Won't the consumer then never settle the deliveries? Doesn't a basic.cancel call for that, for the client to just re-subscribe and never settle previous deliveries? Without single active consumer that keeps the consumer forever parked out, right?

deps/rabbit/src/rabbit_amqqueue_process.erl
```diff
@@ -1715,6 +1801,52 @@ handle_info({drop_expired, Vsn}, State = #q{args_policy_version = Vsn}) ->
...
+            %% timeouts carry the ctag `none' and park no consumer, so
+            %% `get_blocked/3' returns `undefined' and they are skipped.
+            ParkedEntries =
+                [Entry || {ChPid, CTag, _AckIds, _RawTags} <- Expired,
+                          (Entry = rabbit_queue_consumers:get_blocked(
+                                     ChPid, CTag, Consumers1)) =/= undefined],
+            notify_consumer_transitions(ParkedEntries, blocked, State2),
+            notify_decorators(State2),
+            State3 = requeue_and_run(AllAckTags, false, [], State2),
+            noreply(ensure_consumer_timeout_timer(NextDeadline, State3))
```
LoisSotoLopez, line 1847: I'm not sure this `ensure...` is needed. Got to check later in the afternoon.

LoisSotoLopez (commented):
Most of my comments are questions, but would like to discuss.

deps/rabbit/src/rabbit_amqqueue_process.erl
LoisSotoLopez: Got confused because NextDeadline might be `infinity` when passed to `noreply(ensure_consumer_timoeut_timer(NextDeadline, State3))` in line 1487. If all acks expired, the NextDeadline will be infinity. The expired messages are delivered again in 1846 but 1847 still uses infinity.  Thankfully the noreply after that recomputes the timer.

Ah, but if noreply is recomputing the timer, why do we need the `ensure_consumer_timeout_timer` before calling `noreply`?

PR #17046, opened 2026-07-28
Federation: isolate upstream URI parse errors at link init

sauravonwww:
When a federation upstream URI fails to parse, `to_params/2` badmatched `{ok, Params}` and crashed link initialisation. The crash propagated through the link supervisor into broker recovery, so one malformed upstream took the whole broker down on every cold boot.

The trigger is c3b950b3 (atom-table exhaustion hardening), which now rejects unknown atom parameters (`auth_mechanism`, `verify`, ...) via `list_to_existing_atom/1`. That hardening is kept as is; the regression is that upstream definitions accepted by the older parser can now return `{error, _}` and turn a per-link problem into a broker-wide boot failure.

Make `to_params/2` return `{ok, _} | {error, {Info, SafeURI}}` with credentials stripped, and have `init_link/1` log and return `ignore` so one bad upstream cannot fail broker startup. Add unit tests covering the parse-error return and the credential-stripping guarantee.

## Proposed Changes

## Types of Changes

- [ x] Bug fix (non-breaking change which fixes issue #NNNN)
- [ ] Build system and/or CI

## Checklist

## Further Comments

### DIFF
--- a/deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
+++ b/deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
@@ -86,13 +86,27 @@ init_link({Upstream, XName}) ->
     case rabbit_exchange:lookup(XName) of
         {ok, X} ->
             DeobfuscatedUpstream = rabbit_federation_util:deobfuscate_upstream(Upstream),
-            DeobfuscatedUParams = rabbit_federation_upstream:to_params(DeobfuscatedUpstream, X),
-            UParams = rabbit_federation_util:obfuscate_upstream_params(DeobfuscatedUParams),
-            rabbit_federation_status:report(Upstream, UParams, XName, starting),
-            join(rabbit_federation_exchanges),
-            join({rabbit_federation_exchange, XName}),
-            gen_server2:cast(self(), maybe_go),
-            {ok, {not_started, {Upstream, UParams, XName}}};
+            case rabbit_federation_upstream:to_params(DeobfuscatedUpstream, X) of
+                {ok, DeobfuscatedUParams} ->
+                    UParams = rabbit_federation_util:obfuscate_upstream_params(DeobfuscatedUParams),
+                    rabbit_federation_status:report(Upstream, UParams, XName, starting),
+                    join(rabbit_federation_exchanges),
+                    join({rabbit_federation_exchange, XName}),
+                    gen_server2:cast(self(), maybe_go),
+                    {ok, {not_started, {Upstream, UParams, XName}}};
+                {error, {_Info, SafeURI} = Reason} ->
+                    %% Unable to parse upstream URI. Return `ignore` so this
+                    %% link's supervisor keeps its siblings and does not fail
+                    %% broker startup. Publish an entry to the status table
+                    %% so the link is visible via `federation_status`.
+                    ?LOG_ERROR(
+                       "Skipping federation link for ~ts, upstream ~ts: ~tp",
+                       [rabbit_misc:rs(XName), Upstream#upstream.name, Reason]),
+                    rabbit_federation_status:report(
+                      Upstream, #upstream_params{safe_uri = SafeURI}, XName,
+                      {invalid_upstream_uri, Reason}),
+                    ignore
+            end;
         {error, not_found} ->
             ?LOG_WARNING("not found, stopping link", []),
             {stop, gone}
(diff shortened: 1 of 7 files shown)


deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
```diff
@@ -86,13 +86,23 @@ init_link({Upstream, XName}) ->
...
+                    gen_server2:cast(self(), maybe_go),
+                    {ok, {not_started, {Upstream, UParams, XName}}};
+                {error, Reason} ->
+                    %% Unable to parse upstream URI. Return `ignore` so this
+                    %% link's supervisor keeps its siblings and does not fail
+                    %% broker startup.
+                    ?LOG_ERROR(
+                       "Skipping federation link for ~ts, upstream ~ts: ~tp",
+                       [rabbit_misc:rs(XName), Upstream#upstream.name, Reason]),
+                    ignore
```
the-mikedavis, line 107: One possible gap here is that we don't report on these links via `rabbit_federation_status`, so the CLI wouldn't show them for example. I think we want to add a line here like

```erlang
%% ... error log ...
rabbit_federation_status:report(
  Upstream, #upstream_params{safe_uri = SafeURI}, XName,
  {invalid_upstream_uri, Reason}),
ignore
```

deps/rabbitmq_federation_common/test/unit_SUITE.erl
```diff
@@ -174,3 +176,36 @@ await_all_dead(Pids, AttemptsLeft) ->
...
+%% rabbit_federation_upstream:to_params/2 error handling
+%% -------------------------------------------------------------------
+
+%% Regression test for the boot-time crash observed when a persisted
+%% federation-upstream URI contains an `auth_mechanism` value that is not an
+%% existing atom (for example, an invisible U+2028 injected by a copy-paste).
+%% Before the fix, `amqp_uri:parse/2` returning `{error, _}` would badmatch
+%% inside `to_params/2` and crash the link's supervisor. It must now return
+%% an error tuple that callers can handle.
+to_params_returns_error_on_unparseable_uri(_Config) ->
```
the-mikedavis, line 193: For context around this change: before the changes in https://github.com/rabbitmq/rabbitmq-server/pull/16063, a malformed mechanism could pass validation. The federation link would fail to connect in the login step, but if the server was upgraded while that federation config existed, boot would fail during exchange recovery because validation is stricter now.

the-mikedavis: I think the reproduction flow would work like...

1. On v4.2.5, `gmake ENABLED_PLUGINS="rabbitmq_management rabbitmq_federation" run-background-broker`
2. Declare a durable exchange:
   ```
   curl -u guest:guest -X PUT http://localhost:15672/api/exchanges/%2F/fed.test \
     -H "content-type: application/json" \
     -d '{"type":"direct","durable":true,"auto_delete":false,"internal":false,"arguments":{}}'
   ```
3. Setup a federation parameter with a bad mechanism
   ```
   curl -u guest:guest -X PUT http://localhost:15672/api/parameters/federation-upstream/%2F/up1 \
     -H "content-type: application/json" \
     -d '{"value":{"uri":"amqp://guest:guest@localhost:5672?auth_mechanism=totally-bogus"}}'
   curl -u guest:guest -X PUT http://localhost:15672/api/policies/%2F/fed-policy \
     -H "content-type: application/json" \
     -d '{"pattern":"^fed\\.test$","apply-to":"exchanges","definition":{"federation-upstream":"up1"}}'
   ```
4. Stop the broker. Switch to v4.2.6+ or v4.3.0+ tags and run-broker again.

I think this flow should work but I haven't tried it myself yet.

deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl
sauravonwww: good catch, updated!

lukebakken: Good call — this is now handled: both the exchange and queue link error branches report to `rabbit_federation_status` with `{invalid_upstream_uri, Reason}` before returning `ignore`, so the link shows up via the CLI.

deps/rabbitmq_federation_common/test/unit_SUITE.erl
lukebakken: Thanks for the context — that upgrade-during-recovery path (stricter `auth_mechanism` validation after #16063) is exactly what `to_params_returns_error_on_unparseable_uri` pins down.

lukebakken (approved):
Verified: malformed and multi-URI upstreams no longer crash link init, credentials are stripped from both the sanitized URI and the parse-error `Info`, and skipped links are reported via `rabbit_federation_status`. Federation link + unit suites, dialyzer, and xref all pass. LGTM.

michaelklishin (approved):
Thank you for pointing out that this is a regression from prior security work.

PR #17049, opened 2026-07-29
Make AMQP message parsing stricter

ansd:
This PR introduces improvements to the AMQP message parsing layer ensuring stricter adherence to the protocol specification. The parser now explicitly rejects and throws errors on malformed AMQP encodings such as wrong order or cardinality of message sections, ensuring that invalid payloads fail fast.

The performance regression is negligible in the default case where the message contains a single data section. Validating the message is O(N) where N is the number of sections. No sub binaries or Erlang terms are constructed however during validation. We cap the maximum number of sections within an AMQP message to 10k. This should be sufficient for any real world use case.


michaelklishin (approved):

michaelklishin:
I have added a couple of property-based tests.

PR #17053, opened 2026-07-30
Stream: clamp frame_max before authentication and authorization

acogoluegnes:
rabbit_stream_reader seeded the wire parser with the configured frame_max (default 1 MiB, operators can set it higher) from the first byte on the socket. That ceiling governed parsing of peer_properties, sasl_handshake and every sasl_authenticate round-trip, well before the connection was authenticated and authorized.

Authentication actually completes before frame-size tuning: sasl_authenticate succeeds, the server sends tune, the client responds tune, then sends open. open is where vhost-alive, vhost/user/node connection limits and vhost access are checked, so a connection should only be trusted with the full negotiated frame size once open succeeds, not merely after sasl_authenticate.

Clamp the parser to 8192 bytes (or the configured frame_max, if stricter) until open succeeds, then switch to the value negotiated at tune. 8192 comfortably covers realistic SASL PLAIN payloads carrying OAuth 2.0 JWTs with several vhost-scoped claims, while keeping the pre-open buffer well below the default frame_max.

### DIFF
--- a/deps/rabbitmq_stream/docs/PROTOCOL.adoc
+++ b/deps/rabbitmq_stream/docs/PROTOCOL.adoc
@@ -880,3 +880,8 @@ The client answers with a `Tune` frame with the settings he agrees on, possibly 
 from the server's suggestions.
 * Open: the client sends an `Open` frame to pick a virtual host to connect to. The server
 answers whether it accepts the access or not.
+
+Until `Open` completes successfully, the server enforces a low, fixed `frame_max` ceiling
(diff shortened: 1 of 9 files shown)


michaelklishin:
@acogoluegnes I have made the pre-authN limit configurable like we did for AMQP 0-9-1. The default stays the same.

michaelklishin:
I will submit a follow-up PR to match the AMQP 0-9-1 Cuttlefish schema data type so that information unit values like `8K` are accepted.

PR #17055, opened 2026-07-30
Support time-based delete in AMQP 0-9-1 shovel

michalovits:
## Proposed Changes

This PR introduces `(src-)delete-after-duration` as an AMQP 0-9-1 shovel parameter, to allow a shovel's lifetime to be bound by time alongside the existing target message count (`src-delete-after`). Our goal is to handle time-bound lifetimes robustly within the plugin. A few properties fall out of the implementation here that are difficult to orchestrate via the HTTP API:
- The _knowledge_ of how long the shovel should live is replicated
- The _ability to delete_ the shovel can fail over to any node
- The configured runtime is part of the shovel definition and thus visible in the shovel list

`src-delete-after-duration` is a per-shovel parameter. To avoid churn, we add `delete_after_duration_floor` (60s by default) in `rabbitmq.conf`. Any shovel below this value is adjusted to meet the floor, with a hard minimum set at 15s.

Similar to the accounting for `src-delete-after`, the TTL resets on shovel restart. This behaviour seems interpretable and consistent with the existing feature.

Overall, we understand this feature to be in line with the plugin's [stated philosophy](https://www.rabbitmq.com/docs/shovel#benefits): "a minimalistic yet flexible tool in the distributed messaging toolkit that can accommodate a number of use cases," and we can reuse the patterns of `src-delete-after`.

Discussed in https://github.com/rabbitmq/rabbitmq-server/discussions/17016. Happy to follow up with documentation/release notes once this PR is concluded.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

One consideration was whether to produce a validation error for TTLs below the `delete_after_duration_floor`. Considering the fact that `delete_after_duration_floor` depends on the worker node's configuration, and also aiming for well-behaved static shovels, it seems safest for the TTL to simply be clamped to the floor.

We also spent some time on whether the TTL should be expressed as an absolute-time deadline instead of in relative terms. Overall, we think that the relative TTL is consistent with other RabbitMQ features, is interpretable and simple, and avoids introducing concerns about e.g. the viability of static shovels. It also means a shovel can structurally never be started past its deadline.

The intention is that TTL is set on one or a few shovels at a time to perform a time-bound job, hence why this is an opt-in, per-shovel parameter.

On whether this can lead to mass autodeletes -- currently, shovels reconnect and redeclare topology upon restart. With the change, they might also set an expiry deadline; the TTL begins counting down from `init_source/1`. In a usage pattern where very many shovels are configured with the same TTL, a node restart could line up their deadlines to also produce correlated expiries (graceful terminations, at minimum 15s later, per the hard limit).

Out of caution, we reproduced this scenario locally on a 3-node cluster with 1000 shovels. As @mklishin indicated in the preceding discussion, the shovels all started on the same node. We then restarted that node to line up the deadlines, causing all 1k shovels to autodelete at the same time. In a background thread, we polled `ra:key_metrics({rabbit_khepri:get_ra_cluster_name(), node()})` on each node and correlated 1k applied commits with the broker log lines _"Shovel [...] stopping (it was configured to autodelete and transfer is completed)"_. We observed no `commit_latency` spike above 2ms. The broker cluster and plugin also remained in a healthy state. We can share the logs and run more extreme scenarios as needed.


michaelklishin:
Rebased and moved to https://github.com/rabbitmq/rabbitmq-server/pull/17064 which includes a tiny amendment.

Thank you, @michalovits!

PR #17075, opened 2026-07-31
Support if-empty/if-unused safe-deletion flags for quorum queues

### LINKED ISSUE #10543: QQs: support if-empty and if-unused deletion parameters

zanarellidev:
## Summary

Fixes #10543.

Classic queues have long supported `queue.delete`'s `if-unused` and `if-empty` flags to avoid accidentally deleting a queue that still has data or consumers. Quorum queues instead unconditionally rejected any delete that set either flag with a `not_implemented` protocol error — the `TODO Quorum queue needs to support consumer tracking for IfUnused` comment was still sitting right above the rejection.

This implements the semantics requested in #10543: check the queue's current message/consumer count (via `ra:member_overview/2` against the leader) before deleting, and refuse with `{error, not_empty}` / `{error, in_use}` if the corresponding flag's condition isn't met, reusing the existing `ra:member_overview/2` primitive rather than adding new tracking machinery.

As documented inline, this is necessarily best-effort for quorum queues: unlike classic queues, the check and the delete are not one atomic operation across the cluster, so this prevents *accidental* deletion rather than offering atomic check-then-delete semantics under concurrent publishes/deletes. When the leader is unreachable, `if-empty` fails safe (refuses deletion, since message count can't be verified) while `if-unused` alone still succeeds (a queue with no reachable leader can't have live consumers right now) — mirroring the existing down-queue-process behavior already used for classic queues in this same module.

## Test plan

- [x] Rewrote the two `quorum_queue_SUITE` test cases that previously asserted the `not_implemented` rejection to instead exercise the real semantics (delete succeeds when empty/unused, refused otherwise).
- [x] Full `quorum_queue_SUITE` (271 cases) passes clean.
- [x] `dialyzer` passes clean.

### DIFF
--- a/deps/rabbit/src/rabbit_quorum_queue.erl
+++ b/deps/rabbit/src/rabbit_quorum_queue.erl
@@ -37,6 +37,7 @@
 -export([update_consumer_handler/8, update_consumer/9]).
 -export([cancel_consumer_handler/2, cancel_consumer/3]).
 -export([become_leader/2, handle_tick/3, spawn_deleter/1]).
+-export([state_restore_timer/3, spawn_state_restore_timers/2]).
 -export([rpc_delete_metrics/1,
          key_metrics_rpc/1]).
 -export([format/2]).
@@ -173,6 +174,9 @@
 -define(START_CLUSTER_RPC_TIMEOUT, 60_000). %% needs to be longer than START_CLUSTER_TIMEOUT
 -define(TICK_INTERVAL, 5000). %% the ra server tick time
 -define(DELETE_TIMEOUT, 5000).
+%% Must exceed ?DELETE_TIMEOUT so a normally-running conditional delete
+%% finishes before the safety-net timer restores a blocked queue.
+-define(BLOCKED_STATE_RESTORE_TIMEOUT, ?DELETE_TIMEOUT + 1_000).
 -define(MEMBER_CHANGE_TIMEOUT, 20_000).
 -define(SNAPSHOT_INTERVAL, 8192). %% the ra default is 4096
 %% setting a low default here to allow quorum queues to better chose themselves
@@ -1064,6 +1068,12 @@ recover(_Vhost, Queues) ->
          %% rabbit_durable_queue
          %% So many code paths are dependent on this.
          ok = rabbit_db_queue:set_dirty(Q),
+         %% A queue left blocked means a conditional delete was interrupted
+         %% before restoring state; make it usable again on recovery.
+         _ = case amqqueue:get_state(Q) =:= blocked of
+                 true -> set_queue_state(QName, live);
+                 false -> ok
+             end,
          case Res of
              ok ->
                  {[Q | R0], F0};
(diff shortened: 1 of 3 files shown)


kjnilsson (approved):

deps/rabbit/src/rabbit_quorum_queue.erl
```diff
@@ -1096,20 +1096,67 @@ restart_server({_, _} = Ref) ->
...
+    {error, in_use};
+check_delete_preconditions1(_IfUnused, _IfEmpty, _Consumers, _Msgs) ->
+    ok.
+
+-spec query_overview(amqqueue:amqqueue()) ->
+    {ok, Msgs :: non_neg_integer(), Consumers :: non_neg_integer()} | error.
+query_overview(Q) ->
+    case find_leader(Q) of
+        {_, _} = Leader ->
+            case ra:member_overview(Leader, ?DELETE_TIMEOUT) of
```
Ayanda-D: @zanarellidev thanks (we are also very keen [on having this fix](https://github.com/rabbitmq/rabbitmq-server/pull/16996)) - there was a question https://github.com/rabbitmq/rabbitmq-server/pull/16996#discussion_r3625237316 - when concurrent channels are publishing to the same queue, are we guaranteed that the `ra:member_overview/2` reply will be received and following deletion processed, by its calling channel first, before the publishing channel receives its in-flight response to `rabbit_queue_type:deliver/4`. we are concerned a concurrent publish(es) may be confirmed by its channel first, while the deletion is underway on a different channel - which could lead to lost messages. (unless we are overthinking here and this PR solves this, happy to close https://github.com/rabbitmq/rabbitmq-server/pull/16996 in favour of this one)

zanarellidev:
Thanks for the pointer to #16996, and I should have credited it explicitly from the start: the blocked-state mechanism this PR uses (`with_queue_blocked/2`, the safety-net restore timers) is your pattern from there, adapted here.

To your question directly: no, there is no full ordering guarantee at the Raft log level between the delete's precondition check and a concurrent publish's confirm. What this PR actually closes:

1. The blocked state (from #16996) rejects any publish arriving after the block is set, via `deliver/0`. That leaves only the time before blocking, plus whatever was already in flight to the leader at that instant.
2. For that narrower window, `check_delete_preconditions/3` now does a linearizable `ra:consistent_query/3` read, then re-reads once more after a short settle delay. A publish that commits inside that window now causes a safe refusal instead of a silent delete.

I want to be precise about the evidence rather than overstate it. The concurrent-publish stress test (`delete_if_empty_while_publishing`, 5000 publishes) only lands inside the settle window in a small fraction of runs, since it depends on real publish/commit timing, so it isn't a reliable proof on its own. I added `delete_if_empty_race_settle_recheck`, which mocks the two `ra:consistent_query/3` reads to force the exact sequence the re-check exists to catch. It fails against the code before the settle-recheck and passes with it.

The one race this still doesn't close: a publish committing between the second read and the delete command itself reaching the log. Fully closing that would need the check-and-delete to be a single command applied by the `rabbit_fifo` state machine, which felt like a bigger change than this PR's scope. Happy to follow up on that separately if there's interest.

Given kjnilsson's approval, would you be comfortable closing #16996 in favor of this one, or is there anything else you'd want addressed first?

Ayanda-D:
@zanarellidev looks like your commit https://github.com/rabbitmq/rabbitmq-server/pull/17075/commits/16df8926523344c298cdfab810d08a0face4bd21 is just an exact copy of the work done in https://github.com/rabbitmq/rabbitmq-server/pull/16996 which has been awaiting review for about a week - unfortunately we cannot close https://github.com/rabbitmq/rabbitmq-server/pull/16996 at this point (of course its at the maintainers discretion, @michaelklishin & co, to keep PRs open or to close them).

As we all do (in the good spirit of open-source collaboration) you could give us your feedback and input in the first PR https://github.com/rabbitmq/rabbitmq-server/pull/16996 which we opened last week and are currently working on.

zanarellidev:
You're right, and I should have caught this myself before opening this PR: #16996 was open ten days before this one, and the blocking mechanism here ended up mirroring yours closely once I adapted it after your earlier comment. That's not the right way to collaborate, closing this in favor of #16996.

The one piece of independent value here is the race-condition analysis on the precondition check itself (a plain `rabbit_fifo_client:stat/1` read isn't linearizable, so a publish committing at the same instant can be missed). I'll bring that over to #16996 as a review comment, with the fix and a deterministic test proving it, since #16996 has the same read pattern today.

PR #17103, opened 2026-08-04
Stream: add max_uncompressed_sub_entry_batch_size setting

acogoluegnes:
The broker stores a published sub-entry batch as-is, including its declared uncompressed size, without decompressing or verifying it: doing so would defeat the point of end-to-end sub-entry compression.

Add a per-node setting, max_uncompressed_sub_entry_batch_size, that bounds this declared size at publish time. It defaults to 67108864 (64 MiB), matching the default of the same setting on the Java client (maxUncompressedSubEntryBatchSize). Configure it via stream.max_uncompressed_sub_entry_batch_size in rabbitmq.conf; the broker and the clients publishing to it should use the same value, since the broker's setting governs what can be written and a client's setting governs what it will read back.

A publish frame with a sub-entry batch over the limit gets a publish_error with the existing precondition_failed response code instead of being stored. If one sub-entry in a frame is rejected, the rest of that frame is rejected too, so a publisher never sees a gap in its publishing ID sequence.

Also fixes a bug found while making this change: a publish_v2 frame with no filter value, addressed to an unknown publisher ID, crashed the reader instead of returning a publish_error.

### DIFF
--- a/deps/rabbitmq_stream/Makefile
+++ b/deps/rabbitmq_stream/Makefile
@@ -14,6 +14,7 @@ define PROJECT_ENV
 	{initial_credits, 50000},
 	{credits_required_for_unblocking, 12500},
 	{frame_max, 1048576},
+	{max_uncompressed_sub_entry_batch_size, 67108864},
 	{heartbeat, 60},
 	{advertised_host, undefined},
 	{advertised_port, undefined}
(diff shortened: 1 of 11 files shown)


michaelklishin:
@acogoluegnes I have added a very small improvement: instead of relying on a [client-provided value](https://github.com/rabbitmq/rabbitmq-server/blob/57b1f0c1dece56c4969fdb98498132a21702f623/deps/rabbitmq_stream_common/src/rabbit_stream_core.erl#L734-L740), we can compute the number of messages actually handled.

Let me know if there are reasons to prefer the original version.

michaelklishin:
Never mind, I have found a reason to revert my change: it kills the TCO (Tail Calling Optimization) and the "do not depend on client input" improvement is not significant enough to justify that.

PR #17120, opened 2026-08-06
Management, OAuth 2: honor X-Forwarded-* protocol and port when building the proxy URL

### LINKED ISSUE #17110: oauth2 management UI: token_endpoint proxy URL uses http:// scheme even when accessed over https, causing mixed-content block
### Describe the bug

When `management.oauth_initiated_logon_type = sp_initiated` is used with
a resource server that has `use_token_endpoint_proxy` enabled internally,
the discovery document RabbitMQ serves at
`/js/oidc-oauth/token-endpoint/<id>/openid-configuration` renders the
`token_endpoint` field with a plain `http://` scheme, even when RabbitMQ
is only ever reached over https (TLS terminated at a reverse proxy in
front of it). Every other field in the same document
(authorization_endpoint, userinfo_endpoint, jwks_uri — all pointing at
the actual IdP) correctly uses https; only this one self-referencing
field does not.

The browser then refuses to fetch() an http:// URL from a page loaded
over https ("mixed content" blocking), so the OAuth token-exchange step
silently fails right after an otherwise successful identity-provider
login, with no trace in RabbitMQ's or the IdP's server logs at all — the
browser blocks the request before any network traffic is sent.

Confirmed the scheme is not derived from `X-Forwarded-Proto`: port-forwarded
directly to the pod (bypassing the reverse proxy entirely) and compared
requests with and without `X-Forwarded-Proto: https` set manually — the
response is byte-for-byte identical either way. Also found no cuttlefish
config mapping for `use_token_endpoint_proxy` itself in
`rabbitmq_management.schema` — it appears to be computed internally with
no way to override it via config.

### Reproduction steps

1. Deploy RabbitMQ with the management oauth2 plugin enabled:
management.oauth_enabled = true
management.oauth_initiated_logon_type = sp_initiated
management.oauth_client_id = <client>
auth_backends.1 = rabbit_auth_backend_oauth2
auth_oauth2.resource_server_id = <resource_server>
auth_oauth2.issuer = https://<keycloak-host>/auth/realms/<realm>

2. Put RabbitMQ behind any reverse proxy / ingress that terminates TLS
before traffic reaches the RabbitMQ pod/node (ingress-nginx, Traefik,
a cloud load balancer doing TLS termination, etc.) — RabbitMQ itself
only ever receives plain HTTP.
3. Fetch the discovery document for the resource server:
curl https://<rabbitmq-host>/js/oidc-oauth/token-endpoint/<resource-server-id>/openid-configuration

Inspect the `token_endpoint` field in the response.
4. Alternatively, reproduce without any ingress at all: `kubectl
port-forward` directly to the RabbitMQ management port and curl the
same path with and without an `X-Forwarded-Proto: https` header set
manually — the response is identical either way, showing the header
is not consulted.
5. Open the RabbitMQ management UI in a browser at `https://<rabbitmq-host>/`,
click Login, complete the identity provider's login form successfully,
get redirected back to RabbitMQ.
6. Observe the browser console: right after the redirect, a `fetch()` to
the URL from step 3's `token_endpoint` field fails with
`TypeError: Failed to fetch`, and the page URL gets
`?error=TypeError:%20Failed%20to%20fetch` appended.

### Expected behavior

The self-referencing `token_endpoint` URL (and any other URL RabbitMQ
builds pointing back at itself for this proxy feature) should reflect
the scheme the client actually connected with — either by respecting
`X-Forwarded-Proto` (as is common for reverse-proxied web apps), or via
an explicit config option to declare the externally-visible
scheme/host — so `sp_initiated` OAuth login works for the common case of
RabbitMQ deployed behind a TLS-terminating ingress/load balancer.

### Additional context

- RabbitMQ version: 4.2.9 (rabbitmq_management, rabbitmq_auth_backend_oauth2)
- Deployment: Kubernetes (AKS), ingress-nginx behind a Traefik Gateway,
TLS terminated at Traefik
- Related: https://github.com/rabbitmq/rabbitmq-management/issues/560 —
same general class of bug (mgmt UI ignoring `X-Forwarded-Proto` behind
a TLS-terminating proxy) reported years ago in the old, now-archived
`rabbitmq-management` repo.

michaelklishin:
See #17110 for details.

Closes #17110.

### DIFF
--- a/deps/rabbitmq_management/src/rabbit_mgmt_oauth_token_proxy.erl
+++ b/deps/rabbitmq_management/src/rabbit_mgmt_oauth_token_proxy.erl
@@ -9,7 +9,8 @@
 
 -export([init/2]).
 %% exported for testing
--export([inject_client_secret/2, rewrite_token_endpoint/2]).
+-export([inject_client_secret/2, rewrite_token_endpoint/2,
+         external_authority/5]).
 
 -include_lib("oauth2_client/include/oauth2_client.hrl").
 -include_lib("kernel/include/logger.hrl").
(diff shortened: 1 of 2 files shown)


michaelklishin:
@MarcialRosales let me know what you think about the approach. I think it's fairly standard for infrastructure tools and data services.

michaelklishin:
Closing in favor of https://github.com/rabbitmq/rabbitmq-server/pull/17153.

PR #17128, opened 2026-08-06
QQ: fall back to prior machine version for local queries in mixed clusters

shashank-singh-2999:
## Proposed Changes

Same underlying bug pattern as a mixed-cluster test failure seen between 4.0.x and 3.13.7 (`amqp_client_SUITE > cluster_size_3 > last_queue_confirms`): a local query hardcodes the latest `rabbit_fifo` module and crashes against an older state shape still held by a old cluster member. Confirmed this pattern concretely at the 4.2.x/4.3.x boundary instead (`rabbit_fifo` machine version 8 shipped in 4.3.0): during a rolling upgrade, a queue whose Ra group still includes a pre-4.3 member stays capped at the v7 state shape, and `query_notify_decorators_info` / `query_consumers` crash with `function_clause` against it, silently turning into skipped consumer notifications or an empty consumer list

Generalizes the fallback `query_single_active_consumer/1` already had into a shared `rabbit_fifo_client:local_query/2,3`, routes the other affected call sites through it, and removes `rabbit_amqqueue: consumers/1`'s own ad hoc version of the same fix. `stat/2` also drops `ra:member_overview` in favor of the same lightweight helper.

Added `quorum_queue_SUITE:mixed_version_local_queries` (skips outside mixed-version runs) plus `rabbit_fifo_int_SUITE` unit tests.

### DIFF
--- a/deps/rabbit/src/rabbit_amqqueue.erl
+++ b/deps/rabbit/src/rabbit_amqqueue.erl
@@ -1619,19 +1619,9 @@ consumers(Q) when ?amqqueue_is_classic(Q) ->
     end;
 consumers(Q) when ?amqqueue_is_quorum(Q) ->
     QPid = amqqueue:get_pid(Q),
-    case ra:local_query(QPid, fun rabbit_fifo:query_consumers/1) of
+    case rabbit_fifo_client:local_query(QPid, query_consumers) of
         {ok, {_, Result}, _} ->
             maps:values(Result);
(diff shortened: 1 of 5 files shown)


michaelklishin (approved):

michaelklishin:
I don't have an opinion on whether this is the optimal approach (optimal refactoring) but it does address the stated problem, and it is an upgrade safety improvement.

PR #17129, opened 2026-08-06
Add OSS changes for JMS Topic support

ansd:
This commit adds the OSS changes for supporting JMS Topics over AMQ using the AMQP JMS Mapping spec.

Test cases can be found in tanzu-rabbitmq PR 630.

### DIFF
--- a/deps/rabbit/src/rabbit_amqp_session.erl
+++ b/deps/rabbit/src/rabbit_amqp_session.erl
@@ -89,11 +89,15 @@
 -define(PROTOCOL, amqp10).
 -define(MAX_PERMISSION_CACHE_SIZE, 12).
 -define(HIBERNATE_AFTER, 6_000).
-%% Capability defined in amqp-bindmap-jms-v1.0-wd10 [5.2] and sent by Qpid JMS client.
--define(CAP_TEMPORARY_QUEUE, <<"temporary-queue">>).
 -define(CAP_VOLATILE_QUEUE, <<"rabbitmq:volatile-queue">>).
 
--export([start_link/9,
+%% The following capabilities are defined in amqp-bindmap-jms-v1.0-wd10 [5.2]
+%% and sent by Qpid JMS client.
+-define(CAP_TOPIC, <<"topic">>).
+-define(CAP_TEMPORARY_TOPIC, <<"temporary-topic">>).
+-define(CAP_TEMPORARY_QUEUE, <<"temporary-queue">>).
+
+-export([start_link/10,
          process_frame/2,
          list_local/0,
          conserve_resources/3,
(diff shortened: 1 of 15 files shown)


deps/rabbit/src/rabbit_db_binding.erl
```diff
@@ -565,19 +565,18 @@ match(SrcName, Match) ->
 %% match_routing_key().
 %% -------------------------------------------------------------------
 
--spec match_routing_key(Src, RoutingKeys, UseIndex) -> [Dst] when
+-spec match_routing_key(Src, RoutingKeys) -> [Dst] when
```
michaelklishin, line 568: There seems to be a call site in `rabbit_db_binding_SUITE` that still use `match_routing_key/3` (the `/3` version).

deps/rabbit/src/rabbit_amqp_session.erl
```diff
@@ -3774,7 +3961,7 @@ detach_reply(Detach,
        map_size(NewOutgoingLinks) < map_size(OldOutgoingLinks) orelse
        map_size(NewIncomingMgmtLinks) < map_size(OldIncomingMgmtLinks) orelse
        map_size(NewOutgoingMgmtLinks) < map_size(OldOutgoingMgmtLinks) ->
-    [Detach#'v1_0.detach'{error = undefined}];
+    [Detach];
```
michaelklishin, line 3964: This returns the unmodified `Detach` for all clients that are not JMS, whereas previously it was `Detach#'v1_0.detach'{error = undefined}`.

This is a subtle change that clients won't notice under normal operation but per AMQP 1.0 Core spec section `2.7.3`, responding with an unmodified client value seems… not an obvious violation but a weird thing to do.

michaelklishin: This is a tricky area where we've had two behavior changes since native AMQP 1.0 was introduced: in 7baff37f65 and 5c617725dc.

michaelklishin (changes requested):

deps/rabbit/src/rabbit_db_binding.erl
ansd: The call site is in `rabbit_db_binding_SUITE:match_routing_key1/1` which is dead code. I deleted this function.

deps/rabbit/src/rabbit_amqp_session.erl
ansd: Yes, that's to return any error during ending the JMS subscription in:
```erl
                            case jms_topic(end_subscription, [QName, User, PC]) of
                                {ok, PC1} ->
                                    {OutgoingLinks1, Unsettled1, Pending1, QStates0, Detach0, PC1};
                                {amqp_error, Condition, Description} ->
                                    Err = #'v1_0.error'{condition = Condition,
                                                        description = {utf8, Description}},
                                    Detach1 = Detach0#'v1_0.detach'{error = Err},
```

PR #17134, opened 2026-08-06
Shovel HTTP API: process URIs the same way federation does

michaelklishin:


michaelklishin:
Tested locally in two environments. Actions are having a difficult day.

PR #17153, opened 2026-08-10
[mgt] Honor X-Forwarded- headers

MarcialRosales:
This PR updates the OpenID discovery payload to correctly honor X-Forwarded-Proto, X-Forwarded-Port, and X-Forwarded-Host headers when rewriting the token URL. It also introduces new Selenium tests that simulate a TLS-terminating proxy scenario, ensuring RabbitMQ handles environments where the external client-facing host and protocol (HTTPS) differ from the internal proxy-to-RabbitMQ connection (HTTP).


michaelklishin:
@MarcialRosales I don't think respecting query string values is a good idea: it's an untrusted input, and proxies historically use headers (they are not logged as often, more discreet, that kind of reasoning).

michaelklishin:
@MarcialRosales my changes ended up being smaller in scope for the implementation but larger than I anticipated around test coverage.

Please take a look.

PR #17163, opened 2026-08-11
Let a deprecated feature be denied while in use

the-mikedavis:
The motivation for this is the deprecated `transient_nonexcl_queues` feature which moved to `denied_by_default` in 4.3.0. When denied by default, the `is_feature_used` callback is used for three purposes:

* To report deprecated features in use in the management UI and in `rabbitmqctl list_deprecated_features`, etc.
* As a gate for state transition to denied. When true, it prevents the underlying feature flag from being enabled.
* As a gate at boot, failing boot if the denied feature is in use.

The last (gate at boot) means that a single-node broker (or cluster during rolling upgrade) fails to boot if it has a transient non-exclusive queue in the metadata store. This is problematic for single-node since the cleanup of these queues happens in a later boot step, and it's a problem for clusters since an older member on 4.2 might have a transient non-exclusive queue in a mixed version cluster, preventing the 4.3 node(s)' boots.

This change allows marking a deprecated feature flag so that the underlying feature flag can be enabled even if the deprecated feature is still in use. This prevents further declarations and fixes the boot issue. This is appropriate for deprecated features which need no migration, especially if they are transient.

### DIFF
--- a/.github/workflows/test-upgrades.sh
+++ b/.github/workflows/test-upgrades.sh
@@ -67,6 +67,79 @@ case "$scenario" in
 		wait
 		;;
 
+	transient_nonexclusive_queue)
+		# Declare a transient non-exclusive classic queue. The property
+		# combination is denied by default as of 4.3.0, so it is
+		# declared on the old node where it is still permitted.
+		cat > transient-nonexclusive-definitions.json <<-'EOF'
+		{
+		  "queues": [
+		    {
+		      "arguments": {"x-queue-type": "classic"},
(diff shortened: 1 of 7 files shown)


the-mikedavis:
The wording around deprecated features is a bit confusing since it's inverted from feature flags. Put a bit more plainly: with this change, transient non-exclusive queues cannot be declared after an upgrade to 4.3 even if there are some in the metadata store. That's OK for transient entities like this that will not survive the full upgrade anyways.

Issue #16194, opened 2026-04-20
CQ shared store v2 file format


lhoguin:
CQ shared message store has always had slowness issues around compaction/deletion, which were improved on but nothing perfect. To better do compaction/deletion a new file format is necessary because the existing format doesn't have any header or anything that would avoid potential collisions (even if rare) so we can't change the format. I came up with this plan:

## 1. Goals

The current message store ("v1") has significant limitations addressed by
v2:

1. **Hole representation is implicit.** Removed messages leave gaps in segment
   files. To prevent stray bytes from being misinterpreted as messages,
   compaction zero-fills every gap. The cost of that zero-fill is proportional
   to the gap size and is the dominant I/O cost of compaction on large files.

2. **Records are not self-describing.** Every record is a message. There is no
   way to introduce new record kinds (e.g. explicit free-space markers,
   checksummed messages) without breaking compatibility with existing files.

3. **Index entries are kept too long in-memory.** Removing index entries eagerly
   means entries can't be detected and we have to parse one byte at a time.
   Keeping index entries means we have to cleanup on segment deletion.

v2 addresses all by introducing **typed records** with explicit, cheap
**HOLE markers** that the scanner skips in O(1) per hole regardless of hole
size, and by leaving a clear extension path (additional record types).

### 2.1 Decided model

- **Format identity is encoded in the file extension.**
  - `.rdq` -> v1 segment.
  - `.sqs` -> v2 segment.
- **Each store carries a single boundary integer**, `last_v1_file`:
  - `File =< last_v1_file` -> v1.
  - `File >  last_v1_file` -> v2.
  - `last_v1_file = none` -> the store has no v1 files at all.
- **Writes always produce v2** from release N onward. v1 files are only ever
  read, never created.
- **Conversion happens by attrition.** v1 files disappear naturally as their
  messages are removed (compaction reduces them, then `delete_file_if_empty`
  unlinks them).
- **Forced migration** of any v1 tail happens in a later release N+X.

### 2.2 Phased rollout

| Release | Behaviour |
|---|---|
| **N**     | Ship v2 encoder/decoder/scanner, the `last_v1_file` boundary, and dual-format reads. New files are always v2. Existing v1 files keep being read in place. No bulk migration. |
| **N+X**   | On startup, if any `*.rdq` files remain, migrate each to a fresh v2 segment (next integer, swap index entries, delete the old `*.rdq`). Set `last_v1_file = none`. |
[...]

PR #13250, opened 2025-02-12
Optimise msg_store recovery in case of large message file

gomoripeti:
## Proposed Changes

Since 4.0.0 (commit d45fbc3d) the shared message store writes large messages into their own rdq files. This information can be utilised when scanning rdq files during recovery to avoid reading in the whole message body into memory unnecessarily.

This commit addresses the same issue that was addressed in 3.13.x by commit baeefbec (ie. appending a large binary together from 4MB chunks leaves a lot of garbage and memory fragmentation behind) but even more efficiently.

Large messages which were written before 4.0.0, which don't fully fill the rdq file, are still handled as before.

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
If you're unsure about any of them, don't hesitate to ask on the mailing list.
We're here to help!
This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] All tests pass locally with my changes
- [ ] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

## Further Comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution
you did and what alternatives you considered, etc.

PR #17166, opened 2026-08-12
CQ shared store v2 file format implementation

lhoguin:
This was all written by Sonnet 5 in roughly one week. I only removed a few comments and reviewed many times with a lot of back and forth to get the code in the best shape possible.

* [CQ: Introduce a v2 file format for the shared store](https://github.com/rabbitmq/rabbitmq-server/pull/17166/changes/5568c86512e2b49e28ff1bec558f99c8e39ff7de): This is the main part, implementing the spec. Additional checks compared to the ticket were added.

* [CQ: Guard compaction's index offset update against a racing fan-out rewrite](https://github.com/rabbitmq/rabbitmq-server/pull/17166/changes/f132d59ca4812468364c48b3c89a900b2feb048d): This is a hypothetical race protection. The timing to hit this is super unlikely though so I don't think it should be backported.

* [CQ: Reject a truncated file_summary or index dump instead of loading it](https://github.com/rabbitmq/rabbitmq-server/pull/17166/changes/956101ca952d6a816e43349ab8682082fc62f7b5): An additional check in case ets tabs are truncated somehow.

* [CQ: Dispatch dirty-recovery index rebuild jobs inline](https://github.com/rabbitmq/rabbitmq-server/pull/17166/changes/b972543ad0e64733f34a7cebdd803e9081cf7b7f): Fix a potential hang at startup that could be caused via a crash elsewhere. Simplifies recovery in any case.

* [CQ: Update comments](https://github.com/rabbitmq/rabbitmq-server/pull/17166/changes/63a41130eefdcdd7f0f18deef871334776fe8d62): Last batch of AI comments removal.

First part of #16194

### DIFF
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -880,14 +880,8 @@ init([VHost, Type, BaseDir, ClientRefs, StartupFunState]) ->
     State2 = State1 #msstate { current_file         = CurFile,
                                current_file_handle  = CurHdl,
                                current_file_offset  = CurOffset },
-    %% CurFile0 was still current_file (and therefore exempt from
-    %% every delete_file_if_empty/2 check run so far, including the
-    %% one build_index/3 just did for dirty recovery) right up until
-    %% open_current_file/5 above decided to roll away from it because
-    %% it's v1. If it turned out to hold no valid messages, nothing
-    %% else will ever reconsider it once we return here with a
-    %% different file current: check it once, now, while we still
-    %% remember it was just abandoned.
+    %% After an upgrade the most recent v1 file may be empty,
+    %% we delete it now instead of leaving it forever.
     State3 = case CurFile of
         CurFile0 -> State2;
         _        -> delete_file_if_empty(CurFile0, State2)
@@ -896,14 +890,9 @@ init([VHost, Type, BaseDir, ClientRefs, StartupFunState]) ->
      hibernate,
      {backoff, ?HIBERNATE_AFTER_MIN, ?HIBERNATE_AFTER_MIN, ?DESIRED_HIBERNATE}}.
 
-%% The current write file must always be v2. build_index/3 may have left
-%% us pointing at a v1 file, either because the store predates v2 or
-%% because nothing has been written since upgrading: in that case we
-%% force a roll to a fresh v2 file (CurFile0 + 1) before accepting any
-%% writes. Separately, if the recovered v2 "current" file has nothing
-%% but (at most) a partial header in it, we (re)create it fresh via
-%% writer_open/2 rather than let writer_recover/3's truncate conjure up
-%% a correct 64-byte header out of implicit zero-fill.
+%% The current write file is always v2. If the current file is v1,
+%% we create a new v2 current file. We also do if the v2 file has
+%% nothing in it (we overwrite the header if any).
 open_current_file(Dir, CurFile, CurOffset, LastV1File, FileSummaryEts) ->
     case file_format(CurFile, LastV1File) of
         %% v1 files can't take v2 records, so roll over to a brand new file.
@@ -1654,10 +1643,6 @@ mark_handle_closed(FileHandlesEts, File, Ref) ->
 
 form_filename(Dir, Name) -> filename:join(Dir, Name).
 
-%% Format is determined solely by the file number relative to the
-%% store's last_v1_file boundary: writes always produce v2 from this
-%% release onward, so any file above the boundary (or every file, when
-%% the store has no v1 files at all) is v2.
 file_format(_File, none)                       -> v2;
 file_format(File,  LastV1) when File =< LastV1 -> v1;
 file_format(_,     _)                          -> v2.
@@ -1816,7 +1801,7 @@ scan_v1_next_byte(<<_, Rest/bits>>, Fd, Fun, Offset, FileSize, MsgIdsFound, Acc)
 %% as in v1. We never fall back to scanning the next byte: that is
 %% exactly the kind of guess that could land on something that merely
 %% looks like a record.
-
+%%
 %% Any exception raised while scanning (whether one of the deliberate
 %% rabbit_msg_store_v2_scan corruption errors above, or anything else
 %% unexpected) is re-raised with Path attached, so whatever eventually
@@ -1826,12 +1811,6 @@ scan_v2_file_for_valid_messages(Path, Fun) ->
     try
         case file:open(Path, [read, binary, raw]) of
             {ok, Fd} ->
-                %% The fd must close on every exit from the scan below,
-                %% not just success: scan_file_recovering_corruption/2
-                %% retries a failed scan of this same file in a loop,
-                %% and each retry calls back in here, so a leaked fd
-                %% per failed attempt would accumulate for as long as
-                %% the process doing the scanning lives.
                 try
                     {ok, FileSize} = file:position(Fd, eof),
                     Messages = case FileSize < ?V2_HEADER_SIZE of
@@ -1964,11 +1943,7 @@ scan_v2_data(<<?REC_MESSAGE:8, Size:32, Rest0/bits>> = Data, Fd, Fun, Offset, Fi
             PayloadLen = Size - 21,
             case Rest0 of
                 <<MsgIdInt:128, Body:PayloadLen/binary, Rest/bits>> ->
-                    %% A torn write reproducing a syntactically valid
-                    %% Erlang external term by chance is astronomically
-                    %% unlikely, so this catches a Type/Size that happened
-                    %% to validate numerically but isn't really a message,
-                    %% regardless of what MsgId lookup would say below.
+                    %% We only need to know this is a valid Erlang term.
                     try binary_to_term(Body) of
                         _ -> ok
                     catch
@@ -2039,9 +2014,7 @@ index_recover(Dir) ->
     Path = filename:join(Dir, ?INDEX_FILE_NAME),
     %% {verify, true} rejects a truncated dump (e.g. left behind by a
     %% shutdown that was killed mid-write) instead of silently loading
-    %% however many objects made it to disk. The caller already falls
-    %% back to a full dirty-recovery rescan on any error here, which is
-    %% what must happen instead of trusting a partial index.
+    %% however many objects made it to disk.
     case ets:file2tab(Path, [{verify, true}]) of
         {ok, IndexEts}  -> _ = file:delete(Path),
                            {ok, IndexEts};
@@ -2081,16 +2054,8 @@ index_update(IndexEts, Obj) ->
 %% write racing this compaction can delete and reinsert the same msg_id
 %% under a different file (or a different offset in the same file) before
 %% this runs; if that happened, the entry no longer belongs to this
-%% compaction pass and must not be touched -- its own physical bytes were
-%% already copied into the new location regardless (write_compact_file_v2
-%% doesn't know or care whether the index update will succeed), so the
-%% only consequence of skipping is an orphaned, harmless duplicate copy
-%% that scan_and_vacuum_message_file/3's existing fan-out clause already
-%% recognises and discards on this file's next scan.
+%% compaction pass and must not be touched.
 index_update_offset_if_unchanged(IndexEts, MsgId, File, OldOffset, NewOffset) ->
-    %% The extra {...} around the replacement record is required by
-    %% ets:select_replace/2: without it, a literal tuple appearing in a
-    %% match spec's body is parsed as a function call instead.
     MatchSpec = [{#msg_location{msg_id = MsgId, ref_count = '$1', file = File,
                                 offset = OldOffset, total_size = '$2'},
                   [],
@@ -2173,8 +2138,6 @@ recover_index_and_client_refs(true, ClientRefs, Dir, Name) ->
             end
     end.
 
-%% Ground truth for the v1/v2 boundary when it cannot be trusted from
-%% persisted terms (dirty recovery).
 scan_for_last_v1_file(Dir) ->
     case filelib:wildcard("*" ++ ?FILE_EXTENSION_V1, Dir) of
         []    -> none;
@@ -2206,13 +2169,9 @@ recover_file_summary(false, _Dir) ->
                     [ordered_set, public, {keypos, #file_summary.file}])};
 recover_file_summary(true, Dir) ->
     Path = filename:join(Dir, ?FILE_SUMMARY_FILENAME),
-    %% {verify, true} rejects a truncated dump instead of silently
-    %% loading whatever entries made it to disk -- a partial table can
-    %% be missing exactly the highest-numbered (most recently added)
-    %% files, which would otherwise make a stale, lower file look like
-    %% the current one. Falling back to recover_file_summary(false, ...)
-    %% forces the full dirty-recovery rescan, which derives the true
-    %% current file from the real directory listing instead.
+    %% {verify, true} rejects a truncated dump (e.g. left behind by a
+    %% shutdown that was killed mid-write) instead of silently loading
+    %% however many objects made it to disk.
     case ets:file2tab(Path, [{verify, true}]) of
         {ok, Tid}       -> ok = file:delete(Path),
                            {true, Tid};
@@ -2374,13 +2333,8 @@ scan_file_recovering_corruption(Path, Fun) ->
             end
     end.
 
-%% Every rabbit_msg_store_v2_scan reason (wrapped by
-%% scan_v2_file_for_valid_messages/2 in rabbit_msg_store_v2_scan_error)
-%% reports the offset of the record it rejected as its second element.
-%% Only that specific, unambiguous shape is trusted: anything else,
-%% including a tuple that merely happens to also have an integer in
-%% the same position, is not our own error and must not be mistaken
-%% for one.
+%% Every rabbit_msg_store_v2_scan reason reports the offset of the
+%% record it rejected as its second element.
 corruption_offset({rabbit_msg_store_v2_scan_error, _Path, InnerReason})
         when is_tuple(InnerReason), tuple_size(InnerReason) >= 2,
              element(1, InnerReason) =:= rabbit_msg_store_v2_scan,
@@ -2388,8 +2342,6 @@ corruption_offset({rabbit_msg_store_v2_scan_error, _Path, InnerReason})
     {ok, element(2, InnerReason)};
 corruption_offset(_Reason) -> error.
 
-%% Runs in the same process as its caller, rebuild_index/3, rather than
-%% a separate spawned process -- see the comment there for why.
 enqueue_build_index_workers(_Gatherer, [], _State) ->
     ok;
 enqueue_build_index_workers(Gatherer, [File|Files], State) ->
@@ -2420,22 +2372,6 @@ reduce_index(Gatherer, LastFile,
             reduce_index(Gatherer, LastFile, State)
     end.
 
-%% Dispatching used to happen in a separate, unlinked, spawned process,
-%% concurrently with reduce_index/3 below draining the gatherer. If
-%% that process died before dispatching every file -- for any reason,
-%% since nothing linked or usefully monitored it -- the gatherer's
-%% per-file fork count (already bumped for every file up front, just
-%% below) would never reach zero, so gatherer:out/1 would never return
-%% empty and reduce_index/3 would block forever; rabbit_msg_store's own
-%% start_link/5 uses {timeout, infinity}, so nothing would ever time
-%% that out either -- the store just never came up, with no crash and
-%% no log. worker_pool:dispatch_sync/1 paces on free-worker availability
-%% either way, identically whether it's called from a separate process
-%% or inline, so doing it inline here doesn't change how long dispatching
-%% every file takes; it just means a failure part-way through (e.g. the
-%% shared worker_pool itself dying while a dispatch call is in flight)
-%% now propagates directly out of this function instead of vanishing
-%% silently.
 rebuild_index(Gatherer, Files, State) ->
     lists:foreach(fun (_File) ->
                           ok = gatherer:fork(Gatherer)
@@ -2683,27 +2619,10 @@ do_compact_file_v1(Fd, _, [#msg_location{ offset = StartMsgOffset, total_size = 
 do_compact_file_v1(_, Offset, _, [], IndexAcc) ->
     {ok, Offset, IndexAcc}.
 
-%% gap_size == 0  -> nothing
-%% gap_size <  5  -> 1..4 SMALL_HOLE bytes
-%% gap_size >= 5  -> one HOLE record, inner bytes zeroed
-%%
 %% The scanner only ever looks at the HOLE record's own 5-byte header
 %% to know how much to skip; it never reads the inner bytes. We zero
 %% them anyway so that a completed write leaves no message bytes
 %% behind for a raw scan (a salvage tool, for instance) to recover.
-%%
-%% This doesn't add crash resilience: the zeros are the last bytes in
-%% the pwrite, written after the message and the header, so a write
-%% torn partway through is, if anything, more likely to leave old
-%% bytes behind unzeroed than to leave zeros. What a scanner relies on
-%% after a crash is the header's own type and size landing intact,
-%% which zeroing the payload does nothing for.
-%%
-%% It does have one incidental benefit for a *different* failure mode,
-%% though: REC_RESERVED is 0 and has no matching scan clause, so if a
-%% scan ever gets misaligned (say, from a corrupted Size elsewhere) and
-%% lands inside a hole's zeroed payload, it hits the unrecognised-type
-%% crash instead of silently reading zeros as if they were meaningful.
 hole_bytes(0) ->
     <<>>;
 hole_bytes(Size) when Size < 5 ->
@@ -2714,15 +2633,7 @@ hole_bytes(Size) ->
 %% We decide the whole compaction plan first, as pure data,
 %% then write it. This way nothing is ever written based on
 %% information that a later part of the plan could still
-%% invalidate. write_compact_file_v2/2 writes a hole record for
-%% whatever's left of the space each run of moves lands in, so the
-%% file is always fully valid and scannable at every point during
-%% compaction and before the physical truncation that follows it has
-%% actually happened: a crash never leaves an unmarked, ambiguous gap
-%% for the next scan (this one, recovery, or the salvage tool) to
-%% trip over. Messages we leave in place are untouched, including
-%% whatever follows them: we haven't broken anything there, so there
-%% is nothing for us to mark.
+%% invalidate.
 do_compact_file_v2(Fd, StartOffset, Messages) ->
     {Plan, TruncateSize} = plan_compact_file_v2(StartOffset, Messages, lists:reverse(Messages), []),
     IndexUpdates = write_compact_file_v2(Fd, Plan),
@@ -2735,15 +2646,6 @@ do_compact_file_v2(Fd, StartOffset, Messages) ->
 %% this never touches the file: it only decides the final, static
 %% layout, so nothing here can be invalidated by a read or a write
 %% that a later decision hasn't been made yet when it happens.
-%%
-%% A move's own hole size isn't known until we see what ends up right
-%% after it in the finished plan: packing can stop either because this
-%% was the very last message that could possibly move (the terminal
-%% clause below), or because the next candidate from the end of the
-%% file simply doesn't fit and we fall through to leaving a message
-%% where it is instead. Either way leaves a leftover behind this move,
-%% so the hole size is filled in from the finished plan, in
-%% write_compact_file_v2/2, rather than decided here.
 plan_compact_file_v2(Offset, Start = [#msg_location{ offset = StartMsgOffset }|_],
         [#msg_location{ msg_id = EndMsgId, offset = EndMsgOffset, total_size = EndMsgSize }|EndTail],
         Acc)
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -40,7 +40,18 @@
 -define(CLEAN_FILENAME, "clean.dot").
 -define(FILE_SUMMARY_FILENAME, "file_summary.ets").
 
--define(FILE_EXTENSION,        ".rdq").
+-define(FILE_EXTENSION_V1, ".rdq").
+-define(FILE_EXTENSION_V2, ".sqs").
+
+-define(V2_MAGIC, 16#52535153). %% "RSQS"
+-define(V2_HEADER_SIZE, 64). %% bytes
+
+%% Record 0 is reserved forever so that zero-filled regions
+%% never look like valid records.
+-define(REC_RESERVED,   0).
+-define(REC_SMALL_HOLE, 1).
+-define(REC_HOLE,       2).
+-define(REC_MESSAGE,    3).
 
 %% We keep track of flying messages for writes and removes. The idea is that
 %% when a remove comes in before we could process the write, we skip the
@@ -109,7 +120,10 @@
           %% client ref to synced messages mapping
           cref_to_msg_ids,
           %% See CREDIT_DISC_BOUND in rabbit.hrl
-          credit_disc_bound
+          credit_disc_bound,
+          %% Highest-numbered v1 (.rdq) file, or 'none' if the store has
+          %% no v1 files at all.
+          last_v1_file
         }).
 
 -record(client_msstate,
@@ -121,7 +135,8 @@
           file_handles_ets,
           cur_file_cache_ets,
           flying_ets,
-          credit_disc_bound
+          credit_disc_bound,
+          last_v1_file
         }).
 
 -record(file_summary,
@@ -132,7 +147,8 @@
           index_ets,
           file_summary_ets,
           file_handles_ets,
-          msg_store
+          msg_store,
+          last_v1_file
         }).
 
 -record(dying_client,
@@ -149,23 +165,26 @@
                                 index_ets        :: ets:tid(),
                                 file_summary_ets :: ets:tid(),
                                 file_handles_ets :: ets:tid(),
-                                msg_store        :: server()
+                                msg_store        :: server(),
+                                last_v1_file     :: file_num() | 'none'
                               }.
 
 -type server() :: pid() | atom().
 -type client_ref() :: binary().
 -type file_num() :: non_neg_integer().
+-type file_format() :: 'v1' | 'v2'.
 -type client_msstate() :: #client_msstate {
                       server             :: server(),
                       client_ref         :: client_ref(),
-                      reader             :: undefined | {non_neg_integer(), file:fd()},
+                      reader             :: undefined | {file_num(), file:fd(), file_format()},
                       index_ets          :: any(),
                       %% Stored as binary() as opposed to file:filename() to save memory.
                       dir                :: binary(),
                       file_handles_ets   :: ets:tid(),
                       cur_file_cache_ets :: ets:tid(),
                       flying_ets         :: ets:tid(),
-                      credit_disc_bound  :: {pos_integer(), pos_integer()}}.
+                      credit_disc_bound  :: {pos_integer(), pos_integer()},
+                      last_v1_file       :: file_num() | 'none'}.
 -type msg_ref_delta_gen(A) ::
         fun ((A) -> 'finished' |
                     {rabbit_types:msg_id(), non_neg_integer(), A}).
@@ -408,7 +427,7 @@ gc_pid(Server) ->
 -spec client_init(server(), client_ref(), maybe_msg_id_fun()) -> client_msstate().
 
 client_init(Server, Ref, MsgOnDiskFun) when is_pid(Server); is_atom(Server) ->
-    {IndexEts, Dir, FileHandlesEts, CurFileCacheEts, FlyingEts} =
+    {IndexEts, Dir, FileHandlesEts, CurFileCacheEts, FlyingEts, LastV1File} =
         gen_server2:call(
           Server, {new_client_state, Ref, self(), MsgOnDiskFun},
           infinity),
@@ -422,7 +441,8 @@ client_init(Server, Ref, MsgOnDiskFun) when is_pid(Server); is_atom(Server) ->
                       file_handles_ets   = FileHandlesEts,
                       cur_file_cache_ets = CurFileCacheEts,
                       flying_ets         = FlyingEts,
-                      credit_disc_bound  = CreditDiscBound }.
+                      credit_disc_bound  = CreditDiscBound,
+                      last_v1_file       = LastV1File }.
 
 -spec client_terminate(client_msstate()) -> 'ok'.
 
@@ -518,7 +538,8 @@ read_many_file2(MsgIds0, CState = #client_msstate{ dir              = Dir,
                                                    index_ets        = IndexEts,
                                                    file_handles_ets = FileHandlesEts,
                                                    reader           = Reader0,
-                                                   client_ref       = Ref }, Acc0, File) ->
+                                                   client_ref       = Ref,
+                                                   last_v1_file     = LastV1File }, Acc0, File) ->
     %% Mark file handle open.
     mark_handle_open(FileHandlesEts, File, Ref),
     %% Get index for all Msgids in File.
@@ -539,7 +560,7 @@ read_many_file2(MsgIds0, CState = #client_msstate{ dir              = Dir,
             %% Then we can do the consolidation to get the pread LocNums.
             LocNums = consolidate_reads(MsgLocations, []),
             %% Read the data from the file.
-            Reader = reader_open(Reader0, Dir, File),
+            Reader = reader_open(Reader0, Dir, File, LastV1File),
             {ok, Msgs} = reader_pread(Reader, LocNums),
             %% Before we continue the read_many calls we must remove the
             %% MsgIds we have read from the list and add the messages to
@@ -642,8 +663,10 @@ client_read3(#msg_location { msg_id = MsgId, file = File },
     end.
 
 read_from_disk(#msg_location { msg_id = MsgId, file = File, offset = Offset,
-                               total_size = TotalSize }, State = #client_msstate{ reader = Reader0, dir = Dir }) ->
-    Reader = reader_open(Reader0, Dir, File),
+                               total_size = TotalSize },
+               State = #client_msstate{ reader = Reader0, dir = Dir,
+                                        last_v1_file = LastV1File }) ->
+    Reader = reader_open(Reader0, Dir, File, LastV1File),
     Msg = case reader_pread(Reader, [{Offset, TotalSize}]) of
         {ok, [Msg0]} ->
             Msg0;
@@ -659,7 +682,7 @@ read_from_disk(#msg_location { msg_id = MsgId, file = File, offset = Offset,
     {Msg, State#client_msstate{ reader = Reader }}.
 
 %%----------------------------------------------------------------------------
-%% Reader functions. A reader is a file num + fd tuple.
+%% Reader functions. A reader is a file num + fd + format tuple.
 %%----------------------------------------------------------------------------
 
 %% The reader tries to keep the FD open for subsequent reads.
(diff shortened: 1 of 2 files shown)


deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -40,7 +40,22 @@
 -define(CLEAN_FILENAME, "clean.dot").
 -define(FILE_SUMMARY_FILENAME, "file_summary.ets").
 
--define(FILE_EXTENSION,        ".rdq").
+-define(FILE_EXTENSION_V1,     ".rdq").
+-define(FILE_EXTENSION_V2,     ".sqs").
+
+%% v2 segment file header: MAGIC:32, VERSION:8, reserved (zero-filled).
+%% "RSQS" (Rabbit Shared Queue Store), distinct from "RCQS"/"RCQI".
```
lhoguin: Mostly pointless comments.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -40,7 +40,22 @@
...
+-define(V2_MAGIC,       16#52535153).
+-define(V2_VERSION,     2).
+-define(V2_HEADER_SIZE, 64).
+
+%% v2 record types. Type 0 is reserved forever: a zero-filled or
+%% uninitialised region can then never accidentally look like a
+%% valid record.
```
lhoguin: Simplify comment.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -668,43 +695,66 @@ read_from_disk(#msg_location { msg_id = MsgId, file = File, offset = Offset,
...
+    Format = file_format(File, LastV1File),
     {ok, Fd} = case Reader of
-        {File, Fd0} ->
+        {File, Fd0, _} ->
             {ok, Fd0};
-        {_AnotherFile, Fd0} ->
+        {_AnotherFile, Fd0, _} ->
             ok = file:close(Fd0),
-            file:open(form_filename(Dir, filenum_to_name(File)),
+            file:open(form_filename(Dir, filenum_to_name(File, LastV1File)),
```
lhoguin: We already know `Format` no need to recompute it.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -668,43 +695,66 @@ read_from_disk(#msg_location { msg_id = MsgId, file = File, offset = Offset,
...
                       [binary, read, raw]);
         undefined ->
-            file:open(form_filename(Dir, filenum_to_name(File)),
+            file:open(form_filename(Dir, filenum_to_name(File, LastV1File)),
```
lhoguin: Same as above.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -668,43 +695,66 @@ read_from_disk(#msg_location { msg_id = MsgId, file = File, offset = Offset,
...
-reader_pread_parse([<<>>]) ->
+    [binary_to_term(MsgBodyBin)|reader_pread_parse_v1([Rest|Tail])];
+reader_pread_parse_v1([<<>>]) ->
+    [];
+reader_pread_parse_v1([<<>>|Tail]) ->
+    reader_pread_parse_v1(Tail).
+
+%% Locations come straight from the index with exact total sizes, so we
+%% trust Type/Size here rather than re-validating them (that is the
+%% scanner's job).
```
lhoguin: Pointless comment.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -799,16 +851,51 @@ init([VHost, Type, BaseDir, ClientRefs, StartupFunState]) ->
...
+            {ok, Hdl} = writer_open(Dir, NextFile),
+            {NextFile, ?V2_HEADER_SIZE, Hdl};
+        v2 when CurOffset =< ?V2_HEADER_SIZE ->
+            true = ets:insert(FileSummaryEts, #file_summary {
+                                file             = CurFile,
+                                valid_total_size = 0,
+                                file_size        = ?V2_HEADER_SIZE,
+                                locked           = false }),
+            {ok, Hdl} = writer_open(Dir, CurFile),
+            {CurFile, ?V2_HEADER_SIZE, Hdl};
```
lhoguin: Surely there's a less verbose way to do this.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1312,8 +1401,8 @@ write_large_message(MsgId, MsgBodyBin,
                                   file_summary_ets    = FileSummaryEts,
                                   cur_file_cache_ets  = CurFileCacheEts }) ->
     {LargeMsgFile, LargeMsgHdl} = case CurOffset of
-        %% We haven't written in the file yet. Use it.
-        0 ->
+        %% We haven't written anything but the header in the file yet. Use it.
+        ?V2_HEADER_SIZE ->
```
lhoguin, line 1428: Double check whether 0 is possible.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1326,10 +1415,13 @@ write_large_message(MsgId, MsgBodyBin,
     %% Write the message directly and close the file.
     TotalSize = writer_direct_write(LargeMsgHdl, MsgId, MsgBodyBin),
     ok = writer_close(LargeMsgHdl),
-    %% Update ets with the new information.
+    %% Update ets with the new information. Either way the large message
+    %% lands immediately after the header: LargeMsgFile is either the
+    %% current file with nothing but its header written so far, or a
+    %% freshly opened file whose header was just written by writer_open/2.
```
lhoguin: Pointless comment.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1452,35 +1544,48 @@ should_mask_action(CRef, MsgId, #msstate{
 
 -define(MAX_BUFFER_SIZE, 1048576). %% 1MB.
 
+%% Writes always produce v2 (release N onward): writer_open/writer_recover
+%% are only ever reached for the current write file, which is always v2
+%% (see open_current_file/5, which forces a roll away from any v1 current
+%% file before either is ever called).
```
lhoguin: Simplify comment.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1452,35 +1544,48 @@ should_mask_action(CRef, MsgId, #msstate{
...
 writer_open(Dir, Num) ->
-    {ok, Fd} = file:open(form_filename(Dir, filenum_to_name(Num)),
-                         [append, binary, raw]),
+    {ok, Fd} = file:open(form_filename(Dir, filenum_to_name(Num, none)),
+                         [read, write, binary, raw]),
+    %% Truncate first: writer_open/2 is only ever called on a file with no
+    %% valid message data yet, so discarding whatever is already there is
+    %% always safe.
+    {ok, 0} = file:position(Fd, bof),
+    ok = file:truncate(Fd),
```
lhoguin: I wouldn't do that. I would open for append and confirm we are at byte 0 if anything.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1520,13 +1624,32 @@ mark_handle_closed(FileHandlesEts, File, Ref) ->
...
+%% release onward, so any file above the boundary (or every file, when
+%% the store has no v1 files at all) is v2.
+file_format(_File, none)                       -> v2;
+file_format(File,  LastV1) when File =< LastV1 -> v1;
+file_format(_,     _)                          -> v2.
+
+%% Passing 'none' as LastV1File forces v2 regardless of File. This is
+%% used at every site that is only ever writing/creating a brand new
+%% file, since new files are always v2.
+filenum_to_name(File, LastV1File) ->
```
lhoguin: If we already know v1/v2 we shouldn't go through this.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1520,13 +1624,32 @@ mark_handle_closed(FileHandlesEts, File, Ref) ->
...
 form_filename(Dir, Name) -> filename:join(Dir, Name).
 
-filenum_to_name(File) -> integer_to_list(File) ++ ?FILE_EXTENSION.
+%% Format is determined solely by the file number relative to the
+%% store's last_v1_file boundary: writes always produce v2 from this
+%% release onward, so any file above the boundary (or every file, when
+%% the store has no v1 files at all) is v2.
+file_format(_File, none)                       -> v2;
+file_format(File,  LastV1) when File =< LastV1 -> v1;
+file_format(_,     _)                          -> v2.
```
lhoguin, line 1648: Double check whether LastV1 gets `none` if unset in terms.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1602,26 +1741,152 @@ scan_data(<<Size:64, MsgIdInt:128, _Rest/bits>> = Data, Fd, Fun, Offset, FileSiz
...
+%% a HOLE below the 5-byte minimum) are skipped without interpreting
+%% their contents. A HOLE larger than what's currently buffered is
+%% skipped via file:position/2 rather than read, so scanning cost does
+%% not scale with the size of the hole. Only a genuinely unrecognised
+%% Type byte (corruption unrelated to a simple torn write) crashes;
+%% a record whose Size does not fit within the file is treated as a
+%% torn last write and the remainder of the file is discarded, exactly
+%% as in v1.
+
+scan_v2_file_for_valid_messages(Path, Fun) ->
```
lhoguin, line 1810: I need to do a deep review of this function.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1851,8 +2135,15 @@ build_index_worker(Gatherer, #msstate { index_ets = IndexEts, dir = Dir },
...
+        %% A v2 file always has at least its header even when it holds no
+        %% (or no longer any) valid messages; a v1 file with no messages
+        %% genuinely has nothing on disk. Getting this wrong for v2 would
+        %% let a later truncate/recover destroy the header.
         [] ->
-            0;
+            case Format of
+                v1 -> 0;
+                v2 -> ?V2_HEADER_SIZE
+            end;
```
lhoguin, line 2274: What happens if the file to be recovered is 0 bytes in length (or < ?V2_HEADER_SIZE in length, partial header).

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2011,24 +2302,39 @@ compact_file(File, FileSize,
...
+    %% do_compact_file/5 only ever pwrites message bytes: when it moves an
+    %% end message into a hole that the message doesn't exactly fill (the
+    %% common case), whatever is left of that hole is now unmarked,
+    %% arbitrary bytes rather than a HOLE/SMALL_HOLE record. v1's scanner
+    %% tolerates that (it byte-shifts through anything that doesn't look
+    %% like a message), but v2's does not by design, so leaving it would
+    %% make the next scan of this file (recovery, salvage, or another
+    %% compaction) crash on an unrecognised record type.
+    case Format of
+        v2 -> rewrite_holes_after_compaction(Fd, Messages, IndexUpdates);
```
lhoguin: This is wrong. Holes must be filled immediately when moving messages so the file is always consistent. That means do_compact_file should behave differently for v1/v2.

lhoguin (commented):
First pass. I did not review tests. A couple severe issues to address first.

deps/rabbit/src/rabbit_msg_store.erl
lhoguin: I've restricted this truncate to the part that can legitimately use it.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2011,24 +2293,40 @@ compact_file(File, FileSize,
...
-                                 msg_store        = Server }) ->
+                                 msg_store        = Server,
+                                 last_v1_file     = LastV1File }) ->
+    Format = file_format(File, LastV1File),
     %% Get metadata about the file. Will be used to calculate
     %% how much data was reclaimed as a result of compaction.
     [#file_summary{file_size = FileSize}] = ets:lookup(FileSummaryEts, File),
     %% Open the file.
-    FileName = filenum_to_name(File),
+    FileName = filenum_to_name(File, LastV1File),
```
lhoguin: Should not recompute the format.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2011,24 +2293,40 @@ compact_file(File, FileSize,
...
-    {ok, TruncateSize, IndexUpdates} = do_compact_file(Fd, 0, Messages, lists:reverse(Messages), []),
+    Messages = scan_and_vacuum_message_file(Format, File, State),
+    StartOffset = compact_start_offset(Format),
+    {ok, TruncateSize, IndexUpdates} = case Format of
+        v1 ->
+            %% v1 has no typed hole record to fall back on, so leftover
+            %% message bytes are zero-filled up front, before moving
+            %% anything: unchanged from before the v2 format existed.
+            blank_holes_in_file_v1(Fd, Messages),
+            do_compact_file(Fd, StartOffset, Messages, lists:reverse(Messages), []);
```
lhoguin: Rename to do_compact_file_v1.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2011,24 +2293,40 @@ compact_file(File, FileSize,
...
-    Messages = scan_and_vacuum_message_file(File, State),
-    %% Blank holes. We must do this first otherwise the file is left
-    %% with data that may confuse the code (for example data that looks
-    %% like a message, isn't a message, but spans over a real message).
-    %% We blank more than is likely required but better safe than sorry.
-    blank_holes_in_file(Fd, Messages),
-    %% Compact the file.
-    {ok, TruncateSize, IndexUpdates} = do_compact_file(Fd, 0, Messages, lists:reverse(Messages), []),
+    Messages = scan_and_vacuum_message_file(Format, File, State),
+    StartOffset = compact_start_offset(Format),
```
lhoguin: Doesn't need a separate function.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2011,24 +2293,40 @@ compact_file(File, FileSize,
...
+            %% We decide the whole compaction plan first, as pure data,
+            %% then write it. This way nothing is ever written based on
+            %% information that a later part of the plan could still
+            %% invalidate. Every write coming out of the plan either
+            %% moves a message together with a hole record for whatever
+            %% hole space is left after it, or writes a lone hole record
+            %% for a gap nothing moves into, so the file is always fully
+            %% valid and scannable at every point during compaction: a
+            %% crash never leaves an unmarked, ambiguous gap for the next
+            %% scan (this one, recovery, or the salvage tool) to trip over.
```
lhoguin: Comment should be moved to do_compact_file_v2.

lhoguin (commented):
Partial review, push incoming.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2131,10 +2435,107 @@ do_compact_file(Fd, _, [#msg_location{ offset = StartMsgOffset, total_size = Sta
...
+%% gap_size == 0  -> nothing
+%% gap_size <  5  -> 1..4 SMALL_HOLE bytes
+%% gap_size >= 5  -> one HOLE record, inner bytes zeroed
+%%
+%% The scanner only ever looks at the HOLE record's own 5-byte header
+%% to know how much to skip; it never reads the inner bytes. We zero
+%% them anyway so that whatever used to be there (message bytes we no
+%% longer want readable by a raw scan, or unrelated leftover data) is
+%% actually gone, and so that a crash partway through this write can
+%% only ever leave zeros behind, never something that happens to look
```
lhoguin: I don't think that's true because the zeros are at the end of the write so the bytes are likely not zeroed in case of a crash.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1602,26 +1728,218 @@ scan_data(<<Size:64, MsgIdInt:128, _Rest/bits>> = Data, Fd, Fun, Offset, FileSiz
...
+
+%% A HOLE record with a size that does not fit the type's minimum is
+%% corruption, not a torn write (a torn write cannot fully flush a
+%% valid 5-byte HOLE header with a bogus size field).
+scan_v2_data(<<?REC_HOLE:8, Size:32, _/bits>>, _Fd, _Fun, Offset, _FileSize, _MsgIdsFound, _Acc)
+        when Size < 5 ->
+    error({rabbit_msg_store_v2_scan, invalid_hole_size, Size, Offset});
+scan_v2_data(<<?REC_HOLE:8, Size:32, Rest0/bits>>, Fd, Fun, Offset, FileSize, MsgIdsFound, Acc) ->
+    case Offset + Size > FileSize of
+        true ->
```
lhoguin, line 1906: That's an error so should be a comment even if we accept this behavior.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1602,26 +1728,218 @@ scan_data(<<Size:64, MsgIdInt:128, _Rest/bits>> = Data, Fd, Fun, Offset, FileSiz
...
+    %% Type byte seen but the 4-byte size field isn't fully buffered yet.
+    scan_v2(Data, Fd, Fun, Offset, FileSize, MsgIdsFound, Acc);
+
+scan_v2_data(<<?REC_MESSAGE:8, Size:32, _/bits>>, _Fd, _Fun, Offset, _FileSize, _MsgIdsFound, _Acc)
+        when Size < 21 ->
+    error({rabbit_msg_store_v2_scan, invalid_message_size, Size, Offset});
+scan_v2_data(<<?REC_MESSAGE:8, Size:32, Rest0/bits>> = Data, Fd, Fun, Offset, FileSize, MsgIdsFound, Acc) ->
+    case Offset + Size > FileSize of
+        true ->
+            Acc;
```
lhoguin, line 1940: That's an error and we should probably not accept this behavior for messages.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1602,26 +1728,218 @@ scan_data(<<Size:64, MsgIdInt:128, _Rest/bits>> = Data, Fd, Fun, Offset, FileSiz
...
+                    %% A torn write reproducing a syntactically valid
+                    %% Erlang external term by chance is astronomically
+                    %% unlikely, so this catches a Type/Size that happened
+                    %% to validate numerically but isn't really a message,
+                    %% regardless of what MsgId lookup would say below.
+                    try binary_to_term(Body) of
+                        _ -> ok
+                    catch
+                        error:badarg ->
+                            error({rabbit_msg_store_v2_scan, invalid_message_body, Offset})
```
lhoguin: Errors should indicate the file/path of the wrong file so it can be logged.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1602,26 +1728,218 @@ scan_data(<<Size:64, MsgIdInt:128, _Rest/bits>> = Data, Fd, Fun, Offset, FileSiz
...
+                            scan_v2_data(Rest, Fd, Fun, Offset + Size, FileSize, MsgIdsFound, Acc);
+                        _ ->
+                            case check_msg(Fun, MsgIdInt, Size, Offset, Acc) of
+                                {continue, NewAcc} ->
+                                    scan_v2_data(Rest, Fd, Fun, Offset + Size, FileSize,
+                                                 MsgIdsFound#{MsgIdInt => true}, NewAcc);
+                                %% Unlike v1, a record that Fun rejects (e.g.
+                                %% superseded elsewhere) is still known-good
+                                %% data once Type/Size validate, not
+                                %% suspected corruption.
```
lhoguin, line 1967: Isn't this wrong?

lhoguin:
All good as far as implementation goes. Still need to review tests.

lhoguin:
All that remains is a human review and a lot of squashing.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -668,43 +691,86 @@ read_from_disk(#msg_location { msg_id = MsgId, file = File, offset = Offset,
...
-reader_pread_parse([<<Size:64,
-                      _MsgId:16/binary,
-                      Rest0/bits>>|Tail]) ->
+reader_pread_parse(DataL, v1, File) -> reader_pread_parse_v1(DataL, File);
+reader_pread_parse(DataL, v2, File) -> reader_pread_parse_v2(DataL, File).
+
+%% A message record that fails to parse or decode raises with the file
+%% number attached, rather than crashing with a bare badmatch/badarg or
+%% (if nothing here even matches) a function_clause that says nothing
+%% about which file was being read.
```
lhoguin: Should also crash with MsgId (or undefined if none).

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -799,16 +867,66 @@ init([VHost, Type, BaseDir, ClientRefs, StartupFunState]) ->
...
+                               current_file_handle  = CurHdl,
+                               current_file_offset  = CurOffset },
+    %% CurFile0 was still current_file (and therefore exempt from
+    %% every delete_file_if_empty/2 check run so far, including the
+    %% one build_index/3 just did for dirty recovery) right up until
+    %% open_current_file/5 above decided to roll away from it because
+    %% it's v1. If it turned out to hold no valid messages, nothing
+    %% else will ever reconsider it once we return here with a
+    %% different file current: check it once, now, while we still
+    %% remember it was just abandoned.
```
lhoguin: OK shorten the comment.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -799,16 +867,66 @@ init([VHost, Type, BaseDir, ClientRefs, StartupFunState]) ->
...
      {backoff, ?HIBERNATE_AFTER_MIN, ?HIBERNATE_AFTER_MIN, ?DESIRED_HIBERNATE}}.
 
+%% The current write file must always be v2. build_index/3 may have left
+%% us pointing at a v1 file, either because the store predates v2 or
+%% because nothing has been written since upgrading: in that case we
+%% force a roll to a fresh v2 file (CurFile0 + 1) before accepting any
+%% writes. Separately, if the recovered v2 "current" file has nothing
+%% but (at most) a partial header in it, we (re)create it fresh via
+%% writer_open/2 rather than let writer_recover/3's truncate conjure up
+%% a correct 64-byte header out of implicit zero-fill.
```
lhoguin: tl;dr

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1520,13 +1651,28 @@ mark_handle_closed(FileHandlesEts, File, Ref) ->
 
 form_filename(Dir, Name) -> filename:join(Dir, Name).
 
-filenum_to_name(File) -> integer_to_list(File) ++ ?FILE_EXTENSION.
+%% Format is determined solely by the file number relative to the
+%% store's last_v1_file boundary: writes always produce v2 from this
+%% release onward, so any file above the boundary (or every file, when
+%% the store has no v1 files at all) is v2.
```
lhoguin: tl;dr

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1600,28 +1762,249 @@ scan_data(<<Size:64, MsgIdInt:128, _Rest/bits>> = Data, Fd, Fun, Offset, FileSiz
...
+scan_v2_file_for_valid_messages(Path, Fun) ->
+    try
+        case file:open(Path, [read, binary, raw]) of
+            {ok, Fd} ->
+                %% The fd must close on every exit from the scan below,
+                %% not just success: scan_file_recovering_corruption/2
+                %% retries a failed scan of this same file in a loop,
+                %% and each retry calls back in here, so a leaked fd
+                %% per failed attempt would accumulate for as long as
+                %% the process doing the scanning lives.
```
lhoguin: Delete

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1600,28 +1762,249 @@ scan_data(<<Size:64, MsgIdInt:128, _Rest/bits>> = Data, Fd, Fun, Offset, FileSiz
...
+        false ->
+            %% MsgId (16 bytes) is included below; this is the body only.
+            PayloadLen = Size - 21,
+            case Rest0 of
+                <<MsgIdInt:128, Body:PayloadLen/binary, Rest/bits>> ->
+                    %% A torn write reproducing a syntactically valid
+                    %% Erlang external term by chance is astronomically
+                    %% unlikely, so this catches a Type/Size that happened
+                    %% to validate numerically but isn't really a message,
+                    %% regardless of what MsgId lookup would say below.
```
lhoguin: Claim too strong.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1651,7 +2034,12 @@ index_new(Dir) ->
 
 index_recover(Dir) ->
     Path = filename:join(Dir, ?INDEX_FILE_NAME),
-    case ets:file2tab(Path) of
+    %% {verify, true} rejects a truncated dump (e.g. left behind by a
+    %% shutdown that was killed mid-write) instead of silently loading
+    %% however many objects made it to disk. The caller already falls
+    %% back to a full dirty-recovery rescan on any error here, which is
+    %% what must happen instead of trusting a partial index.
```
lhoguin: tl;dr

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1685,8 +2073,27 @@ index_update(IndexEts, Obj) ->
...
+%% compaction pass and must not be touched -- its own physical bytes were
+%% already copied into the new location regardless (write_compact_file_v2
+%% doesn't know or care whether the index update will succeed), so the
+%% only consequence of skipping is an orphaned, harmless duplicate copy
+%% that scan_and_vacuum_message_file/3's existing fan-out clause already
+%% recognises and discards on this file's next scan.
+index_update_offset_if_unchanged(IndexEts, MsgId, File, OldOffset, NewOffset) ->
+    %% The extra {...} around the replacement record is required by
+    %% ets:select_replace/2: without it, a literal tuple appearing in a
+    %% match spec's body is parsed as a function call instead.
```
lhoguin: I know.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1685,8 +2073,27 @@ index_update(IndexEts, Obj) ->
...
+%% i.e. it's still the exact copy this compaction pass moved. A fan-out
+%% write racing this compaction can delete and reinsert the same msg_id
+%% under a different file (or a different offset in the same file) before
+%% this runs; if that happened, the entry no longer belongs to this
+%% compaction pass and must not be touched -- its own physical bytes were
+%% already copied into the new location regardless (write_compact_file_v2
+%% doesn't know or care whether the index update will succeed), so the
+%% only consequence of skipping is an orphaned, harmless duplicate copy
+%% that scan_and_vacuum_message_file/3's existing fan-out clause already
+%% recognises and discards on this file's next scan.
```
lhoguin: tl;dr

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1726,33 +2133,51 @@ index_terminate(IndexEts, Dir) ->
...
+                                 {true, IndexEts, ClientRefs, LastV1File};
                              {error, Error} ->
                                  Fresh("failed to recover index: ~tp", [Error])
                          end;
                 false -> Fresh("recovery terms differ from present", [])
             end
     end.
 
+%% Ground truth for the v1/v2 boundary when it cannot be trusted from
+%% persisted terms (dirty recovery).
```
lhoguin: Pointless

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1778,7 +2203,14 @@ recover_file_summary(false, _Dir) ->
...
 recover_file_summary(true, Dir) ->
     Path = filename:join(Dir, ?FILE_SUMMARY_FILENAME),
-    case ets:file2tab(Path) of
+    %% {verify, true} rejects a truncated dump instead of silently
+    %% loading whatever entries made it to disk -- a partial table can
+    %% be missing exactly the highest-numbered (most recently added)
+    %% files, which would otherwise make a stale, lower file look like
+    %% the current one. Falling back to recover_file_summary(false, ...)
+    %% forces the full dirty-recovery rescan, which derives the true
+    %% current file from the real directory listing instead.
```
lhoguin: tl;dr

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1863,8 +2320,75 @@ build_index_worker(Gatherer, #msstate { index_ets = IndexEts, dir = Dir },
...
+%% updated the index for every message reached before the error, so
+%% guessing wrong here doesn't just lose the file's own data: it
+%% leaves those index entries pointing at a file this then discards,
+%% turning a later, unrelated read of one of those messages into a
+%% crash instead of a clean "not found". v1 has the exact same
+%% principle already: corruption there is handled entirely by
+%% scan_v1_next_byte/7 silently shifting forward a byte at a time,
+%% never raising, so nothing ever reaches build_index_worker for a v1
+%% file except an equally unanticipated error, and it is scanned
+%% directly, unwrapped, for the same reason.
```
lhoguin, line 2314: tl;dr

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1863,8 +2320,75 @@ build_index_worker(Gatherer, #msstate { index_ets = IndexEts, dir = Dir },
...
+            end
+    end.
+
+%% Every rabbit_msg_store_v2_scan reason (wrapped by
+%% scan_v2_file_for_valid_messages/2 in rabbit_msg_store_v2_scan_error)
+%% reports the offset of the record it rejected as its second element.
+%% Only that specific, unambiguous shape is trusted: anything else,
+%% including a tuple that merely happens to also have an integer in
+%% the same position, is not our own error and must not be mistaken
+%% for one.
```
lhoguin: tl;dr

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1863,8 +2320,75 @@ build_index_worker(Gatherer, #msstate { index_ets = IndexEts, dir = Dir },
...
+corruption_offset({rabbit_msg_store_v2_scan_error, _Path, InnerReason})
+        when is_tuple(InnerReason), tuple_size(InnerReason) >= 2,
+             element(1, InnerReason) =:= rabbit_msg_store_v2_scan,
+             is_integer(element(2, InnerReason)) ->
+    {ok, element(2, InnerReason)};
+corruption_offset(_Reason) -> error.
+
+%% Runs in the same process as its caller, rebuild_index/3, rather than
+%% a separate spawned process -- see the comment there for why.
```
lhoguin: Don't care

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1893,16 +2417,27 @@ reduce_index(Gatherer, LastFile,
...
+%% empty and reduce_index/3 would block forever; rabbit_msg_store's own
+%% start_link/5 uses {timeout, infinity}, so nothing would ever time
+%% that out either -- the store just never came up, with no crash and
+%% no log. worker_pool:dispatch_sync/1 paces on free-worker availability
+%% either way, identically whether it's called from a separate process
+%% or inline, so doing it inline here doesn't change how long dispatching
+%% every file takes; it just means a failure part-way through (e.g. the
+%% shared worker_pool itself dying while a dispatch call is in flight)
+%% now propagates directly out of this function instead of vanishing
+%% silently.
```
lhoguin: Don't care, commit message enough

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2011,33 +2546,46 @@ compact_file(File, FileSize,
...
     %% be there for readers. Note that it's OK if we crash at any point before we
     %% update the index because the old data is still there until we truncate.
-    lists:foreach(fun ({UpdateMsgId, UpdateOffset}) ->
-        ok = index_update_fields(IndexEts, UpdateMsgId,
-                                 [{#msg_location.offset, UpdateOffset}])
+    %%
+    %% The update only takes effect if the entry still matches File and its
+    %% original offset: a fan-out write can race this compaction and rewrite
+    %% the same msg_id under a different file (or offset) before we get here,
+    %% and blindly overwriting that live entry's offset would corrupt it.
```
lhoguin, line 2525: Add note that the message in question in that case is no longer referred to in this compacted file and the data will become a hole to fill in next compaction.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2108,33 +2657,185 @@ do_compact_file(Fd, Offset, Start = [#msg_location{ offset = StartMsgOffset }|_]
...
+%% torn partway through is, if anything, more likely to leave old
+%% bytes behind unzeroed than to leave zeros. What a scanner relies on
+%% after a crash is the header's own type and size landing intact,
+%% which zeroing the payload does nothing for.
+%%
+%% It does have one incidental benefit for a *different* failure mode,
+%% though: REC_RESERVED is 0 and has no matching scan clause, so if a
+%% scan ever gets misaligned (say, from a corrupted Size elsewhere) and
+%% lands inside a hole's zeroed payload, it hits the unrecognised-type
+%% crash instead of silently reading zeros as if they were meaningful.
```
lhoguin: tl;dr

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2108,33 +2657,185 @@ do_compact_file(Fd, Offset, Start = [#msg_location{ offset = StartMsgOffset }|_]
...
+%% information that a later part of the plan could still
+%% invalidate. write_compact_file_v2/2 writes a hole record for
+%% whatever's left of the space each run of moves lands in, so the
+%% file is always fully valid and scannable at every point during
+%% compaction and before the physical truncation that follows it has
+%% actually happened: a crash never leaves an unmarked, ambiguous gap
+%% for the next scan (this one, recovery, or the salvage tool) to
+%% trip over. Messages we leave in place are untouched, including
+%% whatever follows them: we haven't broken anything there, so there
+%% is nothing for us to mark.
```
lhoguin: tl;dr

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2108,33 +2657,185 @@ do_compact_file(Fd, Offset, Start = [#msg_location{ offset = StartMsgOffset }|_]
...
+%% that a later decision hasn't been made yet when it happens.
+%%
+%% A move's own hole size isn't known until we see what ends up right
+%% after it in the finished plan: packing can stop either because this
+%% was the very last message that could possibly move (the terminal
+%% clause below), or because the next candidate from the end of the
+%% file simply doesn't fit and we fall through to leaving a message
+%% where it is instead. Either way leaves a leftover behind this move,
+%% so the hole size is filled in from the finished plan, in
+%% write_compact_file_v2/2, rather than decided here.
```
lhoguin: Very confusing

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2108,33 +2657,185 @@ do_compact_file(Fd, Offset, Start = [#msg_location{ offset = StartMsgOffset }|_]
...
+%% valid record (every source at or beyond the terminal move's own
+%% source offset belongs to a message this pass moved) is itself a
+%% real, if now-superseded, record; scan_v2_data's MsgIdsFound
+%% deduplication resolves the resulting duplicates by keeping the
+%% earlier (moved) copy, and physical truncation removes them for
+%% good once it eventually runs. A kept message is left untouched
+%% entirely: we never overwrote anything around it, so there is
+%% nothing there for us to mark. Both tuple shapes carry their offset
+%% in the same, second position, so the next entry's offset can be
+%% read with element/2 without matching out which shape it is.
```
lhoguin, line 2707: tl;dr

lhoguin (commented):
Mostly oversized comments.

deps/rabbit/src/rabbit_msg_store.erl
lhoguin: No let's keep it whole.

deps/rabbit/src/rabbit_msg_store.erl
lhoguin: Not necessary.

deps/rabbit/src/rabbit_msg_store.erl
lhoguin: A bit verbose but let's keep.

lhoguin:
@gomoripeti Hello I can't assign you as reviewer for reasons unknown but I'd appreciate you taking a look as well. Thanks!

lukebakken (commented):
> [!NOTE]
> This review was drafted by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who guided the analysis (including requiring the Erlang/OTP behavior to be verified against the OTP source, and steering the scope down to these comment-accuracy notes) and reviewed it before filing.

Read through the v2 scan/recovery, compaction and reader paths. No correctness concerns: the design choices (never byte-shift past a v2 record, crash rather than self-heal during compaction of a possibly-referenced file, positional access to the tagged scan/plan tuples) are deliberate and well documented at each site. Two small comment-accuracy notes inline.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1600,28 +1750,239 @@ scan_data(<<Size:64, MsgIdInt:128, _Rest/bits>> = Data, Fd, Fun, Offset, FileSiz
...
+%% coincidence. We tighten this with two invariants compaction never
+%% violates, so any violation is treated as corruption rather than
+%% something to recover from: a hole (of either kind) is never
+%% immediately followed by another hole, and a SMALL_HOLE run is never
+%% 5 bytes or longer. A MESSAGE record's body must additionally decode
+%% via binary_to_term/1, which a torn write is astronomically unlikely
+%% to produce by chance.
+%%
+%% A record whose Size does not fit within the file is treated as a
+%% torn last write and the remainder of the file is discarded, exactly
```
lukebakken: Comment-precision note, not the behavior (discard-on-overshoot is the right call): v1 does not actually discard here. Its catch-all byte-shifts one byte at a time (scan_v1_next_byte/7) trying to re-sync, and only ends up discarding because a genuine torn tail has nothing valid left to find. So v1 and v2 match by result only for a real torn last write; for a mid-file Size corruption with valid records after it, v1 can resync onto them while v2 returns Acc and drops them. "exactly as in v1" reads as if v2 keeps v1's resync behavior, which it deliberately does not.

deps/rabbit/test/backing_queue_SUITE.erl
```diff
@@ -378,6 +404,1632 @@ msg_store_read_many_fanout1(_Config) ->
...
+%% window) must not trip over whatever is physically sitting between
+%% TruncateSize and the old end of file. This constructs a plan whose
+%% last move leaves real leftover space behind it (as opposed to
+%% msg_store_compaction_v2_exact_fit, which deliberately leaves none),
+%% mocks the truncate away entirely, and scans the file while it is
+%% still at its original, untruncated size.
+%%
+%% The leftover space is deliberately *not* the last moved message's
+%% own source copy: at the point this write happens the index still
+%% points there (index_update_fields/3 only runs later, once every
```
lukebakken: Stale name: index_update_fields/3 does not exist. The function is index_update_offset_if_unchanged/5, used correctly further down at :676 and in the source comment at rabbit_msg_store.erl:2691.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1863,8 +2282,68 @@ build_index_worker(Gatherer, #msstate { index_ets = IndexEts, dir = Dir },
...
+            end
+    end.
+
+%% Every rabbit_msg_store_v2_scan reason reports the offset of the
+%% record it rejected as its second element.
+corruption_offset({rabbit_msg_store_v2_scan_error, _Path, InnerReason})
+        when is_tuple(InnerReason), tuple_size(InnerReason) >= 2,
+             element(1, InnerReason) =:= rabbit_msg_store_v2_scan,
+             is_integer(element(2, InnerReason)) ->
+    {ok, element(2, InnerReason)};
```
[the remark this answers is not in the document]
lhoguin: Disagree. Before a new record tag actually exists, it is indistinguishable from garbage. We also do not support downgrading. So the right approach is to consider eventual new tags as invalid.

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -2108,33 +2596,151 @@ do_compact_file(Fd, Offset, Start = [#msg_location{ offset = StartMsgOffset }|_]
...
+%% The scanner only ever looks at the HOLE record's own 5-byte header
+%% to know how much to skip; it never reads the inner bytes. We zero
+%% them anyway so that a completed write leaves no message bytes
+%% behind for a raw scan (a salvage tool, for instance) to recover.
+hole_bytes(0) ->
+    <<>>;
+hole_bytes(Size) when Size < 5 ->
+    binary:copy(<<?REC_SMALL_HOLE:8>>, Size);
+hole_bytes(Size) ->
+    <<?REC_HOLE:8, Size:32, 0:(Size - 5)/unit:8>>.
```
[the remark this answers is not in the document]
lhoguin: The v1 cost comes from byte by byte scanning. Zero filling helps avoid issues and is not a big cost as only small regions typically get filled.

deps/rabbit/src/rabbit_msg_store.erl
lhoguin: Dropping "exactly as in v1".

deps/rabbit/test/backing_queue_SUITE.erl
lhoguin: Done.

lhoguin:
Merging this now. Thanks for reviews!

deps/rabbit/test/backing_queue_SUITE.erl
```diff
@@ -378,6 +404,1632 @@ msg_store_read_many_fanout1(_Config) ->
...
+    %% scan_and_vacuum_message_file dispatching to their v1 code paths,
+    %% not straight to deletion.
+    StorePid = rabbit_vhost_msg_store:vhost_store_pid(?VHOST, ?PERSISTENT_MSG_STORE),
+    GCPid = rabbit_msg_store:gc_pid(StorePid),
+    Path = msg_store_file_path(?VHOST, ?PERSISTENT_MSG_STORE, "0.rdq"),
+    SizeBeforeCompact = filelib:file_size(Path),
+    %% count_msg_refs registers refs via ets:update_counter/4, whose
+    %% Default already carries ref_count=1; on a fresh key the Increment
+    %% is added on top of that, so a message recovered this way starts
+    %% at ref_count=2 and needs two removes to be fully dereferenced.
```
gomoripeti, line 1114: what is the intention with this double-counting behaviour? As far as I understand this is independent from file format, so affects v2 too, and it can create an entry/file leak after unclean shutdowns.

I prepared a commit (https://github.com/cloudamqp/rabbitmq-server/commit/fa8c92d2993f5826f67dfa76787952647f704f87) to address this behaviour but then I got unsure if this buggy behaviour needs to be kept, maybe there is something that relies on it?

deps/rabbit/src/rabbit_msg_store.erl
```diff
@@ -1600,28 +1750,239 @@ scan_data(<<Size:64, MsgIdInt:128, _Rest/bits>> = Data, Fd, Fun, Offset, FileSiz
...
+                                %% superseded elsewhere) is still known-good
+                                %% data once Type/Size validate, not
+                                %% suspected corruption.
+                                try_next ->
+                                    scan_v2_data(Rest, Fd, Fun, Offset + Size, FileSize,
+                                                 MsgIdsFound#{MsgIdInt => true}, Acc)
+                            end
+                    end;
+                _ ->
+                    %% Not enough buffered yet; read more.
```
gomoripeti, line 1974: there was a small optimisation in 3.13 that for large messages (compared to SCAN_BLOCK_SIZE) the whole message body is read from file instead of reading in blocks and then appending (https://github.com/rabbitmq/rabbitmq-server/pull/11248)

That was made unnecessary in 4.x because large messages went to their own file (and their body was entirely skipped during scanning) https://github.com/rabbitmq/rabbitmq-server/pull/13250

But with v2 the body needs to be read from disk for term_to_binary check, so the 3.13 micro-optimisation could still make sense

An AI perf test from today

```
┌──────────────────────┬─────────────────┬───────────────┬──────────────────┬─────────────────┬───────┬─────────┐
│        build         │ 64 MB msg: time │ peak binaries │ binaries created │ 16 MB msg: time │ peak  │ created │
├──────────────────────┼─────────────────┼───────────────┼──────────────────┼─────────────────┼───────┼─────────┤
│ 3.13 before PR 11248 │ 67 ms           │ 141 MB        │ 1426 MB          │ 8.0 ms          │ 48 MB │ 179 MB  │
├──────────────────────┼─────────────────┼───────────────┼──────────────────┼─────────────────┼───────┼─────────┤
│ 3.13 after PR 11248  │ 48 ms           │ 128 MB        │ 393 MB           │ 6.0 ms          │ 32 MB │ 95 MB   │
├──────────────────────┼─────────────────┼───────────────┼──────────────────┼─────────────────┼───────┼─────────┤
│ 4.x after PR 13250   │ 0.4 ms          │ 4 MB          │ 8 MB             │ 0.8 ms          │ 4 MB  │ 8 MB    │
├──────────────────────┼─────────────────┼───────────────┼──────────────────┼─────────────────┼───────┼─────────┤
│ current v2 scanner   │ 61 ms           │ 141 MB        │ 1554 MB          │ 11 ms           │ 48 MB │ 195 MB  │
└──────────────────────┴─────────────────┴───────────────┴──────────────────┴─────────────────┴───────┴─────────┘
```

So the saving of PR 11248 is mostly in total allocation and garbage, not in peak memory.

lhoguin: The term_to_binary check is there to better detect problems as v2 starts getting used. I think we can drop it later on, both for normal and large message files. The more use, the more confident we can be and remove the check.

deps/rabbit/test/backing_queue_SUITE.erl
lhoguin: Good catch, I don't think anything relies on it, please send a PR.

gomoripeti: opened https://github.com/rabbitmq/rabbitmq-server/pull/17351

PR #17168, opened 2026-08-12
Selenium suite fixes

michaelklishin:
With these assorted fixes all Selenium suites pass five times in a row without a single failure on an x86 Linux host.

### DIFF
--- a/selenium/bin/gen-env-file
+++ b/selenium/bin/gen-env-file
@@ -53,7 +53,7 @@ generate_env_file() {
     FILE_ARRAY_LENGTH=${#FILE_ARRAY[@]}
   done
   rm -r $TMP_ENV_FILE
-  tail +3 $ENV_FILE > "$ENV_FILE.tmp" && mv "$ENV_FILE.tmp" $ENV_FILE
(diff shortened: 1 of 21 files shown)


selenium/bin/gen-env-file
```diff
@@ -53,7 +53,7 @@ generate_env_file() {
     FILE_ARRAY_LENGTH=${#FILE_ARRAY[@]}
   done
   rm -r $TMP_ENV_FILE
-  tail +3 $ENV_FILE > "$ENV_FILE.tmp" && mv "$ENV_FILE.tmp" $ENV_FILE
+  tail -n +3 $ENV_FILE > "$ENV_FILE.tmp" && mv "$ENV_FILE.tmp" $ENV_FILE
```
michaelklishin, line 56: The `-n` syntax is modern, the position argument one is old.

Ubuntu 22.04's coreutils still support the old syntax. Fedora 44 one's, however, does not and fails.

Therefore this silly looking change is a matter of compatibility with modern distributions and not a drive-by cosmetic tweak.

michaelklishin:
@MarcialRosales I'd like to get this in before the next patch release. We can always revisit these changes as you see fit.

PR #17177, opened 2026-08-13
Management UI: do not coerce x-argument values to strings

michaelklishin:
A note on XSS safety: the key is escaped as before, and the value is escaped by `fmt_amqp_value` (depending on the value type).

In other words, the issue in #17176 was a double encoding one. And with this change we make
the value encoding type-aware, as it should be.

References #17176.

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/formatters.js
+++ b/deps/rabbitmq_management/priv/www/js/formatters.js
@@ -183,7 +183,10 @@ function args_to_features(obj) {
     var res = {};
     for (var k in obj.arguments) {
         if (k in KNOWN_ARGS) {
-            res[k] = fmt_escape_feature(obj.arguments[k]);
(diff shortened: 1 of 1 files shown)


deps/rabbitmq_management/priv/www/js/formatters.js
LincolnPuzey: Known args (like `x-max-length` in my example) go through this `if (k in KNOWN_ARGS)` branch. I think there needs to be a similar fix on this line.

LincolnPuzey (commented):
Thanks for opening a PR - I tested locally, found one thing.

michaelklishin:
@LincolnPuzey thank you, excellent catch. The PR is ready for another round.

LincolnPuzey:
@michaelklishin No worries, tested again, looks good to me

PR #17198, opened 2026-08-17
[mgt] Gracefully degrade list channels when stats are disabled

MarcialRosales:
## Proposed Changes

- Management UI: The Channels list and individual channel pages now gracefully degrade to show basic information when management stats are disabled, rather than returning a 400 Bad Request (#8240).
- Testing: Expanded Selenium test coverage for UI section visibility across various pages, user permissions, queue types, and stats configurations.

## Types of Changes

- [x] Build system and/or CI


michaelklishin (approved):

selenium/test/queuesAndStreams/with-disabled-stats/display-basic-stats.js
```diff
@@ -63,9 +63,10 @@ describe('Should display basic stats even when stats are disabled', function ()
       await overview.clickOnQueuesTab()
       await queuesAndStreams.ensureAddQueueSectionIsVisible()
       await queuesAndStreams.fillInAddNewQueue({"name" : queueName, "type" : "classic"})
-      await delay(5000)
```
michaelklishin, line 66: @MarcialRosales FYI, we should avoid sleeps in most cases (unless we want N seconds to pass e.g. for expiration checks), we have `doUntil` that mimics RabbitMQ's `await_condition/2` and other eventual matchers.

MarcialRosales: Absolutely right. I introduced the construct `doUntil` exactly for these situations. To avoid having a delay. Thanks for spotting it. I am retrofitting doUntil on a separate PR.

PR #9141, opened 2023-08-18
Native STOMP

ikavgo:
**NOTE**: Bazel tests pass locally, I don't like the code yet though. 
**TODO**: Double check Management UI correctly shows Connection information.

## Proposed Changes

Don't use AMQP 0-9-1 as backend for STOMP.

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [x] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
If you're unsure about any of them, don't hesitate to ask on the mailing list.
We're here to help!
This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
- [x] If relevant, I have added necessary documentation to https://github.com/rabbitmq/rabbitmq-website
- [ ] If relevant, I have added this change to the first version(s) in release-notes that I expect to introduce it

## Further Comments

Documentation update: https://github.com/rabbitmq/rabbitmq-website/pull/1713

PR #17202, opened 2026-08-18
STOMP: authorize durable unsubscribe cleanup and preserve delivery settlement

brianr2600:
## Proposed Changes

Durable STOMP topic subscriptions are backed by queues. The native `UNSUBSCRIBE` path currently reconstructs the queue name from the `UNSUBSCRIBE` frame and invokes a plugin-local deletion helper without checking `configure` permission. Because `x-queue-name` is client controlled, the deletion target can differ from the queue associated with the subscription.

This change takes the queue resource from server-side subscription state, checks `configure` permission immediately before deletion, and uses `rabbit_amqqueue:delete_with/6` so the standard internal and exclusive-access checks apply.

It also resolves `ACK` and `NACK` from the pending-delivery ledger rather than the active subscription table. This allows a delivery to be settled after its subscription has been removed, preserves the acknowledgement mode in force when the delivery was sent, and keeps transaction settlement idempotent when a cumulative acknowledgement covers a later queued action.

## Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation improvements
- [ ] Cosmetic change
- [ ] Build system and/or CI

## Testing

- `gmake -C deps/rabbitmq_stomp`
- `gmake -C deps/rabbitmq_stomp xref`
- `gmake -C deps/rabbitmq_stomp ct-system` (84 cases)
- `gmake -C deps/rabbitmq_web_stomp ct-unsubscribe_authz` (2 cases)

## Checklist

- [x] **Mandatory**: I (or my employer/client) have signed the CA
- [x] I have added tests that prove the fixes are effective
- [x] All listed tests pass locally with these changes
- [x] No documentation change is required because the protocol surface is unchanged
- [ ] Release-note handling can follow maintainer guidance

### DIFF
--- a/deps/rabbitmq_stomp/src/rabbit_stomp_processor.erl
+++ b/deps/rabbitmq_stomp/src/rabbit_stomp_processor.erl
@@ -673,11 +673,13 @@ tidy_canceled_subscription(ConsumerTag, Subscription,
 
 %% Client-initiated cancelations will pass an actual frame
 
-tidy_canceled_subscription(ConsumerTag, Subscription = #subscription{dest_hdr = DestHdr},
+tidy_canceled_subscription(ConsumerTag,
+                           Subscription = #subscription{dest_hdr = DestHdr,
+                                                        queue_name = QRes},
                            Frame, State0) ->
     {ok, State1} = tidy_canceled_subscription_state(ConsumerTag, Subscription, State0),
     {ok, Dest} = parse_endpoint(DestHdr),
-    maybe_delete_durable_sub_queue(Dest, Frame, State1).
+    maybe_delete_durable_sub_queue(Dest, QRes, Frame, State1).
 
 tidy_canceled_subscription_state(ConsumerTag,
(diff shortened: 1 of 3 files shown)


michaelklishin:
@brianr2600 ack (no pun intended). Thank you.

This is a non-trivial change that will take some time to review.

michaelklishin:
The general approach is reasonable but some of these LLM-generated comments are not just hard to read, they are straight up confusing and invent terminology neither the STOMP spec nor this plugin really uses.

michaelklishin:
The new tests are extensive enough to spot a subtle deviation from [the spec](https://stomp.github.io/stomp-specification-1.2.html#SUBSCRIBE_ack_Header)
introduced in 300dcee219 #9141.

Multiple acks in the spec must be scoped per subscription but in our case, they are actually per connection. This is fine in practice and not something this PR introduces.

michaelklishin:
Moved to https://github.com/rabbitmq/rabbitmq-server/pull/17208.

PR #17213, opened 2026-08-19
Support customize_hostname_check in AMQP URI query parameters

jiangranwang:
## Proposed Changes

This PR adds support for specifying `customize_hostname_check` query parameter in AMQP URI strings, by extending  `amqp_uri:parse/{1,2}`. Setting this parameter to `https` with `&customize_hostname_check=https` will configure the TLS option `match_fun` to `public_key:pkix_verify_hostname_match_fun(https)` (which cannot be directly specified as part of the AMQP URI string), allowing wildcard certificates to be matched using the same hostname rules as HTTPS clients. 
Ref: [https://www.erlang.org/doc/apps/ssl/ssl.html#t:client_option_cert/0](https://www.erlang.org/doc/apps/ssl/ssl.html#t:client_option_cert/0:~:text=%7B-,customize_hostname_check,-%2C%20%5B%7B)

Without this, a client connecting to a server presenting a valid wildcard certificate (e.g. `*.example.com`) fails hostname verification during the TLS handshake, since the default match function doesn't apply wildcard rules, as follows:

```
2026-08-10 20:38:22.770920+00:00 [error] <0.3977248.0> Shovel 'xxxxxxxxx' failed to connect (URI: amqps://xxxxxxxxx): {tls_alert,{handshake_failure,"TLS client: In state certify at ssl_handshake.erl:2140 generated CLIENT ALERT: Fatal - Handshake Failure\n {bad_cert,hostname_check_failed}"}}
```

Also fixes the doc for `parse` by adding the missing SSL query parameter `server_name_indication`.

## Types of Changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes issue #NNNN)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply.
You can also fill these out after creating the PR.
This is simply a reminder of what we are going to look for before merging your code._

- [x] **Mandatory**: I (or my employer/client) have have signed the CA (see https://github.com/rabbitmq/cla)
- [x] I have read the `CONTRIBUTING.md` document
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] All tests pass locally with my changes
[...]

PR #17221, opened 2026-08-21
By @jiangranwang: #17213 Support customize_hostname_check in AMQP URI query parameters (e.g. for shovels)

michaelklishin:
This is #17213 by @jiangranwang with an addition from me: there is no reason not to support this new key in the AMQP 1.0 client, for shovel's needs alone.

Closes #17213.


michaelklishin:
Note: if a custom `verify_fun` is set (e.g. via `amqp_client.ssl_options` app env, so, noide-wide), Erlang's `ssl` app skips its default hostname check path entirely, so `customize_hostname_check`'s `match_fun` is never invoked.

That's apparently how OTP's `ssl_handshake:validation_fun_and_state/4` works. And it means the two options can't be combined: whichever `verify_fun` you supply takes over peer verification handling and the hostname match part does nothing without any log traces. 

We should probably consider documenting this, even if it's really inside baseball.

PR #17227, opened 2026-08-22
Remove dead fallback clause in plain auth mechanism (Fixes #16259)

### LINKED ISSUE #16259: `rabbit_auth_mechanism_plain:handle_response/2` fallback clause appears to be dead code
## What

`rabbit_auth_mechanism_plain:handle_response/2` has a fallback clause for the case where the passed state is not a `#state{}` record:

```erlang
handle_response(Response, #state{socket = Socket}) ->
    ...;
handle_response(Response, _State) ->
    case extract_user_pass(Response) of
        {ok, User, Pass} ->
            rabbit_access_control:check_user_pass_login(User, Pass);
        error ->
            {protocol_error, "response ~tp invalid", [Response]}
    end.
```

This fallback clause appears to be dead code under normal operation.

## Why it looks dead

`init/1` has a single clause that unconditionally returns `#state{socket = Sock}`:

```erlang
init(Sock) ->
    #state{socket = Sock}.
```

Every caller I checked follows the usual `AuthMechanism:init(Sock)` followed by `AuthMechanism:handle_response(Response, AuthState)` cycle, where `AuthState` comes from `init/1` and is therefore always a `#state{}` record:

- `rabbit_reader:auth_phase/2` (AMQP 0-9-1)
- `rabbit_amqp_reader:handle_sasl_frame/2` (AMQP 1.0)
- `rabbit_stream_reader:handle_frame_pre_auth/4` (RabbitMQ Stream Protocol)

A grep of the code base finds no call to `rabbit_auth_mechanism_plain:handle_response/2` outside this pattern.

## Origin

The fallback was added by 3bcdc0f3596ba4a037f9fa6da77a3247b3218f84 as part of #13818 ("Fallback to original implementation of plain auth_mechanism if socket is not provided"). I cannot see how the `_State` clause is reached under the invariants the readers maintain today.

## Ask

Decide which of the following applies and act accordingly:

1. The fallback is genuinely defence in depth for a code path I have not found. In that case, add a short comment pointing to the call site that can invoke `handle_response/2` without going through `init/1`, and add a test exercising it so it does not become silently unreachable.
2. The fallback is dead code that can be removed without loss of behaviour, simplifying the module.

This came up as a secondary finding while investigating #16255 and its fix (PR #16258).

Vishal-770:
## Proposed Changes

This PR removes the `handle_response(Response, _State)` fallback clause in `rabbit_auth_mechanism_plain.erl`. 
As described in issue #16259, this fallback clause is dead code. The `init/1` function in the module unconditionally returns a `#state{socket = Sock}` record. Every caller of `handle_response/2` gets its state from `init/1`, meaning the state will always match the `#state{socket = Socket}` clause, rendering the `_State` fallback unreachable. 

Removing this simplifies the code and avoids maintaining dead paths.

Fixes #16259

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments
This was a straightforward dead code removal requested in issue #16259.

### DIFF
--- a/deps/rabbit/src/rabbit_auth_mechanism_plain.erl
+++ b/deps/rabbit/src/rabbit_auth_mechanism_plain.erl
@@ -40,14 +40,6 @@ handle_response(Response, #state{socket = Socket}) ->
             rabbit_access_control:check_user_login(User, AuthProps);
         error ->
             {protocol_error, "response ~tp invalid", [Response]}
-    end;
-
-handle_response(Response, _State) ->
-    case extract_user_pass(Response) of
-        {ok, User, Pass} ->
-            rabbit_access_control:check_user_pass_login(User, Pass);
-        error ->
-            {protocol_error, "response ~tp invalid", [Response]}
     end.
 
 
--- a/deps/rabbit/src/rabbit_auth_mechanism_plain.erl
+++ b/deps/rabbit/src/rabbit_auth_mechanism_plain.erl
@@ -40,6 +40,19 @@ handle_response(Response, #state{socket = Socket}) ->
             rabbit_access_control:check_user_login(User, AuthProps);
         error ->
             {protocol_error, "response ~tp invalid", [Response]}
+    end;
+
+%% This clause is reached during the RabbitMQ Stream Protocol's
+%% update-secret feature. In `rabbit_stream_reader:handle_frame_post_auth/4`,
+%% a new `sasl_authenticate` frame is received. At this point, the
+%% `authentication_state` has been set to `done`, so this fallback clause
+%% matches and performs the credentials check.
+handle_response(Response, _State) ->
+    case extract_user_pass(Response) of
+        {ok, User, Pass} ->
+            rabbit_access_control:check_user_pass_login(User, Pass);
+        error ->
+            {protocol_error, "response ~tp invalid", [Response]}
     end.
 
 


michaelklishin:
@Vishal-770 thanks. As a first time contributor, can you please sign [our CLA](https://github.com/rabbitmq/cla)? It's done online, you just need to reach out to our team with your email.

So it shouldn't be a big deal even for such a small cosmetic fix.

Thanks.

Vishal-770:
Thanks @michaelklishin !   I just sent the email to request the CLA document. I will let you know as soon as I receive and sign it.

michaelklishin:
@Vishal-770 the CI failure seems to be repeatable (and not present in `main`). Are you absolutely certain that this is dead code?

Vishal-770:
@michaelklishin you were right to question it—I just looked into why the stream tests were failing and it turns out the _State clause isn't dead code after all.

During a stream protocol update-secret operation, rabbit_stream_reader:handle_frame_post_auth/4 handles the sasl_authenticate frame. Since the client is already authenticated, the stream reader's authentication_state is set to done.

When it calls handle_response, it passes done instead of the #state{} record. So the stream protocol actually relies on that _State fallback clause. Removing it was throwing a function_clause exception and hard-closing the socket, breaking the tests.

I've reverted the deletion and added a detailed comment above the clause for the future, explaining its use in the stream protocol so it doesn't get flagged as dead code again

PR #17232, opened 2026-08-24
Shovel: add destination byte-rate limiting

jgorea:
## Summary
This PR introduces configurable destination byte-rate limiting for Shovel forwarding.

## Motivation
Shovels running over constrained links can saturate bandwidth when forwarding is unthrottled. This change adds an explicit bytes-per-second cap so operators can control transfer rate per shovel instead of throttling the entire source <-> destination connection.
Using tools like _tc_  (traffic control) to achieve per shovel byte-rate limiting is impractical when dealing with multiple shovels declared for the same source / destination / port.

## What Changed
- Added a token-bucket rate limiter module for Shovel forwarding
- Added destination rate-limit configuration support for:
  - AMQP 0.9.1 Shovel
  - AMQP 1.0 Shovel
  - Local Shovel
- Added management UI fields for dynamic shovel configuration
- Added management UI display for the configured limit
- Added release note entry for the feature

## Configuration
- Static shovels:
  - `max_bytes_per_second`
- Dynamic shovels:
  - `dest-max-bytes-per-second`

If unset, the limit is treated as unlimited.

## Behavior Notes
- Rate limiting is applied to message bytes forwarded to destination
- Forwarding order is preserved while throttling
- Overlap between rate-limit delays and other blocking signals is handled so pending head messages are not charged twice
- Dynamic config values are normalized defensively before limiter initialisation to avoid startup failures from malformed persisted values

## Compatibility
- Default behavior is unchanged when no rate limit is configured
- Existing shovels continue to operate without configuration updates

## Validation
- Verified successful compilation/diagnostics for touched Shovel modules
- Verified forwarding path behavior under:
  - Normal forwarding
  - Active throttling
  - Concurrent blocking and unblocking with pending messages
- Verified management UI accepts and displays the new dynamic setting

## Release Notes
- Included feature note in `release-notes/4.4.0.md`

## Checklist
- [x] Feature implementation complete
- [x] UI wiring complete
- [x] Defensive handling for malformed dynamic values
- [x] Release notes updated

### DIFF
--- a/deps/rabbitmq_shovel/src/rabbit_amqp091_shovel.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_amqp091_shovel.erl
@@ -109,13 +109,17 @@ parse(Name, {destination, Dest}) ->
     ATH = proplists:get_value(add_timestamp_header, Dest, false),
     PropsFun1 = add_forward_headers_fun(Name, AFH, PropsFun),
     PropsFun2 = add_timestamp_header_fun(ATH, PropsFun1),
+    MaxBytesPerSecond = parse_parameter(max_bytes_per_second,
+                                        fun parse_positive_integer_or_undefined/1,
+                                        proplists:get_value(max_bytes_per_second, Dest, undefined)),
     #{module => ?MODULE,
       uris => proplists:get_value(uris, Dest),
       resource_decl  => rabbit_shovel_util:decl_fun(?MODULE, {destination, Dest}),
       props_fun => PropsFun2,
       fields_fun => PubFieldsFun,
       add_forward_headers => AFH,
-      add_timestamp_header => ATH}.
+      add_timestamp_header => ATH,
+      max_bytes_per_second => MaxBytesPerSecond}.
 
 parse_source(Def) ->
     SrcURIs  = obfuscated_uris(<<"src-uri">>, Def),
@@ -185,6 +189,7 @@ parse_dest({VHost, Name}, ClusterName, Def, SourceHeaders) ->
                         V =/= none],
     AddHeadersLegacy = pget(<<"add-forward-headers">>, Def, false),
     AddHeaders = pget(<<"dest-add-forward-headers">>, Def, AddHeadersLegacy),
+    MaxBytesPerSecond = pget(<<"dest-max-bytes-per-second">>, Def, undefined),
     Table0 = [{<<"shovelled-by">>, ClusterName},
               {<<"shovel-type">>,  <<"dynamic">>},
               {<<"shovel-name">>,  Name},
@@ -207,7 +212,8 @@ parse_dest({VHost, Name}, ClusterName, Def, SourceHeaders) ->
                  fields_fun => {?MODULE, fields_fun, [X, Key]},
                  props_fun => {?MODULE, props_fun, [Table0, Table2, SetProps,
                                                     AddHeaders, SourceHeaders,
-                                                    AddTimestampHeader]}
+                                                    AddTimestampHeader]},
+                 max_bytes_per_second => MaxBytesPerSecond
                 }, Details).
 
 validate_src(Def) ->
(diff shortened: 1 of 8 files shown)


michaelklishin:
@jgorea we highly recommend starting a discussion before submitting a PR to understand whether the core team is interested in a change or not.

deps/rabbitmq_shovel/src/rabbit_amqp091_shovel.erl
```diff
@@ -465,18 +516,54 @@ handle_dest(#'connection.blocked'{}, State) ->
 
 handle_dest(#'connection.unblocked'{}, State) ->
     State1 = update_blocked_by(connection_blocked, false, State),
-    %% we are unblocked so can begin to forward
-    forward_pending(State1);
+    %% we may be unblocked, but could still be blocked for another
```
michaelklishin, line 519: This comment states the obvious and is not necessary.

deps/rabbitmq_shovel/src/rabbit_amqp091_shovel.erl
```diff
@@ -465,18 +516,54 @@ handle_dest(#'connection.blocked'{}, State) ->
...
+    %% reason (e.g. an outstanding rate-limit delay) -- only resume
+    %% draining the pending queue once nothing is blocking us
+    resume_if_unblocked(State1);
 
 handle_dest({bump_credit, Msg}, State) ->
     credit_flow:handle_bump_msg(Msg),
     State1 = control_throttle(State),
-    %% we have credit so can begin to forward
-    forward_pending(State1);
+    %% we may have credit, but could still be blocked for another
```
michaelklishin, line 527: Ditto, this comment states the obvious and should be deleted.

michaelklishin (changes requested):
The idea may be a reasonable.

Unfortunately the implementation is complex, and because it is bytes/second and not messages/second, I don't see how it can be meaningfully reduced.

Shovels now act as in memory queues with all the complexity and wheel reinvention that come with that.

Some parts of the code reinvent basic Erlang guard functions (`validate_positive_integer/2`). It then duplicates that function between AMQP 0-9-1, AMQP 1.0, local shovels.

michaelklishin:
This PR is a good example of why we ask for a contribution to be discussed first, before any time (or, in this case, seemingly LLM tokens) spent on it.

I've asked for more opinions on the core team about the general idea but this implementation is a bolt-on, not a first class shovel feature.

ansd:
I agree with @michaelklishin to close this PR.

This PR seems to provide some flow control on the number of bytes. For AMQP the receiver already provides flow control for the number of messages how many in can tolerate at any point. There is also flow control on the number of bytes already: https://www.rabbitmq.com/blog/2024/09/02/amqp-flow-control#flow-fields
> Sessions provide a flow control scheme based on the number of transfer frames transmitted. Since frames have a maximum size for a given connection, this provides flow control based on the number of bytes transmitted

I think the proper way should be using AMQP flow control rather than artificially calling `timer:sleep(DelayMs)` in the shovel worker.

jgorea:
> This PR is a good example of why we ask for a contribution to be discussed first, before any time (or, in this case, seemingly LLM tokens) spent on it.
> 
> I've asked for more opinions on the core team about the general idea but this implementation is a bolt-on, not a first class shovel feature.

@michaelklishin Thanks for the feedback, and for asking the core team about the general idea. I am interested in whether they will warm up to the idea eventually.

The intent was to be a bit more than just: "Hey guys, can we please have a byte-rate limiting mechanism for Shovel forwarding?" I added this PoC code that was 100% AI-assisted since I don't speak Erlang.

In previous conversations on this topic, I saw that the suggested solution was to use _tc_, and I wanted to see how hard it would be to bolt on something functional on the Shovel itself.
I understand that in most use cases, putting a hard cap (in bytes/second) wouldn't really make sense, but in our scenario, where network bandwidth is a commodity, it would be very useful to have that option.

michaelklishin:
@jgorea I haven't seen any pushback on the idea, only on the implementation.

PR #17235, opened 2026-08-24
Update Khepri from 0.18.0 to 0.18.1

dumbbell:
Khepri release notes:
https://github.com/rabbitmq/khepri/releases/tag/v0.18.1


dumbbell:
@ansd: I see you added the `backport-v4.3.x` label. The branch is already against `v4.3.x` because the main branch is on Khepri 0.19.0-beta.x. Is this label still helpful to indicate that the patch will be available in 4.3.x, regardless of the target branch? Or should I submit the patch against the main branch instead?

ansd:
I didn't notice that this PR targets already `v4.3.x`.

One question though: Can we backport this PR to `v4.2.x` and `v4.1.x` in the commercial backports repo?

dumbbell:
RabbitMQ 4.2.x uses Khepri 0.17.x, so we can’t backport it directly but it can use Khepri 0.17.8 which brings Horus 0.5.0 and the same optimisations.

RabbitMQ 4.1.x uses Khepri 0.16.x, thus no, unfortunately.

PR #17238, opened 2026-08-24
Fix federation resource alarm blocking (Fixes #16670)

### LINKED ISSUE #16670: Federation links do not respect resource alarms on the downstream node
## Problem

Federation exchange and queue links continue publishing messages to the downstream node during disk and memory resource alarms. This can accelerate the condition that triggered the alarm (e.g., filling disk or exhausting memory).

All other publishing paths in RabbitMQ respect resource alarms:

- Network (AMQP 0-9-1) connections receive `connection.blocked` and cooperatively stop publishing
- AMQP 1.0 sessions have their `incoming-window` set to 0 via a FLOW frame
- The local shovel registers directly with `rabbit_alarm` and buffers messages until alarms clear
- The AMQP 0-9-1 shovel registers a blocked handler on the destination connection

Federation links are the exception.

## Root Cause

Federation's downstream connection uses `#amqp_params_direct{}` (an in-process connection to the local node). The alarm notification machinery is present but unused:

1. `rabbit_direct:connect/5` registers the connection pid with `rabbit_alarm`
2. `amqp_gen_connection` receives `{conserve_resources, ...}` casts and tracks `blocked_by` state
3. When blocked, it sends `#'connection.blocked'{}` to the registered `block_handler` pid

However, federation never calls `amqp_connection:register_blocked_handler/2` on the downstream connection (`DConn`), so the blocked notification is sent to `none` and discarded.

Additionally, federation publishes via `amqp_channel:cast/3` (the `noflow` variant), so it does not participate in credit flow backpressure from queues. This means even queue-level backpressure (quorum queue soft-limit `{block, QName}` actions, classic queue credit flow) does not reach federation. A queue that is overloaded but has not yet triggered a system-wide alarm has no protection from federation publishing.

The server-side `rabbit_channel` has no alarm check in its `handle_method(#'basic.publish'{}, ...)` path - it routes and delivers unconditionally. Neither `rabbit_queue_type:deliver/4` nor the individual queue type implementations (`rabbit_fifo_client:enqueue`, classic queue deliver) reject messages based on resource alarms.

## Impact

During a resource alarm, federation links continue to deliver messages to local queues at whatever rate the upstream provides them. The upstream `basic.qos` prefetch (default 1000) bounds in-flight messages at any instant, but since downstream confirms return quickly (queue acceptance is not alarm-gated), the prefetch window refills continuously.

This means federation can sustain full throughput into local queues during alarms, potentially:

- Filling disk further during a disk alarm
- Increasing memory pressure during a memory alarm
- Prolonging or worsening the alarm condition

Even below alarm thresholds, a backed-up queue cannot apply backpressure to federation because the `noflow` publish path bypasses credit flow entirely.

## Affected Modules

- `deps/rabbitmq_exchange_federation/src/rabbit_federation_exchange_link.erl`
- `deps/rabbitmq_queue_federation/src/rabbit_federation_queue_link.erl`
- `deps/rabbitmq_federation_common/src/rabbit_federation_link_util.erl`

## Suggested Fix

Two related gaps should be addressed together:

### 1. Resource alarm blocking

Register a blocked handler on the downstream connection and pause forwarding when blocked:

1. In `rabbit_federation_link_util:start_conn_ch/5`, after opening the downstream connection, call `amqp_connection:register_blocked_handler(DConn, self())`
2. In both link modules, handle `#'connection.blocked'{}` in `handle_info` - set a blocked flag in state
3. Handle `#'connection.unblocked'{}` - clear the flag and drain any buffered messages
4. In `handle_info({#'basic.deliver'{}, ...})`, check the blocked flag before calling `forward/9` - if blocked, buffer the message

The buffer is naturally bounded by the upstream `prefetch_count` (default 1000). Once the link stops acking upstream messages (because it stops forwarding), the upstream channel stops delivering. When the alarm clears and the buffer drains, upstream delivery resumes.

This mirrors the approach used by `rabbit_amqp091_shovel`, which registers a blocked handler and maintains a `pending` queue with the same bounded-by-prefetch property.

Note: `amqp_gen_connection` already handles the race where an alarm is active at registration time - it immediately sends `#'connection.blocked'{}` to the newly registered handler.

### 2. Queue-level backpressure via credit flow

Switch from `amqp_channel:cast/3` (noflow) to `amqp_channel:cast_flow/3` for downstream publishes. This enables credit flow between the federation link process and the downstream channel, allowing the channel to apply backpressure when queues return `{block, QName}` actions.

The link process would need to handle `{bump_credit, ...}` messages and track whether it is blocked by credit flow (similar to how `rabbit_amqp091_shovel` calls `credit_flow:handle_bump_msg/1` and checks `credit_flow:blocked/0`).

This provides queue-level backpressure before conditions escalate to a system-wide alarm.

Vishal-770:
## Proposed Changes

Fixes #16670

When a node triggers a system-wide resource alarm (memory or disk), federation downstream connections ignore the alarm because they do not register a blocked handler. Additionally, because downstream publishes use `amqp_channel:cast/3` instead of `cast_flow/3`, federation links ignore queue-level credit flow backpressure. This allows the federation plugin to overwhelm overloaded nodes.

This PR addresses both gaps by aligning federation backpressure with the shovel plugin:
* `rabbit_federation_link_util:start_conn_ch/5` now registers the link process to receive `connection.blocked` notifications
* `rabbit_federation_link_util:forward/9` now uses `amqp_channel:cast_flow/3` to participate in queue-level credit flow
* Both the exchange and queue link processes now buffer `basic.deliver` messages when blocked
* Because federation uses `on-confirm` or `on-publish` ack modes by default, delaying the downstream publish inherently delays the upstream `basic.ack`, which throttles the upstream consumer. Note that this bounds the buffer by `prefetch-count` only in those ack modes, and only when `prefetch-count` is not 0, which means unlimited. With `ack-mode: no-ack` there is no acknowledgement to withhold and the buffer is unbounded, which matches the AMQP 0-9-1 shovel

## Commits added after review

> [!NOTE]
> This section was drafted by Claude (Anthropic's Claude Code) under the direction of @lukebakken. The code references were verified against this branch, and the branch history was rewritten so that it contains no add-then-revert pairs: what each commit says is what survives in the final diff.

The buffering feature itself is @Vishal-770's, in the first commit. The commits after it fix defects found while reviewing it, and one gap in `deps/rabbit` that #16670 depends on.

**`Replay cluster-wide alarms to a registering client`** (`deps/rabbit`, worth reviewing on its own)

A resource alarm blocks publishers cluster-wide, because `maybe_alert/5` calls `alert_local/3` for every alarm regardless of which node raised it. But `rabbit_alarm:internal_register/3` replays only `dict:find(node(), alarmed_nodes)` to a newly registered alertee, so a client registering while a *different* node is alarmed is told nothing and keeps publishing at full rate.

Most callers hide this by seeding their own state from the cluster-wide list `register/2` returns. `rabbit_direct` and `rabbit_stream_reader` do not. For federation this is the case that matters most: a link's downstream connection is a direct connection, so a link starting while a peer node was alarmed never received `connection.blocked`, which is what made the plugin's own alarm handling ineffective.

The fix is in the alarm process rather than in each caller, which also fixes stream connections and orders the replay against an alarm clearing during registration. Registration from outside cannot do that: `rabbit_direct:connect/5` runs in an rpc worker rather than in the connection process. A remote alertee is deliberately excluded, because `alert/4` reaches one only via `alert_remote/3`, which runs solely for this node's own alarms, so a peer's alarm replayed to a remote pid could never be cleared. `rabbit_direct` itself is unchanged by this branch.

Other direct connection users, and so other things whose behaviour changes during an alarm on a peer node: the shovel plugin, the HTTP API endpoints that publish and get messages, and the tracing plugin.

**`Revert unconditional basic.qos in federation links`**

Prefetch does not bound the buffer for a `no-ack` consumer on any queue type federation can reach: streams reject `no-ack` consumers outright, quorum queues settle them inside the upstream channel and restore credit immediately, and classic queues skip the limiter. This restores parity with `rabbit_amqp091_shovel`, which also skips `basic.qos` in `no-ack`.

**`Restart a federation link rather than strand its buffered deliveries`**

A link holding buffered deliveries cannot survive losing its downstream channel. `credit_flow:peer_down/1` clears the block, so the next delivery would be forwarded ahead of everything buffered, and neither `bump_credit` nor `connection.unblocked` can arrive on a dead channel to drain the rest. It restarts instead, with `{shutdown, restart}` so `supervisor2` treats it as an explicit restart and logs no crash report.

Resource cleanup had to change with it. Deleting the internal upstream queue on that restart would discard exactly what the restart exists to redeliver, plus everything checked out unacked, so `cleanup_mode/4` keeps the queue when the link is blocked or buffering. It takes the terminate reason as well, because a link stopping *for good* must still clean up: its child spec is removed, so no link is left to delete the queue and it would go on accumulating messages nothing consumes. `disconnect_for_shutdown` goes through the same decision instead of reading `resource_cleanup_mode` directly, where it had been deleting the queue holding the buffered deliveries during a drain while the log line said they would be redelivered.

Both links now treat `closing` as blocked, so a delivery arriving after `disconnect_for_shutdown` is not published onto a closed connection: with `on-confirm` that exits in `next_publish_seqno/1` during a *graceful* shutdown, and with `on-publish` the cast is a silent no-op and the message is then acknowledged upstream and lost.

Also in this commit: `forward/9` no longer acknowledges the upstream on its reject path in `no-ack`, where the tag is not tracked and `rabbit_channel:collect_acks/3` would raise `precondition_failed` and close the channel; and a terminating link reports the buffer it is discarding, warning rather than informing when `ack-mode` is `no-ack` because the upstream kept no copy.

### Tests

Two new suites, `exchange_alarm_SUITE` and `queue_alarm_SUITE`, plus cases added to `rabbit_direct_SUITE` and `unit_SUITE`:

* `alarm_buffers_and_drains_in_order` and `alarm_buffers_no_ack_and_drains_in_order` (both alarm suites): a link buffers under an alarm and drains in publish order once it clears
* `downstream_death_while_buffering_restarts_link`: a link that loses its downstream channel while buffering restarts and the upstream redelivers
* `retiring_a_buffering_link_deletes_the_upstream_queue`: clearing the policy while the link holds buffered deliveries deletes the internal upstream queue rather than orphaning it
* `blocked_by_a_remote_node_alarm`: a direct connection opened during an alarm on a peer node receives `connection.blocked`
* `a_remote_alertee_is_not_told_about_a_peer_alarm`: a remote alertee is not sent a peer node's alarm, which it could never have cleared
* `cleanup_mode_keeps_the_queue_only_for_a_recoverable_restart`: `cleanup_mode/4` asserted as a complete truth table over reason, blocked, buffer and configured mode, in the unit suite because it is a pure function and both of its mistakes are silent

Every one was checked by reverting the fix it covers and confirming that this test, and not merely some test, fails:

| Fix | Mutation applied | Test that fails |
| --- | --- | --- |
| Reason-keyed `cleanup_mode/4` | drop the reason clause | `cleanup_mode_keeps_the_queue_only_for_a_recoverable_restart`, `retiring_a_buffering_link_deletes_the_upstream_queue` |
| `Blocked` in the keep predicate | drop `Blocked orelse` | `cleanup_mode_keeps_the_queue_only_for_a_recoverable_restart` |
| Cluster-wide replay in `client_replay/2` | make it local-only | `blocked_by_a_remote_node_alarm` |
| Remote-pid exclusion in `client_replay/2` | remove the guard | `a_remote_alertee_is_not_told_about_a_peer_alarm` |

The alarm suites also assert that no crash reports appear in the logs.

### Known coverage gaps

Two of the fixes are not pinned by a test, stated here rather than left to be discovered:

* the cleanup decision in `disconnect_for_shutdown`. Reverting it to read `resource_cleanup_mode` directly leaves every test green; instrumenting the branch shows the delete is not reached by way of `disconnect_all/0` in this configuration, so a test written against that path would assert nothing
* the `closing` guard in `is_blocked/1`. It needs a delivery to land after `disconnect_for_shutdown`, which is a race a test does not control

### Deliberately not in this PR

* **#17438** Draining the buffer responsively, and dropping a message before buffering it rather than at drain time. Both were written and then removed, because nothing observes the responsiveness: restoring the whole-buffer loop leaves every test green. That issue tracks the work along with the coverage it needs
* **#17440** A link with an *empty* buffer still survives a clean downstream channel death and keeps a dead pid. Pre-existing, and changing when links restart has a wider blast radius than this PR should take on
* **#4667** is related: that issue's orphaned upstream queue comes from `upstream_queue_name/3` embedding a per-node `cluster_name_default/0`, and the retained queue here interacts with it. See [this comment](https://github.com/rabbitmq/rabbitmq-server/issues/4667#issuecomment-5654120672)

## Types of Changes

- [x] Bug fix (non-breaking change which fixes issue #16670)
- [ ] Build system and/or CI

## Checklist

## Further Comments

`rabbit_federation_link_util:drain_buffer/4` handles the case where the process becomes blocked by credit flow *while* the buffer is draining, for example immediately after a global alarm clears: it stops and preserves the remaining un-forwarded messages, mimicking the pending queue strategy in `rabbit_amqp091_shovel:forward_pending/1`.


lukebakken:
@Vishal-770 I have converted this to draft as I have changes/additions to push to your PR branch.

lukebakken (commented):
> [!NOTE]
> This comment was drafted by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting. Claude verified the code references against branch head 273bf321ff; the decision to set the prefetch in no-ack mode was @lukebakken's, and the open question below is an unsettled design choice rather than a recommendation.

I pushed a follow-up commit (273bf321ff) addressing the code review findings on the resource-alarm blocking work from a56b09b307.

**Code changes**

- `credit_flow:peer_down/1` is now called from the downstream-channel-death branch of `handle_down` in both link modules. Without it, a link that survived a clean DCh death could stay `credit_flow:blocked/0` forever with an undrained buffer.
- The buffer-drain state machine is factored into a single `rabbit_federation_link_util:drain_buffer/4` instead of being duplicated in each link module. It pre-checks `IsBlockedFun(State) orelse credit_flow:blocked()`, so a `bump_credit` or `connection.unblocked` that arrives while the alarm is still active forwards nothing. This also closes a gap in the original code, where the `bump_credit` path checked only `credit_flow:blocked()` and could forward one buffered message past an active alarm.
- The new `handle_info` clauses for `connection.blocked`, `connection.unblocked`, and `bump_credit` now match `#state{}` (the queue link also requires `?is_amqqueue(Q)`), so a message arriving in the not-started state cannot crash the link with `{badrecord, state}`.
- `basic.qos` prefetch is now applied even in `no-ack` mode (see the open question below).
- Restored the `should_forward/2` rationale comment above the `ForwardFun` in the exchange link's `do_deliver/3`, and removed the two `%% Drain the buffer` narrator comments.

**Tests**

Added `exchange_alarm_SUITE` and `queue_alarm_SUITE`. Each publishes while the downstream node's memory watermark is forced to 0, asserts nothing is forwarded during the alarm, then clears the alarm and asserts every message reaches the downstream in publish order, under both `on-confirm` and `no-ack` ack modes. The queue suite subscribes a downstream consumer first, because queue federation only pulls from upstream while the downstream queue has an active local consumer at priority >= 0 (`rabbit_federation_queue:consumer_state_changed/3`).

**One open design question worth a maintainer's opinion before merge**

`no-ack` + classic upstream: the prefetch is now always set, but classic queues do not track `Volume` for `no-ack` deliveries, so the ceiling is effectively ignored there. Quorum queues and streams honour it via credit-based flow. So the buffer bound holds for quorum/stream upstreams but remains best-effort for a classic `no-ack` upstream. Is that acceptable, or do we want a hard buffer cap for that case?

michaelklishin:
This has not been forgotten.

lukebakken:
I moved to draft to do some follow-up work. When that's done I'll change the status and will ping @michaelklishin, thanks.

lukebakken:
@michaelklishin whew, my 🧞‍♂️ struggled with this one. I'll tackle the follow-up issues once this merges.

Issue #17255, opened 2026-08-27
Deleting a destination exchange produces no binding deletions (4.3.x regression)


RodrigoBernardesIbope:
### Describe the bug

When an exchange that is the **destination** of an exchange-to-exchange binding is
deleted, RabbitMQ 4.3.x removes the binding record but produces an empty set of
`rabbit_binding:deletions()`.

Because the deletions map is empty, `rabbit_binding:process_deletions/1` and
`rabbit_binding:notify_deletions/2` appear to have nothing to iterate over, so:

* the `remove_bindings/3` callback of the **source** exchange type is never invoked;
* no `binding.deleted` internal event is emitted.

Exchange types that maintain their own binding state are left out of sync as a result.
With `rabbitmq_consistent_hash_exchange` this shows up as a stale bucket in the hash
ring that keeps routing to the deleted exchange, causing **silent message loss**.

Built-in types (`direct`, `fanout`, `headers`, `topic`) have a no-op `remove_bindings/3`,
so nothing visibly breaks: the binding disappears from listings and publishes correctly
return `routed:false`. The only symptom I could find is the missing `binding.deleted`
event.

Deleting a destination **queue** is not affected in my testing.

This looks like a regression: I don't see the behaviour on 4.2.9.

### Reproduction steps

1. Start RabbitMQ 4.3.4 and enable `rabbitmq_consistent_hash_exchange` and
   `rabbitmq_event_exchange`.

2. Declare the topology (one exchange destination, one queue destination as a control):

   ```bash
   A=(curl -s -u guest:guest -H content-type:application/json)
   B=http://localhost:15672/api

   "${A[@]}" -XPUT  $B/exchanges/%2F/chx    -d '{"type":"x-consistent-hash","durable":true}'
   "${A[@]}" -XPUT  $B/exchanges/%2F/dest-x -d '{"type":"fanout","durable":true}'
   "${A[@]}" -XPUT  $B/queues/%2F/dest-q    -d '{"durable":true}'
   "${A[@]}" -XPOST $B/bindings/%2F/e/chx/e/dest-x -d '{"routing_key":"1"}'
   "${A[@]}" -XPOST $B/bindings/%2F/e/chx/q/dest-q -d '{"routing_key":"1"}'
   ```

3. Check the ring state:

   ```bash
   rabbitmqctl eval 'rabbit_exchange_type_consistent_hash:ring_state(<<"/">>, <<"chx">>).'
   ```

   ```erlang
[...]

PR #17257, opened 2026-08-28
A few related fixes around binding deletion

michaelklishin:
These changes started with a fix for what #17255 describes but some deeper digging found a couple of remaining comparable problems.

This should be safe to backport to `v4.3.x` without a feature flag.

### DIFF
--- a/deps/rabbit/src/rabbit_db_exchange.erl
+++ b/deps/rabbit/src/rabbit_db_exchange.erl
@@ -448,11 +448,29 @@ unconditional_delete_in_khepri(X, OnlyDurable) ->
     delete_in_khepri(X, OnlyDurable, true).
 
 delete_in_khepri(X = #exchange{name = XName}, OnlyDurable, RemoveBindingsForSource) ->
-    Ret = rabbit_db_binding:delete_all_for_exchange_in_khepri(
-            X, OnlyDurable, RemoveBindingsForSource),
-    %% See rabbitmq/rabbitmq-server#17255
-    ok = khepri_tx:delete(khepri_exchange_path(XName)),
-    Ret.
+    UsesUniformWriteRet = try
+                              khepri_tx:does_api_comply_with(uniform_write_ret)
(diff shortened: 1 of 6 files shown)


michaelklishin:
@dumbbell my first attempt turned out to have a regression, so I've left a comment in the part that had to be revisited.

michaelklishin:
The forced push was a rebase on top of `main` to benefit from a few flake fixes from https://github.com/rabbitmq/rabbitmq-server/pull/17280.

deps/rabbit/src/rabbit_db_exchange.erl
```diff
@@ -448,8 +448,11 @@ unconditional_delete_in_khepri(X, OnlyDurable) ->
     delete_in_khepri(X, OnlyDurable, true).
 
 delete_in_khepri(X = #exchange{name = XName}, OnlyDurable, RemoveBindingsForSource) ->
+    Ret = rabbit_db_binding:delete_all_for_exchange_in_khepri(
+            X, OnlyDurable, RemoveBindingsForSource),
+    %% See rabbitmq/rabbitmq-server#17255
     ok = khepri_tx:delete(khepri_exchange_path(XName)),
```
dumbbell: Ideally, you could use `khepri_tx_adv:delete()` here: it would return everything that was deleted, directly or indirectly. I think you could build the Deletions map from that.

The existing call to `rabbit_db_binding:delete_all_for_exchange_in_khepri()` could remain after (i.e. in its current location), just in case there are some tree nodes that don’t have their `keep_while` conditions yet.

michaelklishin:
With the most recent implementation, you can notice that some delete operations will no no-ops in practice because their "deletion targets" are removed by the underlying functions.

I intentionally left that as instead of continuing the refactoring. These no-ops are cheap and safe, and the goal of this PR is to fix a few remaining bugs in that corner of the codebase, not refactor it in a major way.

PR #17260, opened 2026-08-28
Fix minor errors in streams protocol reference doc

the-mikedavis:
This is a collection of small fixes to PROTOCOL.adoc for streams.

* Add missing SASL response codes (0x14 and 0x15)
* Add FilterValue to PublishedMessage version 2. The FilterValue item was previously defined but not included in the PublishedMessage definition. Also note that null is accepted.
* Add missing "*Response" definitions for Subscribe, Create, Delete, CreateSuperStream, and DeleteSuperStream. These commands all use the default response (i.e. just ResponseCode).
* Separate TuneResponse into its own definition. It's the same as TuneRequest but the high bit on the key is set, so the request can't be reused for the response _exactly_ as-is.
* Remove stray "Mechanism" reference in SaslHandshakeRequest. The request half of the handshake doesn't include a requested mechanism.
* Use Mechanisms instead of `[Mechanism]` in SaslHandshakeResponse. Mechanisms is bound to `[Mechanism]` in the definition.
* Fix DeleteSuperStream name in definition.

### DIFF
--- a/deps/rabbitmq_stream/docs/PROTOCOL.adoc
+++ b/deps/rabbitmq_stream/docs/PROTOCOL.adoc
@@ -81,6 +81,8 @@ used to make the difference between a request (0) and a response (1). Example fo
 |Precondition failed|0x11
 |Publisher does not exist|0x12
 |No offset|0x13
+|SASL cannot change mechanism|0x14
+|SASL cannot change username|0x15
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you, @the-mikedavis

PR #17272, opened 2026-08-31
`mc_amqpl`: remove parsing of <= 3.13 AMQP 1.0 compatibility headers

ansd:
Follows through on the decision in https://github.com/rabbitmq/rabbitmq-server/pull/16077#issuecomment-4260962184 to remove parsing of the AMQP 1.0
compatibility headers in 4.4 rather than harden their validation.

RabbitMQ 3.13 and earlier stored a message received over AMQP 1.0 as an
AMQP 0-9-1 `#content{}`, keeping the sections that do not map to AMQP 0-9-1 in
four headers:

```
x-amqp-1.0-properties
x-amqp-1.0-app-properties
x-amqp-1.0-message-annotations
x-amqp-1.0-footer
```

The module that wrote them was replaced in 4.0, so the only messages that carry
them are ones a 3.13 or earlier node stored and no consumer has read since.
`mc_amqpl` still looked all four up on every conversion to AMQP 1.0. Such a
message is still delivered, with these headers now converted like any other
`x-` header, which is what an AMQP 0-9-1 consumer of one has always seen.

Three message annotations have the same history in the other direction:
`x-basic-delivery-mode`, `x-basic-priority` and `x-basic-expiration`.
`rabbit_msg_record` wrote them when it encoded an AMQP 0-9-1 message as AMQP 1.0
for a stream, and it was deleted before 4.0, so only a stream entry written by a
3.13 or earlier node carries them. `x-basic-type` stays, because
`basic.properties.type` has no AMQP 1.0 equivalent.

Two smaller changes to the same conversion code:

* `basic.properties.type` is a free-form AMQP 0-9-1 property, but `convert_to/3`
  took a value of `amqp-1.0` to mean the payload holds AMQP 1.0 encoded body
  sections. It now checks that the payload is well formed and that the sections
  it decodes to belong in a message body, and falls back to a single data
  section otherwise.

* `essential_properties/1` matched `{ok, TTL}` on
  `rabbit_basic:parse_expiration/1`. `rabbit_channel` checks `expiration` before
  building the message, but producers that build a `#content{}` themselves do
  not go through that check, so an expiration that does not parse is now
  ignored. This changes nothing for AMQP 0-9-1.

### DIFF
--- a/deps/rabbit/src/mc_amqp.erl
+++ b/deps/rabbit/src/mc_amqp.erl
@@ -251,18 +251,12 @@ routing_headers(Msg, Opts) ->
     List = application_properties_as_simple_map(Msg, X),
     maps:from_list(List).
 
-get_property(durable, #msg_body_encoded{header = Header} = Msg) ->
+get_property(durable, #msg_body_encoded{header = Header}) ->
     case Header of
         #'v1_0.header'{durable = D} when is_boolean(D) ->
             D;
         _ ->
-            %% fallback in case the source protocol was old AMQP 0.9.1
-            case message_annotation(<<"x-basic-delivery-mode">>, Msg, undefined) of
-                {ubyte, 2} ->
-                    true;
-                _ ->
-                    false
-            end
+            false
     end;
 get_property(timestamp, #msg_body_encoded{properties = Properties}) ->
     case Properties of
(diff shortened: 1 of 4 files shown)


michaelklishin (changes requested):
Should we apply the same idea as in 7111f4d57e to `mc_amqpl:convert_to/3` (lines `341-347`)?

Something like this

```erl
MsgTTL = case rabbit_basic:parse_expiration(Props) of
             {ok, Ttl} -> Ttl;
             {error, _} -> undefined
         end,
```

michaelklishin (approved):

PR #17284, opened 2026-09-01
Add configurable startup delay for client connection listeners

### LINKED ISSUE #13153: A mechanism to delay client connection listener startup

In certain scenarios, most [(in)famously on Kubernetes](https://www.rabbitmq.com/docs/cluster-formation#peer-discovery-k8s) with a sequential `podManagementPolicy`, nodes can start but fail to [sync](https://www.rabbitmq.com/docs/clustering#restarting) its [metadata store](https://www.rabbitmq.com/docs/metadata-store) tables from peers.

This can affect all kinds of environments, for example, https://github.com/rabbitmq/rabbitmq-server/discussions/3837 and https://github.com/rabbitmq/rabbitmq-server/discussions/13151 are examples where the same scenario looks probable, even though we do not have conclusive evidence (or much information about the environment in general).

This can also be a matter of timing: that is, a node does begin syncing its metadata store with peers upon restart but it takes a certain amount of time, even if often this would be fractions of a second,
all while a client connection or CLI tools try to perform an operation that cannot possibly
succeed without all metadata store tables being in place.

Client listeners were moved to the very last step of the boot step graph, and even that does not
help in certain environments. Unless 

A while ago, probably in late 2023, @dumbbell and I have discussed this and agreed to the following feature:
a configurable startup delay for client listeners.

By default, the delay will be set to 0 to not delay node startup and not affect, for example, every integration test suite in both RabbitMQ itself and its client libraries, various tools that start a RabbitMQ node, and so on.

To introduce a non-zero delay, the user will have to opt-in like so:

```ini
# a delay of ten seconds, applied once for all listeners on the node
listeners.startup_delay = 10
```

There aren't that many alternatives.

We could consider checking for metadata table/tree presence before starting the listeners but the risk of
false positives will still exist.

An opt-in delay is very easy to reason about, as we have seen in other parts of RabbitMQ (peer discovery, the Raft algorithm's randomized delay during leader elections).

### Additional context

_No response_

Vishal-770:
This implements the opt-in listener startup delay we discussed in #13153.

### What it does
It adds a `listeners.startup_delay` setting to `rabbit.schema` (defaulting to 0 so we don't affect existing setups). 

During the final boot sequence in `rabbit.erl`, right before `rabbit_networking:boot()` fires up the client listeners, the code checks this config value. If the delay is strictly greater than zero, it logs a brief info message and simply drops a `timer:sleep()` to stall the boot process.

Because this sleep happens inside the isolated `rabbit:start_it/1` setup process, it safely delays the network sockets from opening without messing with the core supervision tree or triggering heartbeat timeouts.

### Testing
I added a snippet test to `config_schema_SUITE_data` to prove the config maps correctly. I intentionally left out an integration test for the delay itself so we don't force the CI runners to sit idle for 10 seconds on every build. 

Fixes #13153


deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -18,6 +18,13 @@
 %% {tcp_listeners, [{"127.0.0.1", 5672},
 %%                  {"[::1]",     5672}]},
 
+%% Delay in seconds before client connection listeners are started.
+%% This can be useful in environments where starting listeners immediately
+%% results in connection errors if the metadata store has not fully synced.
+{mapping, "listeners.startup_delay", "rabbit.listeners_startup_delay", [
+    {datatype, integer}
+]}.
```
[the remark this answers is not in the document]
michaelklishin: This is a good suggestion.

deps/rabbit/priv/schema/rabbit.schema
```diff
@@ -18,6 +18,14 @@
 %% {tcp_listeners, [{"127.0.0.1", 5672},
 %%                  {"[::1]",     5672}]},
 
+%% Delay in seconds before client connection listeners are started.
+%% This can be useful in environments where starting listeners immediately
+%% results in connection errors if the metadata store has not fully synced.
```
michaelklishin: It's not about just the metadata store.

"in environments where applications may begin connecting to the node immediately when the port is enabled, even if some parts of the node are still booting
concurrently".

deps/rabbit/src/rabbit.erl
```diff
@@ -1082,6 +1082,18 @@ do_run_postlaunch_phase(Plugins) ->
 
         %% Start listeners after all plugins have been enabled,
         %% see rabbitmq/rabbitmq-server#2405.
+
+        %% Delay listener startup if configured. This is useful in certain environments
+        %% to allow the metadata store to fully sync before accepting client connections.
```
michaelklishin: This comment is largely useless.

michaelklishin (changes requested):

lukebakken:
I can review this soon as well 👀

lukebakken (commented):
> [!NOTE]
> This review comment was drafted by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who reviewed it before posting. The soft-warning approach was his call; the 300 second threshold and the log wording are Claude's suggestions, verified to compile against the PR branch.

Looks good and matches the design agreed in #13153.

One optional refinement on the delay clause in `deps/rabbit/src/rabbit.erl`. A mistyped large value (say `listeners.startup_delay = 6000`) silently keeps the node unready for that long with only an INFO line to explain it. Emitting a WARNING above a sane threshold makes the misconfiguration visible without rejecting the value, which keeps it consistent with other unbounded delays in RabbitMQ.

A named threshold alongside the existing boot timeout macros (near `BOOT_STATUS_CHECK_INTERVAL`):

```erlang
%% Warn when the configured client listener startup delay is greater than this
%% many seconds, since the node stays unready until the delay elapses.
-define(LISTENERS_STARTUP_DELAY_WARN_THRESHOLD, 300).
```

and the delay clause (around line 1086) using it:

```erlang
Delay when is_integer(Delay), Delay > 0 ->
    case Delay > ?LISTENERS_STARTUP_DELAY_WARN_THRESHOLD of
        true ->
            ?LOG_WARNING("listeners.startup_delay is set to ~b seconds; "
                         "the node will not accept client connections "
                         "until this delay elapses", [Delay]);
        false ->
            ?LOG_INFO("Delaying startup of client connection listeners "
                      "by ~b seconds", [Delay])
    end,
    timer:sleep(Delay * 1000);
```

Vishal-770:
haha good catch on that edge case. thanks for pushing that update in directly. honestly learning a lot from you guys on this PR, appreciate the reviews!

michaelklishin (changes requested):
The new key ended up at the top of the `rabbit.schema` file, which is hard to justify.

I will finish this work in a separate PR.

michaelklishin:
Moved to https://github.com/rabbitmq/rabbitmq-server/pull/17289 with attribution.

PR #17286, opened 2026-09-01
[ui] Run Javascript tests when npm is available 

MarcialRosales:
## Proposed Changes

Runs JavaScript tests during `make tests` when `npm` is installed and `package.json` is present in the plugin. If `npm` is not installed, `make tests` skips the `npm-test` target and runs only the Erlang tests, preserving the build workflow for environments without Node.js.

Split from [rabbitmq/rabbitmq-server#17109](https://github.com/rabbitmq/rabbitmq-server/pull/17109) to keep that PR focused.

This PR introduces infrastructure to facilitate authoring JavaScript tests using the pattern established in `rabbitmq_stream_browser`. Sample reference tests have been added to the following plugins:

* `rabbitmq_management`
* `rabbitmq_stream_management`
* `rabbitmq_shovel_management`
* `rabbitmq_federation_management`

**NOTE**: There is another PR (https://github.com/rabbitmq/rabbitmq-server/pull/17295) that converts the majority of JavaScript files to ES modules, allowing unit tests to import modules directly without VM sandboxing..

## Types of Changes

- [x] Build system and/or CI


CONTRIBUTING.md
```diff
@@ -172,6 +172,17 @@ To stop a previously started cluster:
 gmake stop-cluster NODES=3
 ```
 
+### Running JavaScript Tests
+
+Plugins with a JavaScript-based Web UI can run JavaScript tests alongside Erlang tests when executing `gmake test`.
```
michaelklishin, line 177: `gmake test` does not exist. This should say `gmake tests`.

deps/rabbit_common/mk/rabbitmq-npm-tests.mk
```diff
@@ -0,0 +1,15 @@
+# JavaScript/Node tests: any plugin that ships a package.json at its root
+# opts in automatically, with no changes needed to the plugin's own Makefile.
+# Skipped silently if npm isn't installed, since not every developer works
+# with the JavaScript side.
+ifneq ($(wildcard $(CURDIR)/package.json),)
+.PHONY: npm-test
+npm-test:
+	@if command -v npm > /dev/null 2>&1; then \
+		cd $(CURDIR) && npm install && npm test; \
```
michaelklishin: `npm test` here walks the entire subdirectory tree. Should we specify a conventional path that would be expected to keep the JS tests?

michaelklishin: `npm install` on every run is straightforward but unnecessary. Maybe this is not worth fixing.

deps/rabbitmq_federation_management/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "rabbitmq_federation_management",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "test": "node --test"
```
michaelklishin, line 9: Should we add `"engines"` to require a modern Node.js?

michaelklishin (changes requested):

CONTRIBUTING.md
MarcialRosales: It should . Fixed 5792e634905eca566aca6df48ec03470cfc53463

deps/rabbit_common/mk/rabbitmq-npm-tests.mk
MarcialRosales: Absolutely. Fixed b7607323ed4ea9d4a141d836a8460cd26aa72e09

MarcialRosales: See this improvement 3e1ab277911139a38ce913691eac0cc184932a37

deps/rabbitmq_federation_management/package.json
MarcialRosales: Would you set the engines on each package.json ? 
```
{
  "engines": {
    "node": ">=21"
  }
}
```
This means that if we bumped it up, we need to go thru every package.json. Maybe not a big deal to do it. 

Or we could check the npm version in the makefile? 
```
npm-test:
	@if command -v npm > /dev/null 2>&1 && node -e 'process.exit(parseInt(process.versions.node) < 21 ? 1 : 0)' > /dev/null 2>&1; then \
		cd $(CURDIR) && \
		{ [ package-lock.json -nt package.json ] || npm install; } && \
		npm test -- 'test/**/*.test.js'; \
	else \
		echo "Skipped JavaScript tests for $(PROJECT): Node.js >= 21 and npm are required"; \
	fi
```

which option do you favor?

lhoguin: I would just put them in package.json.

MarcialRosales: Confirmed with @lhoguin that we should go for package.json option. Fixed in 49ec8aea3bf53b818308bf6b52a4e4b6a036cfe1

michaelklishin (approved):

PR #17287, opened 2026-09-01
[management] /init endpoint

MarcialRosales:
## Proposed Changes

Consolidates all the settings required by the management ui into a single /init endpoint, replacing the four separate blocking requests the login flow used to make. Not only that, the /overview endpoint have been used to provide settings required by the ui. We cannot remove them because they are somewhat part of the public api. However, a very recent setting `require_definition_json_extension` has been removed.

Base branch: this PR is based on another [PR](https://github.com/rabbitmq/rabbitmq-server/pull/17286).

Split from [rabbitmq/rabbitmq-server#17109](https://github.com/rabbitmq/rabbitmq-server/pull/17109) to keep that PR focused.

What change:

- New GET /init (rabbit_mgmt_wm_init.erl) returns {settings, vhosts, nodes} in one response. nodes is omitted server-side for non-monitor users, replacing the old client-side ac.isMonitoringUser() check.
- Login used to make four requests before rendering the app: /overview, /vhosts, /nodes, plus a second, redundant /nodes fetch buried inside setup_global_vars. All four collapse into the one /init call.
- Settings computation moved out of /overview into rabbit_mgmt_features:get_settings/1 and get_product_info/0, so /init and /overview share one implementation instead of /overview duplicating it. rabbit_mgmt_wm_overview.erl shrinks from 
- Client side: check_login()/login() fetch /init once into window.app_settings/app_vhosts/app_nodes; load_ui()/setup_global_vars() read from there instead of issuing their own requests. format() now injects settings into every template's render context.
- Fixed a live bug along the way: the #/cluster-name route handler called setup_global_vars() with no argument, which dereferences settings fields with nothing to read them from - throws on main today. It now updates window.app_settings.cluster_name directly and repaints via a new update_cluster_name_ui() helper (carved out of setup_global_vars for this purpose).

Breaking change

- /overview no longer has a top-level `require_definition_json_extension` field - it now lives under a nested definitions object (rabbit_mgmt_features:get_definitions_settings/0), alongside the other settings /init and /overview now share.

### DIFF
--- a/deps/rabbitmq_management/priv/www/js/main.js
+++ b/deps/rabbitmq_management/priv/www/js/main.js
@@ -136,19 +136,25 @@ function check_login () {
     return false;
   }
 
-  // Fetch initialization data synchronously (which sends Authorization header)
-  var rawInitData = sync_get('/init');
-  if (rawInitData) {
-      var initData = JSON.parse(rawInitData);
-      window.app_settings = initData.settings;
-      window.app_vhosts = initData.vhosts;
-      if (initData.nodes) {
-          window.app_nodes = initData.nodes;
-      }
-  } else {
-      console.error("Failed to load /api/init");
-  }
+  load_init_data();
   load_ui(user);
+  return true;
+}
+
+// Sent with the Authorization header, so it must run after the credentials
+// have been set.
+function load_init_data() {
+  var raw = sync_get('/init');
+  if (!raw) {
+      console.error("Failed to load /api/init");
+      return false;
+  }
+  var data = JSON.parse(raw);
+  window.app_settings = data.settings;
+  window.app_vhosts = data.vhosts;
+  if (data.nodes) {
+      window.app_nodes = data.nodes;
+  }
   return true;
 }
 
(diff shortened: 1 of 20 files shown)


deps/rabbitmq_management/priv/www/js/main.js
```diff
@@ -135,6 +135,19 @@ function check_login () {
...
+  var rawInitData = sync_get('/init');
+  if (rawInitData) {
+      var initData = JSON.parse(rawInitData);
+      window.app_settings = initData.settings;
+      window.app_vhosts = initData.vhosts;
+      if (initData.nodes) {
+          window.app_nodes = initData.nodes;
+      }
+  } else {
+      console.error("Failed to load /api/init");
```
michaelklishin, line 147: This is not really error handling. In fact, the code proceeds to use `load_ui/1`.

deps/rabbitmq_management/src/rabbit_mgmt_wm_init.erl
```diff
@@ -0,0 +1,61 @@
...
+    {cowboy_rest, Req2, #context{}}.
+
+is_authorized(ReqData, Context) ->
+    rabbit_mgmt_util:is_authorized(ReqData, Context).
+
+content_types_provided(ReqData, Context) ->
+    {rabbit_mgmt_util:responder_map(to_json), ReqData, Context}.
+
+to_json(ReqData, Context) ->
+    SettingsJSON = get_settings_json(ReqData, Context),
```
michaelklishin: Several other endpoints, e.g. `GET /overview`, `GET /nodes`, `GET /vhosts` have a try-catch in similar settings. That avoids a 500 in certain scenarios where the rest of the endpoints would return a reasonable 400 response .

michaelklishin (changes requested):

michaelklishin (approved):
Everything I personally found was addressed.

deps/rabbitmq_management/priv/www/js/main.js
michaelklishin: This turned out to not be a practical issue I could reproduce, so please ignore it.

michaelklishin:
This is a big change with more non-trivial PRs to follow, so the entire line of work is currently being discussed as 4.4.0-only for its sheer size (blast radius if you will).

PR #17288, opened 2026-09-01
QQ: park delayed messages on the untracked enqueue path

lukebakken:
> [!NOTE]
> This pull request description was drafted by Claude (Anthropic's Claude Code) under the direction of @lukebakken, who provided significant guidance throughout and reviewed the change before opening it. The fix and its tests were verified against the quorum-queue test suite.

## What

Two changes against the `qq-vNext` (rabbit_fifo v9) branch, found while reviewing the deferred-message and message-delay work.

### Fix: park delayed messages on the untracked enqueue path

v9 added delivery-time parking only to the tracked enqueuer clause of `rabbit_fifo:maybe_enqueue/8`. The untracked clause, used by stateless deliveries such as at-most-once dead-lettering, enqueued straight into the ready queue and ignored the `dt` annotation.

`rabbit_quorum_queue:deliver/3` stamps the `dt` annotation on every message, including `{Q, stateless}` targets, and `rabbit_dead_letter:publish/5` preserves `x-opt-delivery-time`. A future-dated message dead-lettered into a quorum queue therefore became immediately deliverable instead of being parked until its delivery time.

The untracked clause now routes through `get_delivery_time/2` as well, so a future delivery time parks the message via `delayed_in/5` and a past or absent one keeps the current ready-queue behaviour. Covered by two new `rabbit_fifo_SUITE` cases mirroring the existing tracked-enqueue tests.

### Docs: document that first enqueue never registers a deferral token

Both enqueue clauses of `maybe_enqueue/8` pass `undefined` as the deferral token to `delayed_in/5`, so an `x-opt-deferral-token` supplied at publish time is ignored; only `should_delay/5` on the return/modified path reads the token. Added a comment recording that this is intentional, since deferral tokens are established solely by a consumer parking a message via the MODIFIED outcome.

## Testing

`gmake -C deps/rabbit ct-rabbit_fifo` passes (169 cases, including the two new ones).

### DIFF
--- a/deps/rabbit/src/rabbit_fifo.erl
+++ b/deps/rabbit/src/rabbit_fifo.erl
@@ -2273,6 +2273,10 @@ maybe_enqueue(RaftIdx, Ts, From, MsgSeqNo, RawMsg,
                                          },
                     {ok, State, Effects0};
                 DeliveryTime ->
+                    %% A deferral token is never registered on first enqueue:
+                    %% tokens are established only by a consumer parking a
+                    %% message via the MODIFIED outcome, so an
+                    %% x-opt-deferral-token set at publish time is ignored here.
                     Delayed = delayed_in(DeliveryTime, RaftIdx,
                                          Msg, undefined,
                                          State0#?STATE.delayed),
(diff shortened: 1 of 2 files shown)


kjnilsson:
fixed in 7097f3959b9893a58bc4dd4d87c374ff68411f85

PR #17295, opened 2026-09-02
[management] Introduce auth providers and a bootstrap pipeline for login/logout

MarcialRosales:
## Proposed Changes

- Replace the scattered oauth.enabled checks and duplicated login/logout/error-handling logic in main.js with a basic/oauth2 AuthProvider abstraction (auth-providers.js), so each mechanism owns its own presentError, signOut, and onUnauthorized behavior instead of that logic being open-coded at every call site
- Add a reversible bootstrap step pipeline (bootstrap-steps.js) that drives page load: login gates may veto with a user-facing reason, init steps run after, and a single completed stack means an init failure correctly unwinds any login gates that already ran
- Add auth-options.js, which separates "strict filters" from "preferred preselects" when deciding which login mechanism(s) to offer - this also fixes a pre-existing bug in login_oauth.ejs where a wrong preference match (object vs string) caused a login section to render as both section-visible and section-invisible at once
- Resolve the active auth provider from the login flow that actually ran, not from sniffing the token type ‚Äî the previous approach misclassified encrypted-basic-auth logins as oauth2, since /login issues a bearer token for basic auth too when management.credential_encryption_secret is set
- Fix a related gap surfaced by Selenium testing: idp-initiated oauth sessions (no click-through login) never declared themselves to the provider registry, so sign-out and unauthorized-handling silently fell back to the basic-auth implementation for those sessions
- Harden the Selenium suite itself: fix several tests whose assert.rejects(...) calls weren't awaited (silently passing regardless of the actual assertion), replace slow/noisy console.error-logging waits with quiet equivalents for intentional "should not be visible" checks, fix teardown() misclassifying a self-skipped (this.skip()) test as a failure, and fix the tls-terminating-proxy component's --add-host=host.docker.internal:host-gateway resolving to the wrong address on non-Docker-Desktop runtimes (e.g. Rancher Desktop)

**IMPORTANT**: This PR must be followed by another PR in tanzu -once it is merged- which refactors all the plugins with UI. 

## Types of Changes

- [x] Build system and/or CI


michaelklishin:

PR #17300, opened 2026-09-03
Support an initial offset when creating a stream

ansd:
A stream always started at offset 0. The new `x-stream-initial-offset` queue argument sets the offset of the first message published to a stream, which is what `first` and `next` resolve to while the stream is still empty.

The offset is fixed when the stream is created, so it is read from the queue arguments only and has no policy equivalent. Redeclaring the stream with a different value fails the existing argument equivalence check. The stream protocol exposes it as the `stream-initial-offset` argument of `Create`, and `rabbitmq-streams add_super_stream` as `--stream-initial-offset`.

A non-zero value requires the `rabbitmq_4.4.0` feature flag because a replica on an older node would start its log at 0 and then reject the writer's first chunk as out of order.

This relies on osiris honouring `initial_offset` for writers, so `rabbitmq-components.mk` still has to be pointed at a release that contains https://github.com/rabbitmq/rabbitmq-server/pull/17300


rabbitmq-components.mk
```diff
@@ -53,7 +53,7 @@ dep_jose = hex 1.11.12
 dep_khepri = hex 0.19.0-beta.3
 dep_khepri_mnesia_migration = hex 0.8.1
 dep_meck = hex 1.0.0
-dep_osiris = git https://github.com/rabbitmq/osiris v1.13.2
+dep_osiris = git https://github.com/rabbitmq/osiris initial-offset
```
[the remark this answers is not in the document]
ansd: Right

deps/rabbitmq_management/priv/www/js/tmpl/stream-queue-arguments.ejs
```diff
@@ -2,5 +2,6 @@
 | <span class="argument-link" field="arguments" key="x-max-age" type="string">Max time retention</span><span class="help" id="queue-max-age"></span>
 | <span class="argument-link" field="arguments" key="x-stream-max-segment-size-bytes" type="number">Max segment size in bytes</span><span class="help" id="queue-stream-max-segment-size-bytes"></span></br>
  <span class="argument-link" field="arguments" key="x-stream-filter-size-bytes" type="number">Filter size (per chunk) in bytes</span><span class="help" id="queue-stream-filter-size-bytes"></span>
-| <span class="argument-link" field="arguments" key="x-initial-cluster-size" type="number">Initial cluster size</span><span class="help" id="queue-initial-cluster-size"></span>
+| <span class="argument-link" field="arguments" key="x-stream-initial-offset" type="number">Initial offset</span><span class="help" id="queue-stream-initial-offset"></span></br>
```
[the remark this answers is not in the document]
ansd: That path serializes to an actual AMQP wire type by JS value type — a JSON string becomes longstr, which rabbit_amqqueue:check_non_neg_int_arg rejects outright (only long/short/etc. pass). Switching it to type="string" wouldn't reduce precision loss, it would break every declare through that UI. Every other numeric x- queue argument in the codebase (x-max-length, x-max-length-bytes, both effectively unbounded) has the identical limitation and nobody's fixed it; a real fix needs a dual-type validator plus coercion at lookup time, which is disproportionate for a range no realistic offset will ever reach. Left as-is.

rabbitmq-components.mk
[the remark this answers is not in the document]
ansd: Blocked on cutting a new Osiris release to contain https://github.com/rabbitmq/osiris/pull/241

deps/rabbitmq_stream_management/priv/www/js/tmpl/superStreams.ejs
```diff
@@ -55,6 +55,7 @@
                   <span class="argument-link" field="arguments" key="max-length-bytes" type="number">Max length bytes</span> <span class="help" id="queue-max-length-bytes"></span>
                   | <span class="argument-link" field="arguments" key="max-age" type="string">Max time retention</span><span class="help" id="queue-max-age"></span>
                   | <span class="argument-link" field="arguments" key="stream-max-segment-size-bytes" type="number">Max segment size in bytes</span><span class="help" id="queue-stream-max-segment-size-bytes"></span></br>
+                  | <span class="argument-link" field="arguments" key="stream-initial-offset" type="string">Initial offset</span><span class="help" id="queue-stream-initial-offset"></span>
```
michaelklishin, line 58: Should this new row use `type="number"`?

deps/rabbit/src/rabbit_stream_queue.erl
```diff
@@ -189,6 +191,29 @@ check_filter_size(Q) ->
             ok
     end.
 
+check_initial_offset(Q) ->
+    Args = amqqueue:get_arguments(Q),
+    case rabbit_misc:table_lookup(Args, <<"x-stream-initial-offset">>) of
+        undefined ->
+            ok;
+        {_Type, 0} ->
```
michaelklishin, line 199: This clause bypasses `rabbit_feature_flags:is_enabled/1`. I assume it's OK since it matches the original behavior.

michaelklishin (changes requested):

deps/rabbitmq_stream_management/priv/www/js/tmpl/superStreams.ejs
ansd: I used type `number` initially. Copilot complained in https://github.com/rabbitmq/rabbitmq-server/pull/17300#discussion_r3925713849 suggesting to use type `string.`

deps/rabbit/src/rabbit_stream_queue.erl
ansd: I pushed a commit that requires this feature flag for `x-stream-initial-offset = 0`.

michaelklishin (approved):

PR #17343, opened 2026-09-07
Shovel: look up non-built-in modules via `rabbit_registry`

### LINKED ISSUE #17341: shovel_protocol registry allows registering custom protocols, but dest-protocol/src-protocol string parsing doesn't support them
### Describe the bug

**Context**

Commit f2d05a4 (https://github.com/rabbitmq/rabbitmq-server/commit/f2d05a4c373b970e3bcdd2d338316143a5f23637) ("Shovel: self-contained shovel protocols", #14904) made rabbit_shovel_behaviour extensible via a proper registry (-rabbit_registry_class(shovel_protocol)), allowing a new protocol to register itself dynamically via rabbit_registry:register(shovel_protocol, <<"my_protocol">>, MyModule).

I tested this mechanism with a third-party module (implementing rabbit_shovel_behaviour, registering correctly at boot — confirmed in the logs: Running boot step rabbit_amqp091_templated_shovel_protocol defined by app rabbitmq_shovel_templated).

**Problem**

Declaring a Shovel with "dest-protocol": "my_protocol" fails validation, with an error suggesting the standard amqp091 protocol was used instead (rejection of my protocol-specific keys, requirement for dest-uri which I hadn't provided).

Root cause: registered_protocol_to_atom/1 in rabbit_shovel_parameters.erl remains a fixed list of 3 values, without consulting the registry:
registered_protocol_to_atom(Protocol) when is_atom(Protocol) -> Protocol;
registered_protocol_to_atom(<<"amqp091">>)                  -> amqp091;
registered_protocol_to_atom(<<"amqp10">>)                   -> amqp10;
registered_protocol_to_atom(<<"local">>)                    -> local;
registered_protocol_to_atom(_)                              -> amqp091.

Permalink v4.3.5 (https://github.com/rabbitmq/rabbitmq-server/blob/0dde27bfdd1984ff7e157226fd97656854a7f359/deps/rabbitmq_shovel/src/rabbit_shovel_parameters.erl#L187-L191) — identical on main as of today.

Any unrecognized protocol silently falls back to amqp091, instead of being resolved dynamically via rabbit_registry:lookup_all(shovel_protocol) (already used elsewhere in this same file, list_all_protocols/0, line 227).

**Effect**

The shovel_protocol registry cannot, as it stands, be used end-to-end by a third-party protocol — only module dispatch (once the atom is known) actually works; recognizing the protocol name from a Shovel definition (management API/JSON) remains limited to the 3 historical protocols.

**Suggestion**

Make registered_protocol_to_atom/1 a dynamic resolution via the registry rather than a fixed case/pattern match, for example by relying on list_all_protocols/0 (or an equivalent lookup) to accept any protocol name currently registered under shovel_protocol, falling back to amqp091 only if genuinely unregistered.

### Reproduction steps

1. Start a RabbitMQ 4.3.5 node with a custom plugin that implements
   `rabbit_shovel_behaviour` and registers itself at boot under a new
   protocol name, e.g.:
```
   -rabbit_boot_step(
      {my_shovel_protocol,
       [{mfa,      {rabbit_registry, register,
                    [shovel_protocol, <<"my_protocol">>, ?MODULE]}},
        {cleanup,  {rabbit_registry, unregister,
                    [shovel_protocol, <<"my_protocol">>]}},
        {requires, rabbit_registry}]}).

```
   Confirmed at startup in the logs:
   "Running boot step my_shovel_protocol defined by app my_plugin"

2. Declare a dynamic Shovel via the management API using this custom
   protocol as `dest-protocol`:
```
   PUT /api/parameters/shovel/%2f/my-shovel
   {
     "value": {
       "src-protocol": "amqp091",
       "src-uri": "amqp://",
       "src-queue": "my-source-queue",
       "dest-protocol": "my_protocol",
       "dest-queue": "my-dest-queue",
       "my-custom-key": "some-value"
     }
   }

```
3. Observe the response

### Expected behavior

The Shovel definition should validate against `my_protocol`'s own
`validate_dest_funs/2` (as returned by the registered module), accepting
`"my-custom-key"` and not requiring any key specific to `amqp091` (such
as `dest-uri`) unless `my_protocol`'s own validation requires it.

Instead, the request fails validation with an error indicating that the
standard `amqp091` protocol's constraints were applied:

```
{"error":"bad_request","reason":"Validation failed\n\nUnrecognised terms
[{<<\"my-custom-key\">>,<<\"some-value\">>}] in my-shovel\nKey \"dest-uri\"
not found in my-shovel\n"}
```

This happens because `registered_protocol_to_atom/1`
(rabbit_shovel_parameters.erl) silently maps any unrecognized protocol
binary to the `amqp091` atom, instead of resolving it against the
`shovel_protocol` registry — so `my_protocol`'s own
`validate_dest_funs/2` is never actually invoked.

### Additional context

_No response_

michaelklishin:
But fall back to AMQP 0-9-1 for unknown [to the runtime, the atom table] atoms.

For shovel types provided by plugins this means "for modules that were not loaded".

Closes #17341.

### DIFF
--- a/deps/rabbitmq_shovel/src/rabbit_shovel_definition.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_shovel_definition.erl
@@ -55,7 +55,7 @@ extract_source_info(Def0, ShovelVHost) ->
 -spec extract_source_info_for_index(Definition :: list() | map(), ShovelVHost :: vhost:name()) -> source_info().
 extract_source_info_for_index(Def0, ShovelVHost) ->
     Def = rabbit_data_coercion:to_proplist(Def0),
-    Protocol = rabbit_shovel_parameters:src_protocol(Def),
+    Protocol = builtin_src_protocol(Def),
     BaseInfo = #{protocol => Protocol, vhost => ShovelVHost},
     case Protocol of
         amqp10 ->
@@ -69,6 +69,16 @@ extract_source_info_for_index(Def0, ShovelVHost) ->
             end;
         _ ->
             extract_amqp091_local_source(Def, BaseInfo)
+    end.
+
+-spec builtin_src_protocol(list()) -> protocol().
+builtin_src_protocol(Def) ->
+    case pget(<<"src-protocol">>, Def) of
+        <<"amqp10">> -> amqp10;
+        amqp10       -> amqp10;
+        <<"local">>  -> local;
+        local        -> local;
+        _            -> amqp091
     end.
 
 %% Parse AMQP 1.0 address using RabbitMQ address scheme v2.
--- a/deps/rabbitmq_shovel/src/rabbit_shovel_parameters.erl
+++ b/deps/rabbitmq_shovel/src/rabbit_shovel_parameters.erl
@@ -184,11 +184,31 @@ protocols(Def) ->
     Dst = dest_protocol(Def),
     {Src, Dst}.
 
-registered_protocol_to_atom(Protocol) when is_atom(Protocol) -> Protocol;
-registered_protocol_to_atom(<<"amqp091">>)                  -> amqp091;
-registered_protocol_to_atom(<<"amqp10">>)                   -> amqp10;
-registered_protocol_to_atom(<<"local">>)                    -> local;
-registered_protocol_to_atom(_)                              -> amqp091.
+registered_protocol_to_atom(Protocol) when is_atom(Protocol) ->
+    Protocol;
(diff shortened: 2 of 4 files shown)


ksuedile:
tested OK (comment in https://github.com/rabbitmq/rabbitmq-server/issues/17341)

thanks !

michaelklishin:
This will need tweaking for Tanzu RabbitMQ.

PR #17349, opened 2026-09-08
Fail the build on end_per_testcase/end_per_suite failures, not just testcase failures

shashank-singh-2999:
## Summary

Common Test reports a test case as passed if its body passed, even when `end_per_testcase`/`end_per_group`/`end_per_suite` subsequently crashes; that failure only shows up as a warning, and `ct_run`'s own exit status stays 0. Both the plain `ct-<suite>` targets and the parallel "master" runs (`ct_master_fork`, used by `deps/rabbit/Makefile` and `deps/rabbitmq_mqtt/Makefile` to fan suites out across peer nodes) have this gap, so a teardown crash can go unnoticed and the build still reports success.

- A new Common Test hook, `cth_fail_on_teardown_crash`:
  - `post_end_per_testcase/4`: a crash in `end_per_testcase` comes back as `Return = {failed, {_, end_per_testcase, _}}`; returning `{fail, Return}` makes Common Test count the test case itself as failed, which on its own fails the build via `ct_run`'s normal exit status.
  - `post_end_per_group/4` / `post_end_per_suite/4`: unlike test cases, returning `{fail, Reason}` here does *not* get reflected in `ct_run`'s exit status. So these two just record whether either one crashed.
  - `terminate/1`: called once, after every suite's own `post_end_per_group`/`post_end_per_suite` above but before Common Test regenerates the cross-run `logs/index.html`/`logs/all_runs.html`. If a crash was recorded, halts with a non-zero status here. The only cost is that those two cross-run index files won't be updated with this run — logged to the console when that happens.
  - Wired in via `CT_HOOKS`/`CT_OPTS` in `deps/rabbit_common/mk/rabbitmq-early-plugin.mk`, applying on CI too. Compiled directly with `erlc`.
- Parallel master runs (`ct_master_fork.erl`): `master_loop/1` returns the list of teardown-only failures collected from the event stream (previously computed but discarded after being printed), and the `ct_master.erl` snippets in `deps/rabbit/Makefile` and `deps/rabbitmq_mqtt/Makefile` fail the run if that list is non-empty, in addition to the existing per-node pass/fail counters.
- Crashes found during `end_per_suite`/`end_per_testcase` teardown should not be silently ignored: if a specific crash is known and acceptable, it should be added to the ignored-crashes list (e.g. via `rabbit_ct_broker_helpers`'s crash-matching), not left to slip through unnoticed because the build didn't fail.

### DIFF
--- a/deps/rabbit_common/mk/rabbitmq-early-plugin.mk
+++ b/deps/rabbit_common/mk/rabbitmq-early-plugin.mk
@@ -39,8 +39,23 @@ endif
 TEST_DEPS += cth_styledout
 dep_cth_styledout = git https://github.com/rabbitmq/cth_styledout.git master
 
+# Fails the build on an end_per_testcase/end_per_group/end_per_suite crash.
+CT_HOOKS += cth_fail_on_teardown_crash
+
 ifneq ($(strip $(CT_HOOKS)),)
-CT_OPTS += -ct_hooks $(CT_HOOKS)
+# `ct_run' expects multiple hooks as `Hook1 and Hook2', not `Hook1 Hook2':
+# the latter is parsed as `Hook1' with `Hook2' as its own init argument.
+CT_OPTS += -ct_hooks $(subst $(space),$(space)and$(space),$(strip $(CT_HOOKS)))
+endif
+
+ifeq ($(PROJECT),rabbit_common)
+# A normal TEST_DEPS entry would rebuild rabbit_common plainly (dropping
+# its -ifdef(TEST) exports), since rabbitmq_ct_helpers depends on it.
+# Compiled directly instead, just for rabbit_common.
+CTH_FAIL_ON_TEARDOWN_CRASH_DIR = $(DEPS_DIR)/rabbitmq_ct_helpers
+test-build::
+	$(verbose) mkdir -p $(CTH_FAIL_ON_TEARDOWN_CRASH_DIR)/ebin
+	$(verbose) cd $(CTH_FAIL_ON_TEARDOWN_CRASH_DIR)/src && erlc -o ../ebin cth_fail_on_teardown_crash.erl
 endif
 
 # We fetch a SECONDARY_DIST if SECONDARY_DIST_VSN is set and
--- a/deps/rabbit/Makefile
+++ b/deps/rabbit/Makefile
@@ -237,16 +237,19 @@ define ct_master.erl
 	peer:call(Pid2, persistent_term, put, [rabbit_ct_tcp_port_base, 20000]),
 	peer:call(Pid3, persistent_term, put, [rabbit_ct_tcp_port_base, 24000]),
 	peer:call(Pid4, persistent_term, put, [rabbit_ct_tcp_port_base, 28000]),
-	[{[_], {ok, Results}}] = ct_master_fork:run("$1"),
+	[{[_], {ok, Results, EventFailed}}] = ct_master_fork:run("$1"),
 	peer:stop(Pid4),
 	peer:stop(Pid3),
 	peer:stop(Pid2),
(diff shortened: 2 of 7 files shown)


lhoguin:
Note that the 'erlang.mk' file is an upstream project and that your changes will get deleted the next time `make erlang-mk` is run. I don't think you need this teardown marker, the CT hooks should provide you with everything you need for this.

I don't yet know whether all end_per_testcase failures should be regarded as test failures. There were too many when I did the parallel-ct stuff so end_per_testcase failures were written but not considered test failures. You will also need to run the test suites many times to ensure there's no end_per_testcase flakes after you have fixed the end_per_testcase failures.

michaelklishin:
This PR modifies `erlang.mk` and wraps CT, so I'll defer to @lhoguin to decide whether the gain is worth the complexity and maintenance burden (`erlang.mk` is upgraded every so often).

deps/rabbit_common/mk/rabbitmq-early-plugin.mk
```diff
@@ -39,10 +39,23 @@ endif
...
 ifneq ($(strip $(CT_HOOKS)),)
-CT_OPTS += -ct_hooks $(CT_HOOKS)
+# `ct_run' expects multiple hooks as `Hook1 and Hook2', not `Hook1 Hook2':
+# the latter is parsed as `Hook1' with `Hook2' as its own init argument.
+CT_OPTS += -ct_hooks $(subst $(space),$(space)and$(space),$(strip $(CT_HOOKS)))
 endif
 
+# Compiled directly (it only needs plain OTP) instead of via
+# `rabbitmq_ct_helpers''s own Makefile, to avoid rebuilding `rabbit_common'
+# plainly as a side effect of a TEST_DEPS entry.
```
lhoguin: This part is curious, I don't really understand. There should be no need to (re)compile this module. Did you run into some sort of circular dependency?

shashank-singh-2999: Not a circular dependency that fails build. I tested adding rabbitmq_ct_helpers to TEST_DEPS in rabbit_common's Makefile, and it reproducibly drops rabbit_common's -ifdef(TEST) exports (e.g. rabbit_misc:decompose_pid/1, used by unit_SUITE). Building rabbitmq_ct_helpers transitively rebuilds rabbit_common plainly first (as one of its own deps), and erlang.mk's mtime-only staleness check then skips rabbit_common's real -DTEST=1 recompile right after, since the .beam already looks up to date. So the explicit compile step isn't just "no dependency exists" — it's specifically avoiding that regression for rabbit_common.

shashank-singh-2999: Worth noting this isn't unique, amqp10_common has the same cyclic TEST_DEPS/DEPS shape with rabbitmq_ct_helpers today, and hits the same race — gmake -C deps/amqp10_common eunit on a clean build silently runs zero tests, since its -ifdef(TEST)-gated EUnit tests never get compiled in.

lhoguin: Right it's an issue we have with _common plugins, but it should only affect them, so I would only apply this workaround to them and not to all applications.

lhoguin (approved):
Commit looks good.

I think next step is to fix some of those errors. I would do many full CI runs of this PR (Re-run all jobs from Test make action page) and compile a list of failures observed and ask the relevant folks to look at them. We don't need to fix all the new failures before merging, but we do need to fix the most frequent ones. The many runs ensure we cover flakes.

lhoguin:
Instead of merging you can `git pull --rebase origin main` that will avoid the extra commit.

PR #17351, opened 2026-09-08
CQ: Count each recovered shared message store reference once

gomoripeti:
## Proposed Changes

There was a bug in the CQ shared message store, the recovery process after an unclean shutdown counted every live message one too many, when it goes through every per-queue index to look for shared messages that are still referenced by any queue.

Bug introduced in commit 18acc01a47 included in 4.0.0

Details: `rabbit_msg_store:count_msg_refs/3` registers references with `ets:update_counter/4` and a default object carrying ref_count=1. The increment is applied to a freshly inserted default as well, so every message came out of dirty recovery with one reference too many. That reference was never released: after a crash, every message that was in the store at the time stayed on disk for good, with its index entry, once all queues had consumed it. Fan-out messages were affected the same way (N queues gave N+1 references).

Start the default at 0. Add a regression test, and make msg_store_v1_compat remove each message once, as its second remove of a zero-reference entry now crashes the store the way it would in production.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

### DIFF
--- a/deps/rabbit/src/rabbit_msg_store.erl
+++ b/deps/rabbit/src/rabbit_msg_store.erl
@@ -2189,12 +2189,14 @@ count_msg_refs(Gen, Seed, State = #msstate{ index_ets = IndexEts }) ->
         %% It can be removed once we no longer support converting from v1 data.
         {MsgId, 1, Next} ->
             index_update_ref_counter(IndexEts, MsgId, +1,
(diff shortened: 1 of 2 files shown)


lhoguin:
Thank you!!

lhoguin:
Did manual backport in https://github.com/rabbitmq/rabbitmq-server/pull/17354

PR #17363, opened 2026-09-09
[oauth2] Validate OpenId discovered attributes

MarcialRosales:
## Proposed Changes

Validate the attributes downloaded from the Idp via the OpenId discovery endpoint. These attributes are:
- The most important URIs used in any OAuth 2.0 Authentication/Authorization Flow: jwks_uri, token_endpoint, authorization_endpoint, end_session_endpoint. 
- The issuer 

**NOTE**: This has been requested by our internal security auditor.

Validation rules implemented by this PR:
- If jwks_uri, or token_endpoint, or authorization_endpoint or end_session_endpoint are present in the OpenId payload, they must use the https scheme. 
- issuer is REQUIRED to be always present and it matches the prefix of the url used to download the OpenId configuration payload. 

Introducing this validation on an existing environment may break it. Because not all Idp are OpenId compliant, or fully. For this reasons, it has been introduced two new settings, at the root level and under individual oauth_providers. 
The settings are (Note that the settings are prefixed with a dot meaning that they are appended to a root setting which can be auth_oauth2 or auth_oauth2.oauth_providers.<idp_name>):
- .discovery.verify_https_endpoints : By default it is true. Set to false, to disable validation of https scheme in the aforementioned endpoints discovered via the OpenId Discovery endpoint
- .discovery.verify_issuer : By default it is true. Set to false, to make it optional and not verifiable. 

This change has to be reported in the release notes as it may be a breaking change in some environments where their IDP does not conform to the OpenId specification. 

## Tasks

- [x] Add  GitHub Security Advisory draft (according to [this](https://docs.google.com/document/d/1kX4NLdjdhOEVNEdZPhkmAe5vGfNEIUfUNMcOzYdRXbY/edit?tab=t.0)) -> https://github.com/rabbitmq/rabbitmq-server/security/advisories/GHSA-g4xh-8cpm-jg28
- [x] Add docs PR -> https://github.com/rabbitmq/rabbitmq-website/pull/2641
- [x] Update release notes. 
- [ ] Release notes
- [ ] Merge PR https://github.com/rabbitmq/rabbitmq-oauth2-tutorial/pull/62 once this PR is merged 

## Types of Changes

- [ ] Build system and/or CI

## Observations

This PR modifies the existing #oauth_provider{} record. Since that record is never sent between nodes, this change carries no risk in mixed-version clusters.

### DIFF
--- a/deps/oauth2_client/include/types.hrl
+++ b/deps/oauth2_client/include/types.hrl
@@ -28,7 +28,8 @@
     authorization_endpoint :: option(uri_string:uri_string()),
     end_session_endpoint :: option(uri_string:uri_string()),
     jwks_uri :: option(uri_string:uri_string()),
-    ssl_options :: option(list())
+    ssl_options :: option(list()),
+    discovery_options = [] :: list()
 }).
 
 -type query_list() :: [{unicode:chardata(), unicode:chardata() | true}].
--- a/deps/oauth2_client/src/oauth2_client.erl
+++ b/deps/oauth2_client/src/oauth2_client.erl
@@ -8,7 +8,9 @@
 -export([get_access_token/2, get_expiration_time/1,
         refresh_access_token/2,
         get_oauth_provider/1, get_oauth_provider/2,
-        get_openid_configuration/2,
+        get_openid_configuration/3,
+        validate_openid_configuration/3,
+        validate_issuer/2,
         build_openid_discovery_endpoint/3,
         merge_openid_configuration/2,
         merge_oauth_provider/2,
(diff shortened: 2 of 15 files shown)


michaelklishin:
Note: UAA is a non-compliant IDP w.r.t. these changes, so it will need an extra configuration key added to `rabbitmq.conf` to disable this new verification.

michaelklishin (approved):

michaelklishin:
Because of the upgrade step above for the UAA users, we have agree to keep this change specific to the `4.4.0` release.

PR #17367, opened 2026-09-10
Fix incomplete case clause in SAC coordinator state_enter/2

kjnilsson:
A disconnected consumer's elapsed disconnection time between 10 and 60 seconds (the default disconnected timeout) had no matching case clause, crashing the Ra state machine whenever a node became the new leader while such a consumer was present.


lhoguin (approved):

michaelklishin:
I have investigated the stream-related failures that also happened on https://github.com/rabbitmq/rabbitmq-server/pull/17373. Will fix three tests there and merge, then rebase this branch.

https://github.com/rabbitmq/rabbitmq-server/pull/17373/changes/51f42c953c16ecd7af0bdeb40ee586138097beec is the most relevant part.

PR #17372, opened 2026-09-10
Redact TLS/PSK secrets from ssl_opts in management API responses

lhoguin:
Redact cert, key, cacerts, certs_keys, password, sni_hosts, stateless_tickets_seed, verify_fun and user_lookup_fun to the atom '...' before formatting, keeping the key so operators can still see that the option is configured. Mirrors the approach already used by ranch_acceptors_sup:hide_socket_opts/1 for the same class of option.


michaelklishin:
The force push was to simplify and shorted a commit message, and remove tooling spam in the commit footer.

michaelklishin (approved):

PR #17375, opened 2026-09-10
Shovel management: stop guessing delete vs restart from the request path

lhoguin:


michaelklishin:
The force push was a rebase, my drive-by flake fixes remain as separate commits.

PR #17391, opened 2026-09-11
amqp_client: enhance frame size and boundary validation

acogoluegnes:
Update the AMQP 0-9-1 client reader to strictly enforce frame_max limits and validate frame end markers earlier in the processing pipeline.

Previously, the client relied on broader protocol handling to manage incoming frame boundaries. This change introduces an early check in amqp_main_reader to reject oversized frames immediately upon reading the header, improving connection teardown behavior when encountering non-compliant streams.

Additionally, the reader now explicitly validates the frame end marker during payload accumulation to ensure strict protocol compliance.

Changes:
* Define a sensible default ceiling during the handshake phase.
* Pass the negotiated frame_max from the connection to the reader.
* Immediately reject inbound frames that exceed the negotiated limit.
* Log and close connections when an invalid frame end marker is detected.

### DIFF
--- a/deps/amqp_client/include/amqp_client_internal.hrl
+++ b/deps/amqp_client/include/amqp_client_internal.hrl
@@ -26,3 +26,8 @@
 
 -define(DIRECT_OPERATION_TIMEOUT,  120000).
 -define(CALL_TIMEOUT_DEVIATION,    10000).
+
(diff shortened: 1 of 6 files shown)


michaelklishin:
The forced push was a rebase on top of `main` plus one small commit that adds a test with more variance in (sequentially published) payload sizes.

michaelklishin (approved):

PR #17394, opened 2026-09-11
Filter cross-vhost queue names out of quorum health checks

kjnilsson:
`GET /health/checks/node-is-quorum-critical` and
`/health/checks/quorum-queues-without-elected-leaders/all-vhosts/` returned queue names and vhosts cluster-wide to any management-tagged user, ignoring that user's per-vhost permissions. Filter the queue list through `rabbit_mgmt_util:filter_vhost/3`, the same mechanism `/api/queues` uses.

filter_vhost/3 expects atom-keyed vhost entries, but these two endpoints return maps keyed by the binary <<"virtual_host">>, so results are tagged before filtering and untagged after. Cluster-wide critical components (rabbit_stream_coordinator, rabbitmq_metadata) aren't vhost-scoped and pass through unfiltered.

### DIFF
--- a/deps/rabbitmq_management/src/rabbit_mgmt_wm_health_check_node_is_quorum_critical.erl
+++ b/deps/rabbitmq_management/src/rabbit_mgmt_wm_health_check_node_is_quorum_critical.erl
@@ -60,15 +60,7 @@ failure(Message, Qs, ReqData, Context) ->
 is_authorized(ReqData, Context) ->
     rabbit_mgmt_util:is_authorized(ReqData, Context).
 
-%% Qs are #{binary() => any()} maps keyed by <<"virtual_host">>, not the
-%% atom-keyed proplists filter_vhost/3 expects, so tag and untag around it.
-%% Critical components (rabbit_stream_coordinator, rabbitmq_metadata) are
-%% cluster-wide, not vhost-scoped, and must pass through unfiltered rather
-%% than be dropped for everyone. They are marked with `type => process`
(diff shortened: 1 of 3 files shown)


deps/rabbitmq_management/src/rabbit_mgmt_wm_health_check_node_is_quorum_critical.erl
```diff
@@ -52,3 +59,20 @@ failure(Message, Qs, ReqData, Context) ->
...
+
+%% Qs are #{binary() => any()} maps keyed by <<"virtual_host">>, not the
+%% atom-keyed proplists filter_vhost/3 expects, so tag and untag around it.
+%% Critical components (rabbit_stream_coordinator, rabbitmq_metadata) are
+%% cluster-wide, not vhost-scoped, and must pass through unfiltered rather
+%% than be dropped for everyone. They are marked with `type => process`
+%% (rabbit_upgrade_preparation:list_with_minimum_quorum_for_cli/0); vhost
+%% names are arbitrary binaries, so matching on the display-only
+%% virtual_host value of "(not applicable)" could be spoofed by a real
+%% vhost of that name.
```
lhoguin: PR is fine and this comment is accurate but perhaps lines 3+ of the comment could be summarized to "Critical components of type process are never filtered."

michaelklishin:
The forced push was a rebase plus one commit that addressed the comment feedback.

michaelklishin (approved):

PR #17398, opened 2026-09-11
Harden exchange federation binding propagation against malformed input

acogoluegnes:
The hop counter carried in binding arguments during federation binding propagation is client-controlled. Compute it with a pure, non-negative function instead of trusting the raw value, and validate `max-hops` as a bounded positive integer, so malformed or out-of-range input can no longer produce unbounded propagation or crash the link.


michaelklishin:

michaelklishin (approved):

PR #17405, opened 2026-09-11
DO NOT MERGE Guard policy re-evaluation against concurrent queue/exchange writes

lhoguin:
rabbit_db_policy:update/3 decided each queue/exchange's new policy from a snapshot read before its own transaction, then applied that decision inside the transaction with no check that the record hadn't changed since. Two policy changes racing on the same vhost could interleave so that the loser's transaction, built from a stale snapshot, silently overwrote whatever the winner's transaction had just committed.

Add update_in_khepri_tx/3 to rabbit_db_queue and rabbit_db_exchange, guarding the write with the payload_version read alongside the snapshot and aborting the transaction via khepri_tx:abort/1 on a mismatch (needed for the abort to actually discard puts already made in the same attempt, rather than let khepri commit them regardless). update/3 retries the whole operation from scratch on that abort, so the retry sees the concurrent write instead of overwriting it.

### DIFF
--- a/deps/rabbit/src/rabbit_db_exchange.erl
+++ b/deps/rabbit/src/rabbit_db_exchange.erl
@@ -40,6 +40,7 @@
          delete_in_khepri/3,
          get_in_khepri_tx/1,
          update_in_khepri_tx/2,
+         update_in_khepri_tx/3,
          clear_exchanges_in_khepri/0,
          clear_exchange_serials_in_khepri/0,
          put_options/1
@@ -259,6 +260,42 @@ update_in_khepri_tx(Name, Fun) ->
             X1 = Fun(X),
             ok = khepri_tx:put(Path, X1),
             X1;
+        _ -> not_found
+    end.
+
+%% -------------------------------------------------------------------
+%% update_in_khepri_tx().
+%% -------------------------------------------------------------------
+
+-spec update_in_khepri_tx(ExchangeName, Vsn, UpdateFun) -> Ret when
+      ExchangeName :: rabbit_exchange:name(),
+      Vsn :: khepri:payload_version(),
+      Exchange :: rabbit_types:exchange(),
+      UpdateFun :: fun((Exchange) -> Exchange),
+      Ret :: not_found | Exchange.
+%% @doc Same as `update_in_khepri_tx/2', but only applies `UpdateFun' if
+%% the record hasn't changed since the version `Vsn' was read, aborting
+%% the enclosing transaction with `mismatching_node' otherwise. Use this
+%% when `Vsn' was read outside of the transaction, e.g. to decide
+%% `UpdateFun' itself from data that could since have gone stale.
+
+update_in_khepri_tx(Name, Vsn, Fun) ->
+    Path = khepri_exchange_path(Name),
+    case khepri_tx:get(Path) of
+        {ok, X} ->
+            X1 = Fun(X),
+            UpdatePath = khepri_path:combine_with_conditions(
+                           Path, [#if_payload_version{version = Vsn}]),
+            %% A mismatching_node condition failure must abort the whole
+            %% enclosing transaction, discarding any puts already made
+            %% during this attempt: khepri_tx:abort/1 does this via an
+            %% exception, unlike returning the failure as an ordinary
+            %% {error, _} value, which khepri would otherwise still
+            %% commit up to this point.
+            case khepri_tx:put(UpdatePath, X1) of
+                ok -> X1;
+                {error, Reason} -> khepri_tx:abort(Reason)
+            end;
         _ -> not_found
     end.
 
--- a/deps/rabbit/src/rabbit_db_policy.erl
+++ b/deps/rabbit/src/rabbit_db_policy.erl
@@ -7,6 +7,7 @@
 
 -module(rabbit_db_policy).
 
+-include_lib("khepri/include/khepri.hrl").
 -include_lib("rabbit_common/include/rabbit.hrl").
 -include("amqqueue.hrl").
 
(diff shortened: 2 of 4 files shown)


deps/rabbit/src/rabbit_db_exchange.erl
```diff
@@ -262,6 +263,42 @@ update_in_khepri_tx(Name, Fun) ->
...
+%% when `Vsn' was read outside of the transaction, e.g. to decide
+%% `UpdateFun' itself from data that could since have gone stale.
+
+update_in_khepri_tx(Name, Vsn, Fun) ->
+    Path = khepri_exchange_path(Name),
+    case khepri_tx:get(Path) of
+        {ok, X} ->
+            X1 = Fun(X),
+            UpdatePath = khepri_path:combine_with_conditions(
+                           Path, [#if_payload_version{version = Vsn}]),
```
dumbbell, line 288: This can be done by the caller which then calls `update_in_khepri_tx/2`. This avoids adding two functions which are used once each.

`update_in_khepri_tx/2` can be updated to abort the transaction instead of raising a `badmatch`.

deps/rabbit/src/rabbit_db_exchange.erl
```diff
@@ -262,6 +263,42 @@ update_in_khepri_tx(Name, Fun) ->
...
+            %% A mismatching_node condition failure must abort the whole
+            %% enclosing transaction, discarding any puts already made
+            %% during this attempt: khepri_tx:abort/1 does this via an
+            %% exception, unlike returning the failure as an ordinary
+            %% {error, _} value, which khepri would otherwise still
+            %% commit up to this point.
```
dumbbell, line 294: There is no need for useless AI comments :-)

deps/rabbit/src/rabbit_db_policy.erl
```diff
@@ -26,30 +27,81 @@
...
+                      Map, ExchangeVsns,
+                      fun rabbit_db_exchange:update_in_khepri_tx/3)
+                    || Map <- Exchanges, is_map(Map)],
+                   [update_queue_policies(
+                      Map, QueueVsns,
+                      fun rabbit_db_queue:update_in_khepri_tx/3)
+                    || Map <- Queues, is_map(Map)]}
+          end, rw)
+    catch
+        throw:{error, {khepri, mismatching_node, _}} ->
```
dumbbell, line 88: Please use `?khepri_error()`.

dumbbell (changes requested):

lhoguin:
@dumbbell The code has been updated. Ignore AI comments for now and please tell me if this is more in line with what you said. I'll do the cleanup afterwards.

deps/rabbit/src/rabbit_db_policy.erl
```diff
@@ -26,33 +27,108 @@
...
         #exchange{} = X1 -> {X, X1};
         not_found        -> {X, X }
     end.
 
-update_queue_policies(#{queue := Q0, update_function := UpdateFun}, StoreFun) ->
+update_queue_policies(#{queue := Q0, update_function := UpdateFun}, Vsns) ->
     QName = amqqueue:get_name(Q0),
-    NewQueue = StoreFun(QName, UpdateFun),
+    Vsn = maps:get(QName, Vsns),
+    assert_unchanged(rabbit_db_queue:khepri_queue_path(QName), Vsn),
```
dumbbell, line 96: This is already handled by conditions in a path, why not use them directly here? `assert_unchanged()` reimplements the logic already present in Khepri.

This would also avoid a second traversal of the tree.

dumbbell (changes requested):

deps/rabbit/src/rabbit_db_policy.erl
lhoguin: Good question — and worth checking empirically before answering, since I already got burned once assuming Khepri's read-side condition behavior. I just verified directly (in a throwaway CT test, not committed): if you attach #if_payload_version{} to the path and pass it into khepri_tx:get/1 (the plain get that update_in_khepri_tx/2 uses), a version mismatch doesn't return not_found or an error — it crashes with {badmatch, []} inside khepri_utils:get_single_node_props/1. khepri_tx:get/2 unwraps its result assuming exactly one match; a failed condition makes the underlying get_many come back with an empty map instead of an error, and the unwrap blows up on that.

So the reviewer is right that assert_unchanged/2 reimplements logic Khepri already has, but that logic only works correctly on the write side. I confirmed this in the previous round too: a condition on a put correctly yields {error, mismatching_node} when it fails, but the identical condition on a get just silently matches nothing — no error, no signal to abort on. That asymmetry is exactly why I ended up hand-rolling the version comparison in Erlang in the first place, instead of just matching on the condition's outcome.

Given that, the actual way to "use conditions directly" without reimplementing anything and without a second traversal is to put the condition on the write, using a Vsn supplied by the caller — which is precisely what update_in_khepri_tx/3 did before I removed it last round. One read (unconditioned, to get X for Fun), one conditioned write (Khepri itself decides mismatch vs. success), no manual comparison, no second traversal.

That leaves the concern from last round — "don't add a function used only once" — but revisiting it: update_in_khepri_tx/2 has zero callers anywhere in the tree, including now. Nobody actually wants the version-unaware form; rabbit_db_policy is the only caller and it always needs the guard. So I'd propose collapsing back to a single function per module — call it update_in_khepri_tx/3 again, taking (Name, Vsn, Fun) — and deleting both /2 (unused) and assert_unchanged/2 (duplicate, and the source of last round's two bugs) entirely. That's one function, doing one read and one conditioned write, with the mismatch check delegated to Khepri's own write-path condition handling rather than hand-rolled.

Does that match what you had in mind, or would you rather keep the two-call shape in rabbit_db_policy.erl and just have it build the conditioned write path itself (inlining /3's body there instead of in rabbit_db_exchange/rabbit_db_queue)?

dumbbell:
I didn’t read the AI answer but trying to implement what I had in mind revealed a bug in Khepri. This is now fixed (but pending a release).

Here is an alternative patch to fix the race condition:
https://github.com/rabbitmq/rabbitmq-server/pull/15332/changes/ce0deb3f107a46aa94316ad54e6071a8853bcd41

It reuses the testcases from this pull request.

lhoguin:
@dumbbell Your fix is accurate. Please open a PR.

lhoguin:
The PR is now only JSP's commit rebased to most recent main. JSP's commit already includes the tests Claude generated.

dumbbell (approved):

PR #17406, opened 2026-09-11
Redact secrets from rabbit_env debug logging

lhoguin:
log_process_env/0, log_context/1, get_used_env_vars/0, and the raw $RABBITMQ_CONF_ENV_FILE dump logged RABBITMQ_DEFAULT_PASS and RABBITMQ_ERLANG_COOKIE in the clear.

log_process_env/0 now only logs the value of variables rabbit_env actually uses (per ?USED_ENV_VARS); other variables log their name only. Known secrets are redacted by exact name, since the set of used variables and context keys is small and known. The raw shell dump in post_port_cmd_output/3 sees the whole inherited process environment, not just RabbitMQ's own variables, so it keeps matching by substring instead of an exact list.


deps/rabbit_common/src/rabbit_env.erl
```diff
@@ -30,9 +30,14 @@
...
--export([parse_conf_env_file_output2/2,
+-export([parse_conf_env_file_output1/2,
+         parse_conf_env_file_output2/2,
          value_is_yes/1,
-         parse_conf_env_file_output_win32/2]).
+         parse_conf_env_file_output_win32/2,
+         post_port_cmd_output/3,
+         redact/2,
+         unicode_characters_to_list/1,
+         used_env_vars/0]).
```
lhoguin, line 40: I'll see if we can reduce the number of test exports.

deps/rabbit_common/src/rabbit_env.erl
```diff
@@ -249,26 +254,61 @@ has_var_been_overridden(#{var_origins := Origins}, Var) ->
...
 
 log_process_env() ->
     ?LOG_DEBUG("Process environment:"),
     lists:foreach(
       fun({Var, Value}) ->
-              ?LOG_DEBUG("  - ~ts = ~ts", [Var, Value])
+              %% `var_is_used/1` only knows about variables rabbit_env
+              %% itself reads; a variable it doesn't recognize may
+              %% still be consumed by the launch scripts before the
+              %% VM starts, so "unused" would overclaim.
```
lhoguin, line 267: OK so we need a list that includes those.

deps/rabbit_common/src/rabbit_env.erl
```diff
@@ -249,26 +254,61 @@ has_var_been_overridden(#{var_origins := Origins}, Var) ->
...
+
+redact(_NameOrKey, undefined) ->
+    %% Keep unset distinguishable from a redacted value.
+    undefined;
+redact(NameOrKey, Value) ->
+    case lists:member(NameOrKey, ?MASKED_VARS) of
+        true ->
+            "******";
+        false ->
+            case lists:member(NameOrKey, ?HASHED_VARS) of
```
lhoguin, line 299: Instead of two lists we could have a map saying "mask" or "hash".

deps/rabbit_common/src/rabbit_env.erl
```diff
@@ -249,26 +254,61 @@ has_var_been_overridden(#{var_origins := Origins}, Var) ->
...
+                false -> Value
+            end
+    end.
+
+hash_secret(Value) when is_atom(Value) ->
+    hash_secret(atom_to_list(Value));
+hash_secret(Value) when is_list(Value) ->
+    %% md5/1 needs Latin-1 bytes; UTF-8-encode first.
+    base64:encode_to_string(
+      erlang:md5(unicode:characters_to_binary(Value))).
```
lhoguin, line 310: Should reuse the function we already have.

deps/rabbit_common/src/rabbit_env.erl
```diff
@@ -1746,10 +1786,57 @@ post_port_cmd_output(#{os_type := {OSType, _}}, UnicodeOutput, ExitStatus) ->
...
+                    case is_sensitive_sh_var(Var) of
+                        true  -> VarPart ++ "=******";
+                        false -> Line
+                    end;
+                [] ->
+                    Line
+            end;
+        _ ->
+            Line
+    end.
```
lhoguin, line 1832: I need to double check the usefulness of this.

deps/rabbit_common/src/rabbit_env.erl
```diff
@@ -1793,10 +1880,12 @@ parse_conf_env_file_output_win32([Line | Lines], Vars) ->
...
         _ ->
-            %% Parsing failed somehow.
+            %% No `=`: could be a continuation line of a secret; log
+            %% length only.
             ?LOG_WARNING(
-               "Failed to parse $RABBITMQ_CONF_ENV_FILE output line: ~tp",
-               [Line],
+               "Failed to parse $RABBITMQ_CONF_ENV_FILE output line "
+               "(~b characters)",
+               [string:length(Line)],
```
lhoguin, line 1888: Should be line number instead.

deps/rabbit_common/src/rabbit_env.erl
```diff
@@ -1818,10 +1907,12 @@ parse_conf_env_file_output2([Line | Lines], Vars) ->
...
                 _ ->
-                    %% Parsing failed somehow.
+                    %% No `=`: could be a continuation line of a
+                    %% secret; log length only.
                     ?LOG_WARNING(
-                       "Failed to parse $RABBITMQ_CONF_ENV_FILE output: ~tp",
-                       [Line],
+                       "Failed to parse $RABBITMQ_CONF_ENV_FILE output "
+                       "(~b characters)",
+                       [string:length(Line)],
```
lhoguin, line 1915: Should be line number instead.

deps/rabbit_common/src/rabbit_env.erl
```diff
@@ -2162,6 +2261,6 @@ unicode_characters_to_list(Input) ->
             String
     end.
 
-log_characters_to_list_error(Input, Partial, Rest) ->
-    ?LOG_ERROR("error converting '~tp' to unicode string "
-                     "(partial '~tp', rest '~tp')", [Input, Partial, Rest]).
+log_characters_to_list_error(_Input, _Partial, _Rest) ->
+    %% A caller's input can be a secret; never log it.
+    ?LOG_ERROR("error converting input to unicode string", []).
```
lhoguin, line 2266: We can do better than that.

lhoguin (commented):
Lots left to do.

deps/rabbit_common/src/rabbit_env.erl
```diff
@@ -249,26 +257,69 @@ has_var_been_overridden(#{var_origins := Origins}, Var) ->
...
+              %% whether rabbit_env itself reads it: it may still be
+              %% consumed by the launch scripts or the CLI before the
+              %% VM starts, so "unused" would overclaim. A handful of
+              %% non-prefixed variables (see `?USED_ENV_VARS`) are
+              %% used too.
+              case var_is_used(Var) of
+                  true ->
+                      ?LOG_DEBUG("  - ~ts = ~ts", [Var, redact(Var, Value)]);
+                  false ->
+                      ?LOG_DEBUG("  - ~ts (unused by rabbit_env)", [Var])
```
lhoguin: Should just say "unused"

deps/rabbit_common/src/rabbit_env.erl
lhoguin: We don't have one easily usable.

lhoguin (commented):
AI comments need removing. Commits need squashing.

michaelklishin (approved):

PR #17408, opened 2026-09-11
Local shovel: enforce read/write ACLs on source/destination

dcorbacho:
The local shovel protocol drives rabbit_queue_type directly instead of going through a real channel, so it never checked whether the URI user had read access to the source queue or write access to the destination before consuming/delivering as the internal rmq-shovel identity.


michaelklishin:

michaelklishin (approved):

michaelklishin:
@mergifyio backport `v4.3.x`

PR #17421, opened 2026-09-11
A small follow-up top #17396 (test improvements)

michaelklishin:
A small follow-up to #17396 whose push failed without me noticing.


michaelklishin:
Will backport via a new `v4.3.x` backport PR for https://github.com/rabbitmq/rabbitmq-server/pull/17396.

PR #17449, opened 2026-09-14
Queue federation: tolerate multiple x-received-from matches

dcorbacho:
A publisher could duplicate hop entries, avoid a link crash on redelivery


michaelklishin (approved):

michaelklishin:
@mergifyio backport v4.3.x

PR #17450, opened 2026-09-14
Improve temporary peer discovery node

lhoguin:
* Disable the distribution listener entirely:
  this ensures that no other node can connect
  to the temporary node.

* Pass the node cookie via peer:call so that
  it doesn't appear when using `ps`.


lhoguin:
Change looks fine, comment cleanup needed.

lhoguin:
Marked it for backport but it's fine to keep at 4.4+ tbh.

michaelklishin (approved):

PR #17466, opened 2026-09-15
Restrict AMQP URI auth_mechanism to a fixed mechanism table

acogoluegnes:
Map auth_mechanism URI values to a fixed table of funs instead of resolving an arbitrary Module:Function pair via list_to_existing_atom/1. This also makes URI parsing independent of which modules happen to be loaded, so a bad value is rejected at parse time instead of failing later with undef when the connection is established.

The colon-separated form is kept only for amqp_auth_mechanisms, since that is the only module ever referenced this way in this repository and is asserted by an existing test.


lhoguin (approved):
Approval from both me and my security scan prompt.

michaelklishin (approved):

michaelklishin:
The forced push was a rebase on top of `main` plus conflict resolution in `4.4.0.md`.

PR #17487, opened 2026-09-16
Gate atom->binary global param key on FF

mkuratczyk:
Follow up to: 39d93b6adb03f6bbc5641ccbcc21b5cf773f8cf0

The original implementation was not safe in a mixed-version cluster. This commit keeps the old behaviour (using atoms) until rabbitmq_4.4.0 feature flag is enabled


lhoguin (approved):
Both me and Claude agree with this change.

PR #17491, opened 2026-09-16
GET->POST in Management API endpoint

mkuratczyk:
Don't allow GET, since arguments will leak through access logs
Add POST support, to provide the same functionality

When GET is issued, a hint about using POST is returned.


deps/rabbitmq_management/src/rabbit_mgmt_wm_hash_password.erl
```diff
@@ -20,15 +21,36 @@ variances(Req, Context) ->
...
 
 allowed_methods(ReqData, Context) ->
-    {[<<"GET">>, <<"OPTIONS">>], ReqData, Context}.
+    {[<<"GET">>, <<"POST">>, <<"OPTIONS">>], ReqData, Context}.
 
 content_types_provided(ReqData, Context) ->
     {rabbit_mgmt_util:responder_map(to_json), ReqData, Context}.
 
+content_types_accepted(ReqData, Context) ->
+    {[{'*', accept_content}], ReqData, Context}.
```
lhoguin: Probably should be <<"application/json">> instead of anything (`'*'`).

lhoguin (approved):

PR #17492, opened 2026-09-16
DO NOT MERGE 5.0 Proposal: Drop the bare unprefixed fallback for RABBITMQ_ env vars

lhoguin:
Because of the nature of the change and the uncertainty around Windows behavior I am marking it for 5.0.


lhoguin:
I'll come back to this later, I can't get agents to agree. I will likely need to manually check a lot of things especially on Windows.

lukebakken:
@lhoguin feel free to hit me up for the Windows testing. I'll keep an eye on this.

PR #17498, opened 2026-09-16
Use binary for user tags

lhoguin:


michaelklishin:
@lhoguin I agree that this is a good idea. There are no reasons for using atoms for tags, it's just a historical artifact.

Note that we have limited the max number of tags back in April or so.

mkuratczyk (approved):

michaelklishin (approved):

michaelklishin:
@lhoguin @mkuratczyk not for this PR but perhaps we should introduce a new module with role predicates. Now that matching on one means matching on either an atom or a binary… even if we assume that in most cases, the read-time coercion to binaries would help.

Even something as simple as `is_tagged_with/2`, if we don't yet have it in the internal user record module.

michaelklishin:
The PR piggybacks on the `4.4.0` umbrella feature flag => we should not backport it.

PR #17501, opened 2026-09-17
Undeprecate RABBITMQ_ERLANG_COOKIE

lhoguin:
This is a safer way to override the cookie than the flag.


lhoguin:
Marked do not backport because we likely want to announce it in release notes (the deprecation was announced in 3.8.6). And since it's not a functional change there's not much point backporting it.

PR #17502, opened 2026-09-17
Use `gen_secure/0` when generaiting the Direct Reply-To pseudo-queue suffix

kjnilsson:
rabbit_volatile_queue:new_name/0 generated the random suffix of the "amq.rabbitmq.reply-to.*" pseudo-queue name with rabbit_guid:gen/0, which is documented as predictable after the first call. That suffix is a capability: anyone who obtains it can consume from the pseudo-queue, so it needs the non-predictable rabbit_guid:gen_secure/0, which the consumer tag generated in the same code path already uses.


michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/pull/17599

PR #17503, opened 2026-09-17
CI: Sha256 the ecs-cli binary before use

lhoguin:


lhoguin:
The AWS failure is expected. The change to the workflow happens before the failure and can be double checked.

michaelklishin (approved):

PR #17512, opened 2026-09-18
Cleanup pre-3.13 code

lhoguin:


lhoguin:
I would keep this 4.4. The diff is not big and unlikely to cause conflicts. The impact is low.

michaelklishin (approved):

PR #17515, opened 2026-09-18
Mqtt QoS0 queue cleanup

mkuratczyk:
Mqtt QoS0 queue cleanup

Repro steps:
1. `make start-cluster`
2. `omq mqtt --uri mqtt://localhost:1883/,mqtt://localhost:1884/,mqtt://localhost:1885/ -x 0 -y 333`
3. `kill -STOP ...` - pause the metadata store leader
4. kill omq
5. `kill -CONT ...` - revive the frozen node

Without this PR, there'll likely be some stale queue entries.


deps/rabbitmq_mqtt/src/rabbit_mqtt_qos0_queue_cleanup.erl
```diff
@@ -0,0 +1,104 @@
...
+    process_flag(trap_exit, true),
+    {ok, #state{}}.
+
+handle_call(Request, From, State) ->
+    {stop, {unexpected_call, Request, From}, State}.
+
+handle_cast({retry_delete, Q},
+            State0 = #state{pending = Pending0}) ->
+    Key = {amqqueue:get_name(Q), amqqueue:get_exclusive_owner(Q)},
+    Pending = [{Key, Q} | Pending0],
```
dumbbell: You don’t need to store this tuple in the `Pending` list: you can compute it whenever you need it when a message is logged.

deps/rabbitmq_mqtt/src/rabbit_mqtt_qos0_queue_cleanup.erl
```diff
@@ -0,0 +1,104 @@
...
+
+handle_info(retry_delete, State0 = #state{pending = Pending0}) ->
+    case Pending0 of
+        [] ->
+            {noreply, State0};
+        [{Key, Q} | Rest] ->
+            case rabbit_queue_type:delete(Q, false, false, ?INTERNAL_USER) of
+                {error, timeout} ->
+                    {noreply, maybe_wait_for_khepri(State0)};
+                _ ->
```
dumbbell: I’m not sure about this catch-all case. Should a success be explicitly considered here before the queue is deleted from the `Pending` list?

Also, I would be optimistic here and put the success case first.

dumbbell (changes requested):

deps/rabbitmq_mqtt/src/rabbit_mqtt_processor.erl
```diff
@@ -1932,7 +1932,12 @@ maybe_delete_mqtt_qos0_queue(
             case {amqqueue:get_type(Q), amqqueue:get_pid(Q)} of
                 {?QUEUE_TYPE_QOS_0, Pid}
                   when Pid =:= self() ->
-                    rabbit_queue_type:delete(Q, false, false, Username);
+                    case rabbit_queue_type:delete(Q, false, false, Username) of
+                        {error, timeout} ->
+                            rabbit_mqtt_qos0_queue_cleanup:retry_delete(Q);
+                        _ ->
+                            ok
```
dumbbell: Perhaps the success return should be handled like in the cleanup process instead of having a catch-all clause.

dumbbell (approved):
Thank you!

michaelklishin (approved):

deps/rabbitmq_mqtt/src/rabbit_mqtt_qos0_queue_cleanup.erl
```diff
@@ -0,0 +1,102 @@
...
+handle_info(_Info, State) ->
+    {noreply, State}.
+
+%% Best-effort: our in-memory `pending' state is lost once we terminate,
+%% so give every pending delete one last try before that happens.
+terminate(_Reason, #state{pending = Pending}) ->
+    lists:foreach(
+      fun(Q) ->
+              _ = rabbit_queue_type:delete(Q, false, false, ?INTERNAL_USER)
+      end, Pending),
```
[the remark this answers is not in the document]
michaelklishin: This is a good find.

deps/rabbitmq_mqtt/src/rabbit_mqtt_sup.erl
```diff
@@ -48,6 +48,13 @@ init([{Listeners, SslListeners0}]) ->
          restart => transient,
          shutdown => ?SUPERVISOR_WAIT,
          type => supervisor
+        },
+       #{
+         id => rabbit_mqtt_qos0_queue_cleanup,
+         start => {rabbit_mqtt_qos0_queue_cleanup, start_link, []},
+         restart => permanent,
```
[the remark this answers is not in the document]
michaelklishin: Fair enough. This can be addressed by doing a bit more work in `init/1`.

deps/rabbitmq_mqtt/src/rabbit_mqtt_qos0_queue_cleanup.erl
```diff
@@ -0,0 +1,101 @@
...
+    gen_server:cast(?MODULE, {retry_delete, Q}).
+
+init([]) ->
+    process_flag(trap_exit, true),
+    Stale = [Q || Q <- rabbit_db_queue:get_all_durable_by_type(rabbit_mqtt_qos0_queue),
+                  Pid <- [amqqueue:get_pid(Q)],
+                  node(Pid) =:= node(),
+                  not is_process_alive(Pid)],
+    self() ! retry_delete,
+    {ok, #state{pending = Stale}}.
```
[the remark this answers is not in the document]
michaelklishin: Yes it is, in practice.

michaelklishin:
@copilot is stuck on the same incorrect observation about the `case` clauses.

I won't request any more reviews because that "high severity" issue does not check out.

I have addressed a couple of small items.

michaelklishin:
@copilot ignore a false positive you keep bringing up in this and all futures reviews https://github.com/rabbitmq/rabbitmq-server/pull/17515#discussion_r4067944506

deps/rabbitmq_mqtt/src/rabbit_mqtt_qos0_queue_cleanup.erl
```diff
@@ -0,0 +1,102 @@
...
+%%
+%% Copyright (c) 2007-2026 Broadcom. All Rights Reserved. The term “Broadcom” refers to Broadcom Inc. and/or its subsidiaries. All rights reserved.
+%%
+
+-module(rabbit_mqtt_qos0_queue_cleanup).
+
+-behaviour(gen_server).
+
+-include_lib("rabbit_common/include/rabbit.hrl").
+-include_lib("rabbit/include/amqqueue.hrl").
```
ansd: Is this import used? My editor says "unused file".

ansd:
I think the change looks good.

(I was first worried that this new gen_server could delete "blindly" the QoS 0 queue of a new clean session/connection with the same client ID by only its queue name. But this worry doesn't hold up since the delete path already handles this correctly:
```erl
    Conditions = case amqqueue:get_exclusive_owner(Queue) of
                     none ->
                         [];
                     Owner ->
                         Pattern = amqqueue:pattern_match_on_exclusive_owner(Owner),
                         [#if_data_matches{pattern = Pattern}]
                 end,
```
)

dumbbell:
I had the same concern and was pleased to find out this was handled in a more generic place :-)

deps/rabbitmq_mqtt/src/rabbit_mqtt_qos0_queue_cleanup.erl
mkuratczyk: indeed. removed

PR #17532, opened 2026-09-20
Fix cleanup after remote projection timeout

### LINKED ISSUE #17531: Intermittent quorum queue residue after remote projection timeout during rolling upgrades
### Describe the bug

During rolling upgrades of a multi-node cluster with Khepri enabled, we intermittently observed a newly declared quorum queue become unusable. The initial declaration failed with AMQP 541 `INTERNAL_ERROR`; however, a later declaration of the same queue name and arguments returned success, while `basic.consume` failed with `noproc`.

The natural trigger is intermittent: it occurs while application topology recovery can declare queues during a node-by-node rolling upgrade, but I have not been able to make a healthy cluster naturally exceed the projection wait window on demand. To make the failure state deterministic, I reproduced it in an isolated three-node RabbitMQ 4.3.6 cluster by forcing the exact timeout exit used by `rabbit_queue_type_util:wait_for_remote_projection/3` for one test queue only.

The quorum queue creation flow persists queue metadata with `rabbit_amqqueue:internal_declare/2`, then waits for the queue projection to become visible on the remote leader, and only after that starts the Ra cluster. If the projection wait reaches its timeout exit, the exception escapes before `ra:start_cluster/3` is called.

The controlled reproduction injects the same exit tuple as the timeout branch for one test queue only:
```erlang
exit({timeout, wait_for_remote_projection, Node, QName}).
```
injected timeout makes the resulting state transition reproducible

After the first declaration failed with AMQP 541 `INTERNAL_ERROR`:

* `rabbit_amqqueue:lookup/1` still returned the queue metadata;
* the stored server ID pointed to the selected remote leader;
* `ra:members(ServerId)` returned `{error, noproc}`;
* a second declaration of the same queue with identical arguments returned success; and
* `basic.consume` failed with AMQP 541 and `noproc`.

`rabbitmq-diagnostics quorum_status` reported `noproc` for every expected Ra member, and `rabbitmq-diagnostics list_unresponsive_queues` listed the queue.

### Reproduction steps

1. Start a three-node RabbitMQ 4.3.6 cluster with Erlang/OTP 27 and enable the `khepri_db` feature flag.

2. Connect a client to Node A. Create several healthy durable quorum queues with `x-queue-leader-locator=client-local` so their leaders are on Node A. This makes a later queue declared with `x-queue-leader-locator=balanced` select a remote leader in this test setup.

3. Exercise a normal node-by-node rolling-maintenance step before declaring the target queue. This models the observed intermittent scenario in which application topology recovery can create a new quorum queue while a node is being upgraded:

   ```bash
   rabbitmq-upgrade -n rabbit_repro_b@localhost drain
   rabbitmqctl -n rabbit_repro_b@localhost shutdown
   rabbitmq-server -detached
   ```

   Wait for Node B to be running and no longer in maintenance mode. The node restart itself does not force the timeout; it establishes the rolling-upgrade context in which the issue was observed intermittently.

4. In an isolated test build, make only the target QName take the same exit that the current timeout branch of `wait_for_remote_projection/3` takes after retries are exhausted:

   ```erlang
   exit({timeout, wait_for_remote_projection, Node, QName}).
   ```

   Keep all other `wait_for_projection/2` calls unchanged.

5. From a client connected to Node A, declare the target queue:

   ```python
   channel.queue_declare(
       queue="repro-projection-timeout-20260920",
       durable=True,
       arguments={
           "x-queue-type": "quorum",
           "x-queue-leader-locator": "balanced",
       },
   )
   ```

   The first declaration fails with `ConnectionClosedByBroker: (541) 'INTERNAL_ERROR'`.

6. Inspect the broker state. The target queue is still returned by `rabbit_amqqueue:lookup/1`, but `ra:members(ServerId)` returns `{error, noproc}`.

7. Repeat the same `queue_declare`. It succeeds with zero messages and zero consumers. Then call `basic.consume` on the queue. The broker closes the connection with:

   ```text
   INTERNAL_ERROR - failed consuming from quorum queue 'repro-projection-timeout-20260920' in vhost '/': noproc
   ```

### Expected behavior

If `wait_for_remote_projection/3` times out while creating a new quorum queue, the operation should fail without leaving a usable-looking queue definition behind.

In particular, after the failed declaration:

* `rabbit_amqqueue:lookup(QName)` should return `not_found`, or the broker should otherwise complete a safe cleanup;
* no incomplete Ra server definition should remain;
* after the transient condition is removed, declaring the same queue name again should create a working quorum queue; and
* a successful `queue_declare` must not be returned for a queue whose Ra members were never started.

### Additional context

* RabbitMQ: 4.3.6, source commit `7a34a0caf4fa5cf4009cb934ea7a94dbe7e08095`
* Erlang/OTP: 27.3.4.17
* Metadata store: Khepri (`khepri_db` enabled)
* Cluster: three local nodes

Representative server-side log:

```text
{{timeout,wait_for_remote_projection,rabbit_repro_c@localhost,
  {resource,<<"/">>,queue,<<"repro-projection-timeout-20260920">>}}, ...}
```

Representative client results:

```text
first queue.declare:
ConnectionClosedByBroker: (541) 'INTERNAL_ERROR'

same-name queue.declare:
success, messages=0, consumers=0

basic.consume:
ConnectionClosedByBroker: (541) "INTERNAL_ERROR - failed consuming from quorum queue 'repro-projection-timeout-20260920' in vhost '/': noproc"

chenshi5012:
Fixes #17531.

When a quorum queue declaration selects a remote leader, a timeout while waiting for its Khepri projection can occur after the queue metadata has been written but before `ra:start_cluster/3` is called.

Previously, this `wait_for_remote_projection` timeout was raised outside the existing declaration failure handling path. As a result, declaration exited without removing the newly created queue metadata. A retry using the same queue name could then fail or observe an incomplete queue definition, even though no Ra cluster had been started.

This change routes the exact `{timeout, wait_for_remote_projection, LeaderNode, QName}` exit through the existing `declare_queue_error/4` path. That path removes the queue metadata via `rabbit_amqqueue:internal_delete/2` and returns a controlled AMQP `internal_error` to the client.

## Proposed Changes

- Move `rabbit_queue_type_util:wait_for_projection/2` inside the existing `try ... catch` block in `rabbit_quorum_queue:start_cluster/1`.
- Catch only the precise remote projection timeout exit:

  ```erlang
  {timeout, wait_for_remote_projection, LeaderNode, QName}
  ```

- Reuse the existing `declare_queue_error/4` cleanup path instead of adding a second cleanup implementation.
- Add a three-node regression test, `projection_timeout_cleanup`, that:
  1. creates seed quorum queues on the declaring node;
  2. uses the `balanced` leader locator so the target queue chooses a remote leader;
  3. injects a timeout only for the target call to `wait_for_projection/2`;
  4. verifies that declaration returns the expected `internal_error`;
  5. verifies that the queue metadata is removed;
  6. removes the injection and verifies that a queue with the same name can be declared, published to, and consumed from successfully.

The cleanup deliberately does **not** attempt to delete a Ra cluster because the injected failure occurs before `ra:start_cluster/3` is invoked.

## Types of Changes

*Put an `x` in the boxes that apply*

- [x] Bug fix (non-breaking change which fixes issue #17531)
- [ ] Build system and/or CI

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. This is simply a reminder of what we are going to look for before merging your code._

- [ ] **Mandatory**: I (or my employer/client) have signed the CA (see https://github.com/rabbitmq/cla)
- [x] Relevant tests pass locally with my changes

## Further Comments

The production symptom is intermittent because it depends on a remote Khepri projection not becoming visible within the wait window. The regression test uses a controlled timeout injection to make that exact boundary deterministic.

The change is intentionally narrow:

- It does not broaden exception handling to arbitrary `exit` reasons.
- It does not alter the successful declaration path.
- It does not introduce Ra cleanup for a cluster that has not started.
- It reuses the existing error-to-cleanup and protocol-error mapping already used when `ra:start_cluster/3` itself fails.


chenshi5012:
@ansd  Hi, this PR is ready for your review. Please take a look when you have time, thanks a lot!

michaelklishin:
Thank you. Moved to https://github.com/rabbitmq/rabbitmq-server/pull/17550.

PR #17534, opened 2026-09-21
DO NOT MERGE

dcorbacho:


lhoguin:
Second commit is unnecessary as OTP already does the right thing.

lhoguin:
Third commit: same.

lhoguin:
Fourth commit is also dropped. Some commits dropped here became OTP PRs instead.

lhoguin:
Sixth commit is in https://github.com/rabbitmq/rabbitmq-server/pull/17580 but it's defense-in-depth so will likely be deferred a bit.

lhoguin:
Seventh is in https://github.com/rabbitmq/rabbitmq-server/pull/17581

lhoguin:
Final commit in https://github.com/rabbitmq/rabbitmq-server/pull/17582

All commits have been addressed one way or another. Closing, thanks again!!

PR #17539, opened 2026-09-21
AMQP 1.0 client: add a configurable socket connect timeout

arrangabriel:
## Proposed Changes

I came across an issue when using the AMQP 1.0 client over TLS. If a server accepts the connection, but then does not complete the handshake, the connection attempt blocks permanently. 

Because the connection starts under `amqp10_client_sup` this blocks any clients in the application from starting at all.

There is a variant of the `ssl:connect` function that takes a timeout, but it is not currently used by the client library. This change adds a `connect_timeout` option to the connection config, which is then passed to the "transport layer".   
When the option is not provided, the default value preserves the existing behavior:
- For websocket based connections, 5000 ms, matching the default value of `gun:await_up`.
- For `gen_tcp` and `ssl` based connections, `infinity`.

Note that the timeout is applied twice by `ssl`, once for the TCP handshake, and again for the TLS handshake.

## Types of Changes

- [ ] Build system and/or CI

## Checklist

## Further Comments

I added a test that checks that the timeout is honored when attempting to connect to a server that accepts connections but does not answer.


michaelklishin:
Good catch. Moved to https://github.com/rabbitmq/rabbitmq-server/pull/17549.

PR #17561, opened 2026-09-22
Nit: Drop dead code in rabbit_auth_backend_ldap

lhoguin:
Unless mistaken since we set the value to `[]` in the Makefile's env configuration this can never be `undefined`.


michaelklishin:
We don't. Some might via `advanced.config`.

michaelklishin:
@lhoguin I suspect that the status quo may produce a Dialyzer warning. I've switched to a more verbose but more explicit and backwards-compatible alternative.

Unfortunately that `undefined` exists because we did see `function_clause`s in the wild.

michaelklishin:
It's probably a wash overall (we do not gain or lose anything) but maybe this version is a little bit more explicit about what the intent is.

lhoguin:
Fair enough. Note that the `{verify, verify_peer}` is the default in all supported versions now so it's not necessary to set it. But it's harmless.

